PR #576, opened 2015-04-01
Added a mysqli connection instead

a13henhu:


a13henhu:
Changed into PDO instead

Issue #496, opened 2015-03-29
Log analysis tools


HGustavs:
When/if we have comprehensive logging implented some kind of logging view would be great. Answering questions like
- who from the course have visited this page
- how many peaple have downloaded this file
- how many failed login attempts have each user made
- which operating systems are most common

a13chrju:
why not just use google analytics :)?

c13andka:
Should we do this ourself ?

b14vikgu:
Some of these have now been implemented on the analytic page, for example operating systems..

b17carbr:
@HGustavs do you know the status on this issue? I can't find any analytic page like the comments above mentions.

a97marbr:
I think that this feature has been forgotten ... BUT it is a great feature ... I would love to be able to get a feel for things like

* How many of the students registered actually visit the course page
* Is the traffic to the course only during scheduled supervisions
* Can we detect potential cut offs where students drop from courses 
* etc.

Give us a sketch on a prospect UI and we'll take it from there ... this could be really good!

a18filkr:
These are being worked on in sub-issues instead and som are already solved. Closing this since new issues are created.

PR #578, opened 2015-04-01
#496

a13chrju:
For more security and statistical purposes I guess. the amount of tries to login are displayed in a div underneath, if the number or tries exceeds 3, the user has to wait for 20 seconds before he can try again.


a13chrju:
"setTimeout(alertFunc, 3000); " is set to 3 seconds right now though, easy fix.

c13chrjo:
The counter count up instead of down

Issue #564, opened 2015-03-31
SQL code is ambiguous, few comments / no documentation


a13karsi:
The SQL code is undercommented and there don't seem to exist any documents specifying requirements, design and other concerns regarding the database.

HGustavs:
One or two rows of text comments for the code and some separate text documentation would be nice.

Perhaps creating a documentation folder?

a13karsi:
Good idea, I'll start with trying to add some sensible comments to the sql file

HGustavs:
The overarching goal of the project is readability.
The sql file should be understandable just by reading the file, but at the same time we do not want too much documentation in the code, so that the code becomes unreadable if it has more comments than code....

Even though... you can never have too much documentation... I suggest .html files in a documentation folder for the really detailed stuff, when the amount of documentation is starting to approach the amount of code.

a13karsi:
I agree, but I thought the database was kind of hard to intuitively approach. For the time being I've just added a few comments at the top of each table to try to make it easier next time I look at it. I'll see whether I can find more application for your documentation ideas as I keep working!

a13karsi:
Oh, I seem to have made a mistake when I made the pull request, the changes were actually merged into the master branch and not our weekly when the pull request was reviewed and accepted. I'll make sure to not let this happen again.

c13chrjo:
It´s equally my fault for not checking this before I merged it.

PR #586, opened 2015-04-01
#564

a13karsi:
Changed nothing functional, just added comments above tables in an effort to make the database easier to approach.


c13chrjo:
Everything seems to be in order.

PR #599, opened 2015-04-01
Created a teacherView in usermanagementview

c13emmjo:
the first was just a test in read but then in usermanagementview has been divided into two teacherView here. 
(and studentView in an another issue) 
this is just a shell and there are lots of thing to be done and change in this later


a13antda:
OK!

PR #605, opened 2015-04-01
Fixed #558

### LINKED ISSUE #558: New course is not visible directly
When you press the new button in the top right corner in the courser editor. The "page" should reload so that the new course you have created is visible for change directly. Right now you have to refresh the page before you can see the course you have added.

c13andka:
Get all courses after you have added new course, so you don't have to update site to see the newly added course.


c13chrjo:
Works

PR #607, opened 2015-04-01
Update editorService.php

a13henhu:


a13jansa:
insert PDO code and the orginal code. The problem seems to be solved.

Issue #614, opened 2015-04-01
Highlighting of parentheses on hover is so subtle that it is nearly invisible


b13gusbe:
Highlighting of parentheses on hover is so subtle that it is nearly invisible. Taken from issue #571.

PR #616, opened 2015-04-01
Update whiteTheme.css Issue #614

b13gusbe:
Changed background color from white to red.


c13caral:
Why did you make this change? I have added your edit to my folder to check exactly what changes to red background. But I can not find where the class ".hi" is used.

Issue #551, opened 2015-03-31
loginButton change question


b13danso:
Should we change the login box to something thats more clear to the user instead of having an icon representing a user with red/green background. It would be easier for the user to understand that the login button/logout button is placed on the top-right corner if the icon represented something like a door with an arrow pointing in(login) and a door with arrow pointing out(logout)?

HGustavs:
Im thinking that the icon is kind of nice, and a bit different style-wise (a good thing?).
Im also thinking that we want very clean and simple iconography. Make me a door icon, and we can compare it to the current one and have a "vote". 

What I am thinking is that perhaps, we can remove the yellow link which is kind of pointless, and replace it with something like an on-hover tooltip.

b13danso:
Icon idea approved by HGustavs, currently working on converting the created icon to SVG.

b13danso:
issue #618 is conntected to this issue.

b13danso:
Completed, pull request #620 created,

b13danso:
Fixed.

Issue #549, opened 2015-03-31
loginButton, navName width sizes wrong in header.


b13danso:
The width sizes of the two classes loginButton and navName in the table placed in header is wrong. The loginButton appears to be very wide in the webbrowser Chrome, Safari and Opera but correct in Internet Explorer and Firefox.
![fel1](https://cloud.githubusercontent.com/assets/11715018/6917618/7a823f1e-d7a6-11e4-94de-eb504828e244.png)

b13danso:
Fixed.

PR #617, opened 2015-04-01
fixed #563 - the position of the login button is now correctly displayed

### LINKED ISSUE #563: Login button shows incorrectly in chrome
The fault does not lie in CSS

a13chrju:
changed the name of one of the "navheader.php" files too while I was at it since it's confusing to work on two files at the sime time that has exactly the same name.


a13chrju:
never mind the name change of the file "navheader.php", just saw that some files still links to "navheader.php" instad of the new "navheader1.php". :)

b13danso:
This is really conflicting with issue #551. The changes made to the css(loggedout-class) is not relevent for this issue. I dont see the point in changing the <td> to <div>s and closing the <table> before the checklogin(). The misspositioning is a matter of css and it is an easy fix which is already stated in issue #549, this is a duplicate.

a13chrju:
I just didnt understand why they would use tables for positioning items in html therefore I changed it to divs and removed some tr-tags etc. :) but you're right, personally thought it was easier to see after my edit

c13chrjo:
Since this issue already exist and a fix has been made I will close this pull.

PR #620, opened 2015-04-01
#551

b13danso:
Added two new icons(login_button and logout_button), updated the css to add the missing hover effect on loggedout-class and changed the hover-effects. Updated navheader.php and added onmouseover/onmouseout events to loginButton.


a12oskri:
OK!

PR #621, opened 2015-04-01
fixed #566 - buttons reset the input fields

### LINKED ISSUE #566: Boxes in "Edit Course" should clear automatically

a13chrju:


a13chrju:
also, the inputs now have placeholders instead of actual text so you dont need to select the excisting text and delete it before you can enter an input value.

Issue #604, opened 2015-04-01
loginButton not clickable 


b13danso:
The loginButton on navheader.php  is not clickable on UMV, works in DuggaSys and CodeViewer.

b13danso:
Fixed.

PR #627, opened 2015-04-02
#604 - login button not clickable UMV

b13danso:
The login button was not clickable. The file(usermanagmentview.php) had a missing import of JavaScript file(dugga.js). The import was added to the head-tag and the login button is now working and the user is able to login/logout.


a13freas:
Looks good! Good job!

PR #636, opened 2015-04-02
#549

b13danso:
Removed the width from the class navName. This solved the issue and the correct width on loggedin/loggedout classes is viewed correctly.


a13freas:
Ok!

PR #638, opened 2015-04-02
"New"-button opens "New course"-window instead of creating empty course

b13alebr:
Update courseed.php
Added window för creating a new course

Update courseed.js 
Added functions to handle the new window for creating a new course. And the function to close window "edit course" renamed. The function newCourse changed so it opens the "new course"-window, and a new function createNewCourse added to save the new course in the database.

Update courseedservice.php
Funtionality to save newly created course is changed to include course name and course code when it is saved to the database.


c13andka:
It works, good job!

Issue #641, opened 2015-04-02
Buttonmenu2 change color when resizing


a12jiyba:
color changes from #5A5757  to #938C8C. it should always be #5A5757

PR #645, opened 2015-04-02
#641

a12jiyba:


a13jansa:
The header had two color before the changes. When you re-size the window to a mobile size the header will have the same color now. Problem solved!

Issue #632, opened 2015-04-02
improvments to make (Highlighting of parentheses on hover is so subtle that it is nearly invisible #614)


a12jiyba:
We should make he parentheses bold when hovering over and use a darker color.

c12jenek:
This issue was not closed and new similar issues were opened, closing this to clean up issues

PR #646, opened 2015-04-02
#632 New background-color in highlights

a13jansa:


a12jiyba:
its working now. good job with the bold thing.

Issue #606, opened 2015-04-01
Forward/Back buttons


a13henhu:
Make a functional forward and backward buttons..... WORK IN PROGRESS

Just a notification, if someone wants to jump in, go on, I won't deny any help

HGustavs:
I added a commit for this and a few unrelated bugs. Unfortunately I added it to master.... so it needs to be re-merged into the cv15 branch.

HGustavs:
Issue #667 is a superset of this issue. If that commit is proved to solve the issues, this issue can also be retired.

PR #674, opened 2015-04-07
Issue #606

a13henhu:
The fix should be applied for the navigation


a12jiyba:
Good to have the navigation problem fixed!

Issue #675, opened 2015-04-07
Database contains improper comments


a13karsi:
The database contains // comments that are improper sql syntax and cause problems when trying to build the latest version of the sql file.

a13karsi:
#677

c13chrjo:
Isseue is fixed and merged

PR #677, opened 2015-04-07
Issue #675 - Fixed faulty sql comments

a13karsi:
#675


c13chrjo:
The faulty comments have corretctlly been replaced.

Issue #642, opened 2015-04-02
Do so there is a cursor pointer over the loginbutton and the username in the header


c13emmjo:
So the cursor pointer is not just over the image tag and around the name

b13danso:
Pull request #682 created.

b13danso:
Pull request: #684

PR #684, opened 2015-04-07
#642 - pointer-cursor over username/loginbutton

b13danso:
Added two selectors: userName and loginbutton with the properties cursor:pointer. Set the cursor to default on the class navName so that the cursor doesnt appear on the whole header and only on the label containing the users username/name.


a12oskri:
Ok!

PR #685, opened 2015-04-07
ds#540

a13karsi:
Constraint naming was inconsistent or nonexistent between tables in the database. This commit fixes those issues by defining names for each table constraint. The tables ´studentresultat´ and ´programkurs´ were left unaltered because another group is planning to do some modifications to that table, those can be changed later to conform to the constraint naming scheme applied in this fix.


c13chrjo:
Contraints and their names are named in an intuative way

PR #686, opened 2015-04-07
Umv#514

c13marme:
Deleted basic.php file from UserManagementView directory

Changed so the location for basic.php file is in shared directory instead of UserManagementView


a13antda:
Good Job!

Issue #665, opened 2015-04-02
Make the whole loginButton td clickable and not  just the icon.


b13danso:
The whole loginButton td should be clickable instead of just the icon itself. The icon is to small to be the clickable source for login/logout. Since the td background changes when the pointer is in the td on mouseover/mouseout, it would be a better user experience to make the whole td clickable so it matches the changing onmouseover/onmouseout events.

b13danso:
Pull request #687 created.

PR #687, opened 2015-04-07
#665 - make the whole loginbutton clickable.

b13danso:
Set the onclick-event on the whole td, making the whole loginbutton area clickable instead of just having the onclick-event on the little icon inside.


a12oskri:
OK!

PR #688, opened 2015-04-07
Code view#531

a13alfbe:
The problem was that the "Home-button" had an incorrect link to "sectioned.php" from the CodeViewer.  I hade to create a selection to handle the "EditorV50.php" specifically.


a12jiyba:
The code seems to be good and working. I believe this will work when we merge this to the master at the end of the week.

Issue #680, opened 2015-04-07
Names and PW at login


c13chrjo:
Currently a username and a password is present at the login window.
The tesxt should say "Username" and "Password"

b13danso:
Fixed.

PR #689, opened 2015-04-07
#680 - Loginbox placeholders.

b13danso:
Changed the placeholders for username and password, from Toddler to Username and kalleDgreat2 to password.


a13antda:
This is some fine work!!

PR #690, opened 2015-04-07
codeview#667

a12jiyba:
Added afterid and beforeid.


a13henhu:
This is in proper working order, i've tested this, and it works

PR #694, opened 2015-04-07
Update loginbox.php

a13henhu:
Fixed the login window


a12jiyba:
I have tested your change and its working properly. window.location.reload() was a good ide to use!

PR #698, opened 2015-04-07
Update whiteTheme.css

b13gusbe:
Changed highlight background color to red. Works in example 2 and 3. Example 1 uses an other class for some reason.


a12jiyba:
I see that the highlighting problem works as it should now in example 2 and 3. I like the font-weight you added, it makes the highlighting more visible.

PR #699, opened 2015-04-07
Umv#666 - Hover effect on loginbutton-td instead of the image inside.

b13danso:
Hover effect on the whole td instead of the image inside, which represent the user/door. A new function that adds eventhandlers(onmouseover/onmouseout).


a13ottne:
Beautiful!

PR #700, opened 2015-04-07
Code view#697

a13alfbe:
A lot of unnecessesary css.


b13danso:
This is already fixed and merged.

PR #702, opened 2015-04-07
Update EditorV50.php

a13erisc:
Instead of creating a default page, I've simply added a link to the first example.


a12jiyba:
The update is a good solution to the bumper " Course or Code Example does not seem to exist!" in EditorV50.php. This update makes it easier to go to the examples instead of writing the url self(?exampleid=1&courseid=1&cvers=2013).

PR #703, opened 2015-04-07
CodeView#626 PDO

b12simsa:
Update to use PDO to connect to the database.


f12danjo:
Looks good. The changes implements PDO to connect to the database. It changes all files related to the code viewer. It does however not implement any error handling but this can be added by a separate update.

PR #705, opened 2015-04-07
#583

a13freas:
Updated the datbase for umv. Added tables to be able to store results for each course and every sub part for each course.


a13antda:
This one is branched from master. Redo into the umv15 branch. DO NOT MERG THIS COMMIT!

PR #707, opened 2015-04-07
Update navheader.php

b13gusbe:
Edited the homebutton section to go back one section in the tree and in to DuggaSys again. It should work from both CodeSys and DuggaSys this way.


a13erisc:
Yep, it works.

Issue #630, opened 2015-04-02
Change the icons for the "play", "next", etc.


c13caral:
Change the icons for better looking icons and additionally change so that all buttons have the same size.

PR #716, opened 2015-04-07
Code view #630 change the icons

c13caral:
I added padding to the menu to provide uniform spacing icons also added padding-left to "navHeading". I've also changed the size and positions of icons in the menu


a12jiyba:
Good job with the icons. It looks much better now with the new icons.

Issue #560, opened 2015-03-31
Bottons in "Edit Course"


c13chrjo:
A more intuative way to show the user what the buttons is for. For example: a toolbox description.

HGustavs:
Are we saying tooltips or a tutorial?

I prefer the use of standardized stuff.

c13chrjo:
I was thinking a tooltip that appears when you hoover the mouse over the button.

PR #718, opened 2015-04-08
Update courseed.php #560 Buttons in edit course

a12majca:
just added titles to the buttons.


b13johfr:
It works and makes the buttons slightly more clear.

PR #720, opened 2015-04-08
Added error handling for function pdoConnect()

b13gusbe:
Added a try-catch for function pdoConnect in database.php.


a13jorla:
Can not merge till DB is changed to PDO but error message will work as it seams.

Issue #719, opened 2015-04-08
Cogwheel icon on third box gets patrially hidden


f12danjo:
When displaying the three box template in the codeviewer (as in exampleID 11) the cogwheel icon on the third, lower box gets partially hidden by the box above.

PR #723, opened 2015-04-08
Update whiteTheme.css

a13jorla:
Added padding-top to make coggweel come out from under other box.


b13gusbe:
Works great and solves Issue #719. Good Job!

Issue #722, opened 2015-04-08
Back button not working properly


a13oscwe:
Back button isn't working properly.

PR #724, opened 2015-04-08
Fixed back button in header

a13oscwe:
Fixed the button


a13oscwe:
https://github.com/HGustavs/LenaSYS/issues/722

Issue #629, opened 2015-04-02
There should be only one searchfield in teacherView 


c13emmjo:
Now it is two searchfields
Change this to one were the teacher can search on student ssn, student username, class and course

a13erilj:
b13davfr has fixed this with #730

PR #730, opened 2015-04-08
Remove ID search box

b13davfr:
I have remove one of the search box for ID. #629


a13erilj:
In your code you use the function loadData(), which only takes two parameters. In your case you pass it three parameters. Please fix this and commit the changes.
Line of code where three parameters are passed:
(Line 46)
onclick="loadData(studyprogram.value,pnr.value, null);"

a13erilj:
Good jobb, now the function is used correctly.

PR #732, opened 2015-04-08
Update codeviewer.js

a13erisc:
Changed the jqueries that still directly call the PHP service use the AJAXService.


a13henhu:
Tested, and it seems to be working.

Issue #725, opened 2015-04-08
Diagram is missing for the database


a13karsi:
There doesn't appear to exist a diagram of the database, having a diagram to look at would be advantageous in order to assess what the current relations look like and for further maintainability.

PR #734, opened 2015-04-08
Added a diagram that reflects the current running database into a new do...

a13karsi:
#725


a13oscwe:
Looks good, it's nice to have some kind of documentation for the database!

PR #735, opened 2015-04-08
Only for]code view#594

a13henhu:
Fixed login, solved database problem (small issues)


a12jiyba:
The changes you have made seems to solve the problem of showing alert message. I dont know why or how console.log works but we can use this as a temporary solution.

Issue #736, opened 2015-04-08
Input fields in "New Course" should clear default values automatically


b13alebr:

c13chrjo:
Issue is fixed

PR #738, opened 2015-04-08
issue #736 - Input fields in "New Course" should clear default values automatically 

b13alebr:
The attribute "value" changed to "placeholder" so input field automaticly clears when you type in new values


c13chrjo:
Changes made correctly

Issue #726, opened 2015-04-08
Consolidate header from duggasys with the rest of the systems header


a13oscwe:
Can somebody integrate the headers so the entire system only uses one header in one file?

c13chrjo:
Issue fixed

PR #739, opened 2015-04-08
Consolidate header from duggasys with the rest of the systems header #726

a13oscwe:
Changed all files to include the shared header and added a "back" text under the back icon.
https://github.com/HGustavs/LenaSYS/issues/726


c13chrjo:
Paths looks correct, new indentations looks better

Issue #541, opened 2015-03-31
Clicking "Forgot password?" does nothing


f12danjo:
On the login pop-up window there is a link beside the "Login"-button called "Forgot password". However when clicking this link, nothing happens.

a13chrju:
Issue is still present

c13chrjo:
Issue has a fix

PR #740, opened 2015-04-08
Ds15 #541

a13chrju:
the "forgot password" link now switches UI when clicked. https://github.com/HGustavs/LenaSYS/issues/541

### DIFF
--- a/DuggaSys/courseed.js
+++ b/DuggaSys/courseed.js
@@ -31,6 +31,21 @@ function updateCourse() {
 
 }
 
+var status = 0;
+function toggleloginnewpass(){
+
+	if(status == 0){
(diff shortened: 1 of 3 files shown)


b13alebr:
I think the new function "toggleloginnewpass()" should be placed i dugga.js in "shared"-folder instead of courseed.js in "DuggaSys"-folder.

c13chrjo:
I agree with b13alebr and you should change the names of the placeholder names to something more professional. Like "Password" for password, "Username" for username and "mail@address.com" for mail.

a13chrju:
now the code lays under the folder "shared". placeholder values changed aswell

c13chrjo:
The framework for forgot password is now in place

Issue #594, opened 2015-04-01
Login - All functionalitys (Login, Logout, Attempts, Press "Enter", "Remember me" and so on


a13henhu:
An issue has been identified, the Login button is not "working". When clicked "Login" it wont close and refresh the browser. And while manually refresh your browser, you'll be logged.

a13jansa:
iam on it!

a13henhu:
NEW CODE: onclick="processLogin('CODV'); window.location.reload();return false;"

a13henhu:
#678

a13erilj:
The problem has not been fixed properly, this is just a workaround.

a13chrju:
can I provide a solution :P?

a13henhu:
![alert](https://cloud.githubusercontent.com/assets/11715477/7061983/e03f6ec8-de98-11e4-8794-cc01e36450c1.png)

The alert works now, but still working on how many attempts you're able to login

a13henhu:
#731

a13henhu:
#920 Remember me, need a fix for this issue

a13henhu:
UPDATE: Attempts try to login is now working

c13caral:
I will close this issue. And open issue #1236 for the forgot password doesn't work. And it already exists a issue for remember user #920 
But this issue is kinda messy. I think it is better to make a lot of small issue for the loggin instead of one big.

PR #743, opened 2015-04-08
Issue #594

a13henhu:
../Shared/Codesyspw.php  changed to ../Shared/coursesyspw.php instead


HGustavs:
Not sure what goes on in these fixes?
There should be only one password file for all of duggasys.... why are we adding another one?

Furthermore, we should leave the commented-out section out of the master branch... it should remain clean.

PR #750, opened 2015-04-08
Update codeviewer.js

b13gusbe:
Commented out the previous line and magically made the button work.


b13gusbe:
The previous solution caused the red cross for the cog wheel to stop working. Added a new sperate function for closing the template window. Should be working now.

b13gusbe:
As I said in my previous comment, commenting that line out made other things stop working. I made a new function for closing the window instead. Nothing is commented out.

PR #751, opened 2015-04-08
Added button for choosing template v2

b13gusbe:
Added a button for choosing which template you want to use.

### DIFF
--- a/CodeViewer/codeviewer.js
+++ b/CodeViewer/codeviewer.js
@@ -1284,9 +1284,15 @@ function changetemplate(templateno)
 
 function updateTemplate()
 {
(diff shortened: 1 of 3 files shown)


HGustavs:
I think that there is an issue for changing from $.ajax to AJAXService (which is sanitized in a better way).
I suggest that for following additions to the code, while you're at it, also change from $.ajax to AJAXService.

b13gusbe:
Do you mean that I should take the changes Erik did in pullrequest  #732 and add it to this pullrequest?

b13gusbe:
Closing this branch because it's old. Making a new one from the updated code.

Issue #745, opened 2015-04-08
Extend database for the UMV group.With what thay need.


a13antda:
* add table for class
*add table for 
*add more data of what results the student has
*add table for subcourse for the courses
*add trigger to count the results for each students

c13chrjo:
This is something you probably should talk to a13karsi about

a13karsi:
Sounds good, I am available to help if you need anything! Just come talk to me in E104!

PR #753, opened 2015-04-08
Update init_db.sql

a13antda:
issue #745


a13karsi:
The sql script executes without errors and after going through the code intensively I've determined that it looks alright and I'll accept and merge the pull request. I am aware that the new trigger doesn't work as intended so make sure you create a new issue and fix that!

Issue #508, opened 2015-03-29
Page metadata/titles


HGustavs:
We need to revise so that all of the different dialogs have headers and so on that:
* Titles correspond to their functionality i.e. user editor should have something like "User Editor" or "LenaSYS User Editor" in its title.
* The metadata should indicate UTF-8 and give 
* Proper metadata for mobile clients.

PR #757, opened 2015-04-09
Added the title 'LenaSYS User Editor' to the teacherView.php - #508

a13erilj:
Fixed the last part of the issue #508 
* Set a title on the teacherView.php - new title: LenaSYS User Editor


c13marme:
Looks fine!

PR #761, opened 2015-04-09
Update duggaed.js

c13chrjo:
Fixed the link to the cogwheel in duggaed.js


b12erijo:
Everything seems to be in order. Merging.

Issue #758, opened 2015-04-09
Trigger that calculates the culmination of credits for a specific user for a specific course


a13karsi:
The umv group needs a trigger that maintains a value that represents how many credits a specific user has finished in a specific course.

PR #771, opened 2015-04-09
Created a view to determine total hp for a user in a specific course

a13karsi:
#758

I've removed the previous trigger solution because it seemed like it would turn real ugly fast. Making it easier to find out how many credits a student has in a course by using a view is a really clean solution. This commit includes comments above the new view that demonstrate how the view could be used.


a13antda:
This is fixing the problem.

Issue #767, opened 2015-04-09
Accessed.php is a blank page, cant add any users.


a13chrju:
the only thing showing up when you go to this page is the name of user that is logged in

c13chrjo:
This is false since the text "Toddler" is hardcoded into the webpage.
But otherwise yes. There should be content here. Just remember that oscwe have already merged the navheader into this php file.

HGustavs:
Its important to use a testing link as the following. Also /Shared/ is missing. I get mine to work if I change the links in the header and use a test link like the following:
/DuggaSys/accessed.php?cid=1&coursevers=2013
...

```
	<link type="text/css" href="../Shared/css/style.css" rel="stylesheet">
  <link type="text/css" href="../Shared/css/jquery-ui-1.10.4.min.css" rel="stylesheet">  

	<script src="../Shared/js/jquery-1.11.0.min.js"></script>
	<script src="../Shared/js/jquery-ui-1.10.4.min.js"></script>

	<script src="../Shared/dugga.js"></script>
	<script src="accessed.js"></script>
```

HGustavs:
The access ed should only work when supplied with a valid course link.

c13chrjo:
You should now be able to add users to the access page.

PR #774, opened 2015-04-09
Ds15 #767

a13chrju:
okey so you should now be able to add users to access the courses
https://github.com/HGustavs/LenaSYS/issues/767

### DIFF
--- a/DuggaSys/accessed.js
+++ b/DuggaSys/accessed.js
@@ -82,10 +82,7 @@ function returnedAccess(data)
 		str="";
 		if (data['entries'].length > 0) {
 
-				str+="<div style='float:right;'>";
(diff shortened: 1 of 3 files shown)


a13oscwe:
Kan du ta bort width: 100% style på #content? Den påverkar hela systemet vilket inte är så bra. Annars verkar det funka att 

Can you remove the width: 100% style on #content? line 79 in shared/css/style.css. It affects the entire system and that's probably not a good idea.

a13chrju:
yes will do!

a13chrju:
okey fixed the width

c13chrjo:
Since this is fixed I will merge this into our branch

PR #780, opened 2015-04-09
New button for choosing template

b13gusbe:
Added a new button for choosing which template you want to use. Also edited in Editor50.php since a previous pullrequest flipped the functions for closing the template window and cog wheel window.


c13caral:
The button works properly, and I see nothing wrong in the code. (Besides the SVG file but it's fixed now)
But the function itself, to select the amount of template does not work. But for this issue, it has done nothing to solve that particular problem. So I will create a new issue which then affects this particular problem

Issue #781, opened 2015-04-09
the course name links in file "courseed.php" doesnt give permission for admins do edit courses


a13chrju:
only the pen icon are at the moment the way for the admin to edit the course, the pen icons doesnt show for users with no access but the admins can click these links aswell. 

I want users with access to be able to edit from the course name links aswell as the pen icon links.

PR #783, opened 2015-04-09
Update courseed.js

a13chrju:
just added code, if it's an active version it doesnt just write out the link to access the course, if first checks if the user has write access.
https://github.com/HGustavs/LenaSYS/issues/781


c13chrjo:
The code seems reasonable

PR #784, opened 2015-04-09
Update codeviewer.js

c13caral:
This solves the problem of "ID does not exist".


a13erisc:
Seems to solve the problem.

PR #785, opened 2015-04-09
Ds#782

b12erijo:
Removed the include of /duggasys/basic.php
Problem was that /shared/basic.php was also included in navheader.php, and i think those files are completely identical.


b12erijo:
Missed some issues.

Issue #728, opened 2015-04-08
The template duggas do not follow code standard


a97marbr:
The template duggas do not follow code standard. Please fix.

PR #778, opened 2015-04-09
Did some refactoring, mostly comments, to better fit the code standard. ...

b12erijo:
Did some refactoring, mostly comments, to better fit the code standard. A LOT of documentation is still missing however. Would take too much time to try and really understand the code and properly document it.

PR #791, opened 2015-04-09
Ds#782 + #728

b12erijo:
The issues  seems to be fixed. Also fixed some other include errors in showdugga.php.

### DIFF
--- a/DuggaSys/showDugga.php
+++ b/DuggaSys/showDugga.php
@@ -1,10 +1,3 @@
-<?php
-session_start();
-include_once "../../coursesyspw.php";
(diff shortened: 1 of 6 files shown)


c13chrjo:
In file DuggaSys/templates/dugga3.js there is an accidental "]" in the end of line 19

a13oscwe:
@c13chrjo it should not break anything since it's inside a comment block.
The changes seem to be similar to the already merged branch **https://github.com/HGustavs/LenaSYS/pull/778**
But with added changes, please keep all similar changes inside one branch next time.

PR #795, opened 2015-04-09
Update code standard for function tokenize() in codeviewer.js

b13gusbe:
Updated code standard for function tokenize().


a13erisc:
It works just fine.

Issue #788, opened 2015-04-09
CodeViewer: Hover info


c12jenek:
When hovering the home as well as settings button there is no "hover info" like there is on forward, previous and play-buttons

PR #797, opened 2015-04-09
CodeViewer Hover Info #788

a13erisc:
Adds the missing hover info to the home and settings buttons.


c13caral:
This works great. Finds no errors. However, But there is no title for the Login button.

Issue #790, opened 2015-04-09
CodeViewer: Tight style


c12jenek:
The start of the comment is a bit too tight to the border, should be widened to ease the view
![tight-style-issue](https://cloud.githubusercontent.com/assets/11715479/7067751/0ccfc110-dece-11e4-9dbf-d0cf413964e8.jpg)

PR #798, opened 2015-04-09
CodeViewer: Tight style #790

c13caral:
I have added more "air" in the layout.
I added padding to both classes: codeborder and normtextwrapper.
The numbers presenting line number and code example was very tight positioned against the edges. I then just added a little air between the edges.


a13erisc:
That fixed it, looks much better now.

PR #802, opened 2015-04-10
codeview#799

a12jiyba:
font-size set to 15px. update in whiteTheme.css.


c12jenek:
Font is now standardized across the boxheaders!

Issue #794, opened 2015-04-09
 Write a generic close window function to close any visible window


c13caral:
It would be great if all the pop-up windows uses the same function when the pop-up window shut down.
For more info check on the comment from HGustavs at issue 749
#749

c13chrjo:
Issue was fixed

PR #803, opened 2015-04-10
Ds15 #794

a13chrju:
made a function that can close all popup windows and applied these to onclick-functions.
https://github.com/HGustavs/LenaSYS/issues/794


c13chrjo:
Should really both "dugga.js" "../Shared/dugga.js" be included? Shouldn't we use just the shared one?

a13chrju:
okey, now files only uses the dugga.js under the document "Shared".

c13chrjo:
Seems to be in order.

PR #804, opened 2015-04-10
CodeView#737 and CodeView#733

a13alfbe:
It's now possible to change Sectiontitle and Runlink.
These values does not get the the value "UNK" on reload anymore.


a13jorla:
Since there was no big changes and no security issues seems relevat i agree for changes.

Issue #800, opened 2015-04-09
There is no title in the login button


c13caral:
When hovering the Login button there is no "hover info" like there is on forward, previous and play-buttons

b13danso:
Pull request #805 created.

b13danso:
Merged.

PR #805, opened 2015-04-10
#800 - title-text on loginbutton-tds

b13danso:
Added title-text to the loginbutton-tds.


a13alfno:
Looks good. Keep it up!

PR #807, opened 2015-04-10
Update navheader.php

a13jansa:
Added a link to homepage!


a13alfbe:
I agree with this implementation since the only thing we want to do is go back to the home page and the simple fix is to change the $noup to COURSE. 

The CodeViewToSECTION if-statement is unused code.

Issue #672, opened 2015-04-07
Delete empty files


a12jiyba:
Delete empy files called #515 , #563, 566

a12jiyba:
#658

Issue #658, opened 2015-04-02
Mysterious files and folders


HGustavs:
* There is a mysterious file called "#563" which seems to be empty. Delete it.
* There is a .idea folder. I suggest removing it or gitignore-ing it.

b13danso:
Fixed and merged.

PR #808, opened 2015-04-10
#658 - Delete .idea-folder

b13danso:
Deleted the file inside of .idea-folder(be able to remove folder). Issue #658 addressed both this and removing mysterious files, but that was fixed with issue #672.


a13alfno:
Good job. Looks like it works without the file.

Issue #759, opened 2015-04-09
When logged in you should be able to click on your name to access umv 


a13freas:
If you are a student you should be redirected to student view and if you are a teacher to teacher view.

PR #810, opened 2015-04-10
#759 - Made it possible to enter UMV by clicking the name.

a13erilj:
Users can now click their names to enter their umv.
* Students will be redirected to studentView.php
* Teacher will be redirected to teacherView.php


a13alfno:
It works when logged in as toddler (teacher). Click on "Toddler" takes me to teacherView.php. Have no student user to test with.

Issue #543, opened 2015-03-31
A redirect file shoudl be present in LenaSYS root


a97marbr:
It is not obvious how to start LenaSYS and display a file structure is not good.
I suggest adding an index.php including the following code in LenaSYS root directory:
``` html
<html>
	<head>
			<meta http-equiv="refresh" content="0; url=http://<your_url_here:port>/<your_repository_name>/DuggaSys/courseed.php" />
			<script>
					window.location.replace("http://<your_url_here:port>/<your_repository_name>/DuggaSys/courseed.php");	
			</script>
	</head>
	<body>
			Redirecting to LenaSys!
	</body>
</html>
```

c13chrjo:
Changing this to "All"

b13danso:
Fixed and merged.

PR #815, opened 2015-04-10
#543 - Create index.php on root-directory.

b13danso:
index.php is located in the root directory, redirecting the user to the LenaSys-frontpage instead of showing the root file structure.


a13alfno:
Tried your version. Works.

PR #818, opened 2015-04-10
Updated mockups

f13marhe:
new layout on mockups.


a13alfno:
The new meters are awsome! I think our customer will be happy.

PR #820, opened 2015-04-10
When you log out you get redirected to the startpage

a13freas:


a13erilj:
There are a few buggs in the solution, for exempel the login-box is shown when logging out. 
Lets take this issue first thing on monday instead.

Issue #509, opened 2015-03-29
Movement of borders


HGustavs:
In code viewer it would be nice if it was possible to move the borders of the areas.
for example, if it is split 50% 50% vertically, this may not be ideal if we want documentation on the left and code on the right.
We may want a 20% 80% split instead.
If we could drag the borders each user could adapt his view to his preferences.

HGustavs:
Local storage to store the setting would be sweet!

HGustavs:
Click and drag of border.

b12alela:
im on it

PR #821, opened 2015-04-10
#509

b12alela:
This adds resizable boxes for the codeviewer. It currently works for template 1 and 3.


a13alfbe:
I agree with this implementation. Works well for template 1 and template 3!

Issue #829, opened 2015-04-13
Update database diagrams to reflect this week's merge


a13karsi:
The database diagrams in shared are now different from the currently running database, for maintainance purposes the diagram in /shared should be updated to fit the currently running scripts.

PR #836, opened 2015-04-13
Changed database diagrams

a13karsi:
The diagrams now reflect the currently running database. #829


a13oscwe:
Nice work, having a updated diagrams every week is pretty sweet.

a13antda:
it gives a nice overview of our db. One of these every weak sounds good. Makes it easier to see what needs to be added.

PR #839, opened 2015-04-13
Added button for choosing template that got lost in merge.

b13gusbe:
Added button for choosing template that got lost in merge.


b13danso:
Tested and reviewed. Works OK. Awaiting merge confirmation from group leader.

PR #844, opened 2015-04-13
Removed redundant files and changed includes

b12erijo:
The files that were duplicate(existing in both /Shared and /DuggaSys) have been removed and the includes corrected.


c13chrjo:
Changes do what they are supposed to do

Issue #845, opened 2015-04-13
the "forgot password" link has wrong "onclick" method


a13chrju:
the method is wrong in the "onclick=''" thingy.

a13chrju:
https://github.com/HGustavs/LenaSYS/pull/846

c13chrjo:
Issue fixed

PR #846, opened 2015-04-13
ds16 #845

a13chrju:
method fix.


a13oscwe:
Looks fine, should the actual password recovery be implemented? If so which group takes on that responsibility? Anyhow that would be a different issue i guess.

c13chrjo:
It should not be functional since we don't have a mailserver available

a13chrju:
yeah, it's only the visuals

a13denli:
Good!

PR #848, opened 2015-04-13
Code view#651

a12jiyba:
Deleted example code that was conflicting with the id of examples we created.
I also added a new example.


b13danso:
Checked, looks OK. Awaiting merge confirmation from group leader.

PR #850, opened 2015-04-13
CodeView#819 Update init_db.sql

a13jorla:
This should fix the problem from issue#819.


a13jorla:
Altered as suggestion and added

a13jorla:
Could not merge.

PR #852, opened 2015-04-13
Refactor shared navhead to work with all groups

a13oscwe:
Refactored how the "home" and "back" buttons work for all the groups. It relies on the $noup variable to see what the back button will link to.
Uses ternary operations.

### DIFF
--- a/Shared/navheader.php
+++ b/Shared/navheader.php
@@ -13,7 +13,7 @@
 			}else if($noup=="SECTION"){
 				$cid=getOPG('cid');
 				$coursevers=getOPG('coursevers');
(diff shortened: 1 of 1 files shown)


a13denli:
A case that does nothing doesn't really look good. I think we need to force $noup to be something or use the default case. Otherwise it looks like we have something in the system that shouldn't be there.

a13oscwe:
I think it makes alot of sense to have a switch statement here for several reasons. This code doesn't just refactor the existing code but adds support for other parts of the system to add their own cases following a pretty clear format. Switch cases are awesome and there shouldn't be any scenarios where the condition needs to be extended since the $noup value seems to just be something that let's the system know what kind of page or subsystem you are in.

a13denli:
Good!

PR #854, opened 2015-04-13
Cv#849

c12jenek:
Changed $noup and added a generic way to create home and back buttons. Changes in navheader.php copied from ds#838


b13danso:
Looks OK!

PR #855, opened 2015-04-13
Did issue 853

c13emmjo:
This is just containers for divs who should be in teacherView.php
so it’s not finished just a shell

Also change the width back to 80% and the border in shared - style.css

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -77,8 +77,6 @@ hr {
     padding: 50px 12px 8px 8px;
     background: #fff;
     min-height: 200px;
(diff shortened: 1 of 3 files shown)


a13antda:
code: the comments gives redundancy when you comment the id. example
	
	<div id="content">

there are class names in the divs and just gives comments instead. Give placeholdernames instead.

Why is it nested lists?

Test: the css changes works.

a13antda:
Kode: ok

test: works

this is what we wanted. Good that you didnt change stuff in shared

PR #859, opened 2015-04-13
Update dugga.js

a12jiyba:
Added "+file" at line 138 so the play demo button works.


c13caral:
Everything looks good. All sample code does not work. Though the fault is not with the play button. The error exists in the database, then it is wrong "linked" in the database.

c12jenek:
Looks good, this was removed during merge as we did not know what it seemed to do!

Issue #747, opened 2015-04-08
Creating a course with special characters can cause problems


b13johfr:
If the name of the course contains an " ' " for example, it does not work to edit the course.

HGustavs:
Does the function use the sanitized ajax service?
If not, that may be the bug.
Otherwise this should be fixed in that function (e.g. AJAXService) so that all dialogs are fixed at once in all places.

a13oscwe:
The regex fix seems like a good idea, it works and it allows only certain characters now which might be a problem depending on what we want. The problem that arises now is that you can add or edit a new course and if you use non-whitelisted characters(according to the regex) the course will still be created but with a blank name. If we could factor out this regex into it's own function and bind it to the save/create buttons and give some kind of "invalid course name" feedback that would be great!

PR #868, opened 2015-04-13
Added a regular expression

b13johfr:
Added some regular expression to the ajax service funktion that only allows white-listed characters. Fixes issue #747 and possibly more related problems that would have had problems with incorrect input.


a13karsi:
Seems like the concensus is that this fix is a good thing but requires some additions in order to give proper feedback to the user. We'll just merge this branch and fix issues #881 and #877 in a separate branch with the solutions that have been proposed.

Issue #879, opened 2015-04-14
Creating containers for studentview.php


c13marme:
Creating div elements for studentview.php where the information about the student will be stored.

PR #880, opened 2015-04-14
Update studentView.php

c13marme:
Removed old divs and created new div, better suited for the mockup

#879


a13erilj:
There is some old test code that has been added and should be removed.

c13marme:
Removed the code that isn't relevant to the divs.

a13erilj:
The new divs look fine.
There is however still som small indentations problems in the header.

PR #882, opened 2015-04-14
Ds16#635

c13andka:
set a cookie so that we can get a variable from php, this cookie will be used so that we can know what page we are on when the loginbox is showin. This cookie is needed so that we can press enter when we want to login and know what page we are on


a13denli:
Remove unnecessary white space from the code added :)

PR #886, opened 2015-04-14
CodeView#860/#501 Create template6.css

a13jorla:
Made this new css-file to meet demand from customer.

### DIFF
--- a/Shared/css/template6.css
+++ b/Shared/css/template6.css
@@ -25,8 +25,8 @@
     right: 0px;
     height: -webkit-calc(40.0%);
     height: -moz-calc(40.0%);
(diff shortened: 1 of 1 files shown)


a13henhu:
I've been looking at this pull request, and there was a small issue, not really that hard to change, but the borders, had some white lines, and i've been trying to see what the problem was, and I could identify that the percentage was slightly off. Either way, it works, and here's a small fix.


For the the width in box2-4 I've changed these lines to:

    width: -webkit-calc(66.0%)!important;
    width: -moz-calc(66.0%)!important;


and for box4 I changed the height to:

height: -webkit-calc(29.3%);

a13jorla:
Have done changes to the code as suggested.

Issue #715, opened 2015-04-07
Warning when adding a test (dugga) if no dugga is present


a97marbr:
When trying to add a test (dugga), i.e., changing a row from a 'label' to 'test' in the course overview, and there has been no dugga prepared (no dugga added and/or no dugga variant present) we should be notified of this and a brief explanation on where we can add dugga/variant should be given.

PR #887, opened 2015-04-14
Ds16 #715

a13chrju:
the admin can no longer create a test if no dugga exists. A message is shown if there is none and the value "test" in the select menu becomes unselectable and gray.
https://github.com/HGustavs/LenaSYS/issues/715


a13denli:
Remove unnecessary white spaces

a13chrju:
updated and removed some white lines.

a13denli:
Good

Issue #881, opened 2015-04-14
Only whitespace input shouldn't be possible 


a13karsi:
For example, creating a course that has nothing but whitespace as input creates a new course on the course list that has no name. Instead, the user should be informed that creating a new course requires a valid name and course code.

HGustavs:
I think this is related to the regex issue.
Im thinking we may want to stop this on the service level and also give feedback on the service level.

I am against code that has to be written for every input box since there is a risk that you might miss some box..... which unknowingly passes bad stuff to the database.

However are there cases when we want white space only in an input?

a13oscwe:
@HGustavs The code is "stopped" at the service level now and i guess that's fine but there should be some kind of frontend validation aswell letting the user know. https://github.com/HGustavs/LenaSYS/issues/747 https://github.com/HGustavs/LenaSYS/issues/877

HGustavs:
Im thinking an alert box... suggestion in comments for issue #877.

a13karsi:
That's the easiest solution for the time being. If it's okay with everyone I would like to have pull request #868 merged and we'll address this in a separate issue, that is, adding an alert box with a notification.

a13karsi:
Merged #888

Issue #877, opened 2015-04-14
Disallow non-whitelisted character input in text boxes


a13karsi:
With the new regex powered parsing that only allows white-listed characters to be accepted as input, see pull request #868 and issue #747 the problem that now arises is that a user may inadvertently input a course name such as "Toddler's epic course" only to find out that the ' character would be removed and the resulting name would be "Toddlers epic course". 

In order to improve the user interface and experience of the user, all user input should be parsed as well right away to indicate to the user that some characters are not allowed as input.

HGustavs:
My suggestion is to do this at the service level so that you do not have to implement this code on every input box. I am hesitant on features that have to be added manually to input boxes since this often results in boxes that have not been fixed which in turn results in potential injections since the site owner thinks it has been fixed.

e.g. an alert that says "some characters have been removed from the input: xxxx".

HGustavs:
Now if there are several boxes at once that have this issue perhaps writing an alert for each field is overkill... so if there is more than one a coalesced output is preferred to me at least.

a13karsi:
I agree that it would be overkill, since it would require so many changes to so many files. Me and Johan were looking at the code and tried looking for if there was a common point of origin where text input was handled but it seems like the ajax service function is the only candidate. 

It is probably the best solution as you mentioned that we alert the user from the service level that some characters have been removed.

a13karsi:
Merged #888

PR #888, opened 2015-04-14
Alert boxes for wrong input

b13johfr:
Alert boxes will now tell the user if there are any illegal characters removed in either the coursename or the coursecode while creating a new course. Also if the coursename doesnt consist any characters at all.

Issues:  #877 and #881


a13oscwe:
Looks good, atleast now the user knows that something went wrong.

a13denli:
Good

PR #889, opened 2015-04-14
Code viewer#871 markdown bold/italics

b12simsa:
Added a markdown parser, and a call to the parser. 
The parser is not yet fully functional, and to test it the variable desc must be reassigned a value before sent into the parser.

### DIFF
--- a/CodeViewer/codeviewer.js
+++ b/CodeViewer/codeviewer.js
@@ -1534,3 +1534,166 @@ function resizeBoxes(parent, templateId) {
 			}
 		}
 	}
+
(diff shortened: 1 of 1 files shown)


a13alfbe:
It works! Just need someone else to take a look at this before i merge it.

b12simsa:
There might to be some problems with Chrome, will do some further testing.

f12danjo:
Works in Chrome now after "new Map()" got changed to "{}" when initializing the maps.

c13caral:
I have tested this and everything works!

PR #892, opened 2015-04-14
Fixed login redirecting to courseed.php as well as login-prompt showing ...

b12erijo:
...when not supposed to

This seems to work for me now at least. The login will also have to be changed though.
The problem is that navheader.php doesn't reload after logging in.

Note that when logging out, there would be problems if a file is located "deeper" than one level.
Eg. DuggaSys/randomDir/file.php would not work. I assume that each document is located in one of the three directories and not on a deeper level.
Give me a poke if this doesn't make sense.


a13denli:
Why do you call urlDivided.pop(); 2 times?

a13denli:
good

Issue #812, opened 2015-04-10
There should be placeholders instead of current implementation in input boxes


c13chrjo:
Text should not have to removed manually in the input boxes when typing in them

b13johfr:
Fixed

PR #893, opened 2015-04-14
There should be placeholders instead of current implementation in input boxes #812

c13chrjo:
The fixes concerns DuggaSYS files only! #812


c13chrjo:
Take notice that @b12erijo changes from #844 is included in this branch. Why this is I don't know...

b13johfr:
The commits should not merge into master!

PR #895, opened 2015-04-14
Version 2: There should be placeholders instead of current implementation in input boxes #812

c13chrjo:
This concerns only duggaSYS files! #812


b13johfr:
This makes editing courses in duggasys more glanceable.

a13denli:
Good

PR #898, opened 2015-04-14
Dropdownmenu UMV teacherView and studentView

c13emmjo:
Issue 857
Done inventory in student view as well.

Feel free to come with comments for the css
i was thinking adding a function in javascript for the hover effects
later on.

And also there is going to be a "location" class so the menu are darker when you are on that page.

(this is just a first layout as a said feel free to comment on tips i have some in mind but nice to se what you are thinking as well)

teacherview:
http://webblabb.iki.his.se:8080/umv/c13emmjo/week16/UserManagementView/teacherView.php
studentview:
http://webblabb.iki.his.se:8080/umv/c13emmjo/week16/UserManagementView/studentView.php

### DIFF
--- a/UserManagementView/studentView.php
+++ b/UserManagementView/studentView.php
@@ -38,10 +38,10 @@ pdoConnect();
     
         <!-- Dropdownmenu -->
         <nav class="rightDropdownMenu">
-            <ul>
-                <li><a href="#" >Nuvarande år</a></li>
-                <li><a href="#" >Tidigare år</a></li> 
-        	 </ul> 
+        	<ul>
+        		<li><a href="#" >Nuvarande år</a></li>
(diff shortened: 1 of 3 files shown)


b13frero:
I have now looked at your dropdown menus. They are working and they look good, But I would prefer to switch the standard color with the hoover color for a more and vice versa.

c13emmjo:
Thats not how ever what we have agreed on with the client . Now as a said there is no location color but there will be. Are going to do that later on. So i think it will look better this way - so it melts better with the header and so on. can also see the example in the mockup :) but this we can change after woods if the client say so.

c13emmjo:
... the inventory changes back to weard... dont get it... :/ pulling my hear of... looks ok in my programs on my computer but puff and it looks weard when i send the pull request..

PR #899, opened 2015-04-14
CodeView#876 Update template2.css

a13jorla:
Made changes to template2 to make the boxes appear over and under eachother.


b12simsa:
The template is now working in Firefox, Chrome, and Opera (it did not work in the Explorer version I tested in). In my opinion this could be merged with the week branch.

a13alfbe:
I also agree with this change

Issue #828, opened 2015-04-13
Codeviewer can't print indentations


a13erisc:
Currently, the files in the "codeupload" folder use spaces for indentation. When replacing the spaces with tab-indentations, they aren't printed properly.

c12jenek:
Sounds like it could be the result of a bug when trying to print from codeupload

PR #900, opened 2015-04-14
Code viewer #828  Can't print indentations

c13caral:
This will fix the tab function for the code examples

You have to use "&#9" to make it work in code examples, for example "js1.js"

// Fix scrolling on touch devices
   
   	var ScrollFix = function(elem) "Grise" {
   	    // Variables to track inputs
   
   	var startY, startTopScroll;
 
&#9;elem = elem || document.querySelector("elem");
 
&#9;//If there is no element, then do nothing

#828

### DIFF
--- a/CodeViewer/codeupload/js1.js
+++ b/CodeViewer/codeupload/js1.js
@@ -3,8 +3,8 @@
    	var ScrollFix = function(elem) "Grise" {
    	    // Variables to track inputs
    
-   	    var startY, startTopScroll;
-   
-   	    elem = elem || document.querySelector("elem");
-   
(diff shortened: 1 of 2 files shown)


c13caral:
I think this is a very "ugly" solution. So do you have any ideas for a better solution would be awsome!

a13alfbe:
I think that the customer wants to be able to use tabs directly in the document and not "&#9". You could possibly use regular expressions to find the tabs?

f12danjo:
Will remove this since it won't get ready for the weekly merge with the master branch. Have spoken with c13caral and she agrees.

Issue #894, opened 2015-04-14
Parse data from db to studentview.php


c13marme:
Parse data from DB to the studentview.php. 

Studentview.php to show title info, what class and what the name is of the student as a title.

a13erilj:
This issue has been solved by c13marme

PR #907, opened 2015-04-15
Umv16#894

c13marme:


c13marme:
#894

a13antda:
GROUP LEADER: code = ok: test: it works.

a13erilj:
There is some old code that should be removed, otherwise it looks good.

c13marme:
The class and name of the student is shown in white text

Issue #489, opened 2015-03-27
At least two loginbox.php files exists -- should only be one


a97marbr:
The loginbox should be the same for all subprojects ... make sure the one in Shared is current

a13ottne:
Solved

PR #909, opened 2015-04-15
Umv16#489

a13ottne:
Solved issue #489 
Removed an instance of loginbox.php and changed all references to '../Shared/loginbox.php'


a13ottne:
Redundant

Issue #903, opened 2015-04-15
database optimization - clear data in table "eventlog"


a13chrju:
When the login attempts are counted and implemented correctly into the system a hugh number of data will be stored in this table whenever a user makes an unsuccessful login.

to make the system go faster it would be nice if the database performed a weekly(or something like that) table reset of table "eventlog".

a13karsi:
I'll take a look at it and see what's possible!

a13chrju:
nice job, I'll use your code my my own website too! hehe :+1:

PR #910, opened 2015-04-15
Added a weekly event in the database, issue #903

a13karsi:
This event should perform a weekly clear of the eventlog table. The eventlog table can grow to a great size and old information in that table may not be relevant after a long time. Moreover, for performance and storage space reasons, this event is also an optimization. #903


b12erijo:
Looks good to me.

Issue #911, opened 2015-04-15
StudentView coding of html and css


b13frero:
First draft of the studentView

http://webblabb.iki.his.se:8080/umv/b13frero/week16/UserManagementView/studentView.php

PR #912, opened 2015-04-15
Umv16#911

b13frero:
#911


f13marhe:
looks like a good start but further development is needed. some names should be changed and so forth.

PR #917, opened 2015-04-15
Umv16#915

a13freas:
updated usermanagementviewservice.php and usermanagementview.js.


a13erilj:
Due to new updates to studentView.php these changes may not work.

Issue #817, opened 2015-04-10
CodeViewer: Templates aren't connected properly


c12jenek:
Choosing templates doesn't result into the correct template, the id possibly isn't equal to the icon of the template.

HGustavs:
Is this after-the-fact selection of template or first-time template selection?

If first-time works and after-the-fact does not, it means that it is only the after the fact one that is broken.
Choosing a template must mean that we also create or update the corresponding box entries and so on so that everything matches.

a13jorla:
Similar to #789 .

PR #919, opened 2015-04-15
Code view#913

a13jorla:
Made new file in shared/icons named template6_butt.svg and added line in EditorV50.php to implement the svg-file(line240). So the table will look like this. 
edittemplate


a13alfbe:
Could not find any problems with this implementation. Looks good!

c13caral:
This looks good. It's sad that we can't test the function and se if the new layout of templates works. But we have already one issue for that specific problem. #817

PR #922, opened 2015-04-15
Code view#806 Edit Example: Before/After links

a13alfbe:
Added functionality to edit before/after example


a13jorla:
The added code works as to go to other examples then ordinary, by using jump between example-buttons. Ready to be merged.

c13caral:
Everything looks good. I didn't find anything wrong with the code or the result. 
But it is a bit hard to not get lost in this. I mean that this is not a user friendly way to organize all the code examples in my opinion

Issue #921, opened 2015-04-15
Comments and width changes in the Student View


b13frero:
More Comments made in the html and changed with on the Main standard bar

PR #923, opened 2015-04-15
Umv16#921

b13frero:
#921

### DIFF
--- a/UserManagementView/studentView.php
+++ b/UserManagementView/studentView.php
@@ -3,32 +3,25 @@ session_start();
 include_once "../../coursesyspw.php";
 include_once "../Shared/sessions.php";
 pdoConnect();
-	session_start();
-	include_once "../../coursesyspw.php";
-	include_once "../Shared/sessions.php";
-	pdoConnect();
 ?>
 <!DOCTYPE html>
 <html>
-
   <head>
-    
-        <meta charset="utf-8">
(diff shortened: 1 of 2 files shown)


c13emmjo:
There are a few thing to be fix on the page so are typing it down but the comments yot have added looks good:) 

There are some strange things that have happened ..
On line 125 you come to "my" side of the teacherview , whats the reason for this? 
"<a href="http://webblabb.iki.his.se:8080/umv/c13emmjo/week16/UserManagementView/teacherView.php">Nuvarande år</a>"


Line 162 and line 170 provides fault. It is missing spaces between id = "ProgressbarG2F" and role = "progressbar" same on line 170 (words before the role it should be a gap here)

br tag at line 185, are we going to use br tags?

On line 188, it is a class caled "my-container-fluid clears fix" that allows the wrapper enclosing the three classes "courseInnerWrapp" that there is margin of 25 to the left but not the same on the right so if you can see both the progress bar and the three years floats a little more to the right, 
these should be in the middle of the page with the same margins on both sides.
in css it looks 
.my-container-fluid {
	position: relative;
	top: 10px;
	left: 25px;
	width: 100%;
	
}

Maybe you can fix this problem in another issue?

a13antda:
This pull request is to far behind our weakly branch. It will be closed.

PR #924, opened 2015-04-15
Code view#872 markdown headings

f12danjo:
Implemented headings functionality with markdown.


c13caral:
There was just one problem. when you added Three stars (***) it became like h1 e. g. 

*** Test

Is the same as

<h1>Test<h1t>

a13alfbe:
The only thing i could find was that the headers
Example:
\# ## Header 1 Should be:
# ## Header 1
But instead it is like this:

\# 
## Header 1

f12danjo:
It is true that if you input "# " (any number of #) followed by more "# " on the same row, you get several headings, each on their own row.

The question is: Is this reason enough to not merge this branch?

The code doesn't break anything and the headers works if you write correct markdown syntax :)

b12simsa:
Maybe merge and at the same time create a new issue about headers not being parsed correctly in all circumstances?

c13caral:
There is some problems but we gonna create new issue for them  #959 and merge this pull request.
Because the issue is solved but one bug was found

Issue #840, opened 2015-04-13
Make resizing boxes work for all templates.


b12alela:
Make the resizing work for all templates, currently works for 1 and 3. Should work for templates 1-5.

c13caral:
This issue is now solved

PR #929, opened 2015-04-16
Code view#840

b12alela:
Functionality for resizing boxes for template 1-5.
Fixed the CSS for template 3 as it was just a copy of template 4.
Removed redundant overflow properties.
Removed a link to template1.css in EditorV50.php, that was overwriting the css for other templates when they are loaded. Causing them to display incorrect structure.

### DIFF
--- a/CodeViewer/EditorV50.php
+++ b/CodeViewer/EditorV50.php
@@ -67,7 +67,6 @@ EditorV50.php?exampleid=1&courseid=1&cvers=2013
 
 	  <link type="text/css" href="../Shared/css/jquery-ui-1.10.4.min.css" rel="stylesheet">  
 
-		<link type="text/css" href="../Shared/css/template1.css" rel="stylesheet" />
 		<link type="text/css" href="../Shared/css/codeviewer.css" rel="stylesheet" />
     <link type="text/css" href="../Shared/css/whiteTheme.css" rel="stylesheet" />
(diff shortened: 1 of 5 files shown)


a13alfbe:
There is some weird overflow behaviour when resizing the browser window.
untitled

Template 4 & Template 5 breaks when resizing the browser window:

Template 4:
template 4 breaks

Template 5:
template 5 breaks

a13jorla:
If template3.css is changed from the last cv16 branch it will not match the example in choose example button.
template3

b12alela:
The behaviour it presents after resizing the browser window has to do with it not having a defined function for handling that kind of resizing. This should be resolved in a separate issue, which would be solved by something like using the window resize function in jQuery, or forcing exact percentage based positioning via some custom function.

b12alela:
From what i can see the template3_butt.svg icon is the same in both branch CodeView#840 aswell as in the weekly cv16 branch, so no changes to the design of the template seems to have happened. The css that was changed in template3.css was due to it mirroring the design specifications of template4, with the css being identical in template3.css and template4.css. 

In this update i rewamped the css in template3.css to match the layout of template3_butt.svg in the icons folder. Which should display correctly the layout that is being represented in the icon.

c13caral:
I agree with b12alela, We have to create seperate issues for each problem. But I am going to merge this because everything works great if you don't resize the browser window. The code follows the code standard in my opinion. #840

Issue #914, opened 2015-04-15
Dragging a "test" item to another "moment" does not recouple properly 


a13oscwe:
When dragging a "test" item from one "moment" in the course page to another "moment", the "test" item is not recoupled to the new "moment".

To reproduce the bug do the following.
Click on a course from courseed.php press "new" and add two "moment", then click "new" again and add a "test", now click the settings gear icon for the test and couple it with a moment. Then drag the "test" item to another moment or decuople it. By checking the "settings gear" the "test" item does not recouple properly.

a13oscwe:
https://github.com/HGustavs/LenaSYS/issues/937 Issue is solved here.
The changes work well, merging.

PR #935, opened 2015-04-16
Ds16 #914

a13chrju:
dragging a test item from one "moment(English: Seqment)" to another changes the moment that is belongs to in the database. 
I figured the cogweel option should not have an option for changing the moment it belongs to but instead make it easier by letting the admin only make this change to a test by dragging it around and place them under antother "moment".

Also, the word "moment" will have to be changed to Segment or something more fitting.
https://github.com/HGustavs/LenaSYS/issues/914

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -12,6 +12,14 @@ var xelink;
 function displaymessage(){
    $(".messagebox").css("display","block");
 }
+$(document).ready(function(){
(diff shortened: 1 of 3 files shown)


a12majca:
looks good for now but we need to fix the problem with that you need to choose that "the moment is itself", should be done automatically.

a13oscwe:
New changes look good, merging and closing issue https://github.com/HGustavs/LenaSYS/issues/914

Issue #941, opened 2015-04-16
Add " . " " - " and " " @ " to regular expression in dugga.js


b13johfr:
We need to add some characters to the regular expression to be able to add new users, mail etc.

a13oscwe:
Continue this issue here: 
https://github.com/HGustavs/LenaSYS/pull/942

Closing

a13oscwe:
Fixed in merged branch #941

PR #942, opened 2015-04-16
Fixed issue #941

b13johfr:
Added the following characters to the regular expression, @ . - 

#941


a13oscwe:
I get an "Uncaught TypeError: Cannot read property  'match' of null så att någonting verkar vara knas.

a13oscwe:
New changes work, merging.

PR #946, opened 2015-04-16
Added highlighting and tag matching for html

b13gusbe:
Added highlighting and tag matching for html in codeviewer.


f12danjo:
Nice! I downloaded your branch and tested it. Works fine. The code is also well indented and follows the code standard as far as I can see.

a13alfbe:
I have also tested this branch and it worked great. I think this pullrequest is ready to merge.

Issue #926, opened 2015-04-16
CodeView: Edit example popup box design


a13alfbe:
The popup box for edit example looks very weird and when the select-box hold a very long value it looks like this:
![untitled](https://cloud.githubusercontent.com/assets/11715493/7175733/285eefce-e418-11e4-8752-7f7c8d3c99c8.png)

c13caral:
This issue is now solved

PR #956, opened 2015-04-16
Pop up box #926

a13jansa:
Add some new code in CSS and PHP file


a13henhu:
I've checked this request, and everything is working as it should.

c13caral:
Everything works good!
But there are some problems with the code standard.

Like in line 206 it is:
      <td>
	Play Link: 
</td>
		<td>
		Important Words:
		</td>

a13alfbe:
Change the colspan of the two <td> that contains #impword and #playlink to avoid using negative margin to position the #impword input. 

There is also an empty <tr> between the #impword and #impwords

c12jenek:
Code standard checks out, approved to merge!

c13caral:
#926

Issue #954, opened 2015-04-16
resulted.php is displaying the wrong content and an empty div


a13chrju:
the admin cant give grades to students.

PR #957, opened 2015-04-16
Ds16 #954

a13chrju:
the teacher can now give grades to students! yey!
https://github.com/HGustavs/LenaSYS/issues/954


c13andka:
Seems legit

PR #958, opened 2015-04-16
Update dugga.js with autofocus on login inputbox

c13andka:
When the user opens up the login popupbox, it will now autofocus the input username box. so its just to type username


a12majca:
well done!

Issue #959, opened 2015-04-16
CodeViewer Markdown using "#" in headings


f12danjo:
As it is now the input "# ## Hello" would look like:
## Hello
Instead it should look like this:
# ## Hello

You can have "#" in headings if you just don't have a whitespace directly after it, like this: "# ##Hello".
This input would produce:
# ##Hello

The markdown parser should be fixed so that only the first "header symbol" on a line row should be interpreted as markdown. However you should be able to combine header with other markdown styling, like "*".

c13caral:
This issue is now solved! :+1:

Issue #925, opened 2015-04-15
CodeViewer Markdown symbol "***" bold AND italic


f12danjo:
Currently the markdown symbol *** does not have a function. Text between two *** should become bold and italic.

f12danjo:
Fixed by [pullrequest #962](http://github.com/HGustavs/LenaSYS/pull/962).

PR #962, opened 2015-04-16
Codeviewer#959 markdown headings

b12simsa:
Changed the parser so that it uses the replace() function from javascript with regular expression as parameters, Instead of the previous functions.
This also makes the parser able to handle headings, and whitespace after markdown symbols better. The new function needs to be tested with a lot of different markdown inputs since I have not tested it all.

### DIFF
--- a/CodeViewer/codeviewer.js
+++ b/CodeViewer/codeviewer.js
@@ -157,7 +157,7 @@ function returned(data)
 					}
 					
 					/* Assign Content */
-					//Parse markdown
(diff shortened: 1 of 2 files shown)


c13caral:
This looks really great!! It is ready to merge! :+1:

a13alfbe:
Works for me too. Huge improvement, good job!

c13caral:
#959

f12danjo:
This also fixed issue [#925](http://github.com/HGustavs/LenaSYS/issues/925). Closing that issue.

Issue #966, opened 2015-04-17
Pop-up box Top header


a13jansa:
The header or the close button är 1-2 px to big or to small.

a13jansa:
I have change some code in style.css-file. The close button are in correct place.

a13jorla:
After merge the close button is placed right in the header. So now i will close this issue.

PR #970, opened 2015-04-20
Update header in the pop-up box.

a13jansa:
I have made some small changes.
#966

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -479,7 +479,7 @@ font-size: 14px;
 }
 
 /* --------------================################================-------------- *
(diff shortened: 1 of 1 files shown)


a13jorla:
The close button in popup login window was positioned wrong and left some purple line to the right of it. But the fix makes the line disappear.  But you should look with other goup if they want to do the same change, otherwhise it will still look wrong in other places.

c13caral:
This looks really great :+1: !
You just have to talk to the others groups before we merge.

a13antda:
Weakly merge time. Merge your braches before friday 1300. Remake these changes next weak.

Issue #971, opened 2015-04-20
Top border/ Css. (New)


a13jansa:
I made some changes last week but after have updating the new branch it stil look the same.

a13jansa:
I deleted the shadow in top border/header. The one with grey background.

a13jorla:
Since this issue is merged i will close this issue.

PR #974, opened 2015-04-20
Update style.css... Top-header/border. 971

a13jansa:
I deleted the shadow in top border/header. The one with grey background.
#971


c12jenek:
This edit is in a shared file, wouldn't it be more appropriate to add the shadow to the other header then to remove the shadow for all?

c12jenek:
Since the flat color style is adopted by other groups, this edit is approved

Issue #969, opened 2015-04-20
Database diagram for this week


a13karsi:
I'll update the diagrams so that they reflect the current version of the database.

PR #976, opened 2015-04-20
Updated database diagrams

a13karsi:
Database diagrams reflect the current version of the database on master #969


a13karsi:
Something didn't go quite right, I'll try again.

Issue #979, opened 2015-04-20
Code View: Incorrect HTML in Edit Example box


a13alfbe:
![untitled](https://cloud.githubusercontent.com/assets/11715493/7226008/bbf02940-e744-11e4-972a-5aedf4662db4.png)

Colspan is not a proper HTML tag. It is an attribute that is part of the \<td> tag

a13jansa:
I made those changes
left col = <td colspan="1">
right col = <td colspan="2">

HGustavs:
Can we close this issue soon?

PR #987, opened 2015-04-20
Update v50.php

a13jansa:
I made some changes in the code. it should be correct now.
#979


a12jiyba:
your solution to use "colspan" in "td" is better and correct than the old code. good job!

a13henhu:
I've been looking into this, and yeah, there was a statement error, and the fix does work.

Issue #931, opened 2015-04-16
When a user logs out some information still shows in studentview.php


c13marme:
When a guest enters the site, some information still shows up that isn't suppose to.

a13ottne:
Visitor should be directed to start page upon logout

a13erilj:
Issue has been solved at: #989

PR #989, opened 2015-04-20
Added logout redirect

a13ottne:
issue #931 solved


a13erilj:
I have read the code and tested it, everything seems to work.

PR #995, opened 2015-04-20
ds#969 database diagrams

a13karsi:
ds#969 Updated database diagrams to reflect what we're working with this week.


a13antda:
Well documentation is always good. We should fix so you can see which tables the diffrent groups are using. So that we can fix the "loose haning" tables with no relations. Or maybe put some of them toghether.

a13karsi:
Good idea! Right now the diagrams don't convey all the information that they could. Getting some more feedback from each group would be great so we can do better analysis and subsequent optimization. Maybe it would be best to wait with optimization on that level until we're at the point in the assignment where we're fixing stuff and not adding new stuff so we don't accidentally spoil someone's work.

PR #996, opened 2015-04-20
CodeViewer#994

b13danso:
Changes: 
Query uid from codeexample.
Store uid to a variable.
Use the stored uid to compare it to the uid of the logged in user.

Effect:
The if-statement now checks if the user is logged in and/or is the author/owner of the example or a superuser.

Edit:
Added code back that enables access to the cogwheel to users with write-permission on the course.

### DIFF
--- a/CodeViewer/editorService.php
+++ b/CodeViewer/editorService.php
@@ -58,7 +58,7 @@
 		// Perform Update Action
 		//------------------------------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


b13danso:
On hold.

b13danso:
Done now, awaiting review.

b13danso:
Closed, will test this and make another pull request next week.

Issue #744, opened 2015-04-08
It is not possible to choose versions courses


a13oscwe:
When choosing a course from **courseeed.php** the version of the course is undefined and it's not possible to change and view other versions without editing the URL.

a13oscwe:
branch ds#744 started to make **big** changes to the course user interface.

HGustavs:
I think that you can select which course to edit and which course to view in the course list view.

The idea behind this is to make it really difficult to accidentally edit an old version of the course (for traceability reasons).

Students should arrive at the most recent view chosen in the same dialog except for students that are re-taking a course according to an older course plan, in which case those specific students are given a link to that older version.

Students should **not** be able to "choose" which version they see as this adds to the confusion if we make it too easy to see "old" versions of a course.

b13alebr:
There are a few problems with how the versions of a course works. The first is that you assign access to a student to a course, not to only one verison of a course. This leads to that when you click on a course and then changees the coursevers parameter in the URL you can guess into another version of the course.

The other problem is that right now students who took the course last year can no longer go in and see the content from the version that was active when they took the course if a new version for this year has been created.

Ther is currently no good way for the theacher to view or edit other versions than the active one. A teacher can only go to different verisons of a course by changing "student version", but this also chages the link for all students.

The solution can be to assign access to a version of a course instead so that a student always can view that version, even a year later when a new verison of the course has been created. Like you said a student who took the course last year should come to the old version when he clicks on the link to a course. All students should not come too the most recent version like you said, they chould come to the version they have been given access too. Only the theacher should have access to all versions and should be able to jump between them.

PR #998, opened 2015-04-20
Issue #744 - Creating and editing versions

b13alebr:
Big change on how to manage versions of a course. You now manage versions inside a course (sectioneed.php instead of courseed.php), you can also jump between versions so it . The access-button is also moved to inside the courses.
Issue #744

### DIFF
--- a/DuggaSys/courseed.js
+++ b/DuggaSys/courseed.js
@@ -230,18 +230,12 @@ if (data['entries'].length > 0) {
 			}
 			str += ">";
 			
-		if(item['activeversion']){
-		
 			if (data['writeaccess']) {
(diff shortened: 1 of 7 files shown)


a13chrju:
It works like it should. all items are stored under the version that is selected and created from within the course. when adding new students, they get access to the whole course and not the specific version of the course but this can be fixed in antoher issue. 

good job!

b13alebr:
@a13denli  Can you look at this and check if it's OK to merge

a13denli:
Good!

Remember that indentation fixes will create merge conflicts for us at the weekly merge(so I wouldn't let this through, but my issue somehow got reverted so I will let this one get merged).

Issue #990, opened 2015-04-20
Student/Teacher view should not be accessible by guests.


a12oskri:
Currently student/teacher view is accessible by guests, guests should be redirected to the start page or be requested to login or be shown a error message.

a13erilj:
This has been fixed with pull request #999

PR #999, opened 2015-04-20
Umv17#990

a13ottne:
Fixed issue #990 
put loginprompt.php in Shared for use by all subsystems


a13erilj:
The code looks good and I have tested it and it seems to work. In a later issue these changes should also be added to studentView.php and teacherView.php.

Issue #1000, opened 2015-04-20
CodeView: Important Words fix!


a13alfbe:
Adding and removing important words broke in the last weekly merge

b12simsa:
Fixed in pull-request #1001.

PR #1001, opened 2015-04-20
CodeView #1000: Important Words fix! 

a13alfbe:
This fixes the important words and is incorperated with the existing code


b12simsa:
Looks good, a user can now add and remove words.

c12jenek:
The indentation is incorrect between line 196 and 230

PR #1002, opened 2015-04-20
Update init_db.sql in cv17 CodeView#972

a13jorla:
From issue CodeView#972. I have edited with a insert into template on row 225 so access can be done to template6. Also edited the database.


c13caral:
This works great!! We hade to change in the database to se the template6. I looks great! :+1:

c12jenek:
Approved to merge!

PR #1005, opened 2015-04-20
Umv17#997

a13freas:


a13antda:
Code: OK, not much that changed. 

Test: it works, and solves the issue.

Note: make new issue fix for total progress bar same way

PR #1010, opened 2015-04-20
Update dugga.js with keyup handler for esc key

c13andka:
Make so that when you press esc key the popup window will close


c13chrjo:
Looks good

Issue #1013, opened 2015-04-20
Template 6 not added to editorService


a13jorla:
Code is needed in editorservice.php to be able to set template 6 to a example.
![capture](https://cloud.githubusercontent.com/assets/11715474/7230235/13088228-e76c-11e4-97d9-5b809981c4cc.PNG)

c12jenek:
merged!

PR #1014, opened 2015-04-20
Update editorService.php issue #1013

a13jorla:
Made changes to be able to set template6 to a example.


a13henhu:
I've been looking into this request, and it does display template 6 on the editor

image

When clicking on it, it does generate a new template. I can confirm that the button is there, but if someone could double-check, I would appreciate before merging.

c12jenek:
Code looks good!

a13jorla:
Since changes have been made to editorService.php after my edit i will make a new pullrequest from the new weeklybranch and close this one.

Issue #1016, opened 2015-04-20
Sort courses over years in student view


a13alfno:
Fix so the student can see what year the courses are taken/when its due

PR #1017, opened 2015-04-20
Sort courses over years for sutend Umv#1016

a13antda:
Fixed so the courses can be seen for what year in the program the course is taken.
#1016 

http://webblabb.iki.his.se:8080/umv/a13freas/week-17/UserManagementView/studentView.php


f13marhe:
it only shows Year 1 in each col, why?. otherwise gj.

c13marme:
Code: Code looks fine, but there is a varieble with the name "Test" that should change. There still is some console.log code. indentation is fine

Test: Works like it should.

a13antda:
Fix the vatible test. and added so the courses are sorted after which term their are taken

Issue #1012, opened 2015-04-20
Parse data in teacherview.php to show correct classes in the dropdownmenu


c13marme:
Parse data in teacherview.php to show correct classes in the dropdownmenu

- Create a new teacherview.js file
- Fix ajax in dugga.js
- Parse data into the dropdownmenu to show the different programs and every year for that program(year 1 - 3)

c13marme:
This issue has been solved

PR #1018, opened 2015-04-20
Umv17# 1012

c13marme:
#1012 

Fixed the dropdownmenu to show correct classes for the teacherview.php


c13marme:
http://webblabb.iki.his.se:8080/umv/c13marme/week-17/UserManagementView/teacherView.php

Issue #977, opened 2015-04-20
CodeViewer: color scheme for hovering important words


c12jenek:
Create a color scheme for all to follow with the colors for hovering important words etc

a13jansa:
Iam on it!

PR #1027, opened 2015-04-21
Color scheme Hex 

a13jansa:
Color!
issue #977


c13caral:
There are some colors that you have missed.
example that "functions" is blue.
And you haven't added the red color on the border of important words.
skarmavbild 2015-04-21 kl 13 42 25

c12jenek:
Suggest changing some colors ( like the red border, and the fact that there's two important word colors, there should only be one ) and then resubmitting! :+1:

c13caral:
I think this looks great :+1:

PR #1030, opened 2015-04-21
Ds17#952

a12majca:
maybe this makes it a little bit easier to add new users.


b13alebr:
Works like it should and looks good!

a13denli:
Good

Issue #1029, opened 2015-04-21
Fix collumns in DB UMV


a13antda:
Fix collumns so to collerte with studenview.
insert more testdata.

c13marme:
This issue has been solved

PR #1031, opened 2015-04-21
Fix collumn hp in class table, toddler is teacher now

a13antda:
#1029


c13marme:
Code looks fine, and the new primary key works like it should, same with the foreign key
The code indentation looks fine in the SQL file.

Test: Works fine.

Issue #1028, opened 2015-04-21
fix information parse in studentwiev


a13antda:
Correct years for the courses taken.
add all divs for what is needed to look like muckup.
fix parse info on total progress bar.

PR #1032, opened 2015-04-21
Umv17#1028

a13freas:
#1028 
Correct years for the courses taken.
add all divs for what is needed to look like muckup.
fix parse info on total progress bar.


a13antda:
Removed "console.log" And added some comments.

Issue #1023, opened 2015-04-21
Fixed css and some html in studentView


b13frero:
Styling of progressbars and changing of names on ids and classes in css to match.

PR #1034, opened 2015-04-21
Umv17#1023

b13frero:
#1023


c13emmjo:
wrong this should not go in to master

PR #1035, opened 2015-04-21
Umv17#1023

b13frero:
#1023


c13marme:
There is some code in the css usermanagementview_student.css that use ID instead of Classes, we use classes most of the time. 

Progressbar2.css doesn't have any use.

a13antda:
closed to far behind weak breanch and solved in another issue..

Issue #1011, opened 2015-04-20
CodeViewer: Buttons appearing everywhere!


c12jenek:
![buttons](https://cloud.githubusercontent.com/assets/11715479/7229694/f949ce5a-e766-11e4-8b4a-ead3ca56dfde.jpg)
These buttons suddenly appeared, find out where and how they arrived

HGustavs:
The error handling is less than ideal....

When looking at the post in firebug I see the following:
```
<br />
<b>Notice</b>:  Undefined offset: 13 in <b>/var/www/html/codesys/cv17/CodeViewer/editorService.php</b
> on line <b>224</b><br />
```
This is due to an error on line 224 that is not propagated to the service as an error.

Ideally this error is propagated to the server which then receives a 5x0 error which is shown as an alert.

so the error is an uncaught exception.

Either the error can be fixed in situ or error handling be added in addition to the fix.

HGustavs:
In addition to that fix.... it would be a good idea to also by-default hiding most of those buttons in case there are other similar errors.

b13danso:
The error was not on line 224, the problem itself was that the query that collected all before/after-ids was selecting all the exampleids with the same courseid as the set cid, ordering the the output by sectionname and examplename, not by id. The temporary fix for now, because we have different courseids, is to select all data from codeexample and order id by exampleid, this seem to fix the problem and aswell allowing us to navigate with the forward/backwards buttons and dropdown menus on them(by holding down mouseclick).

From:
$query = $pdo->prepare( "select exampleid, sectionname, examplename, beforeid, afterid from codeexample where cid = :cid and cversion = :cvers order by sectionname, examplename;");
$query->bindParam(':cid', $cid);
$query->bindParam(':cvers', $cvers);
$query->execute();

To:
$query = $pdo->prepare( "select exampleid, sectionname, examplename, beforeid, afterid from codeexample order by exampleid");
$query->execute();

b13danso:
Note: I think the problem was that the data stored in the database.

Example:
Codeexample: 
exampleid=14, beforeid=13, afterid =14, cid=2
exampleid=13, beforeid=12, afterid =13, cid=2

Query: take out all rows with cid=2 from codeexample
Result: 2 rows.

The rows gets stored in each array and then put into one single array.

[...]

PR #1037, opened 2015-04-21
Cv#1011 - Navigation problems with the buttons

b13danso:
Fix for issue #1011, read comments there for more info!


c13caral:
This works great!
This allowing us to navigate around and use the buttons :+1:

Issue #1019, opened 2015-04-20
the "Files" page is not working and is a blank page


a13chrju:
the page as of now is blank with no information whatsoever.

the javascript file is constructed in a way that the buttons for adding new files and links are only displayed if there is a file created for the chosen course. the buttons should be there even if no files exists otherwise you can't add any at all if you dont do it directly in mysql. 

files don't even get uploaded and directories aren't created after inserting a new file or link, only "insert" statements are completed and imported to the DB.

furthermore, links to the stylesheets and javascript files aren't correct.

HGustavs:
I believe that this may be due to problems with permissions.

The filereceive service should write the file if it is able due to permissions.

HGustavs:
Showdoc is used to read the file.

Similar functionality to showdoc is needed for issue #1008

a13chrju:
yes, I will (or someone) have to fix the "showdoc.php" file that reads the files aswell

HGustavs:
It should work by the file name, if the file name you give exists as a global file you get that file but if there is a local file with the same name it gets precedence. As for local and global I think that the kind of file is decided on upload.

This way you can have newer versions of a file, and older versions of the same course see the old file....as long as the file name is the same.

a13chrju:
aha okey, yeah I had to change a lil to have the buttons up from the beginning and the global function kinda stopped working but I fixed it with the attribute. I will look into it

a13chrju:
might come a follow up on this later

PR #1044, opened 2015-04-22
Ds17 #1019 - adding "Files" page fix

a13chrju:
with this code the buttons for adding files are displayed even if there is no file in the database under the course.
Javascript and css links are fixed.
Directories were created for where global and local files will be stored in.
couldn't upload files before, now you can. All global files will show up on all courses which is why I needed an attribute "isGlobal" which is either 0 or 1. if it is a global file the attribute is equal to 1.

### DIFF
--- a/DuggaSys/fileed.js
+++ b/DuggaSys/fileed.js
@@ -87,125 +87,114 @@ function hideLoginPopup()
 //----------------------------------------
 function returnedFile(data)
 {
+
 		filez = data;
 		// Fill section list with information
-		str="";
+		str1="";
+		str2="";
+		str3="";
+		str4="";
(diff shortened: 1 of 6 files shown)


a13chrju:
The attribute I need for this pull request was added in this one from Kári: https://github.com/HGustavs/LenaSYS/pull/1042

a13chrju:
https://github.com/HGustavs/LenaSYS/issues/1019#issuecomment-95246353

a13denli:
Remove deadcode(example you have some outcomment code in DuggaSys/filereceive.php)

a13chrju:
I removed some dead code :+1:

a13denli:
Good (Y)

PR #1058, opened 2015-04-24
Ds17#1048

a12majca:
mailto functionality for those who want to get the duggareceipt to there email. the email is not stored in the database as it is now.

### DIFF
--- a/DuggaSys/showDugga.php
+++ b/DuggaSys/showDugga.php
@@ -132,32 +132,23 @@
 			
 	</div>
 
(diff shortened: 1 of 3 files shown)


a13chrju:
gets sent to the email that the user puts in. works like a charm

a13denli:
Remove the style.css commit(indentation fixes shouldn't be here, atleast not shared files)

a13denli:
I will not merge this to our branch, so you will have to fix this to next week.

Issue #960, opened 2015-04-16
CodeViewer: Template 4 & Template 5 breaks when resizing the browser window


c13caral:
![39eb2074-e421-11e4-9853-1ff181249f6d](https://cloud.githubusercontent.com/assets/11715450/7183436/b9c20e1c-e457-11e4-8ed4-31394fcf68e3.png)

c13caral:
It also breaks when you open the webbpage and then increase the size of the browser.

![fel](https://cloud.githubusercontent.com/assets/11715450/7184079/8e269008-e45b-11e4-8791-4fd1351f7848.png)

a13jorla:
Merge is done and worked good.

PR #1059, opened 2015-04-24
Fixed browser resizing problems. Issue: #960

b12alela:
The problem boiled down to jQuery having a prefrence of doing everything pixelbased, and the previous functions were set to work on exact measurments in pixels. However this is now resolved, all measurments as far as i can see have been converted to percentage based ones. Which solves the browser resizing problem.


c12jenek:
This will be merged next week

Issue #1055, opened 2015-04-23
accessed.php - Adding a user gives error message and email is not saved in database


b13alebr:
When a user is added you type in the user's email address, the first part before the @-sign is saved in the database as username but the complete email address is not saved at all, the column "email" in the database has null as value. Also when the user is created a pop-up windows says that the user does not exist.

a13karsi:
Ah, we might need to add the @ character to the regular expression in dugga.js in  /shared

a13chrju:
but it works to insert a @ into the database they just forgot to add the whole input field into the email attribute, I got it to work :)

a13karsi:
Awesome, I just checked the regular expression did allow @ so no worries about that part!

a13chrju:
fixed in https://github.com/HGustavs/LenaSYS/pull/1062

PR #1062, opened 2015-04-24
Ds17 #1055 - storing the email and error fix

a13chrju:
the reason why you always get an error when making a new user is because the debug attribute always get input, it's not under an if - statement. in the accessed.js, the data contained in the "debug" field gets alerted if it's not empty which it never was.

also added so the email gets stored.


a13denli:
Good!

PR #1066, opened 2015-04-24
Ds17#506

b12erijo:
Implemented a timer in dugga1(will be added to all duggas later). Added placeholder box and button to show highscore. Nothing complete yet, will be continued next week.


a13denli:
Good!

PR #1021, opened 2015-04-21
Code viewer saving descriptions as file with markdown text instead of text in the database #473

a13henhu:
Now it's able to read from a .txt file.

#473 

And also creds to @b12simsa for working togheter.

Issue #668, opened 2015-04-06
editorservice contains code standard errors


HGustavs:
editorservice.php should not have passed the review for the merge as it breaks the code standard repeatedly.
* It contains numerous commmented-out sections. We do not use /* for commenting out things in the code, only for headers and box-commenting code for internal testing. Commenting out things is something that we do only for internal testing. The merged branches should not have this.
* Also, the indentation seems off in places but I am unsure about this.

a13jorla:
Since there is many issues about codestandards and this one was specific about editorservice.php i'm closing this issue. Cause it has been merged.

PR #1067, opened 2015-04-24
Cv servergitsync

c12jenek:
Some of Henry's changes in merge #1021 was not on git but remained on server, this merge is a sync of those two versions


c12jenek:
Also, the up to date variablename changes were causing issues, reverted for now

b12alela:
There doesn't seem to be any issues with this version. No breaks witnessed in the codeviewer site, should be nothing else since only editorService file is edited notably.

b12simsa:
The example buttons are working now. Will make another attempt at "fixing" variable names next week (issue #668).

Issue #1069, opened 2015-04-24
Style the studentview


c13emmjo:

c13marme:
This issue has been solved

PR #1070, opened 2015-04-24
did #1069

c13emmjo:
#1069 

Fixed the styling for the student view

studentview:
http://webblabb.iki.his.se:8080/umv/b13frero/LenaSYS-umv17/UserManagementView/studentView.php


c13marme:
Code: Looks fine , only thing is that div in the php file is year1,year3 and year2. But on the site is still shows correct order for every year.

Test: Works

PR #1080, opened 2015-04-27
Code view#789

a13jorla:
Makes it able to choose/change template for an example.


b12alela:
The changes look solid, although there are alot of moving parts on this one. I think these changes fixes some earlier mistakes in the code which made the changing of templates to break. Creds to Jorgen for finding these.

Issue #1085, opened 2015-04-27
Content inside the course pages does not display.


c13chrjo:
Data is there in the database but it does not get retrived by the webpage.
The problem COULD be that someone has change the name of something inside the database.

We have recreated this problem localy and the problem persisted only when using the new database. The old one still functioned like before.

PR #1091, opened 2015-04-27
Fixed faulty sql query

a13karsi:
Changed 'COURSE' to 'course' since mysql is case sensitive on our platform. issue #1085


a13denli:
I'm still not sure how this could cause all the problems, but it's working.

PR #1098, opened 2015-04-27
Updates to Resizable windows!

b12alela:
This update fixes the breaking windows caused by pixel based positioning, by exchanging measurments used for percentage based ones. This also fixes the first time experience of the resizable windows, due to used uninitialized variables the windows behaved very strangely for first time visitors. This should now be solved.


a13jorla:
Everything works fine and all examples with same template has the same size on boxes, if you choose new example.  And no breakup between boxes.

PR #1116, opened 2015-04-27
issue 1109 - basic.php have typing error causing inserts to a column that doesn't exists

b13alebr:
added a "d" so the query inserts to the right column.


a13denli:
Good

Issue #873, opened 2015-04-14
CodeViewer Markdown Ordered and Unordered Lists


f12danjo:
Implement ordered and unordered lists with markdown in the CodeViewer.

PR #1123, opened 2015-04-27
Code view#873 markdown list

b12simsa:
#873 
Using regular expression to replace certain symbols with html list tags.

For ordered lists type any integer value followed by a dot and then some text, e. g.
1. entry 1 
2. entry 2
4. entry 3

For unordered lists type a line, followed by a space, followed by some text e.g.
- entry 1
- entry 2

If you are unsure if the input is parsed correctly try the same input here http://spec.commonmark.org/dingus/. If the our parser translates input differently it is regarded as unexpected behavior.


b12simsa:
Since I am trying to merge this branch into master, I will close this request and reopen it.

PR #1125, opened 2015-04-27
CodeView#796:  Edit Content

a13alfbe:
This fixes and makes the edit content boxes usable. With this implementation you should be able to change the following for each individual "code-box":
* Title
* Kind
* Wordlist
* FIle
* Remove/Add important rows

I also redesigned the box a bit.


a13alfbe:
Found something that i will make into another issue. 
As of now, when the box "Kind" is Document it doesn't change the file directory to the descupload.
The wordlist should also be disabled when "Kind" is Document.

b12alela:
The KIND option HTML in the edit box, totally breaks the box its supposed to edit. This is due to the HTML feature not being fully implemented in previous versions of LenaSys. Other than that the changes are solid and i propose we comment the HTML option out and merge the code.

PR #1135, opened 2015-04-27
Update editorService.php

b12simsa:
#668
Since I think it is risky to change the name of variables and it did introduce some errors last week, this pull request only changes some of the names so it is a semi fix of #668.


c13caral:
Everything looks like it works. Not all the variables is changes just like b12simba says. But is safer to do this in this way :)

c13caral:
I will merge this

PR #1139, opened 2015-04-27
Update showDuggaservice.php

c13andka:
Change to 1 explode


a13denli:
The extra temp is needed to make it work with other version, so good

Issue #1124, opened 2015-04-27
CodeViewer: Missing inserts in testdata.sql


c12jenek:
There's some missing inserts in this weeks testdata sql file

c13caral:
this is now solved :+1:

PR #1144, opened 2015-04-27
Cv#1124 missing inserts

c12jenek:
Added missing inserts, added the updates' data into the inserts and then removed updates #1124


a13jorla:
The changes makes less code that is good and and follows sql syntax. But why just comment away code and not delete it?

c12jenek:
Very true, should be automatically added at the merge, will remove instead

c13caral:
It looks great! :+1:  Its ready to merge!

PR #1146, opened 2015-04-27
Implemented Highscore

b12erijo:
Highscore is now implemented but how the score is actually going to be calculated
is not yet decided. As of now it simply uses the time spent.


a13denli:
Good!

Issue #1159, opened 2015-04-27
id´s and data for course requierments


a13antda:
Set id´s to course divs that makes it easeier to see requiermens for the difrent courses.
Added some info for the DB script. 
added php code for the connection.

a14emeax:
I closed this since this is already merged and also looks more like a note of what have been added rather than an actual issue.

PR #1160, opened 2015-04-27
Umv#1159

a13antda:
#1159 
 
ADded a base for the course rquierments. Its not all done yet. just the base, Jquery is needed


f13marhe:
looks good what i can see.

Issue #1163, opened 2015-04-27
only admins can see the content of courses


a13chrju:
members without write access can't access their tests under a course.

a13oscwe:
Fixed in pull request #1164, closing

PR #1164, opened 2015-04-27
Ds18 #1163 members can now see tests under a course

a13chrju:
members can now see tests under a course - https://github.com/HGustavs/LenaSYS/issues/1163


a13denli:
Good

Issue #865, opened 2015-04-13
Do the smal progressbars in the contentoverview (courses completed and not) in teacherView.php


c13emmjo:
Look at the teacherview mockup. (Not radiobuttons) 
also procent so the teacher can see how many procent has completed that course.

c13marme:
This issue is solved

PR #1165, opened 2015-04-27
umv18#865

c13emmjo:
Did issue #865

teacherView:
http://webblabb.iki.his.se:8080/umv/c13emmjo/week18/UserManagementView/teacherView.php

(The hover effects on the progressbars are not 100% finished and need to ask the customer aswell)


c13marme:
If it the credits reaches over 7.5(100%) it will incorrect in the bars, like 15 points (200%). Hover effect works fine.

Indendation looks good, and I tested the hover effect.

c13marme:
Will create a new issue for the credit problem.

PR #1166, opened 2015-04-28
Ds18#1048

a12majca:
hope the indentation is ok now, let me now if it isn't..


a13oscwe:
Tested, doesn't break anything and looks good with the new indentation.

Issue #1167, opened 2015-04-28
Correctly aligned "edit-button" and title for codeexamples.


a13alfbe:
The edit button and title are missaligned and looks weird.

a13henhu:
Velly gut

PR #1176, opened 2015-04-28
CodeViewer #1167: Missaligned titel and editbutton for codeboxes 

a13alfbe:
This fix also removes some unnessesary borders and corrects some height values for a boxtemplate.

Compare this to the old version and let me know what you think


a13henhu:
After checking this request a few times, I've noticed the changes, and how it aligns correctly now, and actually looks better.

PR #1178, opened 2015-04-28
Codeviewer: #828  Print indentations

c13caral:
Issue #828 
This changes makes it possible to print indentations in the code examples by using "tab" and not 4 spaces.

+ changed to follow the codestandard


a13alfbe:
I have tested the tab functionality and it works great!

Issue #1171, opened 2015-04-28
Style ->  "NO STUDENTS FOUND THAT ARE REGISTERED TO THIS COURSE" 


c13emmjo:
in teacherview the text "NO STUDENTS FOUND THAT ARE REGISTERED TO THIS COURSE" are shown sometimes. Style this so it looks better

PR #1181, opened 2015-04-28
Fixed the style on no_page text #1171

f13marhe:
#1171


f13marhe:
#1171

a13antda:
Looks better. But change the error messages to make more sense. like "no students found for this class" or "no data found for students in this class" someting like that.

Issue #1120, opened 2015-04-27
Creating main progressbar to studentView


b13frero:
Coding and styling main progress bar for studentView.
Uppdated css for progressbar2, js functions for calculating the bars progresses.

PR #1185, opened 2015-04-28
Umv18#1120

b13frero:
#1120


b13frero:
http://webblabb.iki.his.se:8080/umv/b13frero/LenaSYS-umv18/UserManagementView/studentView.php

Issue #1184, opened 2015-04-28
Quick fixes on high score list


a13karsi:
The high score list lists the same person many times if that person has taken a dugga many times, it looks nicer if the list only shows the best score (quickest time) for each person and can list more user names associated with scores.

HGustavs:
I really like this service.

...but...

Not all "dugga"-s are time-based.

It would be nice if we could score differently on different "duggas" i.e. by the javascript submitting a "score" depending on factors... such as the number of clicks in bit-counting. The more clicks the more guesses.... which should give lower scores.

a13karsi:
That's a good idea. I think we were going to implement the timer in all of the duggas but your click counter sounds like a really good idea too.  This service is just at its infancy just yet though so there's a lot of room for improvement!

HGustavs:
Im thinking that each dugga is responsible for its own scoring e.g. that we supply a framework and example code for submitting highscores, and that each dugga decides what to submit in some cases it is time, in some case clicks and in other cases something completely different.

a13karsi:
I'll make sure to talk with Erik about this (He's the magician behind most of the highscore code)

PR #1187, opened 2015-04-28
Update highscoreservice.php

a13karsi:
Added a group by userAnswer.uid to query to display only users only once. Shows the best time for that user. See issue #1184


b12erijo:
Makes sense!

Issue #457, opened 2014-10-20
It should not be possible to resubmit dugga if passing grade has been set (e.g., G or VG)


a97marbr:

a13chrju:
fixed in PR https://github.com/HGustavs/LenaSYS/pull/1189

PR #1189, opened 2015-04-28
Ds18 #457 - resubmit dugga not possible if passing grade has been set

a13chrju:
if the the assignment is graded the user can no longer submit it(graded G, VG, 3, 4, or 6). But if the grade is set to U (failed), the user can still submit it.

https://github.com/HGustavs/LenaSYS/issues/457


a13denli:
Good

Issue #833, opened 2015-04-13
CodeViewer: Part 2 of issue #825, creating new examples


c12jenek:
Part two of Henke's issue #825 ;
"It would also be nice to have a way to create an example from an example dialog....e.g. to create the next example for the course without having to return to the course view."

a13jansa:
I will work on the first part of this issue. New button + pop-up box

a13henhu:
I need to focus on one issue at the moment, so I will unassign myself from this issue for the moment

a17poner:
long time has passed not sure if this is still relevant, closing this and the other part and using #825 to ask henrik if this has been "silently fixed" in the last 4 years and/or is still relevant.

PR #1191, opened 2015-04-28
Part 1 #833

a13jansa:
this is the first part of issue #833, i made a new button and a pop-up box


c12jenek:
Some changes here removes existing code that's not part of this issue (some of carina's implementations) so this branch needs to be updated to a newer version before it can be merged in

a13jorla:
The button does not show right in chrome. And what is Edit fdsfds in the header in the pop-up box?
Maby one should be able to add contents into a number of boxes. Using the template button that now is in the navheader.

c12jenek:
This will be be merged in time for weekly merge thus it'll be closed and have to be reopened next week

Issue #1190, opened 2015-04-28
CodeViewer Markdown hr line


b12simsa:
Add the possiblity to add a line with three '-' symbols.

c12jenek:
merged in and working!

PR #1193, opened 2015-04-28
Codeview#1190 hr line

b12simsa:
#1190
Adds the possiblity to create a hr line with three '-', (also re-adds some lines of code that disappeared in a merge).


c12jenek:
Horizontal line is there and working! Code standard is follwed, eveyrhting looks great!

Issue #1192, opened 2015-04-28
sectioned.php becomes blank if links have special characters in their names


a13chrju:

c13chrjo:
Fixed

PR #1194, opened 2015-04-28
ds18 #1192 - becomes blank if links have special characters

a13chrju:
https://github.com/HGustavs/LenaSYS/issues/1192
beware, the code might be complicated to follow.


a13karsi:
You maniac!

c13chrjo:
Looks ok

PR #1199, opened 2015-04-28
CodeViewer#1188: Edit Content document fix

a13alfbe:
You can now change between Code and Document and the directory containing the correct files will be available. 

When "Document" is choosen the wordlist defaults to "Plain Text".


a13jorla:
The code  follows codestandards and works as the issue suggests. But there is a issue when saving, one have to manually reset the page to see changes who has been done. But as that will be a new issue i will merge and  make new issue if it does not exist allready.

Issue #1148, opened 2015-04-27
Duplicate code in AJAXService


b12erijo:
The same block of code gets run twice in AJAXService, resulting in each parameter being sent twice to the service.

HGustavs:
I assigned highpriority and bug to this issue to get it confirmed quickly... as bugs in this code may affect all parts of the system...

a13karsi:
This is confirmed, apparently we're getting the para string twice. I've looked at the code and I think I have a fix in mind which would accomodate the old and new code (which I don't really get, people need to comment more!)

However, I'm waiting for a pull request of mine to be merged which updates dugga.js before I can submit a new update.

PR #1200, opened 2015-04-29
ds#1148 Update dugga.js

a13karsi:
Fixed bug causing duplicate parameters. See issue #1148 , I've tested this fix myself and seems to work but please test this fix thoroughly before merging as well.


a12majca:
the problem is solved, well done.

Issue #1141, opened 2015-04-27
Placement of elements should be fixed in fileed (firefox)


c13chrjo:
It should look the same on all browsers. This looks wrong on firefox

c13chrjo:
Fixed

PR #1202, opened 2015-04-29
Ds18#1141 - Placement of elements should be fixed in fileed (firefox)

a13chrju:
Looks the same now in all the browsers I tested (IE, Chrome and firefox).
https://github.com/HGustavs/LenaSYS/issues/1141


c13chrjo:
Looks fine

PR #1203, opened 2015-04-29
the duggasysmenu is now easier to use on a smaller device. we still

a12majca:
we still need to make the loginBox responsive, maybe a new issue for that. i
also made some changes to how the different menu choices are displayed,
on the left side there are buttons for editing the course version. on
the right side we have buttons that leads to a new page. ”New” button
is now called ”New Item” and also moved so it’s easier to understand
that it has to do with adding items to the sectionlist.


a13denli:
Good

Issue #1204, opened 2015-04-29
Style the studentview and teacherview and the contentwrapper


c13emmjo:
Fix so the contentUMV is 100% in the height aswell

Hoover effect on the smalprogressbars in studentview 

And change so the color on the progressbars are the same

PR #1206, opened 2015-04-29
umv18#1204

c13emmjo:
Did issue #1204

Fixed the contentUMV wrapper so the height always is 100%

Put in a div in the studentview so the css is working to it

Changed so it is hoovereffects on the smal progress bars in studentview

Changed the student title so it looks like this way instead: (studentview)
http://webblabb.iki.his.se:8080/umv/c13emmjo/week18v2/UserManagementView/studentView.php

Also changed the color in teacherviews progressbars so it is the same
as it is in studentview

And here is the teacher view:
http://webblabb.iki.his.se:8080/umv/c13emmjo/week18v2/UserManagementView/teacherView.php


a13antda:
The css is comming along, There are still changes to be made on the front interms of coloring. 

The indentation on usermanagementview.js line 168 should be fixed before merging. The line is far to long to follow.

a13antda:
Seen the fixes and noticed the relocated files. Good work.

PR #1208, opened 2015-04-29
Update accessed.js

c13andka:
Standards that i have changed the code after that it did not follow.
     - only for very short code we allow single line if-statements and never allow else with single line if-statements.


a13denli:
Good

PR #1234, opened 2015-05-04
Update studentView.css

b13frero:
comments fixed


a13erilj:
Seems to be a few issues with the studentView due to these changes
Fix them before merge :+1:

Issue #1241, opened 2015-05-04
Move the bigprogressbar up so it is closer to the header


c13emmjo:
Move the progressbar so there is less space. Also change the username so it is smaler and ex lays in the left top instead.

PR #1250, opened 2015-05-04
Umv19#1241

b13frero:
#1241


a13erilj:
Everything looks fully functional. @c13emmjo has checked your code and it seems ok

Issue #1048, opened 2015-04-23
mailto self link/button in dugga


a12majca:
Henrik says: I would like a mail-to self link/button on this dialog also so that the student can email the link to himself/herself.

It would be nice if the mail-to could remember which email address one enters (in local storage?) so that one does not have to enter a new email address every time one submits an answer.

c13andka:
You mean when you save the result of the dugga? that he/she can send the information to the persons mail?

PR #1253, opened 2015-05-04
ds19#1048 - answer to "mail-to could remember which email address one enters (in local storage?)"

a13chrju:
Email gets stored in local storage. https://github.com/HGustavs/LenaSYS/issues/1048


b12erijo:
Indentation need to be fixed before merging.

PR #1254, opened 2015-05-04
Codeviewer #668 codestandard editoerservice.php

b12simsa:
Fix for #668.
Changed the name of variables to follow codestandard.


a13alfbe:
Seems to work and only local vairables are changed. This shouldn't effect any other parts of the webapp.

f12danjo:
Looks good. You had missed one variable: "imp". Changed it to "importantRows".

Issue #1228, opened 2015-05-04
Codeviewer Markdown create documentation


b12simsa:
Create a document describing the markdown implementation, also review the comments written in the code (codeviewer.js).

HGustavs:
It would be super-nice if it would be possible to show a markdown legend by clicking somewhere in the markdown editor... like a dropdown or something like that which lists the available markdown strings and their effect.

It should be possible to show this without losing what you have written so far.

b12simsa:
Right now you can´t really write anything directly in the editor, you rather create a file, upload it to the folder, and connect the filename with the example in the database, the content of the file is then loaded into the description box. Maybe this is another thing to discuss during a customer meeting?

f12danjo:
Closed since documentation for markdown has been implemented. A markdown legend is not relevant for the code viewer as it functions right now. If we were to implement an editor this would become relevant.

PR #1256, opened 2015-05-04
Codeviewer #1228 markdown

b12simsa:
Uploading a pdf manual for the markdown parser.
Some minor fixes for the markdown parser.


a13alfbe:
You probably need a newer version of codeviewer.js before you try to implement these changes.

Issue #1239, opened 2015-05-04
Change the radiobuttons to image 


c13emmjo:
In teacherview the radiobuttons are now styled in css. Should create image instead so the dot always is in the middle.

PR #1257, opened 2015-05-04
umv19#1239

c13emmjo:
Did issue #1239

Changed the radio buttons in teacherview to svg images

http://webblabb.iki.his.se:8080/umv/c13emmjo/week19/UserManagementView/teacherView.php


a13erilj:
There is a issue with the image in Safari. It gets partially cut.
However this is a late problem and not a critical issue

Issue #1225, opened 2015-05-04
Code Viewer: Clean code


c13caral:
There are a lot of unused functions and variables. By cleaning the code makes it easier to read and understand the code.

HGustavs:
In PHP unused is fine if it corresponds to the service.

There should always be variables that correspond to each parameter sent from the client e.g. if the client sends a cid and a version we should have variables for those parameters so that the php correpsonds exactly to what is sent from the client even if it is not used at the moment.

I agree that there is stuff that is not used, particularly in the code viewer javascript.

HGustavs:
By the way I think this is a duplicate of issue #464 and issue #465.

It is a good idea to go through the issues from the back and see if there are other duplicates.

c12jenek:
Code has been cleaned as much as currently possible, this issue is then solved

PR #1258, opened 2015-05-05
Cv#1225 2

c12jenek:
#1225 second file to get a small clean up


f12danjo:
Nice with proper indentations! Easier to read and understand the code, looks good.

a13alfbe:
Still works as it should and nice that you changed the name of "cnt"!

Issue #1158, opened 2015-04-27
Unexpected behavior by comments of // type 


a13erisc:
When these comments are used, linenumbers increase by 2 instead of 1.

HGustavs:
Great catch!

HGustavs:
It has to be one of the row++ lines that is over-adding. My prime candidate is this section starting on line 970. 
Try to remove the row++ there and see what happens, if it is better, worse or the same.

```
	        c=instring.charAt(i);
            if (c=='\n'||c=='\r'||c=='') {
                row++;
                break;
            }else{
                str+=c;                
            }
            i++;
```

a13erisc:
Yes, that solved it.

PR #1260, opened 2015-05-05
CodeViewer #1158

a13erisc:
Solved a very small bug which might affect future implementations. Read more about it at #1158


a13jorla:
Where did you place the // comment? to make the line-no increase by 2.

f12danjo:
Nice. Fixes a bug that would be a problem when implementing a buttonborder for collapsing and expanding if-cases and such.

Issue #470, opened 2015-03-22
Support for trees in the code viewer


HGustavs:
One stretch goal feature that would be really appreciated is support for trees in the code e.g. allowing one to click a small plus icon near a function or block which would expand or contract that block to a single line.
e.g. 
`+ if(condition){ 

...   

random code  

...

}

...if we click the plus...

it would be turned into

`- if(condition){ ...`

and if we click the - we would return to the above.

a13erisc:
I've decided to split this issue into several parts, because of its complexity. Right now I've set up a pull request that will cover the first part, which adds a border with clickable buttons next to the code, see #1266. The second part would definitely require group effort.

a13jorla:
We are putting this issue on ice for now. And we are letting other course work on this issue.

a17poner:
@HGustavs is this still relevant and something you want us to work on?

a17poner:
Finally fixed 4 years later, all praise lord joaap!

PR #1266, opened 2015-05-05
CodeViewer #470, Part 1

a13erisc:
This is a very difficult issue, which I've decided to split up into parts. This part will at least implement a new border next to the existing codeborder which holds buttons for each codeblock. See issue #470 for more.

### DIFF
--- a/CodeViewer/codeviewer.js
+++ b/CodeViewer/codeviewer.js
@@ -1265,6 +1265,7 @@ function rendercode(codestring,boxid,wordlistid)
 	bracket=new Array();
 	cbcount=0;
 	cbracket=new Array();
+	ts = new Array();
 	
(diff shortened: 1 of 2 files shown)


HGustavs:
Im thinking we can leave this issue as a "stretch goal", there are more pressing issues to attend to.
If we run out of other issues to work with, we can come back to this issue.

f12danjo:
Looking good. Like you said, it adds a side border where "+"-buttons can be added in the future. Fixed a minor thing so your code follows the code standard: The first bracket of a function is supposed to be on a new line.

a13alfbe:
* There is no need to use float:right when the element is absolutely positioned.
* I found that the height of the buttonborder isn't 100% of it's parent when you scroll the content of the codebox.

f12danjo:
We will close this because, like HGustavs wrote, we should concentrate on other, more high priority issues. If this were to be implemented again, do a new branch from the weekly branch so it's up to date.

Issue #1265, opened 2015-05-05
Give creators of dugga's a choice of how to count high scores


a13karsi:
Henrik mentioned in a previous issue about dugga's that it would be nice to allow a user in an administrator or 'teacher' role to select what type of method to use when calculating high scores for the high score list.

This issue focuses on updating the current dugga creation window as well as any connection to the database to allow the user to select which kind of highscore calculation method to use, be it based on time, number of clicks etc., 

We want this part of the system to be modular and easily flexlible so that more methods of counting high scores can be added later.

PR #1267, opened 2015-05-05
Highscore modes update

a13karsi:
When a user creates a dugga list item, the user is now presented with the possibility of selecting how to count high scores. 
The database table 'listentries' was updated to include what kind of high score mode should be associated with a dugga, currently my thoughts are
    0. None
    1. Timer based
    2. Click based

See issue #1265


b12erijo:
Can't see anything wrong with the code. Looks good! The if cases checking which highscoremode is selected could possibly be if-else instead to optimize but that could screw up the order a bit i guess.

Issue #1242, opened 2015-05-04
Change the username styling in the shared header


c13emmjo:
Change so the text isnt yellow - do the styling more attractive

a13erilj:
Begin by changing the main style, hover effects for entering UMV can be handled in another issue

PR #1271, opened 2015-05-05
Changed color on username and added some css rules

a13alfno:
#1242 
Changed color to white and removed underline if not hover. 
Added .navName:hover for later use of hover. 
Link: 
http://webblabb.iki.his.se:8080/umv/a13alfno/Version%200.4/DuggaSys/courseed.php


a13erilj:
Worth to note is that a shadow appears when hovering over the username.

PR #1272, opened 2015-05-05
Ds#506 timer

b12erijo:
Modularised timer and added a graphical representation. Also implemented it in all duggor.


a13denli:
Dugga3.js has an extra whitespace but not dugga2,js ;) Also remove console.log()

PR #1276, opened 2015-05-05
Cv#1225-4

c12jenek:
#1225 Cleaning some more!


c12jenek:
Hotdogmenu code will be removed in this merge, if this is still desired by customer ( Henke ) please refrain from merging

a13alfbe:
I will wait with this merge until we here from the customer about the hotdog menu.

c12jenek:
Added the hotdog code back into the branch, added comments on why it's still there but never used

f12danjo:
Nice addition of line number were functions are called. Even if these lines changes in the future (most likely) you will at least have an approximation of where the function is called, which is better than having none.

Issue #1207, opened 2015-04-29
In studentview fix the smalprogressbars so they are green when the student have completed all the hp in that course


c13emmjo:
In studentview the smal progressbars are now always "orange" they should be green when it is ex:
7,5 / 7,5 hp 
(like it is in teacherview)

PR #1278, opened 2015-05-05
Change background on progressbar.

a13freas:
Changes the background color to green when the student have finished the course.
#1207


a13alfno:
Ser bra ut. Fortsätt så!

Issue #1211, opened 2015-04-30
CodeViewer: Codeboxes bug


c12jenek:
The codeboxes works fine for one box, but if you wish to have several ones, it's merged into a huge one, e.g.
markdown for codebox
code
code
code
markdown for codebox
text
text
markdown for codebox
code
code
code
markdown for codebox
There should be two boxes, but when running examples there is only one

PR #1281, opened 2015-05-05
Code blocks inclusion and bug fix

f12danjo:
The code for code blocks with markdown has yet again disappeared so I'm adding it back in. This fixes issue [#1211](http://github.com/HGustavs/LenaSYS/issues/1211).


a13alfbe:
Tested and works fine.

Issue #1084, opened 2015-04-27
Debug alert does not seem to work in code viewer


HGustavs:
Some time ago an alert was added for any situation that results in the return of the "debug" object.

It seems like either the error exception is not implemented quite correctly in the code viewer or that the ajaxservice does not alert this error message to the user.

Either way, the error message from the code viewer service is not shown to the user.

```
 Object { debug="ID does not exist"}
```

HGustavs:
This is part of why #1076 breaks. 
If we get an error message we at least know better whats up.

PR #1282, opened 2015-05-05
Error not being alerted when debug is returned

b12simsa:
#1084
Moved the if-statement it is now reached before any attempt to access an element in the array is done.
The script will still break if the debug is returned, but the user will receive the pop-up.


a13alfbe:
Great, it now shows the debug message when you try to access a codeexample that doesn't exitst for example.

PR #1299, opened 2015-05-06
Update testdata.sql

a13alfbe:
- Changed the values beforeid and afterid for the codeexamples so they all are in the correct order.
- Changed the boxid  and example id for the boxes, they were mixed up.
- Removed some important words that caused bugs. There are issues adressing these bugs.


f12danjo:
Fixes the bugs and errors that we had regarding the database and important words. Looks good.

Issue #1285, opened 2015-05-05
Calculate high scores based on user specifications


a13karsi:
Continuing from issue #1265 .

Now a user that creates a dugga in the section list can specify how the dugga should calculate high scores. We want to incorporate the different methods that have been identified to calculate high scores so that they are only activated if they have been selected. 

For instance, if a timer based high score calculation is selected for the dugga, use the timer to store high scores. If something else is selected, use that selection instead.

PR #1304, opened 2015-05-06
Update high score modes

a13karsi:
This update allows the user to select between timer and click based highscores for dugga1.js. Both high score mode calculations are fully functional. Will be refactoring and updating more in next issue. 

See issue #1285


c13andka:
Seems legit

Issue #1301, opened 2015-05-06
Code Viewer: Adding/removing important rows is broken.


a13alfbe:
Adding and removing important words doesn't work anymore.

a13jorla:
Now row highlighting works again.

PR #1311, opened 2015-05-06
Code view #1301   adding/removing imp rows

a13alfbe:
#1301 
The openBoxID was not a global variable.
The main problem was that the list with codeexamples only read files that ended with .js before.


a13jorla:
Works fine with this fix. And now yo can change in php, js and html document boxes.Merging

Issue #1314, opened 2015-05-06
Code Viewer: Description box content alignment


a13alfbe:
The content of the descriptionbox is a bit misaligned at the moment:
* Add some padding to the description box to increase readability
* Align the text with the codeboxes
![untitled](https://cloud.githubusercontent.com/assets/11715493/7490446/e90ef714-f3e0-11e4-8565-2a1f6a318421.png)

PR #1316, opened 2015-05-06
Code View #1314: Description box alignment

a13alfbe:
#1314 
Just a small fix for the description box.

### Before: 
untitled

### After:
untitled2


f12danjo:
Looking good, a small change, will merge.

PR #1321, opened 2015-05-06
Cv#1225 5

c12jenek:
#1225 added some comments/todo list and more dataflow


c12jenek:
Made smaller pullrequests that could be automatically merged

PR #1322, opened 2015-05-06
Cv#1225 5.1

c12jenek:
#1225 More comments, todo-lists, functionflow added


f12danjo:
Great new error message instead of the old: "If no content appears here... update your browser...". The new one if way more informative. Way more informative comments.

Issue #1312, opened 2015-05-06
Progress bar in studentView is edge to edge with the course-containers


a13erilj:
In the studentView the big progress bar is edge to edge with course-containers
The progress indicator is also hidden behind the course-container which makes it hard to see the completed number of points.

PR #1325, opened 2015-05-06
Update studentView.php

a13alfno:
Added lost div
issue  #1312 
http://webblabb.iki.his.se:8080/umv/a13alfno/Version%200.5%20/UserManagementView/studentView.php


a13antda:
Code. Indentation is fine. 
Test. Its works.

PR #1329, opened 2015-05-06
Codeviewer #1295 highlight length fix

### LINKED ISSUE #1295: Code Viewer: Length as important word.
The word length seems to always be an important word even tho it's not added as such in the database. 

If you add "length" as an important word you get an error on line 1237 in codeviewer.js.

b12simsa:
#1295
Two for loops had been replaced by two smart indexing statements e.g.
if(keywords[tokenvalue] !=null){
  //do something
}
But if tokenvalue was equal to "length", keywords[length] returns the length of the array, and the statment is true.


f12danjo:
Looking good, tested and works as it should. It now works to add "length" to important words.

PR #1330, opened 2015-05-06
Bug fix

### LINKED ISSUE #1328: The hover effect is buged
Fixed a bug when the hover effect sometimes doesent work on the when mouse is over a margin.

a13antda:
#1328

changed .hover to on.('mouseenter') and on.('mouseleave') its an more uppdated hover effect that doesent bug out


a13alfno:
Fix seems valid! Nice!

Issue #1308, opened 2015-05-06
Grey out/remove unfitting options in "Edit Item" dialog


b12erijo:
In the edit item dialog, some of the options are not suitable for all item types. E.g. you can now set a grade system and highscore for a header. The options that are not fit for the item type should be either greyed out or removed completely.

HGustavs:
Do we mean inconsistent and not unfitting? If we mean unfitting I do not quite understand the issue.

b12erijo:
Yes i meant just that, couldn't find the word for it :)

Issue #1305, opened 2015-05-06
Textboxes only has placeholder text in the "Edit Course" dialog


b12erijo:
When editing an existing course, lets say "Operativsystem", the course name and course code text boxes has only placeholder text with the current values, this means that they are in effect empty. As a consequence of this you will always have to retype coursename and course code even if you just want to change visibility of the course.

Issue #1303, opened 2015-05-06
"edit course", "edit version" and "new verison" dialogs should be more UX-friendly


b13alebr:
The layout in these diaolgs is not good, also tables and divs is used in the wrong way. For example there are divs inside tables which is not allowed according to the HTML-standard.

PR #1333, opened 2015-05-06
Ds19#1303 - "edit course", "edit version" and "new verison" dialogs should be more UX-friendly

b13alebr:
Have changed the layout in the dialogs "new course", "edit course", "new version", "edit version", "edit item", "edit dugga" and "edit variant". All tables have been removed and instead added divs for the layout. Some new css for the new layout.

This is fix for issue #1303, #1305 and #1308

### DIFF
--- a/DuggaSys/courseed.js
+++ b/DuggaSys/courseed.js
@@ -101,15 +101,7 @@ function selectCourse(cid, coursename, coursecode, visi, vers, edvers)
 	$("#C" + cid).css("box-shadow", "1px 1px 3px #000 inset");
 	$(".item").css("background", "#fff");
 	$("#C" + cid).css("background", "#EDF");
-	// Set Name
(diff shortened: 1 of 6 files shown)


b12erijo:
What's the purpose of "$("#link").html(iistr);" in the selectItem() function in sectioned.js when the item in question is anything but a link?

b13alebr:
The "iistr" variable is used to store options for the select input. It is used both for type link and test, i think it will also be used for type code. All links or duggas is stored in iistr and then "$("#link").html(iistr);" is used to set all these options into the select input. If you create a new item inside a course and chooses type to link or test and then chooses in the select dropdown and save, it will then show as a link on the course page. 
link

a13chrju:
@b12erijo links, files aswell as tests are inserted as options into the select-tag with id "link".

b12erijo:
That sounds very obvious now that you say it, and it looks pretty obvious in the code as well. I guess i was just tired or something! Thanks for taking time to explain so thoroughly :)

a13chrju:
Good job man, reach the stars!

PR #1334, opened 2015-05-06
Code view #1228 markdown pdf

b12simsa:
Upload Markdownmanual.pdf


f12danjo:
Great, now we have some markdown documentation.

c12jenek:
Great explanation of how to use markdown as well as how it's generally done in the code, looks good!

Issue #1339, opened 2015-05-07
Buggfix with the the linear graph


c13marme:
Buggfix with the the linear graph

PR #1340, opened 2015-05-07
Update umv.css

c13marme:
Removed height:100% from contentUMV

#1339 

http://webblabb.iki.his.se:8080/umv/c13marme/Ver%200.52/UserManagementView/teacherView.php


a13erilj:
Everything looks

Issue #1183, opened 2015-04-28
Logging support in code viewer


HGustavs:
Im not sure if this issue has been reported previously... but it would be nice if we could get logging to work for the code examples so we see whom has been viewing the code examples.

I believe that the dugga-system has logging functionality working, it is "just" a question of hooking it up to the code viewer.

f12danjo:
Fixed by [#1341](https://github.com/HGustavs/LenaSYS/pull/1341).

PR #1341, opened 2015-05-07
CodeViewer #1183

a13erisc:
Adds logging support to codeviewer.


f12danjo:
Have tested it, seems to work great. It logs the viewings a user does and to which code example and at which time and saves it in the database.

Issue #1237, opened 2015-05-04
Everything in UMV should be changed to english


c13emmjo:
Now the buttons and info are in swedish this should be changed into english

a13erilj:
This does NOT include changing the language of data in the database

HGustavs:
Since we strive to use realistic data it is unavoidable that some of the data is in Swedish, but we need to make sure that everything else is in English.

a13ottne:
FFS... DONE

PR #1345, opened 2015-05-07
Umv19#1237

b13frero:
#1237


a13erilj:
I cannot merge this pull request due to the fact that this branch has been branched from master. Please make sure which branch you are on while branching to avoid these issues in the future.

Issue #1147, opened 2015-04-27
CodeViewer: Template 3, wrong default value?


a13alfbe:
The default height for box 1 in Template 3 seems to be wrong
![untitled](https://cloud.githubusercontent.com/assets/11715493/7347238/c7e8dd8a-ece6-11e4-8b35-32218c41c9a2.png)

a13jorla:
Does not seem to be a issue at this time. Seem to be fixed (hight: -webkit-calc(100%);). But the resizing seems to breakup sometimes in firefox.

PR #1351, opened 2015-05-07
Update codeviewer.js

c12jenek:
#1147 Added sanity check to make sure localstorage can't be fetched if empty


a13alfbe:
Works like a charm!

Issue #1300, opened 2015-05-06
HTML tags won't show in php-files


f12danjo:
![1](https://cloud.githubusercontent.com/assets/11722176/7489308/956f558c-f3d5-11e4-9603-0955ed0da081.png)
A php-file viewed in the codeviewer and the same file in Notepad++.

HGustavs:
It seems like the < and > are not escaped in the output....
if we write <html> it is interpreted as a tag. We must write &lt;htlm&gt;

HGustavs:
The tag is there but interpreted as html.

f12danjo:
Yes, exactly and it is the same in the description boxes. Which can cause trouble with the code blocks.

b12simsa:
Added a suggested fix in #1352.

PR #1352, opened 2015-05-07
Format html tags in quotes

b12simsa:
#1300
Add a function that replaces  the "< and the >" part of a tag with "&lt, and &gt". This will prevent the browser from rendering the them as html.


a13alfbe:
The cogwheels for each codebox doesn't appear for me, is it the same for you?

Really nice code tho, simple and clean!

a13alfbe:
Found what the problem was, i will make a pullrequest to fix this!

a13alfbe:
It now works as it should, will merge this to the weekly branch and close the issue.

Issue #1359, opened 2015-05-08
Subissue of #1076 fix null-value crash 


b12simsa:
Copied from #1076
~~~
HGustavs: 
The code below does not work if retdata['before'] is null.

TypeError: retdata.before is undefined
if(retdata['before'].value!=null&&retdata['after'].value!=null){

This code needs to work even if a we get an error from the service.....

Perhaps retdata['before']!=null works better?
~~~

b12simsa:
Suggested fix in pull-request #1360.

PR #1360, opened 2015-05-08
Check if object is null

b12simsa:
#1359
Check if object is null, instead of checking if value is null.


a13alfbe:
This fixes the crash!

PR #1365, opened 2015-05-08
Update sectioned.js

c13andka:
set a little timeout so that it does not execute to fast, it did in firefox


a13chrju:
yey works like a charm, good job.

Issue #1348, opened 2015-05-07
Old data being overwritten in teacherView graph 


a13erilj:
In the teacherView graph that displays data about the progress in the different program courses it seems that data is being overwritten.

a13erilj:
issue has been solved

PR #1368, opened 2015-05-08
Fixed old line graph data staying after redraw.

a12oskri:
Fixed issue #1348, clearRect was missing a [0].


a13erilj:
The solution is a bit ugly but it works :+1: Good job

Issue #1366, opened 2015-05-08
Disable before and after button.


a13jorla:
When on first and last example they should not be usable.

a13jorla:
Closing since this is a old issue #1326

PR #1371, opened 2015-05-08
Fix for errors in #1300

b12simsa:
issue: #1366
old pull-request: #1352
old issue: #1300
The replace function did not replace lines like, "<br> text" correctly.


a13alfbe:
Nice small fix!

PR #1372, opened 2015-05-08
Update umv.css. Fixed the height property of content wrapper

f13marhe:
Fixed the height property "height: calc(100% - 150px);".


a13erilj:
Looks nice. Works as expected.

Issue #1240, opened 2015-05-04
When ”No student data found..” is shown in teacherview -> the radiobuttons and changepages are not going to appear


c13emmjo:

a13erilj:
Solved #1373

Issue #1126, opened 2015-04-27
Fix page numbers on teacherview.


a13ottne:
On the teacherview, page selection is displayed even when there are no additional pages available. Make page selection dynamic depending on amount of data loaded.

a13erilj:
This issue has been solved

PR #1373, opened 2015-05-08
Umv19#1240

a13antda:
#1240 Is fixed.

#1126 
also fixed a bug with rendering students this is now working as intendent.


a13erilj:
It looks fine :+1:

Issue #1375, opened 2015-05-11
Edit Content does not show css files in file drop down box.


a13jorla:
In the weekly merge it got lost. Insert it again.

a13jorla:
This issue is now merged and fixed. Good work a13alfbe.

PR #1376, opened 2015-05-11
CodeViewer #1375: .css example support

a13alfbe:
You can now also find the .css files when browsing which file to show in a code box.


a13jorla:
With this edit it's able to see and choose the css files. Merging this request since it has been merged erlier and it works this time allso.

Issue #1378, opened 2015-05-11
Codeviewer Markdown Headings doesnt work


b12simsa:
After a fix the markdown headings stopped work correctly.

f12danjo:
Fixed [#1379](https://github.com/HGustavs/LenaSYS/pull/1379)

PR #1379, opened 2015-05-11
Fix for broken markdown headings

b12simsa:
Fix for issue #1378.
Removed extra whitespaces.


f12danjo:
Looks good. Some whitespaces had been added to the lines of code in the markdown. This simply removed it, so now headings with markdown works again.

Issue #1326, opened 2015-05-06
Back and forward button.


a13jorla:
When on last or first example the back and forward buttons should be hidden. And when no playable example the play botton should to.

a13jansa:
Problem solved!

PR #1383, opened 2015-05-11
CodeViewer #1326: Back and forward button

a13alfbe:
#1326 
The back/forward buttons gets disabled if the current example is the first or last example in order.


a13jansa:
Its seems to work. On the first page you cant go backward. On the last page you will not be able to continue

Issue #1381, opened 2015-05-11
Subissue of #1232 comment css files.


a13jorla:
Subissue of #1232
Make explaining comments in the css files.

c12jenek:
I'll be commenting bits of the codeviewer.css file, hence the self-assign

f12danjo:
Fixed by pull request [#1425](https://github.com/HGustavs/LenaSYS/pull/1425)

PR #1387, opened 2015-05-11
comments css file #1381

a13jansa:
Added some comment in the css file


f12danjo:
Several comments seems to be in the wrong place or doesn't make sense to the block they are describing. Take a another look at it.

f12danjo:
Need to close this pull request and delete the branch in preparation for the weekly merge. I have downloaded the branch, so it is saved.

Issue #1385, opened 2015-05-11
Small progress bars in teacher view sometimes are placed beneath the student info


a13erilj:
Fix the problem where the small progress bars are places beneath the student info in the teacherView

PR #1405, opened 2015-05-11
Fixed the small progress bar prob.

f13marhe:
Reindented the CSS as well. #1385


c13emmjo:
Looks good but the student_page div need to be 100% and also there need to be padding on the div student so it looks good together with the smal progressbars div

But I have done this in my issue so merge in yours so i can request my issue

Issue #1369, opened 2015-05-08
Add envelopes to the teacherView for the students


a13erilj:
Instead of showing the email-address to the students there should be an envelope instead that can be clicked to send an email to the student.
Related issue #1150 
Refer to mockup for more details of the actual representation.

PR #1407, opened 2015-05-11
umv20#1369

c13emmjo:
did issue #1369

did new envelope images in svg and added them to the studentview.js and teacherview.js
also fixed so when the teacher click on the envelope in teacherview that specific students email are typed in the mail.

teacherView:
http://webblabb.iki.his.se:8080/umv/c13emmjo/week20_1/UserManagementView/teacherView.php

(there was some problem adding my changes and the sync didnt worked ok.. so there is some things that i had to add that already had been added before) 
in teacherview.js there was a function: renderView(data) that didnt worked so it is back were it was before now that function. because data was never shown. so this should be fixed in a another issue


a13erilj:
Due to the use of incorrect/old code in this pull request it cannot be merged.

Issue #1377, opened 2015-05-11
Debug additions


c12jenek:
Subissue of #1076 , adding debugging information to functions and statements

PR #1413, opened 2015-05-12
Cv#1377 debug

c12jenek:
#1377 Added some debug info, changed some alerts to console logs instead as they're not relevant for the user to see


a13alfbe:
The beforeId/afterId debug message always shows UNK/UNK. It shows the correct values if the example is edited. 
A fix for this could be to add beforeid and afterid to the select query on line 62 to get the correct id:s for the example

f12danjo:
Seems to work. Good that you removed the alert when not viewing a example. The debug seems to work as intended.

a13alfbe:
Works! Good descriptive debug messages.

Issue #1391, opened 2015-05-11
Make it possible for the teacher to click on student and enter their studentView


a13erilj:
In teacherView make it possible for the teacher to click on the students info div and be redirected to the studentView for the given student.

PR #1415, opened 2015-05-12
Umv20#1391

a13antda:
Teacher can now click on a student and get their student view.
This pullrequest only fix the feature to do this.
So make a new issue to fix some small navigation changes and maybe some css.

#1391 

http://webblabb.iki.his.se:8080/umv/a13antda/v20/UserManagementView/teacherView.php


a13erilj:
There are a few bug but this will be fixed during feature freeze.

PR #1419, opened 2015-05-12
Umv20#1369

c13emmjo:
#1369

-Did new svg envelope images to studentview and teacherview
-Connected the envelope in teacherview so when the teacher click at the envelope that specific student gets that email.
-Updated the umv.css so the envelope is located correctly and also so the ssn and username is started at the same place

teacherView:
http://webblabb.iki.his.se:8080/umv/c13emmjo/week20_2/UserManagementView/teacherView.php


a13erilj:
Looks fine, will merge

PR #1423, opened 2015-05-12
Ds20#1384

b13alebr:
Tables remoed in "edit user" "add user" and "edit link/file" dialogs.
Also fixed a few functions so the values to the input fileds is set properly.
Small change in the css file so the textinput is the same size as select inputs.


a12majca:
looks good, well done.

a13oscwe:
Restoring branch and opening a new pull request because the changes did not make it into the ds20 weekly branch.

PR #1424, opened 2015-05-12
Removed redundant cookies and use of the variable "loginvar"

b12erijo:
The cookies and the "loginvar" variable were redundant as we have changed the processLogin and processLogout functions.


c13andka:
This is added because we refresh page now on login

c13andka:
Seems legit

PR #1425, opened 2015-05-12
cv#1381

c12jenek:
#1381 Part of commenting css issue, this does just concern codeviewer.css 
Added /* START / STOP */ to the different parts of css, removed dead code


f12danjo:
Very convenient with the START/STOP comments, especially when encircling several blocks. Makes the code easier to read. Overall a good cleanup.

Issue #1296, opened 2015-05-06
Code Viewer: HTML Hightlighting doesn't work (Example 12 & 13)


a13alfbe:
The highlighting is not working properly in these examples. 

It works fine if you delete .toLowerCase() on line 1335 in codeviewer.js but im not sure why it is there and the reason for it.

HGustavs:
can you paste the surrounding few lines above and below that line, as I am unsure if I am looking at the same version.

a13alfbe:
```javascript
if(htmlArray.indexOf(tokens[i+1].val.toLowerCase()) > -1){
    var k = 2;
	var foundEnd = false;
	while(tokens[i+k].kind != "newline"){
		if(tokens[i+k].val == ">"){	//If a > has been found on the same line as an < and the token to the left of < is in htmlArray then it classes it as an html-tag
			foundEnd = true;
			break;
		}
		k++;
	}
	if(foundEnd){
		pid="html"+htmlTagCount+boxid;
		htmlTagCount++;
		if(htmlArrayNoSlash.indexOf(tokens[i+1].val.toLowerCase()) == -1){
			htmlTag.push(pid);
		}
		cont+="<span id='"+pid+"' class='oper' onmouseover='highlightHtml(\"P"+pid+"\",\""+pid+"\");' onmouseout='deHighlightHtml(\"P"+pid+"\",\""+pid+"\");'>"+("&lt" + tokens[i+1].val);
		for(var j = 2; j < k+1; j++){
			cont+=tokens[i+j].val;
		}
		cont+="</span>";
		i=i+k;
	}else{
		cont+="<span class='oper'>"+tokenvalue+"</span>";
	}
}
```

c12jenek:
A solution has been merged in!

PR #1427, opened 2015-05-12
CodeViewer #1296: Highlighting crash

a13alfbe:
#1296 
The error occured when the code tried to use the function "toLowerCase()" on a value that was a number. 
"canvas" was also added to the array of allowed html tags.

Me and @f12danjo worked together on this one.


c12jenek:
HTML highlighting now works on examples where it crashed before ( 5, 6, 9 etc.), the addition of canvas to the html array is great. Codestandard is followed, great! Is ready to merge :)

PR #1428, opened 2015-05-12
Umv20#1170

a13freas:
1170# In teacherview a coursname should be visible when hoovering over a smal progressbar

Its working but the styling have to be done.


a13freas:
http://webblabb.iki.his.se:8080/umv/a13freas/LenaSYS-umv20/UserManagementView/teacherView.php

a13erilj:
Have checked it and everything seems to work, there are a few buggs but these will be handled during feature freeze next week

Issue #1408, opened 2015-05-12
Fix the dropdown menu in teacerView to the navigation header


a13erilj:
Fix the dropdown menu in teacherView to the navigation header so that i follows the user when it scrolls the page.

PR #1432, opened 2015-05-12
Update umv.css

a13antda:
its now fixed to head menu 
#1408


a13erilj:
Have checked the code and tested the site, it seems to work.

Issue #1417, opened 2015-05-12
Final refactorization of high score system


a13karsi:
The high score system is essentially working at the moment at the state at which we were hoping for. The user can select between different high score modes, the high scores are only applied where appropriate and so on.

The only thing left to do for now is to run over all of the code and do final refactoring on various parts of the high score system, it may include just simple things as giving variables descriptive names. 

This update should make sure to include that only users that have performed a graded test with a grade higher than U will get shown in the list.

PR #1434, opened 2015-05-12
Highscore system refactoring

a13karsi:
This pull request marks the final stage at which I will be actively working on the high score system. It seems to function correctly at the moment and we will deal with bug fixes in this system as normal from now on.

See issue #1417


c13andka:
Seems legit

Issue #1337, opened 2015-05-07
Create SQL queries for retrieving data that is to be displayed in the teacherView diagram


a13erilj:
Create SQL queries that can be used to retrieve the wanted data for the graph.
* Create a SQL query that retrieves the courses for the program and how many students have finished and not finished the course.
Part of issue #867

a13erilj:
Solved

PR #1435, opened 2015-05-12
Added querys for fetching the relevant information needed for the line graph.

a12oskri:
Issue #1337. Added querys for fetching the relevant information needed for the line graph and made the line graph show the right values. Also made the graph not exceed its containing div.


a13erilj:
Looks fine, good job

PR #1436, opened 2015-05-12
Update template3_butt.svg

a13jorla:
Just moved the rectangles to represent the css template.


f12danjo:
Great, now the template 3 button matches the actual layout of template 3.

PR #1437, opened 2015-05-12
Ds20#1218 Testdata for duggasys

b13alebr:
Added testdata for duggasys, one course (Webbprogrammering) with codeexamples from codeviewer and one course (Webbutveckling - datorgrafik) with example duggas.


a13oscwe:
Works for me to first run init_db then testdata, no errors. Further testing would be nice on the live server @a13karsi

a13karsi:
Testing this on the live database isn't possible at the moment because there are conflicting entries in it and rebuilding it would probably ruin the day for some people. But I tested both the init_db and testdata file on a separate server and they run without any problems.

Merging this update should be completely safe, the database and testdata could be executed maybe later today when everyone's finished their day's work.

PR #1440, opened 2015-05-12
Ds20#1438 Items type code not working

b13alebr:
You can now add a new item on the coursepage and choose type code and choose codeexamples, the link leads to codeviewer.


a12majca:
it works, great!

Issue #1238, opened 2015-05-04
Move the searchfield up to the header


c13emmjo:
move up the seacrhfield up to the header in teacherview 
(have to go in the shared map to fix this)

also move the title and the totalinfo div up so there is not that much space.

PR #1444, opened 2015-05-12
Umv20 #1238

a13alfno:
#1238
Moved searchfield to nav header. 
navheader.php now have case for Teacherview
Moved no dropdown meny, because it was not on my past branch

Link for searchfield:
http://webblabb.iki.his.se:8080/umv/a13alfno/Umv%2020/UserManagementView/teacherView.php

### DIFF
--- a/Shared/navheader.php
+++ b/Shared/navheader.php
@@ -39,6 +39,13 @@
 				case 'NONE';
 					// This case is used for pages still under development
 					break;
+				case 'TEACHERVIEW';
+					echo "<td class='serachC'>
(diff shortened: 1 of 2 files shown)


a13ottne:
placeholder is in swedish, should be english, also there is both a placeholder "sök" and "sök" to the right of the search box, it should be enough with only the placeholder. Plz take a look at this, otherwise good.

a13erilj:
There also seems to be no button to click to perform a search :( Also the dropdown menu is acting very strange and is placed inside the header :/

Issue #504, opened 2015-03-29
Duggasys magic headings


HGustavs:
If large tables are used in duggasys, headings are not visible.since they scroll out of view. Headings that are fixed when they move out of view would resolve this.

Furtheore, any solution that relies on a fixed unmoving header and scrolling table body is seen as too difficult to use.

We are asking for fully free-moving headers both horizontally and vertically.

c13chrjo:
Issue is not present

a13chrju:
I don't understand which tables we're talking about? YOu mean under the file "sectioned.php"?

a13oscwe:
Having a sticky or "fixed" table heading(thead tr essentially) is very tricky to do when the table exceeds the current screen width since setting the css property to position: fixed sets an absolute width for the table heading row on both the X and Y axis. I tried using JS to set the width's of the TH's to the data TD's but the browser overrides the width because the fixed element cannot overflow on the x axis. This is possible to fix using third-party libraries (https://github.com/sxalexander/jquery-scrollspy) but as i understand, is not allowed.

Given i understand the issue correctly that you want the tables on resulted.php and duggaed.php to have a fixed header.

If somebody else want's to give it a shot feel free.

HGustavs:
There is no easy "automatic way" to do this. The trick is to have two table headings, one that is "locked in place" and one that is free-floating. If you on scroll updates the width of the second table heading to match the width of each of the column in the first table the sizes match up.

This is done with about 25 lines of javascript.

a13oscwe:
@b13alebr has solved this and it looks pretty good. Pull request #1591

PR #1445, opened 2015-05-12
Ds20#504 - fixed heading menu when scrolling out of view

a13chrju:
the page "sectioned.php" has a menu that is only displayed for admins, if there are too many items in the course and you scroll down through the list the menu is not visible. But I have made it "fixed" whenever the user scrolls out of the view.
https://github.com/HGustavs/LenaSYS/issues/504


b13alebr:
Does not work as it should, do not merge.

a13chrju:
sorry myself, ill have to fix it next week

PR #1451, opened 2015-05-13
Ds#1277

a13denli:
Added the example parameter and answer for creating new variants


b13alebr:
The "ö" in "röd" looks strange but otherwise it looks good!

a13denli:
Fixed that

c13andka:
Seems legit!

PR #1455, opened 2015-05-13
Umv20#1398

c13emmjo:
updated teacherview

fixed the changepage so the page the user are in on is markt and updated the css to it

also sent in css that i sent in yesterday but didnt get in. so the style in techerview over the students are looks better. 

teacherView:
http://webblabb.iki.his.se:8080/umv/c13emmjo/week20_3/UserManagementView/teacherView.php


a13erilj:
Looks nice :+1:

Issue #991, opened 2015-04-20
CodeView: Examples working properly with only google chrome


a12jiyba:
some boxes disappears with other webbrowsers.

HGustavs:
This is a very important issue.

Most important of all.... i would say. 
 If this type of general stuff does not work in all browsers, all of the rest of the work is futile.

f12danjo:
Seems to work exactly the same in chrome, firefox and opera. Not in internet explorer though.

a13jorla:
We have now merged branch CodeView#991 Not working in some browsers and this made it possible to say we fixed this bug.

PR #1457, opened 2015-05-13
Code view#991 not working in some browsers

f12danjo:
With this fix, the codeviewer works in Internet Explorer. Fixes issue [#991](https://github.com/HGustavs/LenaSYS/issues/991)

### DIFF
--- a/Shared/css/codeviewer.css
+++ b/Shared/css/codeviewer.css
@@ -13,7 +13,7 @@
     font-family: Arial,Sans-Serif;
     font-style: italic;
 }
-/* DOCUMENT CONTENT STYLE STOP */bo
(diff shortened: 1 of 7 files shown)


c13caral:
the code in Template4 doesn't follow the code standard.
Look at  line 7-8, 19-20, 32-33

a13jorla:
The fix makes all examples work in every browser i had like Opera, Firefox, Explorer and Chrome.

f12danjo:
The indentation looks weird in github sometimes, it has proper indentation if you open it in another editor like Notepad++.

a13jorla:
Because of that we could not find any more reason not to merge this fix we will do it.

Issue #1439, opened 2015-05-12
Incorrect dates when editing dugga's


a13oscwe:
When a dugga is modified with "edit dugga" it's possible to set a deadline date that is before the release date, this should not happen.

a13oscwe:
Fixed in pull request #1460

PR #1460, opened 2015-05-13
Ds#1439

b13johfr:
Fixed that Release date only can be set the same or before Deadline in Edit dugga.
#1439


a13oscwe:
Looks good

PR #1465, opened 2015-05-13
Ds20##1459

c13andka:
fixed some high score stuff


a13karsi:
Looks good, ran the code on my own server, everything works smoothly.

Issue #1145, opened 2015-04-27
Template 6 resizable functionality.


b12alela:
Template 6 isn't resizable.

a13jorla:
Branch #1145 is merged and thus we can close this issue. Cause ti made template 6 resizable.

PR #1467, opened 2015-05-13
Code viewer #1145

b12alela:
This adds basic resizability to template 6 before feature freeze.

The functionality as it stands right now may not be perfect, but everything is operable and it seems relatively stable.


f12danjo:
Looks good. Resizing of template 6 examples works now.

Issue #1127, opened 2015-04-27
Do so the searhfield is working in teacherView


c13emmjo:
Fix so the teacher can search on students ssn and userId

PR #1471, opened 2015-05-13
Umv20#1127 x2

a13antda:
#1127

http://webblabb.iki.his.se:8080/umv/a13antda/v20-v2/UserManagementView/teacherView.php


c13emmjo:
it works with a few bugs but that we can fix next week. 
There are a few console.logs aswell in studentview.js and teacherview.js that should be removed but that we can do next time.

PR #1474, opened 2015-05-13
Cv20#1007 - Drag and Drop Sequence handler

b13danso:
Dragndrop functionality to sort example sequences, their before/afters and add new examples to the sequences. Also to create and delete sequences.


c13caral:
This is not very glanceable but we can merge this and the glanceable issue can we work on next week :+1:

c13caral:
there are also some comments that is missing but we can also work on that next week

Issue #1429, opened 2015-05-12
Fault handling if not logged in and has no example needed


c12jenek:
![username bug](https://cloud.githubusercontent.com/assets/11715479/7585870/11d565d2-f8a7-11e4-853b-aac6e15ba382.jpg)
This appears when you're standing in http://webblabb.iki.his.se:8080/codesys/c12jenek/CodeViewer/EditorV50.php and you are not logged in

HGustavs:
I would really like if the buttons would disappear if there is no backward/forward/play link to go to. Also the undefined:undefined should not appear.

c12jenek:
HGustavs comment is made into smaller issues #1483 and #1482

PR #1481, opened 2015-05-18
Fault handling

c12jenek:
#1429 
Added a username check, giving guests a username which is used to log site traffic


a13alfbe:
Tested and it works! Will merge this.

f12danjo:
You get a new random number every time you refresh the page. Maybe it could be stored in local storage?

Issue #1442, opened 2015-05-12
CodeViewer deletes first php tag


f12danjo:
![1](https://cloud.githubusercontent.com/assets/11722176/7587543/699b05e6-f8b7-11e4-9569-3718caafc95a.png)
The ending tag is displayed as normal though.

HGustavs:
Question is whether it is there and just not shown or if it is really lost.

I suspect that it is there?
If so, have we translated the ? to a &#35;63; entity?

a13alfbe:
FIxed this bug, waiting on the weekly branch to be created. 
Missing && tokens[i+1].val != "?" in the statement on line 1295 in codeviewer.js

f12danjo:
Fixed by [#1485](https://github.com/HGustavs/LenaSYS/pull/1485).

PR #1485, opened 2015-05-18
CodeViewer #1442: PHP tag

a13alfbe:
#1442 
Adds "?" as an exception in the statement checking for the start of an html tag.


f12danjo:
Works great. A simple addition of a != "?" in an if-statement.

Issue #1479, opened 2015-05-18
Generate diagrams of the latest version of the database


a13karsi:
Since we won't do any more major feature additions it is unlikely that the diagram will change all that much, therefore, it's probably a nice idea to throw up a diagram of the database with the latest updates in to the documentation folder in shared.

PR #1488, opened 2015-05-18
Ds#1479

a13karsi:
Updated database diagrams and removed old ones, also removed redundant files such as the svg file. We just need the mwb and png file really.

See issue #1479


a13ottne:
must close this beacuse branch errors

PR #1489, opened 2015-05-18
Updated createboxmenu function!

b12alela:
The wrong item in the array of information about the boxes had been chosen, which made it display a number instead of the boxtitle.

Only changed from item 3 to item 4 in that array to get the proper item.


f12danjo:
Well done. The box titles are now shown exactly the same when you are logged out as when you are logged in.

Issue #1482, opened 2015-05-18
Buttons should dissapear if no examples are available


c12jenek:
When entering CodeViewer site without being logged in and not having an example the next/previous example buttons should be greyed out

f12danjo:
Fixed by [#1491](https://github.com/HGustavs/LenaSYS/pull/1491).

PR #1491, opened 2015-05-18
cv #1482

c12jenek:
Added disfunctionality of skipping through non-existing examples #1482


f12danjo:
Works great, the buttons are now faded when not viewing an example.

Issue #1205, opened 2015-04-29
Pushing save in Edit Contentbox does not uppdate the page.


a13jorla:
Now(2015-04-29) you have to manually update the page to see the result from the changes that have been done.

HGustavs:
If the code uses the ajaxservice and the database is updated, it __should__ all be automatic.

PR #1504, opened 2015-05-19
Page refresh when hitting save in Edit Content box

a13erisc:
see #1205


a13alfbe:
Nice that this finally works!

PR #1506, opened 2015-05-19
CodeViewer #1289: Title fix

### LINKED ISSUE #1289: Code Viewer: Too long titles cause issue with heading
I think it is because of the name of the example code is too big. Maybe overflow can fix it?

skarmavbild 2015-05-05 kl 15 04 37

a13alfbe:
#1289 
* Codeexample title is now vertically centered
* The navigation buttons have "cursor: pointer".
* Codeexample title overflow is hidden
* Responsive title size.
* Removed unnessecary css for login button.
* The table in navheader is now the correct height (50px) instead of 54px.

### DIFF
--- a/Shared/css/responsive.css
+++ b/Shared/css/responsive.css
@@ -5,6 +5,9 @@
     .showdesktop{
         display: none;
     }
+    #navHeading {
+        font-size: 1.5em;
+    }
     #exampleSection{
         display: none;
     }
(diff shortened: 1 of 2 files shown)


f12danjo:
Nice. Fixes the problem with title overflow, when the title was too long for the window, that caused the navigation buttons to become misaligned.

PR #1513, opened 2015-05-19
Add wrappers for divs with floats to fix margins

### LINKED ISSUE #1498: The margin between buttons and the list in "files" is too much
This doesn't really look good, and it's hard to keep track on the buttons.

a13oscwe:
Fixed #1498 
Added wrappers for the divs with floats and removed some inline CSS.


a13denli:
You don't need "float:right" in the divs: alllinks, allglobalfiles, allcoursefiles and alllocalfiles? Just so the divs gets closer to the buttons.

c13andka:
Yes, eliminate that "float:right" from those divs, and we are good to go :+1:

Issue #1478, opened 2015-05-18
DragnDrop has a few bugs and issues


c12jenek:
There's a lack of comments and the behaviour is very strange, moving a box within the pop-up moves the entire pop-up. The "create new sequence"/"Delete sequence"/"Save sequence" should be aligned and have some margin from the surronding edges aswell as follow the standard with purple boxes

c12jenek:
It's also implemented in it's own files which we're not supposed to do, it could and should be implemented in codeviewer.js/editorService.php

a13jansa:
i can start with the "create new sequence"/"Delete sequence"/"Save sequence"

b13danso:
Fixed.

f12danjo:
Fixed by [#1642](https://github.com/HGustavs/LenaSYS/pull/1642).

PR #1514, opened 2015-05-19
Update style.css

a13jansa:
this is just and small change for issue #1478


a13henhu:
I've checked this pull request, and I've seen the changes. The default code for the buttons was not matching with the color-theme. With this fix, i've seen the changes and it is now matching all the other buttons, in that window displayed.

a13jansa:
It went wrong in the branch.

PR #1518, opened 2015-05-19
Umv21 #1238

a13alfno:
Issue #1238 

Added case TEACHERVIEW in nav header. 
Moved searchfield to nav header. 
Moved css rules from umv.css to style.css

Link: 
http://webblabb.iki.his.se:8080/umv/a13alfno/UMV%2021/UserManagementView/teacherView.php


f13marhe:
need further css styling. otherwise (Y)

PR #1519, opened 2015-05-19
Update style.css

a13jansa:
Made some changes for editexample box #1478

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -949,7 +949,10 @@ input.new-item-button:hover {
     margin-left: 4px;
     font-size: 14px;
     float: right;
-	transition: background-color 1s ease 0s;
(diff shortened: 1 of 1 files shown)


a13alfbe:
It seems like you have an older version of this file. Get the newer file and apply your changes.

a13alfbe:
You will need to use more than one transition per class to make it work in other browsers. 

**Example:**
```css
-webkit-transition: background-color 1s ease 0s;
-moz-transition: background-color 1s ease 0s;
-o-transition: background-color 1s ease 0s;
transition: background-color 1s ease 0s;
```

Is there a way to combine the css for **.newSequence**, **.deleteSequence** and **.updateSequence** since they share alot of the same properies?

a13alfbe:
Awesome, will merge this now!

Issue #1456, opened 2015-05-13
Showing NaN in the progressbar when credits are 0


c13marme:
Progressbar shows NaN when a student have 0 credits

PR #1529, opened 2015-05-19
Update studentView.js

c13marme:
Fixed a issue that shows NaN instead of a 0 when the student doesn't have any credits at all.

Log in with b13sceth

#1456 
http://webblabb.iki.his.se:8080/umv/c13marme/Ver%200.61/UserManagementView/studentView.php


a13freas:
Works fine!

Issue #1487, opened 2015-05-18
In teacherView a search should change the list of available students.


a13erilj:
In the teacherView the list of students should change when a search is performed. A search should not lead to the display of the studentsView. The reason for this is that a teacher may wish to send an email to a given student and not see their studentView

c13marme:
This issue has been solved.

PR #1532, opened 2015-05-19
Umv21#search

a13antda:
#1487 

Updated the search field so it isnt buggy,

Still someone needs to make the orientation between the tearcher veiewing a student and navigate back to their own teacher page. Proper solution....


c13marme:
The search works much better now. 
Code looks fine, only some minor spelling errors on some variables

Issue #1288, opened 2015-05-05
Code Viewer:  There is a gap at the end of the code examples 


c13caral:
This is from the code example 14. But there are several code example with this issue.

![skarmavbild 2015-05-05 kl 14 59 02](https://cloud.githubusercontent.com/assets/11715450/7472924/7b723f80-f337-11e4-83e7-bdf5e4565a2c.png)

a13jorla:
This seems now to be a testing issue, trying to see if it appeares again in some browsers.

PR #1533, opened 2015-05-19
Fixed white gap on the codeborder

f12danjo:
A white gap shows up on the code border when viewing certain templates. 
Issue: [#1288](https://github.com/HGustavs/LenaSYS/issues/1288)
Fixed it by removing padding from normtextwrapper.


a13alfbe:
Huge improvement! Will merge this now.

PR #1537, opened 2015-05-20
Replacing document.getElementById: codeviewer.js

a13alfbe:
Part of #513 
* Replaced all document.getElementById("id") with $("#"+id)
* Had to change other lines of code to work with this change. Lines 516 & 554 for example.

PR #1538, opened 2015-05-20
Revert "Replacing document.getElementById: codeviewer.js"

a13jansa:
Reverts HGustavs/LenaSYS#1537


a13jansa:
something went wrong so we tried to Revert the branch, bt by reverting the branch dident solve the problem.

Issue #1448, opened 2015-05-13
Create a styled hover function for displaying course name for progress bar in teacher view


a13erilj:
In teacherView there are small progress bars that when hovered over should display the course name of the course that it represents. 
Currently the course name is being displayed but not in a granceable. It where preferable if there was a bubble that displayed the course name when hovering.

PR #1541, opened 2015-05-20
Umv21#1484

b13frero:
#1448

http://webblabb.iki.his.se:8080/umv/b13frero/LenaSYS-umv21/UserManagementView/teacherView.php


a13freas:
It looks fine in chrome but not in firefox. There is a bug when hovering in firefox browser that needs to be fixed.

b13frero:
Problem is fixed by adding some margin in umv.css

Issue #162, opened 2014-04-23
The play(button) link should disappear if the play link is empty


e12johka:
When typing an invalid link the play button is hidden and the only way show it again is to update the page.
the invalid link is being replaced whit the last valid link, but the play button is not shown

HGustavs:
Fixing "bug" issues does sound like a priority to me, has this been fixed?

a13jorla:
When logged in you allways have the play button pressent at this stage of development. So i do not know if this issue is relevant any longer?

c13caral:
The play button is always pressent. Regardless if you are logged in or not.
But we maybe can fix that if the link for the play button is invalid, instead of getting the not found message maybe it can show an alert message or something that the play button link is invalid. 
But this feels like a really low priority issue

HGustavs:
Im thinking that if the play link is empty i.e. there is nothing to run, the play link should disappear through hiding it with css.

a13jorla:
Linking to issue #1326

PR #1544, opened 2015-05-20
Disable the playbutton

f12danjo:
[#162](https://github.com/HGustavs/LenaSYS/issues/162)
Disables the playbutton if there is no playlink.


a13alfbe:
Tested with null values and it works great. Will merge this!

PR #1551, opened 2015-05-20
Code Viewer #1543: Edit example fix

### LINKED ISSUE #1543: Code Viewer: Displaced/wrong values for elements in Edit example
untitled

a13alfbe:
#1543 

* Rearranged the table a bit and
* Removed unnessecary css 
* Changed the impword to use a placeholder instead of value

**Looks like this now:**
untitled


f12danjo:
Nice! Works just like before. This change cleans up the code well. Also aligns things better in the "Edit Example" window. Good job.

Issue #1512, opened 2015-05-19
Warnings in courseed.php


a13denli:
![image](https://cloud.githubusercontent.com/assets/11715616/7699683/e549de76-fe19-11e4-9c5d-e802d567465f.png)

After downloading the weekly branch, and unpacking it to my web-server, I get this error.

HGustavs:
Should courseed really open files? 

I thought that file was strictly database oriented?

a13karsi:
This only happens if you forget to move the .git folder to your own server directory, the file it wants to load is /.git/refs/heads/master, I added this code to the courseed page so we could load the master hash easily to identify "versions" of LenaSys.

In windows the .git folder is commonly hidden. So you have to go into folder options, enable to see hidden files and folders, and make sure you move the .git folder to your server as well.

My pull request: #1454 
Original issue: #826

a13denli:
Ah okey

c13chrjo:
sorta fixed in #1553

PR #1553, opened 2015-05-20
Handle warning when .git folder is missing

a13karsi:
This update changes the file loading code in the php section of courseed.php to check whether the git/ref/master file exists before loading it, if it doesn't exist then the php error won't be displayed but rather a message under the "master hash" string that states that the .git folder was missing.

See issue #1512


c13chrjo:
Looks fine

Issue #1293, opened 2015-05-06
Code Viewer: Regular expression bug replaceAll().


a13alfbe:
There seems to be a bug when a unclosed parathesis is used in an important word such as "scale(" the error occurs on line 999 in codeviewer.js.

Just add a new important word with the example editor like "scale(" to test this.

a13jorla:
There should not be any not closed brackets. So the question is, should there be fault handling for this? @HGustavs

b12simsa:
#1557 Suggested fix.

f12danjo:
Fixed by [#1557](https://github.com/HGustavs/LenaSYS/pull/1557)

PR #1557, opened 2015-05-20
Codeviewer #1293 Uneven amount of parenthesis break the page

b12simsa:
#1293 
Suggested fix that counts the amount of parenthesis in an important word, if they are uneven the word cannot be added.


f12danjo:
Seems like important words can't be added or removed with this update.

b12simsa:
Some line got lost during the copy+paste operation

f12danjo:
Now it works! It won't allow an important word with an uneven number of "(" and ")", but will allow one with an even number.

Issue #1566, opened 2015-05-21
Codeviewer Resize hadle and box wrong placing in template 1 and 2.


a13jorla:
The hanlde is not placed at the box edge as it should but at the edge of the window.

f12danjo:
Fixed by [#1567](https://github.com/HGustavs/LenaSYS/pull/1567)

PR #1567, opened 2015-05-21
CodeViewer #1566: Resizing & templates broken

a13alfbe:
#1566 
Two comments with wrong syntax messed up the templates.


b12alela:
This certainly solves the problem.

f12danjo:
Good find!

PR #1570, opened 2015-05-21
Loginbox reset

a13karsi:
If the user inputs an invalid username or password combination and closes the loginbox, the next time the user opens the loginbox it will have been reset and the red color in the input box as well as the warning will have been cleared.

See issue #1220


b12erijo:
Good stuff, i had forgotten this issue!

Issue #1559, opened 2015-05-20
version div is not in the same height as the buttons


a13denli:
![image](https://cloud.githubusercontent.com/assets/11715616/7725058/7117a518-fef7-11e4-844f-f10f5c61d0d5.png)

Just align the version with the buttons, so it looks more clean.

a13denli:
Completed

PR #1573, opened 2015-05-21
Top menu alignment fix

a13karsi:
Changed  some css so that the menu items at the top align correctly.

See issue #1559


a13denli:
Good

PR #1576, opened 2015-05-21
Codeviewer #820 Description Box Highlight Fix

### LINKED ISSUE #870: CodeViewer: Highlight between boxes
weird highlights
There seems to be a bug where the highlighting doesnt appear in the comment field aswell as the code field when hovered!

b12simsa:
#870
Fixed a bug that caused the highlight of important words disappear from the description box.


f12danjo:
Works beautifully! Now the codeviewer highlights the same words in description boxes as it does in code boxes.

Issue #1554, opened 2015-05-20
Missing whitespace in list


c13chrjo:
There should be whitespace here and it should match the distance on the other subpages.
![untitled](https://cloud.githubusercontent.com/assets/11715749/7724608/3f3a6c3c-fef3-11e4-965a-0252e0e896cc.png)

a13denli:
Completed

PR #1578, opened 2015-05-21
 Missing whitespace in list

c13chrjo:
Added margin-top:10px to listbar
#1554


a13denli:
Change to margin to 6px so it follows "standard"

c13chrjo:
Done

PR #1580, opened 2015-05-21
Cv#1478 - dragndrop fixes.

b13danso:
This fixed some bugs and css issues with the dragndrop-function. Also, the draggable-function on the whole loginBox is removed and the box is now placed 30% from top(previouslly 40%) and only applies to codeviewer(the change is in codeviewer.css).


f12danjo:
Encountered a bug when moving the example boxes when you have to scroll to move it. I still think this is okay, you can use the "sequencer" as intended anyway. Made a separate issue about it [#1585](https://github.com/HGustavs/LenaSYS/issues/1585).

Issue #1409, opened 2015-05-12
Change the years in studentView to represent the semester instead


a13erilj:
In studentView the courses are represented under a given year (2013, 2014 and 2015). This should be changed to represent the semesters instead. 
For example instead of 2013 it should say HT-12/VT-13 or HT-13/VT-14

a13erilj:
Partially fixed

a13ottne:
Since no data about when courses take place are is present in the database, this is issue will be closed for now

PR #1587, opened 2015-05-21
#1409 Update studentView.js

b13davfr:
https://github.com/HGustavs/LenaSYS/issues/1409

remove and added new code for given year 2013,2014,2015 to HT-13/VT-14 , HT-14/VT-15 or HT-15/VT-16

### DIFF
--- a/UserManagementView/js/studentView.js
+++ b/UserManagementView/js/studentView.js
@@ -56,11 +56,11 @@ function renderStudentView(data)
 		
 	}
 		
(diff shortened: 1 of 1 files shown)


a13antda:
Code: row: 63 " +"/VT-2016 </h3></div>";" is hard coded and wont work if the student has started any other year.  Also all the "vt-" and "ht-" is hard coded and wont work if a student starts in january.
a more versitile solution is needed.

PR #1592, opened 2015-05-21
Deleted redundant loading of files.

b12alela:
The dragndrop.php file loaded files that were already loaded, which made for some unstable behaviour with the resizing functionality of the page. The page would sometimes not recognize the function resizable from jQuery ui, this due to that the framework was redundantly loaded again. The problem did not persist after these deletions.


a12jiyba:
the deletion of the rows seems to solve the problem.

a13alfbe:
The solves the problem!

PR #1593, opened 2015-05-21
Update umv.css

b13frero:
Fixed the position for the Course name indicator


a13erilj:
Close commit due to conflict

PR #1598, opened 2015-05-21
Added form validation for links

b12erijo:
The validation function checks that all links has to start with either "http://" or "https://", this means that we can no longer use relative links. The validation can probably be customised to work around this however should the need arise.


a13denli:
Good

Issue #1577, opened 2015-05-21
login button


a12jiyba:
When a example name is too long then the login button disappear it do not matter if i have the web browser in fullscreen(its different in different resolutions). 
The solution will be that we could make so that the login button is always showing on the right side no matter if the example name is too long.

b13danso:
Didnt merge.

b13danso:
We merged a temporary fix for this, we changed the font in navHeading from 2.5em to 26px, fixing the problem with lower resolutions having to large text and pushing the loginButton away. This cant be fixed now because the whole navHeader-file is structured after a table.

PR #1607, opened 2015-05-21
#1577 - font size issue.

b13danso:
The font size on the class navHeading was defined as em instead of pixels. On lower resolutions, the text would scale up and cover the whole navHeader, hiding the loginButton. Instead of having a dynamic font size, we agreed to change it to  static pixel sixed. From 2.5em to 26px.

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -307,7 +307,7 @@ header td {
 }
 
 .navHeading {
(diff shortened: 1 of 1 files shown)


a12jiyba:
I agree with you that its better to use pixels then em. This will solve the problem for most people with different resolutions.

PR #1608, opened 2015-05-21
Fix(?) for negative number

a13henhu:
You're not able to input negative numbers


a12jiyba:
i tested your fix and its working. you can not use negative numbers now.

Issue #1606, opened 2015-05-21
Code Viewer: There is some design problems with the drag and drop


c13caral:
We need to space optimize the drag and drop in Edit Example

a12jiyba:
some small fixes left..

f12danjo:
Fixed by [#1662](https://github.com/HGustavs/LenaSYS/pull/1662).

PR #1611, opened 2015-05-21
Cv #1606

a13jansa:
Have solve the problem for issue #1606. It was some design issue for the drag and drop box, so i create some more space.


a12jiyba:
I have tested your branch. The new fixes are good.

Issue #1614, opened 2015-05-22
Highlight new color for Important words


a13jansa:
New color for Highlight words

PR #1615, opened 2015-05-22
issue #1614 new color for highlight words

a13jansa:
Change the color of highlight words. issue #1614


a12jiyba:
This is a small change to the color so its matching the theme. Its good so i will merge this but i will leave the branch open if we have another opinion in the group.

Issue #1604, opened 2015-05-21
Top of the content of preview window gets hidden


f12danjo:
It fixes itself as soon as you right-click and inspects the content of the preview box. But it bugs again if you refresh the page.

![1](https://cloud.githubusercontent.com/assets/11722176/7745945/28158b10-ffaf-11e4-935f-c75a51dcf7b9.png)
Before inspect.

![2](https://cloud.githubusercontent.com/assets/11722176/7745948/2e6b7448-ffaf-11e4-8899-65fd21d2627a.png)
After inspect.

f12danjo:
Fixed by [#1619](https://github.com/HGustavs/LenaSYS/pull/1619)

PR #1618, opened 2015-05-25
CodeViewer #1617: Iframe fix

### LINKED ISSUE #1617: Preview option dissapeared after merge.
There is no longer a preview option to choose from in the "Kind".
The css for the preview also seems to be missing...

a13alfbe:
#1617 
Fixes thing that dissapeared in the weekly merge.


a13jorla:
This makes it possible to use iframe exaples in codeviewer. And the code looks allright. But it does cause a new issue #1604 but that will be solved shortly. So i'm merging.

PR #1619, opened 2015-05-25
CodeViewer #1604: Iframe margin

a13alfbe:
* Added a missing peice of code that were used on the other cases ( "CODE", "DOCUMENT" and "NOT DEFINED").
* Also found a **'** too much in the iframe


f12danjo:
Looks good! Now the previews works just like intended.

Issue #1569, opened 2015-05-21
Able to edit the description box via important words


a13erisc:
For example, adding h*e as an important word, causes this to be the output in the description box.

![wwww](https://cloud.githubusercontent.com/assets/11733626/7743834/92faab10-ff9e-11e4-869a-3548fba1482f.jpg)

f12danjo:
Fixed by [#1623](https://github.com/HGustavs/LenaSYS/pull/1623)

PR #1623, opened 2015-05-25
Cv22 #1569 highlight special chars

b12simsa:
Replace all asterisks with html code (&#42;) before highlighting important words, after the highlighting is done change the html code back to asterisks.


f12danjo:
Seems to work! I tried the adding "~~~" as an important word too and it didn't bug out.

PR #1626, opened 2015-05-25
Update codeviewer.js

b12alela:
Adds code that disables the pointer events for all iframes while the resizing function is working. Which means you cant use the iframes while your dragging the handle to resize the page. The functionality is restored when you drop the handle.


a13alfbe:
You can now resize the boxes containing iframes without any problems!

PR #1627, opened 2015-05-25
umv22#1409 remake on years

b13davfr:
https://github.com/HGustavs/LenaSYS/issues/1409


a13erilj:
It works but there is still some fixing to do, therefore the issue will still be open

a13erilj:
Wil merge to save the solution if small fixes can't be fixed before the last merge

Issue #965, opened 2015-04-17
CodeViewer: resizing minimum sizes?


c12jenek:
Being able to resize the window (by the implemented click-n-drag) to the maximum width/height of the browser may be very inconvenient, maybe we should have a minimum or maximum size of the windows

HGustavs:
minimum and maximum resized sizes is good but the window itself should not be limited.

PR #1628, opened 2015-05-25
Cv22 #965 MinMax sizes on resizable boxes.

b12alela:
This makes all resizable boxes have a min and max size, which means they cant be resized into the edge of the screen. There will always be a minimum of 10% of one box showing.


a13alfbe:
Works great!
Just change the css to lowercase letters and then i will merge this.

a13alfbe:
Changed to lowercase letters, will merge now.

PR #1630, opened 2015-05-25
Updated - Now you're not able to input negative numbers, and also not…

a13henhu:
if (rowFrom && rowTo > lineno) { //Cannot add higher number, example we only have 1-10 rows, and it's should not be able to add the number 11 in the row editor
	       alert("You cannot add numbers that doesn't exist in the current row!");
		   editContent.reload();


f12danjo:
Seems like you are still able to add negative numbers in the first input box as long as the input in the second box is positive.

f12danjo:
Now it works. Found another bug though. Sometimes you are able to add important rows with a bigger value in the start row than the last row, which shouldn't be possible. Will make a new issue about that.

PR #1635, opened 2015-05-25
Ds22#1562 - Course status should be yellow aswell

a13chrju:
if a test is started the course icon should become yellow. check if the other grade icons work for when a grade has been set.


a13denli:
Try to make so it also turns yellow if a dugga was already started

Issue #487, opened 2015-03-26
Show which courses that are prereq


a97marbr:
In the study program student view hovering over a course with prereq should highlight in some way which courses that are prereq.

a13erilj:
This issue is assigned also to @a13freas, he will work together with @a13antda on the issue

a13antda:
This issue isnt completly done dont close. The function is there just needs to fiqure out what to do except for the hover effect!

a13ottne:
would like a separate smaller icon for each course on which to put the hover effect. So that you are still able to move the mouse somewhat freely without courses constantly fading.

a13ottne:
if "easily done" the icon should be as follows:
if student has points in all pre-reqs, but not completed all = warning triangle
if student has zero points in any of the pre-reqs = stop sign
if student has completed all pre-reqs = green checkmark

a13erilj:
This issue is handled by @a13freas and @a13antda

c13marme:
This issue is solved

PR #1637, opened 2015-05-25
Umv22#487

a13antda:
#487 

this varning signs now show proper.
The hover effect is only used on the courses that actually has a pre req course


a13antda:
http://webblabb.iki.his.se:8080/umv/a13freas/LenaSYS-umv22/UserManagementView/studentView.php

c13marme:
There is some console.logs still
Code looks okey

Issue #1613, opened 2015-05-21
Code Viewer:  It doesn't work to move around the others pop-up boxes 


c13caral:
It doesn't work to move the pop-up boxes like Login and Edit Content.

This problem I think is something that was made by us in code viewer.

c13caral:
Maybe just add a class to Edit Content. Now it is hard to example add important row when the code is hidden by the pop-up window. You maybe don't remember which row you want to highlight

f12danjo:
Fixed by [#1641](https://github.com/HGustavs/LenaSYS/pull/1641).

PR #1641, opened 2015-05-25
cv#1613

c12jenek:
#1613 Draggable is back into the party


f12danjo:
Nice! Now all windows except "Edit Example" are draggable.

Issue #1629, opened 2015-05-25
Select Sequence beforeid saves wrong value.


a13jorla:
This makes the before and afterbutton to stop sending you back and forward in the sequence.

b13danso:
Fixed.

f12danjo:
Fixed by [#1642](https://github.com/HGustavs/LenaSYS/pull/1642).

Issue #1585, opened 2015-05-21
Unable to move a sequence box if you have to scroll


f12danjo:
You are unable to drag a sequence box from "Unused Examples" into "Sequence Examples" if you have to scroll to get it there.

b13danso:
Got a fix for it, will be able to make a pull request next week.

b13danso:
Fixed.

f12danjo:
Fixed by [#1642](https://github.com/HGustavs/LenaSYS/pull/1642).

PR #1642, opened 2015-05-25
Cv22#drag and drop fixes

b13danso:
This is a fix for issues #1629, #1585 and #1478.

CSS added in codeviewer for the scroll, fixing the bug with scroll not enabling the dragging of example to boxes. Removed duplicate old css from style.css that messed up and overwrote the css from codeviewer.css. 

Fixed the issue where the beforeid got updated wrong and was set to the current example id instead of the example id before it.


f12danjo:
Works good. Came across some bugs but will make seperate issues for them.

Issue #1631, opened 2015-05-25
Disable the box resizing on mobile theme


a13alfbe:
The resizing should be disabled when mobile view is enabled.

c13caral:
this is now solved

PR #1643, opened 2015-05-25
Update codeviewer.js #1631

b12alela:
Adds functionality to destroy the resizable functions when the user is swapping over to mobile view. 

The functionality is restored when the user swaps back to desktop mode.

### DIFF
--- a/CodeViewer/codeviewer.js
+++ b/CodeViewer/codeviewer.js
@@ -213,6 +213,8 @@ function returned(data)
 	}
 	// Allows resizing of boxes on the page
 	resizeBoxes("#div2", retData["templateid"]);
+	//mobile resize fix
+	mobileDesktopResize("#div2", retData["templateid"]);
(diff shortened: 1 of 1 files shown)


a13alfbe:
I get the following error message in my console when resizing the browser window:
**Uncaught Error: cannot call methods on resizable prior to initialization; attempted to call method 'destroy'**

c13caral:
#1631

b12alela:
The problem is with still it still trying to call resizable functions when you are under 1100px to destroy those functions, that is why you get the error. A check need to be put in place here to check if the resizable functions are infact initialized and running. The more i look at it i see more problems i didn't see yesterday, the parentwidth should be window width. The same check i described in this post should also be applied to the else statement, because we don't want to initialize resizeboxes again if its already running.

c13caral:
I will close this pull request

Issue #1639, opened 2015-05-25
Important rows - Able to input a bigger number in the start row than the last row


f12danjo:
![1](https://cloud.githubusercontent.com/assets/11722176/7794957/80071932-02d0-11e5-94b9-8de864f98813.png)
Sometimes able to input a larger number in the start row than in the last row. Have happened when trying to add negative row numbers and then changing the input.

PR #1644, opened 2015-05-25
Update codeviewer.js #1639

a13henhu:
#1639

### DIFF
--- a/CodeViewer/codeviewer.js
+++ b/CodeViewer/codeviewer.js
@@ -469,7 +469,7 @@ function editImpRows(editType)
 	var rowTo = $("#improwto").val();
 	var row = $("#improwfrom").val() + " - " + $("#improwto").val();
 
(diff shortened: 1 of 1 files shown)


a13alfbe:
The code looks as follows: 
```javascript
if (rowFrom < 0) { //Negative numbers alert
	alert("You cannot input the negative numbers " + rowFrom);
	editContent.reload();
}
		
if (rowTo < 0) { //Negative numbers alert
	alert("You cannot input the negative numbers " + rowTo);
	editContent.reload();
}

if (rowFrom && rowTo < 0) { //Negative numbers alert
	alert("You cannot input the negative numbers " + rowFrom + " and " + rowTo);
	editContent.reload();
}
		
if (rowFrom && rowTo > lineno) { //Cannot add higher number, example we only have 1-10 rows, and it's should not be able to add the number 11 in the row editor
    alert("You cannot add numbers that doesn't exist in the current row!");
    editContent.reload();
}
```

It's a bit redunant to have this many if-statements. A simple way would be to add all these statements to the parent if-statement, and make one single alert box popup if the input is incorrect.

The second and third if-statement is also incorrect.

a13henhu:
@a13alfbe and I have been checking on this togheter, and we found an easier solution.

a13alfbe:
It is now bit less redundant, will merge now!

PR #1647, opened 2015-05-25
Update style.css

b13danso:
Updated the font-size of navHeading from 2.5em to 26px.


a12jiyba:
this solution works better in different resolutions. Good!

PR #1654, opened 2015-05-25
CodeViewer #1645: Before/after buttons fix

### LINKED ISSUE #1645: Sequence - When using the last example
When you come across the last example (example 14) the "next button" gets disabled even if it's not the last example in the sequence.

a13alfbe:
#1645 
The before/after buttons now work with the sequences.


f12danjo:
Good! Now the first example in a sequence has it's "back button" disabled and the last example has it's "next button" disabled. No matter which example id they have.

PR #1656, opened 2015-05-25
Removed vertical-align:baseline

a13denli:
Was creating some weird alignments on the website


c13andka:
works for us with out it

Issue #1653, opened 2015-05-25
Code Viewer: Tag not highlighted correctly (mouseover)


b12simsa:
When the last html tag is moused over the correct tag is not highlighted. The body tag at line 12 is highlighted instead of the html tag on line 2.

![errordumphighlightmouseover](https://cloud.githubusercontent.com/assets/11715509/7796522/57e7f4b6-02e0-11e5-9a51-df0342800f0c.png)

f12danjo:
Fixed by [#1660](https://github.com/HGustavs/LenaSYS/pull/1660).

PR #1660, opened 2015-05-25
Codeviewer #1653 Mouseover Highlight Bug in Codeboxes

b12simsa:
#1653
The <!-- comments where not recognised as comments.


f12danjo:
Works great. Now all HTML tags in "" comments gets ignored by the tag highlighting.

PR #1662, opened 2015-05-25
Cv22:1606 Design changes of edit example box

a13jansa:
#1606


f12danjo:
Very nice. Now the items in the "Edit Example" window are better aligned. Looks much better.

Issue #1599, opened 2015-05-21
Highlight bug in description box


b12simsa:
There seems to be a bug with mouseover highlight for certain inputs.

![errordumphighlightdesc](https://cloud.githubusercontent.com/assets/11715509/7745614/e9ec6450-ffac-11e4-9fff-da942c290be4.png)

b12simsa:
Cannot find this example anymore so I think we could close this one.

b12simsa:
Bug reoccured.

b12simsa:
The word appendChild existed twice in the sql-file used to initialize the database.

f12danjo:
Fixed by [#1663](https://github.com/HGustavs/LenaSYS/pull/1663).

PR #1663, opened 2015-05-25
Codeviewer #1599 Important word exists twice

b12simsa:
#1599
There where two occurences of the word appendChild, which caused som weird behavior.


f12danjo:
Nice find. A simple removal of the duplicate.

PR #1664, opened 2015-05-25
Cv#1651 and 1646

b13danso:
This fixes the issues regarding error message when there is no sequence and the user is promoted to create a new sequence. It also resets the codeexample table if there is no sequences active, reverting it to its former, orginal structure where ever example is listet after eachother until the final example is reached.


a13alfbe:
Works great. Nice that it goes back to the original order if there are no sequences!

Issue #1634, opened 2015-05-25
Stop dugga timer as soon as user turns in a dugga


a13karsi:
In the dugga ui, when the user finished a dugga and hands it in, the timer keeps on going. This could get confusing. So stopping the timer as soon as the dugga is handed in, even if the user is still on the dugga page should be a good fix.

PR #1669, opened 2015-05-26
Dugga stops timer when handed in

a13karsi:
With this update the dugga timer should stop immediately after a user has handed in a dugga. The code 'clearInterval(self.timer);' was not working for some reason and I don't know enough javascript to debug this without hassle so I just implemented a pretty naive solution which stops the timer updating the ui.

See issue #1634


a13chrju:
good job :+1:

PR #1670, opened 2015-05-26
Cv22#1665 - Icons and coloring of the sequence.

b13danso:
Added green/red color to the divs inside the example/sequence-boxes. Added three icons; one for when the normal state(no hover) is active, one where the user hovers the sequencebox and one when the user hovers the examplebox. Added nessesary jquery code to handle the mouseover/mouseleave-events.


f12danjo:
Works really good. The colors makes it really clear which examples are in the current sequence and which are not.

a13alfbe:
Works great! A lot clearer whats going on and it guides the user!

Issue #1343, opened 2015-05-07
Site icon and better site title


f12danjo:
I would be nice with a site icon (the icon on the tab in the web browser the codeviewer is open in) and a proper site title (right now the title is "Result Editor"). Perhaps the title should always be the title of the code example currently viewed in the codeviewer.
![1](https://cloud.githubusercontent.com/assets/11722176/7512708/14763bc8-f4af-11e4-9184-07b5d8c70510.png)

c12jenek:
The title is change in pull request #1344 to Code Editor v5, hope that an okay name!

f12danjo:
Fixed by [#1671](https://github.com/HGustavs/LenaSYS/pull/1671).

PR #1671, opened 2015-05-26
Adds site icon and dynamic title

a13erisc:
Currently using a placeholder icon that is prone to change. Title may also be changed accordingly. 
See #1343


f12danjo:
Strangely enough, this update seems to change the opacity of the "Edit Example" window.

f12danjo:
Now it looks normal again. Strange bug. 
The title changes as it should when flipping through the code examples. As for the site icon... I lold.

PR #1672, opened 2015-05-26
Added fixed sizes widths and changed the order

a13denli:
Added fixed sizes widths where you need to have them and swapped password and settings.


a13karsi:
Changes look good, tested.

Issue #491, opened 2015-03-28
Error message when no template has been chosen


HGustavs:
Currently, no appropriate error message is given when no template has been selected ant the user is
not an administrator (i.e. does not have write access).

If logged in we do get a message telling us to choose a template.

A good error message would be something like "The administrator of this code example has not yet chosen a template"

HGustavs:
Something like when we choose a course that does not exist.

a13jorla:
Is this implemented @c12jenek ?

c12jenek:
There are no examples where the template has not been chosen so am unsure, try adding an example for it and check

f12danjo:
Fixed by [#1675](https://github.com/HGustavs/LenaSYS/pull/1675)

PR #1675, opened 2015-05-26
Edited error message when no template exists

a13erisc:
Changed the error message to a more neutral message
See #491


f12danjo:
Nice. These are better error messages. Especially the one that shows when you are not logged in.

PR #1676, opened 2015-05-26
Cv#1674 - Move-cursor and draggable window header.

b13danso:
Makes only the window header, loginBoxHeader, draggable instead of the whole window being a draggable object. Added move-cursor to loginBoxHeader so that the user is alerted that clicking and moving will drag the window.


a12jiyba:
Good job. I like the when the cursors icon change and that the box is now draggable without "bugs"!

Issue #714, opened 2015-04-07
Adding 'duggor


a97marbr:
There is no way of adding a new dugga. Users should be able to use the example-duggas from the DuggaSYS/template folder, but this folder should not be used in the actual production system.

a97marbr:
For example:
```
../../duggor/
```
Could be used as a folder outside the LenaSYS code to hold all duggas.
There should exist a dialog that allows for uploading all files relating to a new dugga.
* new_dugga.html [required]
* new_dugga.js [required]
* potential other files, e.g., the png-files for the different colors in dugga2 in the template folder

To get a better structure on the web server I would suggest creating a sub-folder for each new dugga:
```
../../duggor/new_dugga/new_dugga.html
../../duggor/new_dugga/new_dugga.js
```

a97marbr:
After some discussions with Henke the following has been descided:
* We should use the existing file upload tool.
* Duggas should therefore reside with all the other files that can be uploaded

a13arnhe:
Currently there exists a way of adding a new dugga and a new dugga template into the template folder, I think that pretty much covers the issue in question.

PR #1677, opened 2015-05-26
Ds22#714 - adding duggor fix

a13chrju:
changed the filepath for Global files to reside under the files of the duggas so an administrator can upload javascript files and html files which later can be used for tests under "Dugga editor".

https://github.com/HGustavs/LenaSYS/issues/714


c13andka:
Seems legit

PR #1679, opened 2015-05-26
Fixed click counter

b12erijo:
Click counter now counts clicks properly in dugga 3,4 and kryss


a13denli:
Good

c13andka:
Works for me ! seems legit

Issue #1483, opened 2015-05-18
Undefined : undefined showing


c12jenek:
When not logged in and not at an example Undefined : undefined is the title f the examples, this should  be changed

a13jorla:
As for now, when you are going through duggasys you wont get this line in the navheader. So @c12jenek  should we close this issue?

c12jenek:
Is up to the groupleader, since when going with a direct url it's still shown

f12danjo:
Fixed by [#1680](https://github.com/HGustavs/LenaSYS/pull/1680).

PR #1680, opened 2015-05-26
Added null check

b12simsa:
#1483
If the object stored in the array is null, do nothing instead of printing the null value.


f12danjo:
Nice, now it says: "Example Section : Example Name" instead of "undefined : Exname".

Issue #1667, opened 2015-05-26
Pre-requirements in studentView don't work properly


a13erilj:
The pre-requirements in studentView do not work properly, the only course for which it works is operativsystem. Which has a stop-sign. When hovering over the stop-sign the required courses are not shown.

a13erilj:
Solved

PR #1682, opened 2015-05-26
Umv22#1667

a13antda:
#1667 

Fixed some reduncency in the DB where coursecode was stored at myltiply places. 
Also updated the sql query where this information is used.


a13erilj:
Looks good, seems to have fixed the bug :+1:

PR #1686, opened 2015-05-26
Update whiteTheme.css

a13jansa:
change the color of important words! issue nr #1614


a13alfbe:
Shouldn't the descriptionbox have the lighter color for important words?

a13alfbe:
Great, these colors look better!

Issue #1684, opened 2015-05-26
Cannot update template for examples


b12simsa:
If an example does not have a template, it is not possible to update the database with a template number.

PR #1687, opened 2015-05-26
Not able to change template

b12simsa:
#1684
A variable was named cversion intead of cvers.


a13alfbe:
Nice catch! It now works as intended.

Issue #1673, opened 2015-05-26
Edit example re-design


a13alfbe:
There are alot of weird sizing and css for the edit example box. Change the layout to make it look more like the edit content box.

f12danjo:
Fixed by [#1688](https://github.com/HGustavs/LenaSYS/pull/1688).

PR #1688, opened 2015-05-26
CodeViewer #1673: Edit example redesign

a13alfbe:
#1673 
* Changed the layout for Edit Example 
* Made it a bit more which part of the box belongs to which buttons


f12danjo:
This is great. The Edit Example window looks way better now with everything aligned. The addition of borders around the two distinct parts of the window and the relocation of the "Save" button makes the windows function much clearer.

Issue #1668, opened 2015-05-26
Search shows users that are not really available.


a13erilj:
The search filed in teacherView does not show the correct users in the dropdown suggestions. This leads   to problems when an incorrect user is selected.

c13marme:
This issue is solved

PR #1690, opened 2015-05-26
Fixed so that only users for the current program show in the search options.

a13freas:
#1668 

http://webblabb.iki.his.se:8080/umv/a13erilj/week22%231668/UserManagementView/teacherView.php


a13erilj:
Looks fine, should probably remove the console.log :+1:

Issue #1490, opened 2015-05-18
Guest ID fix


a13alfbe:
At the moment you get new guestID everytime your refresh the page. It would be nice if the ID was stored in local storage.

HGustavs:
This sounds like a very worthwhile effort.

a14ponse:
Since all the people that are going to use the page are students or teachers and therefore has an account and can log in we cant see any use in storing a guest ID for any one. And its no point to save an guest ID since you cant modify anything ether.

HGustavs:
It would be nice to be able to track users even when not logged in.

So I would like this feature together with the logging.

A local storage id that is used for users that do not have them. This way we can track usage patterns for users that do not log in.

HGustavs:
I think this is an issue for the logging group (group 3?)

a16joare:
Is this issue still an issue?

a17poner:
Bump @HGustavs

d17marle:
Minor issue needs to be changed, doesnt give out a new guest id once logged in and then logged out again.

b17carbr:
@HGustavs I have implemented guest users ids that persist for one hour by storing the id in cookies. At login the guest user cookie gets deleted. This was inline the original request of this issue, but further down in the comments there is talk about having the original issue together  with logging. Should logging be part of this issue, or moved to another issue? I can only find one place where guest are logged at this movement which is in codeviewer.

PR #1693, opened 2015-05-26
Cv#1490

c12jenek:
@a13henhu @b12simsa 
#1490 Added guest username logging capabilities


f12danjo:
Nice with the Guest ID showing in the top right corner. However it seems the Guest ID still changes every time you refresh the page or changes code example.

a13alfbe:
We don't have any time to finish this in time so i will delete this.

Issue #1696, opened 2015-05-26
cv: wrong font size


a12jiyba:
navheading have wrong font size. it should be 26px.

f12danjo:
Fixed by [#1697](https://github.com/HGustavs/LenaSYS/pull/1697).

PR #1697, opened 2015-05-26
Update style.css

a12jiyba:
changed font-size to 26px #1696


c13caral:
change this so this changes is on codeviewer.css instead of style.css

a12jiyba:
its done.

f12danjo:
Looks good. Makes the oversized heading a more reasonable size. This makes the site look better on smaller screens.

PR #1703, opened 2015-05-27
ds22#1701 Problem with scroll bars on result page

b13alebr:
Added so that the width on the div "tableheaderwrapper" updates depending on the total width of all duggas.


a13denli:
Working

PR #1704, opened 2015-05-27
Ds22#1501

a12majca:
Instead of using a prompt, if the user clicks "save receipt" the input for email is placed inside the receipt box. We also send an actual receipt now.


a13denli:
Working (Y)

PR #1709, opened 2015-05-27
ds22#1708 - Quiz dugga only show one question

b13alebr:
Added query that gets what type of dugga it is and fixed so that all questions on quiz duggas shows like thay should.


c13andka:
Seems to work

PR #1710, opened 2015-05-27
Disables resizable functions in mobile theme

b12alela:
Does as the title says, this updated version of the function now features checks before every action if it should be run or not.

The parent width have been changed to window width.

Now, if the window becomes smaller than 1100px the mobile theme triggers, and the layout of the site dramatically changes. This abolishes the need for the resizable functions, and this destroys those functions if they are found. However if the site later becomes bigger than 1100px it also reinitializes the resizeboxes function and makes the codeboxes resizable again.


a13alfbe:
Works now! This makes it easier to work on the mobile theme later.

PR #1717, opened 2015-05-27
Update testdata.sql (added shader-testdata to listentries)

a12majca:
added shaderprogrammering to listentries..


a12majca:
when you review this you need to have the latest files from code viewer.

Issue #1692, opened 2015-05-26
Confirmation for sequence deletion


a13alfbe:
There should be a popup before a sequence gets deleted that asks the user if the sequence really should be deleted. 

The sequence gets deleted immediately at the moment.

f12danjo:
Fixed by [#1718](https://github.com/HGustavs/LenaSYS/pull/1718).

PR #1718, opened 2015-05-27
cv#1692 - Confirm popup when deleting a sequence.

b13danso:
Simple fix to popup the user a confirmation popup when he/she presses the delete button instead of just deleteing the sequence right off. Issue #1692


f12danjo:
Looks good. Works as intended.It's good to require confirmation from the user for deleting sequences, so it is not deleted by mistake.

Issue #1689, opened 2015-05-26
Code Viewer:  Dropdown list


c13caral:
It would be nice if there is a dropdown list that presented all the playlinks in the edit example.

a15andau:
![screenshot 2017-04-06 kl 08 45 40](https://cloud.githubusercontent.com/assets/26709847/24741546/b2cfea84-1aa6-11e7-9ad9-1f6f6316239c.png)

Do you know the purpose with the play links @HGustavs ? As you can see in the image I've done a dropdown list but I'm not sure what they want the play link to do when Save is clicked. The issue is rather old too so it might not be topical anymore?

a17poner:
The playlink dropdown has already been added and currently exists, closing.

PR #1721, opened 2015-05-27
Cv22 #1689 runlink list

b12simsa:
#1689
Changed the runlink field, to a dropdown list. 
Thx b13danso


f12danjo:
This doesn't really have any functionality right now. However the play link drop downs functionality can be expanded in the future. Before this update there was free text, and that was not a good solution. Now it is a dropdown box. It does not link to the "codeupload" folder however. And it only displays the file that is in the database under "runlink". That means the file that will be run by the play button cannot be changed from the "Edit Example" window.

Issue #1702, opened 2015-05-27
Draggable funktion


a13jorla:
![draggable](https://cloud.githubusercontent.com/assets/11715474/7830573/f15db666-044e-11e5-9ef7-97125929136e.JPG)

If you drag the edit example box out of the webbrowser you can't get it back and you have to reset the page. Because the draggable handle is only on the loginBoxheader and if it is put outside the window you will not be able to get it back.

PR #1722, opened 2015-05-27
Draggable box boundaries

a13erisc:
Prevents the user from dragging the draggable boxes outside of window-space. See #1702


a13alfbe:
Nice, this works really well!

PR #1723, opened 2015-05-27
ds22#1712 - cant turn in dugga answers cause variable not defined

a13chrju:
https://github.com/HGustavs/LenaSYS/issues/1712


a13denli:
Good

PR #1930, opened 2016-03-20
This update supports placeholders in markdown files use by duggas. Ho…

a97marbr:
…wever, a major rewrite spanning many php service files has been done. We must verify that it works before merging with master!


a97marbr:
If we decide to use this code **we must update the database**

The following update must be done:
~~~ sql
UPDATE variant SET param = REPLACE(param, '*##*', '&quot;');
~~~
This will remove the homemade escape sequence '\*##\*' in favor of the html variant for " for all dugga variant paramters

PR #1993, opened 2016-04-05
Issue 1921

g13johbj:
Issue 1921, fixat kanter på hidden td's och lade till ikon på hidden


a97marbr:
We use English here on GitHub. Also, the icon looks very much like copyrighted material and needs to be removed immediately. We create all our graphical material in-house. Use the icon as a source of inspiration, but do the graphics yourselves.

a97marbr:
I removed the icon. Please proved one that is made in-house.

PR #2005, opened 2016-04-05
Grp1w14 issue 1746

c14emian:
Set a default value for HP


a14nikau:
You my also want to look at this again

c14emian:
Not sure what there's to look at, I've added a default value for HP which was the issue. There are a few other issues concerning the adding of courses but this specific one has been fixed.

Issue #1963, opened 2016-04-03
SQL code does not follow Code Standard.


HGustavs:
A lot of the SQL code (in all parts of the project) does not follow the code standard i.e. capitals on SQL code.

so code should read
~~~
SELECT filename FROM submissions WHERE type=2 AND length=2;
~~~

and not 
~~~
SELECT filename from submissions where type=2 and length=2;
~~~
...

furthermore we should avoid star in selects and list columns instead

It is much less error prone to write
~~~
SELECT filename FROM submissions;
~~~

than

~~~
SELECT * FROM submissions;
~~~

Also. all inserts should read 

~~~
INSERT INTO submissions(ColA,ColB,ColC ) VALUES(1,2,3)
~~~

and not...

~~~
INSERT INTO submissions VALUES(1,2,3)
~~~

even if that line does not produce errors.

And all sql code should end with ;

HGustavs:
Updating the code standard accordingly if some of this is missing...

Any other suggestions for code standard for SQL?

HGustavs:
I added this issue since I tripped up mystelf in the sql code....

I need to be really close to the screen to differentiate and vs. amd but I can see the difference between AMD vs AND very clearly.

a14didda:
In the cases where a select (or whatever query) will take all values from a table, do we still need to include every column instead of the star? Is it necessary to change this?

a97marbr:
Yes. Due to the fact that a table might get more columns in the future. Being specific is a very good approach, albeit cumbersome at times ;)

HGustavs:
It is more self-documenting that way, the star does not say anything about what we are using it for

~~~
SELECT * FROM person;
~~~

compared to

~~~
SELECT firstname,lastname FROM person
~~~

a14didda:
Okay! We will take care of that. However I need to ask: where is this benchmark table? I´m unable to find it in the database, but it is used in the parseagents.php file.

HGustavs:
It may have been removed. I am not sure that all that code works as intended.

a14didda:
Then I´ll just leave it there for now and if it´s needed, it will be taken care of in #1926.

a14didda:
After changing everything, the following lines produce errors if we do not use the stars there: 	

`$query=$pdo->prepare("SELECT COUNT(*) AS unmarked FROM userAnswer WHERE cid=:cid AND grade = 1 AND submitted > marked;");`

[...]

PR #2021, opened 2016-04-06
Grp2w14 issue #1963

a14didda:
Fixed SQL syntax in editorService


b14davbo:
Sweet. Seems to be working, good job!

Issue #1843, opened 2015-12-17
Document viewer / Code viewer title.


HGustavs:
Currently when one is viewing documents / code in tabs, the name of the document is not visible.
Adding the name of code example / document to title would be a worthwhile effort.

a13henlo:
This issue seems to be related to which browser you use, if you're using Firefox you'll get the result you want but it does not appear to be supported in Chrome/Explorer (not sure about other browsers)

HGustavs:
Im thinking this comes from the title element - is the title element static in some views?

The section editor seems to not do this correctly, but the code viewer seems to work at least in firefox.

a13henlo:
I believe we're thinking about two different matters here, could you take a screencap of the page you're on where the title isn't working?

a13henlo:
@HGustavs Unable to move forward with this issue without a sceencap

b14davbo:
If you press the back button from a file (or box) in a LenaSYS course, the URL doesn't show you the attribute "coursename" which leads to an empty title in the head tag.

a13henlo:
I will look into it

b14davbo:
Merged. Good job.

PR #2027, opened 2016-04-06
Better solution of #1843 with sessions

a13henlo:
This is a better solution of the previous one. Check here @b14davbo


b14davbo:
Seems ok.

Issue #1816, opened 2015-11-01
A left view split template in code viewer


HGustavs:
We lack a template for situations when there are two views on the left e.g. a short description and a preview in the left half of the screen and a longer piece of code on the right.

So in short, a mirror flip of the one large two smaller template.... making it a two smaller one large split instead.

b14davbo:
This issue has been fixed and was merged during weekly merge 7th of April 2016.

PR #2036, opened 2016-04-07
Grp2w14 issue #1816

a14jonal:
Added two new layout templates to the codeviewer to address issue #1816.


a14jimso:
It looks okay.

Issue #2026, opened 2016-04-06
If you add important word documentviewer does not work on school computers


a14jimso:
We do have a solution ready to implement and fix this, so this issue will merely be for follow-up and documentation for what have been done, and who fixed the issue.

For some reason, which we suspect is due to the schools computers not being up-to-date, the document viewer does not work when adding important words. 
It has been tested on students laptops, both mac and windows, and they work fine there. 
The error that is thrown is
"TypeError: can't supply flags when constructing one RegExp from another" in CodeViewer.js. Row 959.
According to Mozilla* documentation regarding RegExp, this is fixed in ECMAScript 6 and later, so this shouldn't have occured had the schools computers been up-to-date with support for this. Seeing as it worked on all the students personal computers that were tested.

a14jimso:
Related to #2008

HGustavs:
exactly what regexp is it that is broken? (quote code please so anyone can offer opinions if we have any)

a14jimso:
The code in question is this function in CodeViewer.js
`function replaceAll(find, replace, str)
{
    return str.replace(new RegExp(find, 'g'), replace);
}`

a14jimso:
An issue has now been merged.

PR #2041, opened 2016-04-07
fixed issue #2026

a14jonal:
changed one line from replaceAll() to replace() in order to avoid having a regExp inside another regExp further down in the code.


a14jimso:
Almost works, remove case insensitive.

PR #2056, opened 2016-04-07
Grp2w14 issue 1816 2

a14jonal:
added two new layout templates to the codeviewer. Issue #1816


b14davbo:
Wrong amount of commits, should be more.

Issue #1890, opened 2016-02-26
Code viewer shows select template button even if user is not an administrator


HGustavs:
![screenshot 2016-02-26 09 59 11](https://cloud.githubusercontent.com/assets/6978801/13347717/d3ee2012-dc6f-11e5-87a2-dd21ff054edf.png)

The select template button (third from the left) should not be shown if the user is not a super-user of some kind. The button is confusing for regular users as it has no effect unless you are an administrator. 

Furthermore the play button should not be shown if there is no data in the play button field (this as also been discussed in another issue)...

The only buttons that should be available in this screenshot is home, down, left, right buttons.

d11karjo:
After closer examination, this issue is not part of DuggaSys. I would like to remove the label, but it won't let me.

a14jimso:
This has been fixed now.

PR #2078, opened 2016-04-11
Fixed issue #1890

a14jonal:
Made sure that the choose template button only shows for super users. Also made sure that the play button is greyed out if there is no content to be played.


a14jimso:
Good job

PR #2114, opened 2016-04-13
Fixed Issue #1890

b14antfr:
Admin and superuser will both now see the template button and the other
relevant buttons that guests etc can't


a13henlo:
seems ok

Issue #1967, opened 2016-04-04
Fix installation


b14vikgu:
The installation script is broken right now. (entering a database other than imperious does not work since the SQL scripts have imperious hardcoded)

Implement a new installation method, perhaps using a web interface like other systems (ex Wordpress)

HGustavs:
This is related to other issues such as #479 that refer to the installer. It may be a good idea to link these issues.

HGustavs:
Web-based installer seems like a good idea.

Keep it simple though, and use the SQL file.

Installers that build the full sql using for example a php script, will make hand-building of the system without the installation script difficult.

So, an installer that builds the database using the sql file is fine. But that SQL file needs to be used.

b14vikgu:
> This is related to other issues such as #479 that refer to the installer. It may be a good idea to link these issues.

How do I link issues?

b14davbo:
@b14vikgu You type a hashtag (#) followed by the issue ID, e.g. for this issue you would type ```#1967```

b14davbo:
Install script has been fixed, see the branch "install-issue" https://github.com/HGustavs/LenaSYS/compare/install-issue

You can now chose another database name than "imperious" and also get to chose if you want to include testdata.sql when you initiate the database. I can make a pull-request of this into group 2 since we're handling the database.

b15linek:
A bit problematic with a new install when a database already exists. Must add an option to write over existing database.
Must make it easier to execute the bash commands at the end of the installer. 
Should write out the last SQL line when error occurs.

a15eriho:
Current issues that need fixes:
* Permissions for all of LenaSYS needs to be, at least, "www-data" for the group. Manually do the "chgrp -R www-data <LenaSYS-dir>" will do. This should preferable be done immediately after the initial copy of the Lena-dir. 
* Create directories for uploads. "LenaSYS/courses/global"
* Create test files specified in the test-data. These files are specified in "testdata.sql".
- Maybe these should be shipped with the installation, in a "testdata"-directory, and copied with the installer-script if the installer want them installed.
- The SQL-file "testdata.sql" maybe should be in this directory also.

Please comment things need be created by the installer.

b15linek:
A suggestion is to move the installer to an install directory inside LenaSYS and put all files needed inside it (SQL files, example files etc.) All installation will be make from inside this folder and the user is free to delete it once installation is finished. Is this a good idea? @HGustavs

HGustavs:
[...]

PR #2116, opened 2016-04-13
Install issue #1967

b14davbo:
You can now chose another database name than "imperious" and also get to chose if you want to include testdata.sql when you initiate the database.


b14davbo:
Related to #1967

b14davbo:
Wrong base branch.

Issue #1886, opened 2016-02-25
Duggas that are hidden in the section view should not be part of the result overview


a97marbr:
Duggas that are hidden in the section view should not be part of the result overview.

PR #2120, opened 2016-04-13
Duggas that are hidden in the section are no longer part of result ov…

a14maxry:
…erview


a14maxry:
its issue Duggas that are hidden in the section view should not be part of the result overview #1886

PR #2204, opened 2016-04-20
issue 1765

a14ponse:
Fixed the down bumping of the code when it was higlighted.


b14antfr:
Looks good, easy and nice solve!

Issue #2237, opened 2016-04-25
Markdown toggle gif not correct size after loading page


b14davbo:
After you load a page with a gif that is toggable the size is not correct. It is way to big.

a14jimso:
Should be fixed

PR #2238, opened 2016-04-25
Fixed issue 2237

b14rebst:
Markdown images were shown in bigger size than thumbnail, all images
should be thumbnail size at first glimpse. Therefore this code could be
deleted.


b14rebst:
#2237

a14jimso:
Well done, have a cookie.

PR #2251, opened 2016-04-27
fixed so link is added to item

a14maxry:
items of type "links" did not get the link value before


d11karjo:
Solves the exakt same problem as Pull Request #2250.

PR #2302, opened 2016-05-02
Grp2w18 issue 1917

b14antfr:
Removed and fixed undefined css stylings.


a13henlo:
seems ok 
@a14jonal check it out

Issue #1876, opened 2016-02-23
Font size setting


HGustavs:
It would be nice to be able to set the font size for the code view pane (decided for each pane by superuser)

Depending on course, verboseness of the languages differs greatly.

One could argue on a larger font size for PHP and a smaller one for Android in Java.

HGustavs:
But not a "setting per language" since some other course using php/javscript/etc may have another preference.

a97marbr:
I like this

PR #2357, opened 2016-05-09
Solved issue #1876

b14davbo:
Fixed issue.
Database update. Major CSS updates. It is now possible for a superuser to change font size on different boxes. Fixed wrong HTML comments which used 3 "---" instead of 2 which caused my IDE to render code incorrectly.


a14jonal:
Seems to work, good job.

PR #2469, opened 2017-03-28
Deleted redundant file

a15freku:
Suggestions should be placed in issues, not in the source code!


a15elibr:
Alright, seems logical!

PR #2484, opened 2017-03-29
Added column fontsize to init_db

a15freku:
Added column fontsize to table box.

### DIFF
--- a/Shared/SQL/init_db.sql
+++ b/Shared/SQL/init_db.sql
@@ -1,6 +1,6 @@
-drop database grupp01;
-create database grupp01;
-use grupp01;
+drop database imperious;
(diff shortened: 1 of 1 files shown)


Shared/SQL/init_db.sql
```diff
@@ -1,25 +1,25 @@
-drop database imperious;
-create database imperious;
-use imperious;
+drop database grupp01;
```
a15elibr: You realise this isn't going to work when we do the weekly merge?
I guess we could merge this now, and change it right before we do the actual merge.

a15oscwe: This is now solved!

PR #2490, opened 2017-03-29
#1967 create installer

b15linek:
Created installer in php. Will not work without removing database usage in sql files.


c14emian:
Seems great, creates a fresh and clean installation of a new database by filling out form. Gives the option to automatically fill the database with testdata and gives simple intructions on how to get things started with copy/pastable data. It's a simple and easy to use way to create a working database without any SQL knowledge.

PR #2494, opened 2017-03-30
changing default value of term in table user_course in init_db.sql

d15johro:
changing default value of term from "vt16" to "1" in table user_course in init_db.sql


a15wilhe:
checked default value is set to 1 of term

PR #2497, opened 2017-03-30
Updated readme to include install instructions.

b15linek:
Not nicely formatted but it tells you what to do.


a15andau:
The instructions works and are easy to follow.

Issue #2489, opened 2017-03-29
Modify SQL-files for using with new install.php.


a15eriho:
Is required for installer fix: #1967

Removes selection of database in beginning of SQL-files.

PR #2500, opened 2017-03-30
#2489 removed lines

a14phili:


b15linek:
This works fine, the correct lines are removed.

PR #2519, opened 2017-03-30
Checking name works when pullrequesting

c14joalu:
Name didn't work properly when pullrequsting, name displayed as UNKNOWN


b15matvi:
testing the new name

Issue #1236, opened 2015-05-04
Login: Forgot password doesn't work


c13caral:
The forgot password function doesn't work.

b14davbo:
Students will never be able to get an email since the smtp port is blocked on the server, we have no good solution to this right now. One way would be to store all forgot password requests in a database table and then the teacher manually has to send the student a new password?

Any other ideas?

a15tobli:
New screens added for forgotten password in merge #2528.

Issue #746, opened 2015-04-08
Making the "forgot password" window work


c13chrjo:
Add the possibility to add a mail to a user in the database
Send password or a link to reset password to said mail

c13chrjo:
This issue will be present after 10-04-2015 week merge

c13chrjo:
There is already a column for mail addresses in the databases

HGustavs:
Sending email is not supported by the system. Meaning that the current solution (mailto links) must be available as an option, perhaps sending email can be added as an optional solution somehow.

HGustavs:
Not high priority as it is useless for us (we are the only konown user of the system at this point).

a15tobli:
Closing issue as #1236 brings up the same issue and mail is not supported by the system.

PR #2528, opened 2017-03-30
Added screens for forgot password

a15ponja:
Hello! The screens for changing password using a security question has been implemented, however additional styling may be necessary. The code driving the functionality still needs to be added.

Referring to issue #1236 and #746

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -979,7 +979,7 @@ input.submit-button {
     font-family: "Helvetica Neue",Helvetica,Arial,sans-serif;
     background-color:#614875;
     border:0px;
-    width:110px;
+    width:120px;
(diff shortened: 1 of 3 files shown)


e14johho:
Hi! 
1. Add some comments in the code to make it easier for other programmers to understand the code. 
2. The "BTW: maybe not the perfect solution to use a label for the question" dont really belong on the website on the security question page.  
3. Maybe we can move down the question so it is displayed under the text "Your security question is:".
4. The abort text that is displayed on the security page, what do you think about changing that to "Log In" so its the same everywhere? Abit easier to understand I think. 
5. After you answered the security question you are redirected to a page where you write in the new password. The button "Change password" isent displayed properly.

PR #2538, opened 2017-03-30
Installation to fill with test data

b15linek:
Fix for original issue #1967 to create new installation script.
This is now fully working and fills database with test data.


a15jorpa:
Code seems ok. Instructions incomplete, user is dropped into sqlite shell and has to manually exit. sqlite 
DB file is not created after command is run. Path to sqlite file missing last slash.
Suggestion: Require directory to be writeable by web-server during install, possibly tightening up permissions when install is done.

b15linek:
Solved issues with instructions and added bash script to easier create the coursesyspw.php file.

PR #2549, opened 2017-03-30
Added default value to fontsize

a15freku:
In case it is not specified


a15freku:
This is not tested yet

a15wilhe:
Looked good! I checked that default value is now 9. Hope it wont collide with master.

Issue #2537, opened 2017-03-30
Move inline CSS to the right file


A15cecal:
There's some inline CSS in contribution.php that should be moved to the right external file

HGustavs:
This is a very good work-task. Feel free to do this at your convenience.

a15elibr:
Closing issue because another issue has been created for the fix.

PR #2551, opened 2017-03-31
fix-for-issue-#2537

A15cecal:
fixing issue 2537. Moving inline css in DuggaSys/contribution.php to external css file Shared/css/style.css


a15elibr:
Closed the original issue #2537.

Issue #2420, opened 2016-11-02
We need some kind of "past deadline" notice


a97marbr:
When a student is submitting a dugga that has a deadline that is passed there should be some kind of configurable message.

Some courses would say "Deadline has passed. No more grading will be done until next time the course is given"
Another course could have a different approach: "Deadline has passed. Grading will be done if the teachers have time. Otherwise next time the course is given"

I propose some kind of icon, a warning sign symbol?, together with the yellow stop light to indicate that this dugga is not guaranteed to be marked.

a15tobli:
Hello! We have split up this issue into smaller issues. As our solution, a text string is added on the receipt when a deadline for a dugga is missed. The string implies that the dugga might not be graded for a while, which was implemented in issue #2558. However, this string is not configurable at the moment. Opening issue #2569 which will be used to implement a feature for the teachers to set messages on the receipt.

Also, a warning triangle has been made, #2567. The idea is to put this triangle at the position where the lamps used for displaying status of duggor is set in the dugga menu.

a15tobli:
We have added a standard message for when the deadline have been missed on a submitted dugga. Also added warning icons to the receipt and dugga viewer. Working on support which allows teachers to configure the message which is shown for each dugga in the dugga editor. 
We have an idea to allow grouping of duggor, which makes it easier to set the same message for several submission, e.g. selecting every dugga in a course to show a specific message.

a15tobli:
This issue should be solved now, the messages can be set in the comment field for every dugga and it displays properly when a deadline is missed. 
The only thing not implemented is functionality to set message for several duggor at once, is this something that is desired? If not, this issue can be closed.

a15ponja:
Closed due to issue cleanup, was done.

PR #2558, opened 2017-03-31
Added text if deadline has past 

a15nikho:
Text has been added to the pop-up window if the deadline has past.


a15alica:
Looks pretty good, adds an extra message if you hand something in after deadline. Just a few changes needed.

<p>-tag isn't closed after the added sentence. (In Shared/dugga.js.)

Would be more clear what things do if the comments were places above the code line instead of after the code (on the same line).

Possibly add spaces when assigning value to variable (as done in Shared/dugga.js).

a15nikho:
Changes has been made. There is a comment on issue #2420 aswell.

Issue #2557, opened 2017-03-31
Removing resultlisted in DuggaSYS, since it is a placeholder only


b15sarha:
removing fowolling files:
resultlisted.js
resultlisted.php
resultlistedservice.php
* the resultlisted was never implemented and is no longer needed.

a15elibr:
Closing issue since it has been resolved.

PR #2559, opened 2017-03-31
w13-fix-for-issue#2557

b15sarha:
these files no longer needed


b15sarha:
Removing resultlisted in DuggaSYS, since it is a placeholder only #2557

PR #2567, opened 2017-03-31
Icon in png-format

a15filba:
A warning triangle to show if a dugga is not guaranteed to be marked.
Now in PNG for better use than SVG.


a15tobli:
This icon will be used for #2420

PR #2594, opened 2017-04-03
Moved remaining example data

a15oskst:
The array points started with example values. Moved this to diagram_example.js


b15matvi:
Looks good.

Issue #2527, opened 2017-03-30
PHP example file not found


a15freku:
"File: PHP_Ex2.txt not found."
"File: PHP_Ex2.php not found."

Files exists and are located in /LenaSYS/DuggaSys/templates folder but the webpage can not find the files for some reason.

HGustavs:
Im thinking not added to file list with uploaded files or permission problem?

More details.

Can other files be read?

a15freku:
The filename is mentioned in the database, all files have group:www-data and g=rwx permissions. Still not working. Which "file list" do you mean?

a15freku:
Fixed this by doing: "insert into fileLink (fileid, filename, kind, cid, isGlobal) values (2,"PHP_Ex2.txt",2 , 1, 1);"
It's the descriptionbox in php example 2, now we just need to do the same for the code box and all the other examples. I suggest to add this insert to testdata.sql until better solution is available. So we can focus on the markdown issues now that we can test the information in the boxes.

To learn more read the code on line 420 in codeviewerService.php

a15elibr:
Closing issue since the issue has been resolved.

PR #2600, opened 2017-04-03
Added testdata for code box

a15freku:
Added data to fileLink table to make connection between file and code
box in php example 2 see issue #2527


a15freku:
To make this work you have to create the file that is displayed and this is because "/courses" is in the gitignore-file.
Path to file:
LenaSYS/1/45656/PHP_Ex2.php

Issue #2548, opened 2017-03-30
Add table in dugga editor


a15osmla:
Add a table to show submission ways in a list and you should be able to delete rows in the table.

Related to #2547, #2545, #2544, #2546

HGustavs:
Dont quite understand this. More details please.

HGustavs:
Are we saying that a table should be created that contains the names of each of the submission types e.g. PDF, ZIP, LINK etc? 

That sounds like a good idea, if we want to add more submission types a first step would be to insert a row into that table....

I suggest 

Storing an ID, a short name, a long name and a description

...

We can show the name in the dropdown but also have a description as a tool-tip.

b15matol:
Hello HGustavs,

This and the referenced issues together make up a larger issue. We had some problems interpreting the JSON-dugga problem. Yesterday we realized that when editing a dugga, JSON is used as a "variant" and edited through a popup window. 

So the person editing must write the variant as JSON manually which is a painstaking task. Instead of writing it manually like that, we created these issues to come up with a solution to dynamically create the JSON through a form, or table, of input elements. Some javascript function can then create a JSON string out of the inputted data. 

Something like this, the old solution is at the top and a draft of the proposed new the bottom of the picture: 

![capture](https://cloud.githubusercontent.com/assets/26713381/24543073/69b8f964-15fe-11e7-81e5-3998165fb36f.PNG)

Hopefully this will result in the fact that the editor does not need to edit the JSON string manually, but that is created through the form instead. 

BR b15matol

HGustavs:
![screenshot 2017-03-29 20 01 27](https://cloud.githubusercontent.com/assets/6978801/24550013/76c7c78e-161c-11e7-8357-d9e8b4baf923.png)

Something along these lines from the image above, dont remember which issue I posted that image in...

a15tobli:
Merged in #2601, closing issue.

Issue #2547, opened 2017-03-30
Add forms in dugga editor


a15osmla:
Add instruction and submission form to current JSON-editor. Instead of having to write the JSON-code by hand you can fill in the forms and get the JSON written for you.

(Related to #2544, #2545, #2546 and #2548)

a15tobli:
Merged in #2601, closing issue.

PR #2601, opened 2017-04-03
Dugga editor - forms and table

a15alica:
Added forms and table for easier change in instruction text and
submission ways.
(Issues #2547 and #2548)


a15nikho:
Seems good, there is some problems in duggaed.php.

On row 82, there is a "<" thats not needed. It is also red marked in Github.

On row 88, indentation is needed for div with id duggaSubmissionForm.

a15nikho:
Everything looks good.

PR #2602, opened 2017-04-03
Revert "Moved remaining example data"

b15matvi:
Reverts HGustavs/LenaSYS#2594


b15matvi:
This fix didn't work, reverting

PR #2604, opened 2017-04-03
Fixed #2580 - correct mousepos for chrome,edge and

### LINKED ISSUE #2580: Diagram doesn't work properly with Chrome
When trying to move objects in the diagram in chrome the mouse is aligned all wrong. Needs fixing.

c14joalu:
Changed diagram.js and diagram.php so that mousepos is working correctly for chrome,edge and firefox


a14davlo:
Mats Gustavson review it please

PR #2611, opened 2017-04-03
Bugs, animation and clarity of collapsable sections

a15mikla:
Fixed the bug where the sections are created into div's without classes.
Separated content for "header" and "section". Also added an animation
when the section collaps. "Sections" are now collapsable to. "Moments"
and "Sections" now have a pointer as a cursor and arrows to clearify
that they are clickable to collapse the content related to them.


d15eriso:
Looks fine and code is consistent with the changes.

PR #2612, opened 2017-04-03
W14 close tag was missing

b15sarha:
in show ShowDugga.php


a15herny:
Seems to successfully close the head tag. I branched and checked other instances and found no errors with head tags.

a15herny:
Seems to successfully close the head tag. I branched and checked other instances and found no errors with head tags.

a15oscwe:
workspace.xml was included. We need to revert this merge.

PR #2614, opened 2017-04-03
syntax error

b15sarha:
syntax error in fileed.php line 78


b15sarha:
a semi colon was missing in that file.

Issue #2521, opened 2017-03-30
Change markdown processing for javascript


a15elibr:
~~Right now markdown processes line by line but it would be better to process an entire file instead.~~

Right now markdown processes an entire file instead of line by line. Would be better to have it line by line.

a15elibr:
This issue should be split into smaller parts.

d15johro:
it's the other way around. should process row for row.

a15freku:
Yes you are right its the other way around

a15elibr:
My mistake, got it wrong. Fixed now.

a15freku:
Great!

d15johro:
Where to test markdown processing related issues in LenaSYS? You can test markdown when adding a user in "/LenaSYS/DuggaSys/accessed.php", owever, this feature doesn't work at the moment... can you test markdown somewhere else?

a15freku:
Yes in webbprogrammering PHP Example 2 in the descriptionbox. You have to edit the PHP_Ex2.txt file in the templates folder.

d15johro:
issue is closed

PR #2632, opened 2017-04-03
Changed markdown processing in markdown.js

d15johro:
markdown.js is now processing file line by line when not console block or code block. this is handled by function parseLineByLine().


d15johro:
Closes #2521

PR #2633, opened 2017-04-03
Asking browser permission for sending notifications

c14caran:


a15timbe:
Looks good, found nothing wrong with the styling of the code, and the code itself also seems correct.

Issue #2652, opened 2017-04-03
Scaling a diagram


a15admsa:
A possibility to zoom out and zoom in would be nice for the diagram. This is part of the issue #2472.

b15matvi:
Issue resolved.

PR #2655, opened 2017-04-03
Fix for issue #2652

a15admsa:
I have made the canvas size responsive so that it looks good for everyone plus the possibility to zoom.


a15chrau:
You forgot to pull G4 before commiting, causing recent merges to revert.

Issue #2642, opened 2017-04-03
Application console window should be possible to show/hide


b15matvi:
The application console window should be hidden from PC and possible to show on mobile for debugging purposes for when we start testing touch

PR #2664, opened 2017-04-03
#2642 : partial done

c14joalu:
Fixed button for hide and show console, also made the canvas area resized when console is hidden/shown, when pressing the button


b15matgu:
Looks good.

Issue #2546, opened 2017-03-30
Interpret form data and create a JSON string


b15matol:
For use with the dugga editor.

Related to #2544, #2545, #2547 and #2548

b15matol:
Was merged in #2666.

PR #2666, opened 2017-04-03
Javascript logic to extract JSON from the dugga submissions form, and some redesign of said form.

b15matol:
Some functions that generates a JSON string from a form. This is used
in duggaed.php and .js. The functions sets the generated JSON string to
the "param" field. To make it work, the newly created form needed some
redesign aswell.


b15matol:
Is part of https://github.com/HGustavs/LenaSYS/issues/2546

a15osmla:
Code looks good :) following the LenaSYS coding standard. The JSON editor produces valid JSON. The JSON string is updated accordingly while pressing "create JSON" and it's possible to add additional submissions.

Issue #2668, opened 2017-04-04
Fix for issue #2581, removed unnecessary files.


b15matgu:
IDE-files accidentally added. This fix removes them and adds .idea to the gitignore-file.

a14davlo:
The files have been successfully removed from the LenaSys website.

PR #2669, opened 2017-04-04
Fix for issue #2668

b15matgu:
Removed unnecessary files. .idea


a14pethu:
Seems to fix the issue.

Issue #2665, opened 2017-04-03
Create some basic CSS classes like .left and .right


b15matol:
For easier aligning using float.

HGustavs:
Floats should almost *never* be used.

Floats are for positioning images inside text paragraphs, like images in magazines, any other use of floats should be considered a bug. 

Most use of floats like this breaks when the window/page is resized.

HGustavs:
If what one considers styling is in the shape of a table e.g. file list, course list, student list,result list etc we use tables...

Currently we also use tables for layout, since flex is a relatively new development...which did not work in internet explorer or android browser, something that has improved recently, If anything, designing menus using flex instead of tables should perhaps be an issue.... together with removing any use of floats for relocating images in text paragraphs.

http://caniuse.com/#search=flex

b15matol:
Hello HGustavs. The idea with this is not so much the floats, but to use basic classes to try and get away from use of inline CSS. However, i think the availability of such .left and .right class could be useful to the project, aswell as a no-float-class that removes float from the HTML objects. I also think it could be useful to have a .bold class, for example, that sets font-weight to bold.

a15tobli:
Hello Henke!
Would your suggested method consist of removing the floats from this issue and make a separate issue to rewrite the layout so that no current floats are used? 
Or could this be implemented as of now to make the placement of elements by using the current layout, and then add the issue for rewriting this as something that can be done at a later point? 

The later option would require some extra work in the future, but the classes added could be used right away to add new features in the layout.

HGustavs:
Question is how widespread the problem is, I am unsure as to where in the user interface we are talking about.... we first need some kind of triage, then we can work on mitigation.

I would say, that making it easier to use floats is not a solution to anything.

Is this in a menu or more generally used in the interface?

a15tobli:
The editors have been using this quite alot it seems. I haven't really looked into how spread this issue is above that, but it appears to be something that's mostly affects them. One example is the current dugga editor that is filled with floats for aligning the content. I'm not sure how big of a task it would be to rewrite those to remove the usage of floats, but that definitely sounds like something that could be added as a new issue.  

Note taken, the css will be fixed so that the float classes are removed before merging the  added classes to the stylesheet.

a15tobli:
Closing issue, it was a good idea to add classes for styling, but they have no use and would be too complex to use when each class contains such small amounts of code.

PR #2672, opened 2017-04-04
Created some basic CSS classes like .left and .right

b15matol:
Those can later be used to replace some of the extensive inline used in
LenaSys.

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -1794,7 +1794,41 @@ input.large-button {
     line-height: 47px;
 }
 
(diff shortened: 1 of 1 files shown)


b15matol:
Refers to issue https://github.com/HGustavs/LenaSYS/issues/2665

a15tobli:
Hello! Did you read the comments Henke wrote in issue #2665 about using floats as alignment in the css code?

a15tobli:
Holding off this pull request until we recieve further response from Henke about using floats in the code in issue #2665.

PR #2673, opened 2017-04-04
added a function to remove diagram and a remove button

a14pethu:
i have fixed the problem that were in my old commit.


c14joalu:
Everything looks good!

PR #2681, opened 2017-04-04
added-updategfx-for-button-actions

c14joalu:
added updategfx, so that the display updates when changes are made


b15matvi:
Good!

Issue #2572, opened 2017-03-31
Add warning icon to dugga viewer


a15tobli:
Referring to the mockup in #2420. The warning icon should be implemented to the dugga viewer when a deadline is missed.

a15tobli:
Merged in #2682, closing issue.

PR #2682, opened 2017-04-04
Warning triangle in the dugga view. Issue #2572

a15filba:
A warning icon is now visible in the dugga viewer to warn the user when
he has submitted after the deadline has past.


a15ponja:
When i tested it worked fine. Might be worth to remove the else clause since its empty anyhow.

Issue #1937, opened 2016-03-28
Layout/Styling of Assignments


HGustavs:
This is the layout I am currently using for Shader Programming

![screenshot 2016-03-28 11 49 11](https://cloud.githubusercontent.com/assets/6978801/14075664/40d7774a-f4db-11e5-89d9-e68d5bb1f370.png)

I would very much be able to discard the header as it does not make sense with the layout as it is right now. I feel that the header is out of place since the content is basically the same (Assignment X: Content of assignment X) as the header....

Preferrably I would want the styling of the header to also appear on the assignment. This way we get a clear division without having to add a header that is not useful in this case.

So, in short I vote for also adding the "button" style (I am undecided if same color or divergent color ) to the assignments, so we do not also need to add a header to the assignment to clearly show the division between the assignments.

HGustavs:
Also, is not material design-ish.

I would like to see if it is possible to add material design drop shadows on the assignments/headers if possible without changing the html too much e.g. keeping the table layout mostly as is.

a14gabon:
Please clarify the issue. Do you want a header? Or do you want a button?
On the live version of shader programming right now there isn't a header on the assignment.

HGustavs:
Either something like this
![screenaex](https://cloud.githubusercontent.com/assets/6978801/14242955/f891d494-fa52-11e5-8e6e-f61f6e19476f.png)

or even better something like 
![screenbex](https://cloud.githubusercontent.com/assets/6978801/14242965/062e1306-fa53-11e5-9bb5-ff9962686eb9.png)

If we show the assingments clearer, like in the screenshots, no need for an additional header.

So. my suggestion is to keep the headers as is (as they have other uses).

But to change the assignment one to look more like the header.

a97marbr:
I like the second mockup! :+1:

a15mikla:
I've added an option where you can choose if a dugga should look as a standard dugga or a header. If header is chosen, the dugga is styled according to the first mock-up you suggested. @HGustavs

a15mikla:
Duggas can now act like headers, according to the image:
<img width="184" alt="somethingissue" src="https://cloud.githubusercontent.com/assets/26705198/25222993/61ffeba8-25ba-11e7-980d-d2dca9d33cf2.png">

a15magkn:
@HGustavs  @a97marbr 
[...]

PR #2683, opened 2017-04-04
#1937

a15denbi:
Created border-radius around the sectionalist-elements in order to make
them feel more true to the Material Design principles.


a15andau:
This works, look good!

PR #2686, opened 2017-04-04
Field has been added in Edit Dugga View

a15nikho:
A field has been added where teachers could add comment on duggor.  The comment will display on the pop-up window together with the reciept. A connection to the database has also been added.


a15tobli:
Closing pull request, as a little bit of extra fixes before merge have been found.

Issue #2479, opened 2017-03-29
Add challenge question and answer for change of password


a15tobli:
This issue would require 2 new columns in the user table containing the information in the database.

Every user would have an individually set challenge question which will be used as an extra step of verification when attempting to change password. By storing this information in the database it can be used to change password.

a15tobli:
Closing issue as this was implemented in #2687.

PR #2687, opened 2017-04-04
Added two columns to the database and functionality to fetch a user security question 

a15ponja:
Added functionality to fetch a user´s security question via the forgot password button. And also changed the database in order to have a question to fetch. Solves #2479 and makes progress on #1236 and also #746


a15nikho:
Everything looks good! Logical names for variables are used which follows the code standard. The database changes looks fine and the install script will work with the changes.

Issue #2657, opened 2017-04-03
Close diagram figure. [Sub issue of Figure]


b15matki:
A function must be implemented so that the user must be able to close the figure.
Root issue: #2584

b15matvi:
Issue resolved

Issue #2645, opened 2017-04-03
Create "tolerance areas". [Sub issue of Figure]


b15matki:
When drawing a figure it might be complicated to be precisely at the correct pixel. Therefore a tolerance area should be implemented so that a user might be able to complete a figure just by selecting the starting line's tolerance areas. Example:

![example2](https://cloud.githubusercontent.com/assets/26705132/24609695/0ab36a6e-187c-11e7-801d-e781d8484456.png)

The blue colour represents the figure, the red rings represents the tolerance areas. Root issue #2584.

b15matvi:
issue resolved.

PR #2699, opened 2017-04-04
Fix for issue #2645 & #2657

a15chrau:
An area for tolerance has been added, and is defined by the variable tolerance. A figure can now be closed, but it does not exit the session.


a14davlo:
Well done Chrau!

PR #2704, opened 2017-04-04
Analysis page created

a15andau:
Made an analysis page that is empty right now. Can be used for statistic analysis and data.


d15eriso:
I looked through the code and it seems to be consistent and correct.

Issue #2616, opened 2017-04-03
Add attribute for course table in database


a15tobli:
Add a varchar to the course table in the database. This varchar will be filled with the value from the text field in  issue #2615. This text will be shown on receipts of missed deadlines in every dugga in the course.
EDIT: The way of setting the receipt message have been changed from the course view to the dugga view, which makes it possible to assign a different message to each dugga. This makes it harder to set the same message for every dugga in a course however.

a15tobli:
We changed the idea to add the comments from the course view to the dugga view. This makes it harder to set the same message to every dugga in a course. We are suggesting to implement a way to mark several duggas as groups, or set a message to all duggas. We are kind of stuck of which solution would be the best one, or if there is another solution we haven't been thinking on.

a15ponja:
Closed due to issue cleanup, not necessary anymore.

PR #2705, opened 2017-04-04
Added a comment field in Edit Dugga View

a15nikho:
A field has been added on the Edit Dugga View where teachers are able to
write a comment that will show upp for students when they submit a
dugga.  It uses the column that was created in #2616 for storing the
comment in the database

### DIFF
--- a/DuggaSys/courseed.js
+++ b/DuggaSys/courseed.js
@@ -231,6 +231,7 @@ function createVersion(){
 	var makeactive = $("#makeactive").is(':checked');
 	var coursevers = $("#course-coursevers").text();
 	var copycourse = $("#copyvers").val();
+	var deadlinecomment = $("#deadlinecomment").val();
 
 	if (versid=="" || versname=="") {
 		alert("Version Name and Version ID must be entered!");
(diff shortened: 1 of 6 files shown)


a15alica:
The code looks good and it works nicely on the page.

I just have one comment about how it works. Right now the comment is shown when you hand something in, no matter if the deadline has passed or not. (Code in Shared/dugga.js, from line 159) It would be better if the comment would only be shown when the deadline has passed.

a15tobli:
Hello! After the new commit this looks good, I have two minor feedback points.
The comment in the if-statement can be a little bit confusing, "//Check if deadline has past" is set on a selection that checks if the current time is actually less than the date for the deadline. However, the code works as it should, since the deadline message is set in the else-statement where the deadline is less than current date. A suggested edit could be that setting "Check if the deadline is active" as a comment to the first statement, and moving the current comment to the else-statement. That is not really something important, but could make the code a little bit easier to read.

The other feedback point is that the standard message when a comment is not set is that duggor 
will be reviewed in the next time the course runs, line 173 in Shared/dugga.js. Maybe this message can be set to something similar to "duggor blir rättade i mån av tid"? As far as I know, duggor are usually graded before the course runs again. Therefore it may perhaps not be appropriate as a standard comment when none is set.

a15tobli:
Changes looks good!

Issue #2706, opened 2017-04-04
Select Area In Editor Tool


a14davlo:
This is a subissue from an issue that is a little too big.

Make a function that can select an area with ER objects inside.

#2620 root issue

b15matvi:
Issue resolved, possible to select multiple objects at once.

PR #2708, opened 2017-04-04
#2706 fixed

c14joalu:
Fixed so you can target multiple targets, by checking if Topleft point and bottomright point is inside the dotted draw box


b15matvi:
Seems ok, merging

Issue #2637, opened 2017-04-03
Handle ordered lists in markdown.js


a15ramha:
markdown.js is currently not handling ordered lists, as well ordered lists. This should be implemented in function parseLineByLine(). Possible create new function for handling this issue and invoking it in parseLineByLine().

a15ramha:
This issue has been solved.

PR #2717, opened 2017-04-04
Added orded list for markdown.js for issue #2637

a15ramha:
Now the ordered list should be working according to the markdown commands.


a15ramha:
Latest commit breaks out code in function, as well as fixing indentation in code.

Issue #2694, opened 2017-04-04
Functionality for double clicking on an attribute/entity


b15matvi:
In order to solve the dropdown menu we need the functionality for double clicking on an object.

Subissue of #2684

HGustavs:
If there are any other suggestions for how to trigger the editing dialog, I am flexible.....

it has to be a gesture that is natural both on mobile and on computers....

e.g. right click is not suitable since it does not work well on mobile....

a14davlo:
I have designed a structure on the issues to make it cleaner, don't worry, this issue is still here, in fact I edited the issue this one relates to be this exact one.

PR #2724, opened 2017-04-04
Fix for issue #2694

b15matki:
Makes it possible for the user to double click


b15matvi:
Reviewed it, ready to merge

b15matvi:
Tested the code with the G4 branch, appears to have problems. Can't let go of an object.

PR #2725, opened 2017-04-04
#2706-merge-fixed-correctly

c14joalu:
fixed the error occured due to some wierd merging overwriting old pull, added solution to #2706.


b15matvi:
Reviewed, no testcounter visible.

Issue #2712, opened 2017-04-04
Able To Move Selected Objects


a14davlo:
When you've selected an area with objects inside you should be able to drag it around and move it.

#2620 root issue

b15matvi:
Issue resolved.

PR #2729, opened 2017-04-04
#2712 Fixed

c14joalu:
Made so that you can move multiple object at the  same time.


b15matvi:
Reviewed it and it looks good.

Issue #2722, opened 2017-04-04
Delete Selected Area In Editor Tool


a14davlo:
When you select an area in the diagram editor tool, you should have the option to remove all the objects inside the area. Basic stuff, we need this.

#2620 root issue

b15matvi:
Issue resolved.

PR #2738, opened 2017-04-05
#2722

c14joalu:
Fixed so that all targeted objects can be removed at the same time


b15matvi:
Looks good, should solve the issue.

PR #2745, opened 2017-04-05
made a universal deleteobject function

a15chrau:
The function is called upon from other functions that also delete objects.


b15matvi:
Reviewed it, unified delete functions.

PR #2749, opened 2017-04-05
2665 - On hold

b15matol:
This branch was renamed to 2665 from css-basic-classes

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -1798,14 +1798,6 @@ input.large-button {
  *                             Basic simple classes                             *
  * --------------================################================-------------- */
 
-.left {
-    float: left;
-}
-
-.right {
-    float: right;
-}
-
 .top {
     margin-top: 0px;
 }
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -1830,6 +1830,10 @@ input.large-button {
     text-align: justify;
 }
 
+.no-float {
+    float: none;
+}
+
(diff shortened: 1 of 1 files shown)


a15alica:
Syntax wise it looks good. I'm not sure the new classes would be that useful though since they only contain one row each. It seems like it's just as much work (but with fewer lines) just adding the css lines to specific elements when needed instead of having classes this general.

a15tobli:
Hello!
I like the suggestion, altough Alice makes a good point, this code has currently no good use. The idea to use classes in such manner in the css is a good idea, however, when the new classes contain only a line of code each this won't really help out much. If new elements would be assigned these classes, some lines of css would be saved. But the problem with this implementation is that alot of extra classes would be added in the html elements to be using these styling changes. 
The code would be harder to read when an element has assigned several classes with a single line of code, instead of simply adding those lines directly to the class or id. 

Implementing these classes would actually create extra work to when assigning classes to existing elements and would make the code a bit harder to read.

However, if more general CSS-classes would be added, with more styling that has a wider area of use, this would be a great idea to implement. My suggestion is to think about how these classes could be expanded so that they could be used more frequently and save work progress. Leaving the merge request open as of now, if you have a good idea how these classes could be used in an extended way which would make the styling in the code easier, I see no problem with merging this.

a15tobli:
Closing this request, as such classes will be implemented if we reach a point where we would need them and they have no use as of now.

Issue #2684, opened 2017-04-04
Double Click To Make Anything Able To Appear


a14davlo:
The task is to make any kind of message appear when you double-click any object in the diagram editor tool.

Once we're able to do that, we'll combine that with a dropdown menu which is a separate sub issue.

#2576 root issue

HGustavs:
Not necessarily a dropdown, I prefer dialog since dialogs work better on a wide variety of screen sizes.

HGustavs:
![screenshot 2017-04-04 10 36 13](https://cloud.githubusercontent.com/assets/6978801/24658498/2a82b5c4-194a-11e7-8b64-42427e5f2c75.png)

a14davlo:
Henrik G. comment on this matter.
"If there are any other suggestions for how to trigger the editing dialog, I am flexible.....

it has to be a gesture that is natural both on mobile and on computers....

e.g. right click is not suitable since it does not work well on mobile...."

a14davlo:
Matthias Kinnander, I edited this sub issue to be relevant to what you're working with right now, the "double-click an object" interaction.

b15matki:
Fixed a function that enables the user to double click in order to bring up a dialog window. At this moment there is only a simple window that pops up with no functions within it.

a14davlo:
Issue resolved

PR #2750, opened 2017-04-05
Fixed for issue #2684

b15matki:
Enables the user to double click on the canvas in order to bring up a window.


b15matki:
Forgot to include the function at the canvas element.

Issue #2607, opened 2017-04-03
Lines in diagram over attribute


a15admsa:
A line in the diagram is always over an attribute. It would be nice if the line is under the attribute, so that you can read the text in an attribute.

HGustavs:
![screenshot 2017-04-03 10 53 49](https://cloud.githubusercontent.com/assets/6978801/24602160/e3f05a56-185b-11e7-9633-6520a2b41af9.png)

I agree.

The drawing of the diagram needs to be done in two steps, first connectors and then the symbols. Each step is executed for each symbol.

#1 For each symbol - Draw lines connecting symbols
#2 For each symbol - Draw symbols themselves on top of those lines (remove code that draws the connectors from this iteration)

This is MUCH easier than figuring out where the connector line ends on the oval symbol.

b15matvi:
Issue resolved, closing.

PR #2755, opened 2017-04-05
Fixed-#2607

c14joalu:
Fixed so lines are rendered before other objects, so they are always behind


a14davlo:
All of it looks fine to me.

PR #2762, opened 2017-04-05
4: implemented that a dialogwindow shows up when a enitity is created

a14pethu:
also commentet on the variable ghostingcrosses, on what it does.


b15matgu:
Looks good.

Issue #2754, opened 2017-04-05
sublist bug on ordered lists in markdown.js


a15ramha:
The sublists are not organized.

a15elibr:
It would be nice with a description to more clearly understand what the issue is.

a15emibo:
Issue has been resolved

PR #2768, opened 2017-04-05
The bugg for ordered lists is fixed in markdown.js for issue #2754

a15ramha:
This bug is now solved.


a15elibr:
Would be nice if you had described how you solved it!

PR #2770, opened 2017-04-05
Fixed submission of dugga.

a15filba:
The submission of a dugga is now sensetive to the time of the day and
not only by the date.


a15timbe:
Yes, the fix seems correct. The styling of the code also seems to be up to standard. I don't see any conflicts, good job!

PR #2776, opened 2017-04-05
W14 fix for issue#2605

b15sarha:
added delete button for dugga in the test view, also closed a head tag in showDugga.php


b15sarha:
pair programming with @a15oscwe

PR #2777, opened 2017-04-05
Added textfields in dialog to change name och attributes and entities.

b15matgu:
Added textfields in dialog to change name och attributes and entities.


a14pethu:
Seems fine.

PR #2780, opened 2017-04-05
Fixed arrow icons in the JQuery datepicker

b15matol:
Added two complementary CSS classes that adds icons.


b15matol:
Result: 
skarmavbild 2017-04-05 kl 13 49 55

PR #2787, opened 2017-04-05
Error message fixed in codeviewer.

b15linek:
Error message is fixed but the switching of templates does not work. This needs to be fixed in a new issue.


b15linek:
Functionality of this should be fixed, see new issue #2789

Issue #2773, opened 2017-04-05
Implement changes to the JSON editor


a15tobli:
Related to #2547. New suggestions to change how the JSON parsing is done. The table in #2548 has been made, which makes it very redundant in the new changes and should be removed. Add new styling to make the changed content better aligned.

PR #2790, opened 2017-04-05
#2773

a15alica:
Fixed styling of the instruction and submission forms, deleted the table
that was used before and added a delete-button for the submission rows.


a15tobli:
Will create a new issue to handle responsiveness of this form.

Issue #2791, opened 2017-04-05
Double-Click Menu Options "Ok" And "Cancel"


a14davlo:
For the small menu to feel complete we'll need options where you're able to confirm the edits you've selected and the text you've written.

We'll need an OK button and a CANCEL button. They're self-explanatory.

b15matki:
Added the buttons and fixed the buttons functionalities. Also made the window adapt to the object the user selected.

b15matvi:
Issue resolved.

PR #2793, opened 2017-04-05
Fixed issue #2791

b15matki:
Fixed "Ok" and "Cancel" buttons and their functionalities. Also fixed that the window adapts to which entity the user selects.


b15matvi:
Should be working as intended

Issue #2707, opened 2017-04-04
When closing a figure, fill the figure with colour


b15matvi:
We want to be able to fill the area from the figure with colour. 
Try looking at the example code for the A figure for help how to solve it.

b15matvi:
Issue resolved.

PR #2795, opened 2017-04-05
Fix for issue #2707

a15oskst:
Figures now get filled. When created.

Also fixed som bugs when drawing the lines of figures.


b15matgu:
Looks good.

Issue #2775, opened 2017-04-05
Add sorting to file lists in teachers view


a15elibr:
Implement a way of sorting the files in the teachers view, by name, date etc. Currently however, the files does not have any date or size displayed to them. Therefor, ideally, issue #2525 and #2524 should be solved first, since they will implement date & file size.

a15elibr:
Some sorting code is already available in marking view, this code can be re-used.

a15freku:
The problem is solved.

Issue #2445, opened 2017-01-10
Marking view sorting functionality in all tables


HGustavs:
There are tables all over in the interface.

I think that the interface could benefit from using the sorting functionality from the marking view.

For example, the file lists, being able to filter, or to sort on date/user name etc would be a great feature for courses with hundreds of files.

HGustavs:
Implementation details suggestion in  #2545

PR #2797, opened 2017-04-05
Added increment for sorting in fileed.js

a15wilhe:
I reused the basic code from resulted.js for later on sorting functionality in fileed. Since the sorting functionality is not done i choose to not print out the simple icons. Later on the function setupSorting will be excuted onload on body.

In resulted.js the function setup is in start of the file (since that is the correct code standard) but here i selected to have it at the end. I also then choose to name it setupSorting instead. Will it be a problem? I was afraid of collisions with other groups if i put it at the start of file.


d15johro:
Issue not solved but an increment of issues #2775 and #2445

PR #2798, opened 2017-04-05
Changed value of buttons

a15wilhe:
In the fileed.php the value of the input buttons are correct but in the returned function in fileed.js when rewriting the table after AJAX call the value of the buttons were wrong. Simply changed the value from 'add link' to 'add file'.


a15wilhe:
Something went wrong, The first commit "added increment for sorting in fileed" should not be in this pull request. I will close this one.

Issue #2736, opened 2017-04-05
Remove connected lines when removing the connected object


b15matvi:
In the hardcoded example, if you remove the "person" entity all the lines will be out of place. If the base is removed, the connected lines should also be removed.

a14davlo:
All done! Well done Joakim

PR #2806, opened 2017-04-06
Small-bug-fix-#2736

c14joalu:
By mistake a earlier merge from local branches removed 4 lines of code, that removed the "crosses" when deleted object, added them again


a14davlo:
Minor fix, see no problem with it.

PR #2816, opened 2017-04-06
Fixed so the JSON generator is working

b15matol:
Also added icons for the add / remove buttons and made sure that if
there is only one submission row, that row cannot be deleted.


a15nikho:
Now everything is fine and the pull request is ready for merge.

Issue #2737, opened 2017-04-05
Move around in a big diagram


a14davlo:
To make it possible to create big diagrams, there has to be a possible way to move around in the Canvas-element. Make it easier to navigate.

#2472 root issue

b15matvi:
Issue resolved.

PR #2821, opened 2017-04-06
Part-fix for issue #2737

a15admsa:
You are now able to move around in the diagram. The problem is that there is an offsett that will be solved in the next fix.


b15matvi:
Merged in, but you'll have to fix the rest.

PR #2825, opened 2017-04-06
Issue#2619

a15chrau:
Can connect items with lines where connector- top, left, right and bottom are defined. Objects where symbolkind == 2 also work, as they are connected to their center.


b15matvi:
Looks good

Issue #2744, opened 2017-04-05
When creating a new attribute open a dialog window


b15matvi:
Open a dialog window when creating a new attribute. This window should only have functions connected to the attribute (no entity options)

b15matvi:
Issue resolved

PR #2846, opened 2017-04-07
Fix for issue #2744

a15gunfu:
Implemented a dialog window that pops up when new attribute is created.


b15matvi:
Reviewed, looks good

PR #2853, opened 2017-04-07
fix-for-issue#2835

b15matki:
Remove checkboxes and the cancel button due to it being redundant.
Added CSS to better structure the window and make it convenient for the user.


b15matvi:
Looks good

Issue #2837, opened 2017-04-07
No actions should be allowed when entity/attribute menu is up


b15matvi:
When opening up the menu to change entity/attribute appearances no other actions should be allowed in the background. If you click outside the menu, the menu should close.

b15matvi:
Issue resolved.

PR #2857, opened 2017-04-07
Revert "Solved issue #2837."

b15matvi:
Reverts HGustavs/LenaSYS#2856


b15matvi:
Was a bug, reverted the implement

PR #2858, opened 2017-04-07
fix-for-issue#2835 v.2

b15matki:
Fixed problem


b15matki:
Accidentally included old code that is incorrect. Closing pull request.

Issue #2839, opened 2017-04-07
Implement functionality to be able to delete selected object(s) by using the delete key


b15matvi:
It should be possible to delete selected objects by using the delete key on the keyboard aswell as the delete button in the GUI

b15matvi:
Issue resolved.

PR #2879, opened 2017-04-07
can delete object with delete key

a14pethu:
Solved #2839


b15matvi:
Reviewed, should work.

Issue #2874, opened 2017-04-07
Update init_db.sql to include new usergroup and mapping user_usergroup tables


b15matol:

b15matol:
This is part of #2441.

a15tobli:
Merged

PR #2880, opened 2017-04-07
Changed the database to include two new tables, usergroup and user_usergroup

b15matol:
This will later be used to be able to add students to groups for
assigning with duggas and such. Not entirely sure about the constraits
in user_usergroup though. Subject for review.


b15matol:
This refers to issue #2874.

Issue #2525, opened 2017-03-30
Display the date of a file when it's been uploaded in teacher file uploads


a15elibr:
Display date of file after it's been uploaded. This is a smaller part of issue #1835.

HGustavs:
We mean "when it was" uploaded rather than "when it's been" uploaded ?

Otherwise I dont quite understand what we are trying to achieve.

a15emibo:
It seems to be working already or am I misunderstanding the issue?
![what](https://cloud.githubusercontent.com/assets/26706727/24505108/ffa13abe-1558-11e7-8fb3-de469e804656.png)

HGustavs:
Other direction... not student file uploads but teacher file uploads.

![screenshot 2017-03-30 21 45 24](https://cloud.githubusercontent.com/assets/6978801/24523233/a4ae67a0-1592-11e7-869e-73dc46ada87a.png)

We see the files sorted in three groups, perhaps a joint table with all groups, and sorting on type of upload (e.g. global), date of upload, file types (e.g. pdf) etc

HGustavs:
Implementation details in  #2545

a15elibr:
A column for saving date in the database has been made & merged, with pull request #2883.

a15wilhe:
@a15elibr This issue is fixed now after the 3 merges above.

PR #2889, opened 2017-04-07
Fileedservice now fetches uploaddate from database

a15oscwe:
Filledservice now requests uploaddate from database so that it can be viewed.


a15oscwe:
Partfix for #2525

Issue #2845, opened 2017-04-07
Free draw figure starts filling colour before it is closed


b15matvi:
When using the free draw figure tool it starts to fill the colour before the figure is closed.

b15matvi:
Issue resolved.

PR #2900, opened 2017-04-08
fix-for-issue-#2845

a15oskst:
Figures now renders correctly.
Did some clean up of comments and variables related to this function.


b15matvi:
Looks good.

Issue #2901, opened 2017-04-08
Add select menu to choose between figure forms


a15oskst:
Add a select (drop-down/dialog) menu to be able to choose between different forms of figures to draw.

For example:
 - Square
 - Circle
 - Free-draw (Placing Points)

Root issue: #2844

a14davlo:
Resolved

PR #2902, opened 2017-04-08
fix-for-issue-#2901

a15oskst:
Added the select drop-down menu for creating figures. Also added example values that hasn't been implemented yet.
Separated the function createFigure() to be able to add forms easier.


b15matvi:
Reviewed, looks good.

PR #2908, opened 2017-04-10
Initially draws the diagram

a15chrau:
A one line fix, diagram.php loads the updategfx function at body onload


b15matvi:
Looks good

Issue #2735, opened 2017-04-05
Zoom in and out in a diagram


a15admsa:
Making it possible to zoom in and out with buttons. These buttons should zoom in and out in the Canvas-element, not the whole site. 

Root issue is #2472

b15matvi:
We want the zoom in button as a toggle, when you click it you get a magnifiyng glass effect similar to that of paints zoom feature. You will continue to zoom when you click until you untoggle the button.

a15admsa:
Issue is now fixed!

PR #2911, opened 2017-04-10
Fix for issue #2735

a15admsa:
This pull request solves the problem in i ssue #2735. You are now able to move around in the diagram. 

There are some bugs right now, but these will be solved when the zoom-function is implemented.


a14pethu:
Seems to be okey.

Issue #2524, opened 2017-03-30
Display the size of a file after it's been uploaded in the teachers view


a15elibr:
After uploading a file, the size of the file should be displayed.

This issue is connected to #1835

HGustavs:
We mean "after" uploading file, right, not "when" ??

HGustavs:
Is size represented in the database? In that case a size column needs to be added, it is the best place to do this when we upload the file...

a15herny:
Where exactally is this being implemented? Is it the diary functionality?

a15elibr:
This is in the teachers view. Updated title and description for more clarity.

PR #2916, opened 2017-04-10
filereceive.php now inserts filesize into database

a15oscwe:
Filereceive.php now inserts filesize into database when a new file is added.
Partfix for issue #2524


a15ramha:
The file size is now is getting added to the data base.

Issue #2898, opened 2017-04-07
Overflow in submissions div in dugga editor


a15alica:
When you add multiple submissions the div expands. If enough submissions are added the div gets too long and does no longer fit in the box. A overflow scroll should be implemented to fix this problem.

HGustavs:
Has this been solved?

a15tobli:
This has been solved in pull request #2935, forgot to close the issue.

PR #2935, opened 2017-04-10
#2898 - fixed scrollbar

a15alica:
Added so that a scrollbar will appear if needed when adding many submissions in the Edit variant box in the dugga editor.


a15timbe:
Not much to say since only a line of code was added, but this seems to fix the issue so good job. Your code is correct whilst keeping to the standard.

PR #2938, opened 2017-04-10
Table in file view for teachers listes file size

a15wilhe:
Filesize from database is now fetched in fileedservice. Function formatBytes converts the Bytes to ex. The table in file view prints out in human readable format by that function.


a15wilhe:
this fix relates to issue #2524

Issue #2918, opened 2017-04-10
Fix deadline handler for duggor


a15tobli:
This code was broken after the weekly merge. Fix the deadline handling so it works properly again. Related to #2420.

a15tobli:
Solved in #2947, closing issue.

PR #2947, opened 2017-04-10
Fixed deadline handler for duggor.

a15filba:
The warning for the dugga view and the dugga submittion is fixed and works as it should. From issue #2918.


a15tobli:
Merging pull request and I will create a new issue for merging line 754 and 755 together.

PR #2950, opened 2017-04-10
Fixed more specific error messages when checking username in forgot password

a15ponja:


e14johho:
Closing the pull request after talking with a15ponja, this pull request creates more issues then it solves because something has gone wrong with the commits.

Issue #2638, opened 2017-04-03
Add form to change password on user page


a15tobli:
Extend the user setting page with a form that allows change of password for the user.

HGustavs:
Good idea.

HGustavs:
Is there a way to only allow this on https?

c14caran:
Make sure to limit new password lengths to 72 characters. Give an error if the new password is longer than that. This limit is from bcrypt, the password hashing function that the system might change to by implementing issue #1964 . As 72 characters is already very big it should be fine to limit to this even before that issue is fixed and implemented.

a15tobli:
The forms and max password length have been added in #2958 and #3070. Functionality to change password needs to be added to the form.

PR #2958, opened 2017-04-10
Changed styling of profile page

a15alica:
The styling of the profile page has been changed. A field has been added where you can enter an answer to the challenge question. A field for entering a new password again has also been added.

To be able to change either the challenge question or the password a field has been added where you have to enter the current password.

This commit only contain styling/design fixes so no functionality has been added. (Partially related to issue #2638)


a15timbe:
Looks good. Everything is nicely structured now and the code doesn't seem to break anything previously implemented.

PR #2964, opened 2017-04-11
Added Scale Function

a14pethu:


a14davlo:
The code seems to be fine now, well done a14pethu!

Issue #2836, opened 2017-04-07
Delete all has stopped working


b15matvi:
The delete all button has stopped working, but should be a quick fix.

a14davlo:
Issue resolved

PR #2965, opened 2017-04-11
Fix for issue #2836

a15chrau:
Fixes the delete all function


a15chrau:
Divides the delete function into several sub-functions, in case they need to be used at different locations, Move points as well as delete lines, also defines a Symbol erase as well as a Figure erase. Adds a function to diagram that returns all the lines symbols.

The fact that we don't delete points can be a problem in the long run and should definitely be looked at, at some point.

b15matvi:
Looks good!

Issue #2741, opened 2017-04-05
Edit Color On Text In Objects


a14davlo:
In the text editor we'll need the option to change the color used. Basic color options like green, blue, yellow, red, black and grey. 

#2586 root issue

b15matvi:
Issue resolved.

PR #2971, opened 2017-04-11
Fix for issue #2741

a15perbl:
Added code to make it possible for users to change color of the text inside an object.


a14davlo:
The code seems to be working, well done a15perbl!

Issue #2970, opened 2017-04-11
Delete object and delete all no longer uses updateGFX()


b15matvi:
Simple fix, they need to call on updateGFX when clicked on.

a15chrau:
ples

PR #2976, opened 2017-04-11
Fixe for issue #2970

a15chrau:
Fixes issue #2970 and also makes eraseSelectedObject run the eraseObject function instead of directly calling upon object.erase(), resulting in consistency when running the different delete functions.


a14davlo:
Seems to work, well done a15chrau

PR #2978, opened 2017-04-11
Fixes so that the forgot password box gives more desciptive errors

a15ponja:
Since we should not do anything special if a user does not have a security question it just returns an error.


a15ponja:
Now the problem should be fixed.

PR #2985, opened 2017-04-11
Added Primary attribute function

a14pethu:
Added a function that let you mark attributes as primary keys.


b15matvi:
Looks good!

Issue #2598, opened 2017-04-03
Edit Font On Text In Objects


b15matvi:
In the text editor we'll need the option to change the font used. A few distinct options like arial, calibri, comic sans and etc.

#2586 root issue

a15chrau:
For a user to be able to change the font inside entities whenever they please, or for us to change the one that is currently used as standard?

b15matvi:
Issue resolved.

PR #2993, opened 2017-04-11
Edit font of text in objects (#2598)

a15perbl:
Added code to make it possible for users to change fonts on text inside objects.


b15matvi:
Looks good!

Issue #2912, opened 2017-04-10
Diagram Movement Using Spacebar


a14davlo:
To move the diagrams around it would be better to make use of the spacebar button.

HGustavs:
Does not sound very mobile friendly.... also supporting space-bar would be a better way of doing this?

a14davlo:
Hmm, you're right about mobiles, didn't think about that. 

Is there any spacebar alternative for mobiles that suites our issue?

a15admsa:
My idea is that we have one button. When the button is clicked you'll be able to move around in the diagram. The button changes color when pressed in and not. 

With the use of shadowing, this implement will be pretty straight forward. What do you think @HGustavs?

HGustavs:
A button first then we can add other types of gestures later on.

I would really like to see updates of the diagram while moving... I think it is just callig redrawgfx every few mouse moves.

a15admsa:
Issue has been solved with a single button. The button makes it possible to move around in the diagram. You can start the possibility to move around and then close it. The button changes color to indicate that the option is chosen. This issue is now solved and it can be closed.

PR #2998, opened 2017-04-11
Fix for issue #2912

a15admsa:
Removed a button and it's now possible to use one button to open and close the possibility to move around.


b15matvi:
Looks good!

PR #2999, opened 2017-04-11
Added some testdata

b15linek:
Added 15 students and assigned these to courses.
Added 5 teachers and assigned these to courses.
Added 'datorns grunder' as course (empty) for testing.


a15freed:
it's working

PR #3002, opened 2017-04-11
Added a magic header solution

c14emian:
Added position:fixed on relative elements, it's working but still needs
styling to look decent.


a15freed:
I downloaded 2461 branch from G3 but I couldn't se the changes you showed on your computer. what I get is this, still with two "#"  :)

image

PR #3006, opened 2017-04-11
Fixed so the back-button is back in grouped.php

b15matol:


a15timbe:
Yes, this looks good. Seems to keep the code standard. Since not much was changed there isn't much that can be broken. Good job!

PR #3009, opened 2017-04-11
Dropdown menu for font styles

a15perbl:
Users can now only choose the style of the text inside an object from a list of predefined values.


a14davlo:
Minor fix, seems to work, well done a15perbl!

Issue #2920, opened 2017-04-10
Diagram Engine Functionality - Json Text Document


a14davlo:
The diagram engine should work mostly like the engine from google docs.

Each instance is represented as a json text document.

#2454 root issue

c14joalu:
@HGustavs , I tried some on this issue, but i keep getting the problem that if you use JSON.stringify, on for example diagram to retain the state, it only gets a shallow version i.e. it does not reference it functions anymore. wouldn't this issue be alot easier to solve if you use a deep clone version of the arrays, and then convert them to string and be able to make a file of that string?

c14joalu:
I found a way to make it possible to go from a stringify JSON to the array put in before, making it possible to save and load states from JSON text files, disregard my previous comment.

a14davlo:
Well done!

Issue #2841, opened 2017-04-07
Load/Save functionality for diagrams


b15matvi:
We would like to be able to Save and Load different diagrams as textfiles that you store locally on the computer.

PR #3010, opened 2017-04-11
#2920 fixed save states, load state, export files , import files and get canvas as a image

c14joalu:
Created a drop down box for selecting :
To get a .png of the canvas.
To save the state of the canvas.
To load from the previous saved state.
Created button for exporting diagram, diagram.constuctor.name, and points. 
Created button for importing diagram,diagram_names,points so that you can import a canvas.

Created functions for everything, still needs to remove some comments and make it more convenient to handle, since you have to press export x3 to get the 3 files and suchs, and you can only save state once every instance or it will stringify to much


c14joalu:
This should not be merged yet, since i need to talk to @HGustavs , about how he wanted JSON text for the instance, and with the group leader about #2841, since this will basically solve both issues at once, hopefully.

PR #3011, opened 2017-04-12
Dropdown list for text color

a15perbl:
Users can now only choose from a predefined list of colors when changing the text inside an object.


a15oskst:
Ok

Issue #3014, opened 2017-04-12
Add the ability to change font and color in the attributes


b15matvi:
Add the functionality existing in the entity to the attributes aswell.

b15matvi:
Issue resolved.

PR #3017, opened 2017-04-12
(#3014) Added functionality to attribute dialog window

a15perbl:
Attribute dialog window now has the same options as the dialog window for entities.


c14joalu:
Everything looks good!

Issue #2931, opened 2017-04-10
When adding a Variant to a dugga, no feedback to the user. 


b15sarha:
When pressing the plus sign to add a new Variant to a dugga, it does not toggles out. Upon adding new variants, they should toggle out, so that the user can sees them.

HGustavs:
This is a priority issue as it makes the dugga editor quite hard to use...

a15tobli:
Group 2 have been working with this page, there is also a problem with not being able to close those added rows. 
When the + icon is clicked, a new variant is added, but the toggle makes them hidden.
When anywhere else in the row is clicked, all added variants will be toggled visible, but they cannot be toggled back to hidden.
The + icon should always toggle to make the new row visible, and clicking anywhere else should toggle between the two options depending if variants currently are visible or not.

a15tobli:
Issues is merged in #3020. The dropdown is now working as it should with the added dropdown icon for toggling instead of the whole tr. The + icon now always toggles out the added variants. Minor fixes are to be implemented with the appearance of the icons, but those fixes does not affect how the dropdown works.

PR #3020, opened 2017-04-12
#2931 Added functionality to show/hide variants.

a15osmla:
The user should now be able to show/hide variants when a dugga is clicked on in the test page.  New variants and deleted variants should now  also cause the the corresponding dugga to give feedback to the user by showing that a variant is added or removed in the table.


a15tobli:
Code works as it should and the added borders made it easy to distinguish the dugga rows from each other.

PR #3021, opened 2017-04-12
'Verison' -> 'Version'

a15eriho:
Where appliable


a15freed:
The spelling seems ok now

PR #3022, opened 2017-04-12
Added code to check if user is a superuser or teacher

a15ponja:
However the error message should be changed before the merge into master, the error message now is to descriptive.


a15ponja:
I think that all the messages should be the same in the end so that anyone trying to figure out information on how the check works does not get any useful information. I am thinking that all error messages should be Username not found.

Issue #1781, opened 2015-07-16
Markdown tables would be super nice


HGustavs:
Using pipe characters. 

Colons can be used to align columns.

| Tables | Are | Cool |
| --- | :-: | --: |
| col 3 is | right-aligned | $1600 |
| col 2 is | centered | $12 |
| zebra stripes | are neat | $1 |

The outer pipes (|) are optional, and you don't need to make the raw Markdown line up prettily. You can also use inline Markdown.

| Markdown | Less | Pretty |
| --- | --- | --- |
| _Still_ | `renders` | **nicely** |
| 1 | 2 | 3 |

a15ramha:
This issue has been solved.

PR #3025, opened 2017-04-12
Week15 issue#1781 d15johro a15ramha

d15johro:
this pull-request partially solves the markup for tables in markdown.js. tables with tables rows and table data are parsed curroectly, however, there is no support for alignment or table-headers, etc. This branch and Issue #1781 should therefore not be closed.


a15wilhe:
It would be nice with a border on the table element.

a15wilhe:
Not closeing issue and branch as on request.

PR #3030, opened 2017-04-12
Droplist for selecting color on test-row

a15mikla:
A droplist for selecting backgroundcolor when creating or updating a
dugga (test) is added. The values are generated from the JS-file.
Part of a solution for issue #1937


a15freed:
Looking good

PR #3037, opened 2017-04-12
Stop light logic is not correct for moment that has duggas

a15andau:
Stop lights now displays correct for duggas. If grade is pass - green, if fail - red. If not marked - yellow.


a15freed:
Looks ok, nice colors

PR #3039, opened 2017-04-12
Now displaying the security question for the specific user

a15timbe:


a15tobli:
Hello! The code looks fine, creating a new issue for generating different error messages when no security question is set.

PR #3043, opened 2017-04-13
Added more testdata

b15linek:
2 more courses added with students and techers.
1 example dugga added to one course. The dugga gives error but this will be solver later. At least it shows up.


a15freed:
looking good!

Issue #3051, opened 2017-04-13
Remove "Console.log" from alla diagram.*-files


a14davlo:
There is a lot of console.log in the files overall, which makes it hard for the developers to see which one is their own. It would be nice if we can remove all the console.log-lines so that it becomes easier for the developers right now.

b15matvi:
Issue resolved.

PR #3057, opened 2017-04-13
Removed all Console.log - Fix for issue #3051

a15admsa:
I have now removed all unnecessary loc where we have had console.log() without any purpuses. This fixes issue #3051.


b15matvi:
Looks good!

Issue #3065, opened 2017-04-18
Zoom value and coordinate value for the diagram in top left and bottom corners


b15matvi:
We want to be able to see how much we have zoomed in and also on what coordinates we are on.

HGustavs:
I enclosed a picture and code for a "ruler" in some other issue.

a15perbl:
Issue closed.

PR #3067, opened 2017-04-18
Fix for issue #3065

a15admsa:
This commit solves issue #3065. Under the canvas you can now see the values of zoom and X + Y coordinates.


a15chrau:
Looks good

PR #3068, opened 2017-04-18
W16 issue 1888 a15emibo

a15emibo:
Probably fixed tokenizer not being able to recognize for example #34edaf (two numbers and an 'e') as a colour in certain situations.


a15elibr:
"Probably"?

Issue #3015, opened 2017-04-12
Normalize the buttons in attribute and entities


b15matvi:
The layout of the buttons in attributes and entities should be normalized. Should be only one OK button and no other redundant buttons. All changes should be commited when you click OK.

b15matvi:
Issue resolved.

PR #3083, opened 2017-04-18
Fix for issue #3015

a15gunfu:
Removed all redundant buttons on create attribute and create entity, now there is only a OK-button to accept all choices that the user has made.


a14pethu:
Looks good.

PR #3089, opened 2017-04-18
The tables are now working correctly according to the markdown cheat sheet

a15ramha:
This syntax should work properly now.
| Tables        | Are           | Cool  |
| ------------- |:-------------:| -----:|
| col 3 is      | right-aligned | $1600 |
| col 2 is      | centered      |   $12 |
| zebra stripes | are neat      |    $1 |


a15wilhe:
@a15elibr Pull request is good. But there needs to be a css class for adding the border to the table. Please add a issue for that. After that issue 1781 is fixed.

Issue #3027, opened 2017-04-12
Properly display dropdown icons in dugga editor


a15tobli:
The dropdown icons in dugga editor needs some further improvements for a better user experience. There are 2 things that should be fixed:
- Rotate the icon 90 degrees when the variant dropdown is closed.
- Only display the dropdown icon on duggor that have a list of variants added

At the moment, a dropdown icon is added for every dugga. Remove the icon on duggor who does not have any variants in a dropdown view.

a15tobli:
Solved in #3104, closing issue.

PR #3104, opened 2017-04-19
#3027 Properly display dropdown icon in the dugga editor

a15osmla:
Dropdown icon should now only be displayed on the rows that have a variant to show in the dugga editor. Icon now also changes when clicked upon to give feedback.


a15tobli:
The fixes for the dugga editor looks really nice and works good as well.

Issue #3062, opened 2017-04-18
Selecting color on row for duggas does nothing


a15mikla:
The option where you can choose standard- or header-color for duggas doesn't do anything. It needs to be fixed in order to solve issue #1937

Image to clarify which option:
<img width="394" alt="colorrow" src="https://cloud.githubusercontent.com/assets/26705198/25118479/daecd3aa-2416-11e7-8fc8-c041b2f50167.png">

PR #3105, opened 2017-04-19
Duggas as headers

a15mikla:
Duggas now have the option to either be styled as a dugga or as a
header. When adding or updating a dugga you can select which kind of
styling in a droplist. Solution for issue #3062 and #1937


d15eriso:
Looks fine, good with consistent names.

PR #3110, opened 2017-04-19
Fixed save/load and export/import #2920 , #2841

c14joalu:
fixed save/load and export/import and that the entire "canvas" is represented in one text file


a14davlo:
All seems alright now

Issue #3073, opened 2017-04-18
Arrow icons still not where they are supposed to be


a15elibr:
References issue #2973. Look there for more specific information. This issue was supposed to be solved and be working after the latest merge, but it isn't, so it needs to be fixed again.

a15elibr:
Either put icon on the right side of the text or on the left, so that they all have the same

HGustavs:
Screenshot?

a15elibr:
![18009621_10210752636824876_814890710_n](https://cloud.githubusercontent.com/assets/26704655/25123474/bb94f6b0-2428-11e7-8f64-969a3ee7a7c1.png)

a15elibr:
Closing issue since it has been resolved.

PR #3113, opened 2017-04-19
Changed place of arrow icon in fileed

a15wilhe:
In most tables the place for the arrow icon has been changed and it is now consistent. You can check it out at my folder on the testsite.


a15wilhe:
If ok this solves issue #3073

PR #3115, opened 2017-04-19
Fix-for-issue#3101

b15matki:
Now the user can select the derived type on attributes. But still alot of other errors remains.


b15matvi:
Looks good...

PR #3119, opened 2017-04-19
Swimlane beta 1.0

b15linek:
Only working for webbutveckling - programmering. Testdata needs to be standardized and courses needs a start date. Right now swimlane always start at week 1 of the year. Should start at week 1 of course but this doesn't exist.


a15freed:
ok merging

PR #3135, opened 2017-04-19
2419

a15jorpa:
Simple custom 403 error page with some php. Based on an existing php file from the system.


a15freed:
ok merging

Issue #3140, opened 2017-04-20
Most of the attribute/enitity options are broken


b15matvi:
- Name is not entered when you open an enitity/attribute

- Primary key does not give the underline as it should

- Multivalue does not give the border around

- Changing text scale does not work

- Weak option does not give the border around the entity

b15matki:
Fixed the following bugs as of right now:

- Primary key does not give the underline as it should.

- Multivalue does not give the border around.

- Changing text scale does not work.

-Weak option does not give the border around the entity.

b15matvi:
Issue is reoslved

PR #3145, opened 2017-04-20
Partial fix for issue #3140

b15matki:
Fixed the following issues:

-Primary key does not give the underline as it should

-Multivalue does not give the border around

-Changing text scale does not work

-Weak option does not give the border around the entity


a15oskst:
Multivalued doesn't work.

When setting a attribute to become a derived attribute, nothing happens. But when you deselect the derived attribute, the attribute becomes dotted as well as all figures and the background grid.

Primary key seems to be working.
Text size seems to be working.
Weak Seams to be working.

a14davlo:
There's no technical errors but these code changes does not resolve all problems in issue #3140.

PR #3150, opened 2017-04-20
Added ENTER-button functionality

a15freku:
Added so that the enter button on the keyboard works for search input
field


a15freku:
Also removed some old console.logs from another fix

a15elibr:
@d15johro

PR #3161, opened 2017-04-20
Partial fix for issue #3140 v.2

b15matki:
Fixed the following issues:

-Primary key does not give the underline as it should

-Multivalue does not give the border around

-Changing text scale does not work

-Weak option does not give the border around the entity


b15matvi:
This is a partial fix for the issue, a new issue will be created for the derived attribute option bug.

Issue #3118, opened 2017-04-19
Shadows on headers


a15mikla:
The shadows from "moment's" in sectioned look wierd. How are they supposed to look? I'm thinking that the shadow should appear on the whole row, instead of how it looks now. How do you want it @HGustavs ?
Part of issue #1937

Image to clarify how it looks now, the shadow is only visible on date and cogwheel:
<img width="173" alt="issueorsomething" src="https://cloud.githubusercontent.com/assets/26705198/25172299/d35e5908-24f0-11e7-8090-1fdf9740f148.png">

a15mikla:
We made the decision to add a shadow on the whole row. @HGustavs

PR #3167, opened 2017-04-20
Shadow on rows that are headers

a15mikla:
Rows that act like headers now have a shadow.
Issue #3118


b14marki:
Looks fine :)

a15freed:
ok meging

PR #3172, opened 2017-04-20
Added keyUp filtering to fileed

a15wilhe:
The "one" view table with all files and the search table now has filtering on keyUp. 

Function keyUpSearch is run onclick on button for switching table view. 

When user hits search the other keyUp functionality is runed. 

You can check out the functionality at my folder.


a15wilhe:
pull request had conflicts but is now resolved

Issue #2917, opened 2017-04-10
Diagram Engine Functionality - Semi-Unique Hash


a14davlo:
The diagram engine should work mostly like the engine from google docs

Each instance of a diagram has a unique hash. 

#2454 root issue

b15matvi:
Issue resolved.

PR #3174, opened 2017-04-20
#2917

a14pethu:
added hash function


a14davlo:
Seems like you've accomplished the task of adding hash, coming issues will build more upon this.

Issue #3095, opened 2017-04-18
Move inner html to diagram.php


b15matvi:
All unecessary inner html code, such as buttons etc should be moved to diagram.php

HGustavs:
Yes. I would consider buttons and dialogs inside the .js file a bug.

b15matvi:
Issue resolved.

PR #3181, opened 2017-04-21
Fix for issue #3095

a15admsa:
I've now moved everything out from innerHTML, with the exception of the canvas-element. Because we want the size of the canvas-element to be decided whenever the screen size is changed.


b15matgu:
Looks good.

Issue #3183, opened 2017-04-21
Double clicking sometimes highlights the application console text causing issues in the diagram


b15matvi:
Sometimes when you double click the application console text gets highlighted and when you try to draw something on the diagram or select something you get the problem with trying to move the application console text instead.

b15matvi:
Issue resolved.

PR #3184, opened 2017-04-21
Fix for issue #3183

a15admsa:
Fixes issue #3183 and some bug-fix that doesnt belong to this issue. Everything should work now.


b15matvi:
No new implementations, just moving around and fixing. Looks good!

Issue #3162, opened 2017-04-20
Creating a new attribute and applying derived to it will change the grid and figures aswell


b15matvi:
Creating a new attribute and applying derived to it will change the grid and figures aswell and this is something that must be dealt with.

a14davlo:
Nice fix b15matki

PR #3197, opened 2017-04-21
Fixed issue #3162

b15matki:
Added a line of code that prevents other elements to inherit the "derived" type.


a14davlo:
Nice fix b15matki

PR #3202, opened 2017-04-21
Add output if filter list is empty

### LINKED ISSUE #2440: If blank filter list we need some output
We need a "filter list is blank" output if filter list is blank, currently nothing happens....
which may confuse the user to thing that the code is broken.

a15matab:
This fixes #2440.

### DIFF
--- a/DuggaSys/resulted.js
+++ b/DuggaSys/resulted.js
@@ -461,7 +461,7 @@ function process()
 		if (clist){	
 				clist=clist.split("**"); 
 		} 
-		
+
 		// Create temporary list that complies with dropdown
 		momtmp=new Array;
 		var momname = "tore";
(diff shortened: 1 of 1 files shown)


a15wilhe:
It looks a bit wierd when it's outputed to the body. Should it not be printed to the filter dropdown with id dropdownc? It is the "popup" box on hover on the "tratt" icon.

a15matab:
Do you mean inside the actual dropdown, like the example below?

namnlos

a15wilhe:
hmm is this for when data has been filtered (and there is no data to output equal to that filtering) or when there is nothing to filter on (when the actual dropdown list is empty)

a15matab:
It would be after the data has been filtered. E.g if no checkboxes are selected in the dropdown and the user presses "Filter" the "Filter list is blank" text would be outputed. So instead of a completely blank page which could make the user think it's broken (I assume this is what the issue meant) it shows the text.

a15wilhe:
ok. then I will approve your commit.

a15wilhe:
close #2440

PR #3214, opened 2017-04-21
Changed which duggor are shown in grouped.php

a15nikho:
Made the question to the database easier to understand and removed things that are not needed. Also changed what is shown on the site, now only duggor with the "gruppduggor/gruppseminarie" mark are shown on the site.


a15tobli:
New pull request has been requested due to major changes in the code, closing.

PR #3219, opened 2017-04-21
Swimlane beta 1.1

b15linek:
New design and mouse-over effects on duggas. A red line added in swimlane to show what week you're on right now. (still only works with 'Webbutveckling - datorgrafik' because of no database standard).


a15freed:
Looks awesome

PR #3222, opened 2017-04-21
Grid and standard size

a15chrau:
Do not delete branch

Implements a grid with a minor visual bug related to camera position. Does not yet align object borders to the grid, but this will be implemented at a later stage. 

Gives some objects a standard size for when the user creates an object that has a size that is considered "too small".


b15matvi:
Tested and looks good!

Issue #3045, opened 2017-04-13
Generate error messages on profile page when no security question can be retrieved


a15tobli:
Related to #2824. When a security question is not set, generate feedback messages when the question is being fetched on the profile page and it cannot retrieve it. Messages could contain information be that no question is set or the user is a teacher account and should not have one. There should be no security problems with messages stating that the user has a teacher account here, as you would already be logged in to reach the profile page.

a15tobli:
Solved in #3237, closing issue.

PR #3237, opened 2017-04-24
Added error message to the profile page for security question

a15timbe:
If for example a security question is not found, a error message will be displayed and red color will highlight the error.


c14caran:
Fixes issue #3045

Issue #3126, opened 2017-04-19
table-bug in markdown.js


d15johro:
the first "| ----- | :----: | ---- | ---- |" should always come after the first table row. "thead" should have "th" instead of "td"

d15johro:
If you try to align the thead as follow: 

| ----- | :----: | ---- | ---- | <<--- table cannot start like this!!
| head0 | head1 | head2 | head3 |
| dsa1 | dsa1 | dsa1 | dsa1 |

every row will be nested in thead instead of tbody.

d15johro:
EDIT: the issue is not an issue. It should not be possible to start a table with the following syntax:
| ----- | :----: | ---- | ---- | 

However, the first "| ----- | :----: | ---- | ---- |" should always come after the first table row.

PR #3240, opened 2017-04-24
Week16 #3126 a15ramha d15johro

a15ramha:
The issue is solved and the tables should work with the new setup where the second line in the tables determine the alignment of the text in the rows.


a15oscwe:
Closes #3126

PR #3241, opened 2017-04-24
Fix mirrored sizing regarding relations.

b15matgu:
Fixed sizing regarding relations. It now mirrors the selected points location.


a14davlo:
Good fix

Issue #3245, opened 2017-04-24
Problem with resizing attributes and entities


b15matvi:
When we try to resize attributes or entities we can't see in real time how the object changes in size. Only after we let go we see the new size.

b15matvi:
Issue resoleved.

PR #3248, opened 2017-04-24
Fix live resize update. Issue #3245

b15matgu:
Fix issue regarding not getting live update when resizing attributes and entities.


b15matvi:
Issue resolved.

PR #3252, opened 2017-04-24
Removed loginprompt.php and added helper function to show the login box

### LINKED ISSUE #2446: loginprompt.php
Does loginprompt.php really do anything?

Content of file seems useful, but is it used?

Should we put that contents in other file?

c14caran:
Fixes #2446


c14caran:
As this pull request does not show any visible changes, manual testing is needed.

Open the file courseed.php and add the following rows, this will force show the login prompt even if logged in or not (remember to remove afterwards)

capture

Issue #3032, opened 2017-04-12
Functionality to move and delete Figure


b15matvi:
We can't delete a figure without deleting everything on the diagram, we also can't select and move a figure.

b15matvi:
Splitting this issue up, deleting part is done

PR #3267, opened 2017-04-24
Partial fix #3268 #3032 [Do NOT close the issue]!

### LINKED ISSUE #3268: Code Cleanup In Diagram.js
Double check so that the code is consistent, coherent and fulfills its purpose. See if you can find any errors, bugs or things missing.

a15oskst:
Did some clean up of some of the functions that might have something to do with this issue.
(Was going crasy over the code because of the bad formating)

Don't know if it will stay as this, but at least we can try to go after the code standards and keep the indentation?

While i did the clean up I found some missing semi-colons, missing var declaration on local variables and also some performance issues that i fixed.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -932,113 +932,100 @@ function updateActivePoint(){
     activePoint = null;
   }
 }
-function mousemoveevt(ev, t){
-		xPos = ev.clientX;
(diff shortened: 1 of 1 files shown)


a14davlo:
Alright, seems good. I'd like to speak with you about this @a15oskst.

PR #3272, opened 2017-04-25
adding styles to markdown table

d15johro:
wazzup?


a15elibr:
Wazzup is not appropriate as a description for pull requests. It shoud be "what is up".

Issue #3253, opened 2017-04-24
When in the File view, unable to download/open js. and css. files, change BROWSER


b15sarha:
When trying to open a file in the File view, we get the message" ctype undefined" this is because I'm using Safari browser. However this works fine on other browsers like firefox.

b15sarha:
This is already resolved.

a97marbr:
Previous fix has been reverted. A new fix is required.

PR #3273, opened 2017-04-25
Added forced download

a15freku:
Added forced download for files ending with .html .js .css

fix for issue #3253


a97marbr:
This fix breaks the ability to display html files. I revert to the old code and reopen issue #3253. Bug #3253 has to be resolved in some other way.

PR #3274, opened 2017-04-25
404-text updated

a15herny:
The 404-page now has an error message. The error message is displayed in the header as before but without "h1" tags.


a15elibr:
Could have a more explaining description, this pull request doesn't say what is fixed.

PR #3288, opened 2017-04-25
Filter option for teachers in resulted view.

a15eriho:
* Yellow background on teachers.
* Saves to local storage
* Updates when the filter menu is closed


a15freed:
ok merging

Issue #1878, opened 2016-02-23
Play link do not appear to be grayed out?


HGustavs:
The play link should only be visible if there is something to play.

Also, I think that for the play button the links should be from a file list dropdown, not a text field (like code examples and so on).

a97marbr:
I second this

A15cecal:
@HGustavs @a97marbr It seems like there's a fix for this in codeviewer.js, in the function returned(data), but this one is not used anywhere. Is this correct?

A15cecal:
We will ignore the first part of this issue for now. It might besolved already, and either way, there's no good way of testing this untill another issue related to this has been solved.

Fokusing on part two: 
Creating a dropdown list in playlink instead of a text field. This will be done with the same code as for the dropdown in code examples where you choose a file.

a15elibr:
@A15cecal Since part two seems to have been resolved, is part one of this issue do-able?

A15cecal:
@a15elibr It seems like @a15oscwe solved this as he was assigned an issue similar to this one.

a15elibr:
OK, closing issue then.

PR #3292, opened 2017-04-25
Playbutton is grayed out if there's no playlink and also redirects correctly

a15oscwe:
Playbutton will now be grayed out if the playlink is empty or if the file does not exists.
Also fixed the behavior of the playbutton when clicked. The page will now redirect based on if the file is global or not.


a15oscwe:
Related to issue #1878

PR #3300, opened 2017-04-25
Moved checkbox from sorting to filter

d15eriso:
Moved "onlypending" checkbox to the the new menu, still works and uses
local storage, I also and made it match the style of the menu for   a e
s t h e t i c   purposes.


a15freed:
ok merging

PR #3306, opened 2017-04-25
Icons/buttons in section editor

a15andau:
Changed the look of the "tab-like" buttons in section editor to look more like actual buttons


a15freed:
ok merging

PR #3309, opened 2017-04-25
Issue2813

a15chrau:
Removes the x:s from objects connected to objects that become deleted. Not completely done, when attributes become erased from an entity, the x is not removed from the entity. However when another attribute is erased, those two x:s become merged.


b15matvi:
Looks good!

PR #3311, opened 2017-04-25
Partial fix for issue #3032

a15oskst:
- Deletion of figures now works.

 - Moving them does not work.

Maybe we should splitt this issue into two issues? One for deletion and one for moving them. The reason for splitting them is that they are two problems and not one as we thought from the beginning.
If this is done the issue for deleting figures could be closed!


b15matvi:
Tested, seems to work fine. Will split the remaining issue.

PR #3315, opened 2017-04-26
Swimlane close button - changed pointer when hoover

a15andau:
Changed the mouse pointer from a marking  pointer to a "finger"-pointer when hovering over the X to close swimlane.


a15freed:
looks good

PR #3324, opened 2017-04-26
Fixed syntaxerror.

b15linek:
Fixed syntaxerror on all other pages than sectioned.php (because there it worked).


a15freed:
ok merging

PR #3356, opened 2017-04-27
Temporary fix of Hash Removal

a15oskst:
Temporarily fixes the problem with clearing hash. Clicking remove hash and delete all, now runs both functions to make sure that the hash gets cleared.


a15gunfu:
Looks good.

Issue #3088, opened 2017-04-18
Attribute and entity menu options should stay selected when you open them again


b15matvi:
The selected options get "deselected" when you open them again, this needs to be fixed.

HGustavs:
Every time that the dialog is opened it should be filled with the data from the objects... if it works on double-click it should work automatically.

b15matvi:
As the name part was bugged out from an earlier pullrequest it needs to be fixed inside this issue aswell

a15perbl:
Issue resolved.

PR #3385, opened 2017-04-27
Fix for issue #3088

a15gunfu:
All changes made in dialogue box is saved so when the user double click the symbol they can see the choices they've made.


b15matgu:
Looks good.

Issue #3375, opened 2017-04-27
Add functionality to 'New Group' button in grouped.php


a15nikho:
When the button is pressed a pop-up window should appear where the teacher can name the new group.

a15timbe:
Solved in #3389

PR #3389, opened 2017-04-27
Created a pop-up window when 'New Group' is clicked

a15nikho:
A pop-up window will now appear when 'New Group' is clicked in grouped.php. The point is to be able to create a new group in the database when 'Save' is clicked but that is a later issue.


a15timbe:
Good job, this solves issue #3375 and progress towards issue #3392

Issue #3382, opened 2017-04-27
Add functionality to the change password on the profile page


a15timbe:
This means saving the new password into the database, this will be similar to the change security question functionality

a15timbe:
Solved in #3642

Issue #3142, opened 2017-04-20
Add functionality to security question box on the profile page


a15timbe:
You need to actually be able to change your security question by submiting the form on the profile page. You will need to load and save to the database to enable this functionality.

a15ponja:
Since we changed from allowing the user to input their own security question to a selection of predetermined questions, which was the plan from the start. We should maybe consider to revamp how we save the question in the database, since we will save a lot of duplicate data if the system stays as it is now. I would say that we start saving questions as a relation between possible security questions to a user instead, i would like to hear your thoughts on this tough @HGustavs

a15ponja:
I talked with Henrik about my last comment and we are to keep it the way we have it now, since its more flexible.

a15timbe:
Solved in #3544

PR #3390, opened 2017-04-27
New files added for user settings

a15tobli:
These files will handle functionality for user settings at the profile page.


a15timbe:
Yes, this is a good start towards solving the issues #3382 and #3142 (they will not be closed). In the future please refer to the issue(s) in the name of the branch.

Issue #3365, opened 2017-04-27
Zooming in/out does not work correctly


a15perbl:
When using the zoom functionality it seems to not scale correctly. The canvas doesn't add to the original value, it adds to the currently stored value. This needs to be fixed.

a15perbl:
Issue resolved.

Issue #3270, opened 2017-04-25
Offset value when zooming in or out needs to be fixed


b15matvi:
When you use the zoom in or out buttons in the diagram the offset of where you click on the diagram is off and it makes it almost impossible to use the diagram. This needs to be fixed

a15perbl:
Issue resolved.

PR #3395, opened 2017-04-27
Fix for issues #3365 and  #3270

a15admsa:
The zoom function is now fixed, and the mouse point is also fixed. You can now zoom in and out without any problems. When zoomed in, you can press and move any figure or symbol that you want.


b15matgu:
Looks good.

Issue #3281, opened 2017-04-25
Offset when moving the diagram is broken, can't click any objects


b15matvi:
When you move the diagram, the offset to click an object is off. The selection border offset seems fine, but the actual offset to click an object is not fine.

a15perbl:
Issue resolved.

PR #3411, opened 2017-04-28
Fix for issue #3281

a15admsa:
The offset is now fixed. This fixes the issue #3281.


a14pethu:
tested it, seems to work without issues.

PR #3414, opened 2017-04-28
Issue#3081

a15chrau:
Any line drawn from a weak object will be drawn with two lines after clicking in the canvas.

Also changed the declaration of the index variable in most of the loops within diagram.js, in order to avoid future bugs.


b15matvi:
Tested it, works as intended.

PR #3415, opened 2017-04-28
Issue # Relation symbol should not be scalable

b15matgu:
Relation symbol is not scalable. Bug regarding resizing and marking figure is present.


b15matvi:
Tested, looks good.

PR #3441, opened 2017-05-02
Course Ed now has ellipsis (...) at the end of the headline, and the courses.

a15eriho:
The icons will line up when shrinked, and no wrapping will occur on small width screens.


b14marki:
Looks fine :)

Issue #3407, opened 2017-04-28
Remove unnecessary code in diagram files


a15perbl:
There's a lot of redundant code left behind in diagram.js and diagram_symbol.js, for example console.log(), probably used by developers to test the functionality of what they've implemented. If they're no longer needed they shouldn't be there anymore and should be removed.

a15perbl:
Issue resolved.

PR #3443, opened 2017-05-02
#3407: Removed console messages.

b15matgu:
Removed console messages. Old messages were spamming the console window.


b15matvi:
Looks good!

PR #3444, opened 2017-05-02
Issue#3206

a15chrau:
Done.

Implements a working standard width for all objects within the diagram, except line and figure.


b15matgu:
Looks good.

Issue #3442, opened 2017-05-02
"Delete all"-button should not move points


a15perbl:
When the "delete all"-button is clicked the points moves further up in the canvas to appear as they are deleted. The points shouldn't be moved, but instead removed from the array.

a15admsa:
@HGustavs Could you please come down and help us. We still have some issues with how we should fix the arrays when we remove things.

a15perbl:
Issue resolved.

PR #3459, opened 2017-05-02
Fix for issue #3442

a15admsa:
"Delete All"-button now removes the information about the crosses, but doesn't remove the poistion in the array.


a15gunfu:
Looks good!

Issue #3458, opened 2017-05-02
bugs in showdoc.php


d15johro:
bugs in functions identifier, parseLineByLine, markdownBlock and parseMarkdownBlock. Functions don't match functions in markdown.js

a15elibr:
Closing since solved - #3460

PR #3460, opened 2017-05-02
fixing bugs and refactoring showdoc.php

d15johro:
fixing parsing bugs so that it works correctly aswell as refactoring to match markdown.js


d15johro:
issue #3458

Issue #3200, opened 2017-04-21
Add functionality to reset session timer


a15tobli:
30 Minutes before the session ends an alert pops up and tells the user that the session is about to end. Add a button to this alert that resets the session timer instantly so that a user would not have to log out and log in to do this.

a15timbe:
Solved in #3474

PR #3468, opened 2017-05-02
#3200

b14larek:
Refreshes the session when user interacts with timeout div.


a15timbe:
In the future be sure to have a descriptive title in your pull requests.

Issue #2813, opened 2017-04-06
Remove The X's That Appear From Connecting Lines


a14davlo:
When you connect diagrams with lines, X's appear at the beginning and end of the lines, this issue is about removing them so that it looks better.

#2619 root issue

HGustavs:
Is this not just a question of removing the debug mode?

a15chrau:
It is not. When removing one of the attributes in the example program one of the bordering X:s, in the class it is connected to, should be removed.

HGustavs:
When we say "removed" we mean hidden.

If something is "deleted" from ther list it is not possible to trust the index values.

a15chrau:
Alright so:
They're only meant to be deleted from an object's connector- top, right, bottom or left, removed from the points array.

a15perbl:
Issue resolved._

PR #3496, opened 2017-05-03
Issue#2813

a15chrau:
Fixed the delete function, and removes all the relevant X's for issue #2813


a14pethu:
Seems to work, good work.

Issue #3498, opened 2017-05-03
Selected objects not deleted when clicking "Delete object"


a15perbl:
When a user selects more than one object and clicks the "delete object"-button they're not all removed at once, objects are removed pne-by-one. This means the button has to be clicked several times before everything selected is gone.

a15perbl:
Issue resolved.

a15chrau:
Issue stopped working after merge

b15matki:
Reopened issue due to the function not working as intended anymore,

a15perbl:
Issue is now resolved again.

PR #3504, opened 2017-05-03
Fix for issue #3498

a15chrau:
Created a difference betwen how a figure and a symbol is erase, which solves the bug when using the delete object function to delete figures.

Also added a global standard line-width for all drawn objects, the result is a bit cleaner diagram.


a15gunfu:
Looks good!

Issue #3462, opened 2017-05-02
"Delete" should not move points


a15perbl:
Much like issue in #3442, the delete button does not remove the selected points rather than moving them further up in the canvas. There's already a fix for the "delete all"-button which could probably be used to fix this particular issue.

PR #3508, opened 2017-05-03
Fix for issue #3462

a15chrau:
Waldo is not a point on th diagram anymore, but an empty string, causing the points to disappear completely.


a15gunfu:
Looks good!

Issue #2429, opened 2016-12-13
Marking preview is ugly


HGustavs:
![screenshot 2016-12-13 09 35 47](https://cloud.githubusercontent.com/assets/6978801/21133181/ea06d6c2-c117-11e6-8041-e4384b7c2039.png)

Several things are ugly here.

1. Styling of preview list box with interior shadows and rounded corners. It does not look like anything else. Standardize the look to look like something else we already have.
2. The P button gives us a popup preview, the link gives us a "full screen" pdf document. I would like to reverse this, to place the full screen pdf in the P burtton the preview in the link.
3. A you-tube-like full screen icon instead of a horribly styled button containing a P. The support-person-headset-feedback-icon in the icon folder instead of the feedback text button.
4. The headings in the preview file table do not clearly indicate that they are headings e.g. no borders or colors etc.

a15herny:
Found the _.css_ in _feedback_dugga.js_ and will continue with the task and debugg when fileupload works on the server.

a15herny:
Might wanna change this into 4 different issuse:
1. Changed the css
2. p = download, asdf.pf = preview
3. Will create an icon in illustrator
4. Will change the text from "Submission and feedback view" in preivew to "Preview window

a15freku:
Talk to Alice in group 2 they worked with the testdata for this problem, ask if they finished and if it's now in the latestest group merge.

a15elibr:
2 pull requests with part fixes has been submitted: 

- #3600 
- #3596

Up on til now I've split this issue into smaller ones by making part fixes and several pull requests, but for the documentation value it's more readable and more clear to create new issues for the part fixes instead. 

New issues created:

- #3600 
- #3602
- #3603

Therefor I am closing this issue and all further work will be done with the newly created issues which will solve this issue together.

PR #3514, opened 2017-05-03
Partfix for issue #2429 - restyling of preview list box

a15elibr:
I have as of now changed the styling on the box, by removing the weird inset shadowing. There has also been a slight shift of elements, added new table header for preview etc. 

The issue is not completely solved, as I am working on a new icon for the preview button, and also a restyling of the file submitting form. Since I dont think I'm gonna be finished with everything before the merge, I am splitting up the issue in smaller parts


a15elibr:
Ah, yes, that class needed tog go if the changes i wanted to make was gonna work. Can add though to the issue that all css should be in the shared CSS-files. I've already added som new css (since im still wokring on this issue) that will complement that later. The goal is to remove the inline-styling at the end.

a15oscwe:
Okay, then I'll merge this pull request.

Issue #3431, opened 2017-04-28
CSS styling for the cookies-window


A15cecal:
If you're on a mobile device the window looks weird. Fix it so that it adapts to the screen size for mobile devices.
![p](https://cloud.githubusercontent.com/assets/26705134/25529382/0ea0c702-2c22-11e7-87f1-5d27423a307b.JPG)

a15freku:
Check with @a15eriho it could be that he is working on something similar.

A15cecal:
This issue is almost the same as issue #3399, which is solved. Closing this.

A15cecal:
We were wrong about this issue, this still needs fixing and the other issue was not related to this problem.

a15freku:
Fixed with pullrequest #3559

PR #3550, opened 2017-05-04
fix for issue #3431

A15cecal:
This should fix the problem, works on different devices that's been
tested.

We might want to check the add dugga-button, looks like it might need
some styling for phones because it looks like it's "inside" the cookie
window, but this should be a seperate issue.


a15ponja:
Closed due merge

PR #3559, opened 2017-05-04
issue 3431 a15cecal w18, new

A15cecal:
This should fix the problem, works on different devices that's been
tested.

We might want to check the add dugga-button, looks like it might need
some styling for phones because it looks like it's "inside" the cookie
window, but this should be a seperate issue.


a15wilhe:
There was conflicts with the css classes nowrap, elipsis and showhidetext. I couldn't see any conflicts. I let everything be as it is. It might just have been some line interference.

Issue #3355, opened 2017-04-27
Add functionality to menu for relation symbol type


a15perbl:
A dropdown menu should be added to the dialog box that appears when doubleclicking the relation symbol. A user should be able to choose if the symbol should be weak or strong. Functionality for creating a strong relation already exists, however not for a weak relation.

a15perbl:
Issue resolved.

PR #3562, opened 2017-05-04
Fix for issue #3355

a15gunfu:
When weak is selected, an extra border appears around the figure to signify the weakness of the relation.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1070,6 +1070,11 @@ function dialogForm() {
         form.innerHTML =
             "Relation name:</br>" +
             "<input id='nametext' type='text'></br>" +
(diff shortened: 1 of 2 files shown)


a14davlo:
Okay so I copied your code and I tested it out on my grupp4/a14davlo and here's what I saw.

I saw that there's a dropdown option on Create Relation where you can select Strong and Weak relation type as shown in the photo below.

However, all the ER & UML default examples are gone, there's a completely blank sheet you start with and you cannot create anything. All the "create" buttons does not work, nor does the minus, plus & start moving. 

I cannot test to see if the ER relation is different or not, the buttons does not work.

review pull request example

a14davlo:
Seems to work just fine on grupp4/a15gunfu

Issue #3410, opened 2017-04-28
Comment functionality for existing functions in diagram files


a15perbl:
There's a lot of newly implemented functions without any comments which makes it hard to figure out what they actually do. It would probably be better to have them commented both for current and future developments.

b15matgu:
Issue needs to stay open as not all functions is described.

PR #3569, opened 2017-05-05
#3410 - Added comments

b15matgu:
Added and formatted comments.


c14joalu:
Everything looks good.

PR #3578, opened 2017-05-05
Issue#3097

a15chrau:
Object's top left corner now snap to the grid that is drawn. I attempted comparing the two corners bottom right and top left in order to decide which one to snap to the grid, but it was deemed unsuccessful as it lowered the overall quality.

Suggestion for new issues:
* The grid is not drawn properly (bug)
    - When navigating far outside the start position in the canvas, the grid is no longer drawn
* Snap a group of objects to grid


b15matgu:
Works like it's supposed to.

PR #3587, opened 2017-05-05
show/hide looks better now

b15sarha:


a15wilhe:
Remember to reference the issue in the description.

Issue #2456, opened 2017-02-28
Diagramming engine refresh rates


HGustavs:
Refresh of the diagrams should be timer-based.

* If last edit of diagram was recent, update time should be short also, down to a few seconds.
* If last edit of diagram was not recent (hours to days) minute-based updates should be used
* If diagram has not been edited for several days (a week +), diagram should essentially be read-once only, and update status should only be queried on refresh.

HGustavs:
Updates should only use network if page is visible i.e. tab is in front.

https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API

This way the timers will not update for invisible pages.

If a page becomes visible after having been invisible we should perform a "page refresh"

PR #3595, opened 2017-05-05
#2456

a14pethu:
add a refresh engine. the code prints out console logs so we can see that it works.


a15admsa:
Looks good! Great job!

PR #3604, opened 2017-05-06
Added Event listners for touch

a15oskst:
Added event listners for touch. Tried it on a android 5.0 device with chrome as browser.

Now it at least tries to move the attributes and entitys etc. But almost nothing is actually working as expected on phone. The scaling isn't working. The canvas element scales down aka the boarder fits on the screen, but the diagram it self doesn't scale down very well.


a15oskst:
Screenshot of the website on my android device.

a15admsa:
Looks fine! This can be merged.

PR #3606, opened 2017-05-06
Hamburger menu in Section Ed is now completed

a15eriho:
It is deactivated though, and will be activated once this issue, and the issue with disappearing buttons, has both been merged.


a15eriho:
This function is now activated, and will not break any buttons.

Issue #3607, opened 2017-05-08
Create function that can sort connectors


a15perbl:
To dynamically sort all connectors for all objects in the diagram, the function needs to be reworked. Instead of only sorting connectors for a specific object it would be preferable if it did for every existing element in the canvas.

a15perbl:
Issue resolved.

PR #3610, opened 2017-05-08
#3607 - Sort connectors

b15matgu:
Sort all connectors related to entity dynamic.


b15matvi:
Looks good! Makes it possible to connect lines to entities!

PR #3619, opened 2017-05-08
Remove hashed diagram changes

b15matvi:
Loops through localstorage and removes all localdiagram


b15matgu:
Working like i should.

Issue #3621, opened 2017-05-08
Created relations should snap back to fixed size


a15perbl:
When creating a relation a user can make the object as big as they want. Relations should preferably be of a fixed size, which is not the case as it is right now. If a relation is made smaller or bigger it should snap back to the intended size.

a15perbl:
Issue resolved.

PR #3632, opened 2017-05-08
#3621 - Size of relation is fixed.

b15matgu:
Relation gets standard size even if a selection is made bigger or smaller than default size.


a14pethu:
Seems to work without problems.

Issue #3564, opened 2017-05-05
Dialog window needs to display set values


a15perbl:
When the relation symbol is doubleclicked and the dialog window appears the latest saved settings aren't displayed. It would be better if these did displayed the current values to not confuse the user.

a15gunfu:
Was fixed together with issue #3355 and already merged in. So the set values are all saved when user doubleclick the relation symbol.

a15perbl:
Issue is now resolved after discovering the fix didn't merge in along with another issue after all.

PR #3643, opened 2017-05-08
Fix for issue #3564

a15gunfu:
Settings are saved when the user doubleclick the relation symbol to re-open dialogue box.


a14pethu:
Works fine.

PR #3644, opened 2017-05-08
Can't resolve this issue

a15admsa:
The fixes that have been implemented makes it possible to zoom and move one time. But when you want to move when zoomed in, the offset doesn't work as it should. I'm leaving this issue for a later date.


a15gunfu:
Objects can be moved after zooming, though some problems with the canvas after zooming out. So the part about moving objects looks good!

Issue #3589, opened 2017-05-05
Global settings - Fonts


a15perbl:
Much like a user can individually change font of text inside an object, there should be a function which allows setting a font globally for every present object in the canvas. 

#2473 root issue

a14pethu:
Works fines.

a15perbl:
Issue resolved.

PR #3645, opened 2017-05-08
#3589

a14pethu:
There is now a button to change the font on all entites globaly, except for classes.


c14joalu:
Everything looks good!

PR #3646, opened 2017-05-09
Fix for removeLocal()

b15matvi:
Currently we only store localdiagram without any numbers, that's why "localdiagram" + i did not work. If we however start storing localdiagrams with numbers we would have to revert this change.


b15matgu:
Working like it should.

Issue #3623, opened 2017-05-08
Relations should be scalable but snap back to fixed size


a15perbl:
Currently relations cannot be manipulated in any way. It would be better if the relation could be manipulated and snapped back to the intended size instead of making it completely impossible to click the points. This could potentially also fix the issues with scaling the figures.

b15matgu:
Working like it should.

a15perbl:
Issue resolved.

PR #3647, opened 2017-05-09
Solved issue #3623 - Sizing of entities.

b15matgu:
Solved problem regarding sizing all entities except relations. All other points, including figure, can be moved.


b15matvi:
Tried out, works like it should!

Issue #3427, opened 2017-04-28
Fix cross and overlay


b14marki:
Fix cross and overlay across the page.

PR #3649, opened 2017-05-09
#3427 Fix opacity overlay

b14marki:
Now it is the same opacity as loginbox.php.


a14phili:
pull request was made with old version.

Issue #3655, opened 2017-05-09
Magic headings dissapear when you filter content for results


a15mikla:
When you filter the content in the result-view the magic heading dissapears. When scrolling the heading is visible again but it should stay after filter.

PR #3658, opened 2017-05-09
Magic heading now stays visible after filter

a15mikla:
The magic heading now stays visible after the content has been filtered
in the result-view.
Solves issue #3655


a14phili:
Ok merging

PR #3668, opened 2017-05-09
White spaces are not needed anymore between pipes.

a15ramha:
Now there is no need for spaces like | Head1 | this |Head1| can be used too.


a15wilhe:
remember to reference to issue :)

Issue #3383, opened 2017-04-27
Only allow users that are on HTTPS to change password


a15timbe:
This is in regards to the comment on issue #2638 and has been partly solved with the merge #3305

a15timbe:
Solved in #3673

a15ponja:
This one was not solved correctly.

a15timbe:
Solved in #3822

PR #3673, opened 2017-05-09
Password management now only works on HTTPS connections

a15tobli:
Added functionality which checks if the browser is using a HTTPS connection to LenaSYS. If that's not the case, the password form at the profile page will be hidden. This adds extra security for password management.
Issue #3383.


a15tobli:
The testing for https is impossible for the site as it's not using this. I've tried reversing the selection in the code with the result that the form is displayed properly and it should work without problems if the connection would in fact be set to https.

PR #3686, opened 2017-05-09
Switch cases and ? operator changes

a15andau:
Changed switch cases that had too few cases to if-statements and changed ? operator to if-else statements according to Henriks wishes.


a14phili:
Ok merging

Issue #3624, opened 2017-05-08
Add odd even colors to table and decrease height of rows


a15freku:
![oddeven](https://cloud.githubusercontent.com/assets/26705670/25797873/4f9bc15c-33df-11e7-82d3-08f1d58376f6.png)

a15freku:
Should look similar to files view

a15wilhe:
The height of each row has been reduced and fixed with pull request which is approved. 

I have tried to add odd and even styleing but there are "hidden" tablerows in an unspecific amount so somethingg like "tr:nth-child(3n+2)" won't fix it. I will continue trying to fix it with jQuery after the AJAX renderer.

a15wilhe:
For fixing the odd even background coloring i tried first in jQuery :
`$('#testTable tr:visible .fumo').not('variantInfo').filter(':odd').css({'background-color': 'red'});'red'});
$('#testTable tr:visible .fumo').not('variantInfo').filter(':even').css({'background-color': 'yellow'});`

Then in css:
`.list#testTable tr.fumo:not(.variantInfo):nth-of-type(odd) {background-color:LightGray;}
.list#testTable tr.fumo:not([style="display:none"] ):nth-of-type(even) {background-color:WhiteSmoke;}`

Nothing works the way I want, there are "hidden" rows and innertables that makes it tough.

PR #3690, opened 2017-05-09
Changed padding for table contents

a15wilhe:
Changed (lowered) padding for td, input, label and "buttons" in table in tests (dugggaed) so the height of each row is more compact like other tables.

This is a partfix for issue #3624 .

You can see the changes in my folder.


a15wilhe:
changed a lot more and drastically. Font-size, padding and borders! It is now compact :)

a15wilhe:
changed font-size value in duggaed.js, lowered more padding and border. Also added specific height and color for the input.

a15freku:
Looks good

Issue #3710, opened 2017-05-10
Comment out "create class"-button


a15perbl:
Classes have no relevancy to ER-diagrams and are currently redundant to include in the diagram. However, the functionality might be needed for later use and should be commented out so that it still exists but doesn't show.

a15perbl:
Issue resolved

PR #3717, opened 2017-05-10
#3710 - Commented out class button.

b15matgu:
Commented out class button, as it's not needed at the moment.


a15gunfu:
Looks good.

Issue #3586, opened 2017-05-05
Global settings - Line thickness


a15perbl:
Currently there is no way to change settings globally for all symbols. A user should be able to change line thickness for every symbol present in the canvas with ease.

#2473 root issue

a15chrau:
This can be closed.

Issue #2473, opened 2017-03-29
Global settings for diagrams


HGustavs:
A global settings dialog for the diagrams is needed in some way.

Line thickness
Default Colors
Fonts of text in diagrams

etc.

a15chrau:
Looks finished.

a15perbl:
Closing issue for the moment.

PR #3722, opened 2017-05-10
Issue#2473

a15chrau:
Adds all the global menus into one, breaks down their functionality to be more consistent. Added a menu for editing global line-thickness #3586


a15chrau:
#2473

Issue #3631, opened 2017-05-08
Hash security question answer when saved in profile page


a15ponja:
Right now it saves the answer in the database in plaintext.

a15timbe:
Solved in #3725

PR #3725, opened 2017-05-10
Hashed security question answer.

a15filba:
The security question answer is now hashed when saved in the profile page. (#3631)


c14caran:
I'm not the reviewer but these changes are wrong. You should use the standardPasswordHash function introduced in #3207 instead of just using password_hash. This allows for changes in the hash settings in the future.

Issue #3713, opened 2017-05-10
Remove example objects


a15perbl:
Example objects loaded in the diagram are redundant and does not work like the objects generated by users. They're currently just in the way and it would be better if they were removed.

b15matgu:
Should the whole file be removed or just be commented out?

a15perbl:
Commented out. They might be needed later in the future.

a15perbl:
Issue resolved.

PR #3726, opened 2017-05-10
#3713 - Remove example objects

b15matgu:
Example objects and points is commented out.


a14davlo:
Yeah this is okay, I approve.

Issue #3590, opened 2017-05-05
Global settings - Font colors


a15perbl:
It should be possible to change font color of every object present in the canvas instead of having to doubleclick them individually and change everything.

#2473 root issue

a15perbl:
Issue resolved.

PR #3741, opened 2017-05-10
#3590

a14pethu:
We can now change the font color on all entites at once.


b15matvi:
Tested out, seems to work as intended.

Issue #3707, opened 2017-05-10
Remove console in diagram.php.


b15matki:
The console should be removed since it is not needed at the moment. It doesn't need to be removed as a whole but rather "commented" instead so that the code remains but is not executed.

a15admsa:
This issue is now fixed!

PR #3746, opened 2017-05-10
Fix for issue #3707

a15admsa:
I've now removed console from the site plus made so that the crosses are not presented. The crosses can be presented if we press the "debug"-button


a15gunfu:
Looks good.

Issue #3753, opened 2017-05-10
Set a max/min value for zooming


a15perbl:
Zooming in/out can be done an infinite amount of times right now. Zooming should be limited, e.g set the maximum zoom in to 200% and zoom out value to 50%.

HGustavs:
i say minimum 5% and maximum 1600% seems about enough.

HGustavs:
Not necessarily 10% increments, perhaps from a list of values e.g. 5% 10% 20% 30% 50% 75% 100% 125% 150% 200% 400% 600% 800% 1000% 1200% 1400% 1600% ??

a15perbl:
Original issue resolved. A new issue has been opened about adding the list of values (#3764)

PR #3755, opened 2017-05-10
Fix for issue #3753

a15admsa:
I have now fixed the issue so that a user only can zoom in to 200% and zoom out to 50%.


a14pethu:
seems fine.

Issue #3785, opened 2017-05-11
Clean up markdown.js and refactoring


a15ramha:
There are some unused code that need to be deleted in markdown.js.

a15oscwe:
Fixed with pull request #3788

PR #3788, opened 2017-05-11
The code is clean from unused code.

a15ramha:
The code is now updated and the unused code is deleted from markdown.js


a15ramha:
The issue number were wrong and the correct one is #3785

Issue #3766, opened 2017-05-11
Put code relevant for dialog menus in its own file


a15perbl:
By moving the code which relates to the dialog menus in one file we will make it easier for current and future developers to find and make sense of it. A new file should be created in which the functionality for dialog menus should be put.

a15perbl:
Issue resolved.

Issue #3733, opened 2017-05-10
Generate message when AJAX error occurs


a15perbl:
When an AJAX error occurs the browser redirects the user to a 404 error page. It would be preferable if the user got to stay on the current page if this happens. A function which generates an error message should be implemented.

a15perbl:
Issue resolved.

PR #3794, opened 2017-05-11
Issue #3733 & #3766

a15chrau:
#3766: Cuts out the dialog window functionality into a seperate javascript file.

#3733:
Created a function for loading the forms with a simple AJAX-script.

This also gives the editing within the diagram more visual feedback. It laso removes a lot of redundant code by making functions more streamline.

ID:s and inner variables AttributeType, EntityType and RelationType have been removed and replaced with "object_type".

Fixed a bug where double-clicking anywhere in the diagram (that isn't an object) causes an error.


b15matgu:
Some few bugs like getting the actual name in the dialog and not viewing all the characters is not working properly. Otherwise ok.

PR #3797, opened 2017-05-11
Added progress bar.

b15linek:
There is a progress bar added to show installation progress.


a14phili:
Nice progress bar!!! works fine

Issue #3764, opened 2017-05-11
Add a list of zooming values


a15perbl:
As suggested by feedback from issue #3753, a user should be able to zoom in/out the canvas by picking a value from a dropdown list. Suggested range is 5%-1600% with varying increments between the minimum and maximum value.

a15perbl:
Issue resolved.

PR #3807, opened 2017-05-11
Fix for issue #3764

a15admsa:
The zoom recommendation is now fixed. I've also fixed the grid when you zoom in and zoom out.


b15matgu:
Works like a charm.

Issue #2455, opened 2017-02-28
Offline diagramming


HGustavs:
Since the diagramming engine works through reading or writing json documents, an offline or slow-internet version of the engine should be possible e.g. by storing the diagram in local storage and ony infrequently updating the on-server document.

a14davlo:
c14joalu will be working on this. If this issue is too big for a few days of work, come and talk with me and we'll sort it out and split up the work better.

c14joalu:
@HGustavs , I wondered with this, is it just as simple as that it will update the on-server text document when possible ( or set a timer for it ) and simply have a button for a person to be able to view the canvas text document so they can save it manually on the computer if there is no connection while working on it?

c16carbe:
This is already possible. Issue should be closed.

PR #3809, opened 2017-05-11
#2455

c14joalu:
Fixed save to server/ and fixed frontpage website creating correct folder structure and posting the ID and FOLDER to diagram so it can put the files in correct folder


a15chrau:
Seems to work as intended while not breaking any existing code.

PR #3832, opened 2017-05-12
Added a few comments

a15andau:
For clearification


a14phili:
Solved a conflict but it works fine

Issue #3838, opened 2017-05-12
Remove duplicate styling


a15wilhe:
selector  `.list#testTable input` is set 2 times in style.css. The second one has all the declarations as the first. Remove the first one and hope no other groups will add to the first one.

b15sarha:
resolved.

PR #3844, opened 2017-05-12
Removed duplicate selector and styling

a15wilhe:
Removed a selector and it's styling since its a duplicate. The second one (selector) a few lines down contains more styling and the styling I removed. So this won't interfere what I see. Also let's hope no other group won't style in this selector. This styling is for duggaed (tests) table.


a15wilhe:
this solves is #3838

Issue #3719, opened 2017-05-10
Customization in dialog menu for lines


a15perbl:
A line can be doubleclicked and selected in the diagram but no customization can be done.

![bild](https://cloud.githubusercontent.com/assets/26705103/25891374/2377086e-3571-11e7-8800-c18eb4f7c864.png)

Having lines made customizable is better than altering the lines via entities or relations, especially when you only want a specific line to be doubled. As it is right now, all the lines connected to the entity and relations are doubled when the option 'weak' is picked.

The dialog menu for lines should be changed so that a user can pick if a line type should be weak or strong.

a15perbl:
Issue resolved.

b15matki:
Solution merged into the G4 branch!

PR #3850, opened 2017-05-12
Issue #3719 - Customization in dialog menu for lines

b15matgu:
Customization for lines (strong/weak) can now be done. There is an issue regarding selecting the line and open the dialog, sometimes it works and sometimes it doesn't.


b15matki:
Looks good now that the conflict is gone.

Issue #3852, opened 2017-05-12
Fix relation of lines when relation is weak.


b15matgu:
The offset of the lines is not aligned when the relation is weak.
![relation](https://cloud.githubusercontent.com/assets/26705189/25995753/2efaa4f8-3714-11e7-93a5-a04159eb47e8.JPG)

HGustavs:
At the same time -- take a look at the issue that has to do with resizing the relation symbol so that it is a multiple of the entity grid size, so that the relationship lines are not crooked by default.. those slanted lines are an eyesore....

![screenshot 2017-05-10 16 59 46](https://cloud.githubusercontent.com/assets/6978801/25996137/e9601e6c-3715-11e7-84c9-fa97dcbcb80c.png)

b15matki:
Solution merged into the G4 branch!

PR #3855, opened 2017-05-12
Issue #3852 - Line offset on weak relation.

b15matgu:
Fixed the offset of the lines when the relation is weak.


b15matki:
Seems to work fine now that the conflict is resolved!

PR #3858, opened 2017-05-12
Issue 3756 - Default size of relationship not even multiple of default size of entity

b15matgu:
Changed default size of objects to match and align to relations.


b15matki:
Works just fine!

PR #3861, opened 2017-05-12
Fix for issue #3852 - Line offset on weak relations

b15matgu:
Fixed offset to align on weak relation.


b15matki:
Seems to fix the previous alignment that got affected by the last pull request ( #3852 ). Now it seems to be the desired size.

Issue #3854, opened 2017-05-12
Align checkboxes in installer better.


b15linek:
This is annoying, fix this:
<img width="622" alt="screen shot 2017-05-12 at 13 11 46" src="https://cloud.githubusercontent.com/assets/26706647/25995861/93e937f8-3714-11e7-9c84-f9ad5568852e.png">

PR #3862, opened 2017-05-12
Checkboxes aligned and some media styles added.

b15linek:
Aligned two checkboxes in installer. This solves issue #3854. Also added some media styles for smalles screens.


a14phili:
Looks good

Issue #3826, opened 2017-05-12
Weak entity stopped working.


b15matki:
When selecting an entity to be weak nothing happens. It should create a second outline on an object.

a15chrau:
May be related to pull request #3794

a15perbl:
Issue resolved

PR #3873, opened 2017-05-15
Issue #3826

a15chrau:
Draws derived objects proprely, and doesn't make the rest of the diagram to become derived.


b15matvi:
Looks good!

PR #3875, opened 2017-05-15
Canvas fix

a15admsa:
The canvas didn't load properley before. Now this should work just fine.


a15chrau:
Looks good. Shouldn't cause any problems with other solutions.

Issue #3874, opened 2017-05-15
No need for selecting objects to move points.


b15matki:
It would be better to simply move the points of the objects instead of requiring to select the whole object itself first.

b15matki:
Solution merged into the G4 branch!

PR #3876, opened 2017-05-15
Issue #3874 - Fix for moving points.

b15matgu:
No need to select object to move points.


b15matki:
Awesome! Mergin into the G4 branch.

PR #3879, opened 2017-05-15
No more overlap in mobile view

a15andau:
Now the course name doesn't overlap the Show/hide all-button in Section Editor.


a14phili:
Works fine

Issue #3802, opened 2017-05-11
Set object name no longer displaying when doubleclicking


a15perbl:
The object name is no longer fetched and put in the dialog menu when doubleclicking. The field "name" should always display the set name. Needs to be fixed.

![bild](https://cloud.githubusercontent.com/assets/26705103/25941821/15c4ebc6-363b-11e7-8c37-672d9dc72e8a.png)

a15perbl:
Issue resolved

a15perbl:
Reopening since it's broken again.

PR #3882, opened 2017-05-15
Fix for issue #3802

a15admsa:
I've now fixed so that when you doubleclick on a symbol, it will load the current data from the array into the forms. I've also added some padding to the Name-box. 

NOTE: When you doubleclick, the console will say that there is an error. But there is no error. This is a bug for a later issue.


b15matvi:
Looks good!

Issue #3847, opened 2017-05-12
Make progress bar dynamic for window re-sizing.


b15linek:
Right now the progress bar is incorrect if window gets re-sized during installation. This could be fixed.

HGustavs:
Screenshot plz.

b15linek:
This happens is window gets resized (shold be full):
![namnlos](https://cloud.githubusercontent.com/assets/26706647/25997795/1a72b390-371e-11e7-9600-e311e4f8b70a.png)

a14phili:
solved and merged #3883

PR #3883, opened 2017-05-15
Window resize scales progress bar.

b15linek:
It is now possible to resize the window during/after installation and it will properly scale the progress bar relatively. Fix for issue #3847


a14phili:
works fine

Issue #3842, opened 2017-05-12
Line thickness range overlapping dialog window


a15perbl:
The line thickness range overlaps the dialog window which doesn't look very good. Should probably be fixed.

![bild](https://cloud.githubusercontent.com/assets/26705103/25992194/1f712d12-3706-11e7-812c-893b8b1e7199.png)

a15perbl:
Issue resolved.

PR #3890, opened 2017-05-15
Fix for issue #3842

a15gunfu:
The line thickness option do not longer overlap the dialogue window.


b15matgu:
Looks good.

b15matki:
Mergin into the G4 branch!

PR #3894, opened 2017-05-15
Show/hide all transforms

a15andau:
When window is re-sized down, Show/hide all is transformed to become smaller


a14phili:
It looks great

PR #3895, opened 2017-05-15
Swimlanes are working again, so swim along

a15eriho:
Now works in a proper LenaSYS way.
Keep issue open, will comment and refactor some small variables before closing issue


a14phili:
looks perfect!

Issue #3678, opened 2017-05-09
Create test files for file upload filtering


a15jorpa:
For the new style of filtering file uploads in #3384 we need some files for testing how the allowed file types are detected by php. The following types are allowed and need test files:

txt (done)
pdf (done)
gif (done)
jpeg (done)
jpg (same as jpeg)
png (done)
zip (done)
**rar**
html (get from project)
java (done)
**xml**
js (get from project)
css (get from project)
php (get from project)
md (get from project)
**ai** (Adobe illustrator)
**psd** (Some kind of Photoshop file?)
svg (get from project)
sql (get from project)
doc (done)
docx  (done)
odt  (done)
**xslt**
**xsl**
**sl**
**glsl**

The files used should be ones where we don't need to care about copyright (our own preferably). Where would be a suitable place to put these kinds of files?

PR #3896, opened 2017-05-15
Test files for PHP file type

a15jorpa:
Added directory for test files. Files added as per #3678. Need approval for creation of new directory. To use the test files, open the "_mime.php" file in a way that runs the php.


a14phili:
Works fine

Issue #3743, opened 2017-05-10
Drawing menu


HGustavs:
Currently the draw diagram menu looks dated, I want something like google draw or adobe illustrator

Clear Icons representing each kind of functionality to the top left...  abd the drawing document below amd to the right of the menu butons.

HGustavs:
![screenshot 2017-05-10 14 42 17](https://cloud.githubusercontent.com/assets/6978801/25899179/2a4c23c2-358f-11e7-84d0-4605b743b62c.png)

a15perbl:
A menu with icons are being worked on with issue #3361 if that's what you meant?

a15chrau:
Jag har gjort en liten prototyp: http://webblabb.iit.his.se:8080/grupp4/a15chrau/LenaSYS/DuggaSys/diagram.php

Fig 1.
![image](https://cloud.githubusercontent.com/assets/26707347/26061630/bdfb1220-3988-11e7-91aa-d8caba97846e.png)

Fig 2.
![image](https://cloud.githubusercontent.com/assets/26707347/26061721/fd24b636-3988-11e7-8d07-cadf619a6130.png)

Fig 3. 
![image](https://cloud.githubusercontent.com/assets/26707347/26061749/134307b0-3989-11e7-927e-d2b7b2df8d56.png)

a15chrau:
Här i Fig 4 ser du de kommande ändringar jag planerar att göra. Dessa består av att:
* 
* Flytta upp Change appearance och Change Global Appearance

- [Flytta ned verktygen till vänster del av skärmen.] 

- [ ] 

- [ ] 

- [ ]


 4.
![image](https://cloud.githubusercontent.com/assets/26707347/26061928/9de1728a-3989-11e7-9cef-776bcb911848.png)

HGustavs:
Not too keen on dropdowns, especially on mobile dropdowns are a bit tedious...

* rare things such as "download as svg" can work in a dropdown.
* Common things such as "draw class" should be butons.

After some more thinking, perhaps letting the user choose either floating vertical or static horizontal menu could be a good compromise. 

This horizontal or vertical setting can probably be stored in local storage, so that one does not have to re-work the user interface every time one uses the app on the same computer.

a15chrau:
Would it be better to have a seperate solution for mobile?

HGustavs:
[...]

PR #3907, opened 2017-05-16
Application document menu

a15chrau:
Does not fully solve #3743, preliminary push


a14pethu:
Seems to work.

b15matki:
Good! Mergin it with the G4 branch!

Issue #3908, opened 2017-05-16
Duplicate and needlessly large files in testfiles


a15jorpa:
Some files used for testing are bigger than needed for that purpose, some are included in the invalid directory without being invalid.

PR #3912, opened 2017-05-16
Removed/shortened test files

a15jorpa:
Fixed problems with test files as per #3908. Changes should not affect LenaSYS functionality.


a14phili:
works fine

Issue #3866, opened 2017-05-12
Pressing enter when installing in chrome.


b15linek:
If you press Enter during input of installation data in chrome, it automatically submits the form. This should not happen. Enter key should press the right arrow before the last page. 

It works in safari but not in other browsers (only chrome tested so far).

b15linek:
Seems to be a problem only in Windows. Mac works fine.

b15linek:
Linux seems to have problems too.

b15linek:
Solved.

PR #3926, opened 2017-05-16
Enter now works in Linux and Windows

b15linek:
Pressing enter to move on in installation should now work on different browsers in different operating systems.


b15linek:
Fix for issue #3866

a15eriho:
Works on Windows x64 Chrome.

Issue #3771, opened 2017-05-11
Implement functionality for dashed lines


a15perbl:
Dashed lines for the attribute itself already exists, however dashing the lines between entities is not yet possible. Since a menu for customising lines is being made ( issue #3719 ), adding an option in the dropdown list should also be done.

#3735 root issue

a15perbl:
Issue resolved!

PR #3933, opened 2017-05-16
Fix for issue #3771

a15gunfu:
Enabled a dialogue window that will now appear when line is made and implemented a dashed line called "forced".


b15matgu:
Works like a charm.

Issue #3815, opened 2017-05-11
Change the look for zoom selection


a15perbl:
The zoom selection looks pretty basic at the moment. Styling it up would be a good change, do you have any suggestions on how to change them? @HGustavs 

![bild](https://cloud.githubusercontent.com/assets/26705103/25943178/5236ad84-363f-11e7-9e5c-584c98a1b23d.png)

b15matki:
**Feedback from @HGustavs :** 

The zoom button should lie within the canvas itself to enable the canvas to get as much space as possible. The coordinates should also be placed within the canvas itself rather than having a reserved space beneath the canvas.

**General feedback:** 
It would probably be cool if some basic css would be implemented, such as round corners, a colored "semi-button" where the arrow-down is displayed.

a15perbl:
Issue resolved

PR #3940, opened 2017-05-16
Fix for issue #3815

a15admsa:
I've now added some new things and moved some things in the diagram:

*   I have moved the coordinants+zoom-div so that it is inside the diagram.
*   I have moved the zoom-select so that it is inside the diagram
*   I have styled both the coordinants+zoom-div and the zoom-select so that the user experience gets a boost.
*   I have changed the way that the zoom-select works. We now have a arrow that will rotate when you open the select box.
*   I have moved all the CSS from those into the style.css file.


b15matgu:
Looks awsome.

Issue #3943, opened 2017-05-16
Change dashed line name from "forced" to "derived"


a15perbl:
The dropdown options for lines works correctly, but the option "forced" is named incorrectly. Change the name from "forced" to "derived".

PR #3946, opened 2017-05-16
Issue #3943 - Change name on line types

b15matgu:
Change name on line types, weak to forced and forced to derived.


a15chrau:
Fixes the problem and doesn't break anything.

Issue #3945, opened 2017-05-16
Fix global line color regarding forced relationships.


b15matki:
Currently the third line in the line type that indicates a forced relationship inherits the color of the outer lines. The middle line should always be white.

a15perbl:
Issue resolved

PR #3948, opened 2017-05-16
Issue #3945 - Fix global line color regarding forced relationships.

b15matgu:
The center line in the forced relation is now always white.


a14davlo:
Minor fix, it's okay.

Issue #3932, opened 2017-05-16
Fix the errors that occurs when using the globalsetting-tool.


b15matki:
When using the globalsetting-tool several errors occurs that need to be fixed.

a15perbl:
Issue resolved

PR #3949, opened 2017-05-16
Fix for issue #3932

a15admsa:
This fixes the error messages that are displayed when you use the global apperance-button.


a15gunfu:
Looks good!

Issue #3923, opened 2017-05-16
All lines set to dashed when creating attribute


a15perbl:
When an attribute is created all lines in present in the canvas becomes dashed, however when the dialog menu is closed everything returns to normal. Small bug that needs to be fixed.

![bild](https://cloud.githubusercontent.com/assets/26705103/26097931/86045c20-3a26-11e7-8ec3-e799dc3f830c.png)

a15perbl:
Issue resolved

PR #3956, opened 2017-05-17
Issue #3923 - Fix for dashed grid

b15matgu:
Now the grid doesn't get dashed when creating attributes and lines.


a15chrau:
Should solve the issue

Issue #3916, opened 2017-05-16
weird margin-top on headings in css presented in markdown


d15johro:
parsed markdown have weird margin-top on headings. For example the h4 tag have a margin of "1px 0px 8px 0px" which makes the break-line tag above it useless. Also, the heading merges with the text above which makes it look semantically incorrect.

d15johro:
how it looks:
![bug-css](https://cloud.githubusercontent.com/assets/26705142/26096674/3f4c7654-3a22-11e7-8436-40f5ce784956.png)

increased top margin by 10px: 
![bug-css-fixed](https://cloud.githubusercontent.com/assets/26705142/26096706/60001e6e-3a22-11e7-89b2-6e196059602a.png)

d15johro:
what do you think? @HGustavs

HGustavs:
What is the difference?

Or is it that there is no difference?

HGustavs:
Now I get it...

Yes the bottom one seems more semantically correct?

HGustavs:
It should work mostly as "pre" e.g. a single normal line feed makes a line feed.

d15johro:
I don't understand. What's the preferred way of solving this issue? Changing the css of headings might effect code elsewhere and the "br" doesn't seem to have any effect?

**"br" above and under headings:**
![asd](https://cloud.githubusercontent.com/assets/26705142/26100593/a317a1ac-3a2e-11e7-8709-bc3bb229f24b.PNG)

@HGustavs

a15freku:
Solved

PR #3970, opened 2017-05-17
Week20 #3916 d15johro

d15johro:
boxing markdown with a div that contain a div with a unique id so you can style markdown invidivually. also chaning the heading styles for nicer output.

issuue #3916


d15johro:
some refactoring were also made.

Issue #3958, opened 2017-05-17
Move input to first empty field in installer


b15linek:
It's annoying to click the field you want to enter data into. Move input to the first empty one. By doing this it will be possible to complete installation without using the mouse.

a14phili:
solved and merged #3972

PR #3972, opened 2017-05-17
Autofocus on input fields

b15linek:
Installer will now focus the first empty field on current page. This will make it easier to navigate by keyboard only.


b15linek:
Fix for issue #3958

PR #3975, opened 2017-05-17
Merge pull request #3965 from HGustavs/#3839

a15freed:
The graphical bug in result-view when hovering the column is now fixed making the highlight look better


a15freed:
this went wrong somehow, reposting..

Issue #3968, opened 2017-05-17
Add odd even bkgrd-coloring when searching tables in fileed


a15wilhe:
After the user has searched the tables the odd and even background-color is no longer set since rows are hidden.

a15wilhe:
Fixed with pr above

PR #3977, opened 2017-05-17
Added odd even bkgrdColor after searching in tables in fileed

a15wilhe:
To two keyUp function that filters tables and hide rows, I added odd even bkgrd coloring in jQuery. 

This solves is #3968.

You can see the result at my folder.


a15wilhe:
hmm, to be super clear, this is in fileed

A15cecal:
Got it, checked in the access tab. But yes, then it looks alright. Good job!

Issue #3750, opened 2017-05-10
Diagram layers


HGustavs:
Much like the layers in adobe illustratior I would like layer support (but simpler than illustrator that sometimes does not work).

* Each diagram object should have a layer (which is a number) ... eg Layer 1, 2, 3, 4, ...
* Current drawing operation always is in a specific layer
* Each layer has a spcific default coloring e.g. if an er entity is created in layter 2 it is green by default, if it moves ot layer 3, it should keep its color, but if we draw another entity in layer 3 it hould be for instance orange.
* It should be possible to restrict the diagram display to display only some layers or all layers

o18marka:
Tried to develop a suitable solution, but failed to make anything stable enough to use. I'll have to leave this to a more knowledgeable developer.

a18antsv:
Maybe it would be good to create a new class called Layer. An instance of this class can have an array of objects for all objects in the array and also store properties that should be applied to objects added to that layer. A minimum of one layer is required. Then the diagram array could be an array of layers instead of an array of objects directly. The layers then store the objects.

a18antsv:
In the toolbar it should be possible to hide a layer, add a new layer and move a layer up or down for the order the objects should be drawn. Then default properties for each layer should be possible to set from the global appearance menu. Here a collapsible element, introduced recently could be used for "Layers" then another collapsible for each layer could appear after opening the layer collapsible.

a18petha:
some interesting proposal, although I don't think this would be a good implementation "Then the diagram array could be an array of layers instead of an array of objects directly" issue here is that some functions requires the objects to be at cretin diagram index values to work. Changing this would require some adaptive work for system to be compatible.

a18petha:
to achieve so objects are place in orders compare to layers all that is needed is a forloop that write's the diagram in accordance to layers. Only negative to this is obviously the decrees in performance from running a forloop on the main drawing function.

PR #3978, opened 2017-05-17
PC mode - Issue#3743

a15chrau:
This issue is not yet finished, since it requires a solution for mobile.

Implements Issue#3743 for PCs. Tried to maintain the cohesive design, but does have some minor differences that will be changed at a later date.

It is worth noting that some buttons don't have a place within the toolbox yet, such as the delete and change appearance functions. The delete function, alongside the change appearance function, will likely be moved within the layers toolbox. See issue #3750 for the issue tackling layers.

I suggest moving the global change appearance function into the edit drop down menu.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -559,11 +559,10 @@ function lineMode() {
     uimode = "CreateLine";
 }
 
-function figureMode() {
+function figureMode(mode) {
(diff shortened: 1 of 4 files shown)


a14davlo:
Will need to make some adjustments in my issue, but it should be fine. I approve.

PR #3983, opened 2017-05-17
Added some titles to various buttons and fields in Section View

a15eriho:


d15eriso:
Looks fine! Good job!

Issue #3748, opened 2017-05-10
Diagram export to svg button / Export to JSON button


HGustavs:
We need a button to download the document as svg or download the current json as a file.

a15admsa:
@HGustavs Does it have to be SVG? If yes; why didn't we make the diagram in SVG instead of Canvas from the first place? All I can find on the internet are different libraries and how to export canvas to svg with them. I have made it possible to export in .jpg. If I want to make it work with svg, I have to use a extended library.

a15admsa:
@HGustavs I asked the question on Stackoverflow and got these answers;
[http://stackoverflow.com/questions/44020028/export-canvas-to-svg-without-library]

a15admsa:
This issue is not yet fixed!

a15admsa:
This issue will have to be fixed at a later date. @HGustavs have made a new issue with some exemple code on how to fix this one. That exemple code is in issue #4033

c16jonha:
This issue will be fully solved through Issue #4033

PR #3985, opened 2017-05-17
Part fix for issue #3748

a15admsa:
I've part-fixed the issue. You can export both JSON and a picture right now. The thing is that the picture is exported in JPG instead of SVG. See comments in issue #3748 for the reason why.


b15matvi:
Tried it, Looks good!

Issue #3941, opened 2017-05-16
Dialog option - Change the lines of the object.


b15matki:
When double-clicking it is not possible to alter the border style of the objects. This needs to be fixed!

a15perbl:
Issue resolved

PR #3992, opened 2017-05-17
#3941

a14pethu:
There are now a dialog option for changing the line colors on entities.


a15chrau:
Seems good, works on Pethu's branch. Shouldn't break anything.

PR #4010, opened 2017-05-18
Made it so the magic header appears earlier.

c14emian:
Making it appear even earlier than this will make it so both the normal
and the magic header is seen at the same time.


b14marki:
Looks good :)

Issue #3745, opened 2017-05-10
Align support in diagram


HGustavs:
If we select more than one symbol we sould be able to do aligment like in illsutrator.
![screenshot 2017-05-10 14 43 34](https://cloud.githubusercontent.com/assets/6978801/25899166/1dcdee00-358f-11e7-9192-f9e671408059.png)

a15perbl:
Issue resolved!

a15chrau:
Found a minor code error that should produce a bug. It'll be hotfixed in the implementation of #4024 

Instead of iterating through selected objects for the align, it iterates through diagram objects, only aligning the first objects within the diagram array, no matter what objects you select.

PR #4025, opened 2017-05-18
Issue #3745

a15chrau:
Implements top, right, bottom and left align, as well as vertical and horizontal align.

These options are added to an additional drop down menu, titled "Align". This is possibly a temporary solution.

Does not work for figures.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -948,7 +948,80 @@ function setRefreshTime() {
     }
 }
 
+function align(mode){
+    var selected_objects = [];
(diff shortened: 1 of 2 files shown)


a15chrau:
I forgot to mention that this also puts the global appearance button within the drop down menu labeled "Edit".

a15ponja:
Closed due to unfinished before merge.

a15chrau:
Re-opened per instruction

b15matvi:
Tested it out, appears to be working.

Issue #1887, opened 2016-02-25
Loading animation for long loading times


a97marbr:
It would be nice to get some kind of feedback when a page is waiting for data.

HGustavs:
We used to have a throbber. 

Perhaps re-introducing a throbber gif for the code viewer and showdoc / showdugga?

b14marki:
The picture is available but it needs to be an animation.

a17fellu:
Added the animation to codeviewer.php in branch: https://github.com/HGustavs/LenaSYS/tree/GruppC2019_1887fix

Will update the other pages as well and then make a pull request

a17fellu:
This could be introduced on all of our pages. Seems a little inconsistent to only have it on a few. I feel like ether we only add it to the codeviewer only or the entire site?

a17fellu:
Pull request: https://github.com/HGustavs/LenaSYS/pull/5829

a17andby:
#5867

PR #4026, opened 2017-05-18
#1887 Loading icon

b14marki:
Have made a svg picture that I will use for animation. It will rotate
and be smaller then used up in the left corner and will appear when a
page is loaded for a long time.


b14marki:
printanimering
Printscreen svg picture.

a14phili:
Looks fine

PR #4038, opened 2017-05-19
Week20 #3624 d15johro

d15johro:
giving odd and even rows in table with id testTable different background colors. removing old css for odd even tablerows. setting thead to table in table rendering.

issue #3624


a15wilhe:
Also the css class fumo should have the background color (for the files view).

PR #4049, opened 2017-05-20
Added code to add repective class to magic heading rows.

c14emian:
Added code that adds the class "hi" or "lo" to the magic heading rows
creating a "striped" pattern in the table.

### DIFF
--- a/DuggaSys/resulted.js
+++ b/DuggaSys/resulted.js
@@ -98,7 +98,7 @@ function redrawtable()
         show=true;
       }
       var strx = "";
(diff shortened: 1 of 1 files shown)


c14emian:
The inline if statement is used when displaying the rows in general, not just in the magic headings (row 101 and 209 in resulted.js).  Would it be better to merge this and then create a new issue requesting the if statement to be changed or close this one and make a new issue concerning both this one and the changes of the if statement?

a15eriho:
Will merge this, and close the issue.

Issue #3661, opened 2017-05-09
Magic heading in the result-view doesn't appear correctly


a15mikla:
The magic heading should be visible as soon as the original header dissapears. Now there is a point where it doesnt show up.

PR #4050, opened 2017-05-20
Changed offset value to make the magic headings appear earlier.

c14emian:
Change offset to +15 instead of -10 making the magic heading appear as
soon as the header text dissapears on the screen.


c14emian:
This issue (#3661) was solved prior to the last merge, not sure if something went wrong when merging or why it wasn't implemented in the new version.

Issue #3864, opened 2017-05-12
The text inside an object should always be inside the object.


b15matki:
At this moment it is possible for the text within an object to exist outside the object itself. An example could be seen below:

![textoutside](https://cloud.githubusercontent.com/assets/26705132/25997410/27ddd4d0-371c-11e7-9c5c-1062a46bace8.png)

The text should be within the object similar to this:

![text inside](https://cloud.githubusercontent.com/assets/26705132/25997464/64c2a650-371c-11e7-9d68-cbd6e5c69a5f.png)

This could be solved either by resizing the object or decreasing the size of the text.

HGustavs:
Clipping is the solution.

The text should be drawn with clipping enabled.

HGustavs:
Basically Like This
~~~
		// Draw Box Outline

		ctx.beginPath();		
		// Draw Box Outline (Again, same outline as alredy drawn but this time only the outline and it will be used for clipping only)
		ctx.closePath();	
		ctx.save();
		ctx.clip();
  		// Draw Text
   		ctx.restore();
~~~

Same code basically, for entity, relation, attribute, class etc. etc.

c16jonha:
This doesn't seem to be an issue anymore. I will therefore close this issue

PR #4067, opened 2017-05-22
Partial fix for issue #3864

b15matvi:
The text in attribute and entity will no longer go outside the borders. Partial fix for the issue. Still not fully fixed.


a15gunfu:
Looks good!

Issue #3208, opened 2017-04-21
Create an undo & and redo button


a14davlo:
In almost all editors you have buttons where you can take a step back and if you've stepped back, you can also take steps forward.

We currently have hash code that can be further developed for this.

b15matvi:
It would be preferred if it's possible to use ctrl + z for undo and ctrl + y to move forward aswell.

HGustavs:
Seems like a great way to do this by havig both a button and a key combination...

This can quite easily be implemented by keeping a history of the json data, or retrieving an old version of json data from server.

b15matvi:
Placing this issue on ice until @a14pethu is finished with the hashing function.

a15admsa:
I've made it possible to save the state you are at and to return to the previous state. The problem is that it won't redraw the diagram and the points. That is the only thing left in this issue.

PR #4071, opened 2017-05-22
Part-fix for issue #3208

a15admsa:
I've made it possible to save the state you are at and to return to the previous state. The problem is that it won't redraw the diagram and the points. That is the only thing left in this issue.

I have also made some fixes for the site overall. Some CSS-tricks. 

Issue #3208 will have to be put on hold.


b15matvi:
Looks good, unable to test but it should be possible to further develop.

Issue #4024, opened 2017-05-18
Distribution support


a15perbl:
Related to issue #3745, functionality for distribution support should be implemented. When selecting multiple objects, an even spacing should be created between them.

HGustavs:
Not super important to implement distribution, I would prioritize alignment, and when alignment works, focus on distribution.

a15perbl:
Functionality for alignment has already been implemented

a15chrau:
Would have been a good idea to inform me that this was the case.

a15perbl:
Issue resolved

PR #4074, opened 2017-05-22
Issue #4024

a15chrau:
Implements distribution functions for vertial and horizontal. This comes with a objects sort function, that sorts objects in a list by their x or y position. A global bubble sort has also been implemented.

A small bug fix for the alignment issue.


a15chrau:
Absentmindedly forgot that @HGustavs isn't a fan of switch cases, should I change this?

a15perbl:
Yes, it should preferably be changed if he's said that

PR #4079, opened 2017-05-22
For For 3361 Button Icon & Color

a14davlo:
This adds svg file icons to entity, line, attribute, relation and move arrows. The color on them is blue and can switch to orange if they're selected.


a14davlo:
Made a mistake, should not go to Master but go to G4 instead, closing it.

PR #4082, opened 2017-05-23
Diagram Svg Buttons & Color

a14davlo:
This includes svg files used for the buttons entity, line, relation, attribute and move arrow.

Also contains the blue color that can switch to orange when the button has been selected.

Known problem: Orange color doesn't disappear after one of the buttons functions been used. Frankly, don't have more time to develop on this and this needs to be merged in.


a15gunfu:
Looks good!

Issue #4091, opened 2017-05-23
Style the button in the "expiremessagebox" so that it's consistent with the "endsessionmessagebox".


b14larek:
Currently the 'OK' button in the expiremessagebox is to the left when its position is supposed to be in the center relative to the expiremessaebox div.

a15ramha:
This issue has been solved.

a15ramha:
This issue i solved now.

PR #4100, opened 2017-05-23
Issue 4091 a15cecal w21

A15cecal:


a15ramha:
This is the solution for #4091

Issue #4088, opened 2017-05-23
Missing icons for buttons in toolbar


a15perbl:
The buttons placed under the title "draw" in the toolbar currently has no icons. Given that the other buttons have svg images, icons for these should be made.

a15perbl:
Issue resolved

Issue #4085, opened 2017-05-23
Selected button should be unselected when clicked twice


a15perbl:
When a user selects a button and then clicks it again it is never unselected, unless they decide to click another button. It would be preferable if the selected button got unselected when clicked twice.

![bild](https://cloud.githubusercontent.com/assets/26705103/26343537/a75bebba-3f9c-11e7-918b-a406c6f9adee.png)

a15perbl:
Issue resolved

Issue #4084, opened 2017-05-23
Match appearance of buttons to LenaSYS general theme


a15perbl:
The blue color doesn't really match the rest of the website. The orange colour that appears when they're clicked is fine but other than that they could use some styling.

![bild](https://cloud.githubusercontent.com/assets/26705103/26343235/5dbfad44-3f9b-11e7-90ff-980cd3737d64.png)

a15perbl:
Issue resolved

Issue #4059, opened 2017-05-22
Add more colour options


a15perbl:
Some more colour options should be added to all the dropdown lists involving choice of colour.

a15perbl:
Issue resolved

PR #4103, opened 2017-05-23
Fix for issue #4084 + #4085 + #4088

a15admsa:
This pull request fixes all three issues. It:

* Matches the appearance of buttons to LenaSYS general theme
* Deselctes a button if the button was selected before
* Adds two new symbols to the toolbox

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -706,7 +706,7 @@ function relationMode() {
 }
 
 $(document).ready(function(){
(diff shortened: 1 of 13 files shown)


a15oskst:
Everything seems to work as expected, except that symbols change colours when pressing "ok" in the appearance dialog that appears when creating anything in the diagram. But this has been proved to happen in earlier versions of the code and is not related to this pull request. I tested code from different pull requests and found that it seems to be connected to the work done in issue #4059 and pull request #4090.

Issue #3849, opened 2017-05-12
Lock student names in results and groups pages


a15alica:
The student names in the tables on the results and groups page should be visible at all times. If you scroll horizontally (to view all moments) the names should stay at the same place. If you on the other hand scroll vertically (to view all students) the list should scroll like it does now.

HGustavs:
Currently the headings work correctly in your version ?

![screenshot 2017-05-12 13 15 43](https://cloud.githubusercontent.com/assets/6978801/25995997/35e0faf0-3715-11e7-9a2b-2349d25a6da9.png)

On the live-site it works for the headings that attach below the menu when scrolling out of view....

We want the same for the number column and the student name column when scrolling left-right.

Im thinking that the javascript can be adapted/copied/rewritten to also handle the number and name columns ...

a16linas:
This feature is already implemented, therefore closing the issue.

PR #4106, opened 2017-05-23
Added magic heading and magic column

a15osmla:
Student names and headings are now always displayed while scrolling vertically and horizontally on the group page. Fix for #3849


a15ponja:
Already merged.

Issue #4110, opened 2017-05-24
Move function drawOval to diagram_symbol.js


b15matgu:
The drawOval function can be moved to diagram_symbol.js. Also, the attributes has a linewidth of 2 and the standard width of symbols is 1.

a15perbl:
Issue resolved

PR #4111, opened 2017-05-24
Issue #4110 - Move function drawOval

b15matgu:
The function drawOval is moved to diagram_symbol.js. Also, the standard line thickness of the objects is changed to achieve consistency.


b15matvi:
Looks good!

Issue #4123, opened 2017-05-24
Make background colours lighter


a15perbl:
The current colours can be a little straining to the eye since they are pretty strong. If they were made lighter they would probably blend in smoother and also make all the text easier to read.

a15perbl:
Issue resolved

PR #4126, opened 2017-05-24
Issue #4123 - Make background colours lighter

b15matgu:
Background colours is now lighter.


b15matvi:
Looks good! Good work Mats!

Issue #3663, opened 2017-05-09
The course-moments should be visible in the magic heading, result-view


a15mikla:
In the result view the magic heading only shows the name of the dugga. It should also show the name of the course-moment, like the standard header that is visible when the magic heading isn't being used.

d15eriso:
Can you show a screenshot of what you mean? @a15mikla

a15mikla:
Without the magic heading you can see the course-moment(In this case, "Biträkningsduggor 1HP"):
<img width="459" alt="1" src="https://cloud.githubusercontent.com/assets/26705198/25988794/af0c68cc-36f9-11e7-9680-4e4aea540296.png">

With magic heading you can only see the name of the dugga:
<img width="323" alt="2" src="https://cloud.githubusercontent.com/assets/26705198/25988810/c0b02cc6-36f9-11e7-98e2-d41e8e373d68.png">

@d15eriso

d16linja:
Already solved

PR #4143, opened 2017-06-06
Added code to make entire header magic Issue #3663

c14emian:
Added course-moments to the magic heading in result view


b18alegu:
Closed as part of issue #7767

PR #4145, opened 2017-06-06
Fix searching including dropdown values

### LINKED ISSUE #4136: The search function results, also searches for the headings
So if you search for teacher it will not filter out anything since everybody has the header of teacher. If you search for a99marbr it will filter out everybody that does not fit that name.

a15matab:
Fixes #4136. Instead of including the values in the dropdown menu (Student, Teacher and None) in the search the ones that aren't selected are filtered out. This makes it possible to search e.g "teacher" and only the teachers will show up.

Also structured the access table with a \<thead\> and \<tbody\> instead of just \<tbody\>.


b18alegu:
Closed as part of issue #7767

Issue #4135, opened 2017-05-24
When adding a security question using profile page, you can not use it to reset password.


a15ponja:
This is probably due to the check in forgot password expects the password to be saved in lowercase, but the profile page saves it as how it was written ex Potato (how it was saved) will never be the same as potato (how it is expected).

c20adael:
@HGustavs Is this issue relevant as of now?

HGustavs:
I think we can mostly remove the profile page as It has no real use currently as far as i know.

Is it possible to check what profile page does  and where it is used and just remove it?

PR #4154, opened 2017-06-11
lcase validation

a15denbi:
komplettering_a15denbi_2
Forcing lowercase validation between the sql db and the php post function when inserting secret question under profile page. This cause more than one file forces lowercase post and lowercase return value so this fix will simply work. Also tweaking the installation data on the preinstalled profiles cause of the lowercase implementation.
Fixes Issue #4135


b18alegu:
Closed as part of issue #7767

PR #4155, opened 2017-06-11
Change default setting

### LINKED ISSUE #4146: Course members inconsistently displayed on "Results" / "Access"
The ordering is inconsistent between the two lists. The data is also inconsistently displayed.

I have no idea what Access is sorted on, but has separate first / last name fields.

Result can be sorted but is right now sorted on First name.

d15eriso:
The list will now allways be sorted by Last name unless the user has chosen something else. By changing the if-statement to set Last name instead of First name when local storage is empty.
Fixes #4146


b18alegu:
Closed as part of issue #7767

PR #4158, opened 2017-06-16
Added magic headings in accessed

c14emian:
Added a div that holds a magic heading, the div is only visible upon
scrolling down and is stuck to the top of the screen.


b18alegu:
Closed as part of issue #7767

PR #4236, opened 2018-03-28
Issue 4176

c16matan:
Fully implemented issue 4176


c16matan:
Implemented sum at bottom of each column on the resulted.php page.

Issue #4211, opened 2018-03-28
Access editor dropdown - part one, make dropdown


a16seplu:
The course version is currently edited/added using a text box (AFAIK).<--- check this first 

We want an editable table dropdown for this functionality.

PR #4239, opened 2018-03-29
fixed issue #4211

b16filor:
Accessed.js row 272
Added dropdown menu for version field.


c16matan:
Remove empty function before merge

Issue #4249, opened 2018-04-03
The buttons for objects has to be pressed for drawing every object


c16jonha:
When one object is selected, the user should be able to draw that object several times without having to press the objects button again. When the objects button is pressed again, then the object is de-selected. 

You should also be able to name the objects after you've drawn them with the object settings dialouge window.

This issue is a sub-issue of issue #4200

HGustavs:
This is a prioritized issue from my standpoint. If this is not fixed the diagrams are nearly useless as the process of making a diagram is VERY slow without it.

c16jonha:
This issue was solved and will now be closed

Issue #4104, opened 2017-05-23
When button is unselected, it should return to normal state


a15admsa:
If you press a button twice, the color will change on the button. This is to indicate if a button is pressed or not. When you unpress a button, the mode is not returning to normal. The user has to draw the figure for it to return to normal. This should be fixed.

a16rasja:
You can unselect the button by pressing it again before you have drawn something. You can't unselect it by pressing it again if you have drawn something

c16jonha:
This issue is now solved since issue #4249 was solved

PR #4257, opened 2018-04-03
G2 a16rasja 4104

a16rasja:
Fixed so the user can unselect the button by pressing the button again, and mode is being set to normal.
Reffering to #4249 and #4104


c16jonha:
Remove the lines of code that were commented out, so that the code will be more clean

Issue #4188, opened 2018-03-21
Group editor


HGustavs:
The group editor functionality and section editor button should both be removed.

Group editor was to complicated to work well issue #3900 - #3901 ... 

New improved simplified suggestion in #4170 among others.

b16karhe:
What do u mean? Weren't we supposed to create functionality for group editor and handle problems with the section editor? And here it says to delete it.

b16karhe:
Question has been answered.

PR #4288, opened 2018-04-04
Fix for issue #4188 to remove the old solution to groups.

b16karhe:
Deleted the button to reach the groups and deleted the files to handle that part of the site: grouped.php, grouped.js and groupedservice.php.


a16filol:
Ser bra ut!

PR #4291, opened 2018-04-04
Issue 4186

c16matan:


c16matan:
Solved and implemented issue 4186. It should work as long as the useranswer in the database is null.

c16matan:
Merged a issue-4298 (a sub issue of this one) into this branch which is also to be reviewed.

Issue #4228, opened 2018-03-28
Mini mode (blue mushroom mode) - part two


a16seplu:
Render all marking buttons inoperative, this is a viewing mode only.

PR #4297, opened 2018-04-04
Hide marking grades in mini mode

b16nilpe:
Issue #4228 
Hiding the images for grades/changing grades when mini mode is checked in the filter menu.


a16jonis:
Seems to be working fine :)))))). Merging this.

PR #4304, opened 2018-04-04
G2 - a16rasja and c16carbe - 4280

a16rasja:
Added a button for minimizing the toolbar.
You are able to press in the vicinity of the button

chrome_2018-04-04_16-29-11


c16jonha:
Remove the white space under the toolbar title when it's minimized. Maybe add some animations as well.

Issue #4147, opened 2017-06-08
Formating in result-view dropdown-menu


d15eriso:
Separating lines are inconsistent. Should be using same format for all parts of the dropdown menus.

d15eriso:
![image](https://user-images.githubusercontent.com/26739514/26929938-e1be91ec-4c5b-11e7-9815-9adb03c56326.png)

a16seplu:
Solved

a16seplu:
Solved

PR #4315, opened 2018-04-05
Fixed separating lines

b16nilpe:
Issue #4147 
Removed extra borders and fixed the sort button showing in the filter menu.


a16katba:
Works good !

PR #4316, opened 2018-04-05
G2 c16linst 3825

c16linst:
It works to select multiple objects with CTRL/CMD, but not together with the 'dotted selector'. When a selection has been made with the 'dotted selector' and another object is selected with CTRL/CMD, the previously selected objects are deselected. Working on a solution.

It isn't possible to deselect an object with the CTRL/CMD button pressed either. It seems that clicks on already selected objects aren't recognized.

Some new issues could be:
- Integrating 'dotted selector' selection with CTRL/CMD selection
- Making it possible to deselect single objects with CTRL/CMD

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -79,6 +79,7 @@ var classTemplate = {
   height: 7 * gridSize
 };
 var a = [], b = [], c = [];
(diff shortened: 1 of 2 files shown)


c16jonha:
Fix the conflict in the diagram_mouse.js or this can't be merged

Issue #4330, opened 2018-04-06
SectionEd: Cannot remove list entry


a15magkn:
A bug in the section editor is that when you click the trash-can icon to remove an entry a window pops up asking if you want to remove the item, but even if you click "ok" the entry remains and is not removed.

PR #4331, opened 2018-04-06
#4330 - SectionEd: Cannot remove list entry

a16filol:
#4330 Was not able to remove listentries.


b16linra:
Found an issue where it is not always possible to remove moment list entries.

Issue #4325, opened 2018-04-05
Refactorizing Diagram code


c16jonha:
We have made a list of suggestions that we think would improve the readability of the code and the codes structure in general.

- Remove redundant code
- Merge together files such as Diagram_figure and Diagram_symbol into one file. The same for the files in the forms folder.
- Renaming or documenting variable names so that they become easier to understand.

If possible we would like to see the original code for the diagram files to get a better understanding of the codes origin. 

Any feedback @HGustavs ?

HGustavs:
I found the seed code by going to the start of last years project in the releases tab. This should be the earliest available.

https://github.com/HGustavs/LenaSYS/tree/382dc230b14463b55a0a4775c78f64ecd9dcc5a5/DuggaSys

HGustavs:
They did a very big suggested renaming of variables which in some cases made the code more difficult to read and process.
 
This "refactoring" was done and documented in two text files...

~~~
-----------------------------------------------------------------
                ##List of changed variable names##
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

#
# This was done to follow the javascript code
# standard
#

Old Name                        New Name
-----------------------------------------------------------------
private_lines                   privateLines
dist                            distance
ind                             index
xk                              xCoordinate
yk                              yCoordinate
dx                              deltaX
dy                              deltaY
dd                              hypotenuseElevatedBy2
selval                          isSelected
ctx                             canvasContext
sx                              startX
sy                              startY
ex                              endX
ey                              endY
x1                              topLeftX
y1                              topLeftY
x2                              bottomRightX
y2                              bottomRightY
px                              width
py                              height
len                             boxHypotenuseElevatedBy2
centerpoint                     centerPoint
connected_objects               
private_points                  privatePoints
selobj                          lastSelectedObject
ev                              event
private_objects                 privateObjects
object_points                   objectPoints
current_hash                    currentHash
diagram_names                   diagramNames
lastDiagramedit                 lastDiagramEdit
timediffrence                   timeDiffrence
last_edit                       lastEdit
line_width                      lineWidth
cx                              currentMouseCoordinateX
cy                              currentMouseCoordinateY
sx                              startMouseCoordinateX
sy                              startMouseCoordinateY
mox                             oldMouseCoordinateX
moy oldMouseCoordinateY-----------------------------------------------------------------
                ##List of changed variable names##
[...]

PR #4365, opened 2018-04-09
Removed arity-code

b16petma:
Arity code that wasn't used is commented out. Part of Issue #4325


b16petma:
Done with moving code to the Slask

PR #4368, opened 2018-04-09
G2 c16carbe refactorization #4325

c16carbe:
* Re-wrote the part where mode is set (it now only has one function instead of 5(?))
* Changed how the setMode() function is called when a button is pressed.
* Changed setType() so that it no longer gets an item for every check.
* Removed all checks because they are not needed.
* Moved two functions (keyup/keydown) and began implementing them at the already existing keyDown for future use.

A part of Issue #4325


c16jonha:
This wasn't branched from the latest G2V02-2018, but this was fixed in #4372

PR #4383, opened 2018-04-09
G1 issue 4360

b16filor:
Fixed bug with feedback box in swimlane sometimes flashing and made the box appear on the left if theres not enough space on the right.


f16linli:
It looks good

Issue #2820, opened 2017-04-06
Add a barrier for wrong username inputs (password reset)


a15tobli:
Part of #2810.
When applying for a password reset the user is prompted to enter their username. Add a barrier for a maximum amount of wrong inputs, which prevents unauthorized visitors to keep guessing usernames until one checks out.

a16seplu:
Solved

PR #4387, opened 2018-04-10
A barrier has been added in password reset requests

f16linli:
A barrier has been added to prohibit someone from requesting to many password resets on diffrent user accounts within a 5 min timeframe.

The current limit is set to 5 attempts before the lockout period begins.


d16linja:
issue #2820

Issue #4341, opened 2018-04-06
Migration: Create the parser


a15magkn:
We need to write a parser that parses the Json-tables created in issue #4340.

The parser should accept the Json and turn it into an SQL-string.
For example the following Json should be parsed into the SQL-format shown bellow.
## Json Example
```
{
    ‘create’, 
    ‘users’, 
    ‘columns’: {
	 {‘uid’, ‘int’, ‘UNSIGNED NOT NULL AUTO_INCREMENT’}
         {‘name’, ‘varchar(30)’, ‘NULL’}
         {‘email’, ‘varchar(30)’, ‘NOT NULL’}
    },
    ‘Version’ => ‘v0.10’
}
```
## Desired SQL-format
```
CREATE TABLE users(  
    uid INT UNSIGNED NOT NULL AUTO_INCREMENT,  
    name varchar(30) NULL,  
    email varchar(30) NOTNULL,  
);
 ```

HGustavs:
Or an even simpler more general way.

One that just creates the table

And the columns are added one by one separately.

This way, adding columns or adding tables use the same code for each column.

So there would instead be
[
{ ‘Version’:‘v0.10’,[
{‘create’, ‘users’},
{'column','users',‘uid’, ‘int’, ‘UNSIGNED NOT NULL AUTO_INCREMENT’},
{'column','users',‘name’, ‘varchar(30)’, ‘NULL’},
{'column','users',‘email’, ‘varchar(30)’, ‘NOT NULL’}
]
},
{ ‘Version’:‘v0.11’,[
{‘create’, ‘student’},
{'column',,'student'‘uid’, ‘int’, ‘UNSIGNED NOT NULL AUTO_INCREMENT’},
]
},
,
{ ‘Version’:‘v0.12’,[
{'column','greger'‘uid’, ‘int’, ‘UNSIGNED NOT NULL AUTO_INCREMENT’},
]
}
]

HGustavs:
~~~
{"insert","users","ssn","20070401-5566","name",'greger'}
~~~

~~~
INSERT INTO USERS(SSN,NAME) VALUES('20070401-5566','greger');
~~~

HGustavs:
Also, the create/alter example above works even if the table has already been created whereas the "complete create" example works only if users table does not already exist.

a16filol:
You are not able to create an empty table. Im thinking of a workaround or do you have any suggestions on how to solve that problem.

HGustavs:
[...]

PR #4408, opened 2018-04-10
#4341 - Added migration functionality with versioning.

a16filol:
#4341  - Parser to parse data from array


a14isshi:
You will not be able to create a table for a M-M relation with two columns.

Issue #2436, opened 2016-12-21
Using a direct link to a page that requires login is shown to guest


a97marbr:
There does not seem to be a proper credentials check for showdoc.

Using a direct link to a page that requires login is served by showdoc to guest

d16linja:
On what example dugga can i see if the changes work or not?

d16linja:
Already solved

d16linja:
Realized what the problem was, reopened the issue

c16matan:
The implemented change was not good from a continuous development perspective, so that merge got reverted. @HGustavs I did not find any other data in $_COOKIE other than the PHPSESSID variable, any other idea of how to solve this? As you mentioned we can do a proper course credentials check, but currently i don't know how this could be implemented to also check for each moment without the actual moment identifier.

c16matan:
Approved verbally from @HGustavs. Closing.

PR #4460, opened 2018-04-12
Working permissions for showdoc and showDugga

d16linja:
Pair programmed with c16matan

Issue #2436 

To test it you need to change permissions via mySQL for the dugga and the course, as there are not templates for the tests in the example data you can not change the visibility in the course/dugga editor, if you try to do this you will remove the test. 
The permission int's is pretty stupidly defined like this:
0 = hidden
1 = public
2 = private (requires login and access to the course)

### DIFF
--- a/DuggaSys/fileed.js
+++ b/DuggaSys/fileed.js
@@ -1,11 +1,11 @@
 /********************************************************************************
-   Documentation 
+   Documentation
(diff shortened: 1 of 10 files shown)


d16linja:
Fixed the link for all the template duggas so they work with the new credentials check

Issue #1783, opened 2015-07-17
Navigating to showdoc produces unk in return link


HGustavs:
so...
I make a link to a document without supplying a course version.... i.e. showdoc?courseid=3&filename=?foo.md

when I click the return button I am returned to the right course but the version is "unk" which is true as no verison was given to showdoc.... however....  adding a version locks the file link to a certain version, which is not the intended behavior either.

A better way is to use the newest version "live" version if no version is provided. This way we can get version-unlocked links to showdoc and still supporting the back button in a user-transparent manner.

The only time when this back-fires is if a teacher is editing a future version and clicks the back button, he or she would then return to a version that is not the same as the originating page.

a17andby:
Fixed and merged. Closing.

PR #4501, opened 2018-04-13
Showdoc now produces a return link to the active course version

b16nilpe:
Issue #1783 
If no course version is provided when navigating to showdoc the return link now links to the active version of the course instead of UNK.

@HGustavs 
Is it OK to make a query in navheader.php to solve this, since there is no suitable service file?


b16nilpe:
Should be implemented some other way, this could potentially produce an unintended behavior of the button.

Issue #4505, opened 2018-04-13
DuggaEd: Translate data into readable values


a15magkn:
Currently in the duggaed, a lot of fields are filled with a value such as 0, 1 etc. These should be translated to something more appropriate depending on the column. 
An example could be "GradeSys" that could have the values, U, G, VG rather than 0,1, 2.

![issa till magnu](https://user-images.githubusercontent.com/37794262/38726279-ffc85bd6-3f09-11e8-9f9c-d403b1d178e0.PNG)

PR #4511, opened 2018-04-13
#4505 DuggaEd: Translate data into readable values

a14isshi:
Changed variablename inside duggaed.js and duggaedservice.php to match the variablenames inside the database.

Both autograde and gradesystem now show the actual value and not an integer.

#4505


HGustavs:
The important bit here is to not change the database. Translations after the database are allowed. If we change the database we may break older versions of the software that uses the same data.

Issue #4486, opened 2018-04-13
Fileed add button


c16jonha:
Right now the add functionality is missing because it was placed among the headers. This functionality should be added once again. It should however be assigned to a button that's seperate from the table headers.

PR #4523, opened 2018-04-13
G2 c16linst 4486

c16linst:
Added two add-buttons - one to add a link and one to add a file. These are located outside the table, but are working like they did before the sortabletable update. There is one problem though - you can't choose whether you are uploading a global or local file. Maybe this could be added as a new issue?

This solves issue #4486


HGustavs:
Screenshot!

c16linst:
skarmavbild 2018-04-16 kl 08 35 11

Issue #4518, opened 2018-04-13
Migration: Make view to choose version


a15magkn:
In order to pick which version of the migration you want to access, this functionality has to be created.

a16linas:
There is now a dropdown box to select version for migration.

PR #4534, opened 2018-04-15
#4518 - Make view to choose version

a16filol:
#4518 - Made a view to select version and execute the migration that requires superuser priviliges.


a16filol:
This pull request is not correct multiple commits that are not part of the request is in it. Will be checking it out later today.

PR #4558, opened 2018-04-16
Remaining "group functionality"

a16linas:
Found one variabel that seems to be linked with the group functionality that was previously removed.


a16linas:
Added a comment on line 800 in sectioned.js where a variable that's group-related is.

Issue #4527, opened 2018-04-13
Fileed upload global or local file


c16jonha:
As seen in #4523 the add button for adding files should give you the option of uploading a global or local file.
![filechoice](https://user-images.githubusercontent.com/37794047/38793155-4f4795f8-4150-11e8-8431-b09550dc2ed0.PNG)

HGustavs:
As I also said in the related issue, screenshot!

PR #4626, opened 2018-04-17
G2 c16linst 4527

c16linst:
Added functionality to choose which kind the uploaded file will be. This is done with radiobuttons. A function is added to determine which radiobutton is clicked. The old functionality which stores the files in the database is left as it was.

skarmavbild 2018-04-17 kl 16 56 29


c16linst:
This solves issue #4527

PR #4641, opened 2018-04-19
Changed dropdown box width and added comments

a16linas:
The dropdown boxes are now 195px and a comment has been added about how the width might needs to be changed if future changes to the span are made.


a14oliri:
Works when editing items, but not when creating new items. 


screenshot from 2018-04-19 09-58-15

Is this intended?

b16linra:
The issue that @a14oliri mentioned has been present before. Last time it was caused by an unclosed div-element, but that might not be the case this time.

Issue #4644, opened 2018-04-19
SectionEd: Markable FAB


a15magkn:
At the moment the text in the FAB-button ("+") is markable. This should not be the case.

## This should not be able to happen
![textfab](https://user-images.githubusercontent.com/37794262/38977052-6c3f5212-43b3-11e8-9c8c-7c31e8bc32ca.JPG)

PR #4650, opened 2018-04-19
#4644 - SectionEd: Markable FAB

a16konka:
#4644

Fixed an issue where you could select the "+" character on the FAB button.


a16thegu:
Looks good.

Issue #4525, opened 2018-04-13
More example duggor


a16seplu:
Create more duggor that can be used for testing purposes

c16matan:
Implemented last week. Closing.

c16matan:
I don't know if i missed click or something, but this is implemented. Closing

d16linja:
#4282 
With this issue we fixed the old duggas so that they fit the new format and show up under the "test" page in LenaSYS. We also added deadlines so that the issues with the time could be worked with more easily, and added the new duggor as listentries so that the show up in a course

c16matan:
Implemented. Closing.

PR #4651, opened 2018-04-19
Reimplemented changes that somehow didn't get implemented in the merge

d16linja:
Issue #4525 
The vers makes the duggor appear in the "tests" window, and also makes it possible to edit the old example duggor without accidentally removing them. The changed deadline makes the deadline time show up in the dugga view so that it can be worked with easier.


d16linja:
The newly merged rows 224-227 needs to be updated, there is a duplicate of pos 15 with vers 97731

c16matan:
Implemented. Closing.

Issue #4479, opened 2018-04-13
Actually removing points


c16jonha:
At the current state of the diagram, no points are actually removed from the diagram. The points are just set to an empty string. While this is a way to "remove" or, more accurately, hide objects from the diagram, it's a shoddy solution because the points are still visible in the json-file as empty strings. 

The points should be removed completely instead.

b16petma:
I have made a bit of progress, but indices of points when they are shared like when a line is connected to another symbol, the indicies gets messed up. 

The solution right now is that when a point is removed from the canvas we remove it from the points-array and afterwards decrement point-indices higher then the index of the removed point.

Example:
points[] = 0 1 2 3 4 5, we remove 3.
points[] = 0 1 2 4 5 the symbol with index to point 4 and 5 now should use index 3 and 4.
Symbol with point 0, 1, 2 are not effected because they have the same index.

When symbols are connected we use the removePointFromConnector() function and I added that it returns a boolean if we succeeded to remove a point or not. This is used to see if a point is shared between symbols and we should not remove it.

Right now we get symbols that have the same index for two different points and we get indices that are outside the points[] length.

The solution is good for when symbols don't share points.

Branch connected to this issue is G2-b16petma-4479
Contributors are b16petma and a16timsc

b16andka:
This will be closed. The points is not removed completely unless the user press "Clear Diagram". However the points are reused in order to keep the JSON file clean. This is the most we could do without changing too much of the existing code.

PR #4653, opened 2018-04-19
G2-b16petma-4479 Clear Diagram clears the point array

b16petma:
When we clear the diagram with clear diagram from the menu, we now clear the points array.
This is a partial fix for issue #4479.
We limit the JSON-files a bit more now.


b16petma:
The issue with diagramNames was that the save function never cleared the array containing constructor names. So it contained old information at indices above current diagram.length-1. This was exported but never used due to the loops uses diagram[] length. This is now fixed!

PR #4662, opened 2018-04-19
Reworked credentials check for issue 2436

d16linja:
issue #2436, Reworked #4460 
Pair programmed with @c16matan 

Instead of using $_GET we now use session variables set in the navheader so that old templates dont need updates to work. 

See the old pull request #4460 for some information for testing

### DIFF
--- a/DuggaSys/showdoc.php
+++ b/DuggaSys/showdoc.php
@@ -359,11 +359,6 @@
 		}else{
 			$userid="UNK";
 		}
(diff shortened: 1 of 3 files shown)


d16linja:
@HGustavs Is this solution acceptable?

d16linja:
Reworked again to a new soulotion #4683 which should work with all versions of LenaSYS and all places that showdoc is used, and also only change one file.

Issue #4643, opened 2018-04-19
Add markdown preview to fileed


a16juleh:
Currently the markdown preview is located in preview.php. We want the code to be located in fileed.php, fileed.js. The CSS in preview.php should be moved to /shared/style.css (Only the CSS related to the popup window should be moved)

The preview window should open when the pen-button is clicked on the fileed site. 

<img width="1080" alt="skarmavbild 2018-04-19 kl 09 16 22" src="https://user-images.githubusercontent.com/37794632/38976715-5ceabaf0-43b2-11e8-83f4-88e080799127.png">

PR #4711, opened 2018-04-20
G2 a16adala 4643

a16adala:
All code from preview.php are now in fileed.php. Some changes had to be made to js, css and html to make it work. Also the css have been moved from preview.php to style.css and all js from preview are now in markdown.js


a16juleh:
This solves issue #4643

Issue #4658, opened 2018-04-19
SectionEd: Unable to delete certain Headings


a15magkn:
At the moment you are unable to delete certain headings in the Section Editor (for example Geometri 2 Hp). 
When you attempt to delete this you get a error-message. 
This needs some looking into.

![errorupdatingentries](https://user-images.githubusercontent.com/37794262/38981075-c27176c8-43be-11e8-9f48-ca68ebfe11c9.JPG)

a16filol:
This is not an isssue. The issue should be to improve the error message. This message is getting reported because the heading has foreginkeys stopping the delete. those entries should not be deleted.

b16linra:
I and @d15marwi are currently working on a solution to improve the error message when it is not possible to delete items.

d15marwi:
We know about this and are currently working on a message for this. The picture shows an error from the database (we will use this information for our solution)
![image](https://user-images.githubusercontent.com/37793970/39118471-c65ed04a-46e9-11e8-89a3-3ea60bd7bed5.png)

d15marwi:
@HGustavs , what more would we need in the alert? This solution should work?
![image](https://user-images.githubusercontent.com/37793970/39124938-8d6ac97e-46fd-11e8-854b-8c475f9ea902.png)

PR #4757, opened 2018-04-23
#4658 SectionEd: Unable to delete certain Headings

d15marwi:
#4658 
We removed the second error message and created a new message with the PHP error code and message for this case(the problem was that the key for this moment is used in another table row). So the bug was not that the item couldnt be removed but that the alert was bad and shown twice. 

People who worked on this issue: @b16linra and @d15marwi.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -1372,7 +1372,6 @@ function returnedSection(data)
 		showCreateVersion();
 
 	}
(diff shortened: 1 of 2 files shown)


b16linra:
This solution only changes the error message when it is not possible to delete something. In the future, it might be preferable to change the message "Error when updating entries" when performing other actions as well, like inserting or updating.

Issue #4703, opened 2018-04-20
SectionEd: Enter-key = Yes/Finalize


a15magkn:
When you get a pop-up window such as deletion or creating a new item you can press the escape-key to exit, and it would be nice if the Enter-key worked as a OK/Yes/Submit shortcut.

![actuallheaderattributes](https://user-images.githubusercontent.com/37794262/39039748-5d215edc-4486-11e8-80d6-e60bcd5573db.JPG)

![areyousurewindow](https://user-images.githubusercontent.com/37794262/39039661-35393156-4486-11e8-8a27-90618e1f3b6b.JPG)

a14isshi:
Finished!

PR #4761, opened 2018-04-23
#4703 - SectionEd: Enter-key = Yes/Finalize

a14oliri:
#4703 

Added functionality so that enter-press will execute the same functions as pressing Yes/OK/SUbmit/Whatever. added this for both adding new items, editing items and deleting.


a14isshi:
Tested and is working!

Issue #4773, opened 2018-04-23
SectionEd: Pop-up closes when draging & releasing the mouse button (Chrome)


a16konka:
While draging & selecting a text and releasing the mouse button outside the pop-up box on Google Chrome, the browser registers the button release as a click, which then closes the pop-up window because the click is out of range. This should be fixed somehow so that the browser dosn't register the button release as a click. **This issue is not relevant on Firefox browsers.**

## Example
<img width="633" alt="skarmavbild 2018-04-23 kl 4 27 31 em" src="https://user-images.githubusercontent.com/37795608/39132919-41dc55ac-4713-11e8-997e-0a89280e23ac.png">

PR #4780, opened 2018-04-24
#4773 - SectionEd: Pop-up closes when draging & releasing the mouse button (Chrome) 

a14oliri:
#4773 

 Added new function on mousedown that checks if the current clicked element is a loginBox or not, saving this value and checking it when clicking outside the loginbox to see if it should be closed or not


a16linas:
Solution worked as intended

Issue #4712, opened 2018-04-20
Sub-Issue: Submit-button functionallity from #4706


b16karhe:
A divided part of issue #4706.

a16linas:
The submit button works now

PR #4814, opened 2018-04-24
#4712 - Sub-Issue: Submit-button functionallity from

a16filol:
#4712 - Sub-Issue: Submit-button functionallity from. 

Fix the issue so we can post variants.


a16linas:
The submit button works now

PR #4837, opened 2018-04-26
#4833 - SectionEd: Message option doesn't have "tabs" 

a14oliri:
#4833 
 Added so the tabs-box gets shown when creting and editing message-items, also added so it shows tabs when getting created.


a14isshi:
The tabs works as intended, I can thought see a problem that needs to be fixed in another other issue: 

image

Here I created a new Header and added 2 tabs, the tabs does not contain the same color as the header.

I'm closing this issue and merging it as I find that this issue has been completed.

Issue #4741, opened 2018-04-23
DuggaED: Variants window should display dugga name


a16konka:
At the moment the variants window for a dugga doesn't display the dugga name that the variant belong to. The idea was to add this posibility in order to make it easier for the user to see which dugga the variants window belongs to.

## Variants window
<img width="1440" alt="skarmavbild 2018-04-23 kl 10 30 30 fm" src="https://user-images.githubusercontent.com/37795608/39115393-d47a210a-46e1-11e8-8b32-410e6b63765f.png">

HGustavs:
Perhaps by change the popup title from "edit variant" to "dugga name". This way we can get that withoug cluttering things up too much.

Im not super-keen on the look of the headings already... no spacing between headings in table and the edit variant popup header. We dont want _more_ stuff there already.

We want to fit as many variants as humanly possible in view at once.

b16karhe:
It will display the owning dugga through a number.

b16linra:
This was solved in pull request 4859. This issue will be closed.

PR #4859, opened 2018-04-26
#4741 DuggaED: Variants window should display dugga name

b16karhe:
#4741 Removed som code and made sure that fitting titles are displayed for the different edeting window. The duggas will be referd to like "1->XXX".


b16karhe:
fix

Fixed it so that it now displayes names.

Issue #4715, opened 2018-04-20
Sub-Issue: Enable/Disable-button functionallity from #4706


b16karhe:
A divided part of issue #4706.

a14isshi:
Works as intended!

Issue #4706, opened 2018-04-20
DuggaEd: Edit Variant, Disable- and Submit-buttons needs functionality 


a15magkn:
In the currrent version of the Edit Variant menu the disable- and submit-buttons does nothing and this needs to be rectified. 

![image](https://user-images.githubusercontent.com/37794262/39041039-72d22638-4488-11e8-9ab7-e239410f56d6.png)

b16karhe:
_How disabled should look._
![Uploading disabled.png…]()

b16karhe:
The sub-issues are solved

PR #4873, opened 2018-04-26
#4715 Sub-Issue: Enable/Disable-button functionallity from #4706

b16karhe:
#4715

Disable/enable now works for both edeting and adding variants. The visual aspect however has not been implemented.

Made disabled values turn red


a14isshi:
Works as intended!

PR #4885, opened 2018-04-26
Search button is out of place in access

a16katba:
Now the solution has  a capsule-shaped border around the search box and  a looking glass icon.


d16linja:
#issue 4770

Issue #4880, opened 2018-04-26
SectionEd: All options when creating an item should be reverted


a16konka:
After a discussion with the client, all input options need to be visible when creating a new item.

## Current view
<img width="492" alt="skarmavbild 2018-04-26 kl 3 12 27 em" src="https://user-images.githubusercontent.com/37795608/39307870-4cd6d8f8-4964-11e8-89e8-1fc1fd745d10.png">

a16thegu:
Approved, looks good on both new and edit item!

PR #4893, opened 2018-04-27
#4880 - SectionEd: All options when creating an item should be reverted

a14oliri:
#4880 

Now always showing all boxes, whether they have any effect on the item or not, both when creating and editing.


a16thegu:
Approved, looks good on both new and edit item!

Issue #4839, opened 2018-04-26
SectionEd: Arrow icon need to be separated from the Ellipsis effect


a16konka:
When on mobile mode a lot of Moments have the ellipsis effect due to the length of the text. The problem here is that the ellipsis applied both on the title and on the show/hide arrow. This needs to be fixed somehow so that just the title is being affected by the ellipsis effect.

<img width="363" alt="skarmavbild 2018-04-26 kl 9 21 21 fm" src="https://user-images.githubusercontent.com/37795608/39291511-8aa23500-4933-11e8-9d15-fb31e5b33933.png">

a16thegu:
Approved, look very good now!

PR #4894, opened 2018-04-27
#4839 - Fix issue when the arrow is included in ellipsis. Also fixed indentet…

a16filol:
#4839 - Fix issue when the arrow is included in ellipsis. Also fixed indentetion in the files.

### DIFF
--- a/DuggaSys/duggaed.js
+++ b/DuggaSys/duggaed.js
@@ -2,9 +2,9 @@
 
    Globals
 
-*********************************************************************************/
-var sessionkind=0;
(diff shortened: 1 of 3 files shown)


a16thegu:
Cannot approve the solution. The ellipsis is not correct, the yellow moment arrow should push the text behind it to ellipsis and the edit and delete buttons should not overlap the text.

skarmavbild 2018-04-27 kl 13 34 27

a16thegu:
Approved, look very good now!

Issue #4732, opened 2018-04-23
DuggaED: Update variant table after changes have been made


a16konka:
This relates to #4706.

#4716 (Sub-issue)  At the moment when changing a variant entry in a dugga (such as editing or deleting), the table doesn't update until the variant window is closed and reopened. That should be handled so that the table updates at the same time as the change has happened.

HGustavs:
The preferred way to do this kind of stuff is to refresh the table completely e.g. retrieve the information again from the database.

If this operation is very slow e.g. the marking view we change the on-screen data and then refresh, in all other cases we can safely just referesh e.g. the same operation that we do when the page is first loaded.

a16thegu:
Approved, the solution works.

PR #4911, opened 2018-04-27
#4732 - Fix issue when updating variant.

a16filol:
#4732 - Fix issue when updating variant.


a16thegu:
Approved, the solution works.

Issue #4842, opened 2018-04-26
SectionEd: Editing a test while no links available


a16konka:
At the moment it is possible to edit a test without any links available. This is possible when creating a test while links available, deleting links afterwards, doing back to the created test. When trying to edit the dugga the is no message that the are no links available to choose and it is still possible to save the test. This needs to be fixed so that you can either edit the test but not be able to save and the option to delete, or some other option that prevents the user to be able to save the test.

## Editing a test without no links
<img width="1440" alt="skarmavbild 2018-04-26 kl 9 31 45 fm" src="https://user-images.githubusercontent.com/37795608/39292092-670f9040-4935-11e8-8e49-a29499a2690b.png">

a16thegu:
Approved solution.

PR #4917, opened 2018-04-27
#4842 - SectionEd: Editing a test while no links available

a16linas:
added the validate type functon in order to error check the edit/new item box


a16linas:
added the validate type functon in order to error check the edit/new item box

a16thegu:
Approved.

Issue #4928, opened 2018-05-02
Sorting with radio buttons checks other columns than kind


a16juleh:
When we press one of the radio buttons to filter the table, a search for the corresponding string is made (if we press "link", it will act as if we wrote "link" in the search bar). The problem here is that the search functionality goes through all the columns to find a matching string - the radio button filtering should only look at the column "kind".

<img width="736" alt="skarmavbild 2018-05-02 kl 09 25 15" src="https://user-images.githubusercontent.com/37794632/39510665-c128d9ec-4dea-11e8-8831-d69d2c590fb7.png">

PR #4938, opened 2018-05-02
G2 b16andka 4928

b16andka:
When using the radiobuttons to filter the table, only the correct filekinds are shown.
filekindfilter


a16juleh:
This solves issue #4928

Issue #4929, opened 2018-05-02
SectionEd: Items with invalid names can be created


a14oliri:
Items with invalid names can be created. 

## Steps to recreate this:
1. Click create new
2. Enter an invalid name(e.g. @¡@£$)
3. Select another Type
4. Submit button becomes clickable and a new item with an invalid name can be created

![screenshot from 2018-05-02 09-29-32](https://user-images.githubusercontent.com/37795562/39510821-6430705a-4deb-11e8-8296-d1534beab162.png)

a16linas:
Both types now need to be valid in order to submit.

PR #4944, opened 2018-05-02
#4929 -  SectionEd: Items with invalid names can be created

a14oliri:
#4929 

The save button should only be clicked now when both name and type is valid.


a14oliri:
Trying to merge into Master and not G3V06-2018, will remake.

Issue #4746, opened 2018-04-23
Sortable table support for counters


HGustavs:
In some views a line counter column which counts from 1 to number of lines in table is always the first column to the left.

It would be nice to have an option to add this as a non-sortable column.

we can add a showcounter boolean and it should be pretty easy to add this as a column in the code, and running this code if showcounter is true and augment the code with "rowcounter++" for each row of the table.

PR #4945, opened 2018-05-02
Fix for #4746

b16andka:
Moved the rowcounter into SortableTable.js. The variable is set to 0 each time the table is rendered. 
The variable is used and incremented in renderCell().
rownumber


a16juleh:
This is a fix for issue #4746 and solves the problem.

Issue #4862, opened 2018-04-26
DuggaEd: Mark the selected variant when editing


a16konka:
Currently when editing a variant in the variant list there is no clear indication on which variant is currently being edited. This should be fixed so that the user can clearly see which variant is currently being edited.

## Mock-up idea (Color variation could be different)
<img width="1181" alt="skarmavbild 2018-04-26 kl 11 02 44 fm" src="https://user-images.githubusercontent.com/37795608/39296687-dc192cb4-4941-11e8-9276-1eed560258b8.png">

b16karhe:
Done.

PR #4950, opened 2018-05-02
#4862 - DuggaEd: Mark the selected variant when editing

b16linra:
#4862 - Added a background color to the table row that is being edited.

People who worked on this issue: @b16linra.


b16linra:
There seems to be an issue, so the pull request will be temporarily closed.

PR #4952, opened 2018-05-02
#4929 - SectionEd: Items with invalid names can be created

a14oliri:
The save button should only be clicked now when both name and type is valid


a16linas:
The solution works.

Issue #4612, opened 2018-04-17
DuggaEd: Able to save even though the title has illegal characters


a15magkn:
When the name of a dugga/item contains illegal characters (such as ?,!, £) you should not be able to save.
This works as intended in the SectionEditor but at the moment you are able to save in the DuggaEd even if there are illegal characters. 

## When the name contains illegal chars. this button should be grayed out and un-clickable
![screenshot from 2018-04-17 13-55-04](https://user-images.githubusercontent.com/37794262/38868448-4fab1d88-4247-11e8-94fa-ddc999dd9562.png)

HGustavs:
Or the saved version should be "cleaned" from "weird" characters. This way you dont need to clean it up if you mis-press a key.

PR #4962, opened 2018-05-02
#4612 - DuggaEd: Able to save even though the title has illegal characters

a16linas:
Fixed the validatedugganame, added the error check and block submit button when invalid characters are present.


a16thegu:
Approved.

Issue #4956, opened 2018-05-02
SectionEd: Update generic code for FAB in sectioned.js


d15marwi:
We decided that we need to update the code for the diffrent alternatives of the FAB to create a more generic code that is the same all the way. The code needs to changed because some types have exception that need to be validated before created by a function.

a14isshi:
The pull-request for this issue has been accepted and therefore this issue is now closed.

Issue #4840, opened 2018-04-26
SectionEd: Tests can be created without any links


a16konka:
At the moment there is no posibility to create a test through the "+" button when selecting test while having no links. This is still possible when creating a quick test with the FAB button. This need to be fixed so that when pressing the Test button there should be a validation attached and only creating a test if there are any links available.

<img width="1440" alt="skarmavbild 2018-04-26 kl 9 31 45 fm" src="https://user-images.githubusercontent.com/37795608/39291870-a4037850-4934-11e8-93a4-cc1427c1dfe8.png">

PR #4967, opened 2018-05-02
#4956 - SectionEd: Update generic code for FAB in sectioned.js

a16thegu:
#4956 - a16thegu
Updated all sub-FAB in SectionEd (sectioned.js) to be like the solution in #4840, and changed the location of the Group Activity and Message buttons in the FAB to be in a incremental order.


a14isshi:
The code is more readable now and tests have been made to make sure that the prior functionality still existed which according to the test I made, was the case. I will therefore merge and close this pull-request

Issue #4910, opened 2018-04-27
Accessed - Add classes to the class column


a16jonis:
Change the dropdown menu into a dataList in the accessed table. It should give suggestions from what classes the student is registered to. But if something unique is typed into the field it should be possible to create said class.

a16chrsa:
This has been solved and merged in #4972.

PR #4972, opened 2018-05-02
Changed the dropdown menu to a datalist.

a16chrsa:
#4910 
Now it is possible insert new classes which are not suggested.

These new classes will not be created inside the class-table but added to the user-table.


a16chrsa:
Changed back to a dropdown-menu with the option to create a new class. The only problem now seems to be this: When closing down the class-creation popup the value selected is "+ New Class". This value is not inserted into the database, but it is a bit annoying.

Issue #4963, opened 2018-05-02
SectionEd: Creating option to press Enter when reveiving a error/delete message to close. 


d15marwi:
![image](https://user-images.githubusercontent.com/37793970/39524363-048aeff8-4e19-11e8-995c-2b836a254a02.png)

When getting this message (trying to create a test in section Ed. ) it should be possible to close the window with pressing the enter button.  The same goes for the delete window.

![image](https://user-images.githubusercontent.com/37793970/39524522-88274bfe-4e19-11e8-93a0-e57f2a2ffebb.png)

a16thegu:
Approved.

PR #4973, opened 2018-05-02
#4963 - SectionEd: Creating option to press Enter when reveiving a error/delete message to close.

a16linas:
Added a 'duggacheck' so the 'enter' button doesn't close when prompted to delete item


a16thegu:
The variable "duggaCheck" is not a very good name, it doesn't explain what it's for and as boolean it's confusing.

a16thegu:
Approved.

Issue #4623, opened 2018-04-17
DuggaEd: New Dugga, all fields must be filled in order to create new dugga


a15magkn:
In order to create a new dugga you need to enter something into every field, otherwise you can't create the new dugga. This can be a bit annoying so this restriction could be lifted. 

## Maybe some of these fields could be left blank 
![hejdu](https://user-images.githubusercontent.com/37794262/38876110-218a94c8-425c-11e8-94a6-5f25842a6213.png)

HGustavs:
Only mandatory is Name as far as I can see.
Empty deadlines and release dates correspond to immediate publication or no deadline message
If no start date is given we assume continous

I suppose that we need some kind of a default template... and default grade system and autograding off as the default.

a14isshi:
Pull-request #4976 is accepted and therefore will this issue now be closed.

PR #4976, opened 2018-05-02
#4623 - DuggaEd: New Dugga, all fields must be filled in order to create new dugga

a16filol:
#4623 - DuggaEd: New Dugga, all fields must be filled in order to create new dugga.


a14isshi:
@a16filol 
1) This pull-request has some merge conflicts 
2) For this to be seen as a good solution, some kind of feedback should be given to the user when the dugga is not going to be created.
3) This pull-request is at the moment very hard to test as everything seem to be failing, when I try to create multiple duggas after each other, they does not get the name I apply to them.

a14isshi:
New tests were made and seem to be working as intended. Instead of a confirmation message of why the dugga wasn't created, it's no longer possible to push the create button without a name which was the stated issue.

Issue #4971, opened 2018-05-02
DuggaED: Error when closing windows with escape.


b16karhe:
**There is a error when windows are closed using the escpae-button in duggaED.** 

_A picture of the issue:_
![issue](https://user-images.githubusercontent.com/37793806/39525970-c64f78bc-4e1d-11e8-95b2-b77daef489c3.png)

PR #4977, opened 2018-05-02
#4971 - DuggaED: Error when closing windows with escape.

a16filol:
#4971 - DuggaED: Error when closing windows with escape. 

The only line that needed to be effected was removing the line that called on showSaveButton(). The function showSaveButton is only used in sectioned. Therefore the function call should be inside that file. (In this case there is no need to call showSaveButton when pressing esc.)


a16thegu:
Approved.

Issue #5071, opened 2018-05-04
SectionEd: Fix code in sectioned.js, should refer to message instead of group


d15marwi:
Looks like a mix up. Type 7 is message and not group.  Should be changed as fast as possible

![image](https://user-images.githubusercontent.com/37793970/39626590-da009f7c-4fa2-11e8-8ce4-b56e8008e0e9.png)

PR #5078, opened 2018-05-04
#5071 - SectionEd: Fix code in sectioned.js, should refer to message instead of group

b16linra:
#5071 - Changed the ID prefix and class name.


b16linra:
A page refresh caused an error with the pull request. This pull request will be closed.

Issue #5015, opened 2018-05-03
Search in several specific columns


a16juleh:
This can be done when issue #5003 is solved.

Add functionality to search for several specific columns by adding && for "and". 
Example search to find files with kind "global" and have a size of 200:
`kind::global && size::200`

PR #5080, opened 2018-05-04
G2 b16andka 5015

b16andka:
It is now possible to specify multiple columns to search from.
Ex. kind::global && extension::jpg (This will find all global files that has jpg as extension)

multicolumnsearch


a16juleh:
This solves issue #5015

Issue #4431, opened 2018-04-12
Saving .md files on the server


c16jonha:
When we press the save button the updated markdown should be saved on the server.

This is a sub-issue of Issue #4174.

PR #5097, opened 2018-05-07
G2 b16andka 4431

b16andka:
It is now possible to save the file when it has been edited in the markdownpreview.

The save button will need some style fixing.

markdownsave


a16juleh:
This solves issue #4431

Issue #4584, opened 2018-04-16
SectionEd: Undeclared variables in sectioned.js


a16thegu:
There are some undeclared variables that need to be declared using the keyword _var_;

- at the top of the file, if the variable is used in multipule functions
- at the top of the function, if the variable is just used in that function.

Optional:
You can add the line "use strict"; at the top of the file and save.
Then run the page in the web browser to see in the console (right click on the page -> Inspect element -> console) where validation errors occurs.

Remember to delete the "use strict"; line when doing the last commit before the pull reqest.

b16linra:
It might not be necessary to delete "use strict" if no more errors occurs after the file has been fixed. Leaving "use strict" will prevent new issues in the future.

a16thegu:
Approved.

PR #5106, opened 2018-05-07
#4584 SectionEd: Undeclared variables in sectioned.js 

a15magkn:
#4584 
@a15magkn

Declared undeclared variables using "use strict".
"use strict" was later removed.


a16thegu:
Approved.

Issue #5081, opened 2018-05-05
Styling of the markdown and file editor


a16juleh:
The markdown editor (and the regular text editor) could use some additional styling. It would be nice if the text area always matched the size of the box that’s holding it, regardless of if the window changes size and choice of browser (the most regular ones; firefox, chrome, safari). We would also like if there were some additional space between the text and the border of the text area, to increase readability.

There are three buttons to close the markdown editor (cancel, close, and “x”). If some of them aren’t absolutely necessary, they could be removed. If there are any other styling properties that could increase the user experience of the editor, feel free to apply them.

<img width="927" alt="skarmavbild 2018-05-05 kl 16 29 30" src="https://user-images.githubusercontent.com/37794632/39664304-8304120a-5081-11e8-9213-5d918c1d71d1.png">
<img width="1035" alt="skarmavbild 2018-05-05 kl 16 29 17" src="https://user-images.githubusercontent.com/37794632/39664305-832af834-5081-11e8-919f-92630c1b405b.png">

HGustavs:
In my minds eye, this is what I see:
* Slightly wider view that adapts to the extents of the window
* Minimum borders around dialog. (<16 pixels from end of dialog to end of screen)
* Minimum border (is 2 pixels achievable??) between editor and preview

Issue #4743, opened 2018-04-23
Make the preview window resizable.


c16jonha:
One should be able to shrink and enlarge the preview window in file editor.

HGustavs:
Well, I prefer a fluid design. As large as possible and adapting to screen borders. No need to make it movable, that tends to be annoying in the long run.

a16juleh:
Instead of manually shrinking and enlarging the preview window, this should be done automatically when the window is resized. A new issue for mobile mode will be made.

PR #5144, opened 2018-05-08
G2-a15setno-b16nicbr-5081

a15setno:
The textarea (in markdown and file editor) will now always match the size of the box that is holdig it, no matter which browser the user is using or the size of the window. We have also added some additonal space between the text and the border of the text area. The "close"-button is not placed correctly any more (new issue?), yet it still works. 

The regular text/file editor 
---
editor

The markdown
---
markdown

This is a fix for issue #5081


a16juleh:
This also solves #4743

Issue #5040, opened 2018-05-04
Copy objects with line doesn't work propertly


a16juleh:
If objects are copied with a line between them, the line isn't copied. The line should be connected to the new objects. One should not be able to copy only lines. 

<img width="367" alt="skarmavbild 2018-05-04 kl 09 30 51" src="https://user-images.githubusercontent.com/37794632/39616891-de485bb2-4f7d-11e8-99b2-5f695503aa71.png">

a16juleh:
The underlying structure of the code makes this issue way bigger than it should be. New issues will be created, that will focus on fixing the underlying problems behind this first.

a17vikab:
IF you copy objects with a line connected, it will later attach to the new copied object. See gif-link for example: https://gyazo.com/df13477f3f2cf12d8d4f5cab90e0c22a

a14oscna:
After recent meeting we decided that the logic for copying as well as the logic for drawing lines must be remade to work as requested. When objects are copied we basically only want to copy the information about the properties of the object(s). Then we want to use the same draw function as we use when we draw a new object, but apply the copied information to mimic the appearance of the copied object(s). Within the information that is copied there should be data which tells the program which objects a line was drawn **from** and which objects a line was drawn **to**, which in turn can be applied to the newly created copies of these objects to determine where the new lines should be drawn. We assume that in order to implement such logic for copying lines, we must recreate the logic for how lines are drawn initially to have access to such data required to essentially copy lines.

a16tobfr:
#6977 will continue on this. Several connected issues were solved in this issue but when it comes to lines, the problem seems to be bigger than expected. May need to refactor how lines are made in order to get this to work properly.

PR #5147, opened 2018-05-08
G2 c16carbe 5040

c16carbe:
This pullrequest solves #5040 (kind of)

This solution is in a very, very early stage and should be changed.

It works to copy two objects that has a relation with a line.
It currently does not work to copy two objects where the line was originally drawn from an ER attribute.
UML connections don't work very well either.

The code in it's current state is hardcoded to only support the first connection an object has (therefore copying objects with multiple lines attached to it will not work).


a16juleh:
We are closing this one. The underlying structure of the code makes this issue (#5040) way bigger than it should be. New issues will be created, that will focus on fixing the underlying problems behind this first.

Issue #4352, opened 2018-04-09
Part 6 - Check of code standard in HGustavs/LenaSYS/DuggaSYS/sectioned.js


a16thegu:
Check if below function follow the code standard stated in HGustavs/LenaSYS/code_standard.md.

- updateItem()
- createLink()

a14chrve:
Variable names should be declared. Unnecessary comment should be removed.

PR #5195, opened 2018-05-14
#4352 Part 6 - Check of code standard in HGustavs/LenaSYS/DuggaSYS/sectioned.js

a15magkn:
#4352 
@a15magkn

Declared some variables that were undelcared, because of the codestandard. Nad removed a function that never was used.


a14isshi:
The variables don't need to be global and therefore the new code does change anything. An old function is removed and it does not seem to be used at all. Merged!

Issue #4169, opened 2018-03-21
Group editing as part of the access editor (mock-up)


HGustavs:
Group codes should be added/modified from a pre-existing list of codes and group number/values/codes.

* A json object column called groups should be added in the access editor (connecting student to course with a certain set of goup assignments for that course).
* If a group code is assigned a value we add this to the group json object and conversely, if cleared we remove it from the json object.
* This should mainly be achieved through sortabletable plus a static form and some code for reading to/from the json data as a text string.

![mockup_dropdown](https://user-images.githubusercontent.com/6978801/37728529-c6ef4efe-2d3a-11e8-873c-83944a3c92a4.png)

HGustavs:
We never need to se the json on-screen. No json display or copying, just something that looks like the dialog above.

d15marwi:
New Mockup after our last meeting. We will add 3 kinds of group-types each made of 8 groups:

First kind[ 1,2,3,4,5,6,7,8]
Second kind [A,B,C,D,E,F,G,H]
Thrid kind [I,II,III,IV,V,VI,VII,VIII]

The Teacher can than decide whos is in each group. 

![image](https://user-images.githubusercontent.com/37793970/39130771-390bcd40-470e-11e8-99b8-91736ee06573.png)

a16thegu:
@HGustavs, can you take a look at this mock-up?

d15marwi:
Splitt the issue into smaller parts
1) Creating a Group Column with sortableTable  #4827 

....more issues coming later...

b16linra:
Issue #4170 is related to this issue.

d15marwi:
Update on this issue, we will remove all old group related tables and start over on this. We will create diffrent mockups and prototypes for this to controll the quality before implementing this into the system. The idea is to build functionality that is scalable in the future and easy to understand. This is very important because of the problems in the past with this. 

The goal  in the coming weeks will be to build a group functionality so that we can asign groups to a student i Access and then assign a group to a item in section editor.

a14isshi:
Pull-request closed.

PR #5218, opened 2018-05-14
#4169 - Create mutliple select for access to users.

a16filol:
#4169 - Create mutliple select for access to users.


a14isshi:
Looks good and seem to work as intended. Merged and closed.

Issue #5082, opened 2018-05-07
Diagram objects can be inverted


c16linst:
When you grab an object in the diagram in one corner and drag it past the opposite side, the object will flip and show its backside (no text will be visible). It shouldn't be possible to invert any objects.

![skarmavbild 2018-05-07 kl 08 34 41](https://user-images.githubusercontent.com/37794714/39688132-92362cda-51d1-11e8-965f-727033e4fde2.png)
![skarmavbild 2018-05-07 kl 08 34 48](https://user-images.githubusercontent.com/37794714/39688133-925762ec-51d1-11e8-9969-222ca60c66af.png)

HGustavs:
One solution is to change places/positions of point 1 and point 2 if the extents of the box are negative.

The fix should ___not___ be to have different draw options depending on whether the extents are positive or negative.

PR #5227, opened 2018-05-14
G2-b16ponpe-5082

b16ponpe:
When there is an object on canvas you are not able to invert the object. You can't minimize the object smaller than the original size of the object. Added some if statements to check if you are able to minimize the objects.


a16juleh:
This solves issue #5082

Issue #5232, opened 2018-05-14
DuggaED: Implement further the search functionality


a16konka:
The search functionality needs to be implemented further. To work with the search functionality it is needed to write "rowFilter" on ROW 285 in duggaed.js and then look at ROW 163-165 in duggaed.php and ROW 766-773 in duggaed.js

b16karhe:
**THIS refers to the search engine for the variant editor.**

PR #5247, opened 2018-05-15
#5232 DuggaED: Implement further the search functionality

b16karhe:
#5232 Made it possible to search in the variant editor through:

- **_Removed unused functions._**
- **_Fixed the filter function._**
- **_Made the filter reset when the window is closed and a new one is opened._**
- **_Added some comments._**

### DIFF
--- a/DuggaSys/duggaed.js
+++ b/DuggaSys/duggaed.js
@@ -776,7 +776,7 @@ function compare(a,b) {
 
 //Filtering from the searchfield.
 function rowFilter(row) {
-	if (row.param.indexOf(searchterm) != -1){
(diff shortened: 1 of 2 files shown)


b16karhe:
There are some errors when pressing enter but they stem from "sortableTable" and is not something that i have control over.

a14isshi:
It all seem to work as intended and the errors are inside sortabletable. This however does not affect the functionality but should be taken care of. Another false I found is that the search-function is case-sensitive which it probably shouldn't be and therefore this should be fixed before this is merged.

a14isshi:
@b16karhe

b16karhe:
_@a14isshi_ **[DONE]** Transformed all characters to lowercase to stop the matching from being case sensitive.

Issue #5207, opened 2018-05-14
ResultEd - grading bug


a16jonis:
In the result editor view

Follow these steps to replicate:

Navigate to ResultEd -> Filter after "Only Pending" or "Show Teachers" -> Try to grade any dugga. 

What ends up happening is that the dugga doesn't appear to get graded. If you update the page and check the dugga it actually is graded, though. It seems like grading after filtering causes the grade buttons to not change colour when grading.

The two filters mentioned above are only examples of filters known to cause the bug. Additional filters could also cause the bug.

a16jonis:
A screenshot would not show much in terms of how it looks. I can clarify how to replicate though:

Enable either of these filters: 
![filter](https://user-images.githubusercontent.com/37793947/39995435-7768170a-577c-11e8-9ac1-6f2ae32a1926.PNG)

Grade a dugga (the yellow one - it wont work): 
![grade](https://user-images.githubusercontent.com/37793947/39995447-7ea949f8-577c-11e8-9c65-067970979791.PNG)

Refresh the page (It's actually graded): 
![graded](https://user-images.githubusercontent.com/37793947/39995463-8dc4ff0e-577c-11e8-8a65-f844217e756f.PNG)

b16nilpe:
The bug was introduced when a change was made to prevent rebuilding the entire table. This bug will be corrected in the related issue #5237

a16chrsa:
Now works.

#5265.

Issue #5237, opened 2018-05-14
DuggaEd - dugga grading bug


a16jonis:
In Dugga editor view:

Sometimes duggas that you grade wont receive the correct colour. This is due to local data not being updated.

HGustavs:
Since this has to do with student results I consider it high priority to fix this issue.

a16chrsa:
Fixed.
#5265

PR #5265, opened 2018-05-15
Local data is updated when grading in resulted

b16nilpe:
Issue #5237 
The colors should now update correctly when grading.
When setting the grade G and then removing the grade, the color of the cell would change to purple instead of yellow/orange as it should.

The changes made also fixed issue #5207 since most of the issue was due to local data not being updated correctly when grading.


a16chrsa:
Works.

Issue #5210, opened 2018-05-14
Styling for markdown preview window


a16juleh:
For now, the styling isn't what the costumer wants. if it could emulate the picture below with fieldsets instead of windows in the editors, it would be more consistent with the rest of LenaSYS. The styling for the buttons should also follow the example. 

![32375958_10209676186140691_3049710432685129728_n](https://user-images.githubusercontent.com/37794632/39985721-cdc82d00-575f-11e8-9162-a0c5d0ef75ce.png)

b16nicbr:
Where is a good spot to set the shortcut buttons?, on the legend line (instruction file) or edit variant or as buttons or if you have another suggestion of where to set the shortcuts @HGustavs

HGustavs:
Not an easy question to answer. I would put the buttons somewhere near the top.

It does not have to emulate this specific dialog, but I do believe that there are not many side-by-side dialogs in Lena.

Does not have to use fieldsets exactly, but something that looks similar to this so that we are using a style that is somewhat recognizable across the various views.

I would prefer emulating some other dialog rather than inventing yet another styling option.

PR #5285, opened 2018-05-15
G2-b16nicbr-a15setno-5210

a15setno:
The styling is changed now, but we kept the windows in the editor because of the shortcut buttons. Also, we added some small changes - like the scroll in the textarea. And the save and close buttons are more consistent with the rest of LenaSYS. You can see the changes below, the layout for the mobile view is also updated!

The regular view
---
vanlig

This is the layout for the mobile view
---
mobil

This is a fix for issue #5210.

### DIFF
--- a/DuggaSys/fileed.php
+++ b/DuggaSys/fileed.php
@@ -137,9 +137,9 @@ pdoConnect();
 
                     </div>
                 </div>
-                <div class="markText">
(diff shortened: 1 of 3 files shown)


b16nicbr:
we have changed the design of markdown and file editor so it now looks like this.

This is now the normal view 

issue-5210

This is the mobile view

issue-5210-mobile

The file editor looks like this now on the normal view

issue-5210-editor

This is the mobile view of the file editor

issue-5210-editor-mobile

Issue #5284, opened 2018-05-15
Functionality for UML cardinality


c16carbe:
Cardinality isn't really supported between UML objects yet. This should be fixed.

PR #5303, opened 2018-05-16
G2 c16carbe and a16rasja 5284

a16rasja:
You are now able to add cardinality between UML classes.
The dialog menu has been changed so it dynamically changes depending on if it is UML type cardinality or not.
chrome_2018-05-16_09-28-36 skarmavbild 2018-05-16 kl 10 56 18 skarmavbild 2018-05-16 kl 10 56 44


a16juleh:
This solves issue #5284

PR #5349, opened 2018-05-17
Fab button for adding/importing users added to accessed

a16seplu:


a16chrsa:
This change is fine. However it would be better for the options to appear when hovering the fab button.

Currently the options are displayed by holding down the left mouse-button for about a second. It is the same for the fab button at fileed.php. Another thing to consider is if the import-user option should be the default option when the fab-button is clicked. Maybe have add-user instead as the default option?

Will create new issues for these things.

PR #5380, opened 2018-05-17
The title and searchbox follows in result.

a16katba:
Fixed so the title and searchbox follows when you scroll to the right in result .


a16katba:
Made some new changes and created a new branch .

Issue #5409, opened 2018-05-21
AccessED: When a student has been assigned to a group the select contracts. 


a14isshi:
Change so that when a student has been assigned to a group the selects contracts. This can be done by at the same time making a AJAX query to the database to update the new information.

a14isshi:
Pull-request #5418 for this issue is merged and closed. Therefore this issue is also closed.

PR #5418, opened 2018-05-21
#5409 - Implement group relations into accessed.

a16filol:
#5409 - Implement to add group into accessed.


a14isshi:
Manually groups had to be created as the groups didn't exist. When a student were assigned to group, this was done also on the database, as well as when a student were removed from a group. When a student is assigned to a group the checkbox is not checked. This can be seen as another issue and therefore this is accepted, merged and closed.

Issue #5377, opened 2018-05-17
JSON deadline object - Creating new dugga


a16jonis:
Check #5345 for information

The JSON deadline object added in the mentioned merge request seems to work only when editing an already existing dugga, not when creating a new one. This functionality needs to be implemented as well.

HGustavs:
Yes this is quite important as this functionality is not useful without it.

d16linja:
Solved, the first deadline gets added as a deadline as both json and the old way to that everything still works as usual at the webpage

PR #5437, opened 2018-05-21
Added jsondeadline for the createdugga function

d16linja:
Issue #5377 

@b16nilpe knows whats up 

The function for adding 3 deadlines as a json now also works with creating new tests


b16nilpe:
@a16seplu The function NewDugga shows a popup window for creating a new dugga, the submit button in that window runs createDugga on click.

d16linja:
@a16seplu This update requires that the installer have been run since the latest merge, that might be the problem if it does not work when you test it

c16matan:
a16seplu didn't have a database to test the issue with, so the approve from f16linli is enough. Good to be merged 👍

Issue #5429, opened 2018-05-21
Groups: When a course is created, all groups for the course are created


a14isshi:
When a new course is created it also need to contain the groups for is group.

d15marwi:
groups for the group? Do you mean the 3 group types?(Roman numbers, numbers and Letters?)

a14isshi:
#5450 is merged and closed. therefore this issue is now closed as well.

PR #5450, opened 2018-05-21
#5429 - Groups: When a course is created, all groups for the course are created 

a16filol:
#5429 - Fix for adding groups when adding a new course.


a14isshi:
The information courseID should exist inside the "vers" table but as the earlier implementation has not made it as an foreign key this is not possible. This needs therefore be done as you have done it not and 2 different tables are needed for the information "vers" and "courseID".

Issue #5451, opened 2018-05-21
Groups: Add groups for every course if the groups does not exist


a16filol:
To be backwards compataible we need to add gorups if they does not exist for each course.

a14isshi:
#5452 pull-request is merged and closed.

PR #5452, opened 2018-05-21
#5451 Groups: Add groups for every course if the groups does not exist 

a16filol:
#5451 - Implement so groups are seeded when visiting sectioned.


a14isshi:
We also want the attribute vers from the vers-table, it is therefore better to do this check when a new version of a course is created. You will then already have the vers and cid available.

a14isshi:
As it is now, if a new version of a course is created it does not get new groups. This is a part of the issue as well and need to be fixed before this is accepted.

a14isshi:
Fixed! Merged and closed.

Issue #5453, opened 2018-05-21
Groups: Show which groups the user are assigned to.


a16filol:
Show which groups the user are assigned to.

a14isshi:
#5454 is merged and closed. This issue is therefore also closed.

PR #5454, opened 2018-05-21
#5453 - Implement so the checkboxes are checked if user is assigned to the gr…

a16filol:
#5453 - Implement so the checkboxes are checked if user is assigned to the group.


a14isshi:
Tests that were made outside of codecheck: Tested to check some groups for 3 different student. Made sure that the ones I checked on one student didn't get checked on other students. I also shut down the browser and cleaned the cache and returned to the page to make sure that the ones I checked are still checked. It all worked as intended and this pull will therefore be merged and closed.

Issue #5503, opened 2018-05-22
Group: Instead of getting information from course we need to receive it from vers


a14isshi:
Right now we get information as if it is the same for all groups. Instead we want to have to receive groups from each version of a course.

a14isshi:
#5507 pull-request merged and therefore this is closed.

PR #5507, opened 2018-05-22
#5503 Take vers in consideration.

a16filol:
#5503 - Take vers in considiration


a14isshi:
The last problems when changing version inside "Access" has been fixed. This seem to work now as intended, therefore this will be merged and closed.

Issue #4189, opened 2018-03-21
Sectioned group indicators


HGustavs:
There should be a field in sectioneditor for connecting to group-ids.
* Should work for section type, only Group Activity
* Groupid is just a text field
* Text  field contains reference to a group e.g. "Presentation2"
* If the json data for current student in the current course is {"Presentation1":"A","Presentation2":"C"} for the section that has "Presentation2" in its group field, we write "C" next to the text written for the current section ...

and so on for each section...

a16linas:
I'm not sure I fully understand the issue. Can you explain the "group-id"?

a16thegu:
@HGustavs, can you take a look and explain more.

a16thegu:
After talk with client, issue will be on hold.

d15marwi:
design for this icon was made in #5051

a14isshi:
Merged and closed.

PR #5560, opened 2018-05-24
#4189 - Implement group to listentries that shows the group active.

a16filol:
#4189 - Implement group to listentries that shows the group.


a14isshi:
Looks good and works as intended. One problem is that if you want to edit the listentry then u once again have to select the group it is intended for. This is a new issue but can be easily solved by getting the groupID from the database and write it out the coloumn.

Issue #5563, opened 2018-05-24
Mobile browsing - searchbox have too small margin


a16juleh:
The margins for the searchbox in mobile view are too small. 

![33326907_10209726606121159_7823491153973477376_n](https://user-images.githubusercontent.com/37794632/40469973-73b1bde2-5f32-11e8-8d75-a18b839f31d4.jpg)

PR #5565, opened 2018-05-24
G2-c16jonha-5563

c16jonha:
I have added a margin to the sortingDiv when the max-width of the screen is 768 pixels. I also indentated the code a little bit in the css file.

This is a fix for issue #5563


a16juleh:
The result is not fully done and the time is running out. The over all styling is very good so this can be changed later on.

PR #5665, opened 2019-04-05
Removed templates dir. Fixes #5664

### LINKED ISSUE #5664: Remove the "templates" directory and its contents
It seems that a13chrju added the folder for no apperent reason in commit 9b50ec5 four years ago. Now it only serves to increase the kognativ load on new developers trying to get a grip on what every files does in the project.

a17adajo:


a17adajo:
Thanks andreas!

Issue #5644, opened 2019-04-04
Mobile view: Deadline date cuts off


b16johso:
In mobile view, the deadline dates get cut off due to incorrectly styled media queries.
![Capture](https://user-images.githubusercontent.com/37793460/55555403-80357f80-56e5-11e9-8c84-497bb6b7261b.PNG)

b16johso:
Removing the media query in question seems to fix it @HGustavs 

```
@media only screen and (max-width: 1085px),
only screen and (max-device-width: 1085px) {
     . . .
  .dateSize {
    width: 50px;
  }
}
```

![Capture](https://user-images.githubusercontent.com/37793460/55556309-89bfe700-56e7-11e9-8946-13582bfbec50.PNG)

HGustavs:
This looks better?

Whats the verdict?

b16johso:
Merged https://github.com/HGustavs/LenaSYS/pull/5684

PR #5684, opened 2019-04-05
Removed dateSize from the media query for small devices

a17joaap:
Fixes issue [5644](https://github.com/HGustavs/LenaSYS/issues/5644)


b16johso:
Seems to be working

Issue #4244, opened 2018-03-29
Diagram Toolbar should be able change from vertical to horizontal


c16jonha:
![toolbar](https://user-images.githubusercontent.com/37794047/38081547-985d9810-3344-11e8-8a33-3aa245292f86.PNG)
You should be able to change direction by clicking on a button in the toolbars top right corner.

HGustavs:
The toolbar is far from ideal especially in mobile.

Perhaps some kind of docking / pinning mode or as a hotdog menu?

PR #5699, opened 2019-04-08
Fix for issue #4244

c17matjo:


a17okaoz:
Issue #4244 Fixed. They have implemented a button where the rotation takes place horizontally. For each mode, the width of the box changes and where the tools are placed in the right place. The code is well written but comments may be needed if another person is to enter and make changes.

Issue #5696, opened 2019-04-08
New Item button doesn't work


a17poner:
The "new item" button when logged in as a superuser on a course doesn't work. 
When pressing it nothing happens and this error 

> sectioned.php?courseid=4&coursevers=1338:167 Uncaught ReferenceError: selectItemdefaultNewItem is not defined
    at HTMLInputElement.onclick (sectioned.php?courseid=4&coursevers=1338:167)

is thrown in the browser console.
					`	<input id='addElement' style="display:none;" type='button' value='+' class='submit-button-newitem' title='New Item' onclick='selectItemdefaultNewItem();'>`
is the line that appears to be the problem(in sectioned.php), searching in the entire project no javascript function named selectItemdefaultNewItem() exists.

PR #5700, opened 2019-04-08
Grupp c2019 5696

a17poner:
Changed row 143 in sectioned.php, the new item button for superusers should now work (the + at the top of a course page on the same tab as the course name), fixing #5696 .


a17poner:
Making a new pr with the new branch as base

Issue #5595, opened 2018-08-01
Consistency between FABs


a16hamsa:
The FAB in DuggaEd have another style than the other FABs and the FAB in CourseEd doesn't have the same hover-effect as the other FABs in the project. This should be fixed so all the FABs are consistent in style across the project.

a16hamsa:
#5596 All FABs now look and behave the same on all pages.

a17fellu:
Made the FAB on courseed consistent with the others

a17fellu:
Branch with the change: https://github.com/HGustavs/LenaSYS/tree/GruppC2019%235595

Changes to FAB (Button with + sign) on courseed.php

- Removed the zoom effect on hover and the excessive drop shadow is changed to fit the others

- Made the FAB get brighter on hover just like the one on the other pages

PR #5701, opened 2019-04-08
GruppC2019_5595 fix

a17fellu:
Link to Issue: #5595

Changes to FAB (Button with + sign) on courseed.php

Removed the zoom effect on hover and the excessive drop shadow is changed to fit the others

Made the FAB get brighter on hover just like the one on the other pages


a16joare:
seems to be working.

Issue #5661, opened 2019-04-05
Null value appears in result view


c17johha:
broken-down issue from:
Uneconomical use of space in result view #5625

value "null" appears in view. Taking up space.

![52473472-2d809100-2b96-11e9-8359-37aaaab5d133](https://user-images.githubusercontent.com/49199825/55617674-f218d200-5794-11e9-834d-a3e767a8ecfa.png)

c17johha:
![pasted image 0](https://user-images.githubusercontent.com/49199825/55617827-42902f80-5795-11e9-9ca7-e0e0274f0531.png)
![pasted image 0 (1)](https://user-images.githubusercontent.com/49199825/55618206-3a84bf80-5796-11e9-852f-44e592ba83b4.png)

"null is gone"

HGustavs:
This looks much better!

a17pongo:
@c17johha this issue is merged and fixed.
Please close this issue.

Issue #5625, opened 2019-02-08
Uneconomical use of space in result view


HGustavs:
Currently the data in the result view is slightly higher than necessary.

A mockup of the problem:

![screenshot 2019-02-08 11 32 51](https://user-images.githubusercontent.com/6978801/52473472-2d809100-2b96-11e9-8359-37aaaab5d133.png)

The "null" in each student which contains something, and the times graded in the result...

this makes the whole result row one row of text too high.

I would like to add the numer of marked times as just a number enclosed in parenthesis next to the result in some place e.g. (1) or (2) or (3)

HGustavs:
What do the null-s contain?

one is study program?

a17amama:
Null next to studentID contains Class and the null at the bottom contains SetTeacher, should any of them be removed?

a17pongo:
Resolved and merged with groupB branch

PR #5702, opened 2019-04-08
Update accessedservice.php #5625

c17johha:
This pull request is actually a sub-issue of "Uneconomical use of space in result view" #5625.

This is the isse:
Null value appears in result view #5661
https://github.com/HGustavs/LenaSYS/issues/5661


a17pongo:
Needs to create and pull a branch for the issue #5661 for testing.

Issue #3752, opened 2017-05-10
Diagram virtual "paper" support 


HGustavs:
The diagram is infinite.... but it would be nice to be able to "print" the diagram... the first step would be to draw an outline of an A-4 document on screen. 

If this option is enabled we should see a transpared dashed outline of the size and dimensions of a virtual A-4 including the "holes" for putting the a-4 into binders.

PR #5703, opened 2019-04-08
Fix for issue #3752

a17jonro:


a17okaoz:
Issue #3752 Fixed. When i clicked on the "View" button and on "Display Virtuel A4" button a guideline appears on the paper that shows the with that a A4 paper has.

PR #5724, opened 2019-04-08
GruppC2019 Fix _4843

a17alevu:
Changed so that the href is equal to "href = "../DuggaSys/courseed.php""  in the 404.php file. The change now allows you to go to the homepage while keeping your login session.


a17fellu:
Unnecessary fix since it is not an issue on the live site. Only a problem for people running a local build with a different folder structure

PR #5727, opened 2019-04-08
Commented out null values #5625

a17amama:
We have commented out setTeacher (row 757 in resulted.js) because it was holding unnecessary null-values.


a17pongo:
closing request because of duplication

Issue #5720, opened 2019-04-08
Add guidance on how to set mySQL root password in the LenaSYS install guide


a14oscna:
The lenaSYS install.php guide requires users to fill in all text fields in every part o the installation to proceed. If you run a local database for example by using xampp, the default root has no password set, hence you can not proceed the installation as you have to fill in a password when you have none. The guide should therefore instruct the user on how a password can be set for mySQL.
![guide1](https://user-images.githubusercontent.com/37793141/55727227-d5d59900-5a11-11e9-9340-3e4a2b7d0e42.PNG)

a17adajo:
I yhink lenasys should allow an empty root password instead

a14oscna:
I changed the fix to allow the user to proceed without a password rather than enforcing the user to set a password if they do not already have one. However, it is still very bad practice to not have a root password.

PR #5730, opened 2019-04-08
Fix for #5720 install guide text for the install.php GUI

a14oscna:
The guide now instructs the user to set a root password if none is set (default for some applications like xampp). It also shows a sample shell command to set a password. See last 2 sentences in image below: 
guide2
#5720


a17drama:
Would you be kind and implement this with the new "master". Branch it's supposed to merge to is obsolete. Should merge to gruppA2019W16. Also mention this pull request in the new one for history purposes. Once that is done you may close this pull request or inform @a17drama to do it.

Issue #5600, opened 2018-08-19
SectionEd: The small "+" does nothing


a15magkn:
At the moment the smaller "+" button in the top right corner does nothing. 
A suggestion could be that it creates a new default object, similar to clicking the FAB-button.

b17nicsv:
Ready for testing. 

[#5732](https://github.com/HGustavs/LenaSYS/pull/5732)

PR #5732, opened 2019-04-09
GruppB2019#5600

b17nicsv:


c17johha:
#5600

a17amama:
test5732
Code does what it should, pressing it creates a "new code" similar to the FAB-button. Code follows the standard.

PR #5734, opened 2019-04-09
Grupp a2019#5663

a16tobfr:
Everything seems to work as it should after deletion of the two text files:
diagram_new_function_names.txt
diagram_new_variable_names.txt


a17jacsv:
Approved.

Issue #5628, opened 2019-03-29
We need a better dugga specifically for entering time sheets.


HGustavs:
I suggest something like the code for making surveys in surveySYS.

Page 3 in the following research paper:

http://crpit.scem.westernsydney.edu.au/confpapers/CRPITV42Clark2.pdf

HGustavs:
Im thinking it can be used for self-reporting attendence in some way.

a17robda:
![dagbokmock](https://user-images.githubusercontent.com/37792328/55612328-48334880-5788-11e9-987d-8ef30cbba7d2.png)

Started working on a time sheet mockup for the general idea. We would like some feedback on this simple mockup if there is anything we have forgotten. We thought about adding a preview after adding each row.

a17poner:
Is this something we want for all text file submissions? As we understand currently, to add this we need to change in generic_dugga_file_receive.js under the "else if(type =='text')" which would change any possible text file submission in lenasys to use this type of timecheet form, if we're correct in our assessment is this how we want it to be or should we make/add a new type for it?

a17poner:
@HGustavs in order to test this and start working with integrating our form, how do we create a diary(dagbok) in lenasys? Been trying for 30 minutes to get through the superuser ui to create something similar to whats on the live version for this course but with very poor results

a17joaap:
@HGustavs How do you want the data to be saved? Currently there is no infrastructure that supports this tabular data, at least that I can find. Should we create a way for the system to parse this? 
![timesheet](https://user-images.githubusercontent.com/37792198/55990480-0fbfcd00-5ca8-11e9-9198-48d809a980a5.PNG)

a17joaap:
I went ahead and created a way to save user input in a json file, this way it could be parsed and reentered to the table when the user revisits the page, if this is something we want.

HGustavs:
I would say that we create a time sheet table in the main database somehow.

And that we connect this data to the section in the course, this way we can have more than one time sheet in a course.

HGustavs:
Generic names... I suggest Kind or Type and Reference instead of pull request and number

HGustavs:
Week numbers also as a sortable / searchable column

a17joaap:
I created a new table for timesheets, with references to the course, quiz, user, and moment.
![timesheetdb](https://user-images.githubusercontent.com/37792198/56033691-615a6d00-5d25-11e9-938d-3a54ec807e5c.PNG)

[...]

Issue #5618, opened 2019-01-24
Open pdf in new window


HGustavs:
When viewing an exam we want a third option to open pdf in a new window in addition to the current open in same window or open external

HGustavs:
Generic file submission dugga.

HGustavs:
But also in fileed.js

a17poner:
Commited fix and created pull request, closing.

PR #5738, opened 2019-04-09
Fix for issue #5618

a17poner:
Accidentally named the branch after another issue I was working on altough it had nothing commited to it, attempted to fix it but failed but this is a fix for issue #5618 and NOT #5628 which is still being worked on.
image
The icon to the right of the download icon is the added one and should when clicked open the pdf in a new tab. To test this I reccomend making sure you have 1 user which you can log in to ( I used l94lewal after changing his password to a 10 round [bcrypt ](https://bcrypt-generator.com/) and changing it in phpmyadmin) then making a new quiz/test and having it be a generic_file_receive then putting submission type to pdf and uploading a random pdf as the user you chose before going on the superUser again to check in result. (OBS! make sure the person in access is set to the correct version, e.g. for Webbutveckling - datorgrafik it is ht15 that will be default)

### DIFF
--- a/Shared/dugga.js
+++ b/Shared/dugga.js
@@ -1336,12 +1336,18 @@ function findfilevers(filez,cfield,ctype,displaystate)
 							tab+="<tr'>"
 
 							tab+="<td>";
-							// Button for making / viewing feedback - note - only button for given feedback to students.
+
 							if (ctype == "link"){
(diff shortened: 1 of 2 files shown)


a16joare:
I tested it on my local machine and it seems to work. Could not test this on the test server due to server rights.

a17poner:
Making a new pr with a new branch that is based on the week 16 C branch.

Issue #5707, opened 2019-04-08
Temporary solution?


a17celal:
In diagram_symbol.js there is a comment saying that the solution is temporary in the drawERAttribute function. Is it really temporary? Should be investigated.

![Capture3](https://user-images.githubusercontent.com/37792271/55723595-ba669000-5a09-11e9-9490-361d877c7c36.PNG)

PR #5742, opened 2019-04-09
Removed the comment as the solution is a proper fix. Fix for #5707

a14oscna:
This fix is for issue #5707
Investigating further I found that the issue has been previously solved as of issue #4261
Therefore I removed the comment as it was misleading.


a17drama:
Would you be kind and implement this with the new "master". Branch it's supposed to merge to is obsolete. Should merge to gruppA2019W16. Also mention this pull request in the new one for history purposes. Once that is done you may close this pull request or inform @a17drama to do it.

Issue #5626, opened 2019-03-25
Title attribute for sortable tables


HGustavs:
A requested feature for all sortable tables is to introduce title / alt attributes for the table headings so that it is possible to view the headings when hovering etc.

HGustavs:
For example the result view headings that tend to be long...

![Screenshot 2019-03-25 10 29 32](https://user-images.githubusercontent.com/6978801/54908802-fa495580-4ee8-11e9-890c-92d3fa494d3c.png)

... and if we only see last part of heading can be very un-informative.

b17timjo:
There are a great number of potential changes, what is the preferred behaviour?

Ex. 1) Removing "rtl" shows start instead of end 
![remove_rtl](https://user-images.githubusercontent.com/38911528/55705235-a3f70f00-59de-11e9-83ea-e521c0a005ea.png)

Ex. 2) Removing hard width limit of 150 px:
![remove_width_limit](https://user-images.githubusercontent.com/38911528/55705239-a78a9600-59de-11e9-8fc5-906f53b57c32.png)

Ex. 3) Removing nowrap on header (but keeping width limit):
![remove_nowrap_on_th](https://user-images.githubusercontent.com/38911528/55705243-aa858680-59de-11e9-82be-515c72294100.png)

b17timjo:
@HGustavs as you will be the primary user, do you have any preferences?

PR #5749, opened 2019-04-09
Gruppb2019#5626 Headingfix

a17ludca:
Headings are fixed so that the entire name shows, this issue is in resulted.php and it concerns the tableheads.

[#5626](http://issue_5626.gruppb.webug.his.se:20001/DuggaSys/resulted.php?cid=2&coursevers=97732)


a17adasy:
Tested. The title of the tableheads are now fully shown and much more clear. The code standard is followed.

Issue #5746, opened 2019-04-09
Coordinates showing non useful info


a17drama:
![Capture](https://user-images.githubusercontent.com/46029340/55796420-4809b480-5aca-11e9-9799-51fba7bd893e.PNG)
Coordinates should show mouse position. As it is now it's showing something that has no relevance to the user.

HGustavs:
For debug purposes useful.... perhaps relegate to developer view ?

a17vikab:
> For debug purposes useful.... perhaps relegate to developer view ?

On the second thought, it might be useful for both the user and the developer and therefore it might be a good idea to add a second bar or extend the existing bar so it can contain both mouse-coordinates and origo-coordinates?

a17drama:
This is useful for the developer because we move the canvas around and these coordinates show the offset. In any case if it is for the developer it shouldn't be shown until the developer mode is active. Regular users have no use of this info and it's just confusing. This will also be obsolete if #5748 is implemented because the canvas would be static in that case.

PR #5751, opened 2019-04-09
Fix for issue #5746

a17vikab:
Added two new variables that tracks the mouse coordinates in the canvas and later change the innerHTML so it always shows on screen. 

#5746


a17drama:
Would you be kind and implement this with the new "master". Branch it's supposed to merge to is obsolete. Should merge to gruppA2019W16.

Issue #5632, opened 2019-04-01
Magic headings appear wrong in result view


HGustavs:
In the result view the magic headings do not work.

In the result view the first column (if visible) student name etc should be magic i.e. if this columns scrolls out of view it should remain to the left.

The column names (on first row) should not scroll out of view either.

If marking a course with 30+ students and 30+ assignments we need this to see which part of the course / which student we are working with.

HGustavs:
<img width="339" alt="Screenshot 2019-04-03 23 32 00" src="https://user-images.githubusercontent.com/6978801/55514543-b4b72600-5668-11e9-9470-f6c42e048ee4.png">

Currently this is what it looks like in the running version

In the demo code from sortabletable this is what it looks like

<img width="654" alt="Screenshot 2019-04-03 23 33 33" src="https://user-images.githubusercontent.com/6978801/55514656-f942c180-5668-11e9-920c-1da090550b9b.png">

Unsure if it is a configuration thing or if it is a bug?

I was toying with the idea that the magic heading perhaps was hidden below the top navigation menu?

As we can see the top header does not remain at all, and the leftmost column does not size each cell properly...

HGustavs:
How does it appear in the current main branch version?

c17julka:
![felmed](https://user-images.githubusercontent.com/37792321/55612691-14a4ee00-5789-11e9-96f8-cc5339584d45.png)

Error messages appear after clicking "+"-button in right-bottom corner in sectioned, which also leads to resulted recieving error msg. Also removes table/entries in table in resulted.

c17alepe:
![image](https://user-images.githubusercontent.com/38323901/55800860-f4e92f00-5ad4-11e9-9988-6c226615f76c.png)
Fixed Magic heading head and Magic heading for column numbers and names

HGustavs:
The shaky-ness is necessary due to the chane of how onscroll works in browsers, as this functionality has been deprecated due to scrolling issues on mobile chrome browsers....

We had a shake-lesser version before but changed to this shakier version in a previous update of the library.

PR #5752, opened 2019-04-09
Request for issue #5632

c17julka:
#5632


c17alepe:
http://issue5632.gruppb.webug.his.se:20001/DuggaSys/resulted.php?cid=2&coursevers=97732

Here is an url for testing purposes regarding this issue. Make sure to log in.

b17timjo:
On firefox, the headers are off-center vertically:
5632_offset

But they do scroll (shakily, as previously mentioned)

c17alepe:
Fixed the bug in firefox. Commented out<thead> in magic heading for columns. Works on both chrome and firefox now.

b17timjo:
When removing code, please delete it entirely instead of commenting it out; old code should be retreived via the version control system, not by uncommenting.

c17alepe:
gotcha, the code is now removed instead of commented out.

Issue #5633, opened 2019-04-01
Filter for students that have passed certain part of course in resulted


HGustavs:

a17pongo:
Closing pull request.
Because errors in solution was found by the developer.

PR #5754, opened 2019-04-09
Grupp b2019 #5633

a17oscno:
#5633 
Log in as Toddler and go to "Webbutveckling-datagrafik" then press "Results". Then test the "Sorting".
acknowledge that the VG option won't work due to there exists no duggas that can have VG on them and it seems like you can't grade any to VG either.


a17ludca:
FIlters out the grades not chosen but still shows the rows where there is no match. This is a good enough solution and an issue concerning the removal of empty rows has already been created. Therefore this issue is solved.

PR #5756, opened 2019-04-09
Update resulted.js

a17amama:
Commented out setTeacher value, cause the value only held NULL and was taking up unnecessary space


c17johha:
#5661

a17pongo:
https://issue_5661.gruppb.webug.his.se:20002/DuggaSys/resulted.php?cid=2&coursevers=97731

a17ludca:
Works on the standard port 20001 but not on 20002, I don't know if this is an issue we have to fix or not. The code provided works.

a17pongo:
I think this is an error in server configuration, workflow and folder structure on the web server will be refined to prevent this.

Issue #5695, opened 2019-04-08
Menu bar attached to the top of the page


a17drama:
Menu bar should be attached to the top of the page and not effected if the user scrolls.
![Edited](https://user-images.githubusercontent.com/46029340/55711159-6d27f580-59ec-11e9-91c7-56cc6b7e93d7.PNG)

PR #5758, opened 2019-04-09
Fix for #5695

a17celal:
The menu bar is now attached to the top of the page, even if the user scrolls the page.
I added some CSS to fix this.

#5695

### DIFF
--- a/DuggaSys/diagram.php
+++ b/DuggaSys/diagram.php
@@ -123,8 +123,6 @@
                         </div>
                     </div>
                 </div>
-
-                
                 <div class="menu-drop-down">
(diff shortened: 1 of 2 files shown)


a17jacsv:
The menu bar is now attached to the top of the page, even if the user scrolls the page as you presented. It works as it should. But "position: absolute;" is making the "move grid-button's" position is changed. 

_See picture below:_

Skärmavbild 2019-04-10 kl  10 53 12

It should be in the right corner like before your changings.

_See picture below:_
Skärmavbild 2019-04-10 kl  10 57 23

Issue #5740, opened 2019-04-09
install timeout


a17poner:
When using the installer after having made sure every prior required step has been properly filled in and all the dependencies installed sometimes the install process takes to long to run and causes a php timeout error. (had it happen 2 ou tof two times when installing on both laptop and home pc). Setting the php timeout setting to a higher value before installing and then setting it back to default is a way to fix the issue, doesn't make the process run faster but having an installation run for longer than 30 seconds doesn't seem unreasonable.

Issue #5652, opened 2019-04-05
New instanse should create coursesyspw.php


a17adajo:
Make install.php create the config file automaticly and remove the prompt about running this: 

sudo printf "<?php\ndefine(\"DB_USER\",\"nisse\");\ndefine(\"DB_PASSWORD\",\"123\");\ndefine(\"DB_HOST\",\"localhost\");\ndefine(\"DB_NAME\",\"lena\");\n?>" > /app/coursesyspw.php

PR #5759, opened 2019-04-09
Grupp c2019 5652 and 5740

a17poner:
fixes for #5652 and #5740
to test them you will need to either run on linux or use the windows fix that last I checked was pending a merge request in a different group, What i did was to go into the windows related issue and copy the code and replace the code in the current version but removed it before comitting, timeout might be hard to test unless your process takes a long time but the other fix should be a bit more easy to test and verify.

### DIFF
--- a/install/install.php
+++ b/install/install.php
@@ -6,6 +6,13 @@
 </head>
 <body>
     <?php
+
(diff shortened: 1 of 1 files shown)


a17poner:
Making a new one with new branch after the big weekly merge, closing this old one :(

Issue #5649, opened 2019-04-04
New instance should create loglena4.db if missing


b17timjo:
When setting up a new instance of lenasys, if the sqlite log (../log/loglena4.db) does not exist: create it, instead of silently failing to operate.

b17timjo:
"new PDO('sqlite:../../log/loglena4.db');"
in:
lenasys/Shared/basic.php:90
lenasys/Shared/resetpw.php:16 and 63
lenasys/Shared/loginlogout.php:31

b17timjo:
Suggestion: either the installer creates the database (possibly: `new SQLite3('<path-to>/loglena4.db')`) or each location of use takes some action.
Such actions could be checking for an existing database and silently skip attempting to use it (possibly add to some other error log), or perhaps create it (would creating it in this manner lead to data races? probably)

a17pongo:
has this issue been resolved with a17joaap fix?

a17joaap:
Yes!

PR #5760, opened 2019-04-09
Added a function that initializes the logging db in the install script.

a17joaap:
Addresses issue [#5649](https://github.com/HGustavs/LenaSYS/issues/5649)

### DIFF
--- a/Shared/basic.php
+++ b/Shared/basic.php
@@ -84,74 +84,9 @@ function swizzleArray(&$filepost) {
 }
 
 //---------------------------------------------------------------------------------------------------------------
-// connect log database - Open log database and create tables if they do not exist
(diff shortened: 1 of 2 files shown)


a17adajo:
I think the creation of the tables in loglena4.db should not be duplicated according to DRY (https://github.com/97-things/97-things-every-programmer-should-know/blob/master/en/thing_30/README.md)

With this commit SQL statements for the creation of the database will reside in both install.php and basic.php.

When the logfile schema is altered wich file should be updated? both? one of them? How can the future programmer now that there is two locations for the creation of the database in the codebase?

a17joaap:
Could cause problems if log database was for some reason destroyed after installation?

Issue #5685, opened 2019-04-05
Assign key numbers to variables for increased readability


a16tobfr:
diagram.js
<img width="323" alt="keyNumbers" src="https://user-images.githubusercontent.com/49199690/55629848-834b7100-57b4-11e9-8ea0-564e09d8c339.PNG">
Not clear what key each number represents, should assign the key numbers to variables to increase the readability of the code.

HGustavs:
it is crucial that this is declared in the same file, otherwise we are breaking the "no refactoring" clause...

I also recommend using const

a17okaoz:
I have found out what the key values are. Now I´m going to change them to variables

a17okaoz:
Issue #5685 Fixed. I have written a comment about each key variable about which key it is and its keyCode

a17okaoz:
Is it ok for me to change the variables for each key? I know that I will be break the no refactoring rule but if I do that it will increase the readability

PR #5764, opened 2019-04-10
Assign key numbers to variables for increased readability

a17okaoz:
#5685 I have commented beside every key and keyCode on witch key they represent. That will make the code more easy to read and understand.


a17drama:
This merge is outdated and is closed because of that. New pull request has been made merging with the correct version. @a17drama

Issue #5709, opened 2019-04-08
Variables checked against numbers


a16tobfr:
<img width="151" alt="toolbarState" src="https://user-images.githubusercontent.com/49199690/55723725-09142a00-5a0a-11e9-8703-ea7ab0da151e.PNG">
There are variables that are checked against numbers in if statements in various places within the diagram files. It is hard to get a grasp of what the numbers represents.
It would be better if the the numbers were mapped to variables that could be used instead.

HGustavs:
This is already an issue #5685 I think it was.

HGustavs:
Please use moustaches (~ <-- three of those) when quoting code.

Screenshots are for user interface stuff not for code.

~~~
This is how a moustache-encapsulated block looks!
~~~

a17okaoz:
Is it ok for me to change the variables here to? I have just written a comment on what the numbers stands for.

PR #5765, opened 2019-04-10
Variables checked against numbers 

a17okaoz:
#5709 I have inserted the values into variable that will represent each toolbar


a17drama:
This merge is outdated and is closed because of that. New pull request has been made merging with the correct version. @a17drama

a17jacsv:
The functionality works as expected. Though you have a lot of whitespaces. Fix that.

PR #5773, opened 2019-04-10
Grupp b2019#5633

a17oscno:
#5633 

Test the results editor as Toddler, there should be a filter option to see only different grades.


a17oscno:
Go into the Webbutveckling-datorgrafik course, it is the only one with already made assignments
If you want you can also distribute grades to test if the editor updates

PR #5786, opened 2019-04-10
Fixed cdirname() in install.php to not return double slashes.

a17robda:


a17robda:
Forgot to test this enough,

PR #5789, opened 2019-04-10
Added a function that initializes the logging db in the install script.

a17joaap:
Addresses issue [#5649](https://github.com/HGustavs/LenaSYS/issues/5760)


a17joaap:
Could cause problems if log database was for some reason destroyed after installation?

a17poner:
I Looked a bit into your previous conversation about this pr/issue and from my very limited previous experience with sqlite I think its supposed to be good practice to have the creat if not exists run more than once. So maybe having the db initialisation in install.php and then letting the tables get created in basic would be a good idea? That also solves any issues of the table dissapearing as you mentioned above.

a17joaap:
Hmm, I don't know what's best. Executive decision @a17andby ? Create tables if exists for loglena4 in both install.php and basic.php?

a17andby:
It doesn't hurt that you try to create the tables twice, as long as you use "CREATE TABLE IF NOT EXISTS"? 

So I think we should try to create the tables twice, ie both in `basic.php` and `install.php`, in order to possibly prevent future problems that may arise when eg. the database was for some reason destroyed after installation.

a17poner:
Could be a problem when you need to change the log database structure that you have to change in both places but that would be pretty unlikely. But install.php is only run once, basic.php is run on every page so it would be created if it doesn't exist each time the site is used which should make sure it always exists. I can review this in a bit to give you some time to either add it to basic or decide here how you want it to be 🍔

a17joaap:
Merge issue is because of `$initQuery = file_get_contents("SQL/init_db.sql");` on row 675. Can be easily resolved.

a17andby:
@a17joaap Can you elaborate more how we can fix this conflict?

a17joaap:
It's conflicting because we merged a fix that changed 
`$initQuery = file_get_contents("SQL/init_db.sql")`
to
`$initQuery = file_get_contents("../Shared/SQL/init_db.sql")`

I didn't want to change it because it would mess the testing up, but I'll go ahead and do it now

a17poner:
Where do you see what the conflict is? Im guessing its because I have a few changes on the main C branch already that does and adds things in the install script. Creating the coursyspw file and changing the timeout. The implementations themselves shouldn't be to conflicting, but the line of code that was change will probably be roughly the same. I also removed the div that had the warning for coursyspw but not the header so that should still be fine I believe.

a17joaap:
There we go, no more conflict! 💯 

It said on this page that it was install/install.php @a17poner

Issue #5774, opened 2019-04-10
Right clicking on FAB in sectioned.php should not add a new item


b17carma:
If you right click the FAB in sectioned.php it adds an item but this makes it hard to inspect the element since it will add a new dugga when you don't want it to.

PR #5792, opened 2019-04-10
Only act on FAB button presses when left mouse button is  clicked in …

b17carma:
…sectioned.js


b17carma:
Fix for issue https://github.com/HGustavs/LenaSYS/issues/5774

Issue #5723, opened 2019-04-08
Change that the canvas-origo can't move


a17vikab:
The canvas should not be able to be moved so both X & Y becomes negative. It should only be moved when X & Y are positive. There is no reason for it to show a negative window of the canvas when it can scale endlessly.

#5145 root issue

HGustavs:
This seems like a good fix. Hopefully this will resolve the zoom and pan issues.

There has been several attempts at this functionality.

a17drama:
Expand this issue to locking the canvas totally. No moving of canvas should be possible. Making a comment in the existing pull request for this issue as well.

a17drama:
This is one part of the solution for #5826 . Canvas should be static and not move.

a17jacsv:
I close this issue because there is a better solution on the go. #5826

PR #5793, opened 2019-04-10
Fixed issue #5723

a16jenja:
Issue: #5723 
Co-author: @a17okaoz 

In diagram_mouse if the coordinates in the info toolbar get below 0 it will be set to 0.

In diagram if the mouseDiff get below 0 AND the coordinates are at 0 in x or y the canvas will not move.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -781,6 +781,14 @@ window.addEventListener('resize', canvasSize);
 function updateGraphics() {
     ctx.clearRect(sx, sy, (widthWindow / zoomValue), (heightWindow / zoomValue));
     if (moveValue == 1) {
+      // If the canvas is moved towards the left and x coordinate is at 0, dont move the canvas
+      if(mouseDiffX < 0 && sx == 0){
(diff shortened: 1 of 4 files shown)


a17drama:
Issue for this has been updated. Please check the updates.

a16jenja:
Issue: #5723 

No longer able to move the canvas at all and took the liberty to comment out labels and anything else relevant to moving the canvas.

There is a lot of dead code in the system now since I didn't remove anything that was relevant to moving the canvas, I just removed the option to even enter that moveable mode by pressing space.

Let me know if I should remove all the dead code. Since it might be useful to keep, if in the future we want to reimplement the movement of the canvas again.

a17drama:
Prioritize this over #5826 and read the comments from the developer. Important arguments there.

a17jacsv:
Yes, I agree. We put this matter on hold for now.

a17jacsv:
I close this pull-request because there is a better solution on the go. https://github.com/HGustavs/LenaSYS/issues/5826

Issue #5639, opened 2019-04-03
URL changes in diagram view


HGustavs:
When I click the menu items (not the tool bar afaik?) the url adds a hash DuggaSys/diagram.php#

This is bad since it affects the scroll position in the window.

I assume this is due to a-links being used as anchors in the menu?

There must be a better wat to make markup that is clickable without wrecking the scolling in the window?

a17jonro:
Are we able to change the HTML markup? from an anchor to eg a span? You could also remove the Href attribute but that does kinda defeat the purpose of it being an anchor in the first place.

HGustavs:
Yes no problem, this is a local change that will only affect this version ...

also, it fixes a bug.

PR #5800, opened 2019-04-10
Fix for issue #5639

b17robev:
#5639


a17jacsv:
Good solution! Approved.

PR #5801, opened 2019-04-10
Snap to grid menu option moved to "Align" category - Fix for #5763

### LINKED ISSUE #5763: Align to grid is in the wrong menu
Edited
Snap to grid is an alignment when drawing and not a type of view. This should be moved to the alignment menu.

a14oscna:
The menu option was moved from the "View" category to "Align category as seen in the image: 
grid snap fix
Fix #5763


a17jacsv:
Approved.

Issue #5675, opened 2019-04-05
Diagram property reworking


HGustavs:
Many of the properties in the diagram are optional....

such as color etc

I suggest removing all optional properties from the diagram objects for example color

I suggest using an "attributes array" instead of

~~~
{"diagram":[
{"kind":2,"targeted":false,"symbolkind":3,"operations":[],"attributes":[],"textLines":[],
"textsize":14,"symbolColor":"#FFB74D","strokeColor":"#000000","font":"Arial","lineWidth":2,
"fontColor":"#000000","name":"Entity0","key_type":"Strong","sizeOftext":"Tiny",
"textAlign":"center","shadowBlur":10,"shadowOffsetX":3,"shadowOffsetY":6,
"shadowColor":"rgba(0, 0, 0, 0.3)","cardinality":[{"value":null,"isCorrectSide":null,"symbolKind":null}],"locked":false,"connectorTop":[],"connectorBottom":[],"connectorLeft":[],"connectorRight":[{"from":41,"to":11},{"from":18,"to":19}],"topLeft":0,"bottomRight":1,"centerPoint":2,"arity":[],"object_type":"","isHovered":false}
~~~

I would use
~~~
{"diagram":[
{"kind":2,"targeted":false,"symbolkind":3,"operations":[],"attributes":[],"textLines":[],
"cardinality":[{"value":null,"isCorrectSide":null,"symbolKind":null}],"locked":false,"connectorTop":[],"connectorBottom":[],"connectorLeft":[],"connectorRight":[{"from":41,"to":11},{"from":18,"to":19}],"topLeft":0,"bottomRight":1,"centerPoint":2,"arity":[],"object_type":"","isHovered":false,properties:["symbolColor":"#FFB74D"]}
~~~

this way only necessary properties occupy space, and if we want to use a non-standard color for some entity e.g. symbolcolor we add this to the properties array. Setting shadowoffsets and so on every time is a waste of space IMHO.

HGustavs:
Technically a refactoring but we need to reduce the data size of the diagrams before using diagrams becomes useful....

PR #5802, opened 2019-04-10
Fix issue #5675

a17vikab:
Issue: #5675 

As the issue stated, the goal was to reduce the amount of data the variables takes and we solved it by moving alot of variable into a "properties" array with properties.

The dangerous part of this issue is that we are refactoring alot of things in diagram_symbol.js, some things in diagram.js and diagram_dialog.js. (Check commit-history for a complete history.)

Make sure that you, the tester, make sure that everything works as it should and if something breaks when merging with other branshes, this issue have a large possibility for being the reason if it isn't tested correctly and solved correctly.


a17jacsv:
It seems to work as expected. I agree with c17matjo. Good job!

Issue #5655, opened 2019-04-05
Logout button is unclear


b17annni:
Button is unclear that it is an logout button until hover. Accidental logout is a problem.

HGustavs:
Mockup for how this can be improved?

a17maxsj:
![New Project](https://user-images.githubusercontent.com/49199993/55624378-f1d50280-57a5-11e9-93a4-a47b788e0180.png)

Something like this? Or some other way to show the user being logged in as well as indicating it being a logout button.

HGustavs:
That red/green mess looks horrible perhaps we can change the "man" icon to a login / logoff icon instead?

a16joare:
Removed the hover function, added a fixed color to the button and changed the logout button color. 
When you login: ![image](https://user-images.githubusercontent.com/49194898/55805164-2adee100-5ade-11e9-9356-02a8ea435fa2.png)

When you logout: ![image](https://user-images.githubusercontent.com/49194898/55806508-a6419200-5ae0-11e9-870c-bea9f46b3853.png)

what do you think?

a16joare:
I will close this one but look at #5812

a17maxsj:
Needs some further work

a17maxsj:
done.

PR #5812, opened 2019-04-11
Fix for "Logout button is unclear #5655"

a16joare:
Removed the hover function, added a fixed color to the button and changed the logout button color.

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -651,19 +651,11 @@ header td {
 }
 
 .loggedout {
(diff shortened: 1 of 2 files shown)


a16joare:
I can support it if the hover fuction just changes the color of the image slightly.  I still think that the image chaning hover should be removed totally.

b16johso:
I think it definitely needs a hover effect. Imo, it should change from green to red on hover when logged in and the other way around when logged out. I agree with the icon change though.

PR #5814, opened 2019-04-11
Removed obsolete comment - Fix for #5707

a14oscna:
Removed the comment in question on line 838 in diagram_symbol.js
Fix for #5707


a17drama:
This pull request was missed because lack of tags and the changes have to done again with the new refactored code.

a14oscna:
Will do right now!

PR #5815, opened 2019-04-11
Fix#2 for issue #5746

a17vikab:
Old pull request: https://github.com/HGustavs/LenaSYS/pull/5751
Issue: #5746 

Added two new variables that tracks the mouse coordinates in the canvas and later change the innerHTML so it always shows on screen.


a17jacsv:
Approved. Nothing to add, good job!

PR #5817, opened 2019-04-11
Update sectioned.php fixed upper plus(add new item)

a17poner:
Fixed issue #5696 a second time for new branch, the other solution with creating a specific one might be valid as well but this soluiton works the same was as normally clicking the big orange fab button so its more reused.


a17joaap:
Issue fixed

Issue #5712, opened 2019-04-08
Unused/empty functions


a17celal:
There are some functions in diagram_IOHandler.js that are unused or without content. If they are not used they should be deleted.

![Capture5](https://user-images.githubusercontent.com/37792271/55724210-33b2b280-5a0b-11e9-8283-922521e67806.PNG)

HGustavs:
No idead what this code does... 

There used to be a button that generated a demo diagram... perhaps leftovers from this?

PR #5822, opened 2019-04-11
Fix for issue #5712

a17vikab:
Issue: #5712 

Only removed two unused functions and added the correct spacing between the nearby functions.


a17jacsv:
OK.

Issue #5694, opened 2019-04-08
Developer mode toggle clarified


a17drama:
View menu has a "Developer mode" that is a toggle function. It's not so clear if this is active or not. Solution for this would be to make a marking (suggestion in the green circle in screenshot) in the menu that shows if it's selected.  Another solution is to write out the mode that is currently used in a "status bar" that is attached to the bottom of the page.
![Edited](https://user-images.githubusercontent.com/46029340/55711224-92b4ff00-59ec-11e9-88c5-2657fe9d2378.PNG)

a17drama:
I missed that the snap to grid feature should have the same solution. Do both while doing this issue.

a17drama:
There will be more functions with toggle implemented from other issues and it would be preferable to be able to just attach this toggle indication to these as well. In other words it would be preferable no to hard code this for each choice if possible.

c17matjo:
![image](https://user-images.githubusercontent.com/49195031/55873968-4e169880-5b91-11e9-9c28-186d791d7de8.png)
Screenshot of my implementation. The checkboxes are not specified in the html document but are instead added to the page from javascript in a setCheckbox method. This can be called on any element that needs a checkbox element.

a17jacsv:
Content from this pull-request was not included after the merge and needs to do over again.

c17matjo:
Turns out that the merge was correct but that some other commit (https://github.com/HGustavs/LenaSYS/commit/3d30fb1f3d377346c91e40f4b8a74f9605c35ab6) in gruppA2019W16 had changed the menu-item implementation enough to break checkboxes.

c17matjo:
https://github.com/HGustavs/LenaSYS/pull/5904 should fix the issue

a17jacsv:
Okay, I understand. We have some guys working with trying to decrease the number of files regarding the diagram. Unfortunately, you'll have to do this solution again when they are done as we can't editing at the moment. I'll be in touch.

c17matjo:
Okay 👌

a17jacsv:
Now you can fix this issue with a refreshed branch from our branch (gruppA2019W16) and create a new pull request.

PR #5824, opened 2019-04-11
Add checkboxes to diagram menu - Fix for issue #5694

c17matjo:
image
The checkboxes are not specified in the html document but are instead added to the page from javascript in a setCheckbox method. This can be called on any element that needs a checkbox element.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -730,6 +730,7 @@ function toggleGrid() {
     } else {
         snapToGrid = false;
     }
+    setCheckbox($("a:contains('Snap to grid')"), snapToGrid);
 }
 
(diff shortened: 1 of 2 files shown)


a17jacsv:
This issue-branch is from the old branch and should be rewritten in a new issue-branch from our updated gruppA2019W16-branch. Please fix it.

c17matjo:
The branch was rebased onto the gruppA2019W16 branch using the git-rebase command so deleting it and creating a new one shouldn't be necessary.

a17jacsv:
Ok, that should work.

a17jacsv:
It works as expected. There were some conflicts but they are now solved. Good job!

PR #5825, opened 2019-04-11
Update dugga open pdf new second fix

a17poner:
Any pdf submission should be able to open in a new tab using the new added open external button.
This is the second pull request for this, the first one is at #5738 Made a new one now that we have a new base branch. Currently on the master the resulted.php is broken due to some javascript so it might be a little bit hard to test but the icon also shows up and works the same way when looking back at a already submitted pdf (i.e submit a pdf, go back onto the dugga and see if the button is there next to download and if clicking it opens a new tab with the pdf in it.

### DIFF
--- a/Shared/SortableTableLibrary/sortableTable.js
+++ b/Shared/SortableTableLibrary/sortableTable.js
@@ -559,7 +559,6 @@ function SortableTable(param)
         			document.getElementById(children[i].id.slice(0, -1)+"f").style.boxSizing = "border-box";
       		}
           document.getElementById(this.tableid+DELIMITER+"tblhead_mh").style.height = Math.round(document.getElementById(this.tableid+DELIMITER+"tblhead").getBoundingClientRect().height)+"px";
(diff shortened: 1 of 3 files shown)


a17alevu:
I went into "Datorgrafik" course and loaded up a pdf file to test if the button work. The buttons works fine and will open the pdf file on a new tab in the browser. The button has no image though, as you can see on the picture below. So only the "visual effect" is not accomplished 
githubissue

a17poner:
Updated with the picture for the graphical part. Would request a re-review but you didn't actually use the github review system the first time (I think its at the top right) so just a review I guess 😄  But remember its good to use the github review system regardless of accepting or failing for both documentational reasons and notifications as it will show what chosen and make it easier for groupleaders to quickly see and assess and know to merge or not without. 😄

a17andby:
Fixed and merged

a17alevu:
Now it works. The buttons work and the images are showing. 


image

Issue #5782, opened 2019-04-10
Pressing escape in sectioned.js removes the FAB used to add items


b17carma:
When the escape key is pressed in sectioned.js with no dialog open, the fab used to add new items is removed which does not appear to be intended behaviour

HGustavs:
This must be a bug

b17carma:
This issue seems to have reappeared?

a17robes:
Just tested it again and there are no problems for me, works like the fix said it would! Can anyone else confirm? @a17alevu ?

a17robes:
Might be that the edited code have been changed after my fix of this issue. Going to check this out!

b17carma:
Fixed in https://github.com/HGustavs/LenaSYS/pull/6170

PR #5827, opened 2019-04-11
Fixed the issue #5782

a17robes:
Removed "e.style.display= "none"; in dugga.js,
in the if(index_highest>0). It caused the escape button to remove/hide the fabBtn.

https://github.com/HGustavs/LenaSYS/issues/5782


a17robes:
Please test this again. I think i solved it.

PR #5829, opened 2019-04-11
Gruppc2019W16_1887 Codeviewer Loader

a17fellu:
Added a Loading GIF on codeviewer.php that displays the loading gif as the page is loading its content in the background. When the content is loaded it will hide the loader GIF and show the content instead


a17fellu:
I will add this to the other pages in question in the same way and update the issue branch

a17andby:
Requested a re-review since the pull and been updated since last review.

Issue #5711, opened 2019-04-08
Large code segments are in comments


a16tobfr:
<img width="259" alt="outCommentedCode2" src="https://user-images.githubusercontent.com/49199690/55724107-fd753300-5a0a-11e9-9f9f-10e7dce4f6aa.PNG">
There are a lot of out commented code in the diagram files. If not used, the segments should be deleted.

a17adajo:
I agree, if needed, the code can be retrived in the git history.

HGustavs:
These screenshots are barely readable ... when citing code it is more convenient to paste code encapsulated with thee moustaches ( ~ <- three of those ) and then a code block and then three more moustaches.

And then we get a nice block
~~~
code
~~~

a17drama:
I agree to disagree. Screenshot is readable if you click the picture. But for convenience sake it's better to implement the code into the comment. Can be done so in the future.

HGustavs:
It is relatively common for people to have problems seeing colors which makes low-contrast text very difficult to read if it is in the darker side of the spectrum... it is less of a problem if it is black on white or white on black... so it is an accessibility issue.

PR #5830, opened 2019-04-11
Fix for issue #5711

a17jonro:
#5711


a17jacsv:
Good addition. There were some conflicts but they are now solved.

Issue #5747, opened 2019-04-09
Cardinality numbers can be placed behind lines


a16tobfr:
<img width="174" alt="kardinalitet" src="https://user-images.githubusercontent.com/49199690/55796419-450ec400-5aca-11e9-918d-fdc0f356c426.PNG">
It would be good if the cardinality was visible when situations like this occur.

HGustavs:
Three ways of doing this (are there more?)
* Draw black text with white outline
* Draw white square around text
* Measuring when line is likely to pinch text box and moving it accordingly (comparatively difficult)

Option 3 is my favourite ...

<img width="397" alt="Screenshot 2019-04-09 22 12 28" src="https://user-images.githubusercontent.com/6978801/55831911-96dc3c00-5b14-11e9-9009-f29adbfab057.png">

If one figures out how far down the line the width between the side of the rectangle and the line is wider than the width of the text box A ... we know how far out we need to place the box ...

Then we only need to do this equation for all 8 pinch directions and text box placements ...

HGustavs:
My vote is that we can try option 1 and see if it looks crappy, and move option 3 to the back burner and work on that if we are running out of issues towards the end of the project.

PR #5831, opened 2019-04-11
Fix for issue #5747

a17alfcl:
Seems to be working, copied code from old merge but didn't seem to cause any problems #5747


a17jacsv:
Approved. Good job!

PR #5836, opened 2019-04-11
Changed textfield constraints for root password - Fix for #5720

a14oscna:
Added the root password textfield as an exception for the function that checks if each textfield is filled before continuing the installation process. You can now complete the installation without entering a root password. Fix for #5720


a17jacsv:
OK.

Issue #5662, opened 2019-04-05
"Times Graded: x" taking up space in result view


c17johha:
![52473472-2d809100-2b96-11e9-8359-37aaaab5d133](https://user-images.githubusercontent.com/49199825/55618435-ceef2200-5796-11e9-8b36-c40bf23eca96.png)

I would like to add the numer of marked times as just a number enclosed in parenthesis next to the result in some place e.g. (1) or (2) or (3)

HGustavs:
Do we have a screenshot of a proposed solution? 

Question is whether the (N) number is better placed aftetr the VG or after 12:40:17

a17alian:
<img width="977" alt="Skärmavbild 2019-04-08 kl  10 23 30" src="https://user-images.githubusercontent.com/37792256/55709420-8c248880-59e8-11e9-8887-79d2b46a4469.png">

Proposed solutions for "Times Graded". Solution 1 is the preferred approach with only a single number  and the removal of "Times Graded", but if you would rather have the "Times Graded" specified, the solution 2 is what we propose.

PR #5838, opened 2019-04-11
Grupp b2019 w#5662

a17amama:
Moved Times graded next to the grading buttons, added the class "WriteOutTimesGraded"  for styling, and added margin-right: 10px; inside the stylesheet at .VGh


a17amama:
#5662 

https://amanda.gruppb.www.webug.his.se:20002/DuggaSys/courseed.php

a17ludca:
When the number is added, the gradebuttons are pushed to the right and as an example in "Geometri 2", the buttons are too close to the edge. In other words, columns that are narrow will look odd since the "U"-button will hit the edge.

Issue #5614, opened 2019-01-18
e-mail all students taking course


HGustavs:
We need a function for e-mailing all students in a course by generating a compose email in the same way as we do reset password but with many recipients

HGustavs:
variants:

* Mail all students on a specific course version (e.g. all registered to latest version)
* Mail all students in all course versions
* Mail all students who have not passed the course

HGustavs:
This is very important and considered very high priority.

HGustavs:
email students that have not had a result in specific part of the course e.g. home exam.

...

There is a dialog for exporting to csv at top of screen, this should also work as source for email. (there is other issue on this?)

a17anter:
There is now a "mail" button which opens the default mail application and fills in all e-mail adresses for the participating students with a subject.

Participating students are filtered by current URL keys "courseid" and "coursevers".

HGustavs:
All visible in result view

PR #5843, opened 2019-04-11
Grupp b2019 w16#5614

a17anter:
#5614


c17johha:
Branch was not up to date

PR #5847, opened 2019-04-11
Pull request to branch gruppB2019W16 #5614

a17anter:
Code updated with the new branch from the complete new GruppB2019W16 branch.
#5614


a17adasy:
Tested. The function is there and works. But if you try sending a mail in a course with no students the mail still opens. Maybe it could be a error message here instead.

a17pongo:
I will merge this and approve it but create another issue for when you try to send a mail with no students.

Issue #5833, opened 2019-04-11
After merge an error is thrown from sortable table on result and access and the "pens" on test no longer works


a17poner:
![image](https://user-images.githubusercontent.com/49194938/55945811-ccd00c00-5c4b-11e9-8451-811ada3316a4.png)

When going onto a page that has a sortable table the first error in the console is thrown. When pressing a pen the second error is thrown. Access however is fine despite having a sortable table but it doesn't have an edit pen the same way the other two tables do.

HGustavs:
Is it any sortable table or just this specific sortable table?

a17poner:
It's accessed and test, both of them when viewed throw the first error shown in the screenshot and then when the pen is clicked they both throw the second error, not yet sure what causes this since it happens in two different places I assume it will be something in sharedtable.js. Git blaming that file non of the rows listed in the stacktrace have been altered in the last 8 months so guessing it would be something above in the chain that causes it.

HGustavs:
It this related to #5853 ?? seems very similar?

a17poner:
Yes this is the exact same issue as #5853. There is a pull requset pending for this issue but if @a17joaap need to be able to click the pen to test your own issue currently being worked on all you have to do is remove the line at 562 that tries to add styling to tableid+delimiter+"tableheadmhv" (pay attention to the v at the end) and the error should go away.

a17andby:
Fixed and merged. Closing.

PR #5854, opened 2019-04-11
Removed styling on element in sortableTable.js

a17poner:
removed styling that was causing #5833
sortable tables should now work properly and the pen icon should work (no errors thrown in console)


a17poner:
What errors are you getting and where? if you could send a printscreen of the console or something :) This should solve the errors found in issue #5833 (there is a screenshot in it)

a17robda:
err

When clicking the pen icon does not do anything and throws errors in console.

a17robda:
However it could be a problem on my end using the school servers.

a17poner:
Are you sure you checked out the correct branch? row 562 in sortabletable.js shouldn't be there and is the row that I removed in my commit/fix. Shows it as removed int he "file changed" tab as well so im pretty sure I pull requested the correct branch. Would youu mind just doublechecking with git status or so to make sure you are on the branch "gruppC2019W16_5833"

Issue #5705, opened 2019-04-08
Swimlanes should be removed or merged into sectioned


HGustavs:
The swimlanes part of section-ed should be part of the sectioned js file.

Currently there is a service swimlanes.php that should not be used?

If it is used it should be merged into sectioned.

If it is not used it should be removed from the repo.

PR #5862, opened 2019-04-12
#5705: remove unused swimlane

b17timjo:
- swimlane.php was only used by navheader.php;
- -- it consisted of an unused div and a reference to swimlane.js
- swimlane.js was referenced in an ajax handler in dugga.js;
- -- said ajax call was only made from swimlane.js itself
- no functions declared in swimlanes.js were referenced elsewhere
- swimlaneservice.php only called by itself


a17pongo:
This will reuquire some extensive testing to make sure nothing on other sites where messed up.

b17nicsv:
The commit were only removal of the unnecessary swimlane files and styling in css so code holds the same standard as before. The functionality of the swimlanes is still there and still works. Other sites still works as they did. Safe to push into GruppB2019W16 branch.

Issue #5804, opened 2019-04-10
Menu bar moving when resizing inspection window


a17drama:
![Edited](https://user-images.githubusercontent.com/46029340/55885148-af4a6600-5ba9-11e9-8699-3d6eaaf46632.PNG)
Menu bar is moving (Marked with red arrows) when inspection window is resized (Line marked with green is draged). This has been tested in both Edge and Chrome and it's the same phenomena. Menu line should be static.

HGustavs:
I would remove the white space above completely.

a17drama:
Normally there is no white space. It only shows up when the inspection window is dragged out like in the screenshot.

Issue #5821, opened 2019-04-11
Menu items position is changed when changing direction of the toolbar


a17jacsv:
Menu items position is changed when changing direction of the toolbar. It should not affect this menu **only** the toolbar.

_See picture below:_
<img width="268" alt="Skärmavbild 2019-04-11 kl  10 11 37" src="https://user-images.githubusercontent.com/37792795/55941391-68f51580-5c42-11e9-99cd-6d9ec4960e37.png">

_See picture below:_
<img width="408" alt="Skärmavbild 2019-04-11 kl  10 15 59" src="https://user-images.githubusercontent.com/37792795/55941634-de60e600-5c42-11e9-9e48-2833bfa405ec.png">

a17drama:
This issue is redundant.

a17drama:
This could have to do with #5804 .

PR #5863, opened 2019-04-12
Fixed issue #5804 and #5821 (which was referenced in issue #5804)

a16jenja:
Issue #5804 and #5821 is fixed.

The menu bar won't float downwards anymore when rezising the window. Got help from @a17okaoz with this issue.

The menu bar won't be affected anymore when changing the direction of the toolbar. The problem was that the toolbar and the menu bar used the same span for the labels, which affected css.


a17jacsv:
Works as expected. Good work!

Issue #5864, opened 2019-04-12
Contirbution.js shows data for a student


a17maxsj:
userid in contribution.js has been set to a student id, this is not ok and should be set to @HGustavs

a17fellu:
It is not exactly correct to choose a specific user at all. This should not be hardcoded. It should be pointing to the user id of the user that is logged in

a17joaap:
Henrik explains in this comment: https://github.com/HGustavs/LenaSYS/issues/5806#issuecomment-481846720

Last year they got a database with HGustavs as one of the users, we will be given a similar database next week

a17adajo:
We will get the database this week

a17maxsj:
This has been fixed now. Closing

PR #5865, opened 2019-04-12
Fixed userid from a StudentID to "HGustavs" #5864

a17alevu:


a17fellu:
The user data it shows should not be hardcoded. It should point to the user that is currently logged in. HGustav is not a user in the database

Issue #5534, opened 2018-05-23
Problem with drawing squares 


a16juleh:
If you are drawing squares, accidentally draw outside the diagram and then keep on drawing squares in the diagram, the drawing collapses. It starts to draw the squares from where the last click happened instead of where you're drawing (see picture below).

![33186831_10209722164250115_6352397969567776768_n](https://user-images.githubusercontent.com/37794632/40412959-f4dd84dc-5e74-11e8-9bfc-6ecd93a0a79b.png)

PR #5869, opened 2019-04-12
Fixed the drawing squares bug - Fix for #5534

a14oscna:
The problem behind the issue was that the program did not register that the user lifted the mouse key and instead believed it was still pressed when it wasnt. This reversed the order of execution in events which bugged out the square drawing logic. 

Solution: Added the onmouseup listener that calls the mouseupevt function to the body tag.

Fix for #5534


a17jacsv:
Works as expected. Good job!

Issue #5872, opened 2019-04-12
Syntax error in resulted.js


b17carma:
There appears to be a syntax error in resulted.js which completely breaks it as shown in the image below

![image](https://user-images.githubusercontent.com/37834592/56029756-661a2380-5d1b-11e9-9c33-32c9c5a8c27a.png)

PR #5873, opened 2019-04-12
Removed unreachable code with syntax error & improved overall indentation in resulted.js

b17carma:
Fix for issue https://github.com/HGustavs/LenaSYS/issues/5872

### DIFF
--- a/DuggaSys/resulted.js
+++ b/DuggaSys/resulted.js
@@ -1,655 +1,719 @@
-
 /********************************************************************************
-   Globals
-*********************************************************************************/
+ Globals
+ *********************************************************************************/
 var sessionkind = 0;
 var querystring = parseGet();
 var filez;
 var mmx = 0, mmy = 0;
 var msx = 0, msy = 0;
 var rProbe = null;
-var subheading=0;
+var subheading = 0;
 var allData;
-var allowedRegradeTime = 24*60*60*1000;
+var allowedRegradeTime = 24 * 60 * 60 * 1000;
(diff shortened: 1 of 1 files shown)


DuggaSys/resulted.js
```diff
@@ -770,84 +841,82 @@ function renderCell(col,celldata,cellid) {
...
+			if (celldata.ishere === true && celldata.timesGraded !== 0) {
+				str += "Times Graded: " + celldata.timesGraded;
+			}
 			str += "</div>";
 		}
 		str += "</div>";
 		return str;
 	}
-
-  // When Filtering is activated then this "hides" all other data than what is specified.
```
b17carma: Code segment with syntax error

DuggaSys/resulted.js
```diff
@@ -770,84 +841,82 @@ function renderCell(col,celldata,cellid) {
...
-  else {
```
b17carma: Code segment with syntax error (lines 842-850)

a17poner: This is supposed to be fixed plus fixing the issue(adding a feature) that caused it in this branch https://github.com/HGustavs/LenaSYS/tree/TempBranchResultMerge.
Its the one that went bad during the big merge. GruppB already merged this into their branch so I'd suggest as a fix for your issue that you instead create a new pull request from this branch so we at gruppC also get the same fix + some result table functionality to hopefully minimize conflicts once we do the next merge

Issue #5693, opened 2019-04-08
The install script should create the database in its final form without the use of ALTER statements.


a17adajo:
At the moment the database are created with CREATE statements and then at the bottom some tables are then altered.
Those tables should be created as they should be directly wich would simplify the script.

Example of lines that could be removed:

635: alter table user_course alter column result SET DEFAULT 0.0;

HGustavs:
The alters should be there as documentation ... ... but in deltachange.sql

The process is, if the running system presents an SQL error, this means that the tables have been augmented compared to the running system. The error message and the alter table should then be found in deltachange.sql

The init_db sql should be the "current" create-at-once-database, any alters necessary for patching any older system running an older database should be available in deltachange.sql together with the corresponding error message.

If the alter table in the init_db is NOT of that kind, but just a result of the work process or the export of some tool... then it should be incorporated back into init_db.sql

HGustavs:
Running the init_db is the edge case and the running system with the "old" database is the main case. ...In many other development paradigms this is the other way around...

a17poner:
merged.

PR #5880, opened 2019-04-12
New time sheet dugga

a17joaap:
Part of issue [5628](https://github.com/HGustavs/LenaSYS/issues/5628), needs further feedback from the client, but is a starting point for further development.

The dugga is under 'Webbutveckling - datorgrafik' and is called 'Daily Minutes'.

You can make any dugga a time sheet dugga if you open the test editor for the course and change the template to 'daily-minutes'. You also need to open the variant editor (click the pencil) and change the submission type to 'Timesheet'.


a17poner:
Should we wait for henrik to respond with the feedback you mentioned before reviewing or do you want me to review it as is and then potentially have it merged by bylund before you get feedback? :) Also minor thing that im not sure of but shouldn't database additions/changes be in the delta sql for some sort of history or am I mistaken and what you did by putting the new table in init_db correct? :)

a17joaap:
I think we can review it and possibly merge, and any feedback from Henrik could be implemented later on. Not sure about deltachange, but it sounds right! I'll fix that

a17joaap:
Will take a look and fix!

a17joaap:
Fixed clearing input when adding new row. Solution goes against the code standard, but it's simply impossible to preserve that data when using strings and innerHTML.

Added a button on new rows for removing the row. It's very ugly but could be improved later on.
removerow

Added the submission preview that dugga uses, so the user can see his/her latest upload,
feedback


Lastly, added the 'required' keyword to the inputs which does not accept empty values. This is supported by all modern browsers (and IE 10.0!). If we have any cavemen using the site, they will just get a SQL error from filereceive_dugga.php if they submit empty values.

a17joaap:
Oops, missed the required attribute on new rows. Not sure why it's not working locally, are the files being saved in DuggaSys/submissions/2/97732/1/*username*/ ?

a17poner:
image
Yeah they're added to the folder but I still can't see them.
P:\xampp\htdocs\LenaXampp\LenaSYS\DuggaSys\submissions\2\97732\21\Kong_Toddler_Toddler
is where they're saved.
Altough everything else seem to work now with required

a17joaap:
Changed the adding of new rows to conform to the code standard if you want to take a look. Not sure about your issue, it's handled by the findfilevers function in dugga.js (row 1308).

a17joaap:
> When uploading I get the error shown in the printscreen above
Do you get that on the server too? Or only local? Don't know what to do about that.

> The form has a fixed size and on smaller screens it goes off screen
I vote we make this a separate issue, this pull request is already too big by Henriks standards.

I fixed the issue when removing middle row and adding a new one.

a17joaap:
I looked around a bit and tried to recreate the issue with the file uploads. I noticed that if you omit '"fieldname"="timesheet"' you get a similar error, since 'findfilevers' uses the fieldname as a parameter.

a17poner:
Yeah agree, this issue itself was way to big to begin with, so styling, maybe the thing with the fileverse to see uploads if we can find a reliable way to recreate it can all become issues. Also thinking that maybe having each upload become one file with several rows instead of one file per submitted row could be a good issue for the future since it would be easier that way to keep working with it on the big contribution update/feature we will work on later in the project. Im still seeing that php error when I submit/upload. But the the adding and removing rows works properly now. I will just double check to see what it is on my end that could cause it before submitting the actual review. Also yeah Im just doing work and testing locally, I haven't done a single test on the webserver yet (haven't even set myself up with an account).
image

a17joaap:
Sounds good! 
The uploaded files actually already contain all rows that are in the table when pressing upload

a17poner:
Not sure how to test this currently, I'll try it on my computer at home tomorrow and if I still get the error I'll just submit an issue about it I guess, I'll approve it, the styling and possibly something to do with seeing uploaded files could make for issues as well.

a17joaap:
Moved the timesheet table creation to init_db since deltachange is used for [other things](https://github.com/HGustavs/LenaSYS/issues/5693#issuecomment-480734934). Merge conflict can be resolved by clicking resolve conflicts.

Issue #5868, opened 2019-04-12
Merge files related to diagram


a17alfcl:
Decrease the amount of files related to diagram functions.
Seven files are currently related to diagram and some need to be removed before remaking the canvas structure.

> "There are too many files" - Henrik 2019

PR #5909, opened 2019-04-15
Fix for issue #5868 Merge files related to diagram

a17vikab:
Issue: #5868 

Merged js-files into diagram.js and diagram_objects.js.


a17vikab:
There are some issues that didn't work before we did this fix and therefore, we know that new issues need to be created. Make sure that you create issues of problems that not already exist.

a17jacsv:
Approved. Good job!

Issue #5619, opened 2019-01-24
Section editor styling


HGustavs:
* We want a subtle difference between examples and other links so students can see the kind of link without navigating to that link. 
* We also want a subtle difference between internal and external (to other non-university pages) links

Preferrably very subtle, not using icons or other exceedingly overt notations.

c17julka:
![image](https://user-images.githubusercontent.com/37792321/56132177-9f57cb00-5f89-11e9-81d5-e58a05ad8e98.png) ![image](https://user-images.githubusercontent.com/37792321/56132216-b5fe2200-5f89-11e9-8fbc-c005c29ed2d6.png)
Fixed and merged. Above pics shows change in text color when hovering over certain links (example links, external links and internal links. Pics show example link + internal link).

PR #5926, opened 2019-04-15
Pull request for issue #5619

c17julka:
Solution for #5619

Test site: https://c17julka.gruppb.webug.his.se:20002

Changes are on sectioned.js (bottom of file) and style.css (rows 396 - 410)
Added functions which add classes to certain links, which have a hover colour effect.


c17julka:
Since you can't define if the added link on sectioned, you can't tell if it's external or internal. Which means testing of this would go through the main dugga page on the console.

c17alepe:
Issue tested and approved. The links change color when being hovered in the sectioned site. If Henrik approves the hover solution it should be good to go.

Issue #5910, opened 2019-04-15
Icons: Icon_Cross.svg and Icon_Tick.svg 404 error


a17andby:
When clicking on a `First name` and `Last name` record in the `accessed.php` table the icons `Icon_Cross.svg` and `Icon_Tick.svg` return a 404 error.

![Screenshot from 2019-04-15 09-58-48](https://user-images.githubusercontent.com/49235724/56116259-b1277700-5f65-11e9-9c49-73464d2f4d89.png)

a17andby:
Fixed and merged. Closing.

PR #5930, opened 2019-04-15
Moved icons and updated src path

a17poner:
Moved icon cross and ticks to the shared/icons and updated the src path accordingly fixing issue #5910
(e.g. in access when doubleclicking a name cell in the table you should be able to edit it and now see a green tick and a red cross icon).


HGustavs:
Screenshot for fixed functionality for documentation.

a17poner:
image
Here's a screenshot of the fixed functionality. Put the icons for it in shared/icons instead of in shared/sortableTableLibrary to keep all icons in the same place in case a cross and a tick has any use outside of sortableTable in the future.

Issue #5875, opened 2019-04-12
Cannot access filter elements that doesnt fit on screen. An scroll bar should be implemented


c17johha:
![image](https://user-images.githubusercontent.com/49199825/56033373-bea1ee80-5d24-11e9-874c-e55b574908af.png)

PR #5932, opened 2019-04-15
Update style.css #5875

a16hiwah:
#5875


a17amama:
The testning has been made, code follows standard and works fine

PR #5944, opened 2019-04-15
Tabel sorting fix for the access, test and files tab.

a16joare:
This is a solution that solves the sorting problem for the tables in access, test and  files tab on course front page. Instead of using 3 different compare functions i made one compare function that all sortingTables can use. The compare function is universal and can be used for more tables later on.

This solution works for the tabels at the pages stated above but do not work for the results table under the "Results tab" on the course front page.

My suggestion is that we take this solution and then adding the results table as an new issue and trying to solve that one with this solution.

### DIFF
--- a/DuggaSys/resulted.js
+++ b/DuggaSys/resulted.js
@@ -1,3 +1,4 @@
+// Test comment
 
 /********************************************************************************
    Globals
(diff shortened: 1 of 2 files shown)


HGustavs:
There is a default sorting function used in sortabletable one idea would be to replace this default with this function if this works better.

This way all default-search functions are automatically better.

a16joare:
This solution only makes the compare function for each table to work universal. Before this all tables had an own implementation of compare function. I just made one compare function that works for all tables.

a17oscno:
Ok to merge, following the code standard and the code works

a17pongo:
Please rewert my change if this merge is not ok, I was a little quick to accept.
sorry!

Issue #5886, opened 2019-04-12
Problem with aligning objects when lines are marked


a17alfcl:
Lines should probably not be included when determining distance between objects when aligning.
One solution would be to check for symbolKind when checking for overlap, and treating lines differently.

Related to issue #5686

a17jonro:
This is a logical change since lines might not need to be considered when aligning. However this doesn't seem to change anything since alignment is based on the top or bottom most coordinates which will always be of one of the objects that is not a line, since lines are always between two objects. There is a possibility though that this change affects the secondary part of the align functions which is adding spacing when objects overlap, since the lines are not considered. Not sure about this though since without the changes the overlap doesn't seem to work correctly for lines anyway.

Issue #5897, opened 2019-04-12
Selecting entities in the diagram doesn't work correctly


a17jonro:
You can select entities but they don't show the outline of being selected. Selecting multiple entities doesn't seem to work at all. Using the control key makes selection work as intended however. The issue in #5869 seems to be the reason for this.

PR #5946, opened 2019-04-15
Grupp a2019 w16#5886

a17alfcl:
Fix for issue #5886 
There are some issues with symbolkind but I still believe this should be the fix.
Problems that might occur with symbolkind should perhaps be a new issue if that is still a problem.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1423,7 +1423,7 @@ function lockSelected(){
 
 function align(mode){
     for(var i = 0; i < diagram.length; i++){
-        if(diagram[i].targeted == true && selected_objects.indexOf(diagram[i]) > -1){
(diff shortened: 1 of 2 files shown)


a17alfcl:
Might be some problems due to #5897 
Maybe needs a rebase to enable proper testing of this issue but idk

a17jonro:
This is a logical change since lines might not need to be considered when aligning. However this doesn't seem to change anything since alignment is based on the top or bottom most coordinates which will always be of one of the objects that is not a line, since lines are always between two objects. There is a possibility though that this change affects the secondary part of the align functions which is adding spacing when objects overlap, since the lines are not considered. Not sure about this though since without the changes the overlap doesn't seem to work correctly for lines anyway.

a17alfcl:
Closing this as the branch do not have changes that is needed for proper testing.
Making a new branch

Issue #5776, opened 2019-04-10
Help dropdown items in diagram doesn't do anything


a17jonro:
Help dropdown items looks clickable but they don't have any functionality. They should have the functionallity they represent.

PR #5951, opened 2019-04-15
Fix for #5776

a17alfcl:
Fix for #5776 
Changed class for div to "drop-down-text-non-clickable", which is currently non-existent but might be useful in the future.


a17jacsv:
Approved. Good job!

PR #5952, opened 2019-04-15
Fix for issue #5563

a17okaoz:
Issue fixed! #5563 I have changed the margin the searchbox and moved up search-box and serach-button 5px up


a17jacsv:
OK.

Issue #5896, opened 2019-04-12
Having the class box change in real time while changing the appearance of the class box


a16jenja:
![chrome_u6C11xZg43](https://user-images.githubusercontent.com/49199287/56038087-57d70200-5d31-11e9-90ce-fccd872d7f23.png)
Right now when changing the class name in the appearance options on a class box, the change will happen in realtime. You can see in the box when you change the class name, without pressing the "OK" button. Which is a good addition.

![chrome_K6QqDfZUju](https://user-images.githubusercontent.com/49199287/56038108-68877800-5d31-11e9-9537-0c99f79cdc67.png)
Therefore it would be nice to able to see realtime change when also changing in the attribute and operations fields in the appearance options. Right now it doesn't do that.

PR #5954, opened 2019-04-15
Fix for issue #5896

a17jonro:
#5896


a17jacsv:
It works as expected! Nice work!

PR #5957, opened 2019-04-15
Removed obsolete comment - Fix for #5707

a14oscna:
Comment was not needed as the fix has been used permanently for the black box problem. 
#5707


a17jacsv:
OK.

PR #5959, opened 2019-04-15
Add checkboxes - Fix for #5694

c17matjo:
For realsies this time. The branch has been moved to the new gruppA2019W16 and should merge without issues. Fix for #5694

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -811,7 +811,7 @@ function toggleGrid() {
     } else {
         snapToGrid = false;
     }
-    setCheckbox($("a:contains('Snap to grid')"), snapToGrid);
+    setCheckbox($("span:contains('Snap to grid')"), snapToGrid);
 }
 
 function toggleVirtualA4(){
@@ -1249,7 +1249,7 @@ function debugMode() {
         ghostingCrosses = true;
     }
     updateGraphics();
-    setCheckbox($("a:contains('Developer mode')"), !ghostingCrosses);
+    setCheckbox($("span:contains('Developer mode')"), !ghostingCrosses);
 }
 
 /******************************************************************************************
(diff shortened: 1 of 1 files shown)


a16tobfr:
Has there been a new branch made from the new groupA2019W16 branch since the last merge or did you continue on the one originating from the old group branch (results in lots of conflicts when pulling the branch)? As of now, if it's the gruppA2019#5694 branch, there are tons of conflicts when pulling the code.
An entirely new branch has to be created since after the merge. My experience is that these kinds of conflicts happens otherwise. The {choose branch to merge into} option in github desktop does not seem to be enough.

c17matjo:
No, this is still the old branch. I used git-rebase to move the commits onto the new groupA2019W16 branch so there shouldn't be any merge conflicts. According to github, merging this pull request will only add the three commits and won't cause any merge conflicts so the rebase seems to have worked. 

image
...
image

a16tobfr:
Okey, even if this logically should work, this is the reslut which occur when doing this:
conflicts
This is why it is better to simply create a new branch and then just copy paste your changes from the old branch into the new branch. It should be a quick fix for this issue, and allows me to start testing the changes.

c17matjo:
There shouldn't be so many changed files in just this pull request. This PR only modified diagram.js, the other files shouldn't be there. Even if I had created a new branch it would still have created a conflict with all those changed files. Just to make sure, I tried installing github desktop and performing the merge on my laptop and there were no conflicts. Where did the other changed files come from?

a16tobfr:
In every case that I am aware of where this approach has been made (my own first approach to a previous issue from before the last merge as well), we've seen these types of conflicts occuring. The solution have always been to make a new branch from the new group branch and manually add the changes to the new branch. I will wait with testing this issue until I have discussed with Jacob/Drazen on how to move forward from here.

a17jacsv:
To prevent eventually problems caused by last weeks merge, I suggest that you remake this solution on a new branch from the updated version of our branch (gruppA2019W16). It wasn't many rows of code and should just take a few minutes to fix. Please do that.

PR #5961, opened 2019-04-15
GruppC2019W16_5870 Username Shadow changed

a17fellu:
Changed the Shadow of the username text in the navbar to black instead of green


a17adajo:
Nice work!

Issue #5840, opened 2019-04-11
Add whitespace before brackets


a17celal:
It should be a whitespace before the first bracket in a function. This should be fixed in several diagramfiles.

`function loadLineForm(element, dir){`

PR #5962, opened 2019-04-15
Fix for issue: #5840 "Add whitespace before brackets"

a17vikab:
Very small quality fix for the code related to the diagram-files.

Issue: #5840


a17jacsv:
Looking good. Approved.

Issue #5708, opened 2019-04-08
Change comments to the right standard


a17celal:
Some comments in the digram files does not follow the code standard. Should be changed to the right format.

![Capture4](https://user-images.githubusercontent.com/37792271/55723867-63ad8600-5a0a-11e9-81e8-138a869e589a.PNG)

HGustavs:
Yes this is worthwhile since it makes the code more readable for all.

Also, when making screenshots of code, use moustaches (~ <- three of those) and then the code pasted into the document and then three new moustaches.

Screenshots are for user interface stuff, for quoting code github and mardown has helped us...

~~~
Then it looks like this
~~~

a17johpe:
I've come across php files containing html code and as such only html comments work .
How should comment standard look for html comments? Always single line, or should a comment such as 
~~~
<!-- The Appearance menu. Default state is display: none; -->
~~~
before a large portion of nested divs be changed into, e.g.:
~~~
<!---------------------------------------------------------
    The Appearance menu. Default state is display: none; 
---------------------------------------------------------->
~~~

PR #5970, opened 2019-04-16
Fix for #5708

a17johpe:
Comments in all files named "diagram" have now been updated to follow the comment standard. In the PHP files there are HTML comments, and as there is no comment standard provided these have not been changed.

https://github.com/HGustavs/LenaSYS/issues/5708


a17jacsv:
OK.

Issue #5942, opened 2019-04-15
Show origo in developer mode


a17drama:
Origo should be shown i developer mode and no it should be static. That means it's pinned to the canvas and no matter what it's stuck on the same point as it was from the start. Please use the proper "icon" for origo.

HGustavs:
Im thinking that we show the "page" or "the first grid line with a thicker colored line" so that it moves
with scrolling.

Below is a mock-up of my two options....

![Screenshot 2019-04-16 08 24 17](https://user-images.githubusercontent.com/6978801/56186824-08d6e880-6021-11e9-89ff-96105a1add65.png)

a17drama:
Sounds like a good idea with the lines. Implement those please.

PR #5971, opened 2019-04-16
Fix for issue #5942

b17robev:
#5942 

Shows lines for (0,0) so it is clear where the initial origo is.  Further the functionality to toggle origo on and off based on developer-mode.


a17jacsv:
Looking good. Nice work!

Issue #5933, opened 2019-04-15
Remove unsued files, Sortable


c17johha:
These files are never referenced:

/Shared/SortableTableLibrary/fixedtest.html
/Shared/SortableTableLibrary/sortableTableDemo.html
/Shared/SortableTableLibrary/sortableTableDemo_columnfilter.html
/Shared/SortableTableLibrary/sortableTableDemo_edit.html
/Shared/SortableTableLibrary/sortableTableDemo_highlight.html
/Shared/SortableTableLibrary/sortableTableDemo_rowfilter.html
/Shared/SortableTableLibrary/sortableTableDemo_simplest.html
/Shared/SortableTableLibrary/sortableTableDemo_simplest_w_magich.html
/Shared/SortableTableLibrary/sortableTableDemo_sortadvanced.html
/Shared/SortableTableLibrary/sortableTableDemo_sortsimple.html
/Shared/SortableTableLibrary/sortableTableDemoDB.html

Sub-issue from: #5692

PR #5984, opened 2019-04-16
Deleted non-referenced pages

c17alepe:
11 non used files deleted. 

#5933


a17amama:
Removed files that aren't used, however i believe there are more files that aren't used, perhaps "fixedtest" is one of them?

PR #5989, opened 2019-04-16
Grupp a2019 w16#5941

### LINKED ISSUE #5941: Show coordinates without decimals when zooming
I would like the coordinates should show like the picture 1 below and zooming should not affect the writings of the coordinates. Remove the decimals for a smoother view and use. It should view as the second picture below.

Skärmavbild 2019-04-15 kl  13 35 49

Skärmavbild 2019-04-15 kl  13 35 25

a16tobfr:
Issue is fixed #5941 
Now there are always only one decimals that is printed in the cordinator bar.


a17jacsv:
The functionality works as expected. You should add a comment and describe what the function does.

a17jacsv:
Perfect! Good job.

PR #6000, opened 2019-04-16
merge fixing: removes syntax errors, erroneous structure and call before definition

b17timjo:
+autoformat, with tabs (as per code standard, probably)


a17alian:
Code tested - No errors appears in console on resulted page

Issue #6006, opened 2019-04-17
Virtual a4 buttons still uses anchor tags


a17jonro:
Fixes for issue #5639 have been merged in the wrong order and therefore some changes have not been applied as they should. Changes in this issue has affected issue #5694 as well since anchor tags were used.

PR #6007, opened 2019-04-17
Fix for issue #6006

a17jonro:
#6006


a17jacsv:
Works as expected. Good job!

Issue #5533, opened 2018-05-23
ResultEd - Option to filter away entire moments


a16jonis:
In #4899 you can see what has been implemented in terms of filter options, in the current state. As mentioned in the thread the option to select/deselect all columns belonging to an entire moments is missing. 

For example: if you're in the course _Webbutveckling - datorgrafikIT118GHT15_ you'd probably want to be able to click a checkbox for "Biträkningsdugga 1HP". 

![filter](https://user-images.githubusercontent.com/37793947/40412539-df825834-5e73-11e8-86d1-f46b4a753d4b.PNG)

Doing this should have the duggas "Biträkningsdugga 1" and "Biträkningsdugga 2" be filtered away too since they are in the same moment.

![result](https://user-images.githubusercontent.com/37793947/40412566-f616eff6-5e73-11e8-9eb7-2cbb3ca3e547.PNG)

PR #6011, opened 2019-04-17
fix merge typo/code ended up outside original function from #5533

b17timjo:
Reinstates functionality from implementation of #5533


b17timjo:
see #6035 instead

PR #6014, opened 2019-04-17
Grupp a2019 w16#5886

a17alfcl:
Fix for issue #5886
Removing lines from each align function now instead as that seemed to work better
Closed previous pull request as the branch had an old version which made testing harder


a17jacsv:
OK.

PR #6033, opened 2019-04-17
Fix for else error in sortabletable

a17adasy:
fixed curly brackets so they match the if else statements. So now the sortabletable should be working.

### DIFF
--- a/Shared/SortableTableLibrary/sortableTable.js
+++ b/Shared/SortableTableLibrary/sortableTable.js
@@ -43,19 +43,18 @@ function defaultRowFilter() {
 
 
 // Global sorting function global
-function sortableInternalSort(a,b)
-{
-    var ret = 0;
-    //var colname = sortableTable.currentTable.getKeyByValue();
-    var colname = sortableTable.currentTable.getSortcolumn();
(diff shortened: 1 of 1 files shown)


Shared/SortableTableLibrary/sortableTable.js
```diff
@@ -105,44 +104,44 @@ function clickedInternal(event, clickdobj) {
...
-      var xscroll = window.pageXOffset;
-      var yscroll = window.pageYOffset;
-
-      popoverelement.style.left = Math.round(lmnt.left+xscroll)+"px";
-      popoverelement.style.top = Math.round(lmnt.top+yscroll)+"px";
-      popoverelement.style.minHeight = (Math.round(lmnt.height)-5)+"px";
-      popoverelement.style.maxWidth = "fit-content";
-      popoverelement.style.display = "flex";
-    }
+		var rowelement = event.target.closest("tr");
```
a17adajo: Its hard to see whats really changed when the indentation are changed across the file

a17pongo: is it ok to merge these fixes and re-do some testing?
@a17andby @a17jacsv

Issue #6039, opened 2019-04-17
Duplicated code in sortableTable.js


c17julka:
Rows 398 - 459 in sortableTable.js consists of duplicated code. Should be enough to remove this duplication?

PR #6047, opened 2019-04-17
Weekly merge

a17andby:
Merge group A, B and C to master.


c17johha:
Unfortunately there is a problem with this branch in sortableTable.js, with duplicated code. But the issues is solved and will be available in the next weekly merge.
#6039

Issue #6020, opened 2019-04-17
Diagram.php uses space that isn't used


a17vikab:
In diagram.php, the canvas does not need all the space that is available on the page. I suggest removing some of the spacing above, below and on the sides. 

Related to #5820

a17jacsv:
I totally agree. It would be a smoother view.

PR #6057, opened 2019-04-17
Fix for issue: #6020 "Diagram.php uses space that isn't used"

a17vikab:
Issue: #6020 

Fixed that you don't have to scroll up/down & sideways, in other words, you always see 100% of the screen without scrolling.


a17jacsv:
Looking good! Nice job.

PR #6059, opened 2019-04-17
Removed console prints.

a16joare:
Removed test prints from sortableTable.js


a17andby:
This will be implemented from an other group (https://github.com/HGustavs/LenaSYS/pull/6067) . Closing this pull request.

Issue #6038, opened 2019-04-17
Key-binds are not mapped correctly for arrow keys


a14oscna:
Currently in the diagram when you want to move any drawn object using the arrow keys, their bindings are wrong. They also interact with the scrollbar on the page if present, which should not happen if an object is selected. 
**Up**-arrow moves object **down**.
**Down**-arrow moves object **right**.
**Left**-arrow moves object **up**.
**Right**-arrow moves object **left**.

EDIT: I did not attach the W17_test label as the bug was confirmed present pre-merge.

a17jacsv:
Oh yes, that needs to be fixed...

PR #6071, opened 2019-04-17
Fix for issue #6038

a17jonro:
#6038


a17jacsv:
Works as expected. Nice job!

Issue #5887, opened 2019-04-12
Resulted, Mail button to all visible students in the list


c17johha:
#5614 Mail button was implemented in sectioned, but it should be in resulted
It should be possible to create a mail that will be sent to all students that's currenlty visible in the list.

HGustavs:
Im thinking we use something that works like the csv export functionality in sortabletable but only for the email column (we need a setting to tell which column is the email and a callback for generating the email headers etc)

PR #6072, opened 2019-04-17
#5887 Pull request for the Mail button in the Result view.

a17oscno:
For more commits and to easier see the history of the development look on branch "GruppB2019W16#5887".

Site: https://a17ludca.gruppb.webug.his.se:20002/DuggaSys/resulted.php?cid=2&coursevers=97732

In the Result tab there is a mail button that will mail all students in the "view" (the current table that is showing). And you can use the search bar to search for different things and then click the mail button to open the mail with those students emails already assigned.

issue = #5887

### DIFF
--- a/DuggaSys/resulted.js
+++ b/DuggaSys/resulted.js
@@ -1210,3 +1210,32 @@ function closeLadexport() {
 function updateTable() {
 	myTable.renderTable();
 }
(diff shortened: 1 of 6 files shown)


a17amama:
Works fine and follows standard

Issue #6049, opened 2019-04-17
Add hover effect on new item button


a17robda:
There should be a hover effect for the button to add a new item in dugga to increase visibility?
![img](https://user-images.githubusercontent.com/37792328/56284290-5c762e80-6114-11e9-8353-b82add7585be.png)

a17andby:
Fixed and merged. Closing...

PR #6077, opened 2019-04-17
Modified add button to become lighter on hover.

a17robda:
For #6049


a17robes:
Seems like the code for hover effect has been added to the css but the affected class wont listen to the css changes made. Not sure why this happens?

a17poner:
Think I found the correct file, it was /shared/css/style.css while you had edited /shared/sortabletablelibrary/style.css, this project is a mess with css files but the section add button has a hover now on my machine. Used the same styling and everything that you had made, just copied it into a different css file.

Issue #5997, opened 2019-04-16
Menu getting overwritten by the toolbox


a17drama:
When loading a fresh page where the toolbox is in default position makes the file menu act funny. If the toolbox stays in the default position and you are trying to chose something from the file menu it will collapse the file menu every time you hover in over the toolbox. Menus should be placed on top of the toolbox instead of the other way around.

PR #6079, opened 2019-04-17
Fix for issue: #5997 "Menu getting overwritten by the toolbox"

a17vikab:
Moved the toolbox downwards towards the center rather than it starting in the upper-left corner.


a17jacsv:
It works well and as expected! Good job.

Issue #6012, opened 2019-04-17
"Toggle A4 Holes" should not be displayed if "Display Virtual A4" isn't active


a17vikab:
In View, it should not be possible to see the option to Toggle A4 Holes while Display Virtual A4 isn't active. It creates confusion for the user at the moment and should be hidden until the display virtual A4 is active.

![issue1](https://user-images.githubusercontent.com/37792464/56270989-b5839980-60f7-11e9-9bb3-59bddfa4791e.png)

a17jacsv:
I agree...

PR #6087, opened 2019-04-17
Fixed the toggle on A4 holes - Fix for #6012

a14oscna:
Toggling the virtual A4 holes is not dependant on if the virtual A4 is toggled on or off.
Added styling for when the option is disabled, hover effect also removed when disabled.
disabled option
Fix for #6012


a14oscna:
While I agree that my solution may not be the best, i strongly disagree that hiding the option completely is a better solution. Perhaps we could show the dependency in another way?

a17jonro:
Yes, just making it more obvious that the 'virtual a4 hole' option is tied to the 'display virtual a4' and therefore minimizing confusion while retaining the button itself would work great.

a17jonro:
This is great, but unfortunately this has slightly broken the checkbox on the 'Toggle A4 Holes' button.
image

a17jacsv:
Looking good. Works as expected. Nice job!

PR #6094, opened 2019-04-17
Time sheet: Changed day from DATETIME to DATE

a17joaap:
Gets rid of the trailing 0's in the submission view

### DIFF
--- a/DuggaSys/filereceive_dugga.php
+++ b/DuggaSys/filereceive_dugga.php
@@ -65,28 +65,28 @@ if($ha){
 		// Create folder if link textinput or file
 		$currcvd=getcwd();
 		if(!file_exists($currcvd."/submissions")) {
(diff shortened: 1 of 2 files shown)


a17joaap:
It should work now

a17maxsj:
All old entries shows UNK! what could be the cause of this?

image

a17joaap:
Your entries are being saved in the database:
dbtest

The problem seems to be with the submissions folder

a17joaap:
@a17maxsj this is because of different branches using the same database 'Databas2019'.

If I upload a submission on branch A, it will save it to the correct folder with the name timesheet1.txt.
If I then switch branch to B, and upload a new submission, it will have the name timesheet2.txt.

Both of those submissions are saved in the submission database table, which is why they show up in the tab. But timesheet1.txt only exists on branch A and can not be viewed when on branch B.

Issue #6022, opened 2019-04-17
Move searchbar inside  "resulted"


a17amama:
The searchbar is way too close to the rest of the content making it look awkward, move it up.

a17amama:
Changed styling on searchbar to move it from corner and from other content, however i've noticed another issue (not caused by this issue), the searchbar has a border inside of it, which should be removed. 
Here are the results of the moved searchbar however:
<img width="220" alt="searchbarmove" src="https://user-images.githubusercontent.com/37793130/56286846-0ce73100-611b-11e9-8f24-15027df40e93.PNG">

PR #6095, opened 2019-04-18
Grupp b2019 w17#6022

a17amama:
Moved up the searchbar in "resulted" to better separate it from the rest of the content and also make it look less awkward.

#6022


a17pongo:
Do not merge to master! please :)

Issue #5143, opened 2018-05-08
Forced lines in diagram


a16juleh:
For now, forced lines need to be added manually. In the future, it would be nice to automatically add this when objects are changed. 

<img width="459" alt="skarmavbild 2018-05-08 kl 08 52 48" src="https://user-images.githubusercontent.com/37794632/39742285-35b9be78-529d-11e8-8fad-f4f9944b5f2e.png">

PR #6099, opened 2019-04-18
Grupp a2019 w17#5143

a16tobfr:
Should now automatically make lines forced between relations and entities when setting the relation to weak
#5143


a17jacsv:
Works as expected! Good job!

Issue #5956, opened 2019-04-15
Make the searchfield bigger


a17okaoz:
The serach-area is to small. Make so the grey border around will get bigger and make more space for the letters
<img width="259" alt="b1" src="https://user-images.githubusercontent.com/37792681/56136113-653ef700-5f92-11e9-8070-f3453837779c.PNG">

The letters get exactly the place in the box I will be better if there was little space between the letters and the border

HGustavs:
I agree.

We should also examine the possibility of putting an optional search box (usable in many places) in the navigation menu...

or at least making a test version to see if it looks like crap.

Currently when the search field scroll out of view they are nearly useless.

PR #6100, opened 2019-04-18
Grupp b2019 w17#6022

a17amama:
Moved searchbar in resulted to separate it from the rest of the content and make it less awkward
#6022


b17timjo:
The CSS change will alter the search bar position in both resulted.php and fileed.php. The new position looks fine in both.

A comparsion (top is old, bottom is changed):
6022_compare_oldnew
The increased margins of `searchBar` has the desired effect of separating the bar from the table.
The top/bottom margins in `searchInput` moves the actual input field inside the container, causing the line seen in the bottom example of the image above, which is probably not intended (and does not look very nice)

a17pongo:
@a17amama 
Please make the requested changes to searchbar.

a17amama:
The lines inside the searchbar has been removed, which i believe was already an issue on its own, but it is resolved now anyway and is ready for testing.

a17pongo:
@a17amama if an issue for this already exist could you make a reference to it here?

a17amama:
The issue that was supposed to handle this is issue #5956, which will be closed now.

Issue #6028, opened 2019-04-17
Cardinality numbers gets partially hidden when having weak entities


a16tobfr:
Cardinality number gets partially hidden since the rectangle gets bigger when setting it to be 'weak'.
Offset of the cardinality number should be slightly bigger for weak entities
![cardinalityPartiallyHidden](https://user-images.githubusercontent.com/49199690/56276733-5deb2b00-6103-11e9-844f-e67dbcb08147.JPG)

PR #6109, opened 2019-04-18
Grupp a2019 w17#6028

a17okaoz:
Issue #6028 Fixed. Changed the x and y values so the cardinalities dosen´t gets hidden by the weak entities


a17okaoz:
Issue #6028 Fixed. The cardinality dosen´t get hidden by the weak entity anymore. But I have written a issue about the start positon of the cardinality in diagonal position

a17jacsv:
Approved.

Issue #6083, opened 2019-04-17
Space between menu and dropdown


a17vikab:
There is a space between the menu and the dropdown and if e.g the toolbox is below the menu it can break the dropdown. Look at the image for more explanation. it looks like it is a few pixels that are between the dropdown and the menu and it should not be any space at all.

![issue2](https://user-images.githubusercontent.com/37792464/56291463-9ac81980-6125-11e9-9fe2-c235e5b124e4.png)

a17drama:
This issue could be related to #5997 or even the solution for it. Removing the space could help a lot.

PR #6116, opened 2019-04-18
Fixed issue #6083

a16jenja:
Issue #6083 is fixed.

Moved the menu a bit higher when hovering over a menu option.

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -1337,7 +1337,7 @@ input {
 
 .menu-drop-down span.drop-down-label{
   display: block;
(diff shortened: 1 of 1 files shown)


a17drama:
Make sure this test covers the alternative when the tool box is moved to the upper left corner of canvas. This was causing an issue #5997 and the solution was to move the tool box. Initial problem was that there was some space between the drop down and the menu where the mouse could hover over the tool box and that caused the issue #5997 .

a17jacsv:
Good job. It works as expected.

Issue #6098, opened 2019-04-18
Remove duplicates of functions


b17robev:
Duplicate declaration of functions in files seems to be related to #5868 

digram.js ...
```
function initToolbox() {
    var element = document.getElementById('diagram-toolbar');
    var myCanvas = document.getElementById('myCanvas');
    var bound = myCanvas.getBoundingClientRect();
    element.style.top = (bound.top+"px");
    toolbarState = (localStorage.getItem("toolbarState") != null) ? localStorage.getItem("toolbarState") : 0;
    switchToolbar();
    element.style.display = "inline-block";
}
```

```
function initToolbox() {
    var element = document.getElementById('diagram-toolbar');
    var myCanvas = document.getElementById('myCanvas');
    var bound = myCanvas.getBoundingClientRect();
    element.style.top = (bound.top+"px");
    toolbarState = (localStorage.getItem("toolbarState") != null) ? localStorage.getItem("toolbarState") : 0;
    switchToolbar();
    element.style.display = "inline-block";
    /*
    element.style.height = (bound.bottom-bound.top-200+"px");
    element.style.height = (400+"px");
    */
}
```

a17jacsv:
I agree. This should be fixed.

PR #6117, opened 2019-04-18
Fix for #6098 - duplicate code

a17johpe:
https://github.com/HGustavs/LenaSYS/issues/6098

The code that has been removed should be almost the exact same code as the one that follows directly below it with only minor differences. The code that has been removed seems to be older, since it doesn't use consts as key codes, and doesn't contain some of the minor fixes that has been done the past week (such as updated comment standard).


a17jacsv:
Agreed. Good addition and makes the code more readable.

Issue #6082, opened 2019-04-17
Make some room between the text when a file is uploaded. 


a16joare:
The filename and upload date is to close to each other. I suggest that we make some space between them to make it more readable.

![image](https://user-images.githubusercontent.com/49194898/56291084-da423600-6124-11e9-9229-4c13d8d9671a.png)

HGustavs:
Adding some padding is not a bad idea.

Also, long (several lines/pages?) feedback tends to look weird.

a17maxsj:
This has been merged, closing...

Issue #6002, opened 2019-04-17
Time sheet: add week number when uploading data


a17joaap:
As per Henriks [request](https://github.com/HGustavs/LenaSYS/issues/5628#issuecomment-482473714).

When posts get uploaded to the database, the week number of the post should be added.

This means adding week to the insert statement on row 146 in filereceive_dugga.php, and calculating the week number from the 'day' value and adding it with bindParam.

a17joaap:
Fixed in #6255

a17andby:
![fixed](https://user-images.githubusercontent.com/49235724/56724652-b77ad780-674b-11e9-8a90-853eeb9d1140.jpg)

PR #6120, opened 2019-04-18
Fix for issue #6082 and #6002

a16joare:
Made some room between cell headings and added "week" as a table heading(issue #6082). I then made it possible to see which week an assignment was uploaded (issue #6002). 

image


a17joaap:
Oh, I thought Henrik meant a week column in the database. That way they could run 
`SELECT * FROM timesheet WHERE uid=1 AND week=15`
to see what users did that week.

weekdb

If we want this, we should add week to the insert statement on row 146 in filereceive_dugga.php, and use your week number calculation to add it with bindParam

Issue #6073, opened 2019-04-17
Add menu option to reset origo


a14oscna:
Right now the only way to reset origo to its original position in the top left corner is to refresh the page. Moving it back by dragging it manually lacks precision, hence there should be an option reset it. The menu option could be added somewhere under the Edit category.

a17drama:
There is another issue with similar problems #6031 . Maybe these can be solved with the same "Reset" of the page trough the menu?

a17jonro:
I don't think this issue has anything to do with #6031 since the toolbox is not part of the canvas, which is what this issue concerns. These functions could possibly be used together by maybe using the same button if that's desirable, since their functions are indeed quite similar.

a17drama:
This is my thought exactly to use the same button for resetting both origo and the tool box. Person working on #6031 can piggy back on this solution.

HGustavs:
I would suggest both in tool box and menu. 

Something like zoom extents in autodesk software.

https://knowledge.autodesk.com/support/3ds-max/learn-explore/caas/CloudHelp/cloudhelp/2016/ENU/3DSMax/files/GUID-B27D7DCC-EFDF-4D27-B6BA-F6E6BA1B7E3C-htm.html

Issue #5826, opened 2019-04-11
Camera object instead of moving the canvas


a17drama:
Moving the canvas around creates a load of problems. This can be avoided by locking down the canvas and creating a camera object that hovers over the canvas to display what user wants to see. This will help solving several other issues. #5145 , #5721 More info about this idea can be found in #5748

a17jacsv:
Issue closed.

PR #6126, opened 2019-04-18
Fix for issue #6073

a17jonro:
#6073


a17jonro:
Yes, this issue is present in the current solution. The solution for this issue uses the same way of moving around as how the drag function is implemented and therefore it should work. This issue is also present in the virtualA4 as well, probably because they both use ctx.translate. Both of these should be solved if a solution for issue #5826 is implemented.

a17jacsv:
There aren't any big changes made so it's not a great risk, so with that said we’re merging it and to test it again when the "camera-functionality" is implemented.

Issue #6101, opened 2019-04-18
Automatic mail function should not display all e-mail addresses


a17anter:
When generating an automatic e-mail all the addresses are visible to all recievers. This should be confidential and a student should not see any other addresses.

![automatic mail function](https://user-images.githubusercontent.com/49204317/56346245-fe078980-61c1-11e9-8a90-e057b5a63a22.PNG)

a17anter:
Updated to that it will be sent out as a "blind copy" which means that the recievers won't be able to see any other recievers.

![Screenshot for blind copy mail](https://user-images.githubusercontent.com/49204317/56368076-fca48400-61f6-11e9-8b01-b1d11afba89f.PNG)

PR #6132, opened 2019-04-18
Grupp b2019 w17#6101

a17anter:
#6101


b17nicsv:
Follows code standard, is very readable and works as intended. A mail recipient can't see other recipients.

Issue #5630, opened 2019-04-01
Hiding ssn


HGustavs:
As a general rule we want to hide ssn numbers except for export of data.

This is due to privacy EU rules.

a17pongo:
@HGustavs 
this seems like something that should be implemented now. 
Could you give more information on eg. where it should be hidden?

a17alian:
I've found visible ssn in 'Results' and 'Access' and masked them out. This solution still makes it possible to sort by SSN in heading. 
<img width="374" alt="Skärmavbild 2019-04-20 kl  15 38 55" src="https://user-images.githubusercontent.com/37792256/56458193-cb0cf380-6383-11e9-8ee3-266090f1e06c.png">

<img width="212" alt="Skärmavbild 2019-04-20 kl  15 47 08" src="https://user-images.githubusercontent.com/37792256/56458194-cb0cf380-6383-11e9-8072-89fe6353cf77.png">

@HGustavs please let us know we missed pages where ssn needs to be masked

PR #6133, opened 2019-04-20
Masking ssn

a17alian:
Masking ssn in resulted and access page - solving issue #5630 - hiding ssn


a17ludca:
This seems to be working as intended. 👍

Issue #5964, opened 2019-04-16
Possible to remain and access the access page as guest.  


a16joare:
If you go to the access page as admin and logout is it still possible to remain on the page. You can't see the content on the page, but a relocation to the front page would maybe be better?

It's also possible to copy the url and access the page. 

Here is a image when logged in:
![image](https://user-images.githubusercontent.com/49194898/56191212-3ecd9a00-602c-11e9-8af0-922358d4125c.png)
Here is a image when you logout:
![image](https://user-images.githubusercontent.com/49194898/56191243-4d1bb600-602c-11e9-9a0c-99def787529f.png)

b17annni:
#6134

a17alevu:
I feel like when you logout you sahould directly be moved to the homepage as a guest.

PR #6135, opened 2019-04-21
Grupp c2019 w17 5964

b17annni:


a16joare:
I checked this one aswell and it seems to work for me. (made the test on the test-server). It works for the access tab but not for the others except tests but it worked before this fix. Should we fix all of the tabs before we merge this one @andreas-bylund

a17maxsj:
might have something to do with the URL on local machines vs webserver, check the changed code, "changeURL" something

a17andby:
After a conversation with @b17annni I will close this pull request. 

https://github.com/HGustavs/LenaSYS/issues/5964

Issue #5990, opened 2019-04-16
The decimal is always 0, 3, 5 or 7


a17okaoz:
The decimal in the coordinate bar is always showing 0, 3, 5 or 7. Make it so the decimal is more accurate

a17jacsv:
I agree...

c17matjo:
I've looked into it and it seems like this behaviour is caused by the zoom values chosen and the mouse position data being reported in integers (numbers of pixels in the x- and y-direction). The mouse location values are divided by the zoom values but since the mouse values are integers and most of the possible zoom values are 3/something (150% = 3/2, 75% = 3/4, ...) the final coordinate ends up with a value ending with .0, .333 or .666 which is rounded to .0, .3 and .7 respectively. This isn't really a bug. 

TLDR: The issue is caused by low precision on the position values we get from the browser, not because of an issue with the diagram. I suggest we either leave it as is, or we remove the decimal digit from the coordinate bar. I think it would look the best to round it off to the nearest integer.

c17matjo:
I've made a PR with my suggested solution (round off to the closest whole number). #6137
![bild](https://user-images.githubusercontent.com/49195031/56561273-1ce19380-65a7-11e9-9da6-3f36e1be89b2.png)
(Screenshot of coordinate label after PR)

PR #6137, opened 2019-04-23
Remove unnecessary decimal precision for mouse coordinates - Fix for #5990

c17matjo:
Suggested fix for #5990. Since getting more precise mouse input is not possible this PR just removes the unnecessary decimal precision from the coordinate values. 
bild


a17jacsv:
As a17vikab said, smart and well-executed fix! Nice addition.

Issue #6144, opened 2019-04-23
De-select all entities when re-opening diagram


a14oscna:
Currrently if you put entities in your diagram and close the session all the entities remain when you enter the diagram page again. The entity which was selected in the previous session will still be selected visually but NOT functionally. For example if you press delete to remove the selected entity you get the following message: 
![saved entity looks marked](https://user-images.githubusercontent.com/37793141/56565380-831ee400-65b0-11e9-97fa-c8e700fe84ab.PNG)
To fix this the visual representation of a selected entity should be removed as a new session is opened on all entities that remain from a previous session.

HGustavs:
Seems reasonable.

PR #6149, opened 2019-04-23
Grupp a2019 w17#6144

a17alfcl:
Fix for issue #6144 
Running a deselect function when loading a diagram as doing this when saving might interfere with the current state while working.


b17robev:
Removed the comment entirely

a17jacsv:
OK.

Issue #6031, opened 2019-04-17
Menu-item who brings back the toolbar


a17jacsv:
The toolbar in the diagram is shown when opening the file. Sometimes it disappears for unknown reasons and you can get it back by refreshing the page, that shouldn't be necessary. It can be scary for the user to refresh and all their content might be gone.

Please implement a menu-item in the section "view", who will "bring" back the toolbar. 

<img width="375" alt="Skärmavbild 2019-04-17 kl  11 39 34" src="https://user-images.githubusercontent.com/37792795/56277942-968c0400-6105-11e9-8571-d64f9bc35da3.png">

a17drama:
This issue could be solved together with #6073 .

PR #6150, opened 2019-04-23
Fix for issue #6031

a17jonro:
#6031


a17jacsv:
Perfect! Good addition.

Issue #6146, opened 2019-04-23
Indication when resizing an entity


a17drama:
When hovering a corner (point) of an entity there should be a clear indication that it's possible to resize the entity. This should be indicated by changing the mouse icon to:
![Capture](https://user-images.githubusercontent.com/46029340/56566371-bcf0ea00-65b2-11e9-9c3c-72607f26f4d2.png)

HGustavs:
Yes.

Older version had this... and it was very useful.

PR #6158, opened 2019-04-23
Grupp a2019 w17#6146

a17alfcl:
Fix for issue #6146 
Could perhaps have made the comparisons for each corner instead (i.e. topleft, topright etc.), which would make the logic more comprehensible but looks uglier. IDK...


a17jacsv:
Works as expected! I agree with a14oscna comment.

Issue #5960, opened 2019-04-15
Ability to find connected entities


a16tobfr:
It would be good if there was a function which lets you get the objects which other objects are connected to. I.e. that you are able to find the two entities that a line is connected to (in the code).

HGustavs:
This used to be part of the hover functionality I think?

When one hovers a line, both ends should/could be highlighted with for example a blue line color.

PR #6162, opened 2019-04-23
Fix for #5960:  Ability to find connected entities 

a17johpe:
There is now a function in diagram_objects.js that can be used, e.g. to evalutate which kinds of symbols a line is connected to.

The function is called on the line in question and returns an array with the two symbol objects that is connected to the line.

https://github.com/HGustavs/LenaSYS/issues/5960


a17johpe:
Note: to test if this function works it is necessary to do a console.log somewhere (I did it at the end of drawline() but this updates on every frame the cursor moves)

a17jacsv:
OK.

Issue #6155, opened 2019-04-23
Can't lock after alignment


a17drama:
Selecting an entity or several entities and aligning breaks the locking function. The entities are still selected after the alignment and can be deleted if so wished but they can't be locked. Locking should be possible even after alignment.

PR #6165, opened 2019-04-23
Fix for issue #6155 Can't lock after alignment

a17jonro:
#6155


a17jacsv:
Works as expected. Good work!

PR #6172, opened 2019-04-23
Grupp a2019 w17#5826

a17alfcl:
First step for fixing #5826 
No zooming function is implemented at the moment.


a17drama:
Some issues have been discovered with this pull request and are currently being worked on. Setting the status "On-Hold" while this is being solved. This pull request will be updated.

Issue #6181, opened 2019-04-23
Setting entity to weak does not change lines between entity and relation to forced in some cases


a16tobfr:
![lineNotForced](https://user-images.githubusercontent.com/49199690/56614153-9e253e80-6618-11e9-91cf-db5aa5824543.jpg)
Easy to fix and the source of the problem is identified. Simply a matter of adding a = sign after an x < y check in some for loops.

~~~
for (let y = 0; y <= attributeMidPoint.length; y++) {
for (let k = 0; k <= attributeMidPoint.length; k++) {`
~~~
instead of:
~~~
for (let y = 0; y < attributeMidPoint.length; y++) {
for (let k = 0; k < attributeMidPoint.length; k++) {
~~~

PR #6182, opened 2019-04-23
lines between entity and relation to forced #6181

a16tobfr:
#6181
The lines between a weak entity and a relation should now be changed to 'forced' whenever the entity is set to be weak.


a16tobfr:
It means that the code is run even if an attribute is not connected to the entity, which was the problem with the previous solution. This makes sure that the code is run in every case. Even when the arrays are empty.
 This previously resulted in lines not changing to be 'forced' in the situations where there was no attribute present etc.

Have you detected any case in where the lines are not changed to forced right now?

a17jonro:
Ok, that clarified the problem, making sure that the code runs anyways is good. But i would advise a note of this in the comments so that this behaviour is not removed by misstake.

a17jacsv:
Approved.

Issue #5690, opened 2019-04-08
Create lines with "kinks" rather than diagonal


a17jacsv:
**Create lines that are using "kinks" rather than diagonal lines. This is needed in UML-diagrams and would benefit the application.**

See pic below: 

<img width="438" alt="Skärmavbild 2019-04-08 kl  09 28 51" src="https://user-images.githubusercontent.com/37792795/55706135-be31ec80-59e0-11e9-9588-ea6405525fca.png">

This should require "anchors" to adjust the lines. See pic below:

<img width="327" alt="Skärmavbild 2019-04-08 kl  09 43 07" src="https://user-images.githubusercontent.com/37792795/55706953-bd01bf00-59e2-11e9-88b3-a02788edede8.png">

a17johpe:
The example image goes against how the lines currently work for all symbols, i.e. that the start and end points for the line is dependant on which sides of the symbols are closest to each other (see images).
![diagram1](https://user-images.githubusercontent.com/37792313/55709797-6946a400-59e9-11e9-8f4b-ae87804f3ed5.png)
![diagram2](https://user-images.githubusercontent.com/37792313/55709798-6946a400-59e9-11e9-8354-46d150fd0aeb.png)

Do we want this to still apply to lines between classes but add the anchor point to make the line kinked, or do want to  somehow force which sides the line starts and ends at?

PR #6183, opened 2019-04-23
Grupp a2019 w17#5690

b17albbe:
Fix to issue https://github.com/HGustavs/LenaSYS/issues/5690
Added new button for UML lines, with "kink" line functionality. Created new symbol (7) for UML Lines, and also new uimode called "CreateUMLLine".


a17johpe:
We'll make this into a separate issue. Right now it's important that it just works and the rules for how the lines work can be changed

a17jacsv:
Works as expected. The issue a17vikab is talking about, please write an issue of it. Good job!

Issue #6179, opened 2019-04-23
Using draw square and draw free currently breaks diagram functionality


a14oscna:
A recent merge has added new complications where using the "Free" drawing objects breaks functionality. This has been tested over multiple PC:s and multiple browsers. When any of these two entities are drawn, any other type of object instantly disappears and now ONLY draw square and draw free entities can be drawn. Outlines that are shown while you are drawing to show how big the object will be are also gone. Some uncertain but probably related instances of where an object sticks to the cursor also occurred. If you refresh the page at this stage the Toolbar will also disappear. Clearing cache does not resolve the problem. This is a snapshot of the console when the error occurs: 
![diagram corners](https://user-images.githubusercontent.com/37793141/56608778-d8d4aa00-660b-11e9-8fa0-537d01bdbb2f.PNG)

PR #6186, opened 2019-04-24
Update diagram_objects.js #6179

a16tobfr:
#6179
Seems to work now. 
However, there is an error still present. It does not affect performance and I am not sure of why we get this error. I'll set it up as a new issue.
segments


a17jacsv:
Approved.

Issue #6191, opened 2019-04-24
Spaceing for coordinates and zoom-placements


a17vikab:
The placement of the coordinates/zoom boxes is right now on the canvasline while there is available space below.

See images for explanation:

![image](https://user-images.githubusercontent.com/37792464/56642794-8a112980-6678-11e9-9432-1acee59d23e9.png)

![image](https://user-images.githubusercontent.com/37792464/56642879-b88f0480-6678-11e9-9e90-d347635cdd75.png)

PR #6194, opened 2019-04-24
Grupp a2019 w17#6191

a17vikab:
Fix for: #6191 "Spaceing for coordinates and zoom-placements".


a17jacsv:
Looking good. Approved!

Issue #6177, opened 2019-04-23
Textbox selection does not have a good indication of when it is selected


a16tobfr:
When selecting and clicking on a  text object, there is no indication of it being the currently selected object. The hover and selected appearance are the same. Can be confusing.

It would be good if there were some indication that the text object really has been selected.

The lines around the textbox could for example change from dashed to a full line

![textSelected](https://user-images.githubusercontent.com/49199690/56588378-81224880-65e3-11e9-9bd1-40bfc3a537eb.JPG)

![noObjSelected](https://user-images.githubusercontent.com/49199690/56589102-e460aa80-65e4-11e9-9d71-3ae1f5e7d3cc.JPG)

PR #6227, opened 2019-04-24
Fix for issue #6177: Textbox selection does not have a good indication of when it is selected

a17jonro:
#6177


a17jacsv:
Looking good! Approved.

Issue #6016, opened 2019-04-17
Coordinates are not relative to origo / Y-value always negative


a14oscna:
Currently the coordinates on the diagram are relative to the diagram frame and not the origo. However, if top-left is supposedly the 0,0 cooridnate, moving the cursor further down should result in a positive Y-value, but that is not the case. In the image below origo was moved down close to the coordinates frame and the cursor was at the position of the red circle. See that neither the X or Y values are related to the position of origo and also that the Y-value is negative. 
![needscursor](https://user-images.githubusercontent.com/37793141/56272581-5758b580-60fb-11e9-84bb-28add6bbde04.PNG)

a17jacsv:
Interesting...

PR #6228, opened 2019-04-24
Changed Y-value to positive

a14oscna:
The X and Y values are still relative to the canvas top left corner and not the virtual origo. Now the Y value is no longer negative. 
#6016


a17alfcl:
RIP
Solved in new canvas logic so this fix is not necessary

Issue #5945, opened 2019-04-15
Move-tool hover in diagram


a17jacsv:
**See comments below...**

Create a move-button which makes it clear for the user of when they can move objects or not. Example of this is in "Photoshop" which is using this functionality.

 <img width="33" alt="Skärmavbild 2019-04-15 kl  13 50 13" src="https://user-images.githubusercontent.com/37792795/56130553-7fbea380-5f85-11e9-92e1-f06485f9e273.png">

<img width="238" alt="Skärmavbild 2019-04-15 kl  13 49 47" src="https://user-images.githubusercontent.com/37792795/56130538-78979580-5f85-11e9-96b6-08947d031472.png">

At this moment you can move when you not have selected an item in the toolbar which is not ideal in my opinion. This functionality gives the user more clarity of when and how they can move an item.

HGustavs:
The hover functionality can alter the cursor using css.

Unclear if it looks really bad or if there is a cursor that we can use to move things that doesnt cover up other actions such as clicking small stuff like relationship ends.

HGustavs:
Further info, we dont need a specific button since we are always in movement mode.

Anything that results in additional mouse clicks is not useful.

If we hover an existing entity etc, if we click we should be able to move/resize any object on-screen except for relationship ends.

HGustavs:
There used to be hover detection of ends of objects e.g. we are hovering right side of entity, we should then resize right side if we click button and move mouse.

a17jacsv:
I understand the purpose with limiting the ”clicks”, but i still think it would benefit the user with this tool.

However, changing the cursor when hovering might be good solution that should make it easier for the user instead of creation of another tool.

a17jacsv:
Something like this.
![Skärmavbild 2019-04-16 kl  09 50 42](https://user-images.githubusercontent.com/37792795/56191850-72f58a80-602d-11e9-9a89-6cc91d164114.png)

PR #6231, opened 2019-04-24
Update to diagram.js #5945

a16tobfr:
#5945
The cursor now changes appearance to make it more clear that the object is movable.


a16tobfr:
Good catch! Might be console.logs that I have removed and missed to remove the empty row afterwards. The issue should definitely not be merged into the Master branch. Good that you noticed this.

Changes have been made and the base branch is now set to gruppA2019W18 as it should be

a16tobfr:
I agree. This should not be possible. Good find, I'll start looking into it

a17jacsv:
Approved! Well done!

PR #6232, opened 2019-04-24
First step to fix canvas #5826

b17robev:
Partial fix for issue #5826 
Zoom not working at all but that should be another issue as it never worked properly


a17alfcl:
Pushing a new solution with zoom instead.
Closing this so they don't conflict

a17jacsv:
Good idea.

Issue #6256, opened 2019-04-24
Filter and sorting popups start in the navbar


a17ludca:
This looks odd and they should be moved down under the navbar.
![image](https://user-images.githubusercontent.com/37799006/56665854-218f7000-66ab-11e9-9953-f2218341a0a6.png)

HGustavs:
Agree

PR #6257, opened 2019-04-24
GruppB2019 w18 #6256

a17ludca:
Quick adjustment of the dropdown-containers

### DIFF
--- a/DuggaSys/resulted.js
+++ b/DuggaSys/resulted.js
@@ -214,8 +214,8 @@ function process() {
 		var lid = moments[j].lid;
 		var name = momtmp[j].entryname;
 		var truncatedname = name;
-		if (truncatedname.length > 12) {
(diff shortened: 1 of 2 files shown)


b17nicsv:
The popup still starts at the navbar. Additionally, why doesn't the funnel or the sort function work in the Access tab if they are visible? The popup still needs correction and I would suggest (perhaps making another issue) to remove the funnel and sort icons from Access since they don't work and there's already a way to sort the different columns.

a17pongo:
the funnel and sort got broken during one of our first merges, im gonna check with henrik on what he thinks the current workflow should be.
@b17nicsv if there are changes that should be made on the current issue then please as a reviewer mark the pull requests as "changes requested"

a17pongo:
What is the status on this?
also, hiwah has made a pull request in this issue that supposidly fixes it.. is it ok to close this pull request?

b17nicsv:
Hiwahs pull request solved the problem. It is okay to close this pull request.

PR #6260, opened 2019-04-25
Fix with zoom #5826

a17alfcl:
Fix for issue #5826
Hopefully does not break any major parts of the program.


a17jacsv:
> You need to add documentation on the new variables that you have created/changed. Make sure that other functions are well documented so other people can work from this as well without confusion.

I agree...

a17jacsv:
> The solution seems to fix some of the coordinator issues that was previously mentioned. The problems that are still present will be divided into new issues.
> The added documentation also looks good and will make it easier to maintain the code in the future.
> Well done!

Agreed! Nice job! 👍

Issue #6151, opened 2019-04-23
Aligning with locked items doesn't work propperly


a17drama:
![Edited](https://user-images.githubusercontent.com/46029340/56568860-27585900-65b8-11e9-83f7-149038e74180.PNG)
This is the result of vertical alignment. As the picture shows the alignment is a bit off.

a17vikab:
seems to be fixed.

PR #6275, opened 2019-04-25
Fix for issue with aligning locked objects #6151

a16jenja:
Fix for issue #6151 

The alignment should not be a bit off anymore. It should instantly align correctly when pressing the vertical and horizontal alignment button with objects locked now. Before you had to press the button multiple times for it to align properly.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -931,12 +931,12 @@ function toggleVirtualA4Holes() {
 }
 
 //---------------------------------------
-// resetToolbar: resets the toolbar to it's original position 
+// resetToolbar: resets the toolbar to it's original position
 //---------------------------------------
 function resetToolbarPosition(){
     var myCanvas = document.getElementById('myCanvas');
     var bound = myCanvas.getBoundingClientRect();
-    //Assign position for the toolbar according to the canvas bounds 
+    //Assign position for the toolbar according to the canvas bounds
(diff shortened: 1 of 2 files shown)


a16jenja:
I kinda agree with you, but it's hard to decide what the lock on an object should actually do. Since it can be beneficial to be able to align every object even if they are locked and avoid the hassle of unlocking every object when you want to align them. But it can also be the opposite of course, and be an annoying function.

I am not sure about which direction we should go, regarding if an object that is locked should move when aligned but not be moved by dragging or not be able to move at all. I would like to get @a17drama and @a17jacsv opinion on this. And I will make the changes accordingly.

a17drama:
Locking means that all possibilities to edit the entity are gone. When it's locked it's locked.

a17jacsv:
> Locking means that all possibilities to edit the entity are gone. When it's locked it's locked.
I agree with @a17drama. Aligning with locked entities included, the locked entities should be excluded when aligning items.

a16jenja:
Thanks for the clarification, I will get right on it.

a16jenja:
Ignore the code added before since it only makes it so the lock will temporarily unlock and the locked object will align with the rest of them, but as previously stated we don't want the lock to move at all when aligning.

a17alfcl:
Creating new branch after merge

Issue #6222, opened 2019-04-24
Content div in sectioned.php has excessive margin


b17carma:
In my opinion, the left, right and bottom margin on the content div in sectioned.php should be completely removed, but I would like to know what other people think.

![Capture](https://user-images.githubusercontent.com/37834592/56657631-d79d8e80-6698-11e9-9e2d-181ada1d7b82.PNG)

a17maxsj:
I think this might have something to do with material design papers, not sure if it would make the page better or worse. Mockup maybe 😄

a17andby:
Just merged a fix for this issue. Closing.

PR #6276, opened 2019-04-25
Adjust margin of content div & TopMenuStatic in sectioned.php

b17carma:
Fix for issue https://github.com/HGustavs/LenaSYS/issues/6222


a17joaap:
I liked the breathing room on the left and right, but it looks better without the bottom margin.

b17carma:
Re-added left & right margins 😃 @a17joaap

Issue #6246, opened 2019-04-24
Exporting diagram as image should be PNG


a17johpe:
When exporting a diagram as an image, the default file type is a JPEG. This should be exported as a PNG file instead, since it contains text and graphics.

HGustavs:
Yes this is better....

PR #6281, opened 2019-04-25
Changed image file from .jpg to .png - Fix for #6246

a14oscna:
A simple fix which changed the file type for the export image menu option from .jpg to .png

#6246


a17jacsv:
Good addition! Approved 👍

Issue #6210, opened 2019-04-24
Cookie/MOTD Message in courseed.php overlaps the new course FAB


b17carma:
![Capture](https://user-images.githubusercontent.com/37834592/56656390-aec7ca00-6695-11e9-9257-a54624ea62af.PNG)

a17maxsj:
maby put the motd bar under the navbar? could work

b16johso:
+1

a17maxsj:
Rough mockup, maybe with a nice fold out effect
![eb761795c0be0904bef9f1fcefc68555-png](https://user-images.githubusercontent.com/49199993/56659582-cd31c380-669d-11e9-9d4a-c8da33870c75.png)

a17maxsj:
Maybe add error messages also in this bar at a later stage

b16johso:
After discussing this with some other group members, we decided we want the MOTD message to appear at the top of the page, under the navbar, but leave the cookie message at the bottom. However, this means the cookie message still overlaps the FAB, so my suggestion is to change the styling of it. Here's what I came up with:

![bild](https://user-images.githubusercontent.com/37793460/56725942-1e00f500-674e-11e9-8b63-5b556cd0be2a.png)

b16johso:
Also animations for showing and hiding the messages would look nice.

a17maxsj:
Looks good, further input from the group would be nice!

a17joaap:
Looks good, the MOTD was really annoying before but this makes it less so

a16joare:
I think it looks much better, GJ!

b17annni:
I approve

a16lisri:
I think it looks better too. Nice!

b17carma:
I agree, this looks a lot better 😃

a17maxsj:
Was this implemented? @b16johso  i think it should be added before merge!

a17andby:
I approve as well. Merge this solution to our group branch before the merge.

PR #6284, opened 2019-04-25
Grupp c2019 w18 6219

b16johso:
https://github.com/HGustavs/LenaSYS/issues/6210


b16johso:
Updated this. Not the best solution, but it was the only one that worked without rewriting the code.

Issue #6029, opened 2019-04-17
Coordinates to do not update when mouse is right below the menu bar


a16jenja:
![chrome_E9x2Z5TIM8](https://user-images.githubusercontent.com/49199287/56277347-7740a700-6104-11e9-905c-b71bdc1d0498.png)
Here it works like it should.

![chrome_0CPxU7qGar](https://user-images.githubusercontent.com/49199287/56277433-a1926480-6104-11e9-8d57-5bdcd5f9002d.png)
Here the x-value should be about the same as the other image, but its not. The problem is that the coordinates do not update when holding the mouse right below the menu bar. Therefore it's likely that the issue has something to do with the menu bar.

a17drama:
This issue is older than the merge. It has nothing to do with today's merge to do so I'm removing the tag. But the issue should definitely be solved.

PR #6286, opened 2019-04-25
Fix for issue #6029

b17robev:
Fix for issue #6029

Removed fixed position as it is no longer needed because scrolling isn't possible anymore. 
Added margin-bottom as well for style points and improved UX.


a17jacsv:
Good fix. Approved!

Issue #6015, opened 2019-04-17
Hover effect on buttons in toolbox


a17okaoz:
When the pointer is above a button in the toolbox make so the button have a hover effect that will show witch button you are on.

a17drama:
You mean like a tool tip?

a17okaoz:
Tool tip already exists. I think like the background color will change to when the pointer is over a button. The hover color could be the same as when a button is clicked

a14oscna:
If hover is the same as button clicked, you would not be able to differentiate between a click and a hover, which could be annoying for people using a touchpad who generally have less physical feedback already. I suggest just changing the cursor icon or similar.

a17okaoz:
The color could be a lighter tone of the existing color that is now. Just a simpel visiuell effect that will help the user a little if they miss the tool tip below

PR #6299, opened 2019-04-25
Fixed issue with clearer hover effects on toolbox buttons #6015

a16jenja:
Fixed issue #6015 

I changed the cursor icon to a pointer when hovering over the toolbox buttons. I could add a small visual effect aswell, but I feel that the pointer is clear enough to let the user know that they are hovering over a button.


a16jenja:
You are correct, should have changed in the button div instead of the image div. It should be solved now.

a17jacsv:
Works as expected! Well done.

Issue #6105, opened 2019-04-18
Remove all-option in the toolbar


a17jacsv:
Remove the all-option in the toolbar as it isn’t needed. It just makes it difficult for new users who trying to learn to model.

<img width="238" alt="Skärmavbild 2019-04-18 kl  10 23 26" src="https://user-images.githubusercontent.com/37792795/56347353-65263d80-61c4-11e9-9b7e-dd8b2776715f.png">

a17drama:
Would be appreciated if we could keep the ALL alternative while we are in developer mode. This makes developing/testing easier when an experienced developer doesn't have to switch between different modes.

Issue #6108, opened 2019-04-18
Move the changing of model-type in the menu instead of toolbar


a17jacsv:
**This option should be moved to the menu instead of the toolbar since very rarely we change type in a diagram as Henrik said.**

Instead of changing the types here, move it to the menu instead.
<img width="187" alt="Skärmavbild 2019-04-18 kl  10 33 15" src="https://user-images.githubusercontent.com/37792795/56347934-aec35800-61c5-11e9-806a-6d81eaa760a2.png">

 I suggests under the view section.
<img width="372" alt="Skärmavbild 2019-04-18 kl  10 36 17" src="https://user-images.githubusercontent.com/37792795/56348015-da464280-61c5-11e9-8b78-5f14b3c12185.png">

HGustavs:
Much better in menu

Issue #6106, opened 2019-04-18
Remove the free-option in the toolbar


a17jacsv:
The free draw-option-items should be visible in the ER/UML section of the toolbar as it should be available in both. Remove the free-option in the toolbar as well.

<img width="174" alt="Skärmavbild 2019-04-18 kl  10 27 50" src="https://user-images.githubusercontent.com/37792795/56347491-a9194280-61c4-11e9-868a-bccaef4cf681.png">

HGustavs:
Yeah. free is not a diagram type for sure.

It would be a really strage work-process to highlight something without drawing some diagram to highlight....

Issue #5934, opened 2019-04-15
Toolbar should have fixed buttons for switching diagram modes


a17johpe:
When in the toolbar's horisontal mode, switching between different diagram modes (e.g. ER, UML) results the left and right arrow buttons moving around since it's centered relative to the size of the toolbar.
![toolbarhoris1](https://user-images.githubusercontent.com/37792313/56127944-9dd4d580-5f7e-11e9-930e-ccb77ec2b490.png)
![toolbarhoris2](https://user-images.githubusercontent.com/37792313/56127947-a0372f80-5f7e-11e9-856d-4bb8c6be4a51.png)
The problem is that when you're trying to get to a specific toolbox mode, you have to move the cursor every time to find the buttons again. This is a minor user experience issue.

HGustavs:
I think this option should be moved to the menu instead since very rarely we change type in a diagram

Separate issue

We should not allow ER entities in a uml diagram and vice versa

Free draw should be available in both / either

a17jacsv:
Remove all-option in the toolbar: #6105 
Remove the free-option in the toolbar: #6106

a17drama:
@a17vikab Considering this is a parent issue for those that you are assigned this goes to you so we can tag all of them in the pull request and close all of them.

a17jacsv:
I approve this...

PR #6301, opened 2019-04-25
Fix for #6105 #6106 #6108 #5934

a17vikab:
This should fix: #6105, #6106, #6108, #5934.


a17jacsv:
Approved! Looking good!

Issue #6166, opened 2019-04-23
 Make variables more readable by adding values ​​to attributes


a17okaoz:
Put all the values in variables to make the code more readable. Insted of symbolkind == 3 put the value 3 into a variable. 

<img width="260" alt="b1" src="https://user-images.githubusercontent.com/37792681/56581505-6b0d8b80-65d5-11e9-9b59-3e0dae60e902.PNG">

PR #6307, opened 2019-04-26
Fix for issue #6166: Make variables more readable by adding values ​​to attributes

a17jonro:
#6166


a17jonro:
I also found kind being assigned to numbers in this.recursetest in diagram_objects.js and changed them as well.

a17johpe:
Looks good. Be aware that of this change in future pull requests and merges though, in case people still use the old numbers as codes instead of the variables.

a17jacsv:
Approved! Good work.

PR #6308, opened 2019-04-26
Version ID removed from hover

### LINKED ISSUE #6291: cogwheel hover on courseed.php display a strange title.
If you hover the cogwheel for a course on courseed.php a kind of strange title will be shown.
Image of the title: 
image

The value after the course name in the title is linked to the Version ID of the course and not the course code as i first thought.. 
If you click to edit the course this window will appear:
image

@HGustavs is this on purpose or is it some kind of misslinking?  Feels strange that the version ID will be shown in the title when you cant change it when you press the cogwheel. 
I suggest that the title will be changed to include the "course name" + "course ID" or all of it is removed and the title is changed to "Edit Course". To be honest i prefer the last suggestion..
What do you think?

a17alian:
Version ID removed from title on cogwheel hover in courseed. Only course name should appear. 
Closes #6291
Test site: https://a17alian.groupb.webug.his.se:20002/DuggaSys/courseed.php


a17anter:
Test OK!

Issue #6274, opened 2019-04-25
Improve rules for UML Lines


b17albbe:
The UML Line tool creates an unnecessary amount of kinks and breakpoints in certain situations, and should adapt depending on certain positions to make the best line path possible. Here's an example on how that could look like:
![11db3d4ba6b44c4db0f713ec66379b0b](https://user-images.githubusercontent.com/38030926/56729734-442a9300-6756-11e9-904b-2119a82fa7f6.png)
![11db3d4ba6b4hj4c4db0f713ec66379b0b](https://user-images.githubusercontent.com/38030926/56729749-50165500-6756-11e9-86db-21485ea4789d.png)

a17jacsv:
I agree...

PR #6310, opened 2019-04-26
Fixed UML lines: #6274

a17johpe:
https://github.com/HGustavs/LenaSYS/issues/6274

UML lines are now smarter, using only as many kinks as are necessary.
(That the line breaks in when the UML classes get too close to each other is a separate issue that will be solved later)


a17jacsv:
Looking good. Well done!

a17johpe:
Created issue https://github.com/HGustavs/LenaSYS/issues/6322 about the broken lines that were mentioned.

Issue #6202, opened 2019-04-24
In sorting-options, swedish characters are not available


a17ludca:
When opening the sorting-options, some characters are not showing as intended. Also the way of shortening the words is mildly confusing. 
![image](https://user-images.githubusercontent.com/37799006/56648698-107f3880-6684-11e9-94a3-4870a0997703.png)

HGustavs:
Unclear what happens....

could it be that it cuts up the text inside &ouml; ?

We want to see the end and beginning of the items if possible with a reasonable amount of code.

If the duggas are called

Test Duggasvar Fråga 1
Test Duggasvar Fråga 2
Test Duggasvar Fråga 3
Real Duggasvar Fråga 1
Real Duggasvar Fråga 2
Real Duggasvar Fråga 3

If we shorten first part we only see

Test
Test
Test
Real
Real 
Real

If we show only end we see

Fråga 1
Fråga 2
Fråga 3
Fråga 1
Fråga 2
Fråga 3

...

By shortening the middle but neither beginning nor end we get best of both worlds...

b17timjo:
It is almost certainly the case that an "&auml;" is string-replaced with "..."

PR #6311, opened 2019-04-26
Grupp b2019 w18#6202

a16hiwah:
if -statement on line 217(resulted.js) has been changed to 30. and entryname.length from 8 to 30.
Test site: https://a16hiwah.webug.his.se:20002/DuggaSys/resulted.php?cid=2&coursevers=97732

issue_6202

lssu_6202

### DIFF
--- a/DuggaSys/resulted.js
+++ b/DuggaSys/resulted.js
@@ -219,9 +219,9 @@ function process() {
 		}
 
 
-		dstr += "<div class='checkbox-dugga checknarrow ";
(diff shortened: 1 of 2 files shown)


c17johha:
#6202

a17ludca:
It definitely looks better but the sorting still has no functionality so while this issue is resolved, a new one should be created about the functionality.  Try to include the number on for example, "Biträkningsdugga 1". 
Now it's Biträkningsdug but it could be Biträkningsdug 1.

Issue #6180, opened 2019-04-23
Objects move if you attempt to resize them smaller than their default size


a14oscna:
When resizing any entity / attribute / class / relation they **move** if you attempt to resize them smaller than their default size. The result is different for when you drag to resize by each of the four corners and here is a brief summary:
**Top-left**: horizontal and vertical movement
**Top-right**: vertical movement
**Bottom-left**: horizontal movement
**Bottom-right**: NO movement

The issue is prevalent for 3/4 corners. Every corner should mimic the functionality of the **bottom-right** corner which has no movement. 

This issue has already been covered partly for the relation object in issue #5884

PR #6316, opened 2019-04-26
Fix resizing objects in diagram - #6180

c17matjo:
#6180


c17matjo:
Okej. Jag har lagt till ett call till updateGraphics() som jag tror borde fixa problemet. Testa nu

c17matjo:
I talked with @a17drama and we decided that since this is a minor bug and since the problem described in the issue was solved, a new issue should be created for the new bug.

a17jacsv:
> I talked with @a17drama and we decided that since this is a minor bug and since the problem described in the issue was solved, a new issue should be created for the new bug.

I'll merge this and the bugs will be resolved in another issues.

a14oscna:
Alright, looks good then!

Issue #6211, opened 2019-04-24
Errors when pressing Enter in sectioned.php


c17julka:
![image](https://user-images.githubusercontent.com/37792321/56656250-4b3d9c80-6695-11e9-8d4d-13bb0274f4a4.png)
![image](https://user-images.githubusercontent.com/37792321/56656267-5c86a900-6695-11e9-955a-52cf1ee59cd9.png)
This error appears if you press Enter while editing an item in sectioned.

![image](https://user-images.githubusercontent.com/37792321/56656314-7aeca480-6695-11e9-9708-69bf38aafa50.png)
If you press Enter in just sectioned, this error appears.

Relevant code is in the $(window).keyup function. The functions "isNameValid()" and "isTypeValid()" aren't defined anywhere.
`if (saveButtonDisplay == 'block' && editSectionDisplay == 'flex' && isNameValid()) {
      updateItem();
    }`

`else if (isTypeValid() && testsAvailable == true) {
      confirmBox("closeConfirmBox");
      testsAvailable = false;
    }`

c17julka:
Same error with isTypeValid() appears when you press enter while editing/adding course.

HGustavs:
The validation code should be removed completely in that view as it is not possible to validate in a succint way, and I may have not done that removal correctly ...

c17julka:
The errors when pressing enter still appears, as they are still used in the key event function.

a17adasy:
https://a17adasy.gruppb.webug.his.se:20002/DuggaSys/sectioned.php?courseid=2&coursename=Webbutveckling%20-%20datorgrafik&coursevers=97732

The errors do no appear for me on my test site which has the issue 6211 branch right now.

c17julka:
Seems my branch had messed up and not updated correctly, my bad.

a17pongo:
@c17julka so did I just merge something that did not fix the issue?

a17pongo:
@c17julka so did I just merge something that did not fix the issue?

PR #6325, opened 2019-04-26
Removed functions that were never created

a17adasy:
#6211

The functions isTypeValid and isNameValid were used in if statements but were never created in the file, which caused errors. They are gone now and so are the errors.

This caused errors in the sectioned part of dugga.


a17amama:
Look inside duggaed.js  and search for isNameValid, look into if it is actually used or not and if it can be removed.

a17adasy:
The duggaed.js file is never linked in the sectioned.php file, so it shouldnt matter!

a17amama:
Looks good, can be merged.

Issue #6143, opened 2019-04-23
Create a smoother "zoom-changer"


a17jacsv:
Change zoom-button to picture below to change the zoom of the diagram instead of clicking a button. This solution decreases the number of clicks and makes it more user-friendly and a smoother view.

(The picture below is from Word)
<img width="200" alt="Skärmavbild 2019-04-23 kl  10 05 40" src="https://user-images.githubusercontent.com/37792795/56565185-173c7b80-65b0-11e9-8168-7fa39bee0e60.png">

HGustavs:
Yes!

But stepped if possible... or with automatic re-centering of diagram.

Soft zoom without auto-centering is motion-sickness-inducing as content moves latteraly when zooming without dynamic centering of view.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range

a17drama:
Check out #6018 and make sure these don't destroy each other.

a17jacsv:
Please delete the zoompart in the "coord-bar" as it gets redundant...
<img width="304" alt="Skärmavbild 2019-04-25 kl  10 32 37" src="https://user-images.githubusercontent.com/37792795/56721604-88616780-6745-11e9-8434-494d4618d85a.png">

PR #6333, opened 2019-04-26
Grupp a2019 w18#6143

a17okaoz:
Issue #6143 fixed. The zooming is now a range that you can drag to zoom in and out. There are buttons beside the range that can zoom in and out. The Zoom value are now moved beside the zoom in button and updates the zoom value. Have removed some unnecessary code that have no use anymore.


b17robev:
Create new pull request as this one has the old canvas logic

Issue #6269, opened 2019-04-25
Modification of the virtual A4


a17drama:
When showing the virtual A4 the user should be able to chose if it's in vertical or horizontal mode.

PR #6336, opened 2019-04-26
Fix for issue #6269

b17robev:
Fix for issue #6269

Toggle for orientation of A4, icon should be changed when toggling and hid when a4 is not displayed


a17jacsv:
Looking good! Well done.

PR #6306, opened 2019-04-26
Grupp B - tooltip for searchbar that shows keywords

a17adasy:
#6248

Hover over the questionmark next to the searchbar on the resulted page.

Link for testing: https://a17adasy.gruppb.webug.his.se:20002/DuggaSys/resulted.php

Issue #6253, opened 2019-04-24
Search bar in resulted.php should be sticky


c17julka:
Scrolling horizontally back and forwards can be annoying. Would be easier if the search bar was sticky so you don't have to scroll this much.

HGustavs:
Or up in the already static navigation menu.

I think that there is some other issue (other service than this?) that has already discussed this, and if I am not mistaken I asked for a mock-up of how this may look.

c17julka:
![image](https://user-images.githubusercontent.com/37792321/56720429-1b4cd280-6743-11e9-978a-53de14ab86ef.png)

@HGustavs Here's a mock-up picture of a suggested move. Would this be a viable move?

PR #6337, opened 2019-04-26
Search bar moved to header

c17julka:
#6253 
In resulted, accessed, fileed and duggaed, the search bar is now in the header. 

Might have issues with the tooltip fix (#6306). The tooltip from that pull request was implemented into this branch with some adjustments. If there are any search bar elements left in any of the .php files (like the tooltip might be), they should be removed.

Test site: https://c17julka.gruppb.webug.his.se:20002/


c17alepe:
looks very gooood. test ok

Issue #6195, opened 2019-04-24
Weird breakpoint when trying to select lines


a17johpe:
![marktoolline](https://user-images.githubusercontent.com/37792313/56644250-a1054b00-667b-11e9-8f61-7ccae2edaab2.png)
When using the cursor to draw a rectangle to select multiple objects, there is a weird breakpoint where lines aren't highlighted/selected if the select box isn't wide enough.

Edit: notice in the image how the selection tool encompasses the attribute, the entity, and the line, but the line is not highlighted to mark it as selected.

PR #6338, opened 2019-04-26
Fix for #6195 - Selection is now pixel perfect

a14oscna:
#6195 
Removed two odd if-statements which required the selection box to cover a much larger area than needed to select objects within it.


a17jacsv:
OK.

Issue #5748, opened 2019-04-09
Right click drag scrolling


a17drama:
Implement a new function where you can right click the canvas and drag to explore the canvas. In other words instead of scrolling you should be able to click and drag.

a17drama:
There is a similar function that activates when spacebar is pressed but this moves the canvas around. My suggestion involves a camera object that is moved instead and canvas is static. This is related to issue #5721. This would also solve the problem with #5723. Moving canvas causes a lot of problems and takes unnecessary computing power to recalculate the new positions on the canvas with the offset that is caused because of the move.

HGustavs:
The canvas should be static and the graphics move inside the canvas IMHO.

a17drama:
This issue is related to #5826 and is a part of the functionality for that issue.

PR #6339, opened 2019-04-26
Grupp a2019 w18#5748

a16tobfr:
#5748 
The solution to the issue seems to work well. You can now right click and drag to move the camera
More comments about the functionality are made in the latest commit.


a17jacsv:
Approved! Well done.

Issue #6327, opened 2019-04-26
Reset view to origin doesn't work 


a17jonro:
Some functionality seems to have been removed when implementing the camera. Zoom also seems to break functionality.

HGustavs:
If implemented correctly the reset should only set the xoffset to 0 yffset to 0 and set the zoom to 1.0

...

and then refresh the view, and it should be mostly automatic from there?

Or are we talking about some other reset?

a17jonro:
No, that is correct. The reset view to origin function only needs to be adapted to the new canvas implementation, since it was different before the change.

PR #6341, opened 2019-04-26
Fix for issue #6327: Reset view to origin doesn't work 

a17jonro:
#6327


a17jacsv:
Looking good. Works as expected!

Issue #6263, opened 2019-04-25
While not zoomed on 100%, the panning is not stuck on the cursor.


a17vikab:
While not zoomed on 100%, the panning is not stuck on the cursor. The best way to describe this is to show the result:

GIF of 50%: https://gyazo.com/a070e9f95eded5651ad0bc68f675df68

This is the result we want on all the zoom-options: https://gyazo.com/eec60e1d3d1022dfadc1e5ad544ffe68

This issue is related to: #5826

PR #6345, opened 2019-04-26
Fix for issue #6263

b17robev:
Fix for issue #6263 

Math now properly included


a17jacsv:
Looking good. Works as expected.

PR #6365, opened 2019-04-26
Fix for issue#6362

b17robev:
Uses function on-input rather than onChange, minor UI change to border-radius on buttons


a17jacsv:
Approved!

Issue #6309, opened 2019-04-26
Should not be possible to change appearance when locked


a17drama:
Locked items shouldn't be editable in any way. No moving, no resizing, no change of appearance.

a17jacsv:
I agree...

PR #6368, opened 2019-04-26
Fix for #6309 - Locked objects can not be interacted with

a14oscna:
You can no longer resize or open the appearance menu for locked objects.
#6309


a17jacsv:
Approved!

PR #6373, opened 2019-04-26
Grupp a2019 w18#5945

a16tobfr:
#5945 
Fixed the issue again after the realization that the hovobj variable had been changed for some reason. Now it seems like it works as it did before the issue was merged into the group branch.


a17jacsv:
Looking good! Well done

Issue #6220, opened 2019-04-24
Cant exit preview window in showdugga.php


c17alepe:
![image](https://user-images.githubusercontent.com/38323901/56657124-89d45680-6697-11e9-83c6-35b41df00c85.png)

Whenever you open a text or pdf file that you have submitted in showdugga.php, you cant close it any other way than clicking the back button, which takes you to the previous site.

The image is an example of opening a textfile and being trapped.

HGustavs:
This must be some kind of styling issue?

There used to be buttons here?

PR #6374, opened 2019-04-26
Grupp b2019 w18#6220

c17alepe:
before
image

after
image

Moved a header which was hidden behind the absolute submission window so that you can escape the previewpopup. It might be a good idea to see if any other boxes gets messed up by the style.css change. the loginbox class is for some reason used for the previewwindow.

#6220

test site, log in and click the text submission to see the change:
https://c17alepe.webug.his.se:20002/DuggaSys/showDugga.php?did=74&cid=324&coursevers=12324&moment=2328&segment=2386&highscoremode=0&comment=UNK&deadline=2019-05-15%2007:15:00

NOTE TO ISSUE-MASTER: The branch i created has the wrong issuenumber. it says 6022 but it should be 6220. Let me know if this is a problem

### DIFF
--- a/DuggaSys/showDugga.php
+++ b/DuggaSys/showDugga.php
@@ -207,7 +207,7 @@
 
   <!--<div id='previewpopover' class='previewPopover' style='display:none;'>-->
   <div id='previewpopover' class='loginBoxContainer' style='display:none; align-items:stretch;'>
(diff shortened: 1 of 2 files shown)


a17pongo:
@c17alepe the wrong issue number fro branch name is ok since they will be deleted and yoy have references the correct issue in this pull request

Issue #6375, opened 2019-04-29
Add a hover effect to the zoom in and out buttons


a17okaoz:
Add a hover effect to the zoom in and out buttons. To follow the structure of the other buttons they have a cursor change when the the pointer is over the button.

PR #6378, opened 2019-04-29
Grupp a2019 w18#6375

a17okaoz:
Issue #6375 fixed. Added cursor: pointer; to the Zoom increase and Zoom decrease buttons.


a17jacsv:
OK.

Issue #6264, opened 2019-04-25
Icon for moving canvas is not placed correctly.


a17vikab:
The icon should be inside the canvas and not on the edge of the browser.

![image](https://user-images.githubusercontent.com/37792464/56721321-eb063380-6744-11e9-8cb1-22436466d204.png)

a17drama:
Good to see I'm not the only one getting bugged by this.

PR #6381, opened 2019-04-29
Fix for #6264 Changed style value for move button

a14oscna:
The move button has be adjusted a bit to the left to fit at the corner of the canvas.
#6264


a17jacsv:
Nice addition. Good job!

Issue #6380, opened 2019-04-29
Sectionedservice.php broke from a merge


a17poner:
You can't log as a normal user registered on a course and then enter the course you are registered to. Found the cause as a merge conflict between group B and C which was dealt with poorly and resulted in duplicate code except the duped code ended up further down in the file with the wrong query data causing undefined index errors.

a17andby:
After two reviews I merged the solution. Closing this issue

PR #6383, opened 2019-04-29
Removed duplicate code and moved a code section to the correct line

a17poner:
fixing issue #6380 by moving and deleting code. also deleting some comments that clutter the code.
Should be able to log in as a normal registered to the course user and see the coursepage instead of "UNK".


a17poner:
testdata.sql added user s94sveri with password "password". a lot of others but that is the user that I've used to test and confirm this with myself.

Issue #6354, opened 2019-04-26
Deselection of draw line


a17drama:
After the line is drawn the tool for drawing line should be deselected. This is irritating when the user is trying to move an object after the line is drawn and it starts drawing a new line.

a17jonro:
Are we sure about this? You might want to be able to draw multiple lines without having to click the button multiple times. At the moment it is irritating if you want to move an object after drawing a line but if we change how it works then it will become irritating the other way around. And since all of the other create/draw works this way it would be weird to only change this one.

a17drama:
I fully agree but our customer (@HGustavs ) has a wish to have it this way and the customer is always right. For the future to work around this problem we could implement hotkeys for tool bar to chose the different tools. This will remove the need for moving and clicking but the user will still have to do something to chose drawing lines. If there is a better suggestion it would be appreciated if this came up as an issue.

PR #6384, opened 2019-04-29
Fix for issue #6354: Deselection of draw line

a17jonro:
#6354 
The line tool is only deselected when the line has been created, so that if you don't draw a line to an object then the line the tool will not be deselected and you can draw as it was intended. This behaviour can be changed though, depending on how we want it to work.


a17jacsv:
User-friendly addition. Good job!

Issue #6390, opened 2019-04-29
Incorrect comment


a17okaoz:
Change the comment so it is correct with the code. Now the comment dosen´t tell the right thing. The code is for moving a object when it are pressed and the comment says 
"Handler for when pressing arrow keys when space has been pressed".

<img width="333" alt="b1" src="https://user-images.githubusercontent.com/37792681/56886374-308d6e00-6a6f-11e9-97e4-b7098959b83b.PNG">

PR #6393, opened 2019-04-29
Grupp a2019 w18#6390

a17okaoz:
Issue #6390 fixed. Have updated a comment in the function arrowKeyPressed so the comment tells exactly what the function does


a17okaoz:
Some changes happend that i haven´t done

Issue #6391, opened 2019-04-29
Second search bar in resulted and accessed


c17julka:
remove the one not in header

PR #6396, opened 2019-04-29
removed second search bar

c17julka:
#6391 

Test site: https://c17julka.gruppb.webug.his.se:20002/DuggaSys/sectioned.php?courseid=2&coursevers=97732


c17alepe:
It looks very clean now! Test ok

Issue #6268, opened 2019-04-25
Move the "Filtrera efter" dropdown


a17adasy:
![image](https://user-images.githubusercontent.com/37799016/56722600-b9db3280-6747-11e9-8a6e-9d3876f0a89f.png)

the "Filtrera efter" dropdown should probably be moved elsewhere since it is not connected to the ladexport button like the other options next to it.

PR #6403, opened 2019-04-29
moved the 'filtrera efter'

c17alepe:
before
image

after
image

#6268 

I moved the select so its right of the buttons. i also expanded the margin on the buttons for a more consistant look.

testsite: https://c17alepe.webug.his.se:20002/DuggaSys/resulted.php?cid=2&coursevers=97732


c17julka:
Looks good!

a16hiwah:
test ok is looks better no

Issue #6357, opened 2019-04-26
Line thickness when zooming


a17drama:
When using zoom the lines for all entities should be affected. Zooming in makes lines thicker and zooming out thinner.

PR #6404, opened 2019-04-29
Zoom value implemented everywhere

a17alfcl:
Fix for issue #6357


a17drama:
Do a detailed check of this so we don't miss anything. ER/UML modes and all possible entities including the free draw so all behaves correctly.

a17jacsv:
Looking good! Approved.

Issue #6322, opened 2019-04-26
UML lines break when classes are too close to each other


a17johpe:
![brokenumlline](https://user-images.githubusercontent.com/37792313/56794405-5cf67f80-680e-11e9-8230-c1ad15d08dac.PNG)
When classes are placed too close to each other, the lines break. This is because of fixed breakpoints and needs to be solved.

PR #6416, opened 2019-04-29
Fix for UML lines #6322

a17johpe:
https://github.com/HGustavs/LenaSYS/issues/6322

Set breakpoints to same as start and end points when classes are too close to each other.


a17jacsv:
Approved.

Issue #6121, opened 2019-04-18
Maximize utilization of work space


a17drama:
![Edited](https://user-images.githubusercontent.com/46029340/56360642-919e8180-61e5-11e9-9403-60c49f0adce3.PNG)
The margins around the canvas are not equal and is way to big on the down side. Margins are a waste of work space and should be minimized to use the screen as efficiently as possible. Screenshot is placed on the black background so it's easier to see where the margins end.

HGustavs:
Efficient use of spce is a priority since we want this to be used on for example tablets for drawing content.

PR #6417, opened 2019-04-29
Fix for #6121

a17celal:
#6121 

Made the canvas work space bigger and fixed so that the margins around the canvas are equal.


a17celal:
Not sure what you mean.. Which boxes?

a17celal:
> Could you make it so that the grey corner of the boxes covers the entire black corner of the canvas?

Not sure what you mean.. Which boxes?

a17jacsv:
Approved!

PR #6418, opened 2019-04-29
Update diagram.js

a17okaoz:
Issue #6390_2 Fixed. The comment is updated. Why there are many changes is beacuse I use Atom. It will 
automatically remove whitespaces


a17jacsv:
OK.

PR #6423, opened 2019-04-29
Updated import instructions in access

### LINKED ISSUE #5940: Improve user importing instructions in accessed.php
image

As mentioned in #5698 the importing interface is fine as it is but the instructions/example for how to import is wrong. Now its a type of csv (not sure what it looks like) and not what it currently says on the page it is. It should probably be updated to help people new to lenaSys in the future.

a17joaap:
Fix #5940

### DIFF
--- a/DuggaSys/accessed.php
+++ b/DuggaSys/accessed.php
@@ -59,11 +59,10 @@
       		</div>
       		<div class='note'>
                   <p>Users must be separated with a linebreak and the format required for each user is as follows:</p>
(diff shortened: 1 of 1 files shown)


a17andby:
Won't merge this pull request before we get a response from @HGustavs on https://github.com/HGustavs/LenaSYS/issues/6425 issue.

HGustavs:
Make it work with given data.

If it does not generate the corresponding individual/individuals ... wash rince repeat.

Issue #6382, opened 2019-04-29
Warning when switching modes


a17drama:
User should be warned when switching between ER and UML mode. This is to help prevent unintentional faults in diagrams. Dialog box with some kind of confirm should be sufficient.

a17jonro:
Should this only appply to the ER and UML modes and not the developer mode? Since i feel like either it could work differently when you're in developer mode, such as displaying no warnings OR the warnings could appear in developer mode "display all tools" as well. What do you think @a17drama ?

a17drama:
Developer mode is for experienced users and a developer should have knowledge of these risks already. The goal with the whole issue is to cover the normal use when beginners are using the diagram. It's to help them so they don't mix the modes and do things wrong (draw in ER entities in a UML diagram or the other way around). Thinking of it now we maybe should even lock the possibility of switching when there is an entity drawn. Switching should only be possible when the diagram is clear. That way we can prevent errors with a poka-yoke.

PR #6426, opened 2019-04-29
Fix for issue #6382: Warning when switching modes

a17jonro:
#6382


a17jacsv:
Looking good! Well done.

Issue #6289, opened 2019-04-25
Arrow keys are not functional in MoveAround mode


a16tobfr:
When pressing the spacebar key, you should be able to move around by pressing the arrow key (up, down, left, right). This is not functioning right now.

a17jacsv:
I agree. It's a good addition!

PR #6443, opened 2019-04-30
Grupp a2019 w18#6289

a17okaoz:
Issue #6289 fixed. Can move around with arrowkeys in "MoveAround" mode.


a17jacsv:
Works as expected! Well done.

Issue #5627, opened 2019-03-28
Self assessment tool group ranking and sorting


HGustavs:
Currently the self assessment tool has course-wide rankings, it would be interesting to also add group-rankings.... If a developer performs well within the group but not necessarily globally this should be visible.

So in short adding group to the sorting in the table....

HGustavs:
Also are we using sortabletable? 

if not....

a17poner:
@HGustavs Im currently unable to access contribution in the development environment, heard there is something about a database missing cause currently the contribution.php page is nothing but blank when I access it on my local or the dev webserver.

a17adajo:
A test database will be provided soon by henrik

a17poner:
merged

PR #6444, opened 2019-04-30
Self assessment tool group ranking 5627

a17poner:
Adding the feature requested in #5627. In contribution you should now see a tab at the top where the rankings are called "grouprank". The grouprank should be "NOT FOUND" if the user does not have a group in the course (user_course) for the ccorrect version. each group the user has is saved and then each student who belong to the same group(s) are pulled into an array and a the user gets ranked among them. The color function for the cell currently has the same call as the normal one, however it is hardcoded as a random number in it, creating an issue about that specificly so ignore that part ;)


a17poner:
Forgot to close the pr before I rebased the branch, don't review or merge this yet, need to fix it back first :)

a17poner:
Should now have the correct base and the merge conflicts resolved.

a17poner:
Yeah you're right, when I redid the pr I missed adding the $j = 1 on row 170 and im not sure if I ever had the author --> blameuser correct. Think i fixed both of them now.

Issue #6314, opened 2019-04-26
When scrolling horizontal in resulted. search tools, filters etc. doesn't follow


c17johha:
![image](https://user-images.githubusercontent.com/49199825/56791688-f53d3600-6807-11e9-9b9f-ca17ecad02ed.png)

is it efficient to have the components "Delkurs" "Betygsskala" Betygsdatum" etc. to follow when scrolling the view horizontal?

c17julka:
#6253 talks about how the search bar should become sticky (currently in progress)

c17johha:
![image](https://user-images.githubusercontent.com/49199825/56883035-c3c1a600-6a65-11e9-861b-58dac2e7cbb2.png)
@c17julka Does #6253 cover more than this search bar?

PR #6445, opened 2019-04-30
Made filter options sticky horizontally

c17julka:
#6314 

Test site: https://c17julka.gruppb.webug.his.se:20002/DuggaSys/resulted.php?cid=2&coursevers=97732


a16hiwah:
test is ok! It definitely looks better

Issue #5629, opened 2019-03-29
Integration of time-sheet data with progress self-assessment tool.


HGustavs:
After having made the time sheet tool ...

I would like to integrate that data with the progress/self assessment module

HGustavs:
Currently the time sheet is handled by feedback_dugga in the templates folder.

HGustavs:
Passive view in sortable table for each time sheet in course.

It is better that teachers view progress here than in each time sheet dugga.

a17poner:
Merged now, closing

PR #6448, opened 2019-04-30
Contribution: Added sortableTable with integration of time sheet data

a17joaap:
Issue #5629 

The table is in the bottom of the contribution dugga. It will show all time sheet entries of the user in that course and verision.

Sorting is messed up on week and reference, but that's an issue for the group that covers sortableTable


a17alevu:
I get Errors when i enter the contribution page. 

image

Issue #6439, opened 2019-04-30
Commented out code in sectioned.js


a17alian:
Line 777-816 is commented out, last edited in commit a483625. Should be safe to remove

PR #6451, opened 2019-04-30
deleted commented out code

c17alepe:
There shouldnt be any trouble, but heres a test site just in case

#6439 

https://c17alepe.webug.his.se:20002/DuggaSys/sectioned.php?courseid=2&coursevers=97732

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -774,46 +774,7 @@ function returnedSection(data) {
               }
             }
           }
-          /*
-          // Alt 2
-          var grp="UNK";
-          let grpmembershp=data['grpmembershp'];
-          let grptype=item['grptype']+"_";
-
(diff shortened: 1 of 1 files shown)


a17pongo:
@c17alepe did you check why this code was commented and if it potentially could be used anywhere?
this is important to check before removing code even though it was commented

c17alepe:
@a17pongo It seems like it was alternative code for a solution and has a version of it working directly above it

a17pongo:
looks like a + sign was added accidentally in the commit

c17alepe:
@a17pongo its githubs + sign for the row added, just like the red ones with - signs. the row is empty for visual reasons. weird place for github to place the + and - signs either way.

a17amama:
Checked the code and it looks fine. The code that was commented out was, as already stated, just an alternative  for the already functioning code, removing it does not cause any issues and it is approved for merge.

PR #6273, opened 2019-04-25
Fix for issue #6166: Make variables more readable by adding values ​​to attributes

a17jonro:
#6166 
I believe that i've changed the most important variables, however if there are more that should be added feel free to request them.

PR #6454, opened 2019-04-30
Fix for issue #6166: Make variables more readable by adding values ​​to attributes

a17jonro:
#6166 
The previous solution somehow removed when revert was performed and a new pull request was opened. This should have the same solution as the 2 previous pull requests. #6273 that wasn't implemented before and #6307 that was implemented.


a17jacsv:
Approved!

Issue #6440, opened 2019-04-30
Unintentional move of entities


a17drama:
If the users intention is to right click and drag to move the camera and the user accidentally clicks on an entity two things happen. Camera is moved but so is the object. Right clicking and raging should not move entities.

a17adajo:
I dont know, maybe "Right clicking and **raging**" should move the entities ;)

a17drama:
Maybe we could do this without rage and change the word to dragging. :D "Right clicking and dragging".

PR #6462, opened 2019-04-30
Added canvasRightClick

a17alfcl:
Fix for issue #6440 

Right click does not move objects anymore


a17jacsv:
Looking good. Approved!

PR #6463, opened 2019-04-30
Grupp b2019 w18#6401

a17oscno:
Removed the system changeing peoples ssn if the year was not specified in the "ladexport" in the result view.

### DIFF
--- a/DuggaSys/resulted.js
+++ b/DuggaSys/resulted.js
@@ -1162,7 +1162,7 @@ function exportCell(format, cell, colname) {
   var ssnYearDigits = parseInt(cell.ssn, 10);
 
   console.log(currentYearDigits);
-  console.log(ssnYearDigits.toString());
+  console.log("substringed: "+ssnYearDigits.toString().substring(0,2));
 
   // if lenght = 10 and cell.ssn last 2 ex 98 < current year last 2 digits then str = first 2 digits of current decenium
   // else str = 2 first digits of decenium
--- a/DuggaSys/resulted.js
+++ b/DuggaSys/resulted.js
@@ -1167,7 +1167,7 @@ function exportCell(format, cell, colname) {
 		if (colname == "FnameLnameSSN") {
 			if (cell.ssn.length > 11) {
 				str = cell.ssn + ";";
-      } else if(cell.ssn.length == 11 && currentYearDigits > ssnYearDigits){  // If the person is born on a date after the current it must be last century.
+      }else if(cell.ssn.length == 11 && currentYearDigits > ssnYearDigits){  // If the person is born on a date after the current it must be last century.
 				str = todaysDate.getFullYear().toString().substr(0,2) + cell.ssn + ";";
       }else{
         str = (todaysDate.getFullYear() -100).toString().substr(0,2) + cell.ssn + ";";
--- a/DuggaSys/resulted.js
+++ b/DuggaSys/resulted.js
@@ -1167,6 +1167,7 @@ function exportCell(format, cell, colname) {
 		if (colname == "FnameLnameSSN") {
 			if (cell.ssn.length > 0) {
 				str = cell.ssn + ";";
+      }
       // }else if(cell.ssn.length == 11 && currentYearDigits > ssnYearDigits){  // If the person is born on a date after the current it must be last century.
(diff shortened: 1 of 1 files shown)


a17oscno:
Did not try the code first.

a17oscno:
Have now tested it myself and seems to work, gonna reopen again.

https://a17oscno.gruppb.webug.his.se:20002/DuggaSys/resulted.php?cid=2&coursevers=97732

a17oscno:
@b17timjo Something like this? still need a quote from @HGustavs  to know what's correct.

DuggaSys/resulted.js
```diff
@@ -1156,16 +1156,15 @@ function onToggleFilter(colId) {
...
 	if (format === "csv") {
 		if (colname == "FnameLnameSSN") {
-			if (cell.ssn.length > 11) {
+			if (cell.ssn.length > 0) {
 				str = cell.ssn + ";";
-			} else {
-				str = "19" + cell.ssn + ";";
-			}
+      }else{
+        str = "XXXXXXXX-XXXX";
```
b17timjo: The real crux is that the `else` does not output a semicolon:
```suggestion
        str = "XXXXXXXX-XXXX;";
```
If that else ever ran, the csv would not have a terminating ";" for that column, and there would be a column mismatch.
Since it would take a missing ssn for this to occur, "this should never happen", but better safe than sorry I say.

DuggaSys/resulted.js
```diff
@@ -1156,16 +1156,15 @@ function onToggleFilter(colId) {
...
 	str = "";
 	if (format === "csv") {
 		if (colname == "FnameLnameSSN") {
-			if (cell.ssn.length > 11) {
+			if (cell.ssn.length > 0) {
 				str = cell.ssn + ";";
-			} else {
-				str = "19" + cell.ssn + ";";
-			}
+      }else{
```
b17timjo:
```suggestion
			}else{
```

DuggaSys/resulted.js
```diff
@@ -1156,16 +1156,15 @@ function onToggleFilter(colId) {
...
+        str = "XXXXXXXX-XXXX;";
```
b17timjo:
```suggestion
				str = "XXXXXXXX-XXXX;";
```

DuggaSys/resulted.js
```diff
@@ -1156,16 +1156,15 @@ function onToggleFilter(colId) {
...
+      }
```
b17timjo:
```suggestion
			}
```

DuggaSys/resulted.js
a17oscno: ofcourse

a17pongo:
19 is needed in external resources, because of a fix ladoc did you now need to have full length on ssn to make sure people born 2000. full ssn is needed because otherwise it is not possible to export.
"YYxxxxxx-xxxx" so ssn should follow this structure

b17timjo:
In that case, I think this merge should be rejected.
The problem persists and requires a slightly more complex solution, as students born in 2000 are now 18-19 and may exist in the system (or may soon).

a17oscno:
Depending on how advanced the fixed needs to be and how it needs to work to me legal this solution might or might not be but i made a short test yesterday with a function that would calculate (or atleast try to) the birth-century of a person if that was not included in the ssn provided. It  probably needs some fixing but it might be a start. Will paste the code here for reference. 

var todaysDate = new Date();
var currentYearDigits = todaysDate.getFullYear().toString().substr(2);  // returns 19 if e.g year is 2019
var ssnInt = parseInt(cell.ssn, 10);
var ssnYearDigits = ssnInt.toString().substring(0,2)  // Same as currentYearDigits but for the ssn

if (cell.ssn.length > 0) {
   str = cell.ssn + ";";
}else if(cell.ssn.length == 11 && currentYearDigits > ssnYearDigits){
	str = todaysDate.getFullYear().toString().substr(0,2) + cell.ssn + ";";
}else{
     str = (todaysDate.getFullYear() -100).toString().substr(0,2) + cell.ssn + ";";
}

As i said i don't know if this is the correct solution (and it can certainly be inproved) but it might be a start.

a17pongo:
I suggest closing this pull request and referencing this from the issue so the conversation is not lost, until we find and agree on a proper solution.

Issue #6449, opened 2019-04-30
Remove background color of chosen diagram


a17jacsv:
Remove the background color of "UML" & "ER" as it can be expected to be a button...

<img width="160" alt="Skärmavbild 2019-04-30 kl  10 39 32" src="https://user-images.githubusercontent.com/37792795/56949854-4f096d00-6b34-11e9-9c67-1e5d4f868d52.png">

PR #6464, opened 2019-04-30
Changed text styling for mode - Fix for #6449

a14oscna:
#6449 
Changed the styling so the text no longer looks like a button and also clarifies that it shows the selected mode.


a17jacsv:
Approved!

Issue #6469, opened 2019-04-30
Grid not working properly on even zoom values


a17alfcl:
Some of the math does not work

PR #6470, opened 2019-04-30
Rouning error solved

a17alfcl:
Fix for issue #6469 

Javascript is still bad at math


a14oscna:
I need a better description of the problem before I can make a proper review of the fix. What math problems existed before?

a17jacsv:
OK.

Issue #5613, opened 2018-11-19
Not yet exported selection in marking view


HGustavs:
Once a result has been exported in marking view we should be able to "tick that result off" if it was a pass or better result.

Currently all results are shown even if they have already been exported.

We want a way to only show (and therefore only export) previously not exported result items.

b17timjo:
@HGustavs what is the expected behaviour? Is the intent that when the export button has been pressed, only changed results should show up if export is pressed again (for that session? or adding a persistent marker to the database if a given result has been exported?)
And/or some function to manually toggle if a given individual should be present in the export window?

HGustavs:
e.g. a last exported date for each item.

Perhaps this is very hard, but I would imagine a service function that receives a number of identifying information for each result that was exported.

HGustavs:
Searching for results that have not yet been exported is the general goal.

Unclear if this is relatively easy or extremely complicated.

If complicated, I say back-burner.... far back on the stove.

HGustavs:
We want to see the results that have come in since last export.

b17timjo:
As the export button only displays a csv-text with the information, there is no way of actually verifying if something has been exported. The closest would be putting the responsibility on the user by adding a button to that window along the lines of "Mark all as exported"

a17alian:
accidentally closed issue

b17timjo:
The current state of this issue is:
* the "answers" seen in resulted have a field in the database (userAnswer.gradeLastExported) which is intended to hold the last time the answer was exported via resulted
* there is a service "getunexported" in resultedservice.php which returns all answers which have not been exported, graded, or were updated since they were last exported (this may need to be updated to provide the required functionality)

Todo:
* userAnswer.gradeLastExported is never updated
* resulted does not update anything in the database when the "Lad Export" button is clicked. Suggestion: add a "copy to clipboard and mark as exported"-button in the popup
* the view shown when exporting should use the service to filter for only un-exported answers

a17alian:
The sub issue #7289 made it possible to update gradeLastExported in the database. But the Last Exported date is updating all entries with the same date. What needs to be implemented in the future is a new SQL query where the date is different dependning which "delkurs" is selected in the drop down. And the last exported date is updated only for the selected delkurs.

b18fella:
Some of the ToDo:s in this issue has been solved in #9014. The ones that are solved are:

- userAnswer.gradeLastExported is never updated

[...]

PR #6472, opened 2019-04-30
Implements and closes #6466 - a sub-issue to #5613

### LINKED ISSUE #6466: Add a service required for #5613 - Get unexported grades in resulted
As a sub-issue to #5613, this will add necessary database columns and implement a new service in resultedservice

b17timjo:
To test these changes, a database update is required. See the committed delta-file.

This will close #6466 (apparently the auto-close-reference does not work in the title?)

### DIFF
--- a/DuggaSys/resultedconstants.php
+++ b/DuggaSys/resultedconstants.php
@@ -1,0 +1,8 @@
+<?php
+namespace resulted;
+
+class Constants
+{
+	// POST parameter identifier for a service retrieveing unexported grades
+	const getunexported_service_name = "getunexported";
+}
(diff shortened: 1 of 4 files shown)


a17pongo:
@b17timjo so the test should make sure that the changes did not break anything? and is there a way to test this new service in a good way? it's a little unclear on what these changes did or mean.

b17timjo:
First and foremost, yes; it should not break anything.
The added service can also be tested manually by sending a POST request to resultedservice with opt=getunexported(and with an active session from a user with w- or superuser permissions)

b17timjo:
As for the purpose, see parent issue #5613, which will require quite a few changes - something like this service being one of those changes

PR #6477, opened 2019-04-30
Grupp b2019 w18#6398

### LINKED ISSUE #6398: Email vulnerability allows an attacker to retrieve all emails associated with a given course
The following code runs without checking if the user has a valid login:
https://github.com/HGustavs/LenaSYS/blob/0b3972bc464df6240a4088ec74e41fcd70bad63f/DuggaSys/resultedservice.php#L72-L104

This allows a malicious actor to retrieve all emails for a course (students and teachers alike) using only publicly known information, by simply sending a POST request with the correct parameters.
This may be a breach of GDPR, as email addresses may be used to identify individuals.

This is mitigated by the ability to block direct access to this page via webserver configuration, but that should not be relied upon.

a17ludca:
Fixes the issue of being able to access emailbutton while not logged in or not logged in as superuser.
closes #6398
test site: https://a17ludca.gruppb.webug.his.se:20002/DuggaSys/resulted.php?cid=2&coursevers=97732

### DIFF
--- a/DuggaSys/resulted.php
+++ b/DuggaSys/resulted.php
@@ -68,7 +68,13 @@ pdoConnect();
         <input id="laddate" type="date" style="font-size:12px;">
         </div>
       <button onclick="ladexport();" style="margin:5px;">LadExport</button>
-			<button onclick="mail();" style="margin:5px;">Mail students</button>
+			<?php if (checklogin()){ ?>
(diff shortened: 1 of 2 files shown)


a17ludca:
No information is gained by doing that though.

c17julka:
That is true. Then everything seems ok!

Issue #6377, opened 2019-04-29
Move cursor displayed when it shouldn't be


a17drama:
Move cursor is displayed when hovering a line and lines shouldn't be able to move. They are "static" between entities. It also displays when a draw tool for line is chosen and this is very confusing. Is the user drawing a line or moving the entity?

a16jenja:
![image](https://user-images.githubusercontent.com/49199287/56888264-da6ef980-6a73-11e9-9ed0-6e502fc73b5b.png)
I might be misunderstanding the issue, but when I hover over a line it does not show a move cursor. And I can't move the lines either. @a17drama 

Although when the draw tool for line is chosen it does show a move cursor, so that I could try and fix.

a17drama:
Might have been an issue that got resolved by something else. First time I tested I got a move cursor hovering the line. Focus the line drawing then and check if there is something else being wrong with the move cursor. Remember the UML line.

a16jenja:
You are right, I forgot to test the UML line, there is a move cursor and you can move that one. So I'll get right on that.

a17drama:
A part of this issue is written as a separate issue #6507 . When this is done tag both issues in the pull request so both get closed.

PR #6480, opened 2019-04-30
Fixed issue #6377: Move cursor was displayed when it should not be

a16jenja:
Issue: #6377 

Fixed the issue when hovering over a umlline and when trying to draw lines between entities it would show a move cursor. Now it shows the default cursor when these actions are being done.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -747,7 +747,8 @@ window.addEventListener("mousemove", function()
     let indexOfHoveredObject = diagram.indexOf(diagram.checkForHover(currentMouseCoordinateX, currentMouseCoordinateY));
     if (indexOfHoveredObject != -1) {
             for (let i = 0; i < diagram.length; i++) {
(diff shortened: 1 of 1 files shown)


a16jenja:
I noticed this too, and I think you are right, I'll get on it.

a16jenja:
I noticed that you can still move the lines between uml entities, but the cursor is now default. But changing the lines to be static might be a separate issue from this one in my opinion.

a17jacsv:
> I noticed that you can still move the lines between uml entities, but the cursor is now default. But changing the lines to be static might be a separate issue from this one in my opinion.

I agree. Please create an issue of these problems and we'll merge this PS for now.

a17jonro:
The conflict wasn't resolved correctly so the wrong changes were implemented and then merged, this needs to be fixed.

a16jenja:
Agreed, I created a new issue regarding this since I did not know what to do since its been merged already.

Issue #6407, opened 2019-04-29
Zoom-buttons hover effect should not be able to be hovered


a17vikab:
Zoom in / out should not be able to be hovered and should disappear when the pointer leaves the button.

![image](https://user-images.githubusercontent.com/37792464/56895458-79eab700-6a89-11e9-98a0-c2b939d6e427.png)

PR #6481, opened 2019-04-30
Fixed issue #6407: Zoom-buttons hover effect should not be able to be hovered

a16jenja:
Issue: #6407

When the mouse cursor is not hovering over the zoom-buttons, the tooltip should disappear correctly now. And it's no longer possible to hover over the tooltip text the zoom-buttons hover effect generate.


a17jacsv:
Approved!

Issue #6386, opened 2019-04-29
Can't access username on duggaed.php on mobile


c17julka:
![image](https://user-images.githubusercontent.com/37792321/56885328-7e54a700-6a6c-11e9-9ab6-72508acd0f14.png)

Can't scroll to the right on mobile unlike the other table .php files, so you can't access your username.

PR #6482, opened 2019-04-30
Grupp b2019 w18#6386

a17adasy:
fixes issue #6386 ,

added styling in media queries so the searchbar input will be smaller.


a16hiwah:
testning is ok , code follows standard and works fine. it looks better.

PR #6483, opened 2019-05-01
Moblie friendly form added to Daily-minutes

a17fellu:
Made a new form for smaller devices. If the window width is bigger than 1120px the old form will show. If the window width is smaller than 1121px the new mobile friendly form will show and the other form will be hidden.

**Normal form:**
normal form

**Mobile friendly form:**
mobile friendly form


a17fellu:
This currently interfere with the normal form. Will fix it

a17fellu:
The mobile form doesn't interfere with the normal form anymore

PR #6488, opened 2019-05-02
Grupp b2019 w18#6223

a17oscno:
Fixed the problem with the missing icon when creating a message, due to not being able to find which old icon was used i decided to use the warning triangle to make a clear signal that this was important (the message not the issue).


a17oscno:
https://a17oscno.gruppb.webug.his.se:20002/DuggaSys/sectioned.php?courseid=333&coursename=Motorblocksgjutning%20med%20Kent&coursevers=1

use this for testing

a16hiwah:
The testning has been made, code follows standard and works fine. it looks better now.

PR #6532, opened 2019-05-02
Create the database in its final form directly

### LINKED ISSUE #5693: The install script should create the database in its final form without the use of ALTER statements.

a17adajo:
Fixes #5693


a17poner:
Looks alright, but before I merge in deltachange.sql what are(were) the lines 62 --> 100 doing. Looks just like someone put the printouts from executing a bit of sql into the sql code which then makes sense to have it removed altough it seems very odd for it to have ever gotten there.

a17adajo:
Yeah it is VERY weird that the file thats supposed to handle the migration of the tables had lots of output from the SQL command line. Thats why i removed it.

Issue #6507, opened 2019-05-02
Move cursor is showing when trying to draw UML lines


a16jenja:
![image](https://user-images.githubusercontent.com/49199287/57066459-8448c300-6ccc-11e9-8758-36690f236e56.png)

The issue #6377 was supposed to adress this problem. I was assigned to that issue, but I only fixed the problem when drawing ER lines. I forgot about the UML lines. My bad.

a17drama:
This issue is covered by #6377 . Marking as redundant and tagging this issue in the older one.

a17drama:
This issue is covered in #6377 . Tagging this one in the old one and marking it redundant.

PR #6539, opened 2019-05-02
Fixed issue #6377 and #6507: Move cursor displayed when it should not be

a16jenja:
Issue: #6377 and #6507

Fixed issue regarding wrong mouse cursor when trying to draw UML lines and when hovering over the points at the edge of the lines. Before it was a move cursor, now it should be a default cursor.


a14oscna:
Looks good!

Issue #6524, opened 2019-05-02
Increase spacing between error-texts.


a17robda:
![Screenshot_2019-05-02_13-26-43](https://user-images.githubusercontent.com/37792328/57072511-45bc0400-6cde-11e9-89d2-bb663994def6.png)
Right now the bold "Bummer!"-text seems too close to the rest of the error-message. Maybe add a few pixels between these?

a17poner:
Has this been fixed? And if so in what pr?

a16lisri:
Forgot to close this. This was fixed in https://github.com/HGustavs/LenaSYS/pull/6540

PR #6540, opened 2019-05-02
Fix for #6524

a16lisri:
Added a space between error-texts.


a17poner:
sectioned is grupp B responsability, we could merge it as well and then have it in the big merge or if tehy want to take it to their branch first, up to @b17nicsv

a17poner:
Talked with @b17nicsv and we will merge this into groupC but avoid sectioned in the future :)

Issue #6494, opened 2019-05-02
Infinite side scroll on resulted


a17adasy:
If the ladexport container is bigger than the viewport of the screen (if the screen is <720px) you can scroll right infinitely.

On the resulted page of dugga.

PR #6546, opened 2019-05-02
Fix for infinite scroll

a17adasy:
Look at issue #6494.

Will now stop scrolling, also changed left to transform: translateX() since that property is supposed to be better for animating. Which I think resulted in smoother animation for the ladExportContainer when scrolling horizontally.

for testing: https://a17adasy.gruppb.webug.his.se:20002/DuggaSys/resulted.php?cid=2&coursevers=97732


a16hiwah:
test is ok , and works fine.

Issue #5750, opened 2019-04-09
Prioritization of object placement (front/back)


a16tobfr:
<img width="205" alt="toFront" src="https://user-images.githubusercontent.com/49199690/55797584-1a723a80-5acd-11e9-81bb-45f5e422d1d3.PNG">
It is not possible to decide which object is in the front of the canvas space and which object that gets stacked in the back of the canvas space. This functionality might be useful in the future.
<img width="205" alt="toFront" src="https://user-images.githubusercontent.com/49199690/55798087-5954c000-5ace-11e9-9bb5-054b4f3a9e74.jpg")>
To further clarify the problem, the Z-axis in the image above is the vector in which objects should be able to be moved.

HGustavs:
It should draw back to front depending on the list of objects, it is a drawing order thing.

If we implement "move to front" this should mean put last in object list.

If we implement "move to back" this should mean put first in object list.

HGustavs:
Objects are commonly drawn from first to last, and then lines after the objects.

PR #6547, opened 2019-05-02
Can now toggle front or back, Fix for issue #5750

b17robev:
Fix for issue #5750 

Free draw objects are always drawn in the background but placement in list between symbol is now possible and also between free draw objects.

Toggle front with key 1 and back with key 2.


a17alfcl:
Added constants and dropdown options in a new branch so I'm closing this one

PR #6550, opened 2019-05-02
Fix #6495 global button styling

### LINKED ISSUE #6495: All the buttons should have the same styling
The buttons do not have the same styling overall. It should be a standard with small variations depending on the situation.

a17vikab:
Fixed so every button behave the same, eg: hover-effect and styling in change-mode-dialog

Issue: #6495


a17jacsv:
Looking good. Well Done! (Can’t approve on phone)

Issue #6218, opened 2019-04-24
Unable to change Course Version ID 


a17maxsj:
reverts back to the former value after save, see image...

![image](https://user-images.githubusercontent.com/49199993/56657198-b8eac800-6697-11e9-880c-610fbd3859d6.png)

a17joaap:
Version ID is a primary key for the 'vers' table, what this dialog does is change the version name, start date, end date, and defaultness of the version with the version ID you put in. So this is working as intended, you can't change the version ID.

HGustavs:
We could even remove the version as it is immutable imho.

PR #6559, opened 2019-05-02
gruppB2019W19#6218

b17nicsv:
[Issue](https://github.com/HGustavs/LenaSYS/issues/6218)

The greyed out color might be something to change. If so, please respond with a suitable hexcode for the color.


c17julka:
Having the standard background colour ("var(--color-background)") as the background colour for the "lo" would make it more consistent with the table and also brings more contrast.

Otherwise it looks really nice

b17timjo:
If you have change requests please make a formal request using the tools, or if you approve please set it as such -- the automation depends on these.

Issue #6139, opened 2019-04-23
CSS use of !important


b17timjo:
A number of CSS rules use `!important.` When possible, specificity should be used instead.

HGustavs:
one of many problems in css.

...

b17nicsv:
Closing this because from what I understand this issue is a large refactorisation and this isn't something which is allowed (?) in this project.

PR #6561, opened 2019-05-02
Using CSS Specificity instead !important.

a16hiwah:
like for test : https://a16hiwah.webug.his.se:20002/DuggaSys/resulted.php?cid=2&coursevers=97732

i removed all !important  and used CSS Specificity instead. all changes has been on style.css


a17anter:
#6139

Issue #2463, opened 2017-03-18
Group assignment dugga template


a97marbr:
I would like a group assignment dugga template.

The major features of this is that we should have the ability to 
* assign groups from the course rooster
* marking one submission in the group should automatically grade all group members' submissions

a15alica:
I discussed this with @HGustavs and we came to the conclusion that the group assignment duggas should work as if it was just one person. The group is seen as one person but all students in the group can hand things in, and see what has been handed in (both by themselves and the other members of the group).

When marking a submission the grades are saved for all members of the group.

d15marwi:
didnt have time to implement this but it should be easy to solve with the new group structure

PR #6562, opened 2019-05-03
Group assignment dugga template

a17joaap:
For issue https://github.com/HGustavs/LenaSYS/issues/2463

Work in progress


a17joaap:
The issue was when no files where uploaded to the dugga, there is nothing to sort. Fixed.

a17joaap:
That is because of incorrect permissions on your server directory, not caused by the code.

Issue #6490, opened 2019-05-02
Developer mode does not work anymore.


a17vikab:
![image](https://user-images.githubusercontent.com/37792464/57063233-d7b61380-6cc2-11e9-88d5-e994ecee43df.png)

PR #6571, opened 2019-05-03
Fixed developer mode logic and bug, Fix for issue #6490 

b17robev:
Fix for issue #6490 

Toolbar state had a bug in it which broke the developer mode in it.


a14oscna:
Confirmed!

Issue #6509, opened 2019-05-02
Incorrect styling for search bar


c17julka:
![image](https://user-images.githubusercontent.com/37792321/57066502-a17d9180-6ccc-11e9-82dd-127599d931d6.png)
![image](https://user-images.githubusercontent.com/37792321/57066539-bce89c80-6ccc-11e9-9c90-2e94cf3f6406.png)
Above is how it's supposed to be like.

![image](https://user-images.githubusercontent.com/37792321/57066574-d0940300-6ccc-11e9-89b8-e2e287542ca5.png)
![image](https://user-images.githubusercontent.com/37792321/57066589-dbe72e80-6ccc-11e9-9053-dad41b969339.png)
Currently is like this instead. 

Should be enough to copy the styling from the first picture.

PR #6572, opened 2019-05-03
Grupp b2019 w19#6509

a17oscno:
For issue #6509 

Fixed the styling of the searchbar.

Works in all the sites exept Resulted due to another bug for that perticular site.

https://a17oscno.gruppb.webug.his.se:20002/DuggaSys/courseed.php

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -1021,7 +1021,7 @@ header td {
 
 input {
   border-radius: 2px;
-  border: 2px solid var(--color-border-2);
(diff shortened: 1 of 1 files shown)


b17nicsv:
That would make it more accessible for change later down the line. I like this change, but then I would like to see every border color using that color to be changed to use the variable.

b17nicsv:
The change above isn't necessary and to keep things going we will merge this pull request and might add an issue for this later.

Issue #6534, opened 2019-05-02
Every other cell in resulted is transparent


c17julka:
A line of code has been removed from dugga.css which makes even numbered cells in the resulted table have a white background. They are now transparent after the removal of this line.

`table tbody tr:nth-child(even) {background-color:#F0F0F0;}` is the code missing

PR #6584, opened 2019-05-03
Grupp b2019 w19#6534

a17oscno:
pull request for issue #6534 

Fixed the colors and which in turn removed the transparancy on the resulted page when you would move the the right (or left) with the row headings (where the name, ssn and class is).

link: https://a17oscno.gruppb.webug.his.se:20002/DuggaSys/resulted.php?cid=2&coursevers=97732


b17nicsv:
Looks great and solves the issue. Code looks good.

Issue #6527, opened 2019-05-02
resulted - searchbar disappeared and sort icon was moved to another location


a17pongo:
searchbar in resulted disappeared after the merge, probably because in setup-function the wrong code was commented and probably can be un-commented.

PR #6595, opened 2019-05-03
Searchbar, sorting has been moved. 

a17amama:
Re-implemented code from resulted.js and commented out code inside navheader.php that would collide with resulted.js and create issues.


a17amama:
issue #6527

a17ludca:
This seems to work as intended, I consider this solved.

Issue #6587, opened 2019-05-03
Commented code should be reimplemented


c17julka:
In the setup() function in resulted.js, there is commented code which should be reimplemented. Instead, lines 73 - 86 should not be there.

This would probably fix a lot of issues at once: #6527, #6516, #6520 and #6565.

PR #6596, opened 2019-05-03
Reinserted code for search bar and position for filter. #6587 Update resulted.js

c17johha:
#6587


a17amama:
Broken, there is still a sorting button to the right which is also bugged.

b17nicsv:
This issue was fixed by another pull request. Closing this one.

Issue #6457, opened 2019-04-30
VOTING!!! -  Where should the toolbar be? Vote in comments!


a17alfcl:
Leave a print screen of the window with the toolbar at the position that it could be initiated to or statically set to. Thumbs up at the option that you vote for if it already exists

a17alfcl:
My suggestion is to make the buttons static at the top.
![toolbar](https://user-images.githubusercontent.com/49202727/56956803-7ae11e80-6b45-11e9-9677-34eeccc7a84e.png)

a17jacsv:
I do support this addition. This would be a more user-friendly and a smoother view in my opinion. What do you think @HGustavs?

a17drama:
Due to our democratic society this voting is over. Winning suggestion is the one by @a17alfcl with 10 votes out of 14 total. This suggestion is accepted by the council. Thank you for your votes.

a17jacsv:
You're free to go ahead with this solution @a16tobfr.

a16tobfr:
After having had a democratic discussion, the toolbar will instead be moved to the left and be displayed in a vertical manner. This means that the canvas will also have to be moved a bit to make room for the toolbar.
![toolbar](https://user-images.githubusercontent.com/49199690/57129708-8af03c80-6d97-11e9-9dd6-fc076e2800c4.JPG)

a14oscna:
Agreed with the comment above, further inspiration can be taken from the following:
https://www.photoshopessentials.com/basics/photoshop-tools-toolbar-overview/

a17vikab:
Issue solved, closing issue.

PR #6598, opened 2019-05-03
Grupp a2019 w19#6457

a16tobfr:
#6457 
The toolbar is now moved to the left and seems to work as intended.
Moved canvas to the right to make room for the toolbar
Changed size of the toolbar
Removed the minimize toolbar functionality
Removed the align toolbar horizontal functionality
Removed unnecessary functions

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -997,7 +997,7 @@ function initializeCanvas() {
     setInterval(hashCurrent, hashUpdateTimer);
     setInterval(hashFunction, hashUpdateTimer + 500);
     setInterval(function() {Save()}, 10000);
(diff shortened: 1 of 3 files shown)


a16tobfr:
fixed

a17johpe:
There are some small design fixes that can be added to a separate issue. The base fix looks good and nothing seems broken on the webpage or in the code

a17johpe:
Design fixes:
- undo and redo buttons preferably on same row
- Tooltips on hover over button should be moved slightly as it currently covers the undo/redo buttons slightly (see image)
toolbarposition
- On Firefox, the header for UML mode is clashing slightly with the next header (see image)
toolbardialog

a14oscna:
Remember to create new issues. I will merge this now.

PR #6599, opened 2019-05-03
Fixed move to front / move to back, Fix for issue #5750

a17alfcl:
Fix for issue #5750

Previous pull request:
https://github.com/HGustavs/LenaSYS/pull/6547

Added constants and dropdown options + hotkeys

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -148,6 +148,7 @@ const upArrow = 38;
 const rightArrow = 39;
 const downArrow = 40;
 const ctrlKey = 17;
+const shiftKey = 16;
 const windowsKey = 91;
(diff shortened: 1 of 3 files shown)


a17alfcl:
> diagram.js
> line 395 and 547. What is the point of having same thing checked twice?
> line 428. Escape key default function in browser is to stop loading the web page. We should avoid hotkey conflicts and overwrites.
> line 468 and 468. What is the point of creating this variable instead of using "diagram.length"?
> 
> This doesn't support free draw entities as it is now. Should it?

The double check is removed, escapeKey is only a constant and the functionality was implemented before. The diagramLength variable is necessary because the length of the diagram will decrease so the math would be wrong.

Issue #6578, opened 2019-05-03
Move align functionality to distribute


b17robev:
Move parts of align so that it does not create spacing between objects

a17drama:
@a17vikab This has pretty strong connection to #6188 . They are pretty much the same and should be marked as redundant.

PR #6601, opened 2019-05-03
FIX: Move align functionality to distribute #6578

a17jacsv:
We have fixed the distribute-functionality to work as it was expected. Also, the align-functionality works as @HGustavs wanted to be.

_(Me and @a17drama has worked with this issue)
#6578_


a14oscna:
I could not find any issues either! Good job.

PR #6603, opened 2019-05-03
Fix #6017 ER composite behaviour 

### LINKED ISSUE #6017: ER composite behaviour
https://ion.uwinnipeg.ca/~rmcfadye/2914/hypergraph/composite.html
"An attribute is considered composite if it comprises two or more other attributes"

As of now. It is possible to set attributes to composite even if they do not comprise two or more attributes.

It would be better if an attribute that comprises two or more attributes automatically changed to a composite attribute without the need to change it to a composite.

Optionally, when composite is chosen, a notification can be shown (if it is not a valid composite) to the user. The notification could say that a composite should comprise two or more attributes, and that the attribute was set to normal because of this.

a17okaoz:
Issue #6017 fixed. When trying to make a line between two ER attributes an error dialog that none of the objects are Composite.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -2900,11 +2900,6 @@ function mouseupevt(ev) {
          }else {
               //Get which kind of symbol mouseupevt execute on
              symbolEndKind = diagram[hovobj].symbolkind;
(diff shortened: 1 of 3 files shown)


a14oscna:
I have a minor request before I merge this fix, the cancel button has 0 border radius, but when it is hovered there is a border radius added. Could you give it rounded corners at all times please?

Issue #6238, opened 2019-04-24
Filtering and sorting buttons don't do anything in accessed.php


a16lisri:
In accessed.php the buttons for filtering and sorting don't do anything and they also have no text when you hover over them like the home and back button do. (Can test this by going into "Datorgrafik" then "Access").

![086607047d31f405d811ac4ba0c8809d](https://user-images.githubusercontent.com/37792690/56659463-7926df00-669d-11e9-9f12-f598b9aa0fcd.png)

a17poner:
It used to give you a empty white box, can't remember or find the issue that dealt with it but then that fix caused some error logs in chat which where brought up in #6037 and #6052. Which I then made a really bad fix for in #6145. Im not sure what they do in accessed but im assuming they should do the same thing here and do some filtering related stuff. The same functions hovers and hoverc exists in accessed.js except they actually have some code in them so to whoever end up with this issue to make them do whatever they do in accessed (never used them or tested so no clue what they do) that is probably a good place to start 😄

a17poner:
@HGustavs is the sorting arrows still necessary? Feels redundant to have them when you can sort the table by clicking the columns in the same way the arrows are supposed to do.

a17maxsj:
bump @HGustavs

a17poner:
This one is now a week branch behind and there seems to be some problems with the placements of the filtering and sorting buttons, I'll wait for a comment by @HGustavs on to keep the sorting or not and for the placements issue to be worked on before I attempt to rebase the fix for this and perform further actions on it.

a17poner:
Splitting this issue, putting the arrow part in a seperate issue so we can move on and merge the funnel once reviewed without having to wait for a response about the arrows in #6602

PR #6604, opened 2019-05-03
Grupp c2019 w18 6238

a17poner:
Making the funnel work as requested in #6238.
Had to fast forward the branch but looking at code that was changed in the merge it looks like it's going to be fine.


a17poner:
hm I guess it broke when I did the fastforward :/ I did it on the site so I didn't get to test it myself to see if it actually worked outside the code I wrote looking the same way. I'll see if I can find what causes this.

a17poner:
I think i've found the cause for this outside my code, there are some other issues currently dealing with it, I'll fast forward the branch once I've merged them to see if it fixes the problem :)

a17poner:
Actually part of this will be included in the commit and pr to fix the issue with hover so I'll redo this entirely if that doesn't fix it all, closing this for now.

Issue #4193, opened 2018-03-21
Contribution is currently a button in section editor and add a dugga/test instead using an iframe


HGustavs:
Remove contribution button in section editor and add a dugga test   type which loads it as an iframe

a16filol:
Should the iframe be shown if "Dugga/test" is pushed, in the content field?

a16thegu:
@HGustavs, can you take a look and explain.

a16thegu:
After talk with client, not Group 3 responsibility. Will talk with Group 1.

a17joaap:
This can certainly be done, however the database needed to test contribution.php does not seem to be included in the repository.

contributionservice.php:
`$log_db = new PDO('sqlite:../../GHData/GHdata_2018_7.db');`

Without this the file will cause a 500 error. Is this part of another issue @HGustavs ?

a17joaap:
The above solution will link any dugga with 'Link' set to Contribution to an iframe containing contribution.php

a17joaap:
Can be closed @a17andby

a17andby:
Closing

PR #6610, opened 2019-05-04
Changed loading of the contribution iframe to be less naive

a17joaap:
The previous implementation hard coded an option to the 'Link' select box with dugga id 9999, which will cause problems when the amount of duggas reach 9999. 

Previous pull: https://github.com/HGustavs/LenaSYS/pull/5805
Original issue: https://github.com/HGustavs/LenaSYS/issues/4193

This way is better and follows the already established logic of creating duggas. (Creating a test in duggaed with the contribution template)

### DIFF
--- a/DuggaSys/sectionedservice.php
+++ b/DuggaSys/sectionedservice.php
@@ -393,16 +393,6 @@ if($gradesys=="UNK") $gradesys=0;
 			);
 		}
 
-		array_push(
-			$duggor,
-			array(
-				'id' => '9999',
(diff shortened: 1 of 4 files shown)


a17poner:
Before I merge this I have a minor thing on row 42 in contribution.php you ended with "; which isn't needed in html so now there is a "; underneath the table. Otherwise its a nice addition :D

a17joaap:
Oops! Fixed that @a17poner

Issue #6385, opened 2019-04-29
Improve the design of the zoombar


a17jacsv:
Please improve the design of the zoom-bar... Fix all the content to be centered. Please add "border-radius" to the "+" & "-" buttons for a cleaner view.

<img width="304" alt="Skärmavbild 2019-04-29 kl  10 34 19" src="https://user-images.githubusercontent.com/37792795/56884527-5f551580-6a6a-11e9-81f4-7f0abf88f445.png">

a17celal:
I believe that this issue is fixed now?

![Capture10](https://user-images.githubusercontent.com/37792271/56955704-736c4600-6b42-11e9-96f5-976b1eff6bd7.PNG)

a17jacsv:
It's better, but I want it to be exactly centered (Just as much space up as down).

PR #6616, opened 2019-05-06
An improved styling to the zoombar

a17okaoz:
Issue #6385 fixed. The styling of the zoombar is improved. The increase and decrease button are bigger and have the same size. Removed position: absolute and istead use dispay: flex.


a14oscna:
a16tobfr I want you to make sure that the mentioned issue will be created. With that said I will go ahead and merge this solution.

Issue #6619, opened 2019-05-06
Invert login/logout button colors


b17carma:
At the moment it looks like you're logged out when you're logged in:
![image](https://user-images.githubusercontent.com/37834592/57215057-24179100-6fec-11e9-93d4-2f94df572f7a.png)

Using a green color when you're logged in and a red color when logged out would look better in my opinion.

a17alevu:
I dont really agree. I would say that green indicates to login and that red indicates to logout, but maybe i am wrong with this.

a17andby:
Don't agree as well.

b17albbe:
I would say that I agree that the color needs to be changed. In fact I would prefer if the previous login visuals were still used too (the user icon). I don't think anybody is arguing whether or not green indicates login, or if red indicates logout. The problem is how those colors are used in this case. And I think it would work better if those colors were used when you are actually trying to login or logout, when hovering over the button.

b17carma:
I'll close the pull request for this issue since people have different opinions about it 😛

b17annni:
This seems sorted by other issues now. @a17alevu ?

PR #6620, opened 2019-05-06
Invert login/logout button colors

b17carma:
Fix for issue https://github.com/HGustavs/LenaSYS/issues/6619


a17poner:
Is this really a good change? isn't the way the colors were prior to this a lot more intuitive?

PR #6633, opened 2019-05-06
Add check for correct upload_max_filesize during install

### LINKED ISSUE #6215: Add check for correct upload_max_filesize during install
server-requirements.md says that upload_max_filesize should be 128M
Is it possible to verify the current setting during install and notify the user from php?

a17adajo:
Fixes #6215


a17adajo:
Looks like this: 

image

Issue #6607, opened 2019-05-03
Redundant variables [properties array]


a16tobfr:
In diagram_object.js an array called properties was made some weeks ago. In that issue, variables were moved into the array.

In the current version of the software, the old variables still exist outside of the array. They should be removed and the problems generated by this should be fixed.
Properties array:
~~~
this.properties = {
        'symbolColor': '#ffffff',                       // Change background colors on entities.
        'strokeColor': '#000000',                       // Change standard line color.
        'fontColor': '#000000',                         // Change the color of the font.
        'font': 'Arial',                                // Set the standard font.
        'lineWidth': '2',                               // LineWidth preset is 2.
        'textSize': '14',                               // 14 pixels text size is default.
        'sizeOftext': 'Tiny',                           // Used to set size of text.
        'textAlign': 'center',                          // Used to change alignment of free text.
        'shadowColor': 'rgba(0, 0, 0, 0.3',             // The shadow color.
        'shadowBlur': '10',                             // Shadowblur for all objects.
        'shadowOffsetX': '3',                           // The horizontal distance of the shadow for the object.
        'shadowOffsetY': '6',                           // The vertical distance of the shadow for the object.
        'key_type': 'normal'                            // Defult key type for a class.
    };
~~~
Old variables which still exist outside of the array:
~~~
    this.textsize = 14;             // 14 pixels text size is default
    this.symbolColor = '#ffffff';   // change background colors on entities
    this.strokeColor = '#000000';   // change standard line color
    this.font = "Arial";            // set the standard font
    this.lineWidth = 2;
    this.fontColor = '#000000';
    this.key_type = "normal";       // Defult key type for a class.
    this.sizeOftext = "Tiny";       // Used to set size of text.
    this.textAlign = "center";      // Used to change alignment of free text
~~~

HGustavs:
Seems reasonable.

PR #6635, opened 2019-05-06
Fix for issue #6607: Remove redundant property variables

c17matjo:
Fix for #6607. Also fixes some lines that were still accessing properties through the old variables. 

When reviewing, try modifying the properties (through the appearance config window) and see if any errors show up in the console.


c17matjo:
Side note: Turns out that when exporting to SVG format the fillColor would be lost (set to black) on weak ER relation symbols due to incorrectly accessing symbol properties. This PR fixes that.

a14oscna:
Fiddled around a bit but could produce an error. Will merge

PR #6636, opened 2019-05-06
Increase responsivness burger

### LINKED ISSUE #6499: Increase responsivness burger
image

Could be more graceful.

b17annni:
Fixes #6499

### DIFF
--- a/DuggaSys/sectioned.php
+++ b/DuggaSys/sectioned.php
@@ -35,7 +35,7 @@
 	<!-- content START -->
 	<div id="content">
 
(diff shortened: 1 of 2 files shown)


a17poner:
Yeah, I for one prefer it all below since there are some spacing issues when above in the navbar once you go lower resolution (it used to be always below at the beginning of the project) but regardless this burger fix works and looks nice, we can set up a discussion elsewhere altough sectioned.php is groupB territory so maybe we can get them to decide it.

Issue #6379, opened 2019-04-29
Add some styling to the Zoom range. 


a17okaoz:
Add some styling to the Zoom range. Now it´s looks kind of basic. 

<img width="184" alt="b1" src="https://user-images.githubusercontent.com/37792681/56883690-b7d6e380-6a67-11e9-8697-2f852c10e3a6.PNG">

a16tobfr:
What kind of styling do you have in mind? To be honest, I think it already looks good, and it agrees with the overall look of the page.
If you could give me some examples of what to add, I might get a better picture of how to solve this issue

a17okaoz:
The thing you drag to increase and decrease the Zooming can be round instead of the existing shape

<img width="204" alt="b1" src="https://user-images.githubusercontent.com/37792681/56898554-a276af00-6a91-11e9-9059-2da2d3a26d77.PNG">

PR #6637, opened 2019-05-06
fixed styling for zoom bar #6379

a16tobfr:
#6379 
added some styling to the zoombar.
Changed the handle and made it transparent when hovered.

Tried to make it so that only the handle was made transparent when hovered but failed with that approach. Can be made in a new issue in the future if it is requested


a14oscna:
I require further involvement to handle the merge conflicts before I can merge.

PR #6641, opened 2019-05-06
Notifies user when session is running out

### LINKED ISSUE #6272: Notify user when they are logged out, e.g. when the session expires
There is no form of feedback when a user is logged out due to causes other than clicking the logout-button.
Suggestion: use an occasional "ping" over ajax and e.g. display a modal or update the page when the user is no longer logged in.

See also: #5584

a17adasy:
fixes #6272

When the session has 15 minutes left a message box appears and the user has a chance to refresh the session. Otherwise when 15 minutes more pass the user will be logged out and the page refreshed.

For testing you can remove the comment on line 329 and 343 while also commenting out line 330 and 344 in dugga.js. This will set so the session box appears after 1 minute and the user is logged out after 2 minutes.

### DIFF
--- a/Shared/dugga.js
+++ b/Shared/dugga.js
@@ -326,8 +326,8 @@ function setExpireCookie(){
     if(localStorage.getItem("securityquestion") == "set") {
 				var expireDate = new Date();
 				// a test date so you dont have to actually wait 45 minutes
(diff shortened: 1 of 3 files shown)


a17adasy:
Changed the text of the alert box so it no longer is empty when session refreshes.

Issue #6475, opened 2019-04-30
Scroll too sensitive


a17alfcl:
Touch pad zoom is too sensitive

PR #6646, opened 2019-05-06
Grid is broken but now touchpad zoom is smoother, Fix for issue #6475

a17alfcl:
Fix for issue #6475

Touchpad zoom is acceptable but inertia is still a problem. Does not seem to have a good fix.
Grid is broken but it was before so scroll functionality was prioritized in this case. Works better on even values but not good.


a14oscna:
I will approve but I want you to create issue for the known problems before they are forgotten!

Issue #6649, opened 2019-05-06
refactor redudant code in accessed.js but keep search bar


a17poner:
Remove the code from row 20 to 32, keep the var filt = "" and search bar but change the search bar to $("#sort").after(filt); and below add in the following code (currently present in resulted.js but was in accessed.js before the merge where it got lost in a conflict.
`/*  Add filter menu   */

  document.getElementById("sort").style.display = "table-cell";
  document.getElementById("select").style.display = "table-cell";

  // Remove all children, at the time of writing there is only 1 and its unwanted.
  while(document.getElementById("dropdownc").hasChildNodes()){
    document.getElementById("dropdownc").removeChild(document.getElementById("dropdownc").firstChild);
  }

  var customFilterDiv = document.createElement("div");
  customFilterDiv.id = "customfilter";
  var columnFilterDiv = document.createElement("div");
  columnFilterDiv.id = "columnfilter";
  document.getElementById("dropdownc").appendChild(columnFilterDiv);
  document.getElementById("dropdownc").appendChild(customFilterDiv);`
This will place all of the buttons + search bar at the right of the navbar, the fix for it is in navheader.php  to move the menuhook echo from row 58 to below 77 but this will also be done by gruppB so we can probably live with it being in the wrong place for 2 more days :)

a17poner:
merged and closing

PR #6653, opened 2019-05-06
fixed 6649

a17maxsj:
fixed https://github.com/HGustavs/LenaSYS/issues/6649, Searchbar is back in accessed.php!


a17poner:
Hovering them gives undefined something error, hoverc, hovers, leavec and leaves are gone. Another issue up for that so merging this and then solving the hover functionality on its own in  #6657.

Issue #6648, opened 2019-05-06
Broken grid in diagram.


a17vikab:
in PR: #6646 the grid broke in order to fix the zoom-smoothness and now we need to recreate the grid so it works for all the values of the zoom.

PR #6655, opened 2019-05-06
Grid working when looking at it, Fix for issue #6648 

a17alfcl:
Fix for issue #6648 
Math is unpredictable but seems to work in all zoom modes, at least inside the canvas...


a14oscna:
Looks alright to me!

Issue #6643, opened 2019-05-06
Reset toolbar position should be removed


a17jonro:
The reset toolbar position in edit menu should be removed since the toolbar now is static and this option is unnecessary.

![image](https://user-images.githubusercontent.com/49195133/57223534-b167df00-7006-11e9-88b2-31ecacdd14ff.png)

PR #6659, opened 2019-05-06
resetToolbar removed

a16tobfr:
#6643
The resetToolbarPosition function has been removed from diagram.js
The drop-down-item div concerning reset toolbar has been removed from diagram.php


a17johpe:
which issue is this a fix for?

Issue #6461, opened 2019-04-30
Empty functions


a17celal:
At row 41 and 82 in diagram_IOHandler.js there are empty functions, createFolder and getImage. Should be investigated.

a17drama:
Seems like you already have a good idea of what needs to be done so this one goes to you.

PR #6660, opened 2019-05-06
Fix for #6461

a17celal:
#6461 

Deleted the function createFolder since it was never used. Commented out getImage and where it was used. It seemed like this function did not get implemented because of some old issue. I found this comment in the code:
`THESE OBJECTS ARE NOT IN THE TOOLBOX OR THE MENU AS THEY PROBABLY SHOULD BE IMPLEMENTED SOMEWHERE WITHIN ISSUE #3750`


a14oscna:
Nice catch!

Issue #6414, opened 2019-04-29
Positioning of A4 holes


a17drama:
When drawing out the A4 holes user should be able to chose to have them on the left or the right side if the A4 is vertical or up/down if the A4 is horizontal.

PR #6661, opened 2019-05-06
Positioning of A4 holes #6414

a17jacsv:
By clicking "Toggle A4-Holes" you change the position of the holes of the A4 from left-right & upper-bottom (Horizontal and vertical). At the moment you click the menu item to shift sides. A future issue would be instead of a checkmark put an icon who describe which side the holes are at. I will write that issue when it's approved and merged.

#6414

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -94,7 +94,7 @@ var p3 = null;                      // Middlepoint/centerPoint
 var snapToGrid = false;             // Will the clients actions snap to grid
 var toggleA4 = false;               // toggle if a4 outline is drawn
 var toggleA4Holes = false;          // toggle if a4 holes are drawn
-var switchSideA4Holes = false;      // switching the sides of the A4-holes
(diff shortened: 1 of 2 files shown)


a17jacsv:
> The actual function seems to be working well, the holes goes to the other side of the paper in both vertical and horizontal.
> 
> One issue I found though is that when this option is toggled on, when you press spacebar, leftclick anywhere on the screen and zooming in and a bunch of other stuff. It will keep switching, which seems weird.
> 
> Also don't forget to reference the issue in the description of the pull request like this #6414

Okay, that's weird. I will look into it.

a17jacsv:
Fixed the problem by making a new menu item to swift the holes from left to right. The problem was that the canvas is being redrawn every time you move the canvas and therefore the holes kept shifting... Now it should work as expected! 👍

a17jacsv:
The problem with enabling disabled items was before this issue and will be solved in another issue. The other problems will be solved now. Please test it again.

a16jenja:
Oh I missed the other comment, my bad, then it works good!

PR #6667, opened 2019-05-07
Fixed bottom margin of cookie popup msg button

a17fellu:
**Added 4px of bottom margin to the button:**

cookieMsgFixed


a17fellu:
Updated the fix, only removed the position: ablolute on the button in the css

Issue #6613, opened 2019-05-06
Problem with the grid in diagram


a17jacsv:
As you can see in the picture the grid is breaking in the lower parts when you are zooming the diagram. That's annoying and should be solved.

![Skärmavbild 2019-05-06 kl  09 51 42](https://user-images.githubusercontent.com/37792795/57212639-a734e900-6fe4-11e9-9bde-0b0702416741.png)

a17vikab:
@a17jacsv Can you describe the steps you took to achieve this problem?

a17jacsv:
I just zoomed out. I ran the diagram on a high-resolution monitor (3840 x 2160) which caused the problem. Maybe it’s just an issue for higher resolutions?

a17vikab:
> I just zoomed out. I ran the diagram on a high-resolution monitor (3840 x 2160) which caused the problem. Maybe it’s just an issue for higher resolutions?

It seems like the high-resolution may be the reason, exactly like you said. Either that or something for iMacs?

a17jacsv:
I’m not using iMac so i don’t know that matter. Only used a monitor attached to my MacBook Pro and saw this problem. Must be the high resolution that is the problem.

a17jacsv:
@a17vikab & @a14oscna: Seems like this problem has returned? Reopen this issue.

a17vikab:
@a17jacsv I can't recreate the problem, could you give us instructions on how you found the problem. There is a chance that it is hardware related.

a17jacsv:
@a17vikab: Same like before. You minimize the window (not using inspect) and the grid breaks. It was solved before but has returned.

See picture. The one to the left is from a branch I was working on and the other is from the group's branch.
<img width="1010" alt="Skärmavbild 2019-05-20 kl  13 30 27" src="https://user-images.githubusercontent.com/37792795/58018562-77690380-7b03-11e9-84b4-77062ad7de8e.png">

PR #6674, opened 2019-05-07
Should be fixed, Issue #6613

a17alfcl:
Fix for issue #6613 

Cannot test this as we do not have such screen.


a17jacsv:
Now it seems to work! 👍 Well done!

Issue #6622, opened 2019-05-06
Should not be possible to resize objects while they are locked.


a17vikab:
The old issue did not merge correctly, #6250, and the only thing left is that it should not be possible to resize the objects when they are locked.

Old PR: #6368

c17matjo:
It seems like 19802dd8 was the commit that broke this. In diagram.js:2539 an if statement was removed that checked if the selected object was unlocked before resizing. Re-adding that if-statement seems to fix it.

c17matjo:
Solved by #6697

PR #6697, opened 2019-05-07
Fix for issue #6622: Re:Stop resizing locked objects

c17matjo:
A previous fix prevented the used from resizing locked objects but merge commit 19802dd8 removed an important if-statement (diagram.js:2539). This PR re-adds it.


a14oscna:
Good to see it working again. Well done!

Issue #6349, opened 2019-04-26
When clicking the menu option, the menu should not close on 'View', 'Align' and 'Distribute'.


a17vikab:
At the moment, different menu options work differently. Some make the whole menu close and a few lets the menu still be open after being clicked. All options in 'View', 'Align' and 'Distribute' should not close the menu after being clicked.

Example of correct behavior is the 'Toggle A4 Holes', you can click it but the menu does not change.

It might be more menu options in other tabs but at least 'View', 'Align' and 'Distribute' need to change.

PR #6701, opened 2019-05-07
Fix for #6349. Preventing collapsing when clicking on menu items

a17drama:
#6349 is an issue about preventing the menu to collapse after one of the items is clicked. Not all menus prevented. Just those that where decided trough discussion. There was a double undo and redo function so I kept just one. In case you wonder why so much code was deleted.


a14oscna:
Looking good!

Issue #6224, opened 2019-04-24
Tool tip for position of an entity


a17drama:
While hovering over an entity and even while moving it have a tool tip showing the current position of the entity. To make things easier calculate the center of the entity and show the coordinates for that point. Free drawn entities should follow the same pattern. Calculate the highest and lowest X,Y and then calculate the center of the entity. In the tool tip make sure it informs that it's showing the coordinates for the center of the entity.

PR #6702, opened 2019-05-07
Fix for issue #6224: Tool tip for position of an entity

a16jenja:
Issue: #6224 

Added a tooltip for when hovering and moving an object, this tooltip appears above the object at the time. The design is not the most fancy ever, but it gets the job done. I removed the tooltip from lines and text boxes for the time being. I feel like a tooltip for lines seems unnecessary because the lines should be static between objects anyway. Tooltip for the text boxes is probably useful, but right now there is a problem where the tooltip box is not placed correctly above the text boxes. Let me know your opinion about this. There is no tooltip for free draw objects right now, since we are going to change them anyway if I understood correctly.

Let me know if I should change anything with how it looks or if there is something you want me to fix, like the text boxes thing. This is mostly a prototype right now, I just wanted some feedback on the changes I've made so far.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1588,8 +1588,8 @@ function developerMode() {
 var targetMode;     //the mode that we want to change to when trying to switch the toolbar
 
 //------------------------------------------------------------------------------
-// modeSwitchConfirmed: 
-// This function calls the switch methods if the change is accepted, called  
+// modeSwitchConfirmed:
+// This function calls the switch methods if the change is accepted, called
 // when clicking the dialog.
 //------------------------------------------------------------------------------
 function modeSwitchConfirmed(confirmed) {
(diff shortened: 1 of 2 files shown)


a16jenja:
Okay as soon as i put this pull request up I noticed that when you move the camera, the tooltip box does not follow the objects.

a14oscna:
I discussed with the reviewer and issue creator and the change I want to request is that the tooltip is moved to the bottom left corner, and that the entire solution is only enabled in developer mode.

a16jenja:
I'll get right on that

a16jenja:
I already talked to @a17vikab about this, but I feel like we should put this pull request on hold for todays merge and recreate the pull request after the merge so that I can make this better without having to stress out a poor solution to make it to merge.

a14oscna:
Agreed. I will put this on-hold for now and we will resume our work after the merge.

a17alfcl:
`var xPosition = canvas.width / 2 - 175;
var yPosition = canvas.height - 30;`

The position needs to be set in relation to canvas size

a16jenja:
> `var xPosition = canvas.width / 2 - 175; var yPosition = canvas.height - 30;`
> 
> The position needs to be set in relation to canvas size

I agree, good idea.

a16jenja:
The thing I want to implement is kinda the same as the one we have right now with showing mouse coordinates. But instead it will show an objects coordinates when you hover over it. This will be placed at the bottom left corner like we discussed. I tried to implement it using html/css but had some problems with it, therefore right now it's only a label created using javascript. Not as fancy in my opinion. Any tips on how to do that?

Issue #6239, opened 2019-04-24
Ability to lock proportions when resizing entities


b17robev:
One should be able to lock the proportions when resizing entities, could possibly be done while pressing shift and resizing an entity.

HGustavs:
Or locking height.

Entities should have a specific and equal height, attributes should perhaps be slightly more resizeable since the text may not fit all shapes of elipses but, really a maximum/minimum height but largely unrestricted width would be one way to achieve both.

a17vikab:
This issue is being solved when #6309 is done.

a17vikab:
The pull request, #6368, did not merge correctly and therefore this issue is now the main issue regarding this problem. 

Old issue: #6309

PR #6704, opened 2019-05-07
Fix for issue #6239: Ability to lock proportions when resizing entities

a17jonro:
#6239
Okay, so objects "officially" have 3 points, bottomRight, topLeft and middle and then the 2 remaining bottomLeft and topRight doesn't really exist. The functionallity implemented only works when resizing the bottomRight and topLeft points. The other 2 corner points somehow introduces problems and doesn't seem to work the same, and after much testing i couldn't get it to work with them. I might have done something wrong but since they are different and aren't tied to acctual points i'm not entirely sure.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -2505,6 +2505,28 @@ function pointDistance(point1, point2) {
 
     return [width, height];
 }
+//---------------------------------------------------
+// Check if the value causes the objects size to be <= the objects minimum size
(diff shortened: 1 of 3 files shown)


a17vikab:
As mentioned earlier, this is only a part of the solution and another issue should be created later.

a14oscna:
Make sure to create an issue for the remaining corners and I will merge this solution for now.

Issue #6688, opened 2019-05-07
Object names are wrong


a16tobfr:
![wrongObjectNames](https://user-images.githubusercontent.com/49199690/57288899-b0908500-70ba-11e9-85ed-1e4e905221d7.JPG)
The object names are wrong.
Should be: Entity0, Attr0, and Relation0
in this case...

a17vikab:
Connected to #6699.

a17drama:
@a16tobfr How do you want this to work. Chronological that just adds over time for each kind or do you want to reuse numbers. eg You create entity 0,1,2,3 and then delete entity 2 and 3. Next created one should be called 2 or 4? 2 if reused and 4 if we are keeping track of how many entities where drawn total trough out the life span of the diagram.

PR #6713, opened 2019-05-07
Fix for #6688 Numbering of different kinds start at 0

a17drama:
This fix should solve the issue but there is much more to do. Copying doesn't get effected by this change as it is now. Should be a different issue consider what @a16tobfr is wishing for this in the future. I will write a few more issues building on this one with regards to the wishes. All this does now is that it start's different kinds from 0.


a14oscna:
Alright, make sure the issues are created as stated above. I will merge this now.

PR #6716, opened 2019-05-07
Adapted the submission preview table for mobile devices

a17joaap:
iphone5

LenaSYS is obviously not made for mobile viewing, but this makes it at least possible to view the submission table on a phone, the smallest I could find in chrome was the iPhone 5 (pictured).

It still breaks on screens with a width smaller than 285px but let's not worry about people who browse the web on smart watches.


a17joaap:
That is a media query not added by this pull request, you can still click the submission header to show the div.

a17joaap:
The filename does get ellipses if the screen is smaller than 600px, it's just set in javascript so you have to reload the page.

a17poner:
Removed a <<<<<<<<<<<<< head tag that you had forgotten to remove, will merge this now.

PR #6718, opened 2019-05-07
AccessEd: Edit name box needs to be redesigned

### LINKED ISSUE #6567: AccessEd: Edit name box needs to be redesigned
image

First off, the red X icon looks like a delete button, but it's actually a cancel button. Change the icon to something more suitable please.

Second, the green background color is kind of ugly and doesn't really fit.

Third, the cursor should change to the pointer cursor when hovering the icons.

b17annni:
Fixes #6567 style edition. No hover.


b17annni:
It is a reset button.

b16johso:
I guess it is a reset button of sorts, but it's really a close/cancel button. Honestly, I think the best solution would be to get rid of the icons completely and just use text.

Take a look at the inline text editing section here: https://material.io/design/components/data-tables.html

b17annni:
image
New and improved. Went with icons to keep the original feeling.

a17poner:
This is put to be merged into w19, i'll update the base altough that might mean a ton of conflicts will show up

PR #6555, opened 2019-05-02
gruppB2019W19#6321

b17nicsv:
[Issue](https://github.com/HGustavs/LenaSYS/issues/6321)

[Week 18 branch](https://github.com/HGustavs/LenaSYS/tree/Gruppb2019W18%236321)

[Week 18 pull request](https://github.com/HGustavs/LenaSYS/pull/6422)

[Week 19 Branch](https://github.com/HGustavs/LenaSYS/tree/gruppB2019W19%236321)

Issue #6321, opened 2019-04-26
Adding clarity to sorting and filter


a17maxsj:
Making it easier to use every other line should be a different gradient. Like the gradients in other tables.

![image](https://user-images.githubusercontent.com/49199993/56794150-d93c9300-680d-11e9-8f0c-71eb1130da9e.png)

Example of how it could look(colors):
![image](https://user-images.githubusercontent.com/49199993/56794265-0f7a1280-680e-11e9-84be-7a8b5d8d8215.png)

Issue #6520, opened 2019-05-02
difficulties to access the filter options with mouse Resulted


c17johha:
<img width="154" alt="menu" src="https://user-images.githubusercontent.com/49199825/57070802-d1cb2d00-6cd8-11e9-9e86-227bf749b17a.png">

When hovering the mouse to access the filter options, one must be fast with mouse movement unless the options will dissapear. It seems to be caused by space between the hovering logo and the list

a17maxsj:
This was referenced in #6516

a17anter:
After I tried different options with CSS-styles I noticed that not just the "tratt" is affected but also the "arrows". I couldn't get the settings to work for both since there is an extra offset between them which I haven't found. See pictures below where they have the same CSS-styling.
![Tratt](https://user-images.githubusercontent.com/49204317/57296158-9364b200-70cc-11e9-9800-d45ac02defc5.PNG)
![Arrows](https://user-images.githubusercontent.com/49204317/57296159-9364b200-70cc-11e9-9633-3b98c2fb0cee.PNG)

b17nicsv:
Is the original issue solved and ready to be pull requested? @a17anter

PR #6764, opened 2019-05-08
Solution for 6321

c17alepe:
image

This was a lot simpler than I originally thought. I started a new branch and tried without javascript. One change in css and it was solved. The children of class checkbox-dugga is #eee on every odd line.
Solution for #6321

test site:
https://c17alepe.webug.his.se:20002/DuggaSys/resulted.php?cid=2&coursevers=97732

Note to Tim, everything related to this can be closed. This new solution is much cleaner and more consistant #6555 and my old PR #6721

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -2768,6 +2768,10 @@ span.arrow {
   text-overflow: ellipsis;
 }
 
+.checkbox-dugga:nth-child(odd){
+    background-color: #eee;
+}
+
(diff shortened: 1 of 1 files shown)


c17alepe:
I noticed the same problems as well, but those are not related with this fix. I'll check if the issues are mentioned and make the issues otherwise.

b17nicsv:
#6520 There is an issue for it already.

Issue #6387, opened 2019-04-29
Free draw/draw square objects can not be locked


a16tobfr:
As of now, the locking of free-draw/draw square objects is not possible.
It should be possible to lock free-draw/draw square objects. There is no reason why it should not be possible.

PR #6770, opened 2019-05-08
Fix for issue #6387: Free draw/draw square objects can not be locked 

a17jonro:
#6387 

Co-authored with @a17jacsv


a17jacsv:
> The issue here wasn't that the free draw objects couldn't be locked but that the lock and lock tooltip wasn't drawn to indicate that the object was locked (unless that was the case in an earlier branch). The new solution has added these visuals to the free draw objects when they are locked.

To enable the tooltip you'll have to hover the actual lock and not the center of the free-draw-object. The reason for this is because of the use of "paths" instead of "symbols".

Issue #6749, opened 2019-05-08
Menu toolbar is in the wrong position


a16tobfr:
canvas may have to be moved to fix this.
![toolbar](https://user-images.githubusercontent.com/49199690/57376055-5d443280-719f-11e9-9a08-eca4c681e36d.JPG)

Issue #6708, opened 2019-05-07
Reduce whitespace padding left/right for toolbar


a16tobfr:
related to #6699
<img width="58" alt="reduceWhitespace" src="https://user-images.githubusercontent.com/49199690/57358621-20b01100-7176-11e9-8a59-c4f2d89b6145.png">
Make it so that only the whitespace inside the black lines are present. Will need to move the canvas and resize the toolbar in order to make it look good across the entire page

HGustavs:
Screenshot of suggested fix?

HGustavs:
Looks all right. Question is (I have no answer...) if this look is standardized in the interface... is there some other dialog that we can borrow the styling cues from?

a16tobfr:
I am unsure about what you mean with dialog in this context

a16tobfr:
Issue #6749 fixes this issue. When the review for that issue is approved, this issue can be closed.

Issue #6699, opened 2019-05-07
Megathread: Toolbar / Menu Styling


a17vikab:
This is the issue that should fix all the design fixes related to the Toolbar / Menu.

HGustavs:
Before-after screenshots would be appreciated....

a17vikab:
Before:
![image](https://user-images.githubusercontent.com/37792464/57358613-1d1c8a00-7176-11e9-8d8e-c6ed6238f06b.png)

a14oscna:
As the toolbar label issue is yet to be fixed I will reopen this issue.

a17vikab:
After:

![image](https://user-images.githubusercontent.com/37792464/58407638-8a805400-806c-11e9-9f50-f66a1a20d6e1.png)

a16tobfr:
The toolbar labeling 
<img width="188" alt="toolbarLabelIssue" src="https://user-images.githubusercontent.com/49199690/58409318-15af1900-8070-11e9-9744-d3b59af254ee.png">
The labels should be next to each button instead of being in the corner as it is now

a18kaswi:
Unclear whats wanted, closing.

PR #6774, opened 2019-05-08
Toolbar styling changes

### LINKED ISSUE #6750: Left toolbar is no longer centralized
Should be an easy fix. move the position of the toolbar some pixels to the right.
Hard to see in the image but it is not centralized
This might be fixed by simply moving the canvas in #6749 
toolbarNotCentralized

a16tobfr:
#6749
Fixed position of the horizontal/vertical toolbar
This solution also fixes #6750, #6708, and finishes the Toolbar megathread #6699


a16tobfr:
Good find. No, it should not. Have now removed the comment.

Issue #6759, opened 2019-05-08
Hotkey exist but is no described in menu


a17drama:
![Eduted](https://user-images.githubusercontent.com/46029340/57376612-bfe9fe00-71a0-11e9-8a17-a9b585cf6173.png)
Hotkey for delete exists but is not show for the user in the menu. Should be like Undo/Redo function.

PR #6777, opened 2019-05-08
Fixed issue #6759: Hotkey exists but not shown as a label in the menu bar

a16jenja:
Issue: #6759 

Added a hotkey shortcut label to the delete option in the menu bar. Right now I have it as "Delete/Backspace" since you can delete an object with both hotkeys. But I can also change it to only "Delete" or maybe even "Del" if thats preferable. Let me know your opinion.


a14oscna:
I suggest you add "/Backspace" to the tooltip again since the backspace key still functions as a delete key in this scenario. In this case I judge that it is important to label all mapped options and not only the preferred ones.

PR #6785, opened 2019-05-08
Duggaed: Added submission object to submission types

### LINKED ISSUE #6631: Submission types when editing variants in DuggaED doesn't autofill.
Instruction file and Extra parameters is solved in pullrequest #6627. I was not able to solve the Submission types.. The hard part about the left divdialog is that we need to fetch the information from the Generated Param JSON that is stored in the database. The solution for Instruction file and Extra parameters is reading the values from the JSON string at Generated Param JSON field. The Submission types is harder to solve in a good way due to the possibility to have multiple rows. 

One variant selected: 
image
Another variant selected: 
image

a17joaap:
Fixes #6631 

The submission object is now parsed into the submission types area

I also went ahead and made the added submission rows (if any) disappear when the user clicks on update, enable, or disable. The submission row now also resets like you would expect when updating, as it retained the previous input before.


a17poner:
I went ahead and did the conflict resolution for this one before someone started revieweing it since it had a bigger conflict than what most pull requests do and this way the reviwer may find a problem that isn't related to the issue but related to the conflict and we can notice and resolve that before anything gets merged. (added @a17joaap 's changes from var submissionTypes to the end of the for loop he added to the bottom of the if(selected) for left dialog that @a17robda had changed/written)

PR #6791, opened 2019-05-09
issue #6753 - fix for not being able to change coursevers in dropdown box

### LINKED ISSUE #6753: Sectioned: Can't change course version
Nothing happens when I change version in the select box

a17pongo:
Fixes #6753 
Added separate function to specificly handle changing course version which should prevent breaking anything else depending on changeURL function in LenaSYS


a17pongo:
https://a17pongo.webug.his.se:20002/DuggaSys/courseed.php

Any course with more than one version should work

Issue #6795, opened 2019-05-09
Cursor does not change when hovering create line


a16tobfr:
The cursor does not change when hovering this button in either ER or UML mode.
Should be the same cursor as it is when hovering the other buttons in the toolbar
![cursorLine](https://user-images.githubusercontent.com/49199690/57439961-68529d80-7247-11e9-8b2c-b6f9a1711358.JPG)

PR #6797, opened 2019-05-09
added CSS class to fix the issue #6795

a16tobfr:
#6795 
Also removes the need for inline styling in the diagram.php document


a16tobfr:
Fixed

Issue #6780, opened 2019-05-08
showDugga/filereceive: File extension filter for upload window


b16johso:
Add a file extension filter when uploading files. At the moment it allows you to select any file type, but selecting the wrong one will give you an error on submit. If we add a filter, the user won't select the wrong file type in the first place. So for example when uploading a pdf file, only allow the user to select pdf files.

![bild](https://user-images.githubusercontent.com/37793460/57386009-ce8de080-71b3-11e9-8f04-494d95d99017.png)

PR #6801, opened 2019-05-09
Upload file filtering added

a17poner:
Most browsers should give the upload file popup a filter with zip and pdf files properly. fixing #6780

works for most browsers.

https://www.w3schools.com/tags/att_input_accept.asp


a17poner:
Yeah were thinking of making them seperate but they've been dealt with that same way so far so reused it, I don't think they have any other submission types where you actually upload the file. (also it is possible to upload a zip to a pdf upload and vice versa so thats a new issue to deal with later 😝 )

b16johso:
I know rar files are accepted in some cases. Should be possible to check the expected file extension with an if-statement, no? And then change the string with the input tag code accordingly.

a17poner:
I think they're called zip in the params that get sent to generic_file_receive even if its a rar, added every single submissionType (besides timesheet) to a dugga and checked the xhr responses and rar is never mentioned, my guess is that they just accept rar the same as zip on the server side, I'll make sure to add in .rar in the accept tag as well now that I've seperated them.

Issue #6465, opened 2019-04-30
Cursor change to indicate objects are movable in 'MoveAround' mode


a16tobfr:
The cursor is not shown in the print screen, but when you are in 'moveAround' mode and hovering over an ER entity for example, the cursor change to indicate that the object is movable. This is not true when in the 'moveAround' mode, and the cursor should not change when hovering objects here
<img width="287" alt="hoverCursor" src="https://user-images.githubusercontent.com/49199690/56959388-8cc6bf80-6b4d-11e9-9a98-76550fb61ef8.png">

a17jonro:
Exactly this issue doesn't seem to exist in the current branch. However this issue with the cursor changing is present when hovering over the points even when they're not visible. So I'll fix that and after disscussing it with a16tobfr, I'll also remove that you can hover symbols when in "movearound" mode since that is slightly related to this issue. Instead of creating a new issue for it.

PR #6810, opened 2019-05-09
Fix for issue #6465: Cursor change to indicate objects are movable in 'MoveAround' mode 

a17jonro:
#6465


a17jonro:
Read my comment on the issue for an exact description of what issues has been fixed.

a14oscna:
This solution partly solves the visual part of issue #6826 as well. Good job

Issue #6419, opened 2019-04-29
Zooming using the mousewheel should zoom relative to the mouse position


a17alfcl:
Currently using the mousewheel to zoom its zooming based of center point of canvas. This should be changed to be relative to the current mouse position.

PR #6813, opened 2019-05-09
Fix for #6419 Zooming relative to mouse coordinates

a17drama:
#6419 Zoom focuses on mouse pointer now instead of origo. Zooming in and out should not lose the object the mouse pointer is on.


a17drama:
Checked the conflict here. Keep the group branch code. I did not write anything in the snippet marked as "my code" in that conflict.

a14oscna:
Alright will do

PR #6817, opened 2019-05-09
Added Toggle all for checkboxes in Accessed

a17fellu:
Added another checkbox (Toggle all). If this checkbox is checked then all checkboxes will get checked. If the checkbox is not checked all other checkboxes will also be unchecked


a17fellu:
Will remove console log

a17fellu:
whoops will correct

PR #6823, opened 2019-05-09
Changed hamburger menu so that it works

a17amama:
Only changed between row 85-100


b17nicsv:
I agree. We can't merge something that removes includes and sessions. Needs to be put back before merge.

b17nicsv:
@a17amama Status on this?

b17nicsv:
Closing this because we don't know what is fixed, doesn't know which issue it is and the code also removes critical code which can't be changed.

Issue #6729, opened 2019-05-08
Change the cursor when the line button is selected


a17okaoz:
When the button line is selected and when you are going to draw a line between two objects change the cursor so the user knows that it can draw. Remove the cursor pointer or choose another pointer for increase the usability

a14oscna:
Further clarification of requirements are described in my comment on the references pull request #6828

a17jacsv:
> Further clarification of requirements are described in my comment on the references pull request #6828

Created a new issue of this updated issue: #6853

PR #6828, opened 2019-05-09
Change the cursor when drawing lines #6729 

a17jacsv:
Change the cursor to a pen when drawing lines.

Issue: #6729


a17jacsv:
I don't agree with you as you can only create lines when clicked-and-dragged on an entity. If the cursor is immediately changed to a pen it would cause unclarity. This is the best solution in my opinion.

a17jacsv:
> @a17celal: I don't agree with you as you can only create lines when clicked-and-dragged on an entity. If the cursor is immediately changed to a pen it would cause unclarity. This is the best solution in my opinion.

@a17vikab & @a14oscna your view on this?

a14oscna:
We have decided that the cursor should be standard pointer when draw line is the selected mode. No change on hover or during the drawing of a line, just standard pointer as it currently is when you hover over nothing. The change in code would be to ignore the current cursor changes when draw line is selected in the toolbar, rather than adding any new functionality.

a17jacsv:
Created a new issue of this updated issue: #6853 So I consider this PR closed...

a17celal:
@a17vikab @a14oscna Is this PR closed or should I approve?

a17vikab:
This PR will be closed and be continued in #6853.

Issue #6720, opened 2019-05-07
Toolbar labels are in the wrong position


a16tobfr:
As of now, the toolbar labels are positioned wrong. The lables should pop up next to the button, not beneath the entire toolbar.
related to #6699

HGustavs:
Screenshot. Unclear where we are and in what view.

a16tobfr:
<img width="76" alt="label" src="https://user-images.githubusercontent.com/49199690/57358286-5f919700-7175-11e9-8bf5-7c5573c96f98.png">

a14oscna:
An attempt has been made and merged but we are still looking for a solution which places the tooltip next to the hovered button. I will therefore not close this issue yet.

a17vikab:
Closing this issue, if the next generation of developers want to change the location they might need to create a new issue.

PR #6833, opened 2019-05-09
#6720: Styling on toolbox hover label

a17johpe:
https://github.com/HGustavs/LenaSYS/issues/6720

Placement of the tooltip text for when you hover over a button in the left toolbox has been changed to the top left corner of the canvas, as the previous placement doesn't work anymore with the new toolbox design.

Before:
toolboxtip

After:
toolboxtip2

I tried to make the tooltip relative to each button but weird things happened to the buttons on hover. I can try to investigate further on this if this placement of the label is what we really wish for. Otherwise I think it looks good with my solution.
toolboxtip3

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -1765,7 +1765,7 @@ div.submit-button:disabled {
   color: #fff;
   position: absolute;
   left: 70px;
-  top: 11%;
+  top: 55px;
   white-space: nowrap;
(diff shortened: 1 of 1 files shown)


a14oscna:
This solution will do for now, but I want the tooltip at right side of each button when it is hovered over rather than a fixed location.

PR #6840, opened 2019-05-09
Fixed upload instructions conditional

### LINKED ISSUE #6796: Dugga with text upload throws console error (generic_file_receive.js)
Uncaught TypeError: Cannot set property 'innerHTML' of null
    at Object.returnedDugga [as success] (generic_dugga_file_receive.js:79)
    at j (jquery-1.11.0.min.js:2)
    at Object.fireWith [as resolveWith] (jquery-1.11.0.min.js:2)
    at x (jquery-1.11.0.min.js:4)
    at XMLHttpRequest.b (jquery-1.11.0.min.js:4)

is thrown when entering a dugga that has a text upload.

a17joaap:
Fixes #6796 

No error message should pop up in the console now when navigating to a dugga using either of these templates:

- generic_dugga_file_receive
- group-assignment
- feedback_dugga
- daily-minutes


a17joaap:
I get that error on GroupC2019W20 branch so I don't think that is caused by this.

Issue #6738, opened 2019-05-08
When holding down a button like ctrl and switching browser tabs the diagram still thinks that button is pressed down when you return.


a17jonro:
This applies to ctrl and right click specifically, but it probably applies to other buttons as well. So when you return to the diagram browser tab the diagram believes the button is still pressed down, this can be deactivated by clicking the button again, but it can be very irritating.

PR #6844, opened 2019-05-09
Fix for #6738 Resetting key press when focus is lost

a17drama:
#6738 Switching tab or window now resets the pressed buttons. Buttons handled now are Ctrl and Shift. If you wish to add more it's just to put in what keys you wish to add. Very simple. Just check the code.


a17drama:
See if you like this better than the last solution.

a14oscna:
Way more efficient solution. 💯

Issue #6684, opened 2019-05-07
unnecessary log in groupbranch


a16tobfr:
![devOff](https://user-images.githubusercontent.com/49199690/57288398-ad48c980-70b9-11e9-93fe-75048aa5612f.JPG)
This should not be logged.

HGustavs:
console log or database log?

PR #6848, opened 2019-05-09
Fixed issue #6684: Unnecessary console.log in groupbranch

a16jenja:
Issue: #6684 

Removed the console.log command for when developer mode is being disabled.

Should the console.log command for when developer mode is being enabled still exist? If not, I can remove that aswell. Just let me know.


a14oscna:
Removed the log for when developer mode is toggled on/off as well. Do it and re-review it and I will merge thereafter.

Issue #6670, opened 2019-05-07
Access: Old version button no longer working


a17joaap:
These buttons:
![okhenrik](https://user-images.githubusercontent.com/37792198/57286882-8fc63080-70b6-11e9-9b9b-ada97438aee4.PNG)

Are supposed to show when a user has made a submission in an old course version, and should act as a link to the old version. It worked here https://github.com/HGustavs/LenaSYS/pull/6303

But it's broken now. Here is the output client side when console logging the object:
![outputClient](https://user-images.githubusercontent.com/37792198/57282668-d1061280-70ad-11e9-9e80-770499a49661.PNG)

Vers should never be 97732 since it is the active version, 97731 is an old version.

When running the SQL command in the code on the database we get this: 
![outputServer](https://user-images.githubusercontent.com/37792198/57282745-f430c200-70ad-11e9-8425-d22b32cc1846.PNG)

Something is wrong, since no data with the 97732 version is sent from accessedservice.php

HGustavs:
Can we add a slightly bigger screenshot so we can see better what/where the buttons are situated?

a17joaap:
New screenshot added for clarity

a17robda:
It seems like somone switched the id to a class, probably by mistake. It should work now.

a17robda:
![weird](https://user-images.githubusercontent.com/37792328/57531157-2013a800-7339-11e9-9702-a69fcce8cbce.png)
I do not seem to get the same issue. I do however get another issue where users do not retain their assigned older course on log in.

a17poner:
merged, closing

PR #6855, opened 2019-05-10
AccessED: Fixed bug where old version icon didn't show.

a17robda:
Fix for #6670 Someone had replaced the img ID with the class attribute.


a17joaap:
The issue was a little more complicated I'm afraid.

a17robda:
It goes back to the older version of the course if there is one.

a17robda:
I will take a closer look.

a17joaap:
And it's supposed to be 'class' not 'id' because of CSS:
```
.oldSubmissionIcon {
  height: 20px;
  margin-left: 10px;
  cursor:pointer;
  vertical-align: middle;
}
```

Issue #6820, opened 2019-05-09
Box shadow on only some elements


a17celal:
There is a box shadow around the square and the draw free. But all other elements are missing a box shadow. Either all elements should have it, or none.

![Capture15](https://user-images.githubusercontent.com/37792271/57449830-32201880-725d-11e9-8cdb-252558064b6a.PNG)

a14oscna:
Remove box shadow for all objects.

a17drama:
@a14oscna There is loads of code for making shadows on different objects. Do we remove all of that or just the shadows on the objects?

a14oscna:
We do not want shadows set by CSS nor by script functions, so I want you to remove any unnecessary functions related to box shadows.

PR #6860, opened 2019-05-10
Fix for #6820 Remove shadows on free draw objects

a17drama:
#6820 Function for putting in shadows on objects  exists here but the question is does it even work and should we keep it? Deleting dead code is easy in case supreme leader decides so.


a14oscna:
I answered your question in the issue related to this PR.

a16jenja:
In that case I would recommend removing the function and everything related to it that I mentioned in my other comment.

a17drama:
All is done according to the latest requests. @a16jenja Please check again.

Issue #6677, opened 2019-05-07
Navbar: Fix for mobile view


b16johso:
When viewing pages like AccessEd in mobile view, there isn't enough room in the navbar for all the icons and the search bar, so the login button gets pushed off screen.
![image](https://user-images.githubusercontent.com/37793460/57286073-d581f980-70b4-11e9-9ef9-40ebb3de8a2d.png)
Possible solutions:
1. Create a hamburger menu for mobile which contains all the navbar items (home, filter, search, etc).
![image](https://user-images.githubusercontent.com/37793460/57286535-b6379c00-70b5-11e9-8154-629ada850c9e.png)

2. Change the search bar in mobile view, so that it's only a search icon and clicking it opens up a search field.
![image](https://user-images.githubusercontent.com/37793460/57286454-88525780-70b5-11e9-974f-be9df117fe16.png)

b17carma:
I think a hamburger menu would look good :)

b17annni:
Waiting on #6739.

b17annni:
![image](https://user-images.githubusercontent.com/37792766/58020527-ac2b8980-7b08-11e9-97d7-55c5a87e9211.png)

b16johso:
This was fixed, but somehow reverted in todays test merge on branch gruppB2019W22#testmerge

PR #6867, opened 2019-05-10
Fixed functionality on mobile on resulted

### LINKED ISSUE #6739: ResultEd: Missing viewport meta tags
ResultEd doesn't have a viewport meta tag for mobile devices, so it starts way zoomed out in mobile view.

I guess use the viewport meta tag from DuggaEd:
`<meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0, user-scalable=no" />`

c17julka:
Fixes #6739 

Originally the issue was only about implementing the meta tags required for mobile view. I also added so that the filter buttons are placed in a column and the search bar is not in the header on mobile. The aim is to increase the functionality for mobile users.

Search bar inside header for fileed and accessed will also disappear on smaller screens. This will be fixed in #6677.


a16hiwah:
test is ok and the problem has been fixed! it looks better.

c17julka:
Make sure to add a review instead of a comment

PR #6881, opened 2019-05-10
If no timesheet table exists in the database, contribution will load

### LINKED ISSUE #6866: Make contribution function without errors even if there is no timesheet/dailyminutes
Henrik during the meeting mentioned that contribution needs to be able to work even if there are no dailyminutes/timehseets implemented, I think currently an alert is thrown with an error. Making it fail silently if its missing would be a lot more desireable.

a17joaap:
Fixes #6866 

Contribution will only throw an error if the timesheet table does not exist in the database. This simply handles the error by console logging a message saying that no such table exists.

You can simulate not having a timesheet table by changing row 537 in contributionservice.php:
`$query = $pdo->prepare('SELECT day, week, type, reference, comment FROM timesheet WHERE uid=:userid AND cid=:cid AND vers=:vers;');`

to:

`$query = $pdo->prepare('SELECT day, week, type, reference, comment FROM whatever WHERE uid=:userid AND cid=:cid AND vers=:vers;');`

### DIFF
--- a/DuggaSys/contribution.js
+++ b/DuggaSys/contribution.js
@@ -686,7 +686,12 @@ function returnedSection(data)
     return;
   }
   retdata=data;
-  if(data['debug']!="NONE!") alert(data['debug']);
+  if(data['debug']!="NONE!") {
+    if(data['debug']!="TIMESHEET") {
+      alert(data['debug']);
+    }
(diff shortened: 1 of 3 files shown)


a17maxsj:
Seems to be working! Nice :D
image

Issue #6724, opened 2019-05-08
Order numbering of diagram objects


a17drama:
Continue where #6688 ended Fix the issue when copying objects and make sure it pleases @a16tobfr with numbers that are reused. eg If the user only makes 6 entities their default names will be Entity 0, Entity 1, Entity 2, Entity 3, Entity 4, Entity 5. Then the user deletes Entity 0. Next Entity drawn should then reuse the number 0. As it is now the next entity would be called Entity 5 and the user would end up with two of those.

a14oscna:
Im happy with how the solution looks now after the #6890 merge and will therefore close this issue.

PR #6890, opened 2019-05-10
Removed acceptance of duplicates

a17jacsv:
If you delete an object and add a new one it's not one with a duplicated name as it was before. It does not reuse numbers it only adds +1 to so there are no duplicates. This works on all objects.

_This is a partial fix of issue #6724 as all requirements are not done and couldn't solve the re-usage of deleted entities as it was complex._


a14oscna:
I want you to do a few minor changes to the code so that it better follows the code standard before I merge this solution. In the conflict you can see that your indentation is not correct and we also want to avoid using "let" if possible, as written in the code standard guide.

a14oscna:
Looks good now but a "}" is missing on line 3438 to close the for loop.

Issue #6869, opened 2019-05-10
The Mail resulted button isnt purple when logged out


c17alepe:
![image](https://user-images.githubusercontent.com/38323901/57521625-15e5af80-7321-11e9-83bf-986859177382.png)

The purple seems to have different properties when logged out to not be clickable, making it white again.

PR #6892, opened 2019-05-10
Fixed the mail button styling when logged out 

c17alepe:
before
image

after
image

The buttons is now consistant when logged in and out

#6869 

testsite:
https://c17alepe.webug.his.se:20002/DuggaSys/resulted.php?cid=2&coursevers=97732


Shared/css/style.css
```diff
@@ -5088,9 +5088,20 @@ textarea#filecont{
...
+  width: 110px;
+  height: 30px;
+  border: 0px;
+  font-size: 14px;
+  transition: 0.1s background-color;
+  border-radius: 4px;
+}
+
+.resultedbuttons:hover{
+  filter: brightness(1.05);
```
c17johha:
```suggestion
  filter: brightness(1.2); 
```

c17alepe: I agree. Tested it and implemented the change! https://https://c17alepe.webug.his.se:20002/DuggaSys/resulted.php?cid=2&coursevers=97732
if you wanna see

Issue #6857, opened 2019-05-10
Login/out button covers the scrollbar. 


a16joare:
The login/out button is to much to the right and covers the scrollbar. 
![image](https://user-images.githubusercontent.com/49194898/57513582-df069e00-730e-11e9-96f8-12af01e7bb2c.png)

The image also shows that the shadow from the navheader is appearing under the button. Dont know if that is a issue but it looks kind of strange.

a17robda:
Seems strange, Might be a browser related issue.

a17robda:
![loginbar](https://user-images.githubusercontent.com/37792328/57515637-69510100-7313-11e9-9317-b6258cb6538e.png) 
![loginbar2](https://user-images.githubusercontent.com/37792328/57515788-bb922200-7313-11e9-9371-666fa90f669a.png)

I will close this since I could not replicate the issue on multiple browsers. It does not seem to be a css related issue?
This could be reopened if needed.

a16joare:
Test to access the contribution page and look there. Did some tests now and it works for all the other pages but not on contribution apparently
..

a17robda:
![contribar](https://user-images.githubusercontent.com/37792328/57516943-090f8e80-7316-11e9-9a41-641be75fb113.png)

It works for me on the contribution page aswell.

a16joare:
Hmm, seems strange. Will ask some more groupmembers to check this out.

b16johso:
I can't replicate this. Works fine for me in both Firefox and Chrome.

a16joare:
@a17alevu has the same problem..

a17robda:
Using GruppC2019W20 and still cannot replicate the issue.

a16joare:
Are you using the webserver or a local test server ?

a17robda:
Local

a16joare:
hmm very strange. We are some guys here in school and we all got it..

a17robda:
https://stackoverflow.com/questions/13028584/browsers-scrollbar-is-under-my-fixed-div
Skulle kunna vara något liknande. Öppnar så länge.

a17andby:
Tested this on my local test environment and on the test server, both on our weekly group branch. Can't reproduce this.

![image](https://user-images.githubusercontent.com/49235724/57526799-ab3b7080-732e-11e9-9e13-458da5aca1c4.png)

a16joare:
Did you check this on the contribution page?

b17carma:
[...]

PR #6898, opened 2019-05-10
Change contribution to use standard dugga format instead of an iframe…

b17carma:
… since it caused a lot of issues


b17carma:
Fix for issue https://github.com/HGustavs/LenaSYS/issues/6857

This also resolves the problem with there being 2 navbars

a17joaap:
https://github.com/HGustavs/LenaSYS/issues/4193 requested loading it in an iframe

b17carma:
Using an iframe created lots of problems (such as double navbar, problems with scrollbar overlaps etc) & with this solution the end result is the same. ShowDugga is used to show the contribution like any other dugga.

a17joaap:
If the iFrame was needed for some reason, an easy fix for the double navbar and navbar overlapping scroll is this:

Remove from contribution.php:
```
<?php
	$noup="SECTION";
	include '../Shared/navheader.php';
?>
```

Change in contribution.js:
`<iframe style='position: fixed; top: 50px`

a17poner:
A little bit unsure about this change since it does conflict with the issue mentioned by @a17joaap and seems like a pretty big change for what wasn't an issue on over half the groups environments. Is this change really needed?

b17carma:
I think this change is needed for a few reasons:
- It changes contribution to use standard dugga template which I think is preferable
- It minimizes future problems which could be caused by using a iframe
- It fixes the issue with the scrollbar position (not sure if this problem can be avoided completely if using an iframe)

Capture

The screencap shows how the contribution is right now with a margin set on the frame to show the problem more clearly. Even if the navbar is removed from the iframe, the scrollbar wouldn't act the way it does on other pages since it would appear below the login/logout button and not next to it.

Like I said earlier, I don't think there's any benefit to using an iframe and this way feels a lot more proper.

a17poner:
Still very uncertain on this since henrik specificly asked for it to be put in an iframe, not sure what advantages is but for some reason he wanted it to be in an iframe.

b17carma:
This seems reasonable and I'll do what you suggest @a17joaap, but it would be good to get some feedback from @HGustavs to clear up why an iframe is needed :)

a17poner:
That sounds fine, we get a solution that doesn't do to much and risk breaking things and we should hopefully get feedback from henrik if we really need the iframe, if we don't get a response before Friday I could try to remember to bring it up with him during the customer meeting.

b17carma:
Reverted previous changes. The page now resizes the navbar to make space for the scrollbar when not on a mobile device.

Issue #6859, opened 2019-05-10
Shortcut on toolbar items


a17okaoz:
Make so the items in the toolbar have a shortcut command for selecting them.  #6738 Look into this issue as well when doing this issue.  

The commands could be:
Line: shift + L
Entity: shift + E
Attribute: Shift + T
Relation: shift + R
Draw Free: shift + D
Uml create class: shift + C
Help: shift + H
Mode switch: shift + M

a17drama:
shift + D should trigger the developer mode and shift + F could be used for free draw. What is help used for?

a17okaoz:
> shift + D should trigger the developer mode and shift + F could be used for free draw. What is help used for?

a17okaoz:
Yes, that solution is good!. Forget about the help command it is unnecessary

PR #6907, opened 2019-05-10
Fix for #6859 Shortcuts for toolbar and some other things

a17drama:
#6859 Almost according to the list in issue. Switched attribute till Shift + A and put in Text that is coming in the future on Shift + T. Rest is according to the list/comments in the issue. Also ordered the key declarations in key code ascending order.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1773,7 +1773,7 @@ function developerMode(event) {
 
 }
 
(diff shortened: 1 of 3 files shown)


b17robev:
Seems to work correctly

a14oscna:
I do not confirm the CSS changes, we should avoid the usage of !important as well as writing duplicate code.

Issue #6788, opened 2019-05-09
Disabled menu items is visable after recent merge


a17jacsv:
See picture.... 

_"Display all tools", "Toggle A4 Orientation", "A4 Holes Right"_

![Skärmavbild 2019-05-09 kl  09 33 20](https://user-images.githubusercontent.com/37792795/57435371-81565100-723d-11e9-946a-46a8b9b33ca9.png)

Related to issue: #6727

c17matjo:
How should the ER/UML selector work when entering developer mode? Right now, entering developer mode deselects the ER or UML option but doesn't disable them meaning you can still enter ER/UML mode without leaving developer mode. Should the ER/UML options be greyed out when in developer mode?

a17jacsv:
> How should the ER/UML selector work when entering developer mode? Right now, entering developer mode deselects the ER or UML option but doesn't disable them meaning you can still enter ER/UML mode without leaving developer mode. Should the ER/UML options be greyed out when in developer mode?

Yes, that would be the best solution as in developer mode all objects are available.

PR #6909, opened 2019-05-10
Fix for #6788: Gray out options correctly in view menu

c17matjo:
I've tried to make the view options gray out correctly when disabled. I haven't fixed other issues like how you can still click on disabled options but that can be fixed in another issue or pull request. Because of that, the reviewer should ignore problems relating to clicking on disabled options. 

This PR makes the ER/UML options gray out when entering dev-mode and should make the A4 options gray out correctly.


c17matjo:
I think that is correct behaviour. The options are only hidden when they're not supposed to be modified. The "A4 Holes Right" option is visible when both the virtual A4 option is enabled and the A4 holes are visible.

a14oscna:
c17matjo's interpretation is correct. Well done!

Issue #6669, opened 2019-05-07
Possible to destroy JSON string when editing variants. (Duggaed)


a16joare:
When editing a variant an JSON string is generated when values is entered in the fields.  You can change all fields included the "Generated Param JSON" field.. This makes it possible to change the values inside the JSON string and also destroy it by changing values the wrong why, like adding characters outside the JSON string object.. 

![image](https://user-images.githubusercontent.com/49194898/57282579-a1efa100-70ad-11e9-816e-b0ae33911ec0.png)

![image](https://user-images.githubusercontent.com/49194898/57282635-c3e92380-70ad-11e9-9d05-9422e675ebbe.png)

HGustavs:
I think this is desired, since the tool is not fully two-way yet, adding and changing in the list, 100% reflecting the text in both directions, changed text updates dialog etc.

The most common way we work currently is copy-pasting a correct string from a test that works, and making minor changes in that view and then saving.

a17poner:
So the issue should really be to make the changes if possible in the json textarea change the form part @HGustavs ? Seems sort of doable with seeing if there are anything in submission and then updating the submisison rows etc and then doing nothing for unkown params if they were to be misspelled etc.

a17poner:
it now works a bit better twoways, closing this.

PR #6927, opened 2019-05-13
Added reverse/two way functionality to jsonText

a17poner:
supposed to fix the issue part of #6669
(in the comments). Editing the json text area should if possible make corresponding changes to the form on the left.


a17poner:
Yeah, might be some better way of doing it, currently it uses onchange, tried on key up and that created a horrifying experience as after each key pressed it would cause a loop of changing the left div form which caused the onchange over there to update the textarea which placed your cursor at the end of your input. But it should be able to get fixed, thats an issue creation for you if you'd like 😉

PR #6932, opened 2019-05-13
Fixed changing font size in codeviewer  - Grupp c2019 w20 6925

a17maxsj:
Previously when changing font size for a document nothing happened when changing since the code did not exists. Added the code for it and it seems to be working.

You can test this by going in to the course "Packningsbyte på om606 med Bosse bildoktorn" and then the codeviewer dugga (or creating it yourself somewhere else.)

Small and big text
image


a17maxsj:
> It does work, the font size now changes. However, I get this alert everytime I change it:
> 
> 4f21460d258e844fcac16b97ca97a9d3
> 
> Is it something we want to fix or leave it?

This is a global issue currently being investigaed, not specific to this issue.

Issue #6943, opened 2019-05-13
Cleaning code: diagram.js


a17jacsv:
Cleaning the code of file: diagram.js which contains 3782 lines of code.

PR #6944, opened 2019-05-13
Cleaned code of diagram.js #6943

a17jacsv:
Follows the code standard much better now...

Issue: #6943

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -3,8 +3,6 @@
     THIS FILE HANDLES THE MAIN FUNCTIONS OF THE DIAGRAM
 
 ************************************************************/
-
-
 /***********************************************
     CANVAS FUNCTIONS NEEDS TO USE THE MAPPING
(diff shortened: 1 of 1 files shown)


a16tobfr:
It looks like you have improved the readability of the code. However, was the purpose also to make the code follow the standard? In that case, function brackets should not look like this:
~~~
function someFunc() {
  ...
}
~~~
The standard for function brackets is like this:
~~~
function someFunc()
{
  ....
}
~~~
Can be seen in: https://github.com/HGustavs/LenaSYS/blob/master/code_standard.md

a14oscna:
I can not solve these conflicts on my own.

This one is tricky because a lot of changes has been made to diagram.js since you began working on your branch. You will either have to be very delicate while resolving the conflicts or pull a up-to-date version of diagram.js and redo the work in smaller increments to prevent conflicts like this to occur.

a17jacsv:
I've solved the conflicts. Please @a16tobfr: test it again and see if there are any errors in the application's functionality.

Issue #6875, opened 2019-05-10
Diagram lock should be able to unlock when clicking on the lock.


a17vikab:
The customer want to unlock objects when clicking on the lock.

![image](https://user-images.githubusercontent.com/37792464/57524526-5d236e80-7328-11e9-929e-0cf4447a3b68.png)

PR #6961, opened 2019-05-13
Fix for issue #6875: Diagram lock should be able to unlock when clicking on the lock.

a17jonro:
#6875


a17jonro:
Everything from line 3062
`if (uimode == "CreateLine" && md == mouseState.boxSelectOrCreateMode) {`
to 3185
`diagram[markedObject].connectorTop.push({from:p2, to:p1});`
is only an indent change but it looks really weird in git.

PR #6962, opened 2019-05-13
Added sort options to sort dropdown

### LINKED ISSUE #6837: Sort box in accessed is blank
when hovering over the sort button this shows up:
image

a17joaap:
Fixes #6837 

Now possible to sort through the dropdown.


a17poner:
Think the code is alright, not sure if sorting is like filtering where there is a function in sortableTable to populate the dropdown. But if it works it works!

Issue #6858, opened 2019-05-10
Accessed: Strange clipping on hovering "arrows"-icon.


a17robda:
![strangeclip](https://user-images.githubusercontent.com/37792328/57514122-05790900-7310-11e9-9882-35b8b56de812.png)
There is some strange clipping going on with the box while hovering over the "arrows" in accessed compared to hovering over the other icons. I have no idea what it is used for.

a17robda:
Fixed. About to pull req.

Issue #6920, opened 2019-05-13
Accessed: Hovering on the "funnel"-icon menu for filtering is unreliable


a17robda:
WEBM of issue:
https://imgur.com/a/rpyGf5b

Seems unreliable to have the window for filtering disappear when moving the mouse on it.

a17andby:
@a17anter Haven't you worked on a solution for this in resulted?

a17andby:
Found it, https://github.com/HGustavs/LenaSYS/issues/6520 

I will check if their solution (https://github.com/HGustavs/LenaSYS/pull/6816) also fix our issue

a17anter:
Yes I believe that solution also works for your issue.

a17andby:
I don't think that we need to solve this issue since @a17anter have done a solution that will also affect Accessed when we merge our groups together.

I have checked and tested @a17anter solution on Accessed and it will fix this issue as well. 

What do you say @a17poner ? Should we just close this issue, since it will be fixed in the weekly merge with groupB? So we don't get any unnecessary merge conflicts

a17poner:
We'll leave this issue as is and hopefully and likely it will get fixed by @a17anter in the weekly merge

PR #6970, opened 2019-05-13
Accessed: Fixed weird clipping on hover-dropdown on "Double-arrows icon"

a17robda:
Fix for issue #6858

### DIFF
--- a/Shared/navheader.php
+++ b/Shared/navheader.php
@@ -67,9 +67,9 @@
 						echo "</td>";
 
   					echo "<td id='sort' style='display:none' class='navButt'>";
-						echo   "<span class='dropdown-contain;er' onmouseover='hovers();' onmouseleave='leaves();'>";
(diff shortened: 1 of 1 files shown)


a17robda:
I am aware of the issue with getting into the actual "menubox" as it was a bug that was present before the fix. However, the fix is only for the issue which is the weird clipping of the menu compared to hovering over the other icons (Which also have the same issue at the moment in this weeks branch. See issue #6920 for the same bug.).

a17poner:
There is a big change to this being done by groupB, I won't merge this until possibly after wednesday. Hopefully their fix will solve this for all of us and this wouldn't be needed then.

a17robda:
This still seems to be an issue with the weird clipping after merge.

a17poner:
I don't have it after hard refreshing with gruppC2019W21 and it seems to work as intended now.

a17poner:
Seems to be fixed now already for all that tested it on location here, closing this.

Issue #6687, opened 2019-05-07
Drawing lines between copied/original objects


a16tobfr:
drawing lines from original attributes/entities to copied attributes/entities/relations results in wrong connections.
related to #5040

Issue #6686, opened 2019-05-07
lines are not attached to correct objects on copy


a16tobfr:
![lines](https://user-images.githubusercontent.com/49199690/57289918-9eafe180-70bc-11e9-84c0-aa2b34c2a2f0.JPG)
lines are not attached to correct objects on copy
related to #5040

a17vikab:
Related issue #5040, the work will continue on that issue.

Issue #6685, opened 2019-05-07
removal of copied items removes original connections


a16tobfr:
removal of copied item also removes original connections
Related to #5040

a16tobfr:
This was fixed in issue #5040

PR #6979, opened 2019-05-13
Issue#5040  Copy functinality

a17okaoz:
This pull request is a partial fixed. This solved issue #6685, #6686, #6687. Can copy of entities, attributes, and relation but copy the line doesn't work yet.  When removing the copy objects the original line doesn't remove any more.

We have removed "Object.assign" witch only create shallow copies. Instead, we used JQuery.extend witch make deep copies.


a17johpe:
New issue: copying UML classes doesn't copy fields and methods
cop

a16tobfr:
Removed the ability to copy the UML lines. Will be implemented in another issue where it will work as expected.

Issue #6963, opened 2019-05-13
Short command for "Clear Diagram"


a17okaoz:
Implement a short command for the button "Clear Diagram". This will make the process easier. 

@HGustavs Do you have any preference for this command?

a14oscna:
The short command technically already exists as: Ctrl+A followed by Delete, since doing this selects all objects and deletes them. This works exactly as the "Clear diagram" option in the menu, therefore we do not need to implement any additional short command. 

Simply add the tool tip into the menu option.

PR #6985, opened 2019-05-13
Fix for #6963 Shortcut for clear diagram

a17drama:
#6963 Tool tip for "Clear diagram" added according to the comment from the supreme leader.


a14oscna:
Good that you added tool tip for undo and redo as well.

PR #7009, opened 2019-05-14
Fix #6853

### LINKED ISSUE #6853: Remove hover-functionality when drawline-tool is used
When drawing-tool is used, the cursor over the canvas will be "pointer" and no other hover functionality will be used when the tool is active. 

This will be done instead of the former issue: #6729

a17jacsv:
Update fix for issue: #6853

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -2756,22 +2756,22 @@ function mousemoveevt(ev, t) {
                     }
                 }
             } else {
-                if(uimode == "MoveAround") {
-                    canvas.style.cursor = "all-scroll";
-                } else if (uimode == "CreateLine") {
-                    //When CreateLine-button is selected the cursor is "pointer".
-                    canvas.style.cursor = "pointer";
-                    //If objects are hovered while button is selected, the cursor remains the same (pointer).
-                } else if(hoveredObject && !hoveredObject.isLocked) {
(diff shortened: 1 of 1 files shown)


a14oscna:
I dont get it when just looking at the code, it looks like the code you added + some existing code exactly matches the code you removed? See the 2 sections marked in blue:
abc123
What has exactly happened with this fix?

a17jacsv:
@a14oscna: It seems like there was some weirdness in the "files changed" section as there were more things edited as there were shown.  I updated the indentation of the code and now it looks better and you see what is changed from the beginning.

a14oscna:
Alright. Could you explain where exactly unwanted hover effects were still present because when testing? Your fix looks so similar to our group branch that I can not find the issue.

a14oscna:
Nevermind, I think I found it.  Next time make sure to explain in the solution what you did to fix the issue, and explain in the issue what the problem is rather than only explaining what needs to be done.

a17jacsv:
> Nevermind, I think I found it. Next time make sure to explain in the solution what you did to fix the issue, and explain in the issue what the problem is rather than only explaining what needs to be done.

This was a quick fix that I told you yesterday in person to merge right away. But the message did not came through. Sorry for this.

a14oscna:
Thank you for the detailed explanation, I did not mean to sound rude, I just had to postpone due to the weird view in "files changed" and forgot that you mentioned it verbally to me. No harm done, good job!

Issue #6879, opened 2019-05-10
Toolbar options should not deselect when used once.


a17vikab:
It creates frustration when eg the line-tool deselect when used. It should only toggle back when clicking on the button again.

PR #7010, opened 2019-05-14
Removed logic for disabling 'draw line' after a line had been drawn #6879

a16tobfr:
#6879
Removed the logic for disabling the 'draw line' functionality after a line had been drawn.
Now you can draw as many lines as you want.


a16tobfr:
Only related to UML lines. working on it now

a16tobfr:
I've put back the annoying code for UML lines. This **issue should not be closed** but can be merged after approval so the lines work correctly for ER entities when the merge comes. Will continue the work on UML lines but it is not clear what is causing the problem so it might take som time to fix

Have spoken to Oscar about this and he approved this approach.

a14oscna:
Can confirm what a16tobfr wrote.

PR #7016, opened 2019-05-14
Fix for sorting files according to size

### LINKED ISSUE #6930: Fileed: Sorting after size does not work properly
When sorting after size in the fileed view, the largest number ends up on top which is not always desired. 
Example: 
image

a17ludca:
This fixes #6930 the issue of sorting files according to size. 
Testsite : https://a17ludca.gruppb.webug.his.se:20002/DuggaSys/fileed.php?cid=2&coursevers=97732
or branch GruppB2019W20#6930


c17alepe:
This works much butter. I approve

Issue #6876, opened 2019-05-10
Alter sorting by grade in resulted to asc/desc/pending


b17timjo:
See discussion in #6515

Instead of sorting by -/P/U/G/VG where some of the options have no clear purpose, sort by three new alternatives: pending grades, and grades ascending/descending.

Quoting HGustavs full post from #6515 for context:
> Use centric sorting.... off the top of my head I want the most relevant items first.... hopefully my view of relevant is same as everyone elses....
> 
> For all views
> 
>     * We always want un-viewed tests at bottom in all sort orders.
> 
>     * After that visited but not handed in.
> 
> 
> For grade specific views sort in order of grades given the sort order... if U start with U and then falling
> if VG start with VG and then falling.
> 
> No G mode.
> 
> Howabout Ascending grade mode, Descending Grade mode and Pending.

c17johha:
What does it mean when an element is purple like this:
![image](https://user-images.githubusercontent.com/49199825/57684637-77b25c00-7636-11e9-98ae-5d1e8fe42654.png)

Compare to a grey element:
![image](https://user-images.githubusercontent.com/49199825/57684704-9d3f6580-7636-11e9-90fd-ca7a92f7ece6.png)

Is one of these "pending"?
@b17timjo @HGustavs

b17nicsv:
I think the light grey and dark grey both are pending, but the background makes them light/dark.

c17johha:
Firefox and Chrome doesn't call the function "newCompare" the same amount of times for the exactly same conditions. I believe that this is the reason for the inconsistency between Chrome and Firefox

PR #7031, opened 2019-05-14
Sorting by grade in resulted to asc/desc/pending #6876

c17johha:
image

image

image
 PS. if changing grades on an element, one must reload the site, unless wrong order may appear.
Is this the intended behaviour?

### DIFF
--- a/DuggaSys/resulted.js
+++ b/DuggaSys/resulted.js
@@ -1085,11 +1085,11 @@ function renderSortOptions(col, status, colname) {
 			}
 		} else {
 			if (status == 0) {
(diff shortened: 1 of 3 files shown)


a17pongo:
not reviewing but I would recommend using short names, ascending - ASC, I think it would look a little better

b17nicsv:
I agree with a17pongo.

c17johha:
@c17alepe Did you reload the site?

c17johha:
Updated names
image

c17alepe:
Yes i reloaded the site. You can try yourself. its the Ola-Conny who is causing this problem when filtering to Pending on Geometri

c17johha:
@c17alepe 
apparently Firefox is causing trouble for me. Chrome seems to be working fine. My bad, I better fix this.

Issue #6933, opened 2019-05-13
Toolbarstate should be saved in localStorage


b17robev:
The toolbars state should be saved using localStorage. If you change to UML mode and decide to refresh the browser, the toolbar state should stay in UML mode.

HGustavs:
Yes! Good idea.

a17vikab:
Appears to be solved.

PR #7033, opened 2019-05-14
Fixed issue #6933: Added localStorage to toolbar mode

a16jenja:
Issue: #6933 

Added localStorage for storing if ER or UML mode is enabled, therefore when you refresh the page it should stay the toolbarstate mode that was enabled before refreshing the page. Instead of defaulting to ER mode every time.

If you go to F12, then the application tab and then to Local Storage you can see the toolbarState being stored there.

1 = ER mode
2 = UML mode
3 = DEV mode

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1773,19 +1773,22 @@ function developerMode(event) {
         setCheckbox($(".drop-down-option:contains('Display All Tools')"),
             crossDEV=true);                                                             // Turn on crossDEV.
     } else {
+      toolbarState = localStorage.getItem("toolbarState");                             // Change the toolbar back to ER.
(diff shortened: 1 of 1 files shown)


a16jenja:
I tried a lot of different things and it does not seem to work, even if I would find it, developer mode is not technically on when refreshing the page. What I mean is the x's that are on the objects and stuff are not showing up since I think another function handles all that. So maybe it's better that the checkbox is not there? I know the toolbar is still in Developer mode but something to think about. I'll keep trying to find a fix in the meantime, let me know what you think @a14oscna @a17vikab

a14oscna:
Due to lack of time before the master-merge I will merge this solution. I also want you to create an issue for whatever is left for you to fix after this solution is merged.

PR #7034, opened 2019-05-14
codeviewer: Ability to maximize a window

a17fellu:
Only a QOL improvement to codeviewer that lets you maximize the window you want to have maximized with a single button press. Right now it supports all templates except for template 9(template 9 is a little broken)

### DIFF
--- a/DuggaSys/codeviewer.js
+++ b/DuggaSys/codeviewer.js
@@ -259,7 +259,7 @@ function returned(data)
 	}
 
 	//hides maximize button if not supported
-	hideMaximizeButton();
+	hideMaximizeAndResetButton();
 
 	// Allows resizing of boxes on the page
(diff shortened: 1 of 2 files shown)


a17fellu:
I have a reset button i just forgot to push it. Will add that and re-request review

a17robes:
> I have a reset button i just forgot to push it. Will add that and re-request review. As for the single pane template it shouldn't have the maximize function since you cant change its size

True that!

a17fellu:
> 
> 
> Works like a charm! Reset button is a nice touch (exept for template 9, fix later?)

Template 9 itself has a few bugs. This means that it isn't possible to add it to it at this moment. But when it gets fixed its an easy addition.

a17poner:
Nice, is there a issue for fixing the broken things about template 9? if not we should probably make one and have it fixed 😛

Issue #6142, opened 2019-04-23
FAB hover area is a little small which makes it easy to accidentally close the expanded item list


b17carma:
An easy way to fix this is to make FABs follow the principle outlined at https://material.io/design/components/buttons-floating-action-button.html#extended-fab under the "Behaviour" heading.

This change would make the acceptable hover area larger which makes it harder to close the list by accident, especially when there are many expanded actions in the list.

HGustavs:
The problem with this is that the FAB-s tend to cover other more useful interface elements.

We need to make something that works in all of the FAB views.

Can we make a mock-up of this so we can see if it causes other (worse?) problems?

b17carma:
I was thinking something along the lines of how e.g lastpass does it where the tooltips are always visible and can be hovered over as part of the item.

![image](https://user-images.githubusercontent.com/37834592/56565881-a8f8b880-65b1-11e9-9145-04a5533f6dea.png)

b17carma:
In dugga it would look something like this on desktop

![image](https://user-images.githubusercontent.com/37834592/56566650-79e34680-65b3-11e9-90d6-cf824d5a5ff4.png)

![image](https://user-images.githubusercontent.com/37834592/56566679-85cf0880-65b3-11e9-8692-618aeab251fd.png)

And on mobile:

![image](https://user-images.githubusercontent.com/37834592/56566706-9bdcc900-65b3-11e9-952a-1612b14cb7a8.png)

a17maxsj:
I think this is a great idea!

a17maxsj:
Adding some / a lot of  opacity while not hovering over the tooltips and then making them 100% visible while hovering?

a17maxsj:
Was this implemented? any news on this :)

a17maxsj:
bump, @b17carma any news?

b17carma:
Leaving this for now, if anyone else would like to work on this they can

a17robes:
Not sure how to solve this, been trying but have not made any progress. Maybe let someone else try to get this fixed? @a17alevu

a17alevu:
Okey, someone else will be assigned to this issue.

a16lisri:
I've made a solution to this. Shown here: https://i.gyazo.com/6da240ae46796312ab5ac21b6555f48e.mp4
Is this good?

PR #7049, opened 2019-05-14
Fix for #6142

a16lisri:
Made a fix for https://github.com/HGustavs/LenaSYS/issues/6142. The tooltips (in the fablist) next to the icons is now always shown. When you hover over the icons, the tooltips are 100% visible again. Solution shown below:

fablist


a17poner:
Don't forget to label your pull requests, I almost never check for un labeled ones so I will be much faster at merging your code if you label it :)

a17poner:
but nice change 👍

Issue #7052, opened 2019-05-14
Contribution-page broken


a17ludca:
A line of code in sortabletable.js has recently broken the contribution-page. This occurs in the branch of grupp b.

PR #7054, opened 2019-05-14
Fixed sorting week-column in contribution

### LINKED ISSUE #6557: sortableTable: Sorting does not work correctly on numbers
Example:
weeksort

I think this is because the sort function treats the numbers as a string. I suggest checking if the cells are numbers with isNaN() as a generic solution.

a17ludca:
Fixes #6557 
To test, go to "webbplatsdesign" and then click contribution
Testsite: https://a17ludca.gruppb.webug.his.se:20002/DuggaSys/courseed.php
or branch GruppB2019W20#6557-2


b17timjo:
This should fix #7052 as well, correct?

PR #7055, opened 2019-05-14
Codeviewer: Added template options for faster workflow

### LINKED ISSUE #2277: Faster Workflow in Code Viewer
I would like a way to assign files to all views at once.

Right now it is Select Template, Ok, Cog Wheel First View, Select File, Ok, Cog Wheel Second View, Select File, Ok, Cog Wheel Third View, Select File, Ok, Cog Wheel Fourth View, Select File, Ok

...

I would like a workflow that works like this:

Select Template, Below the template selection images, in same view there are file dropdowns for View1, View2 View3 View4 for selecting Kind, Language (if appropriate) and File for each.

This way, most of the work CAN be done in the template selection window, and it can then be changed by the current normal way of doing it with the cog wheels.

a17joaap:
Fixes #2277 

This PR adds a form when picking a template:

Which will update the boxes directly, eliminating the need to edit them separately.
Could be made prettier, but that's another issue maybe.

### DIFF
--- a/DuggaSys/codeviewer.js
+++ b/DuggaSys/codeviewer.js
@@ -502,28 +502,44 @@ function changeDirectory(kind)
 	var dir;
 	var str="";
 
-	var chosen=$("#filename").val();
+	var kindNum;
+	if (kind.id) {
+		kindNum = kind.id.split('_')[1];
+	}
+	
+	if (kindNum) {
+		var chosen=$("#filename_"+kindNum).val();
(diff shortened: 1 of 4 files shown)


a17poner:
Nice, could use some documenting in the code since there are now 4-5 areas required to edit to change the amount of boxes and positioning, I'll add an issue for that to do next "week" and merge this :)

Issue #6573, opened 2019-05-03
Remove draw square functionality


a16tobfr:
After the last meeting, this functionality was deemed unnecessary and will, therefore, be deleted. 

The code handling this is mainly at the bottom of the diagram_objects.js file but some parts of it exist in other places as well.

a17vikab:
#6611 will solve this issue.

a17vikab:
Will be fixed in #7059

PR #7059, opened 2019-05-15
Issue #6573

a17alfcl:
Issue #6573

Removed code for draw square

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1572,7 +1572,7 @@ function eraseSelectedObject() {
 
 function setMode(mode) { //"CreateClass" yet to be implemented in .php
     uimode = mode;
-    if(mode == 'Square' || mode == 'Free' || mode == 'Text') {
+    if(mode == 'Free' || mode == 'Text') {
       uimode = "CreateFigure";
       if(figureType == "Free") {
           cancelFreeDraw();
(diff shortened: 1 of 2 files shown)


a17johpe:
There are still some times in the code where figureType "Square" is evaluated and on row 2562 (see code below) in diagram_objects.js the function that was removed, figureSquare(), is called upon. Will the code ever reach that point? 
~~~
function createFigure() {
    startMouseCoordinateX = currentMouseCoordinateX;
    startMouseCoordinateY = currentMouseCoordinateY;
    if (figureType == "Free") {
        figureFreeDraw();
    } else if (figureType == "Square") {
        figureSquare();
    }
}
~~~

Asking if removal of all references to figureType "Square" should be removed in this issue or if that is a separate issue.

a17alfcl:
> There are still some times in the code where figureType "Square" is evaluated and on row 2562 (see code below) in diagram_objects.js the function that was removed, figureSquare(), is called upon. Will the code ever reach that point?
> 
> ```
> function createFigure() {
>     startMouseCoordinateX = currentMouseCoordinateX;
>     startMouseCoordinateY = currentMouseCoordinateY;
>     if (figureType == "Free") {
>         figureFreeDraw();
>     } else if (figureType == "Square") {
>         figureSquare();
>     }
> }
> ```
> 
> Asking if removal of all references to figureType "Square" should be removed in this issue or if that is a separate issue.

Removed this in new commit

a17johpe:
> My verdict is that if you remove the function, you are in charge of removing all reference points from the code as well. I don't want a merge that adds potential errors so it should be cleaned before I approve.

All references to the function has been removed, it is never called upon anywhere in the code. There shouldn't be any potential errors since the button (the "Draw Square" button) that changes the figureType to "square" has been removed so any if-statements that checks if this is true should never have its code executed for that reason. 

I agree though that we might as well remove all references to the Free Squares in this branch because it is just unnecessary code that will never be used. It's also safer in case there is anything we have missed.

PR #7127, opened 2019-05-15
Fixed height and removed scrollbar for dropdowns in access and result

### LINKED ISSUE #7095: AccessED: Redundant scrollbars on hover-menu.
weird2222
It seems like the scrollbars could be easily removed by resizing components.

a17joaap:
Fixes #7095 
Made dropdown height auto and removed the scrollbar.
Does not seem to break anything in resulted.


a17joaap:
Nice catch! Fixed it now.

Issue #6905, opened 2019-05-10
Drawing lines between attributes should be allowed


a14oscna:
Currently you get an error message that the attributes are not composite an the user has to set one attribute as composite before lines can be drawn between them. This process should happen automatically when lines are drawn between attributes using the following logic:

If one of the attributes is already connected to an entity, this attribute should become the composite one.

If no attribute is connected to an entity, no attribute becomes composite. **However** as soon as on of these attributes are connected to an entity, that attribute becomes composite by default. 

This works in theory since nobody would ever finish a diagram with attributes that have no connection to an entity.

a14oscna:
After the customer meeting held 17/5-2019 we decided to not restrict the draw line to any logic but rather allow the user to decide how they wish to draw their lines.

PR #7142, opened 2019-05-15
Fix for issue #6905: Drawing lines between attributes should be allowed 

a17jonro:
#6905


a17drama:
New requirements from @HGustavs about this after the customer meeting. We should scrap this and make a new issue with the new requirements. @a14oscna mind fixing this?

a17vikab:
Issue reset.

Issue #7075, opened 2019-05-15
Remove space above toolbar


a17jacsv:
**How it looks after merger:**
<img width="71" alt="Skärmavbild 2019-05-15 kl  13 20 48" src="https://user-images.githubusercontent.com/37792795/57771869-4c05a380-7714-11e9-8f5f-992080c5ce10.png">

**How it should look:**
<img width="70" alt="Skärmavbild 2019-05-15 kl  13 20 37" src="https://user-images.githubusercontent.com/37792795/57771882-56c03880-7714-11e9-9a3e-850efc953b57.png">

a17johpe:
There are more things broken with the toolbar styling. The toolbar overlaps with the canvas as well.
Group A branch:
![brokenstyling1](https://user-images.githubusercontent.com/37792313/57772073-d1895380-7714-11e9-87f7-324feb6110c3.PNG)
Test merge branch:
![brokenstyling2](https://user-images.githubusercontent.com/37792313/57772089-db12bb80-7714-11e9-913a-428865b3610c.PNG)

It seems to be that the border has returned?

a17jonro:
It looks like the old toolbar styling is back in some ways? Since the purple bar at the top of the toolbar (which shouldn't be there at all) in the new merge branch looks draggable as the toolbar was before and because of the border, as previously mentioned.

a16tobfr:
This is already fixed

a17vikab:
> This is already fixed no need to reassign

In that case, I am closing this issue.

PR #7148, opened 2019-05-15
fixed vertical toolbar #7075

a16tobfr:
#7075
Removed the application header from diagram.php and style.css since it is not used anymore.

Removed inline styling since it was redundant code and was not needed.

Fixed the position of the toolbar
Reduced toolbar width size
removed toolbar borders


a16tobfr:
Fixed
center

Issue #7157, opened 2019-05-16
locked objects not copied


a16tobfr:
Needs to copy some attributes from the original object to get this to work.
![copyLocked](https://user-images.githubusercontent.com/49199690/57840094-3c479700-77c8-11e9-9caa-be79c376d429.JPG)
related to #7121

Issue #7149, opened 2019-05-15
copy of text does not work


a16tobfr:
The problem is most likely solved by setting the text of the copied text object to some default value
![textClone](https://user-images.githubusercontent.com/49199690/57795145-89365980-7745-11e9-8f77-7402ccffcda1.JPG)
related to: #7121

PR #7166, opened 2019-05-16
Update diagram.js #7157

a16tobfr:
#7157
Added symbol attributes from the original to the copy. Fixes the issue where locked objects could not be copied.


a16tobfr:
Agree that the naming of copies is bad. However, this is fixed in #7149

Issue #7141, opened 2019-05-15
Rework the object labeling function


a14oscna:
The logic for generating labels (e.g. Entity0, Relation2) is different when creating new objects versus copying existing ones. This allows for duplicate names due to a loophole in the counter. The problem can be replicated if you create 3 of the same object, delete the first one and copy the last one. 

We want to have a single piece of code that works for both creating new and copying existing objects when generating their labels. Moreover there should be spacing between the word and the number in the label.

HGustavs:
How about giving it the counter from the point array? 

This array does not allow deletions and as long as each entity has at least one point this should work?

Only disadvantage is that a new entity may be called entity421

a17drama:
This one is related to #7119 and will be fixed there. We will do some pair programming on this.

PR #7197, opened 2019-05-16
Added serialnumbers for labeling objects, Fix for issue #7141

b17robev:
Fix for issue #7141

Now serialnumbers are updated and reset properly

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -2035,11 +2035,11 @@ function loadDiagram() {
     var localHexHash = localStorage.getItem('localhash');
     var diagramToString = "";
     var hash = 0;
-    for(let i = 0; i < diagram.length; i++) {
+    for(var i = 0; i < diagram.length; i++) {
         diagramToString += JSON.stringify(diagram[i]);
     }
     if (diagram.length != 0) {
-        for (let i = 0; i < diagramToString.length; i++) {
+        for (var i = 0; i < diagramToString.length; i++) {
             var char = diagramToString.charCodeAt(i);
             hash = ((hash << 5) - hash) + char;
             hash = hash & hash;         // Convert to 32bit integer
@@ -2049,7 +2049,7 @@ function loadDiagram() {
     if (typeof localHexHash !== "undefined" && typeof localDiagram !== "undefined") {
         if (localHexHash != hexHash) {
             b = JSON.parse(JSON.stringify(localDiagram));
-            for (let i = 0; i < b.diagram.length; i++) {
+            for (var i = 0; i < b.diagram.length; i++) {
                 if (b.diagramNames[i] == "Symbol") {
                     b.diagram[i] = Object.assign(new Symbol, b.diagram[i]);
                 } else if (b.diagramNames[i] == "Path") {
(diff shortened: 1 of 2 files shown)


a17drama:
Are you sure you are on the right branch? GruppA2019W21#7141 Reason why I'm asking is that we can't recreate the problem that you have seen. Could you be more specific and describe how you got to that point with that error. A tep by step description of how you got that error would be best.

a16jenja:
> Are you sure you are on the right branch? GruppA2019W21#7141 Reason why I'm asking is that we can't recreate the problem that you have seen. Could you be more specific and describe how you got to that point with that error. A step by step description of how you got that error would be best.

image
Here I am on the right branch, GruppA2019W21#7141.

image
Here I click "create attribute".

image
Here I just do an ordinary left click and nothing is created.

image
After I did the left click I drag the mouse and I get this free draw thing which should only show up when holding the left click button, correct? But here I am not holding the left click button, not sure why it happens.

I let @a17jacsv try as well and he found the same problem I did, so that's really strange that you can't recreate the problem. Hope I explained it clearly.

a17drama:
We have tried again without any luck. Could you be kind and test this on the schools server instead of your local and see if you get the same error? We have tried a few of us on both local and the schools server and we can't get the same error.

a17jacsv:
Tested on the server and got the same results. See picture below.
Skärmavbild 2019-05-20 kl  09 59 08

a16jenja:
I tried the same thing on the school server now and I let @a17jacsv try as well and we both got the same problem as previous before. Maybe try and let someone else try and review and see if they also get the problem?

a17drama:
Good idea. @a17vikab could help with this. I don't want to sweep the dirt under the rug. We tried on 3 PC's and nobody got the error. You two obviously get it repeatedly so the question now is why. Could it be browser related? I already tried Chrome, Edge and Firefox, local and remote and haven't got a single issue. I don't want this merged if it creates problems for the users.

a16jenja:
> Good idea. @a17vikab could help with this. I don't want to sweep the dirt under the rug. We tried on 3 PC's and nobody got the error. You two obviously get it repeatedly so the question now is why. Could it be browser related? I already tried Chrome, Edge and Firefox, local and remote and haven't got a single issue. I don't want this merged if it creates problems for the users.

Agreed, it is really strange. Hopefully, the next reviewer might bring some clarification around this.

a16tobfr:
Unable to replicate. I am not sure how you are doing to get this error. I got no errors in the console.
<a href="https://gyazo.com/8b4362763c4a693d7e4fd698e5d8a226">Image from Gyazo</a>

a16jenja:
I found the problem why it didn't work for us, apparently, when we both clear our local storage it works with no problems. So maybe it was some kind of variable in the issue that we are working with that got saved in our local storage which caused the problem, I am really sorry about this, my mistake. I will test it for real now.

a17jonro:
I had the same problem at first. It seems that if you have attributes in the diagram that are not created with this new code you will get the error.

a17drama:
Thank you. This is a great thing to read. Great job. Do I have to do fake changes for you to approve it or can you switch the status? We still see changes requested.

a17drama:
Going back to var. let wasn't approved. Good eye there.

Issue #6394, opened 2019-04-29
Group / ungroup function


a17drama:
There should be a option to group entities if two or more entities are selected. When grouped it should be indicated by an icon (similar to the indication of locked items). Entities in the group should be considered "locked" and unable to be modified. If the user is trying to select an entity in the group the whole group should be selected and if the user tries to move the whole group should move as a chunk. There should be an option to ungroup the entities so all regular functionality is restored.

PR #7199, opened 2019-05-16
Fix for issue #6394: Group / ungroup function 

a17jonro:
#6394 

Do we really want no resize and no appearance when objects are grouped?

Then i will need to fix a few things before review is performed. Since this is an older branch where some implementation is missing, such as visual group representation for free draw objects and shortcuts for the new actions.


a17jonro:
This issue should be finished now. 
No resize and no apppearance change is left in per the issue description and since no other edit options use shortcuts i've not implemented them. An image was requested instead of the text: "Group" but the image did not work well with moving around in the canvas so i'll leave it for future implementations.

PR #7209, opened 2019-05-17
Use existing db password if it exist

### LINKED ISSUE #7061: Install should autofill SQL user password if it exists
The other fields are autofilled if coursesyspw.php is present but not password.
Password should also be filled in but as starts/dots of course.

This should speedup the install process when testing.

a17adajo:
Fixes #7061

### DIFF
--- a/install/install.php
+++ b/install/install.php
@@ -108,7 +108,7 @@
 	$dbPassword = "";
 
     $credentialsFile = "../../coursesyspw.php";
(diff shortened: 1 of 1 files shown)


a17adajo:
> Nothing wrong with the functionality but you could make some changes on the code.
> For example line 108, 151-153 are not according to code standard.

I have done changes requested

PR #7212, opened 2019-05-17
reverted ssn, fixes ladexpport - Grupp b2019 w21#6937

### LINKED ISSUE #6937: LadExport in resulted doesn't include SSN - needs redesign
image

The export function in resulted is currently incomplete, with no ssn included. This is because including it with the export's current functionality would be a breach of GDPR.

However, in order to make it fully functional again, and not be against GDPR, it would need to be redesigned.

a17pongo:
Reverted changes made from this pull request by request from henrik: 
https://github.com/HGustavs/LenaSYS/pull/6700

Fixes #6937

### DIFF
--- a/DuggaSys/resulted.js
+++ b/DuggaSys/resulted.js
@@ -128,7 +128,7 @@ function process() {
 		}
 		var student = new Array;
 		// Creates a string that displays the first <td> (the one that shows the studentname etc) and places it into an array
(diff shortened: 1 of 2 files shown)


a17pongo:
The sorting by issue is in another issue so not sure if I should fix that.

c17julka:
nooo i clicked wrong button

PR #7217, opened 2019-05-17
Clicking on file will now show a download link and/or a preview

### LINKED ISSUE #6938: FileEd: Clicking on filename causes error
When clicking on a filename in the fileed view the console shows a href null error. I assume the filenames are supposed to be clickable to show them in either a dialog or on another page.

a17adasy:
fixes #6938 
When you click on a filename a preview pop up will show up.

Test site: https://a17adasy.gruppb.webug.his.se:20002/DuggaSys/fileed.php?cid=2&coursevers=97732

### DIFF
--- a/DuggaSys/fileed.js
+++ b/DuggaSys/fileed.js
@@ -206,7 +206,7 @@ function validateForm() {
 //----------------------------------------------------------------------------
 // renderCell <- Callback function that renders a specific cell in the table
 //----------------------------------------------------------------------------
-function renderCell(col,celldata,cellid, filedata) {
+function renderCell(col,celldata,cellid) {
     var str="";
    
     if (col == "trashcan" || col == "filename" || col == "filesize" || col == "editor") {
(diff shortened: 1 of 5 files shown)


a17anter:
Unable to test on the webserver because there are no files/images etc. to "preview" and it's not possible for me to upload new ones, I only get error "No file found - check upload_max_filesize and post_max_size in php.ini" and I can't find the reason why.

b17nicsv:
Before merging I would like to hear that it works on the webserver aswell.

a17alian:
I think I have the same problem, both uploading and previewing/ downloading file works fine on local:
image

But uploading on webserver casuses error "No file found - check upload_max_filesize and post_max_size in php.ini " and preview looks like this. Downloading results in a "file cannot be found" error. 

image

b17nicsv:
This was discussed on the kundmöte. A place to start is to check apache server settings. There might be some Apache standard settings on the server which needs to be changed.

b17nicsv:
Since no changes were made until now. I would like the branch this is put on @a17adasy so we can archive it for next year instead! 🎉

b17nicsv:
https://github.com/HGustavs/LenaSYS/tree/GruppB2019W21%236938

The branch which contains the code used here.

a17adasy:
The branch with the code: https://github.com/HGustavs/LenaSYS/tree/GruppB2019W21%236938

b17nicsv:
I've set the servers post_max_size to 128M and upload_max_filesize to 64M and it didn't fix the issue. 

To change the values:

* Go to /etc/php/7.0/cli
* Sudo nano php.ini
* Sudo service apache2 restart

Issue #6733, opened 2019-05-08
Recursive relationships for ER and UML


a14oscna:
It should be possible to draw a line from an entity / class back to itself to represent recursive relationships as seen in the image below.
![recursive](https://user-images.githubusercontent.com/37793141/57363473-75a55480-7181-11e9-9cb1-8992b6bb53c8.PNG)
When drawing such lines in for example lucidcharts, the line automatically gets a fixed length outwards from the object before it turns back and connects to itself. We should strive for a similar result.

a14oscna:
Reference to #3739 where this is mentioned.

c17matjo:
Are recursive lines for ER relations actually needed? I don't remember how ER works very well but from a google search it looks like recursive relations still use the relation symbol which means they should already work without needing a special recursive line.
![bild](https://user-images.githubusercontent.com/49195031/58077119-28c37400-7bac-11e9-81c7-cea48497940e.png)

PR #7227, opened 2019-05-17
Fix for issue #6733: Add recursive relations

c17matjo:
I've implemented recursive relations to UML classes and ER entities. Recursive relations don't seem to work in exported diagrams but I think that should get it's own issue as this issue has taken a long time as it is. 

bild

#6733


c17matjo:
> Are recursive lines for ER relations actually needed? I don't remember how ER works very well but from a google search it looks like recursive relations still use the relation symbol which means they should already work without needing a special recursive line.
> bild

Copied from the issue page. I'm not sure if ER entities really need recursive relationship-lines. I'm going to focus on the overlapping issue for now

a17celal:
> > Are recursive lines for ER relations actually needed? I don't remember how ER works very well but from a google search it looks like recursive relations still use the relation symbol which means they should already work without needing a special recursive line.
> > bild
> 
> Copied from the issue page. I'm not sure if ER entities really need recursive relationship-lines. I'm going to focus on the overlapping issue for now

Hm, good question. What do you think? @a14oscna @a17vikab

a17vikab:
> > > Are recursive lines for ER relations actually needed? I don't remember how ER works very well but from a google search it looks like recursive relations still use the relation symbol which means they should already work without needing a special recursive line.
> > > bild
> > 
> > 
> > Copied from the issue page. I'm not sure if ER entities really need recursive relationship-lines. I'm going to focus on the overlapping issue for now
> 
> Hm, good question. What do you think? @a14oscna @a17vikab

Remove the ER recursive lines.

a14oscna:
Recursive lines should exist for ER and it should work just as your image shows c17matjo.

c17matjo:
@a14oscna Since recursive relationships in ER use the regular relation symbols they don't need any special support. My screenshot was made on the normal group branch. What did you mean should be done?

c17matjo:
@a14oscna ?

c17matjo:
I've adjusted the cardinality spacing now. I'm not sure what @a14oscna meant but for now I think you can re-review it.

a14oscna:
Sorry for a late reply. You are correct c17matjo, there should be a relation symbol when recursive relations are made in ER just as your image shows. Only in UML should the line able to be drawn back to the same class.

c17matjo:
Okay. Do you mean that when the user clicks and drags a new line within an ER entity the program creates a new relation symbol with 2 connections to the entity automatically? I think that should get it's own issue as it's more of a shortcut and this issue has a big scope as it is.

a14oscna:
No, you can just keep it as it is, assuming you can drag a line from an entity to a relation, and then an additional line from the relation back to the same entity as your image suggests. :)

c17matjo:
Ah, I see. Then everythings done :)
@a17celal, you cleared to review it now :+1:

a14oscna:
Perfect, will have it in before the big merge today!

PR #7242, opened 2019-05-17
Fix for issue #6905: Drawing lines between attributes should be allowed 

a17jonro:
#6905 
Instead of dealing with the hassle of having composite attributes separate to the "normal" attribute and changing to composite when drawing lines depending on which entities or attributes the attribute is already attached to, I have now entirely scrapped composite attributes and always allow them to be connected instead.


a14oscna:
This solution is exactly as requested by me. Good!

Issue #6849, opened 2019-05-10
Give the option to add Contribution dugga to a course during the install


a17adajo:
Nice for dev/testing/demo purposes as discussed in #6612

a18kaswi:
R.n it boils down to how to beautifuly structure an interactive function that mods the install-db during install.

a18kaswi:
See customers answer in #6612 

In contrast to the Diagram-Dugga, the Contribution-dugga exist. 

This means we can start with adding this to the 'Demo-Course' when that one is implemented.

1. Implement ability to add 'Demo-Course'

2. Add 'Contribution-Dugga' (LenaSYS\DuggaSys\templates\contribution) to 'Demo-Course'.

a18kaswi:
The ContributionDugga is now available in the Demo-Course.

PR #7254, opened 2019-05-20
Give option to add contribution dugga during install

a17adajo:


a17adajo:
Will be added to Datorns grunder course:

image

a17adajo:
Option during install:

image

a17adajo:
Fixes #6849

a17poner:
Looking at the discussion that was had with @HGustavs wouldn't it make more sense to have this included in the testdata? maybe add a whole extra course with enough data and all the various duggas and sections needed to test and showcase/demo all the features that are likely to be worked on during the project?

a17adajo:
I agree! a demo course should be the long term goal!

HGustavs:
Feel free to steel from the running courses on the real server for inspiration.

Any publicly visible content is CC0.

a17poner:
So would it be better then to ignore this and instead have that included in the testdata always when testdata was chosen + the extra visible content for all the other quizes and views that would fit into testdata?

Issue #6821, opened 2019-05-09
Cardinality between entity and attribute


a17drama:
![Edited](https://user-images.githubusercontent.com/46029340/57450118-ede14800-725d-11e9-95f8-1ac2beff0fda.png)
There shouldn't be a cardinality option in the dialog box between an entity and attribute. As it's now it just refuses to put in an cardinality but you can still do the clicks and try.

PR #7267, opened 2019-05-20
Fix for #6821: Remove cardinality option for non entity-relation ER-lines

c17matjo:
Previously one could set the cardinality for lines between attributes and for lines between entities and attributes. These should not have cardinalities so this PR removes the cardinality option for those cases. 

#6821


c17matjo:
Ah, missed that

Issue #7198, opened 2019-05-16
Codeviewer: Reset button does not reset box size after pressing maximize


a17joaap:
Nothing happens when I click the Reset button
![thisbutton](https://user-images.githubusercontent.com/37792198/57889661-af3b2700-7835-11e9-8f48-950549d187c8.PNG)

When I click on it, nothing happens. No error message in log, nothing. Expected behaviour would be for the box to return to normal size.

HGustavs:
Screenshot of what happens? 

Which reset button?

a17fellu:
Was this observed on our live server? It seems to be something wrong on the live server. Cannot replicate this on a local server. If on local server it could also be that you before pressing the reset button have manually changed the size by dragging a window. Reset can only reset the maximizes that have been made not reset the resize made by dragging the boxes.

a17joaap:
This is on the web server.

a17fellu:
See if you can replicate it on a local server. Because i cant.

a17fellu:
On the web-server maximize and reset works for template 1

a17andby:
Tried this on my local test environment with our weekly branch.

I can maximize but the reset button doesn't work. Don't get any error either when I press the reset button.

a17fellu:
Something have gotten broken after the weekly merge. When i run the branch where i added these functionalities there are no issues. Here is that branch:

https://github.com/HGustavs/LenaSYS/tree/GruppC2019W20_mergeconflictfix

a17robda:
Is this something that needs to be fixed or not?

a17fellu:
It has to be fixed. Possibly compare the new branch with the branch i linked to find inconsistencies with these 2 functionalities between those branches

a17andby:
Just tried GruppC2019W20_mergeconflictfix branch. Doesn't work. Same result as I wrote in https://github.com/HGustavs/LenaSYS/issues/7198#issuecomment-493423307

a17joaap:
@a17fellu I tried running that branch, getting the same issues.

a17fellu:
This is super weird because i am running that branch with 0 errors works like a charm xD. This is really strange.

a17andby:
@a17joaap What url did you do your testning? I want to see if I get the same error on your specific url. You could pm me on discord if you don't wanna post it here

a17joaap:
@a17fellu Even with template 3 or 8? Those are the templates calling the function that doesn't exist.
[...]

PR #7274, opened 2019-05-20
Codeviewer: Fixed bug where reset wouldn't work on first maximize

a17robda:
Fix for #7198


a17robda:
Interesting, which version of LenaSYS and browser are you using?

a17andby:
> Interesting, which version of LenaSYS and browser are you using?

Branch: GruppC2019W21_7198
Browser: Brave (Chromium)

a17robda:
I see, it seems like it is an issue with some templates only. I will look into this.

a17robda:
Since there is a new branch for this week I will close this and make a new PR.

Issue #7234, opened 2019-05-17
Only the last drawn line should be selected.


a17vikab:
When lines are drawn, only the last drawn line should be selected. Not all the previous lines as well.

![image](https://user-images.githubusercontent.com/37792464/57926000-640f2b80-78aa-11e9-95ba-20efd5ef026b.png)

PR #7280, opened 2019-05-20
Update diagram.js

a17okaoz:
Fixed the issue where all the lines that had been drawn was selected. Now it's only the last line that is selected


a17celal:
Which issue is this? Can you link it :)

a17okaoz:
Issue #7234

Issue #7220, opened 2019-05-17
Add shortcut for "Reset view to origin"


a17johpe:
The functionality "reset view to origin" doesn't have a shortcut. This is something that is used relatively often and would benefit from having a shortcut.

a14oscna:
After discussing this matter I have decided that we set Shift + O as the shortcut command for this function.

a17okaoz:
Added the short command Shift + O next to ResetToOrigin in Edit

Issue #6297, opened 2019-04-25
Map hotkeys to dropdown functions


a17alfcl:
Create hotkeys for as many of these as possible.
Wanted to create a discussion for which are the most intuitive.
"Reset view to origin" could perhaps be mapped to "Shift+0" etc.

![hotkeys](https://user-images.githubusercontent.com/49202727/56793354-27509700-680c-11e9-9646-dd0f6bb4d35e.png)

HGustavs:
More details plz

a17drama:
I do agree with @a17alfcl here. It's very nice to have hotkeys for as much as possible. Over time people tend to use these and it boosts efficiency compared to clinking menus all the time. Only thing to consider is the collision with existing hotkeys.

HGustavs:
Two rules.

Anything that doesnt clash with browser functionality is in general fine...

and we dont have stuff that has hotkeys but not menu items (apart from enter and escape etc)

a17alfcl:
Align might be "Shift + arrowKey"

PR #7284, opened 2019-05-20
Shortcut to resetToOrigin issue #7220

a16tobfr:
Issue #7220 fixed. The shortcut for resetting the view to the origin. The shortcut is now shift + o

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -176,6 +176,7 @@ const yKey = 89;
 const windowsKey = 91;
 const num1 = 97;
 const num2 = 98;
+const oKey = 79;
 
 // Mouse clicks
 const rightMouseClick = 2;
@@ -486,6 +487,9 @@ function keyDownHandler(e) {
         } else {
             switchToolbarTo("ER");
         }
+    }
+    else if(shiftIsClicked && key == oKey){
+      resetViewToOrigin();
     }
 }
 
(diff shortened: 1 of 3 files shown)


a17jacsv:
> Issue #6297 handles all the shortcut for all the menu options, and also pull-request #7271 is already merged and done so this pull request is kind of useless. We created a new issue #7311 that fixes the label issue that happened after it got merged into the group branch and are working on it right now. @a14oscna

Pair-review-comment...

a16tobfr:
> 
so this pull request is kind of useless.
> 
This pull request adds the functionality for the hot-key which did not exist before. Therefore, it's not useless.

a16jenja:
> `so this pull request is kind of useless.`
> 
> This pull request adds the functionality for the hot-key which did not exist before. Therefore, it's not useless.

image
This is from the group branch right now, reset view to origin is bound to Shift + 0 in this case since another option got Shift + O. And it works as it should, the functionality is implemented correctly and everything. So not sure what you mean. And as I said previously the issue #7311 is handling the placement of the label right now, since it got misplaced after the merge.

a14oscna:
I deem this solution relevant because of the code structure which has been proven useful. I will create a big write-up with more detail and i am sorry if it will cause a lot of confusion, but I should not have merged PR #7271 without further requirements, as there has been a lot of conflicts in interest after the merge. Since #7311 is similar to this, but adds the "alt modifier" which we do not really need, I have decided to close that PR#7316.

Issue #7139, opened 2019-05-15
Remove the hover effect from corner points on locked objects. 


a14oscna:
There is currently a hover effect on the corner points of locked objects. This should be removed as it falsely indicates that the user can resize the locked object.

PR #7285, opened 2019-05-20
Fix for #7139 Wrong cursor when hovering point on locked object

a17drama:
#7139 Fixed this cursor issue for that I'm pretty sure got fixed way back. Could be nice if we don't overwrite code in the future.


a16tobfr:
The code looks good and the solutions seem to work well. The cursor is now in default mode when hovering the corner points of locked objects, well done.

Issue #6874, opened 2019-05-10
Moving objects with keyboard moves the camera


a17drama:
Camera should be static when moving an object. As it is now both the camera and the object is moved at the same time and it's very confusing, especially when zoomed out. Zoomed in all the way makes the object and camera move in proportion.

PR #7286, opened 2019-05-20
Update diagram.js

a17okaoz:
Fixed so that the arrow keys will not move the canvas when objects are supposed to be moved. Camera is now only moved with the arrow keys when you are in the MoveAround mode.


a17celal:
Which issue is this? Can you link it :)

a17okaoz:
Issue #6874

Issue #7068, opened 2019-05-15
Clicking FAB button in sectioned results in error


c17julka:
![image](https://user-images.githubusercontent.com/37792321/57766803-54f07800-7708-11e9-9b32-e3725db720cd.png)
The fab btn is currently non-functional, only this error appears.

a17andby:
As I mentioned in my issue (closed) the error also triggers when you are clicking on this button: 

![Screenshot from 2019-05-15 13-49-51](https://user-images.githubusercontent.com/49235724/57773455-588bfb00-7718-11e9-8fc0-6003a982d998.png)

a17poner:
This issue seems to have been caused by this pr #6799 if its of any help to find a working solution :)

PR #7288, opened 2019-05-20
Clicking Fab button in sectioned w21#7068

c17johha:
#7068


b17timjo:
I see some commits under the wrong account, closing to make way for a revised version.

Issue #7269, opened 2019-05-20
Highlight on HTML tags


a17andby:
Since they were some question marks (https://github.com/HGustavs/LenaSYS/issues/2146) on which HTML tags that work or not when it comes to highlight I have made a summary on what HTML tags that works or not works. 

All HTML tags in the list are from: [w3schools.com](https://www.w3schools.com/tags/). I did choose the HTML tags that followed these criteria:
-  Have an opening and closing tag
- Supported in HTML5

**How it is suppose to work**
When you hover HTML tags in CodeViewer the tag should highlight with purple. See example below:
![Screenshot from 2019-05-20 12-23-35](https://user-images.githubusercontent.com/49235724/58014936-4cc67d00-7afa-11e9-8f0d-802f7d10357b.png)

| Opening tag  | Closing tag   | Highlight works?   |
| ------------ | ------------- | ------------------ |
| `<!--         | -->`           | :x:                |
| `<a>`          | `</a>`          | :heavy_check_mark: |
| `<abbr>`       | `</abbr>`       | :heavy_check_mark: |
| `<address>`    | `</address>`    | :heavy_check_mark: |
| `<article>`    | `</article>`    | :x:                |
| `<aside>`      | `</aside>`      | :x:                |
| `<audio>`      | `</audio>`      | :x:                |
| `<b>`          | `</b>`          | :heavy_check_mark: |
| `<bdi>`        | `</bdi>`        | :x:                |
| `<bdo>`        | `</bdo>`        | :heavy_check_mark: |
| `<blockquote>` | `</blockquote>` | :heavy_check_mark: |
| `<body>`       | `</body>`       | :heavy_check_mark: |
| `<button>`     | `</button>`     | :heavy_check_mark: |
| `<canvas>`     | `</canvas>`     | :heavy_check_mark: |
| `<caption>`    | `</caption>`    | :heavy_check_mark: |
| `<cite>`       | `</cite>`       | :heavy_check_mark: |
| `<code>`       | `</code>`       | :heavy_check_mark: |
| `<colgroup>`   | `</colgroup>`   | :heavy_check_mark: |
| `<data>`       | `</data>`       | :x:                |
| `<datalist>`   | `</datalist>`   | :x:                |
| `<dd>`         | `</dd>`         | :heavy_check_mark: |
| `<del>`        | `</del>`        | :heavy_check_mark: |
| `<details>`    | `</details>`    | :x:                |
[...]

PR #7302, opened 2019-05-20
Added support for the missing HTML tags listed in the issue.

a17andby:
Added support for the missing HTML tags listed in the issue https://github.com/HGustavs/LenaSYS/issues/7269


a17andby:
> All tags except for comments are now highlighted. Not sure if comments need to be highlighted?

I purposely left out comment highlight out. So I agree with you

PR #6839, opened 2019-05-09
Suggestion for how github contribution information can be displayed. Solution for #6615, 6583 and #6668 

a16joare:
Here is a suggestion on how the github contribution information can be displayed.  
With this solution it is easy to get an overview how well a student has performed this week. If you hover a icon a text will show up and tell you that icon means and tell you to click to see more information. 
If the icon is clicked the list of contributions will appear. 

image

PR #7303, opened 2019-05-20
Suggestion for github contribution appearance

a16joare:
This is a new suggestion for issue discussed in #6839. 

This soultion has scrapped the icons and continued with the more text based approach. 
From start when the page is first loaded it checks if a key in localstorage exists and either load a defualt value which will keep every contribution tab open or load from the already existing values from localstorage. This will provide the possibility to choose if you want to see all the content or hide some of them. 

Here is a image when you first load the page (with no existing localstorage key):
image

Closed some of the tabs by clicking on the text "Made 3 commit(s)" etc.. :
image

The solution supports the possibility to choose which div/label to keep open. This gives the opportunity to hide everything, keep some tabs for a specific week oppend or keep all tabs of a specific kind open.
image

Here is only the issues tab shown:
image

Everything closed: 
image

Any thoughts on this @HGustavs ?

### DIFF
--- a/DuggaSys/contribution.js
+++ b/DuggaSys/contribution.js
@@ -179,7 +179,7 @@ function renderBarDiagram(data) {
 
   // Renders the diagram
 
-  var str = "<div class='group1' style='width:100%;overflow-x:scroll; margin-bottom:15px;'>";
+  var str = "<div class='group1' style='width:100%;overflow-x:scroll;'>";
   str += "<svg  class='chart fumho'  style='background-color:#efefef;' width='1300' height='250' aria-labelledby='title desc' role='img'>";
   for (var i = 0; i < numOfWeeks; i++) {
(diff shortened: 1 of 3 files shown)


a17maxsj:
I like the concept a lot. Using local storage is a great idea! The only issue i have right now is that all the links are broken. I think they are trying to be loaded within the frame. Making the link open a new tab should fix this.  target="_blank"

https://gyazo.com/35882c1cc6fbc629228057c5ae69ba9b

HGustavs:
AFAIK this seems very close to what we want.

Very useful for teachers, but students that want to hide some of their activities, can do so.

a16joare:
Fixed the linking problem that @a17maxsj pointed out.. Also fixed the problem that made the tabs close when the links were clicked..

a17maxsj:
Works great now!

a17maxsj:
Also change the color scheme to something like this? i think it looks a lot nicer!
image

a17poner:
Yeah agree, some extra buttons and functionality for this can make some new seperate issues :)

Issue #7319, opened 2019-05-21
Fix UML line logic


a17johpe:
The way the middle point between start and end of a UML line is calculated can be rewritten to be more stable.

PR #7323, opened 2019-05-21
Fix for #7319: UML line code

a17johpe:
https://github.com/HGustavs/LenaSYS/issues/7319
Fixed so that calculation of middleBreakPointX and middleBreakPointY is correct. Turns out I didn't correctly understand the use of Math.abs() when I first wrote the code.
Fixes this issue that happens when moving a class off screen on the Y axis.
umlbreakpoint


a17johpe:
Can you further specify what the problem is? I don't have any problems with drawing any lines.

Issue #7305, opened 2019-05-21
New user rows in access does not have a centered reset PW button


a17joaap:
In Access:
![newuseraccess](https://user-images.githubusercontent.com/37792198/58073182-3a078300-7ba2-11e9-89ec-6c1524fb9fed.PNG)

Probably because the .submit-button class does not have the same CSS as .resetpw-button

PR #7331, opened 2019-05-21
Fixed the button that appears more to the left.

a16joare:
Fix for issue #7305.

This need to be tested on the test-server. The purple button does not appear on my local server...

### DIFF
--- a/DuggaSys/accessed.js
+++ b/DuggaSys/accessed.js
@@ -140,7 +140,7 @@ function addSingleUser() {
 	newUser.push($("#addCid").val());
 	newUser.push($("#addTerm").val());
 	newUser.push($("#addPid").val());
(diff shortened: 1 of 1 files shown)


DuggaSys/accessed.js
```diff
@@ -293,8 +293,9 @@ function renderCell(col, celldata, cellid) {
 	} else if (col == "access") {
 		str = "<select onchange='changeOpt(event)' id='" + col + "_" + obj.uid + "'>" + makeoptions(obj.access, ["Teacher", "Student"], ["W", "R"]) + "</select>";
 	} else if (col == "requestedpasswordchange") {
+		console.log();
```
a17poner: This line should be removed

PR #7339, opened 2019-05-21
Made important word highlighting work with strings

### LINKED ISSUE #1818: The important word highlighting does not work with strings
If for instance I want to highlight a specific identifier e.g. "input1" it does not work neither if one writes "input1" or if one writes input1. Either and/or both should work. 
It would be of great benefit to be able to show how a certain piece of javascript relates to corresponding html using the identifier.

a17joaap:
Fixes #1818 

To test this you need a file containing strings. Upload a file with fileed, HTML for example, and edit some words to have double and single quotes.
codequotes

Next, go to Example settings and add those words to the 'Important words'. This should work with all of these cases:

- "word" highlights "word"
- "word" highlights  'word'
- 'word' highlights "word"
- 'word' highlights 'word'
- word highlights "word"
- word highlights 'word'

### DIFF
--- a/DuggaSys/codeviewer.js
+++ b/DuggaSys/codeviewer.js
@@ -986,8 +986,11 @@ function issetDrop(dname) {
 //----------------------------------------------------------
 
 function highlightKeyword(kw) {
+	kwDoubleQuotes = '"'+kw+'"';
+	kwSingleQuote = "'"+kw+"'";
+
 	$(".impword").each(function () {
-		if (this.innerHTML == kw) {
+		if (this.innerHTML == kw || this.innerHTML == kwDoubleQuotes || this.innerHTML == kwSingleQuote) {
 			$(this).addClass("imphi");
 		}
 	});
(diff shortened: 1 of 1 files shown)


a17andby:
It is supposed to look like this?
Screenshot from 2019-05-21 14-16-54

We have two different highlight on the important words.

With the following settings
Screenshot from 2019-05-21 14-18-29

a17joaap:
Hmm no, both important should have the same highlight as string does. Will take a further look.

a17joaap:
Vad kör du för webbläsare? Kan det ha något att göra med hur regex tolkas i olika webbläsare?

```
var withoutQuote = tokenvalue.replace(/(?:\"|')/g, "");
var withQuote = tokenvalue.replace(/(?:\"|')/g, "&quot;");
var withSingleQuote = tokenvalue.replace(/(?:\"|')/g, "\'");
```

Funkade för mig i Chrome & Firefox

a17maxsj:
seems to be working for me 🤔 
image

a17poner:
Looking at the date this was merged, this seems to have been merged into the previous weeks branch, so not sure if this is included in the current one sicne it missed the big merge in that case, I'll put someone to investigate and see if it was merged to the wrong branch and if so a new pr with a fast forwarded version of this branch could be made.

PR #7341, opened 2019-05-21
added brackets to avoid errors when checklogin = false

a17poner:


a17poner:
should be possible to be logged out and go into codeviewer

a17maxsj:
Did not get an error at all on webserver, is this limited to local instances?

a17poner:
May be related to php error settings? does the webserver react to undefined variable errors thrown or will it be silent about them?

a17maxsj:
@HGustavs

Issue #7240, opened 2019-05-17
Resulted - Sorting radion buttons. A more responsive method


c17johha:
![image](https://user-images.githubusercontent.com/49199825/57927319-ed742d00-78ad-11e9-89e8-3ed75e617c0e.png)

When sorting with the radio buttons, one must click the button "sort" to apply the sorting.

Would it be satisfying that the sorting applies when a valid combination of sorting is checked?

e.g:

![image](https://user-images.githubusercontent.com/49199825/57927142-92dad100-78ad-11e9-8bc3-387676bf3ca8.png)

c17julka:
Or a standard setting of either asc/desc sorting could be selected. Then sort it automatically on click.

c17johha:
@c17julka I'm not sure what you mean? Is it not a standard setting of asc desc that you choose?

c17julka:
@c17johha I mean having one of the asc/desc radio buttons "checked" as a standard. So when you go into resulted, you immediately have 1 of them checked already.

a17oscno:
@b17timjo @c17johha 

This issue can be closed because when you move your mouse outside the sort window it will automaticly "press" the sort button so that the changes are made. No need to press it if you don't want to.

c17johha:
@HGustavs What technique is prefered? Sorting that updates as soon as a radio button is clicked, or update when all desired options been choosed and user clicks outside the sorting table?

c17julka:
Sorting when mouse leaves seems to be pretty inconvenient. I'm no Henke, but I'd say on click would be much more preferred

HGustavs:
Since it is a radio button with only one item active at once leaving on click seems the most applicable solution in this case...

Column filters have the opposite way of functioning.... we can click multiple columns to show, and we do not want it to submit until finished with clicking all the desired fields.

In this case it seems like we are over-standardizing between these two views?

b17nicsv:
The column filter are checkboxes and not radio buttons though. I'd say the best course of action is, as you suggested, when leaving it will automatically press sort.

c17johha:
@b17nicsv I believe there are some miss understanding here.

Question is if the radio buttons should sort when a valid combination is marked
![image](https://user-images.githubusercontent.com/49199825/58024907-8c4d9300-7b13-11e9-8283-3e8dd3549052.png)

[...]

PR #7342, opened 2019-05-21
Sorting radion buttons. A more responsive #7240

c17johha:
#7240


a17ludca:
Also, there seems to be conflicts that need to be resolved.

b17nicsv:
Is removing leaves function intended?

c17johha:
@b17nicsv I have removed the sorting functionality in leaves() since it's no longer necessary. leaves() only closes the sorting window.

Issue #7361, opened 2019-05-22
Cardinality is on the wrong side once again


a17alfcl:
I solved this about seven weeks ago but it seems to have reappeared... 

![cardinality](https://user-images.githubusercontent.com/49202727/58155128-94244900-7c73-11e9-8d5e-5d80014713ac.png)

a17alfcl:
This issue should be solved by storing the id of the connected objects for the line, so that the position of the object that the cardinality symbol should adapt to doesn't have to be calculated each time. It should simply be retrieved from the object's corner() function.

a18petha:
this issue can't be solved before #8236. Currently we can't know cardinality object iscorrectside value without this value implementing a fix here would be redundant. without iscorrectside value any solution done here would probably need to be redesign.

PR #7365, opened 2019-05-22
Functionality back in place, Fix for issue #7361

b17robev:
Fix for issue #7361

Cardinality now only checks which of the line's two connected objects should be used to calculate position.


a17vikab:
We are going to remake this issue.

Issue #6991, opened 2019-05-13
Pressing display all tools under the view tab (while it is disabled) forces you out of the menu


a16tobfr:
<img width="226" alt="displayAllTools" src="https://user-images.githubusercontent.com/49199690/57636009-9ec24c80-75a8-11e9-8cd5-809dd872f66f.png">
Pressing this while its grayed out (disabled) should not give any response. Right now, when you press it, the button closes the menu. This should not happen since it is not a 'close menue' button.

Fix so that the button is not responsive while it is being disabled.

a17drama:
Reopening because of a hick up while doing a merge.

a14oscna:
Yes, good idea.

PR #7371, opened 2019-05-22
Fix for #6991 Restoring merge hick up

a17drama:
#6991 Do not pass go. Do no collect $400. This one is tested already and works tested by @a17johpe


a14oscna:
I will approve without review as it is identical to an earlier fix that was damaged in a conflict fix.

Issue #7368, opened 2019-05-22
Remove duplicates of mouse event functions


a17alfcl:
There should only be one function for each mouse event

a17vikab:
solved.

PR #7372, opened 2019-05-22
Fixed mouse events again, Fix for issue #7368

b17robev:
Fix for issue #7368

Solved once again so should not cause problems


a14oscna:
Because there seems to be an error with github, the required code was already in the system and this solution is not needed.

PR #7373, opened 2019-05-22
Additional fix for #6991

a17drama:
#6991 Additional code for @a14oscna


a14oscna:
Github was spooky on the previous pull request, but again I will marge this small fix in to complete the solution.

Issue #7017, opened 2019-05-14
Make cordinates-bar scale in developer mode


a17jacsv:
Make the **coordinates-bar** scale when minimizing the window to make it available to work at smaller windows, screens or devices in developer mode. It should always be visible regardless of the size of the window/screen.

<img width="503" alt="Skärmavbild 2019-05-14 kl  11 02 00" src="https://user-images.githubusercontent.com/37792795/57688396-9a943e80-763d-11e9-8b93-3bdfe5a49a6c.png">

Related to: #7015

HGustavs:
I would write "hide" in title of issue to clarify

Issue #7015, opened 2019-05-14
Scale zoom-bar when minimizing window


a17jacsv:
Make the **zoom-bar** when minimizing the window to make it available to work at smaller windows, screens or devices. It should always be visible regardless of the size of the window/screen.

<img width="503" alt="Skärmavbild 2019-05-14 kl  11 02 00" src="https://user-images.githubusercontent.com/37792795/57688396-9a943e80-763d-11e9-8b93-3bdfe5a49a6c.png">

Related to: #7017

HGustavs:
I view this as two separate issues, the coordinate bar and the zoom bar.

I think it is safe to only show coordinate bar in debug mode?

Do normal users really need the coordinates bar?

a14oscna:
You are right, I do not think the standard user benefits from the coordinate bar. We will write an issue to move it into debug mode.

PR #7396, opened 2019-05-22
Fix for: #7017  & #7015 

a17jacsv:
**Co-Author: @a16jenja** 
Issue: #7015 Scale zoom-bar when minimizing the window
Issue: #7017 Make coordinates-bar scale in developer mode.

Removed coord-bar from user-mode.
Added media queries to zoombar & coord-bar (dev).


a16jenja:
Because of the function of the reWrite function, when resizing the window you have to do a mouse move event for it to update and resize the coord-bar and zoom-bar.

PR #7403, opened 2019-05-22
Improved mobile functionality on resulted

### LINKED ISSUE #7378: media query should change earlier on resulted
Since a new dropdown was added in the merge the layout should change at ~810px instead of 670px. Right now the new "examinator" dropdown will not be seen if the page has a width of over 670px and under 810px.

image

c17julka:
Fixes #7378

Also changed some inline styling to external


a17oscno:
@c17julka  Could you provide a link for the testing?

PR #7404, opened 2019-05-22
commented code for deleting files because functionality not working yet

a17pongo:
Fixes #6953 although a true solution with working delete functionality, some more work is needed(issues have been created) and until then, the functionality is commented so that to prevent potential mishaps.

Could comment or hide the trashcan also if preferred alt. change the message the onclick event fired.


b17nicsv:
Is it necessary to comment this out? Currently it simply doesn't work and don't cause any mishaps. I think we instead should find a way to make the code to work.

b17nicsv:
Closing this. I don't think it is necessary to or see any benefit to comment out the code.

PR #7407, opened 2019-05-22
Made searchbar compatible with examinatorfiltering

### LINKED ISSUE #7376: Searching in resulted no longer works
Leaves no error in console.

a17ludca:
Fixes #7376 where the searchbar broke, this should work now.

Testbranch: GruppB2019W22#7376


a17ludca:
@a17oscno I dont think that searching on teachers was intended as you only have a certain amount of examinators in each course, so its easy to just choose in the dropdown. As for the code, I had to move the searchblock into the if-statements so that it wouldn't bypass that block.

a17oscno:
@a17ludca Then i see no reason to not merge this.

Issue #7181, opened 2019-05-16
Add UUID to symbols


a16tobfr:
Generate the UUID once and set it to a const in order to make it stick to the object.

This will make it easier to ensure that object connections are valid. 

**EXAMPLE__START**
E.g of a situation where UUID may help: 
This will make it possible to make sure that attributes can not be connected in a circle, which is not accurate according to how ER diagrams are supposed to work. 
![circle](https://user-images.githubusercontent.com/49199690/57851423-d4ea1100-77e0-11e9-864b-761abd05bd28.JPG)
<img width="517" alt="noCircle" src="https://user-images.githubusercontent.com/49199690/57851431-d9aec500-77e0-11e9-87e1-a217d9162186.png">

Names can change but the UUID will not change so this is why it is needed.
**EXAMPLE__END**

Should be done to Attributes, Entities, Relations, and UML classes.

Code example:
~~~
function generateUUID() {
  return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function(c) {
    var r = Math.random() * 16 | 0, v = c == 'x' ? r : (r & 0x3 | 0x8);
    return v.toString(16);
  });
}

console.log(generateUUID());
~~~

HGustavs:
Not a bad idea.

Feel free to add this but... make the id shorter... if we use letters instead of hex, we reach towards trillions of combinations already at 8 characters....

We are trying to make the representation as compact as possible.

a17jacsv:
This issue should be closed? @a14oscna

a17vikab:
Solved

PR #7412, opened 2019-05-22
Global object id added, Fix for issue #7181

b17robev:
Fix for issue #7181

Global object id is saved to localStorage and retrieved in correct order.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -50,6 +50,8 @@ var settings = {
     },
 };
 
+var globalObjectID = 0;       
+
 const kind = {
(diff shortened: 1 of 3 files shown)


a14oscna:
Merge has been reverted as it totally breaks the diagram.

a16jenja:
> Merge has been reverted as it totally breaks the diagram.

I think it breaks if you don't clear all of the local storage you have stored, try clearing it and it will work.

a14oscna:
I was initially mistaken, Local Storage had to be cleared for every user after this solution had been merged.

Issue #7414, opened 2019-05-22
Add a total amount row in the contribution personalRankTable.


a17robes:
![image](https://user-images.githubusercontent.com/38004435/58180012-9e613a00-7ca9-11e9-9a78-72cb8b354c51.png)

As you can see the table foot is not used. For example we could add the total amount of the "Number" column and the avarage value of the "Ranking" and "Group Ranking" columns.

PR #7415, opened 2019-05-22
Added sum to personal Rank

a17poner:
Working with issue #7414
rank and groupRank is average of the ranks. so sum divided by 5 (hardcoded since it should always be 5). However sortaBleTable have some floating point issues and could use some extra rounding and fixing.

image

There is a rounding error but we are fairly certain that is because of floating points and would need to be dealt with in sortableTable, making some sort of param to set decimals and having it round in sortableTable instead of in the makeSum function.

Co-Authored-By: Robin Eskilsson <a17robes@users.noreply.github.com>

### DIFF
--- a/DuggaSys/contribution.js
+++ b/DuggaSys/contribution.js
@@ -1170,18 +1170,18 @@ function makeSumPersonalRank(col,value,row){
     }
   } else if(col == "rank") {
     if (value.rank == "UNK" || value.rank == "NOT FOUND") {
-      var retVal = ((Math.floor(retdata.amountInCourse*100))/100);
+      var retVal = retdata.amountInCourse/5;
       return retVal;
     } else {
-      var retVal = ((Math.floor(value.rank*100))/100)/5;
+      var retVal = value.rank/5;
       return retVal;
     }
   } else if(col == "grpranking") {
     if (value.rank == "UNK" || value.rank == "NOT FOUND") {
-      var retVal = ((Math.floor(retdata.amountInGroups*100))/100);
+      var retVal = retdata.amountInGroups/5;
       return retVal;
     } else {
-      var retVal = ((Math.floor(value.rank*100))/100)/5;
+      var retVal = value.rank/5;
       return retVal;
     }
   }
--- a/DuggaSys/contribution.js
+++ b/DuggaSys/contribution.js
@@ -975,6 +975,8 @@ function createRankTable(data) {
     tableElementId: "personalRankTable",
     renderCellCallback: rankRenderCell,
     renderSortOptionsCallback: renderSortOptions,
+    columnSum:["number","rank","grpranking"],
+    columnSumCallback: makeSumPersonalRank,
     columnOrder: colOrder,
     freezePaneIndex: 4,
     hasRowHighlight: false,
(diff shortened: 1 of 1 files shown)


a17poner:
@HGustavs would you mind commenting on the rounding and floating point issue here? I think we would need to change it in sortableTable but im not sure how other sums have dealt with this.

a17poner:
We had math.floor at one point but because we can't make an average without changing sortableTable (only the one under amount is actually a sum the other 2 are averages) we have tp return value/5 and if we floored that we'd at times return 0. We were thinking of putting something in that empty cell but since one is total and 2 are averages we weren't sure how to pjrase that (plus tje only way we could figure out how to fill that cell from outside sortableTable is more of a hack) we decide to leave it empty, any suggestion on what to make with the total? The hacky solution isn't the prettiest but it would work in lack of other known ways of dealing with sortableTable.

a17joaap:
Looking at it more closely, I see it's just returning the sum of the rank column.
```
} else if(col == "rank") {
    if (value.rank == "UNK" || value.rank == "NOT FOUND") {
      var amountInCourse = 22; // testing
      var retVal = amountInCourse / 5;
      return retVal;
    } else {
      var retVal = value.rank / 5;
      return retVal;
    }
}
```

I don't think we need the Math.floor(x*100)/100 deal, the code above is working as intended. Looks like you forgot to divide by 5 for the amountincourse sum also. This will calculate the average correctly, however it doesn't solve the floating point issue. To do that it looks like we need to edit sortableTable slightly:
Row 478-480
```
str += "<td style='whitespace:nowrap;'>" + Math.round(sumContent[columnOrder[columnOrderIdx]]) + "</td>";
if (columnOrderIdx < freezePaneIndex) {
        mhvstr += "<td style='whitespace:nowrap;'>" + Math.round(sumContent[columnOrder[columnOrderIdx]]) + "</td>";
```

Surrounding the sumContent in Math.round fixes the problem, but I haven't tested if it messes with something else.

a17poner:
Yeah, it was supposed to return ot divided by 5. Yeah the code stuff with rounding it using floor x100 etc was a way tp try and workaround sortableTable not using math.floor/toFixed(2) which didn't really solve it. Ideal would be to change sortableTable but as its outside of our area + used outside of lenasys i was afraid to change it. But I think having decimal as an option would be nice and then having the inputted number end up in the toFixed( X) and then each table could decide and have 0 as a default value.

a17poner:
Removed the two decimal return and made sure everything is divided by 5. I can talk with groupB and Henrik today during the thursday customer meeting about adding some rounding down to sortableTable.

a17poner:
Any idea of what we could put in the leftmost cell other than just "summary" ?

a17joaap:
Hmm, can't think of anything good. "Total / Average" but that could be confusing

a17poner:
Talked with henrik about it and we'll make two seperate issues about sortableTable one to deal with the floats and one to have a way to fill in that field, not sure if we should also implement the javascript solution you sent me in this or not though.

a17poner:
Think we'll skip adding it, can be added as a seperate issue so should be safe to re review now :)

Issue #7222, opened 2019-05-17
Change the 'Global Appearance' functionality


a16tobfr:
Right now the Global Appearance button changes the settings for all previously drawn objects. It would be better if the button instead changed the settings for objects to be drawn from that point on.

This would make issue #7203 irrelevant.

To change all the previous drawn objects, the Change Appearance button should be the way to do this since it should work for all of the selected objects and not just the lastSelectedObject as it is right now. This will be fixed in #7224.

PR #7417, opened 2019-05-22
[Fixed] Change the Global Appearance Functionality #7222

a16tobfr:
#7222 
The issue does now appear to be solved. There may be combinations that we have not thought about so be thorough when testing to try and find the edge cases where this may fail.

However, none were found before making this pull-request so hopefully, this works fine.

All of these attributes can now be used in Global Appearance to set the settings for current objects, and the settings for future objects to be drawn:
_Font family
Font color
Text size
Fill color
Stroke color
Line thickness_

UML objects are intentionally excluded


a14oscna:
We solved it together and I will not proceed to merge.

Issue #6634, opened 2019-05-06
Problem with selecting object after using right-click-move


a17jacsv:
As the title says. After you have moved the "camera" with using right click you can't see the black dotted square you're selecting. See pictures:

![Skärmavbild 2019-05-06 kl  12 28 07](https://user-images.githubusercontent.com/37792795/57220191-ab6c0100-6ffa-11e9-8c20-3429f51b6a11.png)

It works when using "space" instead of right-clicking. You should view this black dotted box:
![Skärmavbild 2019-05-06 kl  12 27 11](https://user-images.githubusercontent.com/37792795/57220199-b6bf2c80-6ffa-11e9-9ab3-26ec890d209d.png)

a17drama:
@a17jacsv What browser where you using? I can't reconstruct this issue. I moved and selected and saw the dotted box every time. What am I missing? I tried both Chrome and Edge and it works fine. Maybe it's a fruit problem?

a17jacsv:
> @a17jacsv What browser where you using? I can't reconstruct this issue. I moved and selected and saw the dotted box every time. What am I missing? I tried both Chrome and Edge and it works fine. Maybe it's a fruit problem?

Haha! I'm using Chrome and it behaves as I described above. You have to move the "camera" with right-click and then stop moving the "camera". When you have done that you can't see the dotted box. I dont believe it's only a Mac-problem...

a17drama:
@a17jacsv Tested on different machines with different browsers (Chrome, Edge, Firefox) without this issue. If we are to solve this we would need to assign it so someone that has a machine from the fruit company. This is the only thing that is different. I can't see the problem on my machine.

a17jacsv:
> @a17jacsv Tested on different machines with different browsers (Chrome, Edge, Firefox) without this issue. If we are to solve this we would need to assign it so someone that has a machine from the fruit company. This is the only thing that is different. I can't see the problem on my machine.

Thats weird. I will test it again and show you tomorrow.

a17drama:
It's definite now. Fruit problem. Caused by the mouse or the software has to be investigated.

a17drama:
Not a fruit issue. This happened on a windows as well but trying to replicate it on my PC is impossible. @c17matjo did it on his PC.

c17matjo:
To clarify, I could replicate the issue on my linux laptop

Issue #6726, opened 2019-05-08
Right click moving the diagram opens context menu


c17matjo:
When rightclick-dragging to move the canvas around the browsers rightclick context menu opens, blocking the canvas. This seems to only happen on some OS's/browsers. 
![image](https://user-images.githubusercontent.com/49195031/57357804-1f7de480-7174-11e9-896c-2bb99510ef91.png)
After rightclick-dragging a context menu is opened. Verified on my linux laptop on both firefox and chrome. 

Related to issue #6634, original rightclick-drag issue #5748

a17johpe:
It works on my laptop with Windows 10 on both Chrome and Firefox so I can't test this.

c17matjo:
It still happens in the GruppA2019W20 branch on my laptop. I've tested it on a macbook I had home and it happens there too. The issue might only show up for linux and mac.

a17johpe:
If the issue only exists on Linux and Mac, someone with either/both OS needs to be assigned to this issue.

a16jenja:
#7422 Fixed this issue.

PR #7422, opened 2019-05-23
Fixed issue #6634: Problem with selecting object after using right-click-move

a16jenja:
Issue: #6634 
Related issue: #6726 
**Co-Author: @a17jacsv**

Removed old solution and disabled contextmenu when right-clicking on the canvas. It should work on both mac and pc now as we have tested it on a different kind. Removed some unnecessary code as well.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -202,18 +202,6 @@ const leftMouseClick = 0;
 const scrollClick = 1;
 const rightMouseClick = 2;
 
-// This bool is used so the contextmenu will be hidden on mouse drag, and shown on right mouse click.
-var dragDistanceReached = true;
-
-// Hides the context menu. Needed in order to be able to right click and drag to move the camera.
-window.addEventListener('contextmenu', function (e) {
-        if (dragDistanceReached) {
-            e.preventDefault();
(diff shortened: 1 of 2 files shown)


a17jacsv:
> The right click menu is removed as intended, but right click move functionality is not working. Right clicking on my machine doesn't do anything now, whereas before it would bring up the menu and allow me to move on the canvas. I'm testing on a macbook air laptop, and i might suspect this is a fruit issue.

I'm using a MacBook and it works without any problems with both mouse and mousepad. Are you really sure it doesn't work?

a17johpe:
Since I can't see the problems for either of these issues in the first place, I am not the best person to review this problem, especially since I don't understand why the removed code has been removed.

b17albbe:
My bad. It seems that I just didn't use the drag option correctly. When right clicking, the menu doesn't get in the way. And when right clicking and holding, i can drag across the canvas correctly. So it does work on my macbook air.

b17albbe:
However, do we want to remove the right click menu for the canvas?

c17matjo:
Maybe it would be better to use the middle click button to move instead of right click. Then we could keep the right click context menu.

Issue #7411, opened 2019-05-22
All tools are available for every toolbar mode in diagram editor


a14oscna:
After recent merge, every tool is available no matter which toolbar mode you are in.

HGustavs:
I assume this has to do with the toolbar in the diagram editor?

a17drama:
Yes. The label says "Diagram". And it just got solved and is up for review.

PR #7425, opened 2019-05-23
Fix for #7411 Show correct toolbar when switching

a17drama:
#7411 This is spaghetti programming. All that surrounds the toolbar. Strong suggestion to refactor this and make it easier in the future. @a14oscna Should I write such issue?


a14oscna:
Sure, we should definitely tidy the code. Go ahead and make separate issues.

Issue #7345, opened 2019-05-21
Can't hover over sort / filter dropdown in resulted


c17julka:
![image](https://user-images.githubusercontent.com/37792321/58094423-ee200280-7bd0-11e9-81cf-ec0724a5b9c7.png)
Dropdown menu disappears too quickly when hovering over the sort / filter icons, making it impossible to reach.

Perhaps this could be solved by moving the onmouse-events to the td element, instead of the span element.

a17amama:
Both accessed and resulted use the same code for the navbar, however they seem to behave differently, even though both are affected by changes. Fixing one, ruins the other. Reason is unknown.

PR #7430, opened 2019-05-23
Grupp b2019 w22#7345

a17amama:
Issue #7345 
Changed so that the hover over navbuttons doesn't instantly disappear when moving mouse.


b17nicsv:
Status on this?

PR #7431, opened 2019-05-23
Grupp b2019 w21#7182

### LINKED ISSUE #7182: Using middle mouse button to press top menu items in sectioned.php opens wrong link
When you press using the left mouse button the correct page opens, but when you use the middle mouse button, it opens the wrong page in a new tab. Illustration:

g1q5R9s0T18

a17oscno:
Fixes #7182 

(Changed branch from last pull request)

It is now possible to use "middle mouse button" to open a new window of the different header tabs (results, tests, files access).

Made it so that the "hrefs" of the buttons are assigned on startup rather when pressing the buttons but due to not knowing if it is ok to remove old code (some old code on the buttons are irrelevant now like the onclick attribute is useless now that my code does the same thing but on startup instead when pressed). But i do not know if i should remove that code or not. If thats the case just say that in the review and i will do it.

link: https://a17oscno.gruppb.webug.his.se:20002/DuggaSys/sectioned.php?courseid=2&coursevers=97732


a17oscno:
@a17anter I have removed the code after some testing and all the functionality still works as it should.

DuggaSys/sectioned.php
```diff
@@ -88,28 +88,39 @@
...
 						<div class='access menuButton'>
-						<a id='accessBTN' href='' onclick='navigatePage(this.id, "accessed.php");' oncontextmenu='javascript:navigatePage(this.id, "accessed.php");'><input type='button' value='Access' class='submit-button' title='Give students access to the selected version'/></a>
+						<a id='accessBTN' href='' oncontextmenu='javascript:navigatePage(this.id, "accessed.php");'><input type='button' value='Access' class='submit-button' title='Give students access to the selected version'/></a>
 						</div>
 					</td>
 				</tr>
 
 			</table>
 
+			<<?php
```
b17nicsv:
```suggestion
			<?php
```

There should be only one less than character.

a17oscno:
@b17nicsv Fixade till det! kolla igen vetja.

Issue #7027, opened 2019-05-14
'Draw Text' box size is always default size


a16tobfr:
The draw text box should most likely be sized to the square drawn by the user. As of now, it is always set to a default size.
<img width="170" alt="drawingTextBoxSize" src="https://user-images.githubusercontent.com/49199690/57693020-81dd5600-7648-11e9-92bc-bc4754848121.png">
Always set to this size:
<img width="89" alt="textbox" src="https://user-images.githubusercontent.com/49199690/57693159-d41e7700-7648-11e9-98d8-6d7aa793c57f.png">

a17drama:
@a16tobfr What is the point of resizing the text box? We have no automatic wrap text so we can draw any use from custom sized text boxes. Text boxes are resized automatically to surround the text. Shouldn't this be enough?

a16tobfr:
1) If we want the indicator square to fill any functionality at all, the size that the user has decided should be the limit for when text wraps. This gives the user some additional ability to decide how the text should be represented.

2) The other alternative is to remove the indicator square entirely and simply make the text boxes appear by clicking on the canvas when that button has been selected.

Right now we have a mix of the two, kind of. The indicator square brings nothing to the functionality of the text boxes, and if we want to keep it this way, we will have to change how text boxes work. 

Otherwise, as stated previously in this comment, remove the indicator square and simple make text boxes appear with their default size when the user clicks on the canvas.

If we go for the first alternative, I suppose we will have to leave this issue for the students of the next year to implement this functionality.

@a14oscna, @a17vikab will have to decide what road to take on this one.

a14oscna:
After mutual agreement the decision is to remove the indicator box. You simply click to create a new textbox, which only adjusts as the user inputs text.

PR #7442, opened 2019-05-23
Fix for #7027 Fixed size for text boxes when creating them

a17drama:
#7027 Text boxes are created with click only. No more drag and get an illusion of creating a custom sized text boxes.


a17drama:
No it doesn't. I jut came to think of a thing that I missed. Selection box isn't showing and I know why. Missed to push that. Fixing now.

a17jacsv:
> No it doesn't. I jut came to think of a thing that I missed. Selection box isn't showing and I know why. Missed to push that. Fixing now.

Haha, I see now. Sorry about that...

a17drama:
It happens. You can check again.

PR #7447, opened 2019-05-23
Added ajax error handling for codeviewer

### LINKED ISSUE #7381: files with bad format can't be json_encoded in codeviwer and a blank is shown
In codeViewer a json_encode is done on the entire array and returned but if the files contained characters not in utf8 but in a bad format it fails silently and no prints are even displayed in the network table at all. Some sort of an error message would be helpful, something like "Things went wrong contact an administrator and show them this error " + the errorMessage from json_encode (json_last_error_msg or something like that.

a17joaap:
Fix #7381

Now throws this error:
errorthrown

### DIFF
--- a/DuggaSys/codeviewer.js
+++ b/DuggaSys/codeviewer.js
@@ -62,6 +62,13 @@ function setup() {
 	}
 }
 
+
+function returnedError(error) {
+	if (error.responseText) {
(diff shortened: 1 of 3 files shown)


a17poner:
wops missclicked, meant to merge sry! its getting merged now!

Issue #7265, opened 2019-05-20
Hide teachers in result view


HGustavs:
We want a hide teachers feature in the results view.

Sort of like a default search option, if enabled the sortabletable automatically filters for only students, no teachers in list, and conversely, only teachers.

Three states, only teachers, only students, and both.

Teachers normally use their own accounts for testing puroposes. In any other case we want them to be removed, especially for exporting to csv.

PR #7467, opened 2019-05-23
Filter for students and teachers #7265

c17johha:
#7265 
"We want a hide teachers feature in the results view.
Sort of like a default search option, if enabled the sortabletable automatically filters for only students, no teachers in list, and conversely, only teachers.
Three states, only teachers, only students, and both.
Teachers normally use their own accounts for testing puroposes. In any other case we want them to be removed, especially for exporting to csv."

Teacher filter was already implemented.
Student filter added.
All three states is possible to aquire
image


c17johha:
@a17adasy Yes that code stopped filtering teacher to work.  In your request changes, has you write that changes? Or shall I write the changes as requested?

a17adasy:
> @a17adasy Yes that code stopped filtering teacher to work. In your request changes, has you write that changes? Or shall I write the changes as requested?

I tried to put in the changes but couldnt find how you do that in github. So you need to write in my changes and commit!

c17johha:
@b17timjo @a17adasy 
Code updated, should new tester be assigned?

b17timjo:
@a17adasy: re-review and check the implementation of your suggestions.

We'll then have a third party review the PR as a whole again.

Issue #7359, opened 2019-05-21
contribution.php github rank table overlap


a17maxsj:
The final tab for each week overlaps to the next one, pls fix 😎 
![image](https://user-images.githubusercontent.com/49199993/58120994-ac11b380-7c06-11e9-93c7-291f82c5aff3.png)

PR #7468, opened 2019-05-23
PR for issue #7359

a17robes:
#7359 Added margins to last element of each week to avoid overlapping elements.


a17alevu:
Is this approved ?

b16johso:
Yes this is approved

Issue #7434, opened 2019-05-23
Code Viewer: The copy button doesnt dissapear when box is too small


a17fellu:
Now when you press maximize the copy button gets hidden. But not if you resize the box by dragging. The copy button should get hidden when the box reaches a specific width

a17fellu:
Working on this over at branch:

https://github.com/HGustavs/LenaSYS/tree/GruppC2019W22_7434

a17fellu:
> 
> 
> Working on this over at branch:
> 
> https://github.com/HGustavs/LenaSYS/tree/GruppC2019W22_7434

The functionality works on this branch with new code pages but there is a bug which breaks old code pages

a18jakme:
@d17marle test kommentar

PR #7472, opened 2019-05-23
CodeViewer: Hides Copy button when there isnt room for it

a17fellu:
The copy button will get hidden when the box reaches a certain width threshold. Also removed the old functionality added in this PR: #7416 as it is now obsolete with this pull request

Fixing issue: #7434

### DIFF
--- a/DuggaSys/codeviewer.js
+++ b/DuggaSys/codeviewer.js
@@ -914,7 +914,7 @@ $(window).resize(function () {
 	var windowHeight = $(window).height();
 	textHeight = windowHeight - 50;
 	$("#table-scroll").css("height", textHeight);
-
+	hideCopyButton();
 });
 
 document.addEventListener("drop", function (e) {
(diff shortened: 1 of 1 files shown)


a17fellu:
> https://imgur.com/a/ygVTFpy
> The functionality for the actual copy-button works but with some unwanted side effects.

Will look into it

a17fellu:
@a17robda i cannot reproduce this bug. Can you give some steps so i can try to reproduce this?

a17robda:
Strange, I am getting this bug in both Firefox and Chrome, Tried restarting the server and updating branches but still the same bug for me.

a17fellu:
I have tried multiple browsers but cannot reproduce this. Maybe we should let someone else look at it as well?

a17robda:
That would be nice I guess. Maybe there is something wrong on my side.

b16johso:
I am getting the same bug as @a17robda.

a17fellu:
I also managed to reproduce this bug on an old installation. The bug dissipated after i cleaned the cache and cookies. If that didn't help i also created a new code element on the course page. But before trying that just clean the cache to see if it 100% is a cache issue.

a17fellu:
After some testing i can conclude that it is not a cache issue. The bug appears when opening a code page that was there before this branch was installed. This is a super weird bug and i am not too sure of how to fix this

a17poner:
Any more luck replicating the issue and solving it? Would be nice if you could get the solution to run consistently for everyone so we can merge it in before wednesday :)

a17poner:
This didn't make it in time for the last merge, closing

PR #7475, opened 2019-05-24
Added tree support for codeviewer

### LINKED ISSUE #470: Support for trees in the code viewer

a17joaap:
Fixes #470 

tree

Collapses and restores blocks (only works on { } blocks). Could maybe hide the buttons until the user hovers over the row if that is something we want.

### DIFF
--- a/DuggaSys/codeviewer.js
+++ b/DuggaSys/codeviewer.js
@@ -1092,7 +1092,7 @@ function tokenize(instring, inprefix, insuffix) {
 	while (currentCharacter) { // currentCharacter == first character in each word
 		from = i;
 		if (currentCharacter <= ' ') { // White space and carriage return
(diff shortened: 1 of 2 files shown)


a17joaap:
Fixes icon centering and the issue with text in first/code in second. The third issue I can't recreate, and from what I understand of tokenizer, is caused by either syntax error in the code you are viewing or some bug in tokenizer itself.

a17poner:
@a17maxsj im unable to hit the re-reviewe button but could you re-review this? :D

a17maxsj:
looking great now!

PR #7477, opened 2019-05-24
Remove instructions about coursesyspw.php and loglena4.db

### LINKED ISSUE #6484: Update readme after changes to coursesyspw handling
After #5652 coursesyspw.php is created by the installer.
Update the readme with proper installation instructions

a17adajo:
coursesyspw.php is created by #5652
loglena4.db is created by #5789

Should now be easier for new students to get up and running with less steps to follow during the install.

Fixes #6484


a17adajo:
I think that can be said about a lot of things. 
Remember: if the instructions were not there in the first place and the installer always took care of this, we would not miss the instructions now.

In the case of an error during the install I think the installer should provide good error messages about what needs to be resolved before running the installer again.

a17maxsj:
If the installet works without fault i am willing to accept this.

Issue #7291, opened 2019-05-20
Timesheet table in Contribution should not display when no data.


a17alevu:
![image](https://user-images.githubusercontent.com/49199285/58022798-8903d880-7b0e-11e9-935e-78d9a0ea8675.png)

The table at the bottom should not be displayed when there is no data.

PR #7488, opened 2019-05-24
Contribution: Fix to only show timesheet table if there is data

a17robda:
Fix for #7291


a17robda:
Probably, will fix.

a17maxsj:
I thint this works a lot better! :D

Issue #7204, opened 2019-05-16
Change appearance button outputs an error message to the console when no object is selected


a16tobfr:
This error message should probably not be seen. The button should simply be unresponsive when lastSelectedObject is undefined and not give an error message to the console.
![changeAppearance](https://user-images.githubusercontent.com/49199690/57859853-07513980-77f4-11e9-951a-d546e1609ce5.JPG)

Edit: Also happens when multiple objects are selected.
Maybe change the change appearance so it works for all of the selected objects instead of just one?

c17matjo:
I've solved the main issue but I'm not entirely sure on how to solve the second. When multiple objects are selected, should the appearance menu still work and have the changes apply to all selected objects? What if there are different types of objects selected with different possible parameters (A line and an entity e.g)? I've looked at the code for the appearance menus and I can't see an easy way of implementing something like that. 

I'm going to make a PR for the main issue for now.

PR #7492, opened 2019-05-24
Fix for #7204: Only open "Change Appearance" menu if 1 object is selected

c17matjo:
Previously the console would print an error if the "Change Appearance" button was clicked with either no objects or more than 1 object selected.


a17vikab:
I have made a big mistake, this functionality should not be merged into the main branch. I am terribly sorry @c17matjo

PR #7501, opened 2019-05-24
Search bar in fileed has background

### LINKED ISSUE #7443: Search bar in fileed has transparent/no background
image

You also can't see the text you're typing in it:
image

b17annni:
fixes #7443.
image


a17poner:
hm I guess since someone from group B said we should and its small enough that a conflict won't be to bad we'll merge this.

Issue #7510, opened 2019-05-24
Switching mode while in developer mode breaks localStorage


a17drama:
When in developer mode and switching modes between ER and UML save state is triggered and this is done more and more often for every change. This breaks the localStorage.

PR #7530, opened 2019-05-24
Fix for #7510 Any keyPress triggered a SaveState() and it's fixed now

a17drama:
#7510 clickEnterOnDialogMenu(ev) triggered SaveState() on every key press. Instead of triggering on every key press it'snow moved to be triggered by the OK button in dialog box for this.


a14oscna:
I fixed the merge conflict with help of the team and will now merge.

PR #7533, opened 2019-05-24
Made search case insensitive, and supported åäö

### LINKED ISSUE #6225: AccessEd: Search function should accept lower case and ÅÄÖ
The search function in the access editor requires an exact match including upper case letters. It should be possible to search for 'conny' and still get results for Conny. 

It's also not possible to search for names including ÅÄÖ, such as 'Åsa'.

a17joaap:
Fixes #6225 

This only works in access but could be copied and implemented in other areas.


a16joare:
Do you think it is possible to get this solution to work for all searchfields? Or is it not possible with the current code?

a17joaap:
I think it's possible. It would mean editing the row filter callback function for every page that uses sortable table

PR #7544, opened 2019-05-27
Added display of current owner of lenasys dir

### LINKED ISSUE #6725: The installer should be able the check that the correct owner is set on the directory

a17adajo:
Fixes #6725


a17adajo:
Looks like this:

image

Issue #7401, opened 2019-05-22
code viewer button minimize


b17annni:
Perhaps a minimize to go with the maximize.
![image](https://user-images.githubusercontent.com/37792766/58173379-808cd880-7c9b-11e9-9cc5-a5f3fed9d849.png)

a17maxsj:
+1

a17alevu:
I agree! =)

a17maxsj:
However the button does not really minimize the section, it reverts to its prior state

a17maxsj:
Adding another button minimizing would probably be a better idea

b17annni:
Adding a minimize button was what I intended with issue.

HGustavs:
Good idea.

a17fellu:
Just keep in mind that when you would try to minimize a box that is maximized you need to hard code the size of every box for every template (-template9 and template10). Since we need to minimize that specific box all other boxes widths and heights needs to be recalculated. At least from my understanding.

PR #7553, opened 2019-05-27
Code viewer minimize button added 

a17maxsj:
https://github.com/HGustavs/LenaSYS/issues/7401 minimize now works for template 1 and 2. I am not sure about the rest, cant seem to get the values right. But should not be harder than adding the maximize templates and fiddling with the values


a17fellu:
Tested this out and the minimize button should be set to display: none when templateid is higher than 2. Since a button without functionality shouldn't be visible to the user

a17maxsj:
Minimize button is now hidden for templates other than 1 and 2

Issue #7500, opened 2019-05-24
Disable menus that don't meet the requirements for use


a17drama:
Menus that need a selected object or objects should not be enabled if there are no objects select. eg. Move selected front, Move selected back, Change appearance, Lock/unlock etc.

PR #7570, opened 2019-05-27
Fix for #7500: Disable menus that don't meet the requirements for use

a17jacsv:
**Co-Author: @a16jenja** 
Issue: #7500

Disable menus that don't meet the requirements for use. 
Also, prevent all menu items to collapse when clicking at them.

As we see it the menu item "Global appearance" should be able to be clicked at even when no symbols are selected. The user should be in our opinion, be able to set all the customization before starting to draw the diagram. That's why we made no changes to it. Let us know your thoughts on this @a14oscna.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -517,7 +517,8 @@ function canvasToPixels(pixelX = 0, pixelY = 0) {
 // Move selected objects to front of canvas
 //-----------------------------------------
 
-function moveToFront() {
+function moveToFront(event) {
+    event.stopPropagation();
     let front = [];
     let back = [];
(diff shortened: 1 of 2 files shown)


a17celal:
> I can't get the options to enable.
> ezgif-1-42f018f027a5
> 
> Also, disabled options can still be clicked but this shoud probably be it's own issue.

It is working for me to enable the options. However I am working on my local server, so maybe that is the reason?

a17jacsv:
It works without any problem for me both local and on the server. I can't replicate your problem @c17matjo.

a14oscna:
I agree with the decision about Global Appearance!

Issue #7569, opened 2019-05-27
Access editor, some "reset pw" buttons are purple


a17maxsj:
![image](https://user-images.githubusercontent.com/49199993/58417091-08038e80-8084-11e9-82a9-962c0c43cafe.png)

b18alegu:
Should this issue be closed? It is overlapping with issue #7722, but that issue is better specified. There is also a pull request in waiting from the previous year.

a18eripi:
Closing due to overlapping and duplicating issue #7722

PR #7579, opened 2019-05-27
Some "Reset pw" buttons are not purple anymore 

a16lisri:
Fix for: https://github.com/HGustavs/LenaSYS/issues/7569


a17joaap:
Isn't that a feature of access? If the user was added less than 24 hours ago, they have a purple button. Not sure what for though.

a17fellu:
Yeah looks like that is the case. But what would be the reason for that feature?

a16lisri:
> Isn't that a feature of access? If the user was added less than 24 hours ago, they have a purple button. Not sure what for though.

Yeah that makes sense, as I couldn't test it locally (where no new users had been added). Doesn't make sense for the button to be purple though.

a17fellu:
Maybe wait with merging this so we maybe can get some answers

a17poner:
yeah I will wait with merging this then

PR #7633, opened 2019-05-28
Added burger menu to codeviewer for mobile

### LINKED ISSUE #2300: Code viewer mobile adaptation
There used to be css that makes strange things with the code viewer if the window is thin or very narrow.
...

What I would prefer is a hotdog-menu / hamburger menu (it seems like food-related metaphors are common for that kind of button).

That would show only one of the parts of the screen.

And I would like this to be handled through the get-link, so if one presses the first menu item we would navigate to the same page again, but with "&showpane=1" added to the url.

Mockup:

screenshot 2016-05-02 11 24 21 copy

On mobile that view could conceivably be the default view.

Perhaps at the button of the dropdown, a show deskop button/icon could be an additional alternative.

This should _theoretically_ be a fairly simple change as we would not need to mess much with the template/multiple view system, and no adaptive css is necessary.

And one could navigate to an expanded view or bookmark one through the get-link.

The three alternatives in the menu should thus link back to the code viewer but with 

&showpane=1
&showpane=2
&showpane=3

This should make the code viewer very useful even on mobile.

a17joaap:
Fixes #2300 

Adds a burger menu to the nav header in codeviewer if the screen size is < 800px. Clicking this will open a menu with all the titles of the boxes as options. Clicking any of these will reload the page with a new get parameter "showPane" as requested in the issue. This will show only the requested box.

I know the navheader gets cramped when a superuser/teacher is logged in, because they have so many extra buttons, but that is something for another issue maybe.


a17joaap:
Error is caused by another PR (https://github.com/HGustavs/LenaSYS/pull/7475) and fixed with this PR https://github.com/HGustavs/LenaSYS/pull/7641

Issue #7036, opened 2019-05-14
UMLlines are selectable when mouse cursor is not hovering the line


a16tobfr:
In the image below, I am hovering the uml object named New1. The line gets highlighted and the line gets selected on mouse down.
The line should only be selected when pressing the line.
<img width="268" alt="umlLine" src="https://user-images.githubusercontent.com/49199690/57695725-4abe7300-764f-11e9-8fd3-fd8142cde5a3.png">

HGustavs:
Seems like we are checking for diagonals? 

The code should use the same code as for diagonals but with three straight lines. Not a specific function for "lines with kinks". If this code uses the general "close to line" code, it should not be able to mis-identify clicks.

a16tobfr:
I realize that my first picture was not fully illustrating where the line was given priority when selecting over other objects in the same area.
This probably gives a better idea of the area where the line is selected.
![umlLines](https://user-images.githubusercontent.com/49199690/57795610-b3d4e200-7746-11e9-9c2e-c2bf5ca19b8a.JPG)

a17johpe:
The pull request that solved this issue has not been merged due to too many major issues with UML line anchors. 

The code that solves it can be found here: https://github.com/HGustavs/LenaSYS/pull/7652/files
the two important functions _should_ be "checkForHover" and "UMLLineHover" in the file diagram_objects.js.

However, this solution only works if anchors are in place as well (which was what the pull request mainly set out to solve). See https://github.com/HGustavs/LenaSYS/issues/6577 for further documentation on that.

b18mikfr:
This issue is was popping up due to lack of functionality of the UMLLine symbol. It was treated like a regular UML symbol and thus i implemented functionality that followed what is drawn on the diagram. Took forever to implement but it is done and ready to review. Every case for each UML Line realtionship is supported. The issue of UML Lines being selectable over other symbols is present and would suggest another issue to solve that particular problem.

Issue #6577, opened 2019-05-03
Add anchors to manually move UML lines


a17johpe:
In LucidChart you can manually move UML lines and add new anchors for splitting up the lines. 
![umlanchors](https://user-images.githubusercontent.com/37792313/57128784-d0f7d100-6d94-11e9-9b30-3959488ff838.png)

With this the user can move lines that may be overlapping as seen in the image below
![umlanchorsdugga](https://user-images.githubusercontent.com/37792313/57128881-14523f80-6d95-11e9-81ec-60ccd26389a2.PNG)

a17johpe:
To clarify for if the line is broken or not: the start and end points of the line should not be movable but the line itself should

a17johpe:
Code with partial solution can be found in stale branch _GruppA2019W22#6577_. During review of pull request https://github.com/HGustavs/LenaSYS/pull/7652 we found too many major issues to be able to merge immediately. See https://github.com/HGustavs/LenaSYS/pull/7652/files for a clear view of what was added/edited/deleted.

Summary of solution:
Added arrays "anchors" and "draggablePoints" to Symbol (in diagram_objects.js) which will contain the extra points on the UML line. Anchors are points for every corner/kink on the line. Draggable points are the points which the user can click on to manipulate the line.
![anchorsumk](https://user-images.githubusercontent.com/37792313/58640021-72bd0000-82f8-11e9-9266-0fe36879ffe7.png)
Above we see the points visible for the regular user when you click on a line. The orange points are starting and end points (not anchors) while the grey point is a draggable point.

![anchorsdev](https://user-images.githubusercontent.com/37792313/58640138-b57ed800-82f8-11e9-8927-0c20753c007a.png)
Here (in dev mode where _all_ points are drawn out as crosses) we see the anchor points as well.

Almost all of the code for this is found in diagram_objects.js. Functions addAnchor and addDraggablePoint are used for creating the points. Function createDraggablePoints is used to automatically create one draggablePoint between every existing anchor.
Function drawUMLLine is what actually calculates everything and draws the line. The anchors and draggable points are created in here once and then their positions are recalculated based on the positions of the line's start and end point + if the user has/is moving on any draggable points.

![anchors](https://user-images.githubusercontent.com/37792313/58641316-435bc280-82fb-11e9-81d0-2614a65a1b69.gif)
Gif of solution

[...]

PR #7652, opened 2019-05-28
Fix for #6577: UML lines can now be moved

a17johpe:
Main issue: https://github.com/HGustavs/LenaSYS/issues/6577
Also solves: https://github.com/HGustavs/LenaSYS/issues/7036

UML lines are now created with anchors (line kinks/corners) and one "draggable point" which is used to manipulate the line. 

Hovering over just the line to select it also works now. 

There are still a few issues but nothing that majorly breaks the line:
- In other zoom views than 100%, startLineDirection and endLineDirection is default "left". This might (probably) be a problem with getConnectedObjects() and how it works with coordinates.
- Currently there is only one draggable point and you can only move the line up and down because of the difficulty of implementing the functionality all at once.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -3128,7 +3128,7 @@ function mousemoveevt(ev, t) {
                 // when in movearound mode or if the point doesn't belong to a selected object then don't display different pointer when hovering points
                 if (uimode != "MoveAround" && pointBelongsToObject) {
                     //Change cursor if you are hovering over a point and its not a line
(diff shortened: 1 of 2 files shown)


a16tobfr:
Draggable point is stuck in the same position when moving all of the objects.
<a href="https://gyazo.com/77fb01b37953b8790a00d052c49cfb7b">Image from Gyazo</a>

a17drama:
Capture 1
Capture 3
Capture 4
Capture 5
Capture

There are some funny situations that can be triggered here

a17johpe:
This pull request should be closed and the branch become stale. Documentation of code will be written in the two issues.

Issue #7710, opened 2019-06-18
Contribution: padding differs from div to div


a16lisri:
The header of "Weekly bar chart" is more towards the middle, if we compare that to the header of "Weekly line chart". They don't align with each other.

![differentpadding](https://user-images.githubusercontent.com/37792690/59676904-ea2fd200-91c8-11e9-83b5-ba4ab539cb7f.png)

a20sacka:
Put a label "on hold". Cant finish issue today as contribution.php doesnt display anything due to database not having any test values.

a20vicad:
Now with the contribution server up it was easily solved.

a20behta:
merge in to master during weekly merge

PR #7711, opened 2019-06-18
Komplettering-a16lisri-7710

a16lisri:
Fixes: https://github.com/HGustavs/LenaSYS/issues/7710. Elements in Contribution should now align with each other and overall look better. 

0f4de317d7aa1f94c60e48812d2774a4


a16silph:
@a17robhe What was the reason that this commit was closed without being merged?

a17robhe:
@a16silph Because it was developed on a branch that was current about a year ago. The compatabillity with the current version of the program is somewhat dubious. But this was done as we cleaned up about 500-700 pull requests that were developed by the previous years students

Issue #7735, opened 2019-06-25
showDugga: Timesheetprev is too small


a16lisri:
The box, where you can see your submissions, is very small. You can only see two of the timesheets and then the third one gets cut. You have to scroll a lot to see them all.

![3047a36d4fffd44da515fe8c4e2059fa](https://user-images.githubusercontent.com/37792690/60110134-d9a0ce00-976b-11e9-923a-8dcbd62089ea.png)

a18oscte:
#7735 The issue was already solved, nothing done.

b18alegu:
I'll let the assignees from this issue in Group-2-2020 be responsible for the review instead since the issue was already solved.

a18oscte:
The issue has been solved and only the necessery lines are edited.

b18alegu:
I'll move this back to ready for review. Please provide the requested reviews.

b18alegu:
This issue is solved and merged into the G2-2020-W14 branch
Issue will be closed

PR #7736, opened 2019-06-25
Komplettering-a16lisri-7735

a16lisri:
Fixes: https://github.com/HGustavs/LenaSYS/issues/7735. Changed height from 100px to 200px. This lets us see more than 2 files before having to scroll. In the solution we now see 6 files and their dates. 

c1f4edb680ad306037dba18a325e0d8c

### DIFF
--- a/DuggaSys/templates/daily-minutes.js
+++ b/DuggaSys/templates/daily-minutes.js
@@ -4,7 +4,7 @@
 
 *********************************************************************************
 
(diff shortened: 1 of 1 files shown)


b18alegu:
Since there is another pull request for this issue (https://github.com/HGustavs/LenaSYS/pull/7803) that is with correct code format, I'll close this request. Joint consultation with @a18oscte

Issue #7762, opened 2020-03-23
Inline code in markdown


a97marbr:
We want inline markdown of code. 

For example, if I want to talk about `var tomten=20;` when describing how to create and set a variable in javascript.

I would suggest backtick (`)  as start-stop character similar to GitHub, Discord and other markdown implementations.

HGustavs:
must be distinct from important word as one piece of code may be both

PR #7781, opened 2020-03-25
#7762 inline code in markdown

danieloskarsson:


a97marbr:
I'm curious as to the change of `code` from block to inline-block. Wouldn't this break old examples where a block element is expected?

Perhaps we could add a class to a code block that is inline?

```css
code {
  ...
  display:block;
  ...
}

.inline-code {
   display:block !important;
}
```
So when we generate the HTML for our code we also add the inline-class if needed. This way old example would stay the same, right?

danieloskarsson:
> I'm curious as to the change of `code` from block to inline-block. Wouldn't this break old examples where a block element is expected?

When I was testing it it looked fine. The difference is that instead of the code block being page wide it only becomes as wide as the content, which in my opinion actually looks better.

code-inline-block

a97marbr:
Yes @danieloskarsson , you are right it should be safe with respekt to old examples. The closing three back-ticks and a newline will make sure that no inline elements could be placed to the right of any code example.

I'm undecided regarding which looks best :)

danieloskarsson:
> I'm undecided regarding which looks best :)
Can I sway you by noticing that from a UX perspective `inline-block` is better than `block` because the former causes less eye movements for the user which in practice means that it is easier to read the content.

a97marbr:
Works as an argument for me.

PR #7791, opened 2020-04-01
Prevented long MOTD to be half shown after change

### LINKED ISSUE #7789: Message of the day covered
After closing the message of the day and then editing the message, it apperas behind the header, making it impossible to see the full message without reloading the page. Make it not appear after change at all or appear at the right place.
motd

a18antsv:
Fixes #7789.
Prevents message of the day from showing after administrator changed the message.
Before, the message relied on being one row as it would appear behind the header. On small screens, there is a big risk of the message being multiple rows making it show parts of it below the header.
Uses the height of element when button is clicked to force it outside the screen with the right amount of pixels.
Function returnedCourse in courseed.js will be called whenever a change is made in courseed, which will automatically show the message again even if hidden before. This is because same function is used to show the right things when site is loaded. This workaround prevents it from appearing.

### DIFF
--- a/Shared/dugga.js
+++ b/Shared/dugga.js
@@ -1594,11 +1594,13 @@ function generateTimeSheetOptions(course, moment, selected) {
 //----------------------------------------------------------------------------------
 
 function hideServerMessage() {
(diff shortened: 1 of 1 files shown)


a17robhe:
Mergeat mot W14 branchen förtestning

Issue #7785, opened 2020-03-29
Resultedconstants


HGustavs:
someone created a file resultedconstants that should not exist according to code standard....
so removing that and fixing any references to constants in that class.

b18alegu:
This issue is solved and merged into the G2-2020-W14 branch
Issue will be closed

a17pioja:
This issue's solution has been reverted due to merge conflicts during week 15
Re-opening issue

b18alegu:
Reviewers should use the linked pull request #8056

b18alegu:
This issue was merged into the W16 master. I'm closing this one now.

PR #7807, opened 2020-04-02
G2 2020 w14 #7785

c14jonfr:
Deleted resultedconstants.php and added constants from that file to resultedservice.php and any references to resultedconstants.php.


c14jonfr:
The change in install.php was made by someone else and pushed/merged to the master branch, which this issue branch was based on. At the time we believed that merge would stay since it arguably is a good change (we even believed the merge was approved by the repository owner, but it was probably made due to a mistake since it has now been reverted). I agree that commit/change should be removed. I will ask course responsible whether we should simply make the change in a new commit or if there is a better way to revert that commit without having to redo our own.

c14jonfr:
I have removed the changes that were made to install.php. All should be fine and dandy now.

Issue #7449, opened 2019-05-23
FAB button positioned too far to the right when viewing an assignment in resulted


b17carma:
![image](https://user-images.githubusercontent.com/37834592/58248968-e5106c00-7d5d-11e9-926b-d18b30607a58.png)

I think it should be moved a bit more up & left so it doesn't collide with the scrollbar

b18alegu:
@b18veran, please link the corresponding pull request to this issue

PR #7809, opened 2020-04-02
Gradebutton moved. Moved partially styling from dugga.css to resulted…

b18veran:
….php. #7449

### DIFF
--- a/DuggaSys/resulted.php
+++ b/DuggaSys/resulted.php
@@ -101,7 +101,7 @@ pdoConnect();
 			<span id="hoverRes" ></span>
 			<h3 style='width:100%;' id='Nameof' onmouseover="hoverResult();"
 			onmouseout="hideHover();" >Show Results</h3>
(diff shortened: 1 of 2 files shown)


c14jonfr:
Is there a specific reason for using inline-CSS for #gradeBtn in resulted.php instead of simply modifying the CSS in Dugga.css?

c14jonfr:
I would also advice against using percentages in this case since (in most cases) 5% vertically is not the same as 5% horizontally. Looking at how positioning is done in the rest of dugga.css pixel values are used so it would probably be prefered to stick to the standard the rest of the document uses.

Also, the change from positioning:fixed to absolute does not seem warranted to me. Since it was using fixed before I would guess it is supposed to be fixed rather than absolute (note that this is merely speculation on my part, to be 100% sure you would have to ask whoever made the design, if that is even relevant anymore).

c14jonfr:
Good change! But maybe something like 55px would be better for the right-value. See image for how the change looks like for me (using 45px for right, bottom positioning). 

image

PR #7814, opened 2020-04-02
G4 2020 w14 issue#7812

a18antsv:
Inserts the right value of read only to the database table settings based on the checkbox (0 -> unchecked, 1 -> checked). Before 0 was always inserted.
Made checkbox represent the previously made setting change by being checked (value in database 1) or unchecked (value in database 0) when settings popup is loaded.

-Test if the checkbox stays checked after page reload if saved as check and if it stays as unchecked if saved as unchecked. It is based on what is in the database so should not be a need to check in the database if it is correct there.

### DIFF
--- a/DuggaSys/courseed.js
+++ b/DuggaSys/courseed.js
@@ -9,6 +9,7 @@ var querystring = parseGet();
 var versions;
 var entries;
 var motd;
+var readonly;
 
 $(document).ready(function(){
     $('#startdate').datepicker({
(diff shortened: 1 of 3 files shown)


a17robhe:
@a18antsv Some instruction as to what needs to be tested is needed

a18antsv:
> @a18antsv Some instruction as to what needs to be tested is needed

Test if the checkbox stays checked after page reload if saved as check and if it stays as unchecked if saved as unchecked. It is based on what is in the database so should not be a need to check in the database if it is correct there.

a17robhe:
Has already been a added to the test server at http://group4.webug.his.se:20001/G4-2020-W14-ISSUE%237812/ for testing @b18simla

DuggaSys/courseed.js
```diff
@@ -237,23 +238,36 @@ function editVersion(cid, cname, ccode) {
 }
 
 function editSettings(){
```
b18simla, line 240: functions should have their opening brackets on a new line with same indentation as function declaration

b18simla: (does not apply to for-loops and if-statements, which all seems good in your code!)

PR #7817, opened 2020-04-02
G4 2020 w14 test

### LINKED ISSUE #7776: Cloning a course version should be done as 1 transaction
When creating a new course version we have the option to clone from a previous course.

Currently if we specify a course version name that already exists we get a duplicate key error for the course name but the cloning of course is still done giving us duplicate entries in the course version that generated the duplicate key error.

Current code i believe does the following:
```
Create the new course version // Could generate a duplicate key error
Make a copy of all the sectioned-items, duggas, etc // This is done regardless of 
                                                    // the success of the previous 
                                                    // create operation.
```
Perhaps create a stored procedure for the clone operation and guard the whole internals of the procedure with a `begin transaction` and a `commit`? Other ideas?

a18kaswi:
fixes #7776

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 LenaSYS
 ==============
 
+TEST
+
 University of Skövde Code Viewer and Course Organization System
 
 # Install Instructions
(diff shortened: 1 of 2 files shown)


a18kaswi:
close #7776

PR #7842, opened 2020-04-03
replaced cid with courseid as fix for issue#7806

a18felen:
This is our proposed fix to the issue#7806 where cid and courseid caused a conflict that made it unable  for filedrecived.php and fileed.js to work when uploading a file.


a18felen:
@a18lukto

Issue #7765, opened 2020-03-23
Download all files as zip


HGustavs:
We want the ability for teachers/administrators to download all files from a course.

A download all files menu option in the top toolbar.

HGustavs:
not all hand-ins but all files i.e code examples etc for active version of course.

b18phibe:
> not all hand-ins but all files i.e code examples etc for active version of course.

Are they group in any special way in the database or where can i find those files?

HGustavs:
I would test, to see where they end up by uploading a file...

but as far as I remember there is a folder "downloads" or something to that effect, either in the lenasys root or in shared (it has been in more than one place historically).

And in that folder  there is a folder for each course and inside that a folder for each version of a course.

HGustavs:
and in that folder the files are stored one by one.

a18jonpe:
There is a folder install/courses that contains two folders, one of which (1) contains the files from the course with ID 1. But other courses are missing. Uploading files does not work either, it redirects to http://localhost/LenaSYS/DuggaSys/filereceive.php but gets stuck there.

a18petha:
Cannot reproduce the test, I found the download icon for each course but it seemingly does nothing. The test has been carried out on chrome and firefox, from what I can tell none of the code that has been implemented has been fired once the user interacts with the new download icon

Following code is what I assume to errorhandlers, but I never interacted with these html messages.

![image](https://user-images.githubusercontent.com/49141878/78536471-30086600-77ee-11ea-931a-7ff811561b27.png)

a18kaswi:
Closing PR for weekly, https://github.com/HGustavs/LenaSYS/pull/7845

This stil needs to be fixed.

a18kaswi:
Oops to fast on the triggerfinger.

PR #7845, opened 2020-04-03
Have made it possible to for teatchers to download there files to a zip

b18phibe:
Now teachers can download there files for the current version of to course to a single zip-file complete with the version independent files. @a18jonpe  ISSUE #7765


a18petha:
Cannot reproduce the test, I found the download icon for each course but it seemingly does nothing. The test has been carried out on chrome and firefox, from what I can tell none of the code that has been implemented has been executed once interacting with the download icon.

Following code is what I assume to errorhandlers, but I never interacted with these html messages.
image

a18kaswi:
@b18phibe @a18jonpe take notes of what should be fixed, and close this PR. Put in a new one when changes are made based PR-comments.

b18phibe:
@HGustavs Do you have any idea why the server gets a html 500 error and does not execute the download.php file when it works on a local environment?

a18kaswi:
**ACTIVATE ZIP-ARCHIVE**
sudo apt-get install php7.2-zip
sudo /etc/init.d/apache2 restart

the above needs to be done on server where LenaSYS are installed to enable this feature.

PR #7853, opened 2020-04-06
Colors on new/oldbuttons changed

a18matna:
Re-commit for this week. If-statement Terms flipped, Button class added. Css added to new button class


c14jonfr:
It seems to me like the solution works, but could you explain how/why the colors for the new users were picked? I would argue they look a bit out of place compared to the rest of LenaSYS which keeps to a somewhat consistent theme. It is also unclear to me how these colors indicate that the user is new, as compared to the previous solution (using purple to indicate new users).

resetpwcolors

a18eripi:
The previous issue stated that it wanted a colour that shows that users are new, on the Reset PW button. It was a strange request and we settled for the connection that green usually signifies new. Maybe we should disregard this and create a flag next to it, or something else to signify that it is a new user. Also, what use of knowing if a user is new is strange to me.

PR #7854, opened 2020-04-06
Added PDO transaction to bundle the queries when using option 'CPYVRS'

a18conch:


a18kaswi:
Base was master @a18conch I changed it, this time!

Issue #7463, opened 2019-05-23
ResultED: Grading window is positioned poorly


c17alepe:
![image](https://user-images.githubusercontent.com/38323901/58254723-a8e40800-7d6b-11e9-9597-57a143126d18.png)

When you click the fist at the grading in resulted, you get the students submission. When you then click the yellow fist button, a grade window appears. This window is positioned poorly. Maybe it should be more centered and smaller?

Screenshot of the problem
![image](https://user-images.githubusercontent.com/38323901/58254799-d7fa7980-7d6b-11e9-9ca1-607f9650e1db.png)

(Sidenote, a console error appears when you open the submission window. I dont think the error is causing this issues problem, so i have created an issue for that)

b18alegu:
Issue solved last week (W15 2020). I'm now closing this issue.

PR #7866, opened 2020-04-06
Change the size of toggleGrade and moved more to center. #7463

b18veran:
Co-Authored-By: a18sebka <a18sebka@users.noreply.github.com>


b18jonli:
Seems to work, but is there a reason for the inline css? otherwise I would put all the css styling in dugga.css

PR #7869, opened 2020-04-06
solution for fileinfo extension

b18emigu:
Now checks if the `fileinfo` extension is enabled by checking if the function exists, if it doesnt it uses the file type given on upload. This stops the application from breaking on a new system without the right `php.ini` settings.


a18felen:
Because there is another issue when uploading a file with cid and courseid issue#7806 the page will give you the following message "Error updating file entriesIncorrect integer value: '' for column 'cid' at row 1" and that is expected because this push does not have the fix from the other issue.

PR #7893, opened 2020-04-06
Made it possible to edit files from codeviewer.

a18jimol:
Added the fileed.js script to codeviewer.php so you can use the fileed function to edit and save the files. Added the window for editing to the codeviewer.php from fileed.php.


a18mahkh:
I couldn't find this issue on the test server

Issue #7520, opened 2019-05-24
Save the examinator filter


a17adasy:
a sub issue to #4178.
The examinator filter does not save if you refresh the page. It should be saved in localstorage so when the user comes back later the filter is still there.

b18alegu:
This issue was merged into the W16 master. I'm closing this one now.

PR #7895, opened 2020-04-06
#7520 Saved Examinatorfilter to local storage

a18oscte:
In updateTable()
Made a variable to save the dropdown choice. saved it in local storage.
updated forloop on row 585 to select the option matching the saved variable in local storage

Co-Authored-By: @a18matna

### DIFF
--- a/DuggaSys/resulted.js
+++ b/DuggaSys/resulted.js
@@ -584,7 +584,13 @@ function returnedResults(data) {
 		teacherList += "<option value='none'>none</option>";
 		for(var i = 0; i < teacher.length; i++){
 			if(!teacherList.includes(teacher[i].id)){
-				teacherList += "<option value='"+ teacher[i].id +"'>"+ teacher[i].firstname + " " + teacher[i].lastname + "</option>";
+				if(teacher[i].id == localStorage.getItem('examinatorFilter')){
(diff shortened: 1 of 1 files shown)


a18ismca:
Unfortunately, ResultEd needs to be fixed up before I can review this pull request and approve it.

c14jonfr:
I am able to see that the store to localStorage works (by manually adding more options to the examinator dropdown) but since the teacher-variable (data.teachers) is empty I am not able to test if the value is taken from the localStorage and applied to show the saved drop down option.

Do you have any directions on how to test this, i.e. how was this tested during development?

a18matna:
> I am able to see that the store to localStorage works (by manually adding more options to the examinator dropdown) but since the teacher-variable (data.teachers) is empty I am not able to test if the value is taken from the localStorage and applied to show the saved drop down option.
> 
> Do you have any directions on how to test this, i.e. how was this tested during development?

We did the same, we manually added options and simply assumed it would work the same way

c14jonfr:
> We did the same, we manually added options and simply assumed it would work the same way

So you were not able to test if the value from localStorage was working (the last selected option shows up as selected on page reload)?

b18alegu:
I'll close this pull request and put this issue back into "In Progress"

PR #7899, opened 2020-04-06
An option has been added to download the diagram in A4 size

o18marka:
A export option has been added to export the diagram in a4 format. The issue was assumed to be solved as HGustavs comment above mentions "Svg export with limit to certain page/coordinate combo". Solution does also download the appropriate oriantation(portrait, landscape) depending on what option has been selected in the view options(Display virtual A4 -> Toggle A4 oriantation).

This solution only downloads the first A4(top left) in the canvas since including more would break the A4 format and make the A4 format export pointless.


o18marka:
As seen in history a fix was attempted, but reverted since more complex problems occurred and  comment by #a18frior is instead made into a separate issue.

PR #7903, opened 2020-04-07
G1 2020 w15 issue#1666

a18geoge:
In codeviewer.js, merged dehighlight function with highlightfunction regarding HTML and op. Replaced jQuery with JavaScript in the affected functions where possible.

Tested by hovering the mouse cursor over tags and the added important words.

Attaching screenshot to illustrate what it looks like.

codeviewerhighlight


a16silph:
Issue is solved~

Issue #7623, opened 2019-05-28
Make it possible to only have one A4-drawn


a17jacsv:
Make it possible to only have one A4-drawn and not only repeated ones. 

Also, the repeated A4-papers should cover the entire canvas and not only +coords as it is right now.
![Skärmavbild 2019-05-28 kl  11 10 45](https://user-images.githubusercontent.com/37792795/58466120-57f05d00-8139-11e9-9198-f46958ef86ed.png)

Issue #7250, opened 2019-05-17
Add form validation for Courseed


a17joaap:
These forms:
![validationeditcourse](https://user-images.githubusercontent.com/37792198/57934493-25846b80-78c0-11e9-9a24-f271ce0eba3c.PNG)
![validationnewcourse](https://user-images.githubusercontent.com/37792198/57934520-2fa66a00-78c0-11e9-8997-7bae2738b432.PNG)

Need validation. Part of https://github.com/HGustavs/LenaSYS/issues/6217

a17adajo:
#7252 can be used as a reference for how it can be implemented

a17amama:
@a17joaap What exactly do you mean by validation? Where should the validation be inserted?

a17joaap:
Catch erroneous input before it is sent to courseedservice. See #7252

PR #7906, opened 2020-04-07
G4 2020 w15 issue#7623 resolved

b18simla:
Added option for toggling between single and repeated A4, ready for testing.

Creating new issue: Repeated A4-papers should cover the entire canvas and not only +coords as it is right now (source: https://github.com/HGustavs/LenaSYS/issues/7623)


DuggaSys/courseed.js
```diff
@@ -459,119 +459,3 @@ function htmlFix(text){
...
-			alert('The course updated');
-			updateCourse();
-			
-
-
-		} else {
-			alert("You have entered incorrect information");
-		}
-	}
-}
```
b18phibe, line 577: How are the deletion of these row a part of this issues solution?

b18simla: Hmm, I'm not sure why that commit shows under this issue. It seems to belong in https://github.com/HGustavs/LenaSYS/issues/7250

PR #7920, opened 2020-04-07
Fixed Issue#7046

a18wilis:
This issue has been solved by removing the ability to edit the title by clicking on it. The only way to edit box titles now are by using the "Edit Content"/"Edit Box Settings (The cogwheel button).


d17marle:
Tested and works fine locally and on the test server

Issue #7886, opened 2020-04-06
[accessed] Swedish characters get messed up when updating user data


a17pioja:
![image](https://user-images.githubusercontent.com/49141519/78558866-a2d80800-7813-11ea-8711-fa7e478c588d.png)
___
![image](https://user-images.githubusercontent.com/49141519/78558834-95228280-7813-11ea-8dd7-facdb827b289.png)

When updating a user's data the swedish characters are represented incorrectly in the database.
The database supports swedish characters, but in the process of updating a user's data using the lenasys UI the swedish characters get messed up. 

The characters appear correct again when retrieved from the database to lenasys, but it's impossible to search for a name like "Åsa" using _SELECT * FROM `user` WHERE firstname='Åsa'_

To recreate:

1. Log in as a super user
2. Go to "Webbutveckling - datorgrafik" course
3. Press the padlock in the upper left
4. Update a users name or other detail using at least one swedish character

b18alegu:
This was merged during W15.

PR #7924, opened 2020-04-07
#7886 Fix problem were HTML encoded letters got into the database.

a18oscte:
Added string replacers to change the HTML encoded letters back to the original latin letters "åäö" before updating the database.


b18jonli:
Still a problem when adding new users, but that is probably a separate issue (will create one), this issue is fixed.

b18jonli:
This should be reverted as stated in #7978

PR #7940, opened 2020-04-07
G1 2020 w15 issue#7773

a16vikva:
This is how the link looks in markdown preview.
image

This is what was added, a drop down menu where one can pick a file from the database. Which will create a link to the code example.
image


d17marle:
image
seems to be an error occuring for this function

Issue #7865, opened 2020-04-06
Fix icon placements in mobile mode


a18oscba:
![image](https://user-images.githubusercontent.com/62877630/78540760-ecfdc100-77f4-11ea-8cd8-afb2765c085a.png)

Could be solved as they display in this issue
https://github.com/HGustavs/LenaSYS/issues/7516

PR #7948, opened 2020-04-08
#7865 Fix icon placement in mobile view

a18jonkl:
Created a new "div" under the header that holds courseDropdown and two buttons in mobile view.


a18jonkl:
Created a separate issue for missing buttons (issue #7950) @a18jenro

PR #7958, opened 2020-04-08
G3 2020#7388

b18emigu:
Implemented functionality for deleting files from the server, previously it would only delete it from the database.


a18felen:
Use teacher3 when testing because he is the only user with permissions to access the fileed page and the permissions too remove files on a few courses

PR #7978, opened 2020-04-08
Fixed problem where the html encode of åäö got sent to database

a18matna:
replaced all Html encoding with proper letter prior to the DB insert.


b18jonli:
The previous behaviour was intended and correct, this change should not be merged

PR #7983, opened 2020-04-08
Implemented guest user cookies so that guest id persist for one hour  - G1 2020 w15 #1490

b17carbr:
While the user is a guest they will get a cookie with their id as value for one hour: 
Screenshot 2020-04-08 at 14 03 47

At login the cookie gets removed:
Screenshot 2020-04-08 at 14 03 37


d17marle:
Minor issue needs to be changed, doesnt give out a new guest id once logged in and then logged out again.

d17marle:
The minor issue was solved by the developer and is ready to be merged

PR #7989, opened 2020-04-08
G4 2020 w15 issue#6005

### LINKED ISSUE #6005: Crossing lines vs merged lines
There should be a distinct difference between lines that are just crossing (not connecting to each other) and lines that are merged. Suggest that we have a filled dot/point where the lines merge and a "lump" or "jump" where the lines just cross and not connect.

b18phibe:
Have fixed so crossing lines will be represented by a half circle lite a line jump, it ain't the most beautiful thing i have seen but it works.  (It doesn't work in UML only in the ER mode, but shouldn't hurt the UML mode)closes #6005


a18kaswi:
@b18phibe closing this PR for weekly.

PR #8049, opened 2020-04-14
Fixed navigation through menu with tab

### LINKED ISSUE #6021: You should be able to access the menu bar above the diagram using only the Tab key
As with any site, you should be able to move through the main content of a page only using the tab key as it allows users with no mouse / steadiness-related disabilities to navigate the page. As of right now if you navigate using tab key, the menu bar above the diagram is completely skipped in the sequence of content, this should be changed. Image below illustrates the menu and the options that should be available, but aren't.
user interraction
Furthermore this issue could be expanded to cover more options, like the log-in button in the top right corner of the page.

o18marka:
Fixes #6021 
Code from previous solution went missing after weekly-merge. This commit adds back missing code.

### DIFF
--- a/DuggaSys/diagram.php
+++ b/DuggaSys/diagram.php
@@ -254,26 +254,26 @@
                             </div>
                             <div class="drop-down-divider">
                             </div>
-                            <div class="drop-down-item">
+                            <div class="drop-down-item" tabindex="0">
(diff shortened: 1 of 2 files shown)


a18kaswi:
@o18marka

o18marka:
@a18jonpe , will take a look at this!

Issue #7868, opened 2020-04-06
Notice of MYSQL_VERSION break design


a18antsv:
MYSQL_VERSION is already defined sometimes as it is defined every time the pdoConnect function is executed. A global variable should only be defined once otherwise it could cause problems. The design is broken because of PHP giving a notice of this.
This could be a problem only after PHP7 that is used now.
![Skärmklipp](https://user-images.githubusercontent.com/49121839/78542606-b2e1ee80-77f7-11ea-85f5-dfbcaf88c0b3.PNG)

a18kaswi:
Again unclear why the merge hasnt closed this issue. Closing but have to investigate further.

a18kaswi:
Didnt make it to weekly unclear why #7870

a18kaswi:
Didnt make it to weekly unclear why #7870

PR #8050, opened 2020-04-14
Fix for MYSQL_VERSION constant already defined, issue #7868

a18antsv:
Fixed the issue where MYSQL_VERSION could be defined more than once when connecting to database. A constant should only be declared once.


a18kaswi:
Need notices activated on test-server to be able to be tested. Closing until that is fixed.

Issue #5780, opened 2019-04-10
Line thickness slider gets hidden in the Firefox browser when draged


a16tobfr:
![lineThicknessSlider](https://user-images.githubusercontent.com/49199690/55876346-b9fbff80-5b97-11e9-9cce-443d9589c4d5.JPG)
When using the Firefox browser, the slider gets hidden while being clicked and dragged. It would be nice if this could be fixed.

HGustavs:
Do we know why?

This must be css styling?

Normal range input does not behave like this......

a16tobfr:
Code located in diagram_forms.php
~~~
<input id="line-thickness" oninput=\'globalLineThickness(); hashFunction(); updateGraphics();\' style="width:100%; margin: -2px; padding: 0px;" type="range" min="1" max="3" value="1">
~~~
A tip is to check out how the zoom slider is made which can be found in the shared style.css document.
.zoomSlider uses the following for adjusting to different browsers
For Chrome, Opera, Safari, Edge:
.zoomSlider::-webkit-slider-thumb {
...
}
and for FireFox:
.zoomSlider::-moz-range-thumb {
...
}

This might help in order to solve this issue in the future.

a18kaswi:
PR #7963 not good because rework of system. Check with @a18antsv

a18petha:
This Issue implementation needs to be redone to be compatible with solution from #5638

PR #8063, opened 2020-04-14
Reimplemented solution for new HTML structure

a18petha:
**For testers**

1. open this address. http://group4.webug.his.se:20001/G4-2020-W16-ISSUE%235780/DuggaSys/diagram.php in firefox

2. test the input elements of type "range"

one of these elements can be found by using the "Draw Free" tool in ER mode

Link to original Issue #5780


a18petha:
Added styling to this issue after specification

Issue #7973, opened 2020-04-08
AccessEd table group dropdown closes instantly after opening in certain scenarios


b18jonli:
When a group dropdown is open and the user clicks to open another group dropdown, all subsequent dropdowns will close instantly after opening. The problem stops if you click outside the dropdown column.

HGustavs:
Confirmed.

This is really annoying...

PR #8071, opened 2020-04-14
Removed duplication of accessed table

### LINKED ISSUE #6068: Check mark in Accessed.php changes the name of the person.
When you press the check mark icon next to the persons name it will change the name of the person. 

Before pressing the check mark on Alex. 
image

image

And after you press the check mark the name will change, in this case Alex Bridgeman will have another name (Anders in this case) when you press on him. 

image

image

b18jonli:
Fixes #6068.
Should improve performance as well as make the table work correctly.

### DIFF
--- a/DuggaSys/accessed.js
+++ b/DuggaSys/accessed.js
@@ -521,6 +521,7 @@ function returnedAccess(data) {
 		hasMagicHeadings: false,
 		hasCounterColumn: true
 	});
+	
 	myTable.renderTable();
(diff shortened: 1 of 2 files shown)


b18jonli:
Seems to also fix #7973

b18jonli:
This introduces (or rather stops hiding) a new issue where the group dropdown menus are not updated visually when selecting groups. Will create a new issue...

Issue #7507, opened 2019-05-24
Refresh alert when active


b17annni:
I get the refresh alert even when I'm active. It's annoying.
![image](https://user-images.githubusercontent.com/37792766/58321171-b8bc2480-7e1d-11e9-9070-8781959881ca.png)

a17adasy:
**Suggestion:** Having the alert show when you havent moved the mouse for 10-15 minutes.

PR #8074, opened 2020-04-14
Implemented refreshing of session cookies when user is active - G1-2020-W16-#7507

b17carbr:
The issue https://github.com/HGustavs/LenaSYS/issues/7507 wanted a solution to prevent the alert being triggered even though the user is active on the site. 

Screenshot 2020-04-14 at 13 42 43

To solve this the session cookies which holds the time when the alert is suppose to trigger instead now gets refreshed by the user navigating the site. Refreshes takes place when navigating to codeviewer.php, courseed.php, and sectioned.php.
Screenshot 2020-04-14 at 13 48 02

### DIFF
--- a/DuggaSys/codeviewer.php
+++ b/DuggaSys/codeviewer.php
@@ -130,6 +130,13 @@ Testing Link:
 					
 				}
 			}
+			else{
+				// refreshes session cookies, thereby extending the time before users sees the alert or get logged out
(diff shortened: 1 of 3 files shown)


a18eliek:
@a18jakme  hej

a18jakme:
@a18eliek hejhej

PR #8092, opened 2020-04-15
G1 2020 w16 issue#7669

a18filkr:
Highlighted words with strings in codeviewer should now work.
If there is a string withing a string, single quotes should be used.
Special characters after/before a word such as "," or "!" should work and not be marked as important.


a18jakme:
Color of a important word in a string is still black, it's better to have it blue when it's in a string?
image

a18filkr:
Color mentioned by tester should now work as intended.

PR #8118, opened 2020-04-15
Update in resulted.php - modification of heading and it's title

a18ismca:
The heading was replaced to far left instead of being centered. Also, the title was changed to increase comprehension of the site.

Co-Authored-By: a17pioja <a17pioja@users.noreply.github.com>

### DIFF
--- a/DuggaSys/resulted.php
+++ b/DuggaSys/resulted.php
@@ -47,8 +47,8 @@ pdoConnect();
 	<div id="content">
 		
     <div id="resultedFormContainer">
-	<div class="titles" style="justify-content:center;">
-			<h1>Result</h1>
+	<div class="titles" style="justify-content:flex-start;">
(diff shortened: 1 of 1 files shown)


a18ismca:
@a18viksa okay! we are going to create another issue that addresses several php files and not only this one. thanks for suggesting :)

a18ismca:
@c14jonfr okay! we will work on it... i think that new issue is already in progress in our project... so if you take a look and then comment what you think about it :)

with @a17pioja

b18alegu:
I'll put this back to in progress since you are working on it 👍

a18ismca:
issue #8131 solves the problems you have addressed in this issue @a18viksa and @c14jonfr... so it would be good if both of you approve this issue so this issue correspond with the other new issue #8131

Issue #7639, opened 2019-05-28
Click and drag vs click


a17drama:
When adding new entities user has two options that are mentioned in the title of this issue. A small addition that would make this even better is to have a certain distance that mouse has to be moved before activating click and drag. As it is now it triggers to easy so the simple click to create an object feels inconsistent.

PR #8132, opened 2020-04-15
added minimum distance before activating drag and click

a18conch:
added minimum distance before enabling drag and click

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -165,6 +165,7 @@ var classTemplate = {
   width: 7 * gridSize,
   height: 7 * gridSize
 };
+var minimumDivisor = 4;                 // Is used when dividing templates for minimum selection before activating dragging mode values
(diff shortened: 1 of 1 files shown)


a18conch:
> Correct me if I'm wrong, but shouldn't the minimum distance be a requirement for all objects in the diagram? As of now it only seems to work on "Entities" in ER

https://github.com/HGustavs/LenaSYS/issues/7639 only mentions "entities" but it could be interpreted either way I guess

b18simla:
> > Correct me if I'm wrong, but shouldn't the minimum distance be a requirement for all objects in the diagram? As of now it only seems to work on "Entities" in ER
> 
> #7639 only mentions "entities" but it could be interpreted either way I guess

It doesn't specify ER-entities, and it also later refers more broadly to creating "objects", which is the reason I mentioned it.  Overall I think it makes more sense to have this function apply to all entities, which @a18kaswi agrees with

PR #8134, opened 2020-04-15
Icon redesign to a more common and modern version

### LINKED ISSUE #7451: Suggestion: Remove handle part of filter icon in accessed
At the moment it looks like this:
image

I think it would look better if you remove the handle part of the filter icon since I personally think it looks out of place. Quick mockup:
Untitled-1

I would like to know what other people think about this though.

a17pioja:
fixes #7451
The icon has been changed to a more common icon for filtering.

Another icon is included but not used, this icon is the same as the original icon but the ring has been removed to optimize it to look more modern. It's there in case it's more desired than the current new choice.

The new icon used is filter_icon.svg
The optimised unused icon is tratt_white_v2.svg

image

Co-authored by : @a18ismca


a17pioja:
The changes made to navheader.php were reverted due to the file belonging to another group's domain.
The margin problem will be adressed with group responsible for the navheader.php file
@a18viksa

PR #8141, opened 2020-04-15
Updated group dropdowns in accessed table

### LINKED ISSUE #8082: AccessEd table: Group dropdown menus are not updated visually when selecting groups
Only an issue after #8071

b18jonli:
- Removed unnecessary global variable "expanded"
 - Better show/close logic for group dropdowns
 - Dropdowns now updates visually
Fixes #8082


c14jonfr:
There is a console.log() at line 578, this should probably be removed before merging.

PR #8143, opened 2020-04-15
On screen message handling

o18marka:
A function called "flash" has been implemented which shows a danger/success message on screen.

This function is now used to show a flash message to the user if he/she tries to make a free draw object with less than 3 points(makes two points then right-clicks).

This function can further be used to show any type of message which may be useful to the user.

### DIFF
--- a/DuggaSys/diagram.php
+++ b/DuggaSys/diagram.php
@@ -360,6 +360,9 @@
                             </div>
                         </div>
                     </div>
+                    <div id="errorBox">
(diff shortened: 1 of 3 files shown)


a17robhe:
@o18marka Can you confirm that Andreas conclusion is correct?

o18marka:
Yes, the success message was tested during development. But as you said that's tough for you to test since it's not used yet. 

But yes, @a18andtu conclusion is correct!

PR #8145, opened 2020-04-15
G1-2020-W17-ISSUE#8144

a18abuah:
Fixed margin problem in codeviewer so the first row is visible.
image


d18mohal:
It seems to be fixed. The first row it no longer invisible. Good jobb! 
Screenshot 2020-04-16 at 09 59 56

PR #8151, opened 2020-04-16
G4 2020 W17 issue#7765

### LINKED ISSUE #7765: Download all files as zip

b18phibe:
You are now able to download files that are version independent to you course as well as the files to the specific version of the course you are in at the moment.  the already uploaded files are in non of those categories and will not be downloaded as of right now. But all new files will be uploaded appropriately. closes #7765

### DIFF
--- a/DuggaSys/downloadzip.php
+++ b/DuggaSys/downloadzip.php
@@ -1,0 +1,114 @@
+<?php
+/***********************DOCUMENTATION**************************
+ * This file should make downloading of all material          *
+ * of a singel course doable given we know the courseid       *
+ * every file sould be put in one zip file and then downloaded*
+ **************************************************************/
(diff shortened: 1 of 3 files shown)


a18kaswi:
Note to self: remember to make changes to readme and etc how to enable zip on php if this pr passes testing.

a17robhe:
@b18phibe If more documentation is needed in the code please provide this, see @b18mikfr 's comment

b18phibe:
> Works and follows the code standard. Some more comments on what happens on ELSE while processing IF satements would be nice otherwise, it works.
@b18mikfr 
What do you mean with " Some more comments on what happens on ELSE while processing IF satements"?

a17robhe:
@b18mikfr Can you check that the new comments

PR #8155, opened 2020-04-16
Reimplemented functionality

b18mikfr:
Took its time but it now works as intended with clipping


b18mikfr:
@b18phibe That method of implementation was not compatible with the previous implementation. I had to rethink my approach but the ResizeToMinumum() was a remnant from last weeks merge

Issue #8094, opened 2020-04-15
Highlightning of word in code shows the code in description.


a18filkr:
If description for example is "Hello world" and Hello is an important word as well as onclick. Onclick will be highlighted in the middle of the span and destroy the span that should've highlighted Hello.

This shows the code in description and looks broken. Hello doesn't work because of this either.
![Screenshot_2020-04-15 Validating form data](https://user-images.githubusercontent.com/49142751/79318148-eb27b200-7f06-11ea-92ac-b4a851da8a9d.png)

PR #8163, opened 2020-04-16
G1 2020 w17 issue#8094

a18geoge:
Fixed the bug with specific important words. Events will no longer be highlighted as important within document examples in codeviewer. 

Testing may be required since there is no popup redirecting to a google search when I click on an important word in a document example but when testing on a18wilis end this works. 

How to test: add for example "onclick" and "hello" (if your example contains the word hello) to the important word list. See issue #8094


d17marle:
Tested on the weekly branch after merge and it works splendid

PR #8164, opened 2020-04-16
g1-2020-w17-issue#7955

b18gussv:
Removed commented out code so that markdown would recognize that a table is created.
bild

Alignment now works and it works together with lists (both unordered and ordered). It follows standard markdown syntax for tables.


a18abuah:
The table is working exceptionally good !!!
image

PR #8176, opened 2020-04-16
G1 2020 w17 issue#7760

a18filkr:
By getting the timestamp, css and js files in codeviewer.php should new be reloaded if they are saved. For the CSS this works in a single array because they are all in the same directory, the JS files should be placed similiarly to get this to work, currently only the jQuery files are used as an array.

To get this to work everywhere, moving js files or buildings function to search in many directories should be done.

### DIFF
--- a/DuggaSys/codeviewer.php
+++ b/DuggaSys/codeviewer.php
@@ -21,7 +21,9 @@
 
     $css = array(
         'codeviewer.css',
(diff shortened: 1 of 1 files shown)


b18gussv:
Looks very good on the test server.
bild
The timestamps change as well upon  a new version being uploaded.

PR #8179, opened 2020-04-16
Fixes #8138

### LINKED ISSUE #8138: Notice: Undefined offset: 1 breaks design - Navheader
The notice in the image is breaking the design on servers having notices activated. Values are access directly from an array. Isset would probably fix this problem since we need to know if there will be something on the position that is accessed in the array before accessing it.
image

a18antsv:
Test if notice Undefined offset: 1 can appear. Be logged in and choose courses that do not have versions.
Link: http://group4.webug.his.se:20001/G4-2020-W17-ISSUE%238138/DuggaSys/courseed.php

### DIFF
--- a/Shared/navheader.php
+++ b/Shared/navheader.php
@@ -100,24 +100,26 @@
 							$query->execute();
 							$result = $query->fetch(PDO::FETCH_ASSOC);
 
-							// Changes format from 'HT20' to numbers to create the URL
-							$array = explode("T", $result['versname']);
-							$array[0]; 
-							$year = "20";
-							$year .= $array[1];
-							if ($array[0] === "H") {
-								$term = 2;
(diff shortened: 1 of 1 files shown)


a18petha:
Can't really understand the issue at hand here, are we only looking for direct match on the errors? I still encounter this error then selecting a course.
anton

is this related to issue#8138 ?

a18antsv:
> Can't really understand the issue at hand here, are we only looking for direct match on the errors? I still encounter this error then selecting a course.
> anton
> 
> is this error message related to issue#8138 ?

That issue is not related to this one. That notice is already fixed to our weekly branch in another pull request but this branch was created before that happened. The important part here is that you do not get this notice somewhere.
image

Issue #7385, opened 2019-05-22
Esc button doesn't work when trying to close the Edit course version/Delete dugga dialogs in sectioned.php


b17carma:
When trying to close the edit coruse dialog in sectioned.php by pressing the escape key, nothing happens.

c17johha:
is it this dialog?
![image](https://user-images.githubusercontent.com/49199825/58174618-964fcd00-7c9e-11e9-921f-62b1d7444685.png)

Works here on chrome and Firefox.

b17carma:
No, I mean the one in sectioned.php opened from the cog on the right of this picture:
![image](https://user-images.githubusercontent.com/37834592/58247765-b644c680-7d5a-11e9-9670-278cd0b17ceb.png)

PR #8180, opened 2020-04-16
Close the "new course version" and "edit course version" windows with ESC button, fix for issue #7385

a17towan:
It is now possible to close the "new course version" and "edit course version" windows with the **ESC** button.

These buttons:
image

And these windows:
image

image


a18jimol:
It works to close "new course version" and "edit course version" with the ESC button. Didn't find any problems or error. Looks good!

Issue #8011, opened 2020-04-09
Check database for course ID 


a18frior:
Validate new course version and edit course version forms by checking datebase for course ID to make sure that the user doesn't get duplicate key error.

a18antsv:
When courses are fetched from database and elements are created for each course in a list, there is the possibility to add a data-attribute (data-code) for each element with the value of the code for the course represented by that element in the list. These codes could later be extracted from the elements into an array and when validating a check could be made for if the value in the input is equal to any of the values in the array. This way would prevent fetching from database on each validation attempt.

a18antsv:
When editing a course one exception must be taken into consideration. You should be able to save the course with the same course code as the course previously had. If a list of all occupied course codes exists and a check against it is made it would prevent you from saving with the same course code again. The code the course actually has could be passed a data-attribute to the input element when the form is opened. This way it would be possible to make the exception to accept the same course code for the course the form is representing with the code in the data-attribute.

a18antsv:
The fix for this issue will also fix the problem with validation blocking necessary course names.

PR #8192, opened 2020-04-16
Fixes issue #8011

a18antsv:
Validation for course code. A message appears if the types course code is used by another course. This will prevent unnecessary ajax calls to server as well. For editing of courses a solution was implemented to still be able to save the course with the same course as it previously had without it getting blocked by validation of codes courses already have. 
Validation was reworked to be more general and code was cleaned up a bit.

Test if all validation works as expected.
- Should be possible to have multiple words in coursename.
- Dash can be in between two words like in Webbutveckling - datorgrafik only.

- When validating course codes it has to follow the format two letters, three digits, one letter. Example: 'IT108G'.
- A message should appear in real time when typing if course code is already in use by another course.
- It should be possible to save a course with the same course code as it previously had when editing a course. Try changing some in the input and then writing the previous course code again (no validation error should appear) and then type a course code that another course has (validation error should appear)

Link: http://group4.webug.his.se:20001/G4-2020-W17-ISSUE%238011/DuggaSys/courseed.php

### DIFF
--- a/DuggaSys/courseed.js
+++ b/DuggaSys/courseed.js
@@ -129,6 +129,10 @@ function selectCourse(cid, coursename, coursecode, visi, vers, edvers)
 	$("#cid").val(cid);
 	// Set Code
 	$("#coursecode").val(coursecode);
+
+	//Give data attribute to course code input to check if input value is same as actual code for validation
+	$("#coursecode").attr("data-origincode", coursecode);
+
 	// Set Visibiliy
 	str = "";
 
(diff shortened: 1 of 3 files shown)


o18marka:
Worth to mention is that course code cannot contain åäö according to validation. Though i assume this should never be a scenario anyways.

a18antsv:
@o18marka Yes, I do not think we would ever want to have a course code containing åäö so that's why I left it out. I just added åäö to the course name validation since it was not possible to create courses with åäö before. We are in Sweden with swedish course names so I definitely think courses could come up on LenaSYS that requires åäö.

PR #8201, opened 2020-04-17
Fixes #8200 - Move button active when not in move mode

### LINKED ISSUE #8200: Move button active when not in move mode
When being in move mode and changing mode (DEV, ER, UML), the uimode will be reset to normal without hiding the move button. This will make it look like move mode is active when it clearly is not.

a18antsv:
Test if move button can be visible when not being in move mode by activating move mode, switching to a different diagram mode (UML, ER or DEV) and see if you will still be in move mode with the button visible (success) or if you will be in normal mode with the button visible (fail).

Also minimized related code that was repetetive to look much cleaner.

Link: http://group4.webug.his.se:20001/G4-2020-W17-ISSUE%238200/DuggaSys/diagram.php

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -381,7 +381,10 @@ function resetToolButtonsPressed() {
     // deselect draw text button
     document.getElementById("drawtextbutton").classList.remove("pressed");
     document.getElementById("drawtextbutton").classList.add("unpressed");
-    uimode = 'normal';
+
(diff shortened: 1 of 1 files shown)


a18antsv:
> I simply can not get it to work while switching between modes. The move icon are still active and the move mode are disabled when switching modes.

Strange. Iwill take a look. It worked yesterday.

a18antsv:
It should be fixed now.

PR #8228, opened 2020-04-17
G4 2020 w17 issue#7003

a18conch:
Added basic rebindable keys feature


a17robhe:
What is the source of these conflicts? @a18conch @g18marma

PR #8242, opened 2020-04-17
Maximize button now works with every template

a17chrka:
Maximize button works for every template, except for template 9 since that template doesn't currently have a maximize button.

Also made sure to update alignBoxesWidth to make it more dynamic.

Made some updates to resizeBoxes for template 5 as well.


a18eliek:
@a18jakme Could you have a look at this conflict? It seems like it is modifying rows that you have made a commit on (https://github.com/HGustavs/LenaSYS/pull/8190/commits/86c77c9c285df3dc29265659026712ddd31a4b93)

a18abuah:
Every maximize button works perfectly!
image

Issue #7583, opened 2019-05-27
UML TEXT size changes in a bad way with global appearance


a16tobfr:
Either keep the text to the same size (as it was previously, uml text was always set to a standard size and not affected by global appearance changes), or change it so that the text is not as weird as in the image below.
![umlTEXTSIZE](https://user-images.githubusercontent.com/49199690/58422674-769c1880-8093-11e9-8db9-f31e07a84cc9.JPG)

a18antsv:
This issue should be solved when solving #7645 probably. I don't see a point in making a fix for only this solution as it would need to be reworked by #7645 later anyway.

PR #8246, opened 2020-04-17
G4-2020-W17-ISSUE#7583

a18petha:
**For Testers**

open http://group4.webug.his.se:20001/G4-2020-W17-ISSUE%237583/DuggaSys/diagram.php
play around with global appearance font-size setting. See if you can get the Attribute box to overlap with operation box or Vice versa with a UML class object

Closing #7583


a18petha:
Fix bugs from testing

PR #8247, opened 2020-04-17
Resulted: Fixed minor styling issues and made functions for dropdowns.

a18viksa:
Fixed the problem that the overflow-text could cause by adding a onmouseover-function that displays a dropdown containing the full text that the ellipsis cuts off. 

image

Also added a min-width to the th-elements to avoid narrow table heads and also removed some styling that disturbed the th-texts formatting. 

image

Co-authored-by: b18fella <b18fella@users.noreply.github.com>


a18viksa:
Note that the text inside the dropdown only updates if there's an actual change to a dugga name. 
The table head that says "Biträkningsduggorrrrrrrrrrrrrr..." is just temporarily added in the developer tool and therefore the text inside the dropdown is not changed.

Issue #8191, opened 2020-04-16
$cid is undefined in navheader.php


b18emigu:
The variable `$cid` is out of scope because it is assigned a value inside of an `if & else` statement as well as other places that changes its value. To fix define the `$cid` variable outside of the `if & else` statements, as an example on line 5 as `$cid = "UNK";` That way the variable wont be undefined when accessed later in the code.

This issue also prevents the page from being loaded properly when the user is not a super user as seen in the picture below, the user used for this was `teacher3`.

![duggasys](https://user-images.githubusercontent.com/49141724/79478052-3b883800-800b-11ea-98b8-088564d03790.PNG)

HGustavs:
General fix. Make the code support both $cid and $courseid this way it works with both old and new releases of the system.
(refactoring rule)

b18emigu:
> General fix. Make the code support both $cid and $courseid this way it works with both old and new releases of the system.
> (refactoring rule)

Okey, we can do that!

b18emigu:
A new commit with a fix has been pushed to the branch. We included the coursevers because it was related to the courseid. They are now both stored in the `$_SESSION` superglobal variable and refrenced instead of the `$cid` variable.

PR #8249, opened 2020-04-17
Variable cid is no longer out of scope.

b18emigu:
With this fix the ` $cid` variable in `navheader.php` should no longer be out of scope.


b18emigu:
Me and @a18felen have now updated the pull request with a fix like @HGustavs requested in the linked issue #8191

b18emigu:
I had forgotten to check if the `cid` index was set using  `isset()` meaning on the start page before entering a course it would cast an error, this has now been fixed.

Issue #8221, opened 2020-04-17
URL in showDugga uses 'cid' instead of 'courseid', which all other pages use.


c14jonfr:
This should possibly be changed for consistency.

**Note: if this change is implemented you have to update the corresponding querystring-statements to use 'courseid' or things will definitely break.** 

1. Screenshot from sectioneditor URL (has courseid, not cid)
![image](https://user-images.githubusercontent.com/49141758/79554377-4213bf80-809e-11ea-9182-51fefa41ceab.png)

2. Screenshot from showDugga URL (has cid, not courseid)
![image](https://user-images.githubusercontent.com/49141758/79554420-51930880-809e-11ea-8901-7bce3ac31873.png)

HGustavs:
for consistency. but, I suggest also supporting the old names to allow for interleaving of old and new files.

b18alegu:
Closed after successful weekly-merge.

PR #8261, opened 2020-04-20
G2 2020 w17 #8221

a18oscte:


b18alegu:
I've been in touch with group-3 and the branch `G2-2020-W17-#8221` can ver merged with weekly so the small changes in sectioned.js should not create conflic.

Kudos to @b18jonli for noticing that another groups file have been changed.

Issue #8086, opened 2020-04-14
Bad design of cardinality in symbols


a18antsv:
Only one object is stored per symbol for cardinality ever. Cardinality is still stored as an array of objects which is very unnecessary when it could be stored as one object only. This takes unnecessary space and makes the code unnecessary messy. It has to be access by writing cardinality[0] instead of simply cardinality.

PR #8270, opened 2020-04-20
Issue #8086

b18phibe:
It seems like it only were to remove all off the [0] and the [ ] surrounding the parameters of this.cardinality. Everything seems to work great. There will probably be weird bugs thanks to this change but i could not find any in my 10 minutes testing 

Closing #8086


a18antsv:
Yes, this was the solution I had in mind when creating the issue. It should not cause any bugs.

a18antsv:
Remove the middle [0] of object[access[0]][0][access[1]]; 

This occurs 2 times. One in setObjectProperties and one in setSelections.

b18phibe:
@a18jonpe Should be resolved now

Issue #7423, opened 2019-05-23
Coordinates bar should be hidden in mobile-view


b17robev:
Currently the coordinates bar are shown in mobile-view it should be hidden, since it is impossible to track a users fingers when its not on the screen. In the picture below it shows the last pressed point. The zoombar should instead cover a wider area on the mobile-view.

![ugly](https://user-images.githubusercontent.com/37845511/58236813-ff3d5080-7d43-11e9-80d1-964e711541fa.png)

HGustavs:
Can someone clarify the "since it is impossible to track a users fingers when its not on the screen"  part. 

Isn the point of that box to show the current or last available coordinate?

a17alfcl:
> Can someone clarify the "since it is impossible to track a users fingers when its not on the screen" part.
> 
> Isn the point of that box to show the current or last available coordinate?

It doesn't bring enough useful information to justify occupying such large part of the screen IMO

PR #8280, opened 2020-04-20
No longer shows developer coordinates in mobile view, issue #7423

a18andtu:
When in mobile view the coordinate box for developer mode takes up a large part of the screen and has been hidden in mobile view.


a18antsv:
It works on android and in chrome mobile view. Maybe a fix is needed for iPhone?

a18petha:
Seems as the test failed on Google chrome mobile extension, maybe this is something that needs to be fix but after carrying out the test on my phone device there was no issues.

PR #8285, opened 2020-04-20
Double scrollbars solved

a17chrka:
The templates that had double scrollbars on some boxes have now been fixed.

Template 6:
bild

Template 7:
bild

Template 8:
bild

Template 9:
bild


a18abuah:
The double scrollbar issue is solved! great work.

PR #8288, opened 2020-04-20
G1 2020 w17 issue#8202

a18filkr:
This fix only removes the description when screens are resized to make sure min max and reset are still possible to be clicked.
To make the best possible solution a bigger rework of the menu should to be done by not using a table. This however looke better than overlapping and doesn't "destroy" the graphic in the same way as before.
Also removed some redundant css.


a18abuah:
works as intended, exquisite work Filip!

PR #8293, opened 2020-04-20
Solution for branch G1-2020-W17-ISSUE#8279

d17marle:
Height is now important, wont be affected by the other boxes. 
image


a18abuah:
The height on template 8 boxes is working well with maximize now! great work

PR #8224, opened 2020-04-17
Added basic service logging for diagramservice and highscoreservice.php - G1-2020-W17-ISSUE#8213

b17carbr:
Result from using the service: 

Screenshot 2020-04-17 at 12 04 45
Screenshot 2020-04-17 at 12 04 54

How to test:
Download https://sqlitebrowser.org/dl/

Choose open database
Screenshot 2020-04-17 at 12 08 08

Open the db file located in the log dir
Screenshot 2020-04-17 at 12 07 14
Screenshot 2020-04-17 at 12 07 25

Run services mentioned in the pull request:
Screenshot 2020-04-17 at 12 10 37
Screenshot 2020-04-21 at 10 58 27

then look up the result in table:
Screenshot 2020-04-17 at 12 12 38

co authored with @a18jakme

PR #8295, opened 2020-04-20
Logging for adding a new item in Section

a18jakme:
Logging for adding a new item in Section.php. Logging works through the "+" button and the shortcut. A new EventType (14) is added and a description on type of the new item.

image

For testing follow guide in https://github.com/HGustavs/LenaSYS/pull/8224

co authored with @b17carbr


a18filkr:
bild
Works good locally.

a18filkr:
bild
Works good on the server as well.

PR #8301, opened 2020-04-20
Solution for branch G1-2020-W17-ISSUE#8282

d17marle:
Now it is no longer visible, like it should
image


a18abuah:
The button are no longer in template 9 and 10, great work!

PR #8304, opened 2020-04-20
G2 2020 w17#7667

a18oscte:
jsjs
This is the new version.


a18oscte:
The easiest way to test this pull request is to log into one of the student account. Example a99majro with password: password. Do a dugga and save it. Then log into brom and give the dugga some feedback. Then if you log in on the student account again you can se ther new feedback log.

Hope that helps!

Issue #7645, opened 2019-05-28
Global Appearance should be mode specific


a16tobfr:
The global appearance should be mode-specific, meaning that when the user is in ER mode, the global appearance should only change the settings for the ER objects, and when in UML mode the settings should only change for the UML objects.

The same approach is also relevant for the three dev modes.

DEV + Display All Tools -> Global Appearance -> affect ER and UML

DEV + ER -> Global Appearance -> affect ER ONLY
DEV + UML -> Global Appearnace -> affect UML ONLY

MODE: ER -> Global Appearance -> affect ER ONLY
MODE: UML -> Global Appearance -> affect UML ONLY

This means that the DEV + ER and ER modes belong to the same category, and the same is true for the DEV + UML and UML modes.

a18antsv:
This is necessary. As of right now there is no way of knowing what will be affected without trial and error. You could also question if UML should even be affected at all by any changes for global appearance. There is no options for changing appearance except for name, operations and attributes for individual UML appearance. If it should be possible to change UML apperance from global apperance there should also be options in the individual appearance form to make it possible to change one uml class apperance more. Otherwise UML classes should not be affected by  global apperance. Same for all other types of symbols.

a18antsv:
I have reworked the whole apperance menu which should make this easier to implement because of the logic is being built by.

a18kaswi:
appearanceDiagram
#7677
#7928
#7645
#8067
#6888

a18antsv:
I have fixed this now. There is just one problem; all objects use the same base settings when being created. So when changing global appearance settings in UML mode, only objects active in the toolbar will be changed. But when creating a new object that is not in the UML toolbar, they will still have the new settings that were set for UML. Or when opening the global appearance menu for a new mode, the settings that were set while in UML mode will be used as presets.

a18antsv:
It could be solved by having a different settings object for each mode.

PR #8315, opened 2020-04-20
Fixes #8314 - Proper check to prevent action on wrong objects

### LINKED ISSUE #8314: Object disappear on appearance form submit
When the appearance form is submitted for objects that are not UML, the object will disappear. This is due to a recent change getting into the weekly branch. This change executes a function when the appearance form is submitted on the previously selected object. The executed function could be executed on objects that breaks from this function. The function is only intended for UML objects. This could be fixed simply by acondition to check that the object type is UML. It could also be intresting to iterate through all the selected objects and execute the function for all objects that are UML instead of only doing it for the last selected object.

Another minor issue that got into the weekly branch is a function getting called on the object when loading the appearance form and causing an error. This function does only exist for symbols, not paths. Trying to call this functions when opening the appearance menu for a path will cause an error making it impossible to open the appearance menu. This should be fixed as well in this issue as it's so small.

a18antsv:
Test if it is possible to change appearance for all types of objects without any errors. UML should resize back to minimum required height when appearance form is submitted. 
The problem before was that a function was executed on an object that the function was not intended for, which would break the object.

Link: http://group4.webug.his.se:20001/G4-2020-W17-ISSUE%238314/DuggaSys/diagram.php

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -4458,14 +4458,13 @@ function loadAppearanceForm() {
     let type = object.symbolkind;
 
     //Get objects connected to uml-line and sets name in appearance menu(used for Line direction)
-    var connectedObjectsArray = object.getConnectedObjects();
-    if(object.symbolkind == 7){
+    if(object.symbolkind == symbolKind.umlLine){
+        let connectedObjectsArray = object.getConnectedObjects();
         document.getElementById('First').innerHTML = connectedObjectsArray[0].name;
         //Selection to check if relation is to the same entity. If so: both are named from object 0
         if(typeof connectedObjectsArray[1] == "undefined"){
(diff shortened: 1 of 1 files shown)


a18antsv:
An extra fix was added for another small problem that got into the weekly branch. This for a function getting called on all objects even though it does only exist for symbols, not paths. Trying to open the appearance menu for a path caused an error as a function that does not exist for it was called. It is better to call this function inside of the already defined if-statement. It is only relevant for UML lines here.

a18antsv:
> Everything works as intended, although a side effect from global appearances tab is that we can now preview styling UML class with it. UML class own appearance does not allow for this to happen maybe this should be removed? or we could implement styling for UML classes as well in a separate issue

Yes, I already had this in mind. This will hopefully be fixed when solving #7645.

PR #8330, opened 2020-04-21
Logging for upload a file in showDugga.

a18jakme:
When upload a file in showdugga there is log with username and filetype. 
image

co authored with @b17carbr


a18filkr:
bild
The files cannot be upploaded even though i choose correct type however they ARE logged so the logging part works for upload on testserver.
bild

Issue #8305, opened 2020-04-20
Template 9 rezizing height not working properly


a17chrka:
A box that is being dragged down can only push down the first box below itself but not the second box.

<a href="https://gyazo.com/e611c029628f0a8b0ff02579ad0ab1e4"><img src="https://i.gyazo.com/e611c029628f0a8b0ff02579ad0ab1e4.gif" alt="Image from Gyazo" width="1916"/></a>

HGustavs:
Seems like a bug.

There is something wrong with that template?

Is the fourth box not the same as the others?

PR #8349, opened 2020-04-21
Issue #8305 - Template 9 rezizing height not working properly - Now fixed

a17chrka:
The fourth window will now be resized as well when pushing down the second and third window.

<a href="https://gyazo.com/2f6f613599f91d710ecae35413fe8c72">Image from Gyazo</a>


a18abuah:
The bugg is now fixed! The rezizing of template 9 works!

PR #8380, opened 2020-04-22
Logging for save and upload in showDugga

a18jakme:
Logging for save and upload a file in dugga.

Upload file dugga - Logs a userid and a description with filename and filetype
Save dugga - Logs a userid and a description with courseid, duggaid, moment and an answer.

image

For testing:
image

co author @b17carbr 

When upload a file, there is an issue with a "bummer" message. However, the log still works and the file is getting uploaded.  
image

### DIFF
--- a/DuggaSys/filereceive_dugga.php
+++ b/DuggaSys/filereceive_dugga.php
@@ -418,6 +418,8 @@ if($ha){
 								 	else echo "Type \"$filetype\" not valid for file extension: \"$extension\"" . "\n";
 									$error=true;
 								}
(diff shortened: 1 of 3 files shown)


a18filkr:
bild
On local testing for windows, uploading a file only gave a white page but the logging seemed to work without a description (see row 2 id 782 on screenshot)
Saving dugga works with description as show on row 1.

a18filkr:
bild
On testserver logging works for both upload and dugga saves.
Error message shows for uploading but as shown in screenshot the filenames are logged.

PR #8384, opened 2020-04-22
The addition of a file-preview window in fileed, fix for issue #6938

a17towan:
You should now be able to preview files in fileed.php by clicking on the **filename** in the list. A fileview window should open containing a preview of the file. 

image

Old/non-existent png and jpg files should not be viewable and the text content should describe that the file is unavailable.

The fileview window is adapted to mobile-view.

Problems with previewing files while testing on the webserver from last year should now be fixed.


a17towan:
**To the reviewer**: When testing on the webserver, try and see if you can upload a file yourself and then preview it.

PR #8386, opened 2020-04-22
Changed all current logUserEvent to log userid instead of username - G1-2020-W18-ISSUE#8337

b17carbr:
All logging using logUserEvent now instead use userid. These change can be seen in the userLogEntries table.

Previously:

Screenshot 2020-04-22 at 10 39 34

After:

Screenshot 2020-04-22 at 11 45 54

These changes encompasses:

* Editing/creating courses in courseed:
* Adding items in sectioned 
* Adding files in fileed 
* Login/Logout/LoginFail
* Reseting passowrd
* Viewing examples in codeviewer 

---
How to test: 
Download https://sqlitebrowser.org/dl/

Screenshot 2020-04-22 at 11 55 52

### DIFF
--- a/DuggaSys/codeviewer.php
+++ b/DuggaSys/codeviewer.php
@@ -190,10 +190,12 @@ Testing Link:
 
 			// Logs users who view example, along with the example they have viewed
 			if ($userid == "00") {
-				logUserEvent($username, EventTypes::DuggaRead, $exampleid." ".$courseID." ".$cvers);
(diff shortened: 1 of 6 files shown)


a18filkr:
bild
Works for editing courseed
bild
Works for adding item in sectioned
bild
Works for fileed
bild
Works for logout

Testing for password reset on server cannot be completed, will test locally.

bild
View examples works.

a18filkr:
bild
Test for reseting password locally works! As does everything else.

Issue #7543, opened 2019-05-27
UML lines disappear when exporting SVG


c17matjo:
When exporting to SVG, recursive relations aren't drawn properly. 

Screenshot of diagram:
![image](https://user-images.githubusercontent.com/49195031/58403707-a59a9600-8063-11e9-9f35-b2e8c7dbfc00.png)

SVG export:
![image](https://user-images.githubusercontent.com/49195031/58403734-b21eee80-8063-11e9-9334-c70991e4da59.png)

a17jonro:
Svg export doesn't seem to work with uml lines at all.

Diagram screenshot:
![image](https://user-images.githubusercontent.com/49195133/58408503-39715f80-806e-11e9-814b-d8feeaf3ff09.png)
Svg export:
![image](https://user-images.githubusercontent.com/49195133/58408538-4f7f2000-806e-11e9-8fc7-3bafb5f02111.png)

b17albbe:
I think this has to do with the symbolToSVG function, in the diagram_objects.js file. Including the UML line with an additional if statement will draw out the line, but just as a straight line.

a18jonpe:
As previously mentioned, copying the code for ER-lines makes the lines appear but the breakpoints are not there. Trying to figure out how to get those in as well. They are calculated in drawUMLLine in diagram_object.js, so that's where I'm looking to try to figure it out. I'm also looking at the text file import and export since that works like it should

a18jonpe:
I'm having a hard time figuring out how to break the line. One possibility is to draw several lines, but I feel like that wouldn't be that easy since the symbolToSVG function would need to be looped somehow. Since the svgPos variable only has properties for start point and end point I feel like either the whole thing needs to be reworked to support more points or the function needs to loop in some way, but it feels like that would also require quite a lot of changes in existing code.

PR #8390, opened 2020-04-22
Exporting as SVG now handles UMl lines

o18marka:
Solves #7543
When exporting diagram to SVG the UML lines should now be displayed, and in hopefully all cases be displayed exactly how they appear in the diagram.

When testing, please be sure to clear localstorage first!


o18marka:
Edit: due to how SVG export currently works, the captured area will always be from from camera-position 0,0 even though you've moved the camera. This messed up the use of breakpoints and was solved by simply moving the camera to position 0,0 before SVG export. Maybe not the best solution, but i believe a new issue should address this if needed.

Issue #7690, opened 2019-06-11
[live_version] logging out users makes you loose progress on grading view


andras-marki:
When you are logged out when grading an assignment, the comments are lost.

This is irritating for the users. It is also so that the page doesn't list that you are logged out (it shows you are still logged in), so it is quite a nuissance.

b18jonli:
@andras-marki As of pull request #8474, users who are logged out are redirected to the start page. While I wouldn't call this a solution to this issue, at least it makes the user aware of the logout when it happens. 

Is this enough to close the issue or do you have some other solution in mind?

andras-marki:
At least that is some form of feedback, which is good. Thanks @b18jonli.

Probably a complete solution would be to have a pop-up asking if you are still active and want to keep your progress.

PR #8404, opened 2020-04-22
#7690 Solved issue where written feedback gets deleted by logout.

b18fella:
Solved the issue by storing the text in the feedback textarea to localstorage when the session expires. And later retrieve the text when opening the feedback textarea again.

Also solved the problem with the page not reloading when getting logged out.

I wasn't able to properly test these changes. So be extra hard on the review to be sure that it all works as it's supposed to.


a18eripi:
Still having the same issue @b18veran was talking about regarding feedback not being saved in local storage during logout. Was this fixed? I looked at the code and it looks good, but its buggy and sometimes saves and sometimes not. Is this local for me or happends for you as well?

b18fella:
I think that is a local bug, would be great to test the code on the weekly branch and see if the same problems comes up when it's on the server. @a18eripi

Issue #8341, opened 2020-04-21
Rework the preview for Zip / Rar file download in showDugga.php


a18oscte:
![hfgkhfghfh](https://user-images.githubusercontent.com/49142069/79859791-d8632080-83d1-11ea-9d3c-96a40a71a162.png)
![22222222](https://user-images.githubusercontent.com/49142069/79859801-dac57a80-83d1-11ea-8a87-f1cc0fca589f.png)
When clicking the link to download the uploaded Zip / Rar file a big preview popsup over the whole page. Maybe rework it so you only need to click the link to download it.

HGustavs:
Yes this preview is more than useless.

I think that this has been an issue numerous times before....

b18jonli:
Merged to master.

PR #8424, opened 2020-04-22
#8341 Removed preview window for Zip/Rar files.

a18matna:
Changed so when you click the link for Zip/Rar files you download it immediately instead of opening a preview with a download link.

Changed the CSS to make the link look more in place.

### DIFF
--- a/Shared/css/dugga.css
+++ b/Shared/css/dugga.css
@@ -356,6 +356,9 @@ svg text {
 .previewTable tr:nth-child(even){
   background-color: #dad8db;
 }
+.previewTable a{
+	color: var(--color-text-label);
+}
 @media only screen and (max-width: 500px) {
 	.previewTable {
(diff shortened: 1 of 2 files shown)


c14jonfr:
Where/How can I test this?

I've just had a quick glance so far, but the behaviour seems to be the same for the issuebranch and the groupbranch, but I could very well be doing something wrong.

a18oscte:
To test this you can log into a99marjo with password: password. Then go to "Rapport inlämning" in "datorgrafik". Then its just to click the link for the zip file and it should download immediately. if it doesnt, try to clear your cache and refresh the page to be sure you have the newest version.

a18matna:
The difference should be that in the Original theres a large gray "preview-window" popping up with the clickable download link inside, whereas now u should get the download by just clicking the zip/rar lnk

PR #8428, opened 2020-04-22
FIxes #8427 - Auto focus in appearance menu

### LINKED ISSUE #8427: Auto focus name input in appearance menu
When opening the appearance menu for objects that have the name input field, it would be nice if the name input field was in focus automatically. This would prevent the user from having to move the mouse and select the input. The name input is arguably the most used option in the appearance menu so having this auto focused would be helpful. The user could start writing the name directly after opening the appearance menu decreasing the number of steps needed to do so. In addition, the name input is always the first input in the form, which will make it easy to tab through the other options and edit them without using the mouse.

Entities, relations, attributes, and UML classes have names, so this applies to all of these. Then for text objects it would be a nice addition to have the textarea for text rows in auto focus. For lines the type option could be auto focused since all lines start with that. For free draw objects the fill color could be auto focused.

a18antsv:
This adds auto focus functionality to the appearance menu when it opens. This makes it easy to start typing or selecting directly with the keyboard instead of having to manually select the input with the mouse first. It is also easy to tab through the inputs in the form as the first option in the loaded form will always be auto focused.

The name input will be automatically selected for entities, relations, attributes, and UML classes (where they exist).
The type attribute will be automatically selected for all types of lines since that is the first and most important option for lines.
The textarea will be selected for text objects.
The fill color option will be selected for free drawn objects.

Test so all of the conditions above works as intended.

Link: http://group4.webug.his.se:20001/G4-2020-W18-ISSUE%238427/DuggaSys/diagram.php

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -4595,6 +4595,7 @@ function loadAppearanceForm() {
             const entities = connections.filter(symbol => symbol.symbolkind === symbolKind.erEntity);
             const relations = connections.filter(symbol => symbol.symbolkind === symbolKind.erRelation);
             typeElement.innerHTML = makeoptions("Normal", ["Normal", "Forced", "Derived"], ["Normal", "Forced", "Derived"]);
+            typeElement.focus();
             if(entities.length > 0 && relations.length > 0) {
                 document.getElementById("cardinality").innerHTML = makeoptions("", ["None", "1", "N", "M"], ["None", "1", "N", "M"]);
                 document.getElementById("cardinalityUML").style.display = "none";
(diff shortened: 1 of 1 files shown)


DuggaSys/diagram.js
```diff
@@ -4557,26 +4557,17 @@ function loadAppearanceForm() {
...
     //Undefined would mean the symbol is actually a path not having symbolKind, 0 is used as default for paths
     if(typeof type === "undefined") type = 0;
 
-    showFormGroups(type);
-
     const typeElement = document.getElementById("type");
     const nameElement = document.getElementById("name");
 
+    showFormGroups(type);
+    toggleApperanceElement(true);
```
a18antsv, line 4567: Needed to be toggled before fields in the form could be focused.

PR #8431, opened 2020-04-23
Logging for reset password in accessed

a18jakme:
Logging when a user reset password.
userid - who did the reset on the account
description - on which account

image

For testing:
image

co author @b17carbr


a18filkr:
bild
Test on local windows machine. Logging shows up when reseting pw.

a18filkr:
bild
Test on server, shows the password reset on last line, eventtype = 20 which shows the reset worked here as well.
Works great!

PR #8444, opened 2020-04-23
G3 2020#8107

a18jenro:
Now one can press F11, F12, resize the browser window and sort the columns, and the filtering will remain correct.


a18jenro:
@a18mahkh also worked on this fix.

PR #8457, opened 2020-04-23
G1 2020 w18 issue#8361

a18filkr:
Fixed issue with overlapping not working on other than "code" documents.

Also added some transition and animation to the menu, both descriptiontext and the hovering on menubutton.
The resize should work on all templates and no matter which "window" you resize.


a18jakme:
The windows is not overlapping anymore. "Title" disappear with an animation, which is a good solution. 
Before minimize:
image


When minimize:
image

The solution works on every template and its tested local and on testserver.

PR #8486, opened 2020-04-24
Solves issue#8478

b18simla:
Default paper size set to A4. To test this, go to http://group4.webug.his.se:20001/G4-2020-W18-ISSUE%238478/DuggaSys/diagram.php and navigate to View > Paper size...


a18antsv:
It could be intresting to store A4 in localStorage, both that it was activated and which paper size was chosen to restore this information onload. That's for another issue.

PR #8496, opened 2020-04-24
Added aria-live such that hint is read by screen reader

a18jenro:


a18jenro:
To test one has to download a screen reader and see if it reads the hint text that pops up in a div when searching in files.

Issue #8287, opened 2020-04-20
Error when trying to grade duggor that are grey.


b18fella:
When grading a grey dugga using the fist as shown below, you get the error message shown below.

![resultEdGrey](https://user-images.githubusercontent.com/49142651/79747727-7d1b2a80-830c-11ea-96c0-ac9a32480f28.PNG)
![resultEdGreyError](https://user-images.githubusercontent.com/49142651/79747728-7f7d8480-830c-11ea-95a6-4253c24d50f8.PNG)

This might be because there's no submitted dugga when it's grey. In that case, the option to grade should propably be removed.

c14jonfr:
I _think_ I know why this happens. The grey/striped "assignments" are not really tests/duggas (assignments), but instead moments (for example "Färgduggor", in this moment Färgdugga 1 and 2 are included but "Färgduggor" is not a dugga in itself). Meaning there is no template associated with them, but the system still tries to load it. You can see the file it is trying to load (not clearly visible in the screenshot but I see it in my console), it is looking for a .js file in the template folder, but the .js file has no name.

HGustavs:
This is some type of bug...

hard to classify which part is the bug.

It should not be possible to mark a course that has not been visited at all.

As soon as something has been handed in for some sub-item it should be mark-able.

Unclear if it is possible to know this without changing the information content ...

b18jonli:
Merged to master.

PR #8501, opened 2020-04-24
#8287 Fixed issue were duggas was unasigned still when a sub-dugga has been handed in.

a18matna:
Added a for loop to connect duggor with its sub-duggor.

Made a variable to save the placeholder grade. 0 if  a sub-dugga has been handed in and Null if no sub-dugga has been handed in.

Made changes to when how many times a dugga has been marked and time when submitted.


b18jonli:
I don't think I understand how #8287 is fixed by this, if one assignment of a moment is handed in, the fist still appears and if clicked generates the error from #8287. 
Screenshot 2020-04-27 at 8 14 32 AM
Screenshot 2020-04-27 at 8 14 42 AM

b18jonli:
I think your solution works according to what @HGustavs described, but it does not actually remove the error, I propose adding a few rows to remove the "fist"-element from moment-columns

PR #8509, opened 2020-04-25
Fixes #7677 - Changing appearance of different types of objects simultaneously

### LINKED ISSUE #7677: Problems with changing appearance for text and other object simultaneously
They don't have the same attributes so there are some null values, which causes problems. 
Works for global appearance but not for "change appearance".

a18antsv:
This is a big rework for issue #7677 so it should be tested thoroughly.

Before, it was difficult changing appearance of different types of objects simultaneously. It only worked decently when the selected objects all had the exact same properties that should be affected. It was impossible to change the names of objects to different things at the same time, they would all get the same name. The same goes for textareas, all would get the exact same text. I imagine these as being properties that should be different. The appearance form would before only show the inputs that should be shown for the previously selected object and ignore all the other selected objects needs. This was not a good way to handle it.

In this rework all types that have the exact same properties in common are grouped under one collapsible menu. If only one type of object is selected only one collapsible will be shown for that object. If other type of objects are selected at the same time, all common properties will be under one collapsible while all the dissimilar properties gets into separate collapsible for the type. This makes it possible to always make changes to everything in one appearance menu, without limitation. It is possible to add many objects of every possible type, select them all, and open the appearance menu and still be able to make correct changes to everything. On top of that it is also possible to give different selected objects different names in the same appearance menu, the names are split up by a comma. Editing on one side of the comma will affect one object, while the other side will affect another. The same works for textareas where a comma plus new line works as a separator. 

- Open the appearance menu with a few different types selected of objects selected and see so styling works for everything. Try a few times with different types of objects.
- Test to create every type of object (entities, attributes, relations, UML classes, paths, text, UML lines, er lines between entity-relation, er lines between attributes and entities/relations) and open the appearance menu with everything selected at the same time. Try to change all types of styling.
- Try to edit name and textareas that are separated by a character to see if it the text applies to the right objects.

Link: http://group4.webug.his.se:20001/G4-2020-W18-ISSUE%237677/DuggaSys/diagram.php

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -4600,6 +4600,21 @@ function createCardinality() {
     }
 }
 
+function createCollapsible(formGroups, types, index) {
+    const collapsibleElement = document.createElement("div");
+    const objectTypesElement = document.createElement("div");
+    const iconContainer = document.createElement("div");
+    const icon = document.createElement("div");
+    const title = document.createElement("div");
+    const formGroupContainer = document.createElement("div");
+
+    collapsibleElement.appendChild(objectTypesElement);
+    collapsibleElement.appendChild(formGroupContainer);
+    objectTypesElement.appendChild(iconContainer);
+    objectTypesElement.appendChild(title);
+    iconContainer.appendChild(icon);
+}
+
 function loadGlobalAppearanceForm() {
     showFormGroups(-1);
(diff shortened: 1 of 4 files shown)


a18antsv:
I am aware of cardinality not working properly for UML. I will fix this in a separate issue. So merge this if no other bigger problems come up.

PR #8530, opened 2020-04-27
G3 2020#8397

b18emigu:
Updated access levels for the `fileed.php` file. We have updated the database with 3 users for testing purposes.

Dont forget to take a look at the code as the **supervisor** role hasnt fully been implemented just yet and weve just implemented simple checking for future functionality.

brom = **superuser**
teacher2 = **supervisor**
teacher3 = **write access**
teacher4 = **studentteacher**

In case anyone has reinstalled the database for their branch you might have to set these yourself.

For testing use the course `Webbprogrammering` as thats the only course weve set the permissions for in the database. (If they have been wiped you can choose whatever course you want)

### DIFF
--- a/DuggaSys/fileed.js
+++ b/DuggaSys/fileed.js
@@ -249,8 +249,10 @@ function renderCell(col, celldata, cellid) {
     }
 
     if (col == "trashcan") {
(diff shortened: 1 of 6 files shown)


a18felen:
Only super users can now upload global files

PR #8543, opened 2020-04-27
Fixes #8513 - Copy path objects

### LINKED ISSUE #8513: Cannot copy Path objects
Paths cannot be copied. When a path is copied, an attempt of creating a symbol of it is made. Calling the copySymbol function should be prevented for all selected path objects to prevent bad symbol objects wasting memory. A new function for copying paths should be created and used on paths instead, to handle this correctly.

a18antsv:
This pull request will make it possible to correctly copy and paste path symbols. Before paths were treated like symbols which would create a dummy symbol wasting memory.

- Test to create several paths, select them, copy them and paste them.
- Create other type of objects, select them at the same time as paths and copy and paste everything.

Link: http://group4.webug.his.se:20001/G4-2020-W18-ISSUE%238513/DuggaSys/diagram.php

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1122,6 +1122,12 @@ function copySymbol(symbol) {
 function copyPath(path) {
     const clone = Object.assign(new Path, path);
 
+    const oldPointIndexes = clone.segments.reduce((result, segment) => {
+        result.push(segment.pa);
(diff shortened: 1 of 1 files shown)


a18antsv:
I will fix copySymbol to be more similar to copyPath which will solve this. This could be merged because of this in my opinion.

PR #8548, opened 2020-04-27
Added functionality to press CTRL to toggle fab button clickable

a18filkr:
In fileed, you should be able to press "ctrl" to make the FAB-button transparent and be able to click through it, this was you can click the buttons in the table behind it.


a18filkr:
bild

This is what it looks like when it's "transparent" and not clickable.

d17marle:
Working as intended localy and on the test server!
image
_Figure1 (before pressing CTRL)_

image
_Figure 2 (after pressing CTRL)_

PR #8562, opened 2020-04-28
Added unique IDs to the checkboxes and connected the labels correctly

b18jonli:


b18jonli:
First names didn't work previously either, this issue is just about the correct checkbox being selected by a click on the label, it shouldn't change the sorting.

Issue #8425, opened 2020-04-22
Hexadecimal input box is bloated in "Biträkningsdugga 1 & 2"


b18alegu:
The hexadecimal input box on **"Biträkningsdugga 1 & 2"** looks bloated compared to the more pleasant look of the same editor in **"Färgdugga 1 & 2"**.

**Biträknigsdugga**
<img width="1151" alt="Skärmavbild 2020-04-22 kl  16 40 20" src="https://user-images.githubusercontent.com/62876260/79996988-5c440800-84b9-11ea-92cd-ecad6ade1998.png">

**Färgdugga**
<img width="1151" alt="Skärmavbild 2020-04-22 kl  16 38 16" src="https://user-images.githubusercontent.com/62876260/79996952-53533680-84b9-11ea-9608-b81d0addaede.png">

Both examples are made using the exact same window width and both screenshots have the same size.

The size of both editors depends on the width of the window, which seems unnecessary and can look ugly. I think the buttons should just be slightly wider than the the current height to look the best. Plus it also makes a below miniscule contribution to fight [musarm](https://en.wikipedia.org/wiki/Repetitive_strain_injury).

b18jonli:
Merged to master.

Issue #8298, opened 2020-04-20
Save/Reset buttons in a bad place on some "duggor"


b18alegu:
On some duggor like "Biträkningsdugga 1 & 2" and "Färgdugga 1 & 2" in the "Webbutveckling - datorgrfik"-course, the **Save** and **Reset** buttons will hover above the hexadecimal-panel when activated. It behaves in the same way on both Chrome and Firefox.

<img width="1558" alt="Skärmavbild 2020-04-20 kl  15 02 08" src="https://user-images.githubusercontent.com/62876260/79755890-1997f980-831a-11ea-971f-50bc20cbf5c6.png">

I think a much better solution would be to place the  **Save** and **Reset** buttons in relation to the bottom of the screen instead so they won't obfuscate the hexadecimal-panel unless the window is really small. Care has to be taken to ensure it works well on mobile devices as well.

HGustavs:
Im thinking its a styling issue

if the upper box is bigger it should automatically displace buttons unless something is awry

HGustavs:
Unless conviced otherwise I consider this a bug

HGustavs:
we dont want fixed pos buttons

a18viksa:
While testing I found this bug as well. Don't really know what to do about it. 

![issue](https://user-images.githubusercontent.com/49142704/79986537-804d1c80-84ac-11ea-9fca-646bf8beeda4.gif)

a97marbr:
So the accordion widget seems a bit off ... yes I think you should investigate this ...

a18viksa:
Wouldn't it be a good idea to just ensure that the panel always follows the accordions movement? Since everything else does. I've tried this in the development tool and it kind of works, but not really in a good way. @a97marbr

a97marbr:
> ...kind of works, but not really in a good way
Examples? What is missing, what is not working in a good way? Screenshots?

a18viksa:
By removing the "top" from the element style, the problem goes away. 

![image](https://user-images.githubusercontent.com/49142704/80141237-8a087a00-85a9-11ea-80ff-1bf5d07211e3.png)

Although it resets after every time you close the hexadecimal box. With "top" removed it looks like this. And I guess this is the desired behavior. @a97marbr 

![issue2](https://user-images.githubusercontent.com/49142704/80141368-bd4b0900-85a9-11ea-897f-de720cf63eb8.gif)

a97marbr:
Yes this is the desired behaviour

[...]

PR #8564, opened 2020-04-28
G2 2020 w18 #8425

b18alegu:
Changed the width of the hexadecimal input box on color dugga and bit dugga.
- Usually the width is much lower now which make it look more nimble
- Minimum width was set higher though since it looks really stupid when the width is too small

To the this login as `Toddler`, use password `Kong` and go to the webbutveckling course and take a look on the relevant duggor.


b18alegu:
Issue https://github.com/HGustavs/LenaSYS/issues/8298 deals with the placement of the `save` and `reset` buttons and are currently being worked on by @a18viksa. I agree though that it is unfortunate that the buttons are placed "above" the hexadecimal box.

PR #8587, opened 2020-04-28
Resulted sorting checkboxes are now properly updated when sorting

b18jonli:
The sorting of Fname/Lname with the use of the checkboxes is still wrong however, but is a separate issue. 

(Ignore the merge commit in the history, the merged branch is already in G2-2020-W18)


b18jonli:
Exactly, that is because the status variable is set to three to set “pending” but this instead represents Lname:Asc when sorting by name since pending doesn’t exist for fname.

PR #8589, opened 2020-04-28
Fixes #8462 - Copy UML-class objects

### LINKED ISSUE #8462: Can't copy UML-class objects
UML-Class objects does not seem to appear when copy-pasting them.

a18antsv:
This reworks the copySymbol function. The function is now quite general so it would work to add new points to objects and just add one line in the function to make it work with that as well. A general check was done to see if an already processed point have the same old point as the point currently being processed. If both points have the same old point only one new point will be created and reused for the other points having that same old point.

This rework should fix both UML and UML-lines. The offset newly copied symbols are created at should also be the same for both paths and symbols, moving down to the right, which makes more sense when the origo is in the top of the screen by default.

- Create all types of objects, copy and paste them individually and together at once.
- Create an ER diagram connected with objects connected with lines, try to copy and paste the objects together with the lines.
- Create an UML diagram and connect classes with lines, try to copy and paste. 

Link: http://group4.webug.his.se:20001/G4-2020-W18-ISSUE%238462/DuggaSys/diagram.php


a18antsv:
I am aware of recursive UML-lines not working exactly as expected. But all other type of lines also seems to be working. Recursive UML can be fixed in another issue as this was not part of this issue to begin with.

Issue #8522, opened 2020-04-27
Make course MOTD look like "Server" MOTD


a18jonkl:
There is a need for more consistency between the MOTD windows. My opinion is to make the course motd look the lite server MOTD.

Server MOTD:
<a href="https://gyazo.com/f740490696ce64385247dcb204918ab2"><img src="https://i.gyazo.com/f740490696ce64385247dcb204918ab2.png" alt="Image from Gyazo" width="1397"/></a>

Course MOTD:
<a href="https://gyazo.com/63a7948a44672f8c27a6c31d41c70496"><img src="https://i.gyazo.com/63a7948a44672f8c27a6c31d41c70496.png" alt="Image from Gyazo" width="1392"/></a>

PR #8604, opened 2020-04-28
G3 2020 #8522

a18lukto:


a18jenro:
This seemed to cause the issues that we have been trying to solve. We will wait with this PR until after the weekly merge.

Issue #7802, opened 2020-04-02
TypeError when accessing course as "teacher1"


b18jonli:
<img width="550" alt="Screenshot 2020-04-02 at 2 03 08 PM" src="https://user-images.githubusercontent.com/62876553/78247962-12848500-74ec-11ea-9146-443083cc1215.png">

url:
/DuggaSys/sectioned.php?courseid=2&coursename=Webbutveckling%20-%20datorgrafik&coursevers=97732

a18jonsj:
@HGustavs @a97marbr Should teacher 1,2,3,4 have access to the navheader like "brom" does, or is it enough to make the typeerror go away to fix this issue?

a18jonsj:
We managed to create a soloution that removes this error. In section.js if a user has access its supposed to create a dropdownmenu, but in navheader.php the dropdown is not created for users with writeaccess. So we added the dropdown in navheader. 

Is it correct that user with "writeaccess" should be able to see the dropdown? 

The dropdown:
![image](https://user-images.githubusercontent.com/49142617/79871042-1b2df400-83e4-11ea-88ef-a45ac4abd449.png)

HGustavs:
Only makes sense for write access users?

Or all teachers?

Definitely not for students. I am sure of that but for the rest of the user options, I am more unsure.

a18jimol:
In the code sectioned.js the code write out the different types of course-versions but don't write out the actuall dropdown in havheader.php so from our view teachers/users with write-access should be able to see the dropdown.

For teachers/users with studentteacher-access can see the dropdown but the code don't generate any possible information of course-versions to the dropdown.

So then this should be an good solution for write-access users because they should be able to see the dropdown

a97marbr:
In LenaSYS we have 4 different levels of access.
1. Superuser - the highest level of user can access/see everything / add new courses
2. Write access - Can change / add course versions / update courses, grade duggas in courses ... I think that write access is (should be?) per course ... 
3. Read access - Can view and submit duggas in courses
4. Guest - Can view courses but not submit

PR #8633, opened 2020-04-29
The extra course version dropdown is now removed

a18felen:
Removed the extra course version dropdown that did not work and only showed for people with write access. The error that this was supposed fix in issue #7802 no longer remains but should be tested by the reviewer just to make sure.


a18jenro:
You've done great work son!

Issue #8381, opened 2020-04-22
Sorting and filtering together does not work in files


a18jenro:
If you have filtered in files on say "course local", if you were to search after that, then all files are searched, not just course local.

![untitled](https://user-images.githubusercontent.com/62876507/79964484-17a17800-848b-11ea-867b-5b2f3f6615d9.PNG)

a18lukto:
This issue have encounters som problems, trying to merge it back in to the weekly branch to fix it

PR #8636, opened 2020-04-29
#8381 branch is replaced

a18mahkh:
The old branch is replaced with a new one because there was some problems with merging into the weekly-branch


a18jenro:
Looks very good!

PR #8639, opened 2020-04-29
G3 2020#8437 branch is replaced

a18mahkh:
The old branch is replaced with a new one because there were some problems with merging into the weekly-branch.


a18jenro:
Very gut!

PR #8642, opened 2020-04-29
Removed 2 comments

a18jonkl:
Deleted 2 comments.
`$("#cookiemsg").css("display", "none");`, seem to be fixed by the code at 250-255.


a18jenro:
Not the most detailed cleaning but we will let this slide

PR #8646, opened 2020-04-29
Added php-tags and changed the message to display group owner

### LINKED ISSUE #8324: No ownership popup in installer
The ownership popup doesn't show automatically and cannot be opened manually, this error appears on load:
Screenshot 2020-04-21 at 10 37 13 AM

### LINKED ISSUE #8644: Installer start-popup shows fileowner but the command is to set filegroup
The command shown changes the group owner of the file, which should also be the value displayed, as it is currently the values won't ever match even if the correct permissions have been set.

b18jonli:
Fixes #8644 and fixes #8324

### DIFF
--- a/install/install.php
+++ b/install/install.php
@@ -68,12 +68,11 @@
             "!!!!!!READ THIS BEFORE YOU START!!!!!!</h1><br>" +
             "<h2>Make sure you set ownership of LenaSYS directory to 'www-data'.<br>" +
             "current owner: " +
-            if(function_exists('posix_getpwuid')) {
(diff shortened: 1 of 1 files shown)


a18ismca:
Before I was in the same branch as you were when editing this, no pop-up window showed up... but when I am in the branch a pop-up window is shown...

b18jonli:
You can try to checkout this commit 22387c58765882cf71449610dbdaf0732ae18322, it should be as the installer was in the beginning of the course, my guess is that (on windows) this doesn't show the name either.

PR #8647, opened 2020-04-29
Added php-tags and changed the message to display group owner

b18jonli:
Same changes as pull #8646


a18ismca:
image
This is how the message looks like after you have made changes. The message shows current owner and how to change group in linux-command. As I can see there is no group owner as it seems right now. I don't know if this is the purpose of this pull request, so I would be good if i "double-checked" with you if the picture shown above is the objective of this pull request, before I approve this.

Otherwise, I think you made a great job...

b18jonli:
Are you sure that the file has a set group? Also, how did this work on windows previously? Did it show the file owner?

Issue #8577, opened 2020-04-28
Add ~/ to .gitignore


a18marax:
For linux users this will be very helpful since that folder will always be updated and shouldn't be pushed.

HGustavs:
Yes no problem.

Make sure this has no other negative effects.

b18jonli:
Alternatively, if this is a problem created by the location of the LenaSys files locally and not for all linux users, you could add it into you .git/info/exclude file. This fixes the problem locally but won't be synced so it cannot affect other participants negatively. (e.g. if someone has lenasys files in their home folder)

PR #8655, opened 2020-04-29
Added the path of ~/ to .gitignore

a18marax:
This solves issue #8577

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,7 @@ sftp-config.json
 .idea
 vendor/
 *.iml
-~/
+*~/
 
 
 # Visual Studio 2015/2017 cache/options directory
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ sftp-config.json
 .idea
 vendor/
 *.iml
+~/
 
 
 # Visual Studio 2015/2017 cache/options directory


a18marax:
As previously discussed the path is not absolut when there is a folder named accordingly, since git checks the folders in which are added and not the folders higher up in the structure this will never be a problem. Also it is not possible to run LenaSYS from the home folder due to the installation process in Linux, for this reason also this will not be a problem. It is true that ~/ is the path to the home folder of the current user but it's is also a path to some sort of backup system for Visual Studio Code in the root directory of the project which should not be included in future updates of LenaSYS, this update would eliminate the risk of this happening by mistake and save Linux users the headache of remembering it and also the first time noticing it and understanding that it's not suppose to be pushed.

b18alegu:
> "It is true that ~/ is the path to the home folder of the current user but it's is also a path to some sort of backup system for Visual Studio Code in the root directory of the project which should not be included in future updates of LenaSYS..."

If I understand you correctly @a18marax, maybe `*~` would be a safer solution to this particular problem since that is a common suffix for temporary files. It can be found in the example [Linux.gitignore](https://github.com/github/gitignore/blob/master/Global/Linux.gitignore) page.

I've googled this issue pretty extensively and I haven't found a single example using `~/` inside the `.gitignore`-file, so I don't think this solves a common problem for Linux users.

a18marax:
Actually it shouldn't matter since .gitignore treats them the same and still the only place tou csn run LenaSYS from on a Linux machine is /var/www/html/ it won't affect anything in the home directory, but if you want I can make the change doesn't matter one way or the other. 
This should however be a problem for anyone using Linux working on this project, the problem here is LenaSYS not allowing the developer to install LenaSYS where ever they please it will stop the installation if not at /var/www/html/.

b18alegu:
> Actually it shouldn't matter since .gitignore treats them the same and still the only place you can run LenaSYS from on a Linux machine is /var/www/html/ it won't affect anything in the home directory, but if you want I can make the change, doesn't matter one way or the other.
> This should however be a problem for anyone using Linux working on this project, the problem here is LenaSYS not allowing the developer to install LenaSYS where ever they please it will stop the installation if not at /var/www/html/.

Just for clarification: `/var/www/html/` is the default root folder of a web server running Apache on Linux. It can be changed in the `apache.conf` file by changing the `DocumentRoot` attribute.

a18marax:
> > Actually it shouldn't matter since .gitignore treats them the same and still the only place you can run LenaSYS from on a Linux machine is /var/www/html/ it won't affect anything in the home directory, but if you want I can make the change, doesn't matter one way or the other.
> > This should however be a problem for anyone using Linux working on this project, the problem here is LenaSYS not allowing the developer to install LenaSYS where ever they please it will stop the installation if not at /var/www/html/.
> 
> Just for clarification: `/var/www/html/` is the default root folder of a web server running Apache on Linux. It can be changed in the `apache.conf` file by changing the `DocumentRoot` attribute.

Yes, but LenaSYS won't install anywhere else on a Linux system

b18alegu:
> > > Actually it shouldn't matter since .gitignore treats them the same and still the only place you can run LenaSYS from on a Linux machine is /var/www/html/ it won't affect anything in the home directory, but if you want I can make the change, doesn't matter one way or the other.
> > > This should however be a problem for anyone using Linux working on this project, the problem here is LenaSYS not allowing the developer to install LenaSYS where ever they please it will stop the installation if not at /var/www/html/.
> > 
> > 
> > Just for clarification: `/var/www/html/` is the default root folder of a web server running Apache on Linux. It can be changed in the `apache.conf` file by changing the `DocumentRoot` attribute.
> 
> Yes, but LenaSYS won't install anywhere else on a Linux system

But isn't that because the **LenaSYS install** is run by the web server?

a18kaswi:
> Actually it shouldn't matter since .gitignore treats them the same and still the only place you can run LenaSYS from on a Linux machine is /var/www/html/ it won't affect anything in the home directory, but if you want I can make the change, doesn't matter one way or the other.
> This should however be a problem for anyone using Linux working on this project, the problem here is LenaSYS not allowing the developer to install LenaSYS where ever they please it will stop the installation if not at /var/www/html/.

This is interesting since the lenasys at testserver is located at /var/www/group4 for example. Can you provide screenshots/gif of the installer halting?

b18alegu:
Since the issue is now closed I'll remove the pull request as well.

PR #8659, opened 2020-04-29
G3 2020#8580 now you can view statisticscontent in mobile-view

a18mahkh:
To test this:

Go to mobile mode and then a arrow should appear on the course header. 

Anteckning 2020-04-29 131320

Click on that arrow and the statistics content should slide down.


a18jenro:
In desktop-view, when making the window smaller, the button for statistics used to disappear, but now it remains, but it does not work.

This causes some weird issues.

PR #8663, opened 2020-04-29
Fix for buttons in analytics

a18filkr:
Width fixed for text and centered buttons.


a18filkr:
Also added flex-wrap for responsiveness when resizing.
bild

PR #8680, opened 2020-04-29
Updated function "updateContent" within codeviewer.js - G1-2020-W19-ISSUE#8679

a18geoge:
Updated function "updateContent" within codeviewer.js to better fit the code standard. 

Test by going to codeviewer.php and navigating to the "Edit Content" window by pressing the small cogwheels and edit the examples in different ways (title, wordlist, file etc), edit the examples through the "Edit file" (the small pencil) icon. Check if for example the different exampleid's give the corresponding examples and titles (e.g. by changing the number in the URL, LenaSYS/DuggaSys/codeviewer.php?exampleid=1&courseid=1&cvers=2013 where exampleid would be changed to =2).

### DIFF
--- a/DuggaSys/codeviewer.js
+++ b/DuggaSys/codeviewer.js
@@ -668,7 +668,8 @@ function editImpRows(editType)
 // updateContent: Updates the box if changes has been made
 //----------------------------------------------------------------------------------
 
-function updateContent() {
+function updateContent() 
+{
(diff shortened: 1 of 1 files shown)


a18geoge:
To be honest I'm not really sure how necessary of a change it was to fix the indentation but the code standard states that you should "Try to keep ajax or jquery calls on a single line".

Issue #8626, opened 2020-04-29
Logout button does not properly close when you press the "X" button


a18lukto:
![image](https://user-images.githubusercontent.com/49182858/80572345-68225380-89fe-11ea-8725-d9b96ba89ef6.png)
When you are logged in and press the logout button, it's not possible to close it at first using the "X" button or pressing escape. It will however work after you have pressed cancel once

PR #8684, opened 2020-04-29
Fixing the X button | Fix for issue #8626

a17towan:
Fix so that you can now close the logout popup window by pressing 'x' without having to press 'cancel' first.

Before this fix, you couldn't press the red 'X' button to close the window the FIRST time you opened the log out box. You had to press the 'cancel' button first, then open the logout box AGAIN and press 'X'. The second time you press 'X', it works. 

I've fixed it so that you can press 'X' the first time you open the logout box and it closes like it's supposed to do.

Press here:
image

This box:
image

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -1241,7 +1241,7 @@ input {
   font-weight: bold;
 }
 
-.loginBoxheader div:hover {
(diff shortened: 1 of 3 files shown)


a18jenro:
Seems to work very fine!

Issue #8632, opened 2020-04-29
ResultEd: Cell backgrounds are not filled properly


b18alegu:
In the result editor some of the diagonally striped backgrounds in the cells are not filled properly. This is probably because there is no date inside those cells.

An example: 
<img width="384" alt="Skärmavbild 2020-04-29 kl  10 11 33" src="https://user-images.githubusercontent.com/62876260/80574349-f21feb80-8a01-11ea-94e8-b4c6eab14fb1.png">

b18jonli:
Closing since this has been merged.

b18fella:
Not sure if this was really an issue to begin with. It would be appreciated if @HGustavs could explain how this is supposed to work and if the screenshot example posted above is how it's actually supposed to look like.

a18eripi:
To add on to what @b18fella said, it might just be that we made a fix for a problem that shouldn't have existed since its an assignment with a deadline set at timeZero. Clarification would be appreciated.

HGustavs:
Im thinking same size always ragrdless of whether we have a date or not.

b18fella:
@HGustavs So what you're saying is that it works as it should even if it doesn't have a date? In that case, this issue wasn't a problem, to begin with.

PR #8687, opened 2020-04-30
#8632 Added text for no deadline

b18alegu:
Added "No deadline given" to cells in the result editor when there simply is no deadline. Makes the cells in the grid at the same height and more pleasing to the eye.


b18alegu:
This pull request did not work as intended.

PR #8691, opened 2020-04-30
G3 2020#8652

a18jenro:
Now, when you sort (by column), filter (by kind) or search, the first column will always be counting from 1 to the total number of entries.


a18felen:
Looks good now, good job!

PR #8714, opened 2020-04-30
minWidth safeguard added, ISSUE#8594

a17robhe:
Added if statement and a absolute minimum value that minWidth can have to prevent it from shrinking too small to contain the text in the UML class.

Untitled


a17robhe:
For reference, this is the old incorrect behaviour
test4

The console output has been removed though

PR #8718, opened 2020-04-30
G4 2020 w19 issue#8716

a18kaswi:
Structured the code so its easier to see what's what. Also indented the code to follow the standard.

Have checked so no conflicts arises with other groups.


a18kaswi:
Test that the installer stil work on testserver and that the code looks good I assume!

PR #8726, opened 2020-04-30
ShowDugga: Added inline styling to position the panel

a18viksa:
The first idea was to adjust the code for the top positioning in dugga1.js and dugga2.js, but with javascript I couldn't get the top position to always be -35px. By just adding "top:-35px" to the div containing the hexadecimal values the problem was solved. 

Now it looks like it should and everything feels more connected. 

image

This fixes the issues in both the "biträkningsduggor" and the "färgduggor".

### DIFF
--- a/DuggaSys/templates/dugga1.html
+++ b/DuggaSys/templates/dugga1.html
@@ -61,7 +61,7 @@
 		</tr>
 </table>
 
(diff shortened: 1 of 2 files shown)


a18viksa:
There's a typo in the first commit. The change is in dugga1.html and not dugga1.js.

Issue #8703, opened 2020-04-30
Fix 'home' and 'back' icons


a17towan:
![image](https://user-images.githubusercontent.com/49142028/80688491-2bbd2900-8acc-11ea-952d-ee5f8ba291e1.png)

The 'home' and 'back' icons don't have a box-shadow effect when you hover over them like the other icons do. This should be an easy fix for consistency's sake.

PR #8729, opened 2020-04-30
Fixing shadows over 'home' and 'back' icons | Fix for issue #8703

a17towan:
Fixed so there is now a box-shadow effect on the 'home' and 'back' icons (see image for reference)

image


a18jenro:
Look good

PR #8734, opened 2020-04-30
solves issue#8435

b18simla:
Added option to show/hide toolbar inside fullscreen and modified the way elements are shown/hidden. Press T to toggle toolbar. Also adjusted some text in popup when entering fullscreen

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -3704,11 +3704,8 @@ function toggleFullscreen(){
             menu_buttons[i].style.display = "block";
         }
         tool_bar.style.visibility = "visible";
-        if(!toolbarDisplayed){
-            inside_toolbar.style.visibility = "visible";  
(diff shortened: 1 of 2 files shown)


DuggaSys/diagram.js
```diff
@@ -5242,55 +5294,23 @@ function loadAppearanceForm() {
...
-            indexes[key] = {
-                current: 0,
-                max: 0
-            };
-        }
-        indexes[key].max += value;
-    });
-
-    //Get all unique types of selected objects
-    const types = Object.keys(indexes).map(Number);
```
b18simla, line 5267: I'm not sure why all of this is shown. I haven't touches these lines... Branch says no conflict so should be fine?

a18antsv: This is absolutely not fine.

Issue #8710, opened 2020-04-30
Dugga Editor: renderCell-function cleanup


b18alegu:
The `renderCell`-function in `duggaed.js` is a bit messy. It has the following issues which should be cleaned up:
-Plenty of `if else`-statements. `switch...case` should be used instead
-Commented code can be removed
-Inconsistent commenting

Will also correct the indentation of the `setup`-function in the same file since it's stupid and annoying at the moment:
<img width="396" alt="Skärmavbild 2020-04-30 kl  11 40 17" src="https://user-images.githubusercontent.com/62876260/80696240-6aa4ac00-8ad7-11ea-8aa9-29f4175d2226.png">

b18jonli:
Closing since this has been merged.

PR #8735, opened 2020-04-30
G2 2020 w19 #8710

b18alegu:
This solves issue #8710. The `renderCell`-function has been trimmed and is now easier to follow and works in a more consistent way.

I have also changed the "Start date", "Deadline" and "Result dates" to show "N/A" inside its cells instead of null and also lessened the precision on the timestamps to exclude seconds.

When testing:
- Compare it to another branch.
- It should behave exactly the same except the changes to the three columns stated above.
- Using **Webbutveckling - datorgrafik** while testing is recommended.
- Don't forget to test the **Variant Editor** too (Clink on the pen icon)

The indentation of the `setup`-function was corrected as well...

### DIFF
--- a/DuggaSys/duggaed.js
+++ b/DuggaSys/duggaed.js
@@ -707,7 +707,9 @@ function renderCell(col, celldata, cellid) {
 		case "qrelease":	// DUGGA-TABLE - Result date
 			if(!celldata) {	// if null - return string "N/A"
 				retString = "N/A";
-			} else {		// else - return date withough seconds
+			} else {		// else - return date without seconds (i.e. last three charachters)
+				var secCutoff = celldata.length - 3;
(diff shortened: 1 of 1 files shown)


b18alegu:
> I wasn't able to test how (if) these changes affect the sorting, since this branch does not have the compare()-fix. As soon as this branch is merged to the group branch this should be tested and if any issues occur a new issue should be opened. **Note: if you're testing this and try to sort, it will fail because there is no compare()-function. If you do not clear the localstorage after this most things in the Dugga Editor will "break", such as creating new variants etc. (it does not actually break, the variants are created but they do not show up in the interface. Just a heads-up)**

This was news to me (not that it breaks but why). This behaviour is the same as before but still flawed of course. I agree that a new issue should be made after weekly-merge to fix this.

PR #8736, opened 2020-04-30
Solution for branch G1-2020-W19-ISSUE#8574

a18abuah:
The errors with the button hiding and not showing the description tag is now fixed.
Maximize:[Image from Gyazo](https://gyazo.com/0e5710e94fd1ff6ad0899812600e64eb)
Minimize: [Image from Gyazo](https://gyazo.com/ef19c3bc568b562b8d4d60c88fd7c319)
Reset:[Image from Gyazo](https://gyazo.com/9f9e5d92e6ef8fc924b8d18c0ea16189)

### DIFF
--- a/DuggaSys/codeviewer.js
+++ b/DuggaSys/codeviewer.js
@@ -2644,6 +2644,10 @@ function maximizeBoxes(boxid) {
 	//For template 6
 	if (templateid == 6) {
 		if (boxid == 1) {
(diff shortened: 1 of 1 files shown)


a18jakme:
Template with 2 windows and 4 windows does not show minimize buttons. Is that something with this solution or is that how it should be?

PR #8755, opened 2020-05-03
Fix for issue #8741

### LINKED ISSUE #8741: Bug when new UML-class is created, then line is drawn to incorrect object(UML to ER)
Something seem to go wrong when a new UML-class is created, then a line is drawn between a UML object to a ER-object(which should not be possible).
issue4

o18marka:
Fixes #8741
Fixed issue by changing selection that decides when UML-line should be drawn or not. Also moved when uimode is changed to outside the selection.

This fixes the bug with the odd line being drawn when incorrect line draw is attempted(ER to UML). This also fixes a other issue when drawLine would break and need to be toggled on and off for it to work again when incorrect drawLine is attempted. This is probably due to uimode before being toggled inside selection and would be unreachable.

Test by drawing a mix of correct and incorrect lines while also creating new objects.


a18antsv:
There was a conflict with weekly branch after a recent merge so quickly resolved this as the conflict was only because of a couple rows.

Issue #5617, opened 2019-01-24
Result view filter with exported dates


HGustavs:
We should be able to only export new entries.

 If a student has passed the course and that result has been exported we should be able to filter to only show/export new entries after that date (or some other export date).

a17adasy:
If I understand correctly, you want an option so you can only export the duggor and grades after a certain date?

b17nicsv:
@HGustavs Bump

HGustavs:
A date filter for tests that have been marked after a certain date.

b17nicsv:
Suggesting a TODO list:

- [Somewhere in resultEd let the examinator know when he/she last exported using ladExport, so he/she doesn't have to remember it]

- [Filter results from when the last export in the specific course was made]
OR
- [Filter results from when the last export was made using a dropdown list asking for a specific date and filter from that date to just now]

If I am not mistaken, in resultEd the time a test was submitted already has a get function 
```celldata.submitted.getTime()```
The filter can use this time and set a specific time since to see tests from "The specific date" to just now.

b18fella:
Things that are left of the suggested ToDo list are: 
- [Filter results from when the last export in the specific course was made]

OR

- [Filter results from when the last export was made using a dropdown list asking for a specific date and filter from that date to just now]

PR #8767, opened 2020-05-04
#5617 Added a way to see the date and time an export last was executed.

b18fella:
The last export date is stored in localStorage and the date in the localStorage is then shown under the LadExport button as shown below. Also added so that the last export date is shown in the footer of the popup.

expDate

The way to test this is to click on the "LadExport" button and then close the popup. Then you can see the date under the button and in the localStorage.

Co-Authored-By: b18veran <b18veran@users.noreply.github.com>

### DIFF
--- a/DuggaSys/resulted.js
+++ b/DuggaSys/resulted.js
@@ -37,6 +37,10 @@ function setup() {
     window.onscroll = function () {
 
 	};
+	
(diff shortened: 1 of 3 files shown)


b18veran:
I agree @c14jonfr the best thing is that it is stored in the database. We have had problems accessing the database. If someone finds a good solution, is it free to solve this issue.

PR #8777, opened 2020-05-04
Saved the correct version name when there are several with the same vers

a18jenro:
Now the correct versionname is shown in edit course version if there are several versions with the same version ID.


a18jenro:
To test, observere a verison Id in some course, create a new course version in another course with the same Id. Then, when pressing edit course version, observe that the correct version names appear.

Issue #8525, opened 2020-04-27
Clean dead code in SectionEd


a18jenro:
Clean dead code in SectionEd. If a function is not used, remove it.

a18oscba:
Apprently code is still there after merge.

PR #8790, opened 2020-05-04
G3 2020 #8525

a18lukto:
Cleaned dead code for SectionED booth the js and php file.


a18jenro:
Functionality seems to be unchanged, good job!

Issue #8761, opened 2020-05-04
Content in Transformationsdugga not positioned properly on smaller screen sizes


c14jonfr:
When doing a Transformationsdugga on smaller screen (or window) sizes the content is not positioned properly. The content should be positioned vertically in some way. Currently it causes the page to scroll sideways, which is almost never a good thing. Additionally, when the screen/window becomes small enough the instructions for the dugga disappears completely.

I believe positioning the content in a better way would make the dugga useable for small screen sizes.

![image](https://user-images.githubusercontent.com/49141758/80945278-be6f0800-8deb-11ea-831e-9bf5f5727184.png)

c14jonfr:
I've played around a bit with both the Chrome and Firefox developer tools and they seem to behave a bit differently compared to each other. For example the breakpoints where things start to look messed up are a bit different between the two browsers which of course makes it hard to judge how it will behave on real mobile devices.

b18jonli:
This has been merged, closing issue.

PR #8799, opened 2020-05-04
G2 2020 w19 #8761 Transformationsduggas should now be more useable on smaller screens

c14jonfr:
Solves #8761.

Transformationsduggor should now be more useable on small screens (<1050px wide). Before this change the content would overflow horizontally. When the screen is small enough the content will now be displayed in a column fashion and the horizontal scroll should be no more.

I also replaced the instructions window with an accordion (the one used in for example bitduggas and färgduggas) for consistency. The transformationsdugga uses an accordion for feedback anyway. This is always displayed on the top of the page which made positioning everything else easier.

**Note:** during development this was tested in both Chrome and Firefox Dev Tools. They did not behave the same way when resizing (for example the canvas would become very small when resizing somewhere below 1000px width in Firefox, this did not happen in Chrome, where the canvas was resized to a more viewable size). I chose to trust Chrome this time.

It should look something like this on small screens:
image

and something like this on bigger screens:
image

### DIFF
--- a/DuggaSys/templates/dugga4.html
+++ b/DuggaSys/templates/dugga4.html
@@ -4,19 +4,17 @@
 				<div id="feedbackTable" style="margin-top: -3px;"></div>	
 		</div>
 </div>
-<div id="container" style="display:flex; justify-content:center;">
-    <div id='duggaInfoBox2' style="display: inline-block; width:300px;">
-      
-          <div class='loginBoxheader'>
-                <h3>Dugga 4 - Transformationer</h3>
-          </div>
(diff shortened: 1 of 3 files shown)


c14jonfr:
@b18veran Yes, I noticed some issues with the Save/Reset buttons during testing as well. I decided to overlook that since there was already an open issue about their placement, hoping that the solutions would meld well together. If there are still issues with the Save/Reset buttons after the merge a new issue should be opened. Maybe it's specific for the Transformationsdugga? I'm not sure.

PR #8800, opened 2020-05-04
Solution for branch G1-2020-W19-ISSUE#8667

a18abuah:
Removed client start and end to show service event date.
[Image from Gyazo](https://gyazo.com/d6042884ffd23cf366591db25a0884e1)
When testing remove or comment out this line in service to make the service crash. [Image from Gyazo](https://gyazo.com/492406dfb152284ec5b5f8c32a467d59)


a18jakme:
We developers felt that ClientStart/End was not necessary to log because we already log the ServerStart/End. That is why it was not possible to print out the date of a ClientStart/End.
Maybe we should ask the customers if they want to have logs about the Client service.

PR #8820, opened 2020-05-05
Added a check to see whether the back button exists

a18jenro:


a18jenro:
To test, simply go to the start page: http://group3.webug.his.se:20001/G3-2020%238817/DuggaSys/courseed.php

There should be no console error there.

Issue #8771, opened 2020-05-04
Add new course version gives success message before error


a18jenro:
When adding a new course version of a course and giving it the same version ID as an existing one, a success message is given, but after that an error message and a console error occurs:

WHAT HAPPENS FIRST:
![untitled](https://user-images.githubusercontent.com/62876507/80948775-13624c80-8df3-11ea-9c52-b203bde99778.PNG)

WHAT HAPPENS AFTER:
![untitled2](https://user-images.githubusercontent.com/62876507/80948777-14937980-8df3-11ea-9f5d-1b80f9b7520e.PNG)

Clearly, the console error must be handled and the success message should not be displayed.

PR #8838, opened 2020-05-05
You can no longer create a new course version with an existing version ID | Fix for issue #8771

a17towan:
It is no longer possible to create a new course version with a version ID that already exists **for the selected course**. 

You can, however, create a version with an ID that exists in **another** course, but not in the same course. I.e. you can have the course version with ID "1337" in both IT118G and DV12G but you can't have **two** course versions with the ID "1337" in  IT118G.

In "**Webbprogrammering DV12G**" there is a course version by the id "1337". So when we try to add a new course version with the same ID, we get a warning and we are unable to create it.
image

image
 -----
In "**Webbutveckling datorgrafik IT118G**" we have a course version with the ID "1337", which also exists in "Webbprogrammering DV12G". This is allowed because they are two different courses with different course ID's.
image

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -1069,9 +1069,9 @@ function returnedSection(data) {
   addClasses();
   showMOTD();
 
-  for(var i = 0; i < data["versions"].length; i++){
+  /*for(var i = 0; i < data["versions"].length; i++){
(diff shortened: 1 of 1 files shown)


a18jenro:
Awesome feature!

Issue #8831, opened 2020-05-05
Change let to var [Group 2: Duggaed]


b18alegu:
According to the `code_standard.md` document, `let` should not be used for variable declarations for browser compatability reasons. Instead `var` should always be used instead.

For group 2 this concerns the following files:
- `dugga.js`
- `duggaed.js`

b18alegu:
I'll extend this issue to include the following files:
- `accessed.js`
- `sortableTable.js`

A separate pull request for this part will be made.

HGustavs:
Old files should not be changed so that the code does not change owner, thereby conflagrating the commit-tree.

As for new files, for the sake of readability, I feel that following a single standard is better than having multiple standards, and that the difference between let and var is sligt.

b18alegu:
I do agree @HGustavs that one standard is the way to go. At the moment the code standard tells us to only use `var` and on Discord I've read that at least `const` can be used as well.

It seems like it's about time to make a decision on how to proceed from now on and update the code standard. As I see it there are two options:
-- Follow the current code standard
-- Update `code_standard.md` and use `let` and `const` instead from now on.

HGustavs:
Well from my standpoint I see a use for const (as it documents that a variable is a constant) but allowing let at the cost of having two competing nearly identical variable declaration modes in my mind seems less useful.

HGustavs:
So my suggestion is to allow const (even changing var declarations to const as we are documenting the code further) but to keep using var for all other cases, in the interest of consistency.

b18alegu:
@HGustavs, I will make an issue about updating `code_standard.md` according to your input. so `var` should be used for legacy reasons and `const` should be used on constants from now on.

I will also close the issues regarding changing `let` to `var`.

b18alegu:
This issue is no longer relevant and will now be closed.

`code_standard.md` will be updated in #9147 to clarify the future uses of variable declarations.

PR #8839, opened 2020-05-05
G2 2020 w19 8831

b18alegu:
A small maintenance fix changing all `let` declarations to `var` in the files concerned in this pull request. See issue #8831 for further details.


b18alegu:
I changed the `const` to a `var` in `dugga.js` as well now @c14jonfr. I can't see it being explicitly mentioned in the code standard but I believe its within its spirit.

The slightly messy "Files changed"-window is of course my bad... But the indentation is much better now even though it looks like a mess here on GitHub.

Issue #8830, opened 2020-05-05
Dropdowns in Access Editor opens on hover


c14jonfr:
This can only be seen/tested after https://github.com/HGustavs/LenaSYS/pull/8737 on G2-2020-W19 (group 2's groupbranch). After the W19 weekly branch this should also be seen on the master branch (unless it gets fixed before the merge).

Currently all the dropdowns except the Group dropdown will open on hover. This feels very unintuitive and it's also quite annoying when a bunch of dropdowns open when you're moving the cursor across the screen.

I suggest this be changed to open on click. Clicking a dropdown should open it. Clicking outside of the dropdown, clicking the dropdown again or selecting a value in the dropdown should close the dropdown. If the "new" click is on another dropdown while a first is open, the first should be closed and the new dropdown should be opened.

b18jonli:
Closing since this has been merged.

PR #8844, opened 2020-05-05
G2 2020 w19 #8830 Accessed dropdowns now open/close on click instead of hover

c14jonfr:
Solves #8830.

Dropdowns in Access Editor now open/close on click instead of hover according to issue suggestions. 

There are still some styling issues with the dropdowns, a separate issue should/will be opened for this. This pull request only fixes behavioural issues for the dropdowns.

### DIFF
--- a/DuggaSys/accessed.js
+++ b/DuggaSys/accessed.js
@@ -777,7 +777,9 @@ document.addEventListener('click', function(e){
 			}
 		}
 	}else{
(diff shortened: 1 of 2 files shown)


c14jonfr:
@b18jonli Yeah, I played around a bit with using the activeElement but I couldn't get it to work like I wanted to so I decided to leave it as is for now. But I agree it would be good if they could be combined. Possible future issue?

Issue #8856, opened 2020-05-05
Give user feedback when attempting to draw invalid line


o18marka:
When a incorrect line (for example UML -> ER) attempted to be drawn by the user the line won't appear, since it's not valid. It would be nice to notify the user of this on the screen. Best solution is probably to use the already existing flash function to display a flash-message.

PR #8858, opened 2020-05-05
User feedback when incorrect line is attempted

o18marka:
Solves issue #8856

Using a flash message on screen the user now gets feedback on why line could not be created.

Tester:
Try to draw all different types of incorrect lines to see if appropriate error message shows up. Also try drawing correct lines to make sure we don't show en error by mistake.
http://group4.webug.his.se:20001/G4-2020-W19-ISSUE%238856/DuggaSys/diagram.php


a18kaswi:
Works very nice, very good uxd making it friendly to new user like me!

good job.

Issue #3698, opened 2017-05-10
Move out javascript to separate file for installer.


b15linek:
Now javascript is located everywhere inside the main php-file. Would be better if it was moved to a separate js-file.

a18kaswi:
Asked around, seems possible to call javascript, from within the php-file. Unclear why this separation hasnt been done before. Will look further into it.

a18kaswi:
This issue is partly solved with #8867

There are now two corresponding .js-files. 

One (install_entry.js) that holds functions that needs to be loaded pre-php execution (functions that are called from within the install.php). 

And one (install_defer) that holds js-stuff that needs to be loaded post-php execution (I dont think php execution is the right terminology but I dont know better wording). 

In #8867 I have extracted code involving the permission-modal and the form. I will continue to work on extracting code from the installation-part the comming week.

The aim of my work has been to make the code more friendly to new developers looking at it, which has involved some new comments. And additionaly to make the code more friendly for further development.

a18kaswi:
This issue is finally done. The installer now consist of three files. The install.php, install_entry.js and install_defer.js. 

install.php holds most of the old installer but are now divided more into functions. This was necessary to enable the withdrawal of several js-functions.

install_entry.js holds javascript-functions that are needed on load, foremost functions that are needed for getting/setting permission-modals. 

install_defer.js holds most of the javascript-functions like onclicks and other stuff that manipulates the ui.

For now the new code can be viewed in https://github.com/HGustavs/LenaSYS/tree/G4-2020-W20/install and the old in https://github.com/HGustavs/LenaSYS/blob/master-copyW18/install/install.php

PR #8867, opened 2020-05-06
Partial solution for #3698

a18kaswi:
Partial solution for #3698

Moved out Javascript-logic to files install_entry.js and install_defer.js

The install_entry holds javascript-logic that needs to be loaded pre php/html execution (probably bad wording). 

The install_defer holds javascript logic that needs to be loaded post execution.

Seperated the logic regarding the permission-modal and form-filling.

Test that everything (buttons, installation in itself) works on branch G4-2020-W19-ISSUE#3698-4


a18kaswi:
Do not delete the related issue.

PR #8871, opened 2020-05-06
G4-2020-W19-ISSUE#3750

a18petha:
*For testers*

Link to test URL http://group4.webug.his.se:20001/G4-2020-W19-ISSUE%233750/DuggaSys/diagram.php

This issue relates to a new functionality within diagram, we should now be able to create elements within layers. To start writing to a layers use the drop-down menu located on the top toolbar to select a layer, only one layer should be able to be written to at a given time. to view a layer use the drop-down view layers, multiple layers should be able to view at the same time.

known bugs intended to be solved next week.

* drop-down menu do not effect canvas before there closed or cursor leaves the drop-down element
* Delete layers is not implemented yet, to delete layers clear localStorage
* if you refresh the page with cap of 10 layers an unidentified layer will appear, this is not intended

most important thing to test here is to try and get the system to write a object to none existing layer, this should never be doable.


a18petha:
Conflicts relates to this being a early build of the weekly branch, it's missing the function "canConnectLine" which seems to have been merged with the weekly branch from pull-request https://github.com/HGustavs/LenaSYS/pull/8843. Think it should be safe to ignore this conflict

PR #8918, opened 2020-05-06
G3 2020 w19#8833 

a18fayha:
Changed the the variable declarations from "let" to "var"


b18alegu:
This pull request should probably be closed since the issue is no longer relevant but I'll leave that decision to group 3s issue manager.

Issue #8789, opened 2020-05-04
Fix shadow for other icons on hover


a17towan:
There are inconsistencies still regarding icon hover effects which sometimes have a shadow and sometimes not.

On pages "Edit student results" and "Give students access to the selected version".
![image](https://user-images.githubusercontent.com/49142028/80960265-d7d27d00-8e08-11ea-8a7f-92a8b5d1c9d0.png)

These icons on "Edit student results" page:
![image](https://user-images.githubusercontent.com/49142028/80960126-8924e300-8e08-11ea-96a4-bdd31c2838a8.png)

And this icon on "Give students access to the selected version" page:
![image](https://user-images.githubusercontent.com/49142028/80960364-fe90b380-8e08-11ea-93be-e1bad623959a.png)

Give them a shadow effect on hover.

PR #8924, opened 2020-05-06
Added a shadow over the "filter" and "sort" icons on hover | Fix for issue #8789

a17towan:
Added a shadow over the "filter" and "sort" icons on hover (mouseover) for the sake of consistency because they had none while the other icons do.

The "shadow" is just a copy of the original image with an added box shadow on to it, and when you hover the mouse over it, the images switches back and forth so that it appears to have a shadow.

The icons are on the pages "resulted.php" and "accessed.php". Click on these icons **to get to the page**s:
image

Gif of the new shadow effects over the icons that had none: 
<a href="https://gyazo.com/2bdf0c251d64a9df1bfcb59d9be2d100">Image from Gyazo</a>

(P.S. To the reviewer: the shadow is kind of small so you have to really look close to see it)


a18jenro:
Good work!

PR #8925, opened 2020-05-06
solves issue#8835

b18simla:
Replaced all "let" with "var" in diagram.js, diagram_objects.js and courseed.js

### DIFF
--- a/DuggaSys/courseed.js
+++ b/DuggaSys/courseed.js
@@ -469,7 +469,7 @@ function htmlFix(text){
 }
 
 
-let activeCodes = [];
+var activeCodes = [];
 
 //Gets all active course codes in the list by the data-code attribute all items have when created based on database values.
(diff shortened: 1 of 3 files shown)


a18antsv:
I do not think this is relevant anymore, since the customer wrote that variables that clearly are constants could be changed to const, meaning they are okay with using the new ES6 let and const keywords. They are not interested in changing all var to let though since this would mean thousands of lines having to be changed and also change owner for no change of functionality.

a18kaswi:
@b18alegu check Antons comment and elaborate on the discussion. Feels weird if these kind of things are not consistent over the board.

b18alegu:
I've made a comment on https://github.com/HGustavs/LenaSYS/issues/8831 @a18kaswi.

But I do agree with @a18antsv that the code standard should be updated to using `let` and `const` from now on.

PR #8926, opened 2020-05-06
Fixed errors

b18emigu:
The errors that used to show up due to the function not being defined has been fixed.

Co-Authored-By: a18felen <a18felen@users.noreply.github.com>


a18jenro:
Good fix!

PR #8933, opened 2020-05-06
Removed buttons from uncorrectly created assignments

a18eripi:
Deadline was set to timeZero on assignments, which is improbable to work at best.

Another issue will be created prompting teachers to remove and remake the assignment, but for now it can not be graded if made in error.

long term solution would be for the creation process to prevent invalid deadlines to be set, but this is a separate issue as well.


b18jonli:
Closing this since the linked issue has been closed

PR #8934, opened 2020-05-06
Further work on error message handling to user

o18marka:
Now error messages works when connecting loose lines as well.

Also fixed so error message now shows up when trying to connect multiple lines between two ER-attributes which realized i missed before.

Tester:
http://group4.webug.his.se:20001/G4-2020-W20-ISSUE%238932/DuggaSys/diagram.php
Loose lines is easiest created by copy-pasting a ER-line, then drag endpoints to different combinations of objects to see if on-screen error handling works properly.


o18marka:
Note: this is a continuation of issue #8856 which incorporates on screen error handling for making invalid lines.

Issue #8942, opened 2020-05-07
AccessEd - Sorting by examiner is broken


b18jonli:
Trying to sort by examiner doesn't work and gives this error:
<img width="776" alt="Screenshot 2020-05-07 at 9 49 14 AM" src="https://user-images.githubusercontent.com/62876553/81268258-04c0a300-9048-11ea-9fff-8f803ae60cd3.png">

a18viksa:
Closing issue after merge to master.

Issue #8909, opened 2020-05-06
AccessEd - It is not possible to select "None" or similar in the dropdown menus


b18jonli:
This was possible previously and probably should be again. Also, the various "null"/empty/"None" values should be handled (and look) the same in the table. 

Now:
<img width="220" alt="Screenshot 2020-05-06 at 2 26 48 PM" src="https://user-images.githubusercontent.com/62876553/81176632-b1961400-8fa5-11ea-9fa3-10a724765706.png">

Previously:
<img width="171" alt="Screenshot 2020-05-06 at 2 27 04 PM" src="https://user-images.githubusercontent.com/62876553/81176643-b8248b80-8fa5-11ea-86d2-1cea5c252d2c.png">

a18viksa:
Closing issue after merge to master.

PR #8949, opened 2020-05-07
Fixed examiner-sorting to be alphabetical

b18jonli:
The current table in accessed is a bit buggy, so while testing one option is to rebase this branch onto `G2-2020-W20-#8909` which fixes some of it **but be very careful not to push the rebase to the branch**.

Another option is to look at the difference between this branch and `G2-2020-W20`, to see i there are any additional bugs present. 

In any case the error shown in #8942 should be gone, and the table should be able to sort by examiner.

### DIFF
--- a/DuggaSys/accessed.js
+++ b/DuggaSys/accessed.js
@@ -915,10 +915,10 @@ function compare(a, b) {
 					if (typeof tempB === "number") {
 						tempB = "";
 					}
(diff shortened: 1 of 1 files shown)


b18jonli:
With invalid data in the database, some values could be null that wasn't handled in the code, so I added some further not-null checks.

PR #8951, opened 2020-05-07
G3 2020 w19#8702

a18fayha:
The search icon is now bigger in all the places except "Show tests". For some reason the changes didn't apply there and i will make another issue about it later.


a18jenro:
@a18fayha and @b18emigu reagrding the positioning, I agree with you Emil, it should be in line with the other icons of course. As for the size, I guess the larger icon is nice, but it should be more or less exactly the same size as the other icons.

a18jenro:
@a18fayha You have made a pull request to the master branch. This is not the first time, please make sure to make PRs only to out weekly branch.

b18emigu:
Also try to have a look at the merge conflicts as well.

PR #8955, opened 2020-05-07
G3 2020#8859 Create an 'new announcement' button is only accessed by the admin

a18mahkh:
- login as admin 'brom'  the button is displayed
- login as student 'teacher2'  the button is hidden


a18jenro:
Seems to work

Issue #8103, opened 2020-04-15
Icon for "Add Dummy Empty file" is missing and upload icons is not functioning


a16silph:
When looking at **fileed, duggaed** and **accessed**, you should see a plus icon in the lower right side of the screen. When hovering over the plus icon, there will be different icons showing the user different ways to upload different types of files. When using _"Add Dummy Empty File"_  button in all php files, nothing seems to happen. It seems that all of the other upload icons don't seem to work either.

![image](https://user-images.githubusercontent.com/49142660/79235637-2923c800-7e6c-11ea-80bc-c77ec0f9b303.png)
**Fig. 1** - Icon for _"Add Dummy Empty file"_ **is** missing and all the upload icon is not functioning.

b18jonli:
I think "add dummy file" should only exist in fileed, not duggaed and accessed.

a18jenro:
As of right now, the button exists only in filed, as it should. The icon is also there. But the functionality is questionable.

a18jenro:
@b18emigu or @a18felen Is it perhaps that the button does not work due to access rights? What would one have to do to make then dummy button work?

a18felen:
@a18jenro I just think that the code for creating a dummy file is not there. not sure what code exactly that is needed but i think its filerecieve, fileed.php, and fileed.js that needs to have code added and fileedservice for the edditing of buttons.

b18emigu:
All of the icon related bits have been dealt with in other issues however creating dummy files simply has no code for it, like @a18felen said. We will implement the code for uploading the files but i do believe we will need a definition of a **dummy file** in the database as well.

b18emigu:
We have now finished implementing frontend validation as well as some server-side validation. We are now leaving the rest of the issue (creating the dummy files) to @a18jimol and @a18jonsj as they were working on related issues and had already started implementing the creation of files #7856 and #7919

Issue #7919, opened 2020-04-07
Calling of file-ed


a18mahkh:
Make sure that we are calling the service for creating files in fileed, so that we are not writing new code for the same task.
This issue is related to:  https://github.com/HGustavs/LenaSYS/issues/7766 and https://github.com/HGustavs/LenaSYS/issues/7917 and https://github.com/HGustavs/LenaSYS/issues/7918

HGustavs:
Either by calling service directly or by a conveniently placed iframe to fileed.

a18jimol:
Now we have the button and dialog to write the name of the file from #7917 , #7918 . Do you want us to use the service of fileed.js and filereceive.php as we are in fileed.php when writing the dummy-file so don't see the reason to use an iframe as we already are in fileed.php.

When tryint to call the service filerecive.php it seems the code don't accept the string of text of the name of the file, we get no errors but it just sends us back to fileed.php without a new file saved. Should we make new functions that sends the data in another way than uploadfile() from fileed.js or can the string be converted to a "file". We also have a problem finding where filereceive.php actuelly get the filename from.

a18mahkh:
@HGustavs @mbrohede

a18jonsj:
We are not sure about this issue, we have gotten to the point where we can convert a text-string in filerecieve.php, but we havent figured out a way that will use $_FILES array for taking the file uploaded by a user, and using the values from the array during the actuall upload.
What we can do is to use all the code that checks if the file is valid, the code that puts the file in a folder and uploads it to a server, and put it all in place made for the dummyfiles only, because there seems to be no way of creating a file in fileed and sending that file with "input type=file".

Do you think that putting all of the code for uploading into two places is a bad idea, or is it an acceptable soloution? Right now this is the only way we are able to make it work, the problem being that the same code exists in several places.
@HGustavs @mbrohede

a18jimol:
We can create a file and save it in the right directory and upload it to the database(on localhost) but we don't know if this really is agood solution. Because we can't use $_FILES and the varibles that you use when creating and uploading files we have a seperate if-statement that handles "dummy-files". This means that we have two places in filerecieve.php that have the same functions and that might not be that good but we don't see another solution.
On top of that we could not use "move_uploaded_file" as that one only handles files that get uploaded by the user and the only check to see the file is allowed is to see if the extension is allowed.
@HGustavs @a97marbr

a97marbr:
Not sure I follow.

[...]

PR #8960, opened 2020-05-07
G3 2020 issue#7919 creating dummy-files

a18jimol:
Now in fileed.php you can create dummy-files by writing the name of the file, ex hej.txt and choosing an type of global, course local and version local.
This pull-request is a combination of issue #7919 and #8103


a18jenro:
Works really well!

PR #8961, opened 2020-05-07
Scrollbars now hide and show when needed for all templates

a17chrka:
<a href="https://gyazo.com/9bd2b4dd22aff8a4140ab110ac5c750f">Image from Gyazo</a>


a17chrka:
Last request should be fixed now

a18filkr:
We will revert this to try if this solves the issues with Max/Min/Reset buttons that occured.

PR #8962, opened 2020-05-07
Solved the bug where the last exported date would say "undefined".

b18fella:
The last exported date does now work as it's supposed.

By solving this issue, at least one more issue surfaced. Those issues will be made into separate issues because they were too big to handle in this issue alone.

### DIFF
--- a/DuggaSys/resulted.js
+++ b/DuggaSys/resulted.js
@@ -686,12 +686,15 @@ function returnedResults(data) {
 // Success return function for LadExport lastGraded
 //----------------------------------------
 function returnedExportedGrades(gradeData){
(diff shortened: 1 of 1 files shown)


b18jonli:
Good! I'm not sure I completely understand _why_ the code works, but it does solve the issue. 

However, I also get the error(row 696) printed in the console when I click the "LadExport"-button, is this intended? If so, should it always be printed or be togglable so it is only printed when debugging?

b18fella:
@b18jonli That is intended at the moment. It has to do with a whole other problem for which an issue will be created shortly. I thought that it was too big to fix in this small issue.

PR #8970, opened 2020-05-07
G3 2020 issue#8874

a18jonsj:
You no longer the error: 
image
after creating a new course-version.


a18jimol:
> * You can successfully create a new course version and you don't get any console error
> * However the page is not redirected to that new course version

Only works if you choose Copy content from None. Will make new issue about that error

PR #8974, opened 2020-05-07
Button to visit analytics - issue#8877

a18eliek:
Added button to visit analytics, and a back button if the user does not like the home button.
This button is only shown in courseed.

image


a18eliek:
This should be tested with:

- Superusers (brom, toddler)
- Teachers (teacher1, teacher2)
- Students (a13andka, a13sveth)

or other accounts as well

PR #8975, opened 2020-05-07
G3 2020#8902

a18felen:
Fixed a php error where cookies were set after the header had been loaded then it turned out there was a second error with loading header. So we moved the including of navheader after code had ran.

This can be tested by going into analytics as superuser and as guest. only superuser should have access to the page.


a18jenro:
@a18felen  and @b18emigu could you resolve the conflicts?

a18felen:
@a18jenro done

a18jenro:
butifull

PR #8986, opened 2020-05-08
Fixed buttons in analytics - G1 2020 w20 issue#8915

a18eliek:
Fixed button width.
Before:
image

After:
image


a18eliek:
This PR also removes all inline css from analytics, and includes a new css file specifically for analytics that can be used for future additions to analytics.

PR #8989, opened 2020-05-08
Examiner options are now updated when users gets/loses teacher access

### LINKED ISSUE #8936: AccessEd - When setting user to "teacher" a reload is required to choose that user as "examiner"
The examiner dropdown menu options should update without requiring a reload.

b18jonli:
Fixes #8936 
The accessEd table is still buggy, but this pull request together with #8949 and #8941 should fix most of them.

### DIFF
--- a/DuggaSys/accessed.js
+++ b/DuggaSys/accessed.js
@@ -270,11 +270,15 @@ function changeOptDivStudent(e,value){
 	keyvalue = e.target.getAttribute('data-value');
 	
 	obj = {
+		uid: paramlist[1],
 		[key]: keyvalue
 	}
 	updateDropdownInTable(e.target.parentElement.parentElement.firstChild, obj);
 	changeProperty(paramlist[1], paramlist[0], value);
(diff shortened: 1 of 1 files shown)


b18jonli:
The reason for this is that the re-rendering of the table happens when the data returns from `accessedservice.php`, it might be possible to hide but I guess the question is whether it is worth it or not. Depending on the latency to the server, the delay will vary in length, but I don’t know if this problem would be common enough to warrant a fix.

a18eripi:
Same findings as @a18viksa when testing, good job.

Agree with @b18jonli that troubleshooting latency can become a rabbit hole not worth the time.

b18jonli:
@a18eripi @a18viksa I thought about this a bit more and is it a second of delay? because that is odd considering the server runs locally, the delay probably shouldn't be noticable. I haven't been able to recreate this either.

b18jonli:
Re-requesting reviews since resolving the merge conflict

PR #9004, opened 2020-05-08
#8683 Fixed

### LINKED ISSUE #8683: Diagram import file - long file name overflow
When importing a file and the name is longer than the very small element it should fit into, the name does not show fully and appears in white behind the button on several rows. In my opinion the popup should be way wider to allow for a longer name in the element. The user cannot do anything else on the site while the popup is there anyway, so there is no point having it this small. If the file name would somehow be so long it would overflow the longer element as well, ellipsis (...) should be used and the rest of the string should not be visible. Never overflow the element.

The issue can be seen in this image.
image

a18frior:
#8683 Styling for Import diagram is now updated. The container and button is now bigger to make sure the text fits.
http://group4.webug.his.se:20001/G4-2020-W20-ISSUE%238683-2/DuggaSys/diagram.php

### DIFF
--- a/DuggaSys/diagram.php
+++ b/DuggaSys/diagram.php
@@ -699,8 +699,8 @@
             <div class='table-wrap'>
                 <div class="importWrap">
                     <div style="width:100%;">
-                        <input type="file" class="import-file-button" id="importFile" accept=".txt, text/plain" style="width:100%" />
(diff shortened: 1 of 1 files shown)


a18kaswi:
Very good change, very stylish code.

APPROVED

PR #9026, opened 2020-05-08
G3 2020#9023 now it shouldn't show the warning alert 

a18mahkh:


a18jenro:
@a18lukto when you go into fileed, dont you see the error message: Unknown error type: [8] Undefined index: errortype ?

If so, that is a separate issue that has been solved in: https://github.com/HGustavs/LenaSYS/pull/9038

a18lukto:
@a18jenro Yes i do get that error message, thought this was suppose to fix it. If that is not the case, this issue is resolved

a18jenro:
@a18lukto It is supposed to fix the error "Undefined index: announcementid". "Undefined index: errortype" is a separate issue.

Issue #8968, opened 2020-05-07
Changing language in some places from swe to en in resultEd


a18ismca:
In some parts of resultEd, the language is on swedish instead of english. It would be better if resultEd was on 100% english, thus reducing inconsistencies in the same page.

With @a17pioja

a18viksa:
Closing issue after merge to master.

PR #9029, opened 2020-05-08
Issue #8968: Changed language from swedish to english

a18ismca:
The language has been changed from swedish to english in these places shown below.

image


a18ismca:
Yes @b18veran, also wanted LadScale to be named "Grading scale" but the dropdown icon below adjusts by itself. Also, writing the word that you suggested will be written on two rows (not on one and the same row as others "options")... that's why it is named LadScale... but thanks for commenting on that one :)

b18veran:
Okey @a18ismca if thats the problem I understand. It can be solved through the css if you change #ladexportContainer in style.css in Shared from width: min-content; to width: max-content; everything gets its place it needs.

a18ismca:
@b18veran now this has been changed successfully.

PR #9039, opened 2020-05-08
G3 2020 issue#9022 Now submit-buttons are in the right position

a18jimol:
now the buttons are in the right place and not moved a bit to the right


a18jenro:
Looks better.

PR #9042, opened 2020-05-08
G4 2020 w20 issue#8475

b18mikfr:
Done, while solving this issue i stumbled upon another issue where the hover would also break when the origo and origin of the map weren't aligned so i got that fixed. There should have been a separate issue for the "origo alignment" but since there was none and it would be a quick solve, i simply solved it within this one too.


o18marka:
Note: Found that it is not possible to hover on recursive UML-line, but that is not due to this issue. Therefore i made a new issue for this: #9048

PR #9101, opened 2020-05-11
G3 2020 issue#9021 validation for quotation marks.

a18jonsj:
The user can no longer type quotation marks in the section-name while editing a dugga. Our solotuion validates both name and date correctly when updating an item in sectioned. The user gets an errormessage and an alert if something is wrong.


a18jenro:
Seems to work really well!

PR #9128, opened 2020-05-11
Fixes #9100 - Scans example diagram folder and shows menu-item for each

### LINKED ISSUE #9100: Example diagrams: add diagrams to a folder, scan the folder and show a side dropdown with all files
It would be good if a folder containing all example diagram files exists. This folder could then be scanned into a side dropdown for the menu-item "Generate example diagrams". It would be a good way of making it possible to easily add new example diagrams into this folder and for them to show up automatically on the page.

a18antsv:
Check Help -> Generate example diagrams. A side drop down should appear with all files scanned in a particular folder. Right now, 4 files should show. Test to generate all of them.

This also generalizes side dropdowns and hot key positions making it possible to use the same class for everything.

Link: http://group4.webug.his.se:20001/G4-2020-W20-ISSUE%239100/DuggaSys/diagram.php


a18antsv:
@a18kaswi That is something that should be fixed for the layer system that @a18petha is working. Points that are referenced by objects in a hidden layer should also not be drawn.

Issue #9113, opened 2020-05-11
When editing a dugga in Dugga Editor containing characters åäö in name, these are not handled properly


c14jonfr:
See screenshot. The name of the dugga being edited is "ööö".

![image](https://user-images.githubusercontent.com/49141758/81558789-9acd3400-938e-11ea-8e70-780b0f2d1124.png)

If this dugga is saved (without changing the name), the name will be saved as `"&ouml;&ouml;&ouml;"`.

![image](https://user-images.githubusercontent.com/49141758/81558919-cf40f000-938e-11ea-8935-082f941c7ed7.png)

Additionally, if `"&ouml;&ouml;&ouml;"` is edited again, name in the textbox will be changed to `"&amp;ouml;&amp;ouml;&amp;ouml;"` (this goes on and on if you keep editing and saving the dugga).

One solution to this is replacing the `"&ouml;`" with "ö" using the replace() function (the same goes for Ö, å, Å etc) but maybe there is a different, better solution.

c14jonfr:
After looking into this a bit it seems that decoding the qname-value using html_entity_decode() before the array is returned from duggaedservice.php is the most clean way to solve this.

a18viksa:
Closing issue after merge to master.

PR #9134, opened 2020-05-12
åäö characters in name are now displayed correctly in Edit dugga window

c14jonfr:
Solves #9113.

Added decoding in duggaedservice.php for qname.

Tested with å/ä/ö/Å/Ä/Ö characters. Not sure how it handles other characters like ß for example. I felt that it was not needed since I doubt such characters will ever be used, unlike åäö. But feel free to look into it if you'd like.

Screenshot shows the dugganame "ööö" after the fix. Before the decoding this would be written out as `"&ouml;&ouml;&ouml;"`.

image

### DIFF
--- a/DuggaSys/duggaedservice.php
+++ b/DuggaSys/duggaedservice.php
@@ -206,7 +206,7 @@ if(checklogin() && (hasAccess($userid, $cid, 'w') || isSuperUser($userid))){
 		$entry = array(
 			'variants' => $mass,
 			'did' => $row['id'],
-			'qname' => $row['qname'],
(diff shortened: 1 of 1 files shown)


c14jonfr:
@b18fella It sounds like you have maybe misunderstood the issue. The issue is not inputting åäö for name, but rather after a dugga has been made containing åäö in name and then editing that dugga. Before the åäö-characters would not show up properly (which they should do now).

b18fella:
I misunderstood the meaning of this issue. But I have tested it again with the correct meaning in mind and it works exactly as it should.

Issue #7577, opened 2019-05-27
Sectioned. Tabs cannot be changed on tests


ghost:
![image](https://user-images.githubusercontent.com/49199825/58418254-93caea00-8087-11e9-806f-bebd5bd396da.png)

![image](https://user-images.githubusercontent.com/49199825/58418318-c96fd300-8087-11e9-8612-1509d8e8bb06.png)

Trying to change the Tabs for a test. But the test never change

a18mahkh:
If you create a new 'Moment' then you don't need tabs on 'Test' for indentation. I think this moment option is there to separate tests??
![Anteckning 2020-04-16 112901](https://user-images.githubusercontent.com/49142349/79439746-7d968700-7fd5-11ea-98be-ef5ca2dae378.png)

@HGustavs @mbrohede

a17towan:
It is my of my understanding that "tabs" is supposed to indentate the circles (the types) further to the right depending on how high of a number you pick. If this is the case then it would seem that "test", or "duggor" already have indentation so they naturally align themselves slightly to the right when you create them; regardless if they are below a "moment" or not. 

It looks fine to me, so personally I'm not sure if the "tabs" option is needed. But I could be wrong? @HGustavs @a97marbr

a97marbr:
A moment is the equivalent of a "delkurs". For example, in the project course we have `Projekt  10,0 hp`, ` Seminarier och individuell rapport  2,0 hp` and `Hemtentamen 3,0 hp`

The discussion if having a dugga that is not part of a moment is a valid placement has not been decided. I know that @HGustavs and myself has had this discussion some years ago.

However,
What if we want to split an assignment into several steps with individual deadlines? Let's say for example, that a course has two assignments in one moment and each assignment has several tasks with individual deadlines?

* Moment
  * Assignment 1
    * Assignment 1 - Subtask 1
    * Assignment 1 - Subtask 2
    * Assignment 1 - Subtask 3
  * Assignmane 2
    * Assignment 2 - Subtask 1
    * Assignment 2 - Subtask 2
    * Assignment 2 - Subtask 3

This we cannot visible show at the moment, right?

a17towan:
Alright yes, then I understand why tabs is important for tests! Good explanation.

a17towan:
@a97marbr 

Curious, what are "1 tab + end" and "2 tabs + end" supposed to do? What position should the dugga be moved to? 

[...]

PR #9155, opened 2020-05-12
New functional feature: Tabs | Fix for issue #7577

a17towan:
When you edit a "dugga" (the "Edit item" window) you can choose from the "tabs" selection box and appoint the current dugga one of the options for "tabs" indentation. This feature helps in creating a clear workflow for tests (duggas) in case there is a need for "sub-tests" which need to be accomplished before the "parent-test" can be passed.

_(You press the cogwheel to the right of the tests in sectioned.php to open this window)_
image

**The indentation is not automatic, which means that the user themselves have to manually order the different tabs options to make it look as intended.**

image

The options with the word "end" in them are the ones that have a curved horizontal line. They are meant to be the "end" of the workflow. The options without the word "end" in them will only have a vertical line.

I added a column for "tabs" in the "listentries" database table for this to work. 

When you choose a "tab" option and exit the "Edit item" window, when you re-open it again, your selected option should be displayed as default in the select-box. This did not happen before.


a18felen:
@a17towan "Error reading entriesUnknown column 'tabs' in 'field list' and "You either have no access or there isn't anything under this course" Seems like i cant access any course as brom who is superuser.

a18jenro:
it works!

Issue #8874, opened 2020-05-06
Error when redirecting to newly created course version


a18jenro:
When successfully creating a new course version (some bugs make it so that creating a new course version fails), then the page should be redirected to that new course version. Instead, the following console error is given:

![untitled](https://user-images.githubusercontent.com/62876507/81150718-91079300-8f80-11ea-8942-eaf65b80c6e7.PNG)

PR #9161, opened 2020-05-12
Changed to changeCourseVersURL from changeURL | Issue #8874

a18jimol:
When creating a new course-version you will get sent to the new version. This will only work when choosing Copy content from None


a18jenro:
Beautiful!

PR #9164, opened 2020-05-12
Added checkbox for demo-course FIXES #8311

### LINKED ISSUE #8311: Implement option to select test-data during installation.
As stated in #2974 the system needs more test data, ie: an empty course, more duggas, more students assigned to courses etc. In issue #6612 it were put forth that it would be nice if the diagram were available after an install of the system. The customer stated in the issue that the request also were viable for several other views. The customer did not state which one thou, but in #6849 it were stated that also a Contribution-Dugga would be nice. These two could form an foundation for how this new improved installer could be. Then once all has been done, we could discuss further improvements (more available views) with customer.
 
In regards to #6612 an idea have been put forth in #6790 how an implementation of an diagram for educational purpose could be implemented. Finally once again in #7260 the request for better test-data were mentioned. 

The title of #6849 formulates an good idea. That is to make it possible during the installation of the system to be able to select which test-data one would want. 

My estimation is that this would take maybe two days for two developers to investigate, formulate a solution for and implement the collective request of these issues.

#2974 
#7260 
#6612 
#6790 
#6849

a18kaswi:
For tester: Go to http://group4.webug.his.se:20001/G4-2020-W20-ISSUE%238311/install/ and run the installer, On the test-data page there should be an option to select "Include demo-course". Tick the box and run the installer. Verify that the democourse is there when you open the branch course-ed.

### DIFF
--- a/install/CSS/install_style.css
+++ b/install/CSS/install_style.css
@@ -97,22 +97,6 @@ input[type="text"], input[type="password"] {
     min-width: 550px;
 }
 
-/* Whole area between header and footer */
-#contentWrapper {
-    height:400px; /* Must hold same height between input page shifting */
-}
-/* Wrapper will get taller on smaller screens to show all text. */
-@media screen and (max-width: 1384px) {
-    #contentWrapper {
-        height: 450px;
-    }
-}
-@media screen and (max-width: 800px) {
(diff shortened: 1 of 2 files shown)


a18petha:
Looks good

PR #9176, opened 2020-05-12
Fixed the issue where coursename is gone when returning from coideviewer

a18felen:
The issue was that coursename never existed when entering codeviewer so when returning it did not know what the coursename is

Now when returning from a dugga or codeviewer with the button in navheader the cid, coursename and coursevers are all there. The issue was that coursename was not there when returning from a codeviewer.

Co-Authored-By: Emil <b18emigu@users.noreply.github.com>


a18jenro:
Nice fix!

PR #9178, opened 2020-05-12
G4 2020 w20 issue#7360

a17robhe:
Removed the redundant code. To test download sqlite and open the database after attempting to reset your password, event 12 is the event created by the request barrier(if the reset request failed because of a invalid username etc) and 13 is a log entry for security question entires after a valid username was entered. I didn't manage to find a valid username in my install after several attempts so I'm not sure how the solution works in there but the code looks the same in both spots.


a18kaswi:
@b18mikfr 

a99linja is a username you can use for testing the resetpw

Issue #9135, opened 2020-05-12
Suggestion: Add visual indicator for selected value in Access Editor dropdowns


c14jonfr:
Currently there is no indicator for which value is/was selected when a dropdown is opened. See screenshot. It would be nice to have some sort of indicator in my opinion. Maybe just a different background color or some small icon to the right?

Note that https://github.com/HGustavs/LenaSYS/pull/9076 changes/fixes the positioning of the dropdowns so that the selected value is visible since the dropdown no longer covers the table cell (this has not been merged yet so the screenshot below is from the "old"/current version).

Since the selected value will be more visible after mentioned pull request has been merged I don't consider this high priority at all. But it would be a nice QOL change.

![image](https://user-images.githubusercontent.com/49141758/81649910-58106800-9431-11ea-8d45-0539d31a6ad1.png)

HGustavs:
Good idea.

Subtle though. How about using the contrast color?

a18viksa:
Closing issue after merge to master.

PR #9190, opened 2020-05-13
Clarified which options are selected in Accessed. #9135

b18veran:
Added iterations to check if an examiner, class, version and access is set or not. "If" sets an id so it could be styled and print out all options. "else" prints out all options. Also an iteration for when an option is empty it will be styled. _See img 1._
Screenshot 2020-05-13 at 10 10 03
_img 1 - Examiner_

Co-Authored-By: b18jonli b18jonli@users.noreply.github.com

### DIFF
--- a/DuggaSys/accessed.js
+++ b/DuggaSys/accessed.js
@@ -347,7 +347,7 @@ function renderCell(col, celldata, cellid) {
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


a18oscte:
The fix looks very nice but it also apply the same id to multiple elements. A simple way to fix this is to change the id "access-dropdown-selected" to be a class instead.

Issue #9197, opened 2020-05-13
The "back" icons have no shadow


a17towan:
The "go back" icon only has a shadow hover effect on the sectioned.php page, and nowhere else. 

![image](https://user-images.githubusercontent.com/49142028/81790039-1a7d0f00-9505-11ea-9e53-90a069dcb482.png)

The "go back" button exists in fileed.php, resulted.php, duggaed.php, and accessed.php inside the navheader. 

Fix it so that it has the same kind of shadow effect on hover (onmouseover) as the one in sectioned.php on all the other pages as well.

PR #9203, opened 2020-05-13
Put shadow on "back" icon | Fix for issue #9197

a17towan:
Before this fix, the "back" icon in the navheader only had a shadow-hover effect on the **sectioned.php** page and not on all the other pages that the icon existed on.

This fix makes it so that the "back" icon has a shadow-hover effect on all the other pages it exists on as well. These pages include:

- resulted.php
- duggaed.php
- fileed.php
- accessed.php

Image for clarification:
image

The applied hover effect on the "back" icon:
<a href="https://gyazo.com/0f2e8c0f911a0a022e1888e313ded251">Image from Gyazo</a>


a18jenro:
So shadowy, can't see any brightness.

Issue #8851, opened 2020-05-05
ResultEd: Table rework


a18viksa:
This issue is a sub-issue to #8535. 

This is everything that needs to be fixed: 

- "only pending" and "late submissions" do not combine, but instead the table prioritises "only pending" (Img 1).

- "only pending" do not show all pending submissions (might be an issue of having removed the grade, should these be visible?) (Img 2, 3).

- in "mini mode", the filters "only pending" and "late submissions" filters rows but not cells (e.g. there are still green cells for "G")(img 4).

1.
![image](https://user-images.githubusercontent.com/49142704/81068436-a75e1200-8ee0-11ea-9215-f50f20bc5853.png)

2.
![image](https://user-images.githubusercontent.com/49142704/81068451-af1db680-8ee0-11ea-8bfe-2c4d09791e2c.png)

3.
![image](https://user-images.githubusercontent.com/49142704/81068465-b5139780-8ee0-11ea-8a01-3de964b091fc.png)

4.
![image](https://user-images.githubusercontent.com/49142704/81068505-c361b380-8ee0-11ea-912b-f09f2109ca4e.png)

a18viksa:
This issue will also include some styling fixes. Will be documented in the pull request.

a18viksa:
Closing issue after merge to master.

PR #9229, opened 2020-05-13
Resulted:G2 2020 w20 #8851

a18matna:
Solved the specified issues in issue #8851, it should now be possible to filter by only pending and late submission simultaneously, in normal mode and mini mode.

Also made some styling changes to the mini mode as well as the legend list.

(The legend list was not supposed to be adjusted here. Can be removed if it creates any conflicts). 

Co-Authored-By: a18viksa <a18viksa@users.noreply.github.com>


a18viksa:
**NOTE:** This is not completely finished. We will continue working on this in a new issue.

Issue #9249, opened 2020-05-14
Validation in Duggaed complain from the beginning. 


b18veran:
Validation in Duggaed complains from the beginning on "Name" because it is empty.

<img width="475" alt="Screenshot 2020-05-14 at 10 49 11" src="https://user-images.githubusercontent.com/49142649/81914228-52518880-95d1-11ea-9e02-830dc385651b.png">

c14jonfr:
A quite simple solution is to check whether the dugga has a qid (if it does it's an existing dugga, if not it's a new dugga) and only run the validateDuggaName() for the already existing ones.

c14jonfr:
Correction/Clarification: new duggas do have a qid, but it is "UNK" (so the check has to be if the qid is "UNK" or not).

a18viksa:
Closing issue after merge to master.

PR #9257, opened 2020-05-14
#9249 now only runs validateDuggaName on selection for existing duggas

c14jonfr:
Solves #9249. 

Added a check for qid. If qid != "UNK", validateDuggaName is run when a dugga is selected in the Dugga Editor which means new duggas (where qid === "UNK") will not have their name validated right away (only onChange).


c14jonfr:
@a18ismca I consider that expected behaviour, since that means the user specifically put an empty string for the name field. The reason for this issue is that it would complain about the name field being empty as soon as you opened the window to create a new dugga, without the user even having a chance to put a name. I don't think that's fair. But if you put an empty name yourself, I would expect it to say that the name is invalid.

Issue #9194, opened 2020-05-13
Shadow over the MOTD


a17towan:
It would be nice to create a little shadow under the MOTD banners in courseed.php and sectioned.php so that they don't look so one dimensional (add a **tasteful** box shadow).

![image](https://user-images.githubusercontent.com/49142028/81789116-cf163100-9503-11ea-94ed-bbab6bd18275.png)

![image](https://user-images.githubusercontent.com/49142028/81789209-eead5980-9503-11ea-8dfa-1bf449b388e7.png)

PR #9268, opened 2020-05-14
Putting box shadow on MOTD banners | Fix for issue #9194

a17towan:
Putting a tasteful box shadow on MOTD banners to make them look less one-dimensional.

image
image


a18jenro:
Cool!

PR #9278, opened 2020-05-14
Resolved data not showing up on page information - G1-2020-W21-ISSUE#9266

b17carbr:
Now logged data is presented again in page information:

Screenshot 2020-05-14 at 12 20 24


b17carbr:
> Everything seems to work as expected.
> One minor fix to move 'else' up at the same line as the end bracket of the 'if', so it follows code standards.
> bild

@a17chrka now moved the 'else' in the latest commit.

PR #9283, opened 2020-05-14
G3 2020#9252

a18jonsj:
Now youre able to create a new course verision without the validation immediately giving an error for the version id.


a18jenro:
Seems to work well!

Issue #9277, opened 2020-05-14
Edit course version does nothing


a18jenro:
When editing a course version, the changes simply do not happen.

PR #9286, opened 2020-05-14
Now edit-course-version work | issue #9277

a18jimol:
When editing an course-version it will now work and the information will be updated.


a18jenro:
Good job bois!

PR #9290, opened 2020-05-14
G3 2020#9053 is solved

a18mahkh:
- if you edit an announcement then the announcement is marked as 'edited' at the bottom of the announcement card

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -1692,7 +1692,15 @@ function retrieveAnnouncementsCards(){
   var xmlhttp = new XMLHttpRequest();
   xmlhttp.onreadystatechange = function() {
     if (this.readyState == 4 && this.status == 200) {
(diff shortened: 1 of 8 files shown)


a18mahkh:
The issue was to mark the announcement if it's edited which is solved. Look at figure below:
Anteckning 2020-05-14 195840


And if an announcement is edited means it's renewed and that's why the notification counter will be visible.

@a18felen

a18felen:
In this branch you have made a commit saying "notification icon added" and "if an announcement is created or updated you get an notification" so with that you could at least implement a working notification instead of creating something that will need an issue later.

Basically, if the announcement has been viewed remove the notification.

a18mahkh:
Those commits are from other branches that this was built on

a18mahkh:
Actually you have to click on the announcement card for notification counter to be gone

a18jenro:
"Those commits are from other branches that this was built on".
@a18mahkh Try to avoid building branches on top of other ones.

PR #9292, opened 2020-05-14
G3 2020#9246 is solved

a18mahkh:
- Removed default update of current time


a18jenro:
@a18mahkh Like @a17towan says, you can't keep merging in much more code than the original issue. It becomes impossible to review.

a18jenro:
But this seems to be working.

Issue #9264, opened 2020-05-14
The borders for end-tabs are off 


a17towan:
The different end-borders for the tabs when you edit a dugga are messed up again.

![image](https://user-images.githubusercontent.com/49142028/81916946-d5281280-95d4-11ea-982f-d26114e30346.png)

Fix it so it aligns and looks good.

PR #9301, opened 2020-05-14
Updating the styling for borders | Fix for issue #9264

a17towan:
The spacerRight border stylings for 'end tabs' were off again. Hopefully they stay the way they're supposed to this time. 

This is what it looked like before the fix:
image

This is what it looks like after the fix:
image


a18jenro:
Better!

PR #9322, opened 2020-05-15
Defined global error where needed FIXES issue#9247

a18kaswi:
Problem:
Before

Because root doesn't have access it cant create stuff. And error was undefinied.

Solution:
After

Now $errors is definied properly.

**For tester:** 
Go to http://group4.webug.his.se:20001/G4-2020-W21-ISSUE%239247/install/ and run the installer. When prompted for root-credentials use:

username: uselessroot
password: galvaniseradapa

Verify that no error-message for undefined $errors are printed.


a18kaswi:
@b18phibe the completeQuery got fixed in pr #9348 have a nice weekend m8!

Issue #9280, opened 2020-05-14
Date inputs in Dugga Editor not styled consistently with other inputs


c14jonfr:
The date inputs look a bit weird compared to the other inputs since they are smaller (in terms of height) and have rounded corners. See screenshot 1.

I suggest adding the class "textinput". Similar inputs in for example Section Editor has this class. An example of how this would look can be seen in screenshot 2. I would also increase the width of the input a little bit so the date is not cut off. An example of how this could look is the "Result release"-input in screenshot 2.

![image](https://user-images.githubusercontent.com/49141758/81928218-3526b500-95e5-11ea-9f07-f4b3cc801eee.png)

![image](https://user-images.githubusercontent.com/49141758/81928755-1674ee00-95e6-11ea-9a5e-a4bce73168a6.png)

a18viksa:
Closing issue after merge to master.

PR #9330, opened 2020-05-15
Added a class and styled dates i duggaed. #9280

b18veran:
Now does dates look much more like the rest of the content in duggaed. 
Screenshot 2020-05-15 at 10 55 53
_Firefox on mac_ 

Screenshot 2020-05-15 at 10 56 05
_Chrome on mac_

When this is reviewed check that there are no major differences on different operating systems. There will be differences but make sure they are not annoying.


b18veran:
I´m not having that much of the cutoff but I think it is better now. @c14jonfr.

c14jonfr:
Looks good to me!

image

Issue #9216, opened 2020-05-13
Icons in fabbutton-buttons are displaced


a18jenro:
Tragic:
![untitled](https://user-images.githubusercontent.com/62876507/81807671-cb43d800-951e-11ea-842b-f0ed14e4c62a.PNG)

a17towan:
The icons seem to be more centered after the weekly merge. The lower fab-buttons (in the lower right) look okay but the ones at the top still look slightly off centered. 

![image](https://user-images.githubusercontent.com/49142028/81919150-becf8600-95d7-11ea-9899-00b7f97c2537.png)

a18jenro:
@a17towan the top icons are still a bit off I guess. So it should be fixed.

PR #9332, opened 2020-05-15
Fixing the fab button icons placements | Fix for issue #9216

a17towan:
The only solution I found to center the icons in the fab buttons was to make the title text slightly smaller. This centers the fab buttons. It now looks like this:

image

Before it looked like this:
image


a18jenro:
Sugio!

Issue #8731, opened 2020-04-30
In 'edit item' window you can type in special characters but not "+"


a17towan:
As of right now you are allowed to type in special characters when naming your item, all except the + sign, which doesn't show up. 

![image](https://user-images.githubusercontent.com/49142028/80712068-aa788d00-8af1-11ea-8424-ae2b53089190.png)

![image](https://user-images.githubusercontent.com/49142028/80712113-c11ee400-8af1-11ea-9480-a13ebf6cfb7c.png)

The + sign should probably be allowed to show up.

a18jimol:
The reason why the "+" gets removed is becauce when it gets send to sectionedservice.php it goes through the function getOP in basic.php and in getOP uldecode is used on the name and removes the "+". Do you want us to fix this so you can use "+" in section-names or is this inteaded?

a17towan:
Hm, well yes, I would like it so that the "+" is visible in the dugga name.

a20chrjo:
Approved and merged by @a20chrjo

PR #9334, opened 2020-05-15
G3 2020#8979 | MOTD allows åäö and special characters +§&%#!,. 

a18jimol:
MOTD in courseed and sectioned course-versions will now allow åäö and special characters +§&%#!,. . MOTD in courseed will allow 100 characters.

The "+" is still not saved as it is removed when the text is encoded/decoded before saved in the database (can see issue #8731).


a18lukto:
@a18jimol @a18jonsj Shouldn't "?" be included or do you think it´s not necessary?

Issue #9335, opened 2020-05-15
Resulted: Make it possible to close the legend list in a better way 


a18viksa:
As of right now you have to click on the question mark " ? " to close the legend. 
A potential idea would be to make it possible to close the legend by clicking anywhere on it.

a18viksa:
Closing issue after merge to master.

PR #9352, opened 2020-05-15
G2 2020 w21 #9335: Made it easier to close the legend 

a18viksa:
It's now possible to click on the legend to close it. Also changed to `onmouseover` to make it more fluid but if the reviewers prefer `onclick` it can always be changed back. 
**But**, if you use `onclick` instead of `onmouseover` you can't close the legend by clicking on the question mark.

### DIFF
--- a/DuggaSys/resulted.js
+++ b/DuggaSys/resulted.js
@@ -1752,7 +1752,7 @@ function showLegend(){
 	}
 }
 
-// Function that makes it possible to click both on and outside the div to close it 
+// Function that makes it possible to click on the div to close it 
 document.addEventListener('mouseup', function(e) {
 	var closeLegend = document.getElementById('resultedLegendContainer');
 	if (closeLegend.contains(e.target)) {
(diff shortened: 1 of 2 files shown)


b18jonli:
I'm not sure about the onmouseover variant, I think onmouseover works if you close it by clicking outside the legend, and prefer onclick if you close by clicking the legend itself. 

Onmouseover also leads to this interesting situation when you try to close it:

a18viksa:
@b18jonli I noticed that too. Perhaps it's best to just use `onclick`.

b18jonli:
I approve this as well, the choice between onclick/onmouseover is really just about opinion, it works either way so you can decide what to go with (maybe make it easily toggleable in the code).

a18viksa:
I will change it to `onclick` and write a comment that (if somebody want to) you can easily change it to `onmouseover`.

a18viksa:
Made some final commits with help from @b18jonli. Now you can click on the " ? " to close the legend.

PR #9361, opened 2020-05-15
Fixed the issue with new code added link being set to"null"

a18jonsj:
giving the link to new code-example the name of section-name and exampleid, so it is no longer set as "null".


a18jenro:
Really good fix!

Issue #9357, opened 2020-05-15
Frågedugga question/answer options not displayed properly


c14jonfr:
Sub-issue of #9310.

Because of an issue with string-splitting in kryss.js the question and answer options are not displayed properly. 

![image](https://user-images.githubusercontent.com/49141758/82050890-2e697200-96b9-11ea-89d0-3e7c34af9eec.png)

a18viksa:
Closing issue after merge to master.

PR #9363, opened 2020-05-15
G2 2020 w21 #9357 Frågedugga should now render properly

c14jonfr:
Solves #9357. 

String inputSplit[i] was split on wrong characters causing question/answer options to render as 'undefined' (see linked issue for screenshot). Corrected this and now the dugga should render properly.

image


a18oscte:
How do i test this? Can't find where kryss.js is used.

c14jonfr:
@a18oscte This is the Frågedugga, it should be available by default from the testdata in the Datorgrafik course. Most if not all users should be able to access this dugga.

PR #9377, opened 2020-05-15
G4 2020 w21 issue#9355 Empty points are no longer drawn in origo.

a17robhe:
Test this by turning on dev mode and delete a couple of symbols. There shouldn't be any red crosses at origo


a18antsv:
There were still red crosses in origo. This was because an attempt of accessing middleDivider was made on the object. MiddleDivider only exists for UML classes so attempting this would cause an undefined property being put into the points array. I made a simple fix for this by only attempting to access middleDivider when the symbolkind is UML.

PR #9382, opened 2020-05-17
G3 2020#9242 is solved

a18mahkh:
- to distinguish the students first you have select a course then the course-version field will be enabled after you select a course-version then you can select the students in that course and course-version that you want to send the announcements to.


a18mahkh:
You have to click on the announcement card then it will be marked as read by blurring the announcement card.

Broadcasting feature is separated to an another issue

a18jonkl:
Should the sender also get the notification ?

a18mahkh:
Yes, the sender gets a copy of newly created announcement

PR #9397, opened 2020-05-18
Removed chrome suggestions for filling navheader search bar

a18jenro:


a18jenro:
Apparently these changes do not work for the search bar in duggaed. I made this a separate issue: https://github.com/HGustavs/LenaSYS/issues/9398

a18jenro:
It should work in fileed.

PR #9425, opened 2020-05-18
G1-2020-W21-ISSUE#9099-V2

b18gussv:
Removed AJAX call and made it with a local array instead. This removes any complications and the need to connect to other sources.
Once again the events table should show evendescription and the name of the event
bild


b18gussv:
Changed name on pull request, sorry

Issue #9262, opened 2020-05-14
Course MOTD is not visible for student.


a18jonkl:
If you login with keree (user with only read access) for example he cannot see motd.

PR #9428, opened 2020-05-18
Now students can see MOTD #9262

a18lukto:
You need to login in to "keree" to test this issue, you can change the MOTD with the "brom" account to check that it shows the right message


a18jenro:
It works!

PR #9429, opened 2020-05-18
G4 2020 w21 issue#9375 Shift resize now works

a17robhe:
Test by shift resizing a symbol on the canvas if the symbol keeps proportions then it's working


a17robhe:
Incorrect pull request, ignore

Issue #9419, opened 2020-05-18
Dropdown could be pushed a little bit lower


a18oscba:
As it looks now the heading option is overlapping with the button.
![image](https://user-images.githubusercontent.com/62877630/82209249-3e798f80-990d-11ea-94e6-2f8bb319ea67.png)

PR #9431, opened 2020-05-18
Styling the top fab button | Fix for issue #9419

a17towan:
The top tab button is now styled to have a larger space from the start of the drop-down and the button itself.

This is how it looked before:
image

This is how it looks after this fix:
image
(the button is only green because it's selected)


a18jenro:
Sweet!

Issue #9421, opened 2020-05-18
The last exported date should be taken from the database instead of localstorage.


b18fella:
As it is right now, the last exported date is set into local storage and is then extracted from local storage to be shown on the website. Instead of this, the last exported date should be taken from the database so it always gets the newest value.

a18viksa:
Closing issue after merge to master.

Issue #9014, opened 2020-05-08
It should only be possible to export grades that has not been exported yet.


b18fella:
This issue is a little followup from #8962 where an error would pop up in the console. This issue will fix that error and it will also make the exported work better.

Co-authored-by: a18eripi <a18eripi@users.noreply.github.com>

b18fella:
@HGustavs When you export grades, should it export all the grades or just the new grades? And when it comes to gradeLastExported in the database, should it only update those whose grades were exported there and then, or should it update all of them?

Feel free to ask if I'm not clear enough.

HGustavs:
Well, it is better to export only new grades.

But an option to export all would perhaps be a good idea if some grade is inadvertently missed.

b18fella:
@HGustavs Does an export count as an export when you click on the lad export button or does it count as an export when you click on copy to the clipboard?

a18viksa:
Closing issue after merge to master.

PR #9433, opened 2020-05-18
#9421 Last exported date does now come from the database instead of local storage.

b18fella:
This solves #9421. This also solves a bug from #9014. 

The last exported date is not stored in local storage anymore. Instead, it's taken from the database. This does so that you can't take away the last exported date by removing the local storage item, and it'll always load the newest date when opening ResultEd.php

Co-authored-by: a18eripi <a18eripi@users.noreply.github.com>

### DIFF
--- a/DuggaSys/resulted.js
+++ b/DuggaSys/resulted.js
@@ -38,11 +38,8 @@ function setup() {
     window.onscroll = function () {
 
 	};
-	
-	if (typeof localStorage.getItem("lastExpDate") != undefined) {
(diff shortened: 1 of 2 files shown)


b18jonli:
I did not notice a delay with the popup appearing, as noticed in a previous issue, running the server locally on a windows machine seems to cause a high latency in server responses. @a18ismca, try this branch on the testserver and see if the delay persists.

a18ismca:
you are right @JonatanLinberg. I shoud test with the test server too and not only the local. I see.

PR #9438, opened 2020-05-19
G3 2020#9383 is solved

a18mahkh:
- You can select and de-select all the recipients at once


a18jenro:
Seems to work really well, good job!

PR #9465, opened 2020-05-19
G3 2020#9260 | Removed styling that caused the error

a18jimol:
This error should not be visible anymore 
image
Removed the styling that didn't work and was not needed as ST/SV users should be able to see the navheader.
Test with these two users

mofre

ryrey

Password is ‘password’.


a18jenro:
Good fix!

PR #9466, opened 2020-05-19
Headings are now placed more to the right 

a18viksa:
Added 10px margin-left to give the headings more space to the left.

image


a18viksa:
I know but it's like that naturally. Can be fixed in another issue.

Issue #9407, opened 2020-05-18
User Information: Number of each event type


b17carbr:
There should be information regarding the number of times a users has preformed each event type. This could for example be added to each page selection.

![Screenshot 2020-05-18 at 11 50 57](https://user-images.githubusercontent.com/62876614/82199361-e25b3f00-98fd-11ea-91e5-4fec5edb3493.png)

PR #9470, opened 2020-05-19
Number of times event has been preformed #9407

a18eliek:
image


a18eliek:
View of other user:
image

Issue #9435, opened 2020-05-19
Courseid in Page Information is not displaying correctly


a18eliek:
The hitspread piechart in page information displays an http url instead of the courseid as the lable mentioned. 

![image](https://user-images.githubusercontent.com/49141837/82290021-3b7ead80-99a6-11ea-9de4-e538eabc0780.png)

a18eliek:
After doing some testing, we could use the following regex to get it cid from the refer: `.+?cid=([0-9])` However seems like there's limitations in SQLite in this regard, we could probably ignore the SQLite part and solve this via Javascript if we pass the full refer, alternatively doing this via PHP.

PR #9474, opened 2020-05-19
Page Information CID #9435

a18eliek:
image


a18eliek:
Due to how different URL might look, the query to get the courseid from codeviewer was not working correctly, I've merged another branch that separates this in a better way and added the correct query to this branch. This should solve all issues with this branch. 
image

Issue #9380, opened 2020-05-16
Points should not show for objects in hidden layers


a18antsv:
In developer mode, the points are always visible, even when only referenced by objects that are hidden. It is very distracting to have many crosses over the screen without an actual symbol being shown there. Only points that are referenced in objects that are part of any visible layer should be shown. Keep in mind that some objects can share points (lines can share the attribute center point). So even if the attribute is moved to another layer and hidden, the attribute center point should be shown as it is also referenced by the line. 

![image](https://user-images.githubusercontent.com/49121839/82131058-898e8800-97d1-11ea-93e3-369cbe514fe4.png)

a18antsv:
Related to #9379 and #9381

PR #9475, opened 2020-05-19
Fix cross dots in developer mode -Solves Issue 9380

a18petha:
Link: http://group4.webug.his.se:20001/G4-2020-W21-ISSUE%239380/DuggaSys/diagram.php

*Needs testing*

*test developer mode crosses should be restricted to layers now

Issue link #9380


a18kaswi:
Awesome, crosses from example l1 are now not visible when swaping to l2 and code looks clean!

APPROVED.

Issue #9472, opened 2020-05-19
Fix layout in Random Clipping Dugga.


b18veran:
<img width="1280" alt="Screenshot 2020-05-19 at 11 59 07" src="https://user-images.githubusercontent.com/49142649/82313088-2d418900-99c8-11ea-902c-f0c05a7e821e.png">
<img width="1280" alt="Screenshot 2020-05-19 at 11 59 15" src="https://user-images.githubusercontent.com/49142649/82313093-2dda1f80-99c8-11ea-91c4-038dd0ae0ab6.png">

a18viksa:
Closing issue after merge to master.

PR #9476, opened 2020-05-19
Change infobox and fixed layout on Random clipping dugga. #9472

b18veran:
Now it looks much better than before. But it looks a bit strange when you scale. But I intend to put a new issue on and fix it on the next weeklybranch. 
Screenshot 2020-05-19 at 12 34 35


b18jonli:
Looks good on both firefox and chrome (on mac), the image is now a reasonable size.

PR #9483, opened 2020-05-19
The distancing of the navbuttons are now symmetrical

a18fayha:


a18jenro:
Does indeed look good!

Issue #9492, opened 2020-05-19
Add the latest versions of JQuery


a18eliek:
Add the latest versions of JQuery and JQuery UI. 
The old versions should remain, but if possible be updated to support the new version.
This is due to an XSS that has been patched in newer versions of JQuery.

PR #9494, opened 2020-05-19
Newer version of JQuery and JQuery UI for analytics #9492

a18eliek:
This updates the version of JQuery, all functionality should remain the same, but an extra look should be done to ensure this.

This is due to an XSS exploit that was patched in latest version of JQuery and the version we currently use is from 2014.


a18eliek:
This error is related to dugga.js, I've updated dugga.js to make use of another onload function that removes this error, howerver, this file is not really one that our group fouces on, and that error itself is harmless for analytics.

PR #9496, opened 2020-05-19
G3 2020#9473 fix

a18jimol:
It now work to change a course-version, the link will work


a18jenro:
Seems to work better again! Good job!

PR #9525, opened 2020-05-20
Added new file for 'Rapport inlämning'

### LINKED ISSUE #9521: Bummer! You have reached a non-navigable link!
When grading a specific dugga, one that can have a file uploaded to it:
untitled

You get this message if you open the instructions:
untitled2

b18jonli:
Fixes #9521
Requires reinstallation to add the file to the system.

### DIFF
--- a/DuggaSys/duggaed.php
+++ b/DuggaSys/duggaed.php
@@ -23,6 +23,7 @@ pdoConnect();
   <script src="../Shared/SortableTableLibrary/sortableTable.js"></script>
   <script src="timer.js"></script>
   <script src="clickcounter.js"></script>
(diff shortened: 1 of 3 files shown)


b18jonli:
This fixes both the problem mentioned in #9521 and also a very similar problem when accessing the same dugga via sectionEd/showDugga.

b18jonli:
That is not an error, that is the actual instruction text as you can see in the file contents, in retrospect that was probably not great as a joke (sorry) since it looks confusing

a18eripi:
Following up on @a18ismca excellent comment, and @b18jonli answear, i am satisfied with the fix. Approved

Issue #9516, opened 2020-05-20
Frågedugga submissions/answers not handled properly in Result Editor


c14jonfr:
See screenshot for an example of what a submission from the Frågedugga looks like in the Result Editor.

My understanding of this is:
* The first "Answered:" is the user answers. These are (should be) marked as red if incorrect and green if correct.
* The second "Answered:" is the correct answers for the dugga. Or should be, rather. It seems like it currently shows the correct answers for ALL the duggas in the course which is obviously wrong.

![image](https://user-images.githubusercontent.com/49141758/82416524-7520e880-9a7a-11ea-93e9-59380e7364bd.png)

Aside from fixing the whole issue with the "correct" answers not being shown properly, I suggest changing the second "Answered:" to something more appropriate. "Correct answers:", perhaps?

I also suggest removing the commas from the list of user answers and using a blankspace as a separator between the answers (if there is a Frågedugga with more than one question, in the testdata there are only Frågeduggas with one question but I take it for granted more than one question should be possible). So it could look something like this: "Answered: A B C D".

I think you're supposed to be able to see the actual dugga in the Grading window too. I'm not sure if any work has been put into this, but the instructions accordion is there, so maybe. Either way I consider this a separate issue.

c14jonfr:
The Frågedugga uses the savequizResult function in dugga.js while other duggas use the saveDuggaResult function. I believe the Frågedugga should be able to use the saveDuggaResult as well.

For simplicity's sake any solution to this issue (as in issue 9516) will use the savequizResult function and a new issue will be created at a later date concerning changing the implementation to use saveDuggaResult instead.

a18viksa:
Closing issue after merge to master.

PR #9526, opened 2020-05-20
G2 2020 w22 #9516 Frågedugga results should now be handled properly in Result Editor

c14jonfr:
Solves #9516.

User answers and correct answers for the specific dugga variant the user was assigned to should now be displayed properly. Works for both single-question variants as well as variants with several questions. See screenshots.

Screenshot one. Example of submission from a variant with one question with a correct answer.
image

Screenshot two. Example of submission from a variant with two questions with one correct answer and one incorrect answer.
image

Screenshot three. Example of how a dugga with more than one question looks. Not necessarily related to the solved issue but I'll include it just to show that the process works all the way from opening the Frågedugga to submitting and opening the submission in the Result Editor.
image

### DIFF
--- a/DuggaSys/showDuggaservice.php
+++ b/DuggaSys/showDuggaservice.php
@@ -564,6 +564,7 @@ $array = array(
 		"files" => $files,
 		"userfeedback" => $userfeedback,
 		"feedbackquestion" => $feedbackquestion,
+		"variant" => $savedvariant,
 	);
 if (strcmp($opt, "GRPDUGGA")==0) $array["group"] = $group;
 
--- a/DuggaSys/showDuggaservice.php
+++ b/DuggaSys/showDuggaservice.php
@@ -366,19 +366,16 @@ if(strcmp($opt,"GETVARIANTANSWER")==0){
 	$variant = $temp[3];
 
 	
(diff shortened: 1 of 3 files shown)


b18alegu:
There might be some part to test this pull request that I'm missing. I've done the following:
- Log into LenaSYS as **a99birol**
- Participate in **Frågedugga 1** and answered the question
- Logged in as **Toddler** to grade **a99birol**

I can then see what **a99birol** answered but I do not get the correct answer.
Skärmavbild 2020-05-21 kl  08 37 29

Any suggestions @c14jonfr?

c14jonfr:
@b18alegu Hm, that does not look right.. Let me have a look. I'll get back to you soon.

c14jonfr:
@b18alegu Alright, I think I've got it figured out. The SELECT-query used for this contained the current userid, in this case 2 (Toddler's userid) and it wouldn't work properly unless Toddler had submitted an answer for the Frågedugga (if not there was no useranswer to select).

The way it works currently, the selected useranswer is not even used, so it could be omitted entirely but I decided to only make the necessary changes for now (= removing userid from the query).

I have tried this on a fresh install using the steps you describe and I am able to get the desired results/behaviour. Pushing a new commit in a second.

PR #9527, opened 2020-05-20
Added section in general stats that displays most recently edited file.

a18wilis:
Added a section that displays the name of the most recently edited file in LenaSYS, along with the time it was edited.

Test by editing a file in the fileed-section and then observing the general stats section of analytic after the edit has been made. The new section should look like the following image:

image

### DIFF
--- a/DuggaSys/analytic.js
+++ b/DuggaSys/analytic.js
@@ -189,6 +189,11 @@ function loadGeneralStats() {
             'Created: ' + data['stats']['newestFileTimestamp']
     	]);
 
+		// Most recently edited file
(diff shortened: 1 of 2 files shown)


a18wilis:
The system no longer attempts to convert nonexistent timestamps

The error was due to faulty code which always made an attempt to convert the timestamp displayed to a better format, even if there were no value in the timestamp. When that was the case, the system simply tried to format a null value, thus causing the error.

 This fix makes it so that the system only tries to format when there's an actual value in the timestamp, and leaves the null-timestamps to display their null value on the table.

Issue #9462, opened 2020-05-19
Navheader issues in mobile view in sectioned.php


a17towan:
Currently there are some issues when it comes to the styling in the navheader in mobile view.
![image](https://user-images.githubusercontent.com/49142028/82305988-b9e74980-99be-11ea-8881-e9b6bf8dafc2.png)

![image](https://user-images.githubusercontent.com/49142028/82306401-4e51ac00-99bf-11ea-87e4-a0bcad475264.png)

As you can see, there are several problems. The icons to the left become smaller while the icons to the right remain the same size. It's inconsistent.

The green box for the 'logout' option overflows the boundraries of the navheader.

When there isn't enough space, the icons to the right move down below the navheader in front of the MOTD banner.

PR #9535, opened 2020-05-20
Sectioned.php navbar mobile view styling | Fix for issue #9462

a17towan:
Did some styling to the navbar in sectioned.php because it looked weird on multiple points when you resized the width.

I made the decision to not scale the icons to the right down because they ended up looking way too small. Instead I just scaled down the "acessed.php" icon when the window gets smaller than 339 px in width.

image
image


a18jenro:
Looks better than before.

PR #9539, opened 2020-05-20
G3 2020#9477 good news

a18mahkh:
- multiselect of recipients works without holding down the ctrl key

!OBS:
the update function did not work before, I had to fix that problem prior to solve this issue, that's why I didn't create an separate issue for that.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -2127,6 +2127,15 @@ function selectRecipients(){
       $("#recipient option").attr("selected", false);
     }
   });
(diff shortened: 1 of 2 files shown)


a18mahkh:
There is a commit saying:
'select and de-select single recipient onclick without holding down the ctrl key is working.'

this would select recipient which is true, but would only send announcement to the last selected recipient.

I detected that problem later on.

Check the commit 'good news' and downwards for fully functional feature.

a18oscba:
Nice feature, works fine!

Issue #9522, opened 2020-05-20
Random clipping dugga need some fixes. 


b18veran:

b18veran:
There are still problems. See img 1 there are white edges at the bottom. See img 2 when adding more than 19 operations. New issues will be added to it. Otherwise everything should work.

<img width="1280" alt="Screenshot 2020-05-20 at 12 46 53" src="https://user-images.githubusercontent.com/49142649/82438350-32bad400-9a99-11ea-84b5-4734e471c5fd.png">
IMG 1

<img width="1280" alt="Screenshot 2020-05-20 at 12 47 22" src="https://user-images.githubusercontent.com/49142649/82438348-32223d80-9a99-11ea-894d-ecd093feae43.png">
IMG 2

a18viksa:
Closing issue after merge to master.

PR #9540, opened 2020-05-20
Fixed layout in Random clipping dugga. #9522

b18veran:
- Moved some inline styling. 
- Added shadows to make a card feeling.
- Cleaned up some code.
- Added a paragraph to the "Rityta ".
- Added a media query to make it more mobile-friendly. 

Screenshot 2020-05-20 at 12 53 11
Screenshot 2020-05-20 at 12 54 16

### DIFF
--- a/DuggaSys/templates/clipping_masking_dugga.html
+++ b/DuggaSys/templates/clipping_masking_dugga.html
@@ -23,7 +23,7 @@
             
     </div>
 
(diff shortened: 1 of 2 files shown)


b18veran:
@a18marax I understand what you mean and it's hard for me to test on larger units than I have. Have created a media query for it now.

Issue #9461, opened 2020-05-19
StudentTeacher and Supervisor users can't see hidden items


a18jimol:
Users with StudentTeacher and Supervisor access can right now create new items(headings, code-examples, duggas) in sectioned but can't see them because the new items have the visiblity hidden and ST/SV users can't see hidden items.

Studentteacher: Got 'ST' access for all courses.
mofre

Supervisor: Got SV access for all courses.
ryrey

Password is ‘password’.

a18jenro:
Seems to have been solved by itself.

PR #9543, opened 2020-05-20
Gave studentteacher access to see hidden items || issue #9461

a18jimol:
Users with studentteacher and supervisor access can now see hidden items(duggas, headings, codes).

test with:
mofre

ryrey

Password is ‘password’.


a18jenro:
Looks good!

Issue #9187, opened 2020-05-12
Unnecessary saveState


o18marka:
Currently the saveState() function is run every time user presses the delete key. State should only be saved is something actually changes in the diagram, in this case when something is actually deleted.

PR #9545, opened 2020-05-20
solves issue#9501

b18simla:
Error removed when pressing delete and no object is selected

Try it: http://group4.webug.his.se:20001/G4-2020-W22-ISSUE%239501/DuggaSys/diagram.php


b18simla:
Also removed one redundant savestate() call in deleteFreeDrawObject(), since this is issue was strongly related to https://github.com/HGustavs/LenaSYS/issues/9187

PR #9556, opened 2020-05-20
Added all pages to page information dropdown - G1-2020-W22-ISSUE#9305

b17carbr:
All pages now appear in page information dropdown, they are all connected to their respective log db query in analyticService.php. The queries needs to be modified to show useful information for each page, this needs to be set done in another issue. 

Screenshot 2020-05-20 at 15 17 48


a18jakme:
drowpdown list sorted and diagram.php is added. Well done!
image

Issue #9158, opened 2020-05-12
Rework how undo/redo works with the local storage 


a18andtu:
Right now the undo and redo functionality loads diagrams from the local storage based on diagram number. This is not compatible with the idea of having only one state at a time saved in local storage mentioned in issue #8823. A solution could be to save previous states in the current state.

a18antsv:
I think my solution to only save changes could solve this issue as well in a good way. I am currently developing it and it looks promising.

a18antsv:
This should be solved at the same time as I work for only saving changes in local storage and building the diagram from them.

Issue #8823, opened 2020-05-05
One diagram in localstorage at a time.


b18phibe:
As the title says we only want one diagram state.(one save of the diagram) in localstorage at a time then when the new save are taken the old one are to be written to a file in the database. The same file as all other diagram states from this diagram. So in the end this diagram you are working on only has one file in storage making it easyer to late make an animation of the diagram and to download the full progress of the diagram. An idea can be to have a single zip file storing the diagram.
 
We don't want the local storage to look like this.
![image](https://user-images.githubusercontent.com/62876531/81044556-dfe7f680-8eb4-11ea-94da-094bc12f93e6.png)

may not be worth doing before #8819 are done.

b18phibe:
This might brake the undo and redo buttons (ctrl + z / ctrl +y), please check so this functionality still works after this issue has been solved.

a18andtu:
Right now the undo and redo functionality loads diagrams from the local storage based on diagram number. If the intention is to only have one diagram in local storage at one time this has to be reworked before going through with the changes for this issue,

a18antsv:
I think it would be fine to store everything in local storage. We just need a solution where only the changes are saved. Then we can build the diagram by these changes. And it would be possible to undo and redo, going back and forth between the changes. This information could all be stored as one string in local storage. I would say this is necessary if we want to write all information to a file and be able to replay what was done in that diagram project. Right now, only the last seen diagram is saved in files which will make it impossible to implement this animation feature. I am thinking to try to work on this only changes saving for issue #8819. I already solved #9231 which compresses the property keys to a 2 character string. This compression could however be of no big use if I can manage to save only changes.

PR #9560, opened 2020-05-21
Fixes #8819, #8823 and #9158 - reworks storage to save only changes

### LINKED ISSUE #8819: Reduce the amount of used storage
in Issue #7902 one of the problems are that each diagram save takes up to much storage space a comment in that very big issue (now split in to several smaller) is: 

> This should not make use of several megabytes.
> 
> If we add an entity.
> How much should that take up? Lets say for arguments sake 200 bytes.
> Then we can have 5000 entities or 5000 state changes on one entity per megabyte.
> 
> Also we do not want to store every single miniscule update of the diagram... If an entity is moved around a lot, and we place it in 10 different positions before adding the next entity. Perhaps some of those places can be ignored. Perhaps we only store changes every x seconds or only when new items are added.
> 
> Also the information is VERY un-compact.
> 
> We store true/false when we could store 1 or 0 ... making the data 5 times larger for no benefit compared to the space it occupies.
> 
> We can call "symbolkind" "sk" reducing the number of bytes of that property by a fifth.
> 
> We also had discussions last year about omitting "empty" data from the textual version.
> 
> Since neither operations:[] or attributes:[] contain anything I would ommit that data completely, reducing that information from nearly 30 bytes to 0 for storing that (lack of) information.

one more thing to do to lower the amount of storage needed are to not save the diagram all the time but instead every 10 or 20 secounds or every time an new object are added or removed (what ever you thinks will work best). How do you handle ideal time say if a student goes to sleep not closing the diagram tab. 

**Important are to not remove the feature of multiple saves because then the key feature  of replaying the work process of how the diagram was built.**

a18antsv:
Issues #8819, #8823 and #9158 can be closed after merging this.

This reworks the way storage is handled completely. Now only changes will be stored in local storage and in files. Then the diagram will be built with the help of the stored changes for each state. This solution will save an incredible amount of storage space so it should not really be possible for it to run out anymore. Undo and redo was also reworked to fit the new way of storing the information. I also fixed so user has to confirm if the current state is not the latest saved state (undo has been made and an attempt to create new objects or do other changes is made) to prevent states to be overwritten.

- Clear local storage.
- Test to create all types of different objects. Copy and paste some. Undo and redo some. 
- Test to undo several times and then create objects. There should be a confirmation dialog telling how many states behind the latest save you are. If confirming, the new state will be applied and changes that are after the current will be invalidated. If not confirming the new state will not be applied and nothing should change. 
- Refresh the page to see so the state is maintained. Test all previous things after refresh as well.
- Test to generate example diagrams. See so they generate correctly and also try to undo and redo in the example diagram process, this was not possible before.

Link: http://group4.webug.his.se:20001/G4-2020-W22-ISSUE%238819%238823%239158/DuggaSys/diagram.php


a18antsv:
I found a problem when undoing lines. So wait with testing this until I have fixed this problem.

a18antsv:
I found a solution. Now I just need to recreate the example diagrams again for them to not break when undoing lines.

Issue #9566, opened 2020-05-22
Quiz/Frågedugga uses it's own function to save results instead of generic saveDuggaResult


c14jonfr:
There should be no problem using the saveDuggaResult for the Quiz/Frågedugga too. This will add a few things "for free", such as the "popup"-window which shows after submitting a dugga with a receipt, a way to leave feedback if it's enabled etc.

a18viksa:
Closing issue after merge to master.

PR #9567, opened 2020-05-22
G2 2020 w22 #9566 Quiz/Frågedugga now uses saveDuggaResult instead of it's own saveQuizResult

c14jonfr:
Solves #9566.

**Note: This is branched from https://github.com/HGustavs/LenaSYS/tree/G2-2020-W22-%239516 since it relies on the changes made there to work.**

Removed saveResultQuiz and replaced it with saveDuggaResult, which the other duggas use.

As the linked issue describes this adds a few things "for free", such as this window upon saving:

image

Before no window would be shown, the only visual feedback given to the user/student was the navheader being updated (the stoplight and timestamp for submission would update). As with the other duggas the Feedback-area of the window would only be show if feedback is enabled for this dugga. Opening feedback from the Section/Course Editor works exactly the same as for all the other duggas.

### DIFF
--- a/DuggaSys/showDuggaservice.php
+++ b/DuggaSys/showDuggaservice.php
@@ -564,6 +564,7 @@ $array = array(
 		"files" => $files,
 		"userfeedback" => $userfeedback,
 		"feedbackquestion" => $feedbackquestion,
+		"variant" => $savedvariant,
 	);
 if (strcmp($opt, "GRPDUGGA")==0) $array["group"] = $group;
 
--- a/DuggaSys/showDuggaservice.php
+++ b/DuggaSys/showDuggaservice.php
@@ -366,19 +366,16 @@ if(strcmp($opt,"GETVARIANTANSWER")==0){
 	$variant = $temp[3];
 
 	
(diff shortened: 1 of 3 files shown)


a18marax:
Found that the input field is to big for the window but seems like that this could be a bigger problem since it works on chrome so maybe it need to adjust to the screen or something like that

c14jonfr:
@a18marax This issue/pull request does not deal with the Duggasvar-window layout what so ever. If it is an issue, a new issue about it should be opened. This pull request only deals with "enabling" the window so to speak as a "side effect" of using the standard saveDuggaResult function.

As for the alert error I am not able to replicate it. The database is updated correctly when I save the result, so it's not an issue of the alert not being shown for me. I will look into this a bit but I have a feeling it's a local problem for you.

c14jonfr:
@a18marax Issue should be fixed now with the latest commit. I've tried it on the testserver and it now works there, please test it again locally on your machine.

Also the feedback has to be enabled on the coursepage for the specific dugga. Click the cogwheel to the right and check "Enable Student Feedback" in this window:

image

PR #9571, opened 2020-05-22
Added what courses user partake in - G1-2020-W22-ISSUE#9356

b17carbr:
Added what courses user partake in user information.

Screenshot 2020-05-22 at 11 13 34


a18jakme:
What is the meaning of the other options in the dropdown? 
Example a student with a page, is it a necessary option?
SC:
image

PR #9574, opened 2020-05-22
Timestamp now added to css and JS files to auto-refresh in fileed.php

a18filkr:
This should now refresh css and js files everytime the php is loaded.


a18jakme:
Im not really sure how to test this.

a18filkr:
bild

The files should have a timestamp in the inspector, if you change something in a file and save, the timestamp should be a new one.

PR #9592, opened 2020-05-22
G3 2020#9464 Search bar in mobile-view looks better

a18jimol:
The issues existed in accessed, fileed and duggaed and now the search-bar should look like this in mobile view:
image
Also added the tooltip box (the one with the "?") to fileed.


a18oscba:
Looks definitely better now, good job.
It seems like we're not supposed to reach that page from mobile but it makes sense that everything works if people change their mind the future.

PR #9612, opened 2020-05-24
Fixes #3744 - Added timeline for diagram states and animation support

### LINKED ISSUE #3744: Diagram editor animation support
We want to see the full history of a diagram as an animation.

e.g. we click a diagram and we se an empty diagram area, and then click a "play" button, at which point we see all of the actions to arrive at the final diagrma as a sequence of animation frames.

a18antsv:
You can test by loading any example diagram (Help -> Generate example diagrams) or by creating your own diagrams. Or do both.

The original issue #3744 was about adding animation support for the diagram. I took it one step further, also creating a timeline making it possible to manually go to any available diagram state. In my opinion this is even more useful. I made sure the timeline works in all scenarios; when rulers are active, when full screen is active, when rulers and full screen mode are active at the same time. **Test** all of these scenarios.

The timeline gives good response for which states that were included, will no longer be included after clicking a hovered part on the timeline and which states that were not included, but will be included after clicking a part. **Test** so this works. Also **test** click on a part and see so the right state is loaded.

**Test** to click the play button to animate the diagram creation. Every second, the next part of the timeline should be loaded and when reaching the end of the timeline, the animation will start from zero again. **Test** to pause it.

**Test** to click the "+" button, an element with a few extra buttons should animate to show. Test to use the step backwards/forwards buttons. **Test** the full screen mode button that is in the end of the timeline. **Test** the reset button, the timeline should have no colored steps and the diagram should be empty. **Test** the speed slider (0,1s - 3s is possible right now) so the correct speed is applied to the animating diagram. It should be possible to change speed while animating and it should be directly applied to the animation. It should also be possible to reset the diagram while animating without any problems.

When attempting to change the diagram (add, move, change property) when not in the latest state, a confirm dialog should appear giving information about how many states will be invalidated. When cancelling this, the correct state should be loaded again, preventing new actions to show (I added this in previous pull request, but the actions were shown but not saved, now they are never shown as well). **Test** this. Also **test** to accept this confirmation, all states after current one should be invalidated, removed from the timeline and one new state added for the new action.
image

A similar dialog should now appear when attempting to clear the diagram (File -> Clear diagram). **Test** to clear the diagram to see so the timeline clears as well and attempt to start creating a new diagram, the timeline should fill up again.

Link: http://group4.webug.his.se:20001/G4-2020-W22-ISSUE%233744/DuggaSys/diagram.php

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -6373,14 +6373,27 @@ function updateTimeline() {
         const part = document.createElement("div");
         part.classList.add("diagram-timeline-part");
         if(i <= diagramChanges.indexes.current) {
-            part.classList.add("done");    
+            part.classList.add("included");    
         }
         timelineElement.appendChild(part);
     }
 }
 
-function timelineHover(e) {
-    const elementIndex = getElementIndexInParent(e.target);
+function timelineMouseOver(e) {
+    const hoveredPartIndex = getElementIndexInParent(e.target);
+    const timelineParts = document.querySelectorAll(".diagram-timeline-part");
+
+    timelineParts.forEach((part, i) => {
+        part.classList.remove("plus", "minus");
+
+        //When hovered part is after current iteration part and current iteration part is not already a shown state, current iteration part should be marked green
+        //When hovered part is before current iteration part and current iteration part is a shown state, current iteration part should be marked red
+        if(hoveredPartIndex >= i && !part.classList.contains("included")) {
+            part.classList.add("plus");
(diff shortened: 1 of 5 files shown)


a18antsv:
I will add a slider for animation speed in a new issue. I could also add a reset animation button which will reset to the state before the one shown on the timeline to start over from nothing.

a18antsv:
There will be another pull request merged into this one, so wait with merging this.

a18antsv:
Now this one can be merged.

PR #9615, opened 2020-05-24
Fixes #9613 - speed slider and reset button to diagram timeline

### LINKED ISSUE #9613: Diagram timeline animation: add speed slider and reset button
It would be good to be able to control the animation speed; the amount of time it will take to switch to the next diagram state. It could also be good with a reset button resetting the diagram to before the first state, so the animation will restart.

a18antsv:
A speed slider and a button to reset the diagram timeline has been implemented. All of the extra timeline controls are now under a "+" button which on click will animate the rest of the control buttons to show. This was a way to save space. When resetting the timeline, the timeline should have 0 colored steps and be empty. The speed slider can go from 0,1 meaning 0,1 seconds per state change to 3 seconds per state change right now and changes by 0,1 for each step. It should be possible to interact with all the controls without stopping the animation.
e87f8fbdee9750ff985ebaace5ac7070

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -6445,22 +6445,34 @@ function toggleTimeline() {
     canvasSize();
 }
 
-function playTimeline(playButton) {
+function playTimeline(isSpeedChanged = false) {
+    const playButton = document.getElementById("diagram-timeline-play-button");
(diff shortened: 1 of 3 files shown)


a18antsv:
I am merging this into my other branch which implemented the timeline and animation functionality. That branch is not merged into weekly yet.

Issue #9595, opened 2020-05-22
Replace update functions in init_db with postgres alternative


a17robhe:
TIMESTAMP DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP doesn't work in postgres, as far as we can tell triggers needs to be used.

PR #9631, opened 2020-05-25
Sql migration w22 issue#9595 Update timestamp updated to postgres syntax

a17robhe:
Updating timestamps are now handled by triggers, a standardized variable name had to be implemented though so the system needs to be refactored for this as well.

Closing #9595


a17robhe:
As we realized, changing variable names would invalidate the old database which we want to avoid. Implementing multiple update functions for the different variable names is preferable

PR #9651, opened 2020-05-25
Update functions implemented without changing variable names #9595

a17robhe:
same as https://github.com/HGustavs/LenaSYS/pull/9631 but without changing the names of the rows in the tables


a17robhe:
http://group4.webug.his.se:20001/SQL-MIGRATION-W22-ISSUE%239595/

Issue #9151, opened 2020-05-12
Headers to sorting and filtering pop-ups in are missing and too small pop-up menus for filtering/sorting (Navheader issue)


a18ismca:
The drop-down menus for sorting/filtering is vertically too long, because they stretch down to more than the middle of the screen. Thus, it would be good to increase these both menus **horizontally.** For example a drop-down menu when you are at Zalando and purchasing shoes, and there is several sub-categories etc.

It would also be good with some headers explaining the meaning of each show/sort in these dropdown menus. For example...

See these pictures below. Headers are missing in both picture. 

It would be good [in the first picture] to have two headers named **"Show..."**, **where the bold-marked are "required" to mark at least the first two ones to have a "maximized" database.**
![image](https://user-images.githubusercontent.com/62876523/81666673-9f532480-9442-11ea-930e-f89df4527a9f.png)

In the second picture, you can select three things to sort. The first is to sort asc, desc. The second is to sort by and the third sort for. It would be good with three headers named "Sort...","Sort by...", and "Sort for..."

![image](https://user-images.githubusercontent.com/62876523/81668177-92373500-9444-11ea-893b-ad0a1c15bf7f.png)

_**Hint: The drawing painted in blue shows an example of where you can add headers that describes each sort/show menu...**_

HGustavs:
Any styling cue should not take up more space (a few pixels is ok) but it should not add new rows to the content as it is already critically cramped for courses with much content.

HGustavs:
If there is a way to reduce the number of rows (wider?) this is even better.

a18ismca:
@HGustavs yes, that can also be a solution btw. If you want I can re-write this issue so it also includes your thoughts. However, I think it should be bigger fonts explaining each section. For example you go into H&M's website and when the filter options are shown up, there are fonts, and that is missing in LenaSYS (navheaders). That is why this issue was made.

a18jonsj:
@a18ismca If i understand,

a18ismca:
@a18jonsj exactly. The dropdown-list (or the menu, idk how to call it) should be wider and have categories depending on how many sections there are. If you see the example above (with the pictures), when you open the sorting menu, it would be good with three "categories"... if you understand what I mean.

a18jonsj:
![image](https://user-images.githubusercontent.com/49142617/82800201-923f2800-9e7b-11ea-852b-fd9662c6c536.png)

[...]

PR #9653, opened 2020-05-25
G3 2020 #9151

a18jonsj:
The dropdowns in "edit-student result" are now split into 2 columns and have a border dividing them, making it clearer. Related to #9151.


a18jenro:
Henke seemed to be pleased and so am I.

Issue #9645, opened 2020-05-25
StudentTeachers and Supervisors should be able to delete items in sectioned


a18jimol:
StudentTeachers and Supervisors are supposed to have the same rights as teachers(writeaccess) see #9423 so the trashcan for deleting items should be visible and work correctly

a18jenro:
@a18oscba Why did you close this issue, nothing was merged?

a18oscba:
network issues idk I did merge

a18jenro:
Yes, now I can see that it was merged. NVM!

PR #9654, opened 2020-05-25
G3 2020 #9645 | StudentTeacher/Supervisor delete items

a18jimol:
Now StudentTeacher/Supervisor users can see the trashcan icon on items in sectioned and delete the items.

test with:
mofre

ryrey

Password is ‘password’.


b18emigu:
Apparently @HGustavs has said how **studentteachers** & **supervisors** should work in the course section. They should be allowed to create, delete, edit and read items, same as regular teachers.

PR #9672, opened 2020-05-26
Added media query for smaller plattforms.

b18veran:
- Moved inline styling to dugga.css. 
- Added media query for smaller plattforms. 
- Removed all tables, tr and td for the content and added div to them. 
- Fixed styling on buttons in #input-col. See img 1 & 2
- Fixed some styling on the preview-col. 
Screenshot 2020-05-26 at 08 30 16
img 1 - After

Screenshot 2020-05-26 at 08 31 12 
img 2 - Before


b18veran:
This dugga looks more similar to the Random clipping dugga now.

PR #9701, opened 2020-05-26
G3 2020#9594 fixed error in fileed when using mobile search-bar.

a18jonsj:
The search function didn't work as intened in fileed in mobile-view below 530px when using the mobile search-bar. We got this error:
image

This is now fixed and you no longer get an error and the search-function works correctly.


a18jenro:
Sugoi!

PR #9705, opened 2020-05-26
minimizeBoxes now better fits the code standard

a18wilis:
Replaced JQuery functionality with native Javascript equivalent and fixed incorrect indentation according to the standard.

Test by checking if the minimizing functionality still works on every box in every template, as it should work without any problems just like before the update.


a18geoge:
Perfect!

a18wilis:
The reason why the mobile functionality was removed was simply because it wasn't used anymore, and thus was considered unnecessary code.

Issue #9688, opened 2020-05-26
User information: load information update and position


a18jakme:
When changing users in dropdown list, the load information are not getting updated.

SC:
![image](https://user-images.githubusercontent.com/49142935/82885549-99cb0380-9f45-11ea-8804-a6c8dcc82443.png)
![image](https://user-images.githubusercontent.com/49142935/82885560-9d5e8a80-9f45-11ea-882c-041684a31751.png)

SC - how it suppose to look like:
![image](https://user-images.githubusercontent.com/49142935/82885624-b23b1e00-9f45-11ea-891e-fb5242025502.png)
![image](https://user-images.githubusercontent.com/49142935/82885635-b49d7800-9f45-11ea-9895-df9bfef2f40c.png)

PR #9717, opened 2020-05-26
Updated counter and page load behaviour on user change. Issue #9688

a18wilis:
ezgif-7-40e080997c31

The information table now always updates on user change.

The counter is now removed on user change and reappears after the table update has been made (displayed in the gif above). This prevents those weird positioning issues with the dropdown lists for users and pages. 

Test by changing user while having selected a page (such as courseed) in the User Information section of Analytic.

### DIFF
--- a/DuggaSys/analytic.js
+++ b/DuggaSys/analytic.js
@@ -1219,7 +1219,7 @@ function loadUserInformation(){
         });
     }
 
-   
+	var currentPage;
     function updateState(users){
         $('#analytic-info > select.file-select').remove();
         var userSelect = $('<select class="file-select"></select>');
@@ -1269,7 +1269,9 @@ function loadUserInformation(){
 				localStorage.setItem('analyticsLastUser', $(this).val());
 			} catch(err) { }
 
-			selectPage.change();
+			if(currentPage != "events" && currentPage != "fileEvents"){
+				selectPage.change();	
+			}
 		});
         $('#analytic-info').append(userSelect);
 		userSelect.change();
@@ -1281,6 +1283,7 @@ function loadUserInformation(){
 			firstLoad = false;
 		} 
         selectPage.change(function(){
+			currentPage = selectPage.val();
             switch(selectPage.val()){
                 case "resulted":
 					updateResultedInformation();
@@ -1318,7 +1321,7 @@ function loadUserInformation(){
                 case "loginFail":
 					updateloginFail();
 					break;
-            }
+			}
         });
     }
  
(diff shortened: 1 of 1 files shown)


a18wilis:
I think this solution could be improved by somehow removing the delay between the removal and the reappearance of the counter. In my opinion this solution is still way better than how the counter and dropdown lists currently behaves.

a18eliek:
> I think this solution could be improved by somehow removing the delay between the removal and the reappearance of the counter. Anyhow, this solution is still, in my opinion, way better than how the counter and dropdown lists currently behaves.

We could replace the text if the element exists, we could possibly maybe implement some some sort of loading functionality, however in my test cases this happens in a few milliseconds and I don't really see a problem with the delay, I've heard of some users on Windows having some extended page loading times, but in my testing this wouldn't be an issue either local or on the test server.

DuggaSys/analytic.js
```diff
@@ -1268,6 +1269,7 @@ function loadUserInformation(){
 				localStorage.setItem('analyticsLastUser', $(this).val());
 			} catch(err) { }
 
+			selectPage.change();
```
a18eliek, line 1272: The following row is executed a couple of times each second and causes after a while a browser crash. I would probably not blame this PR for the cause of this infinite loop, however with this this addition the site becomes unusable after a while. 
This is a console log on the row just under this addition to the codebase:
image

I believe this is also the cause of the issue in the following GIF that makes the dropdown not function after switching pages a couple of times. (My clicks are outlined with the black circle.)
userinfo

a18wilis:
The browser crash was due to `selectPage.change();` running indefinitely due to pages "event" and "fileEvent", both containing foreach-statements which causes `selectPage.change();` to loop.

The page no longer updates on user change if currently selected page is "event" or "fileEvent". This still causes some bugs when changing user on these two pages, but updating the other pages still seems to work without problems, and the browser will no longer crash when updating certain pages.

DuggaSys/analytic.js
```diff
@@ -1268,6 +1269,9 @@ function loadUserInformation(){
 				localStorage.setItem('analyticsLastUser', $(this).val());
 			} catch(err) { }
 
+			if(currentPage != "events" && currentPage != "fileEvents"){
```
a18eliek, line 1272: loginFail causes an infinite loop, the other pages seem to work fine after this change, I'd suggest adding this as well.

a18wilis:
The page no longer updates on user change if currently selected page is "loginFail".

a19hammi:
Outdated pull-request

PR #9731, opened 2020-05-26
Fixes #6888 - Sub-collapsible for styling

### LINKED ISSUE #6888: Remove local appearance from double-clicking
Remove local appearance from double-clicking to a menu item to not encourage random colors and styling on the diagram.

It should be behind an "advanced settings" of some sort according to the customer.

The dubble-clicking need to still be there for entity-type and name and so on.

image

a18antsv:
According to issue #6888 local styling appearance should be removed from from double-clicking to not encourage random colors and styling on objects. 

I did not think it was a good idea to separate the styling options in the appearance menu from the double click appearance menu functionality. It would in my opinion require too many steps to select an item and navigate to the menu-item for advanced settings. I think everything should be accessible from the normal appearance menu.

To not encourage styling on the objects, sub collapsibles will now be generated for these advanced settings. The sub collapsible will only be created if the main collapsible will not only contain advanced properties (can happen when selecting many different objects at the same time as the form groups are divided to many different collapsibles). It would be unnecessary to have empty main collapsibles containing a sub-collapsible for the advanced properties, therefor this exception.

Right now, the 5 basic stylings can possibly appear in an advanced sub-collapsible. New form groups can easily get this functionality by adding the HTML-attribute data-advanced to the form group.

Test the appearance menu. Test to open the appearance menu for single objects and also for many objects at the same time. Test so the advanced collapsible works. It should only appear when a main collapsible is opened and there should not be an advanced collapsible if all form groups in the main collapsible are advanced.

Link: http://group4.webug.his.se:20001/G4-2020-W22-ISSUE%236888/DuggaSys/diagram.php

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -5578,9 +5578,15 @@ function showFormGroups(typesToShow) {
     initAppearanceForm();
 
     collapsibleStructure.forEach((object, i) => {
-        const normalGroups = object.groups.filter(group => typeof group.dataset.advanced === "undefined");
         const advancedGroups = object.groups.filter(group => typeof group.dataset.advanced !== "undefined");
-        createCollapsible(normalGroups, object.types, i, advancedGroups);
+
+        //Create normal collapsible with no sub-collapsibles if there are only advanced properties or no advanced properties.
+        if(object.groups.length - advancedGroups.length === 0 || advancedGroups.length === 0) {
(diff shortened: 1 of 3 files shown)


a18antsv:
This is how it is supposed to work right now. This is to make it possible to style everything that is relevant for each type. It was the best way I could come up with to generate the appearance menu so everything can always be styled.

b18phibe:
Alright. then this is a working solution

Issue #9736, opened 2020-05-27
User Information uses splits from the refer while we have columns that include information needed.


a18eliek:
A bunch of unnecessary string splits:

Getting coursevers:
```
if(row.refer.includes("coursevers=")){
    pageParts = row.refer.split("coursevers=");
    vers = pageParts[1];

    if(vers.includes("&")){
        pageParts = pageParts[1].split("&");
        vers = pageParts[0];
    }
}
```

getting courseid:
```
if(row.refer.includes("courseid=")){
    pageParts = row.refer.split("courseid=");
    pageParts = pageParts[1].split("&");
    cid = pageParts[0];
}
```
This is not needed anymore and the analytic service can instead return the correct courseid and coursevers for us.

a18filkr:
This has been merged, closing.

PR #9739, opened 2020-05-27
Refraction of Cvers and CID to not use splits #9736

a18eliek:
This PR only has code refractions


a18filkr:
Hard to test refactoring but will check so the site works as it should.

a18filkr:
bild
Seems to work on testserver.

Issue #8698, opened 2020-04-30
New feedback icon


a18mahkh:
Need an new feedback-icon prior to issue #8697

a18jenro:
@a17towan This has been in progress for a while, how is it going?

a17towan:
I have ideas in store but cannot implement them until the OP has implemented a 'feedback' functionality where I can test the icons.

a17towan:
The icon is now ready for implementation, but issue #8697 needs to be merged into the weekly group branch first.

PR #9742, opened 2020-05-27
Added a feedback icon | Fix for issue #8698

a17towan:
There is now a feedback icon (with a shadow effect) for the new feedback functionality.
image

You need to be logged in as a student to be able to see it (E.g.  User: keree. Pass: password)


a18jenro:
Very much sugoi!

Issue #9735, opened 2020-05-27
In Access Editor, Reset Password buttons are styled with id's instead of class and new-user styling isn't applied


c14jonfr:
One shouldn't use the same id for several elements. It will still render properly but it is not considered valid HTML. 

The id='reset-pw' should be changed to a class instead. 

As for the new-user styling this should be applied when to users considered new, this is done in accessed.js.

The new-user styling is also expecting the button to be an input, since the Access Editor now uses div's for most things (improves performance when many students are in a course) this should be changed to div.

These problems seemed to have begun when https://github.com/HGustavs/LenaSYS/pull/8737 was merged. As far as I can tell the issues presented here should be the last problems coming from that merge.

a18viksa:
Closing issue after merge to master.

PR #9743, opened 2020-05-27
G2 2020 w22 #9735 

c14jonfr:
Solves #9735. 

reset-pw is now a class instead of an id.

new-user class should now be applied to new users (but it's a bit unclear how to create/add new users at the moment, something seems to have broken at some point). The styling is applied properly, at least (can be tested by manually adding the class new-user to one of the buttons in Developer Tools).


c14jonfr:
I had missed to add the reset-pw class to new users (it should have both reset-pw and new-user, as well as submit-button for classes). Corrected this in the latest commit.

PR #9770, opened 2020-07-26
Fixed button row bottom margin by removing its fixed height in style.css

a16silph:
image


a19hammi:
Outdated pull-request

PR #9771, opened 2020-07-26
Canvas dark mode switching fix

a16silph:
Made the analytic page refresh when it detects that you have switched between light and dark mode in analytic.js.


a19hammi:
Outdated pull-request

PR #9772, opened 2020-07-26
Fixed page information error when there is no values from the database.

a16silph:
Fixed page information error when the selected list element's refer was null by adding a null check to that part of the code. You can now keep browsing Page Information after selecting an empty list element. 

image


a19hammi:
Outdated pull-request

PR #9773, opened 2020-07-31
Rotation function completed

shortformikael:
As it stands, in 40 hours of work I was able to create a working function that rotates the targeted FreeDraw object. The function uses trigonomic math where the unit circle helps to rotate the objects vector on an axis (that axis being the center of the object aka origo). 

Currently the rotation is set to 30 degrees but is easily adjustable in diagram.js (the _theta_ variable of the function rotateFreeDraw()). It works perfectly fine on a square but bugs out on oddly shaped polygons somehow. 

The issue of the displaying an arrow that the mouse could drag to adjust the rotation proved to be more complicated than the rotation of the object. I would suggest separating the rotation of the object and the creation of the arrow into two separate issues.

To test the function:
1. create a freedraw object
2. target the object
3. write in the console _rotateFreeDraw();_
4. Move your mouse in the canvas

Once the changes are done, calling the canvas update function can work. Rather than moving the mouse to the changes.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -7010,3 +7010,33 @@ function sortLayer(){
 function diagramCompare(a,b){
     return (a.isLayerLocked === b.isLayerLocked)? 0 : a.isLayerLocked? -1 : 1;
 }
+
+//---------------------------------------------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


a19hammi:
Outdated pull-request

PR #9808, opened 2021-03-31
Replaced diagram drawing system.

a19hammi:


a19steis:
Code needs to be further reviewed.

PR #9814, opened 2021-04-01
Fixed overlapping of menu bars.

a19chrel:


a19aleda (approved):
Seems to do the trick, nice fix :)

Issue #9811, opened 2021-04-01
Disable text selection when moving an entity


a19nikbo:
Currently when moving an entity, dragging it over other entities highlights their text elements.

a19nikbo:
Fixed

PR #9815, opened 2021-04-01
Disabled text selection #9811

a19steis:
Disabled text selection when moving entities.


a19matma (approved):
The text no longer gets selected and other functionality doesn't seem to be affected in any way by the changes.

PR #9816, opened 2021-04-01
entities overlapping header

a19phith:
Fixed by adding position and z index to diagram-header


a19chrel (changes requested):
The problem is still existing, the diagram editor menu bar (top) has an transperent background. This make the diagram-elements appear behind the text.

**Exempel:**
bild

This problem could be fixed with an background color for the diagram editor menu.

b19jenda (approved):
Entities does not overlap the menu bar anymore. Looks good

PR #9817, opened 2021-04-01
Fix saving new position of entities

a19aleda:


b19jenda (approved):
Everything seems good, entity is working correct.

Issue #9791, opened 2021-03-28
Diagram line overlay layer is above entities not below


HGustavs:
We need to use css to move the overlay layer containing svg lines to a level below the entities. Currently the lines are drawn on top of the entities.

PR #9820, opened 2021-04-01
Group3 2021 #9791 - Fixed diagram line overlay layer

a19vikth:
Added z index: -10 to the line overlay layer.


b19phier (approved):
see no problem, z-index: -10 should fix the diagram line overlay layer

Issue #9829, opened 2021-04-02
Fix proper cursor type for buttons in diagram


a19nikbo:
When hovering over buttons in the new core the cursor stays as the default cursor. Preferably it should be switched to a pointer to make it clearer that an item is clickable.

![cursor_showcase](https://user-images.githubusercontent.com/81541557/113395119-a73cef80-9399-11eb-8673-211284489f01.png)

a19nikbo:
Successfully merged

PR #9839, opened 2021-04-06
Group3 2021 #9829 - Fixed Proper Cursor Type

a19steis:
Fixed a proper cursor type for Zoom in/out buttons in the options panel and a pointer cursor when hovering "Options".

Also disabled user-select for the "Zoom" title in the box.


a19phith (approved):
Looks good!

PR #9841, opened 2021-04-06
Changed bulbs to always be grey.

b19wilfo:
Simplified CSS by removing different stoplight colors in style.css.  Changed If-statements in sectioned.js to only show grey light regardless of submission status.


a19kimgr (approved):
Det ser bra ut, testade ändringarna och de funkade

a19klafr (approved):
Lights does not change color from grey after submission as requested

a19isada:
Unsure if the parts of the code should be saved for later which was realised after the pull request had been made.

Issue #9830, opened 2021-04-02
Diagram flips out when a node is on top of another


a19nikbo:
More specifically, if you drag a relationship/attribute node behind another node two things can happen as far as I know:
    - If you try to move any of the "stacked" nodes, you will grab it but it will have a pretty significant offset from the 
      cursor
    - If you move another node, both of the "stacked" nodes will change positions
    
The first one doesn't seem to happen for entities for some reason, but the second one does. The issue also only occurs if you move a node with a lower z-index "behind" another node, not if you move one in front of another.

![diagram_showcase](https://user-images.githubusercontent.com/81541557/113397441-562efa80-939d-11eb-9f97-4b23b44f5827.gif)

a19hammi:
Shouldn't this issue be closed?

a19nikbo:
Issue fixed

PR #9843, opened 2021-04-06
Group3 2021 #9830

a19hammi:
Solved ISSUE#9830.

Other features were also introduced which were required in order to solve the bug without creating something that would break in the future. I also added some additional comments were I saw fit.

List of changes:
* Objects will no longer freak out when placed on top of each other
* Selection of several objects at once
* De-selecting options
* Selected objects will change in colour (beta feature, should be fixed within its own issue)
* Introduced global variable for indicating when the mouse has been dragged more than a set amount from start X/Y

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -36,6 +36,8 @@ var elements=[];
 
 // Currently clicked object list
 var context=[];
+var deltaExceeded = false;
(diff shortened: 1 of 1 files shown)


a19phith (approved):
Looks good, seems like every issue of this issue is solved

PR #9844, opened 2021-04-06
Updated diagram lines

a19matma:
Removed/fixed issue where lines didn't fully extend to the corresponding boxes.


b19phier (approved):
The issue where the lines didnt extend to the boxes seems to be fixed and the updated code is reasonable for the problem

PR #9852, opened 2021-04-06
Grouping of objects with keypress

a19phith:
Fixed by adding a key detection function  to the entities so that grouping is more easier


a19vikth (approved):
Code works. Holding ctrl key allows multiple elements to be selected.

PR #9861, opened 2021-04-07
added hash

a19gusba:
when clicking save a random number is generated that gets hashed and transformed from decimal to hex.


a18osken (changes requested):
Revert some of the refactoring/indentation so that we do not have too many changes

a18osken (approved):
Looks good

PR #9864, opened 2021-04-07
Removed color from bulbs

a19danla:
Remove the variable grade and replaced the if-else statements controlling the color of the lightbulb with a static greybulb.
Also remove the unused styles for the bulbs in style.css


b18phika (approved):

a18willi (approved):
Functionality works as described

Issue #9792, opened 2021-03-28
Center / Left alignment of text in diagram


HGustavs:
If text is longer than the space inside the symbol we currently have clipping of the text

The issue with this is that we clip both the beginning and the end of the text.

"Procurement" would read "curem" with the rest of the text clipped away.

We should use centering if the text is short so that "Owns" is centerd in the middle of the attribute or relationship symbol for example.

But if the left edge of the text is to the left of the left edge of the symbol the text should remain to the left.

something like place left edge of text half of the width of the text to the left of the center of the symbol, but if that resulting value is less than the left of the symbol we should assign the position of the left of the text to the left edge of the symbol.

The result is centering until the beginning of the text is occluded, and after that we align the left edge of the text to the left edge of the symbol.

The key line drawing function has a function for the estimation of text length.

HGustavs:
This can partially be done through code BUT I do think we need text measurement.

In cavas we have text measurement, but I am not aware of any way to do this when using SVG without some advanced trickery.

My approach is to make a reasonably competent measuring functionality.... that is mostly guessing correctly. It only works for widths relative to the text size and with this specific font.

https://github.com/HGustavs/DiagramEngine/commit/a5022cf

a19nikbo:
Fixed

PR #9871, opened 2021-04-07
Issue #9792 center/ left alignment

a18nadma:
Added if statement so center alignment becomes left alignment after a certain amount of characters.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -864,7 +864,7 @@ function setPos(id, x, y)
 //-------------------------------------------------------------------------------------------------
 
 // Generate all courses at appropriate zoom level
-function showdata()
+function showdata() // TODO : Refactoring??
 {
     var container = document.getElementById("container");
     var containerbox = container.getBoundingClientRect();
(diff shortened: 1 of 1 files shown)


a19frelu (changes requested):
* Does not work with different width of the elements.
* Only work on the entities, not relations or attributes.
* Use calculation which uses the pixels amount of characters/width of the letters > element width, instead of fixed maximum character value (const maxchar = 29;).

a19hammi:
Updated branch to latest weekly build

a19frelu (approved):
The changes of the code now makes the text in the elements behave as wished for. The code aligns the text in the center of the object as long as the element is wider then the text. 
If the text is wider then the edges of the element, the text will align to the left in the element to make it possible to read the beginning of the text in each element.

The text alignment is now also depending on how wide the element is wich also was requested.

And for last. The code is now implemented to act over all of the elements, not only the entities, but also the relations and attribute elements.

Good job!

a19hammi:
Re-opening under new branch

Issue #9804, opened 2021-03-31
Redirection : Url to student assignment


a18osken:
We need a short url to student assignment e.g. dugga.his.se/sh/2fc6e1 that will redirect to saved assignment submission

part of bigger issue #9798

a19kevkh:
Created a unique password generator for 7 characters.
Also created an URL shortener based on the "real" URL. If URL is in the database, that means a shortened URL already exists.

a16erilj:
We have created a solution that relies on a database that have yet been implemented. This database (passwordURL) are currently under development in another issue #9790.

PR #9879, opened 2021-04-07
G1 2021 w14 #9804

B18mohah:


B18mohah:
we used skeleton metod to solve this issue, our code relies on databas that isnt implemented.

a18osken:
Going to temporarily close this to be reopened when the required db elements have been implemented and your solution works with it.

Issue #9818, opened 2021-04-01
Be able to create new elements and add them to the canvas


a19aleda:
In the current version there is no way to create new elements and add them to the canvas. This issue should maybe be broken down in  smaller sections.

HGustavs:
A simpler version of last years version plz.

I suggest a similar or identical menu to that of the old system, sort of like the tool bar in illustrator for example

<img width="36" alt="toolbar" src="https://user-images.githubusercontent.com/6978801/113291579-84f09680-92f3-11eb-95bc-d8d288b799ed.png">

Then we are in "mouse mode" or "entity mode" "relation mode" "attribute mode" ...

Every click on the canvas should create one object with the default properties for the relevant mode.

HGustavs:
In addition to this we need a "line mode" which creates connectors between objects

In that mode I suggest one of the following: 

* starting a "rubber band" line from the center of the object below the mouse when the mouse is pressed down, and if another object is below the mouse when we release the mouse button we create a link to that object.
* we click on an object, this starts the rubber banding from the center of the object below the mouse and when we click again on another object the line object is created

Unclear which of the two feels most natural.

a19nikbo:
We can now add entities, attributes and relations. Edges (lines) are part of a separate issue.

PR #9885, opened 2021-04-08
Group3 2021 #9818

a19aleda:
There is now a way to create elements on the canvas based on with mode you're in. The buttons still needs to be styled but this could be another issue to be created.


a19frelu (approved):
* All new elements has the same name/value (Entity) when implemented in the canvas. Future fix.

PR #9886, opened 2021-04-08
Creates box-markers around selected elements

b19jenda:
when selecting elements in diagram, box-marker appears around selected elements.


a19steis (commented):
Tested and looks fine, the box-marker around the selected items seems to work as intended.

a19steis (approved):

PR #9902, opened 2021-04-08
update showdugga.php

a19gusba:
added inline css to remove resize of textarea


a18osken (approved):
looks good to me

a19rebsa (approved):
Looks good, works fine

PR #9906, opened 2021-04-08
Mouse pointer

a19steis:
Fixed mouse pointer when hovering over mouse mode buttons, also for the "+" sign.

### DIFF
--- a/DuggaSys/diagram.php
+++ b/DuggaSys/diagram.php
@@ -464,14 +464,14 @@
             <div>
                 <fieldset>
                     <legend>Zoom</legend>
-                    <input class="zoomButtons" type="button" value="Zoom in" onclick='zoomin();' />
-                    <input class="zoomButtons" type="button" value="Zoom out" onclick='zoomout();' />									
+                    <input class="paneButtons" type="button" value="Zoom in" onclick='zoomin();' />
+                    <input class="paneButtons" type="button" value="Zoom out" onclick='zoomout();' />									
                 </fieldset>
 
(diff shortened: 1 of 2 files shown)


DuggaSys/diagram.php
```diff
@@ -464,14 +464,14 @@
...
                 </fieldset>
 
-                <input type="button" value="Mouse Selection" onclick='setMouseMode(0);' />
-                <input type="button" value="Entity" onclick='setMouseMode(1);' />
-                <input type="button" value="Relation" onclick='setMouseMode(2);' />
-                <input type="button" value="Attribute" onclick='setMouseMode(3);' />
+                <input class="paneButtons" value="Mouse Selection" onclick='setMouseMode(0);' />
+                <input class="paneButtons" value="Entity" onclick='setMouseMode(1);' />
+                <input class="paneButtons" value="Relation" onclick='setMouseMode(2);' />
+                <input class="paneButtons" value="Attribute" onclick='setMouseMode(3);' />
```
a19chrel, line 474: This change removes the input-type "button" that makes the input go to the default type value of "text". Therefore it is possible to edit the field.

a19chrel (changes requested):
While adding the class to the input fields, the type was removed and makes the input-fields to be text-fields and not buttons. This makes the fields editable.

DuggaSys/diagram.php
a19steis: Correct, added button types.

a19chrel (approved):
LGTM

PR #9907, opened 2021-04-08
Fix: Correct name on new elements - #9889

### LINKED ISSUE #9889: All newly added elements have the same name.
Change each element name to the correct name when adding new elements to the canvas. For entities the name should remain "Entity", but attributes should be named "Attribute" and relations should be named "Relation" when first added.

a19matma:
Fixed an issue where new elements had the same name. Each element is now created with the correct name respectively.
fixes #9889


a19karwa (approved):
Well done! I tested it and it seems to be working fine.

PR #9919, opened 2021-04-09
added hash to url

a19gusba:
removed random url and added the hash instead, also removed the "Hash:" from receipt since it already is displayed in the url.

Also changed the hash from base16 to base64 (youtube style)


e15marol (approved):
Looks good, works on chrome, firefox and edge browsers.

a19arvka (approved):
Decimal to Base64 function is very well made. 
All good!

Issue #9795, opened 2021-03-29
No downloads when viewing submissions (file submission dugga)


HGustavs:
Teachers should still see the normal view but ..
 students should only see the file names and/or file list from zip files.

Other types of duggas are not affected by this change, and should function as always

Issue #9782, opened 2021-02-24
File lists for uploaded zip-files


HGustavs:
If we upload a zip-file, the short-linked preview should show the name of the file and the link but also a directory listing of the zip file.

This way a teacher can see if for example the correct type of files have been uploaded without having to decompress zip file.

a13zanal:
We're working on understanding how zip reading works in JavaScript and how to implement this to dugga.js and resulted.js. We have also asked a question to customer for more clarification _"Should the zip file be able to preview directly in lenasys before downloading? You can do this with Zip.reader but this seems to be a separate library that needs to be implemented. Is this what you are looking for in this issue?"_.

b19ponka:
Since the database structure is not yet defined, going from student ID to hash, we decided to put this issue on hold

PR #9920, opened 2021-04-09
G1-2021-w15-#9795

a19arvka:
Students can't download submitted assignments. They can only see file name, date and file extension. 
Teachers (Super users) have access to the download functionality, either by clicking the pdf link or by pressing the download button. A boolean flag determines whether the download should be enabled or not, depending on the active user if they are teacher or student. 
However, Zip submissions are unmodified. This is a seperately issue #9782.


a17alfja (approved):

a19gusba (approved):
works! 

students can open the pdf and then download it, not sure if that should be another issue or not

PR #9921, opened 2021-04-09
Save Assignment password and hash to db

a19frebo:
Added columns hash and password to the table userAnswer. The password and hash are now generated and stored in the table for new saved dugga.


a19kevkh:
Testing guide:

1. Run the SQl script (alterTable) that is located in /shared/sql using mysql workbench.
2. This will add the columns hash and password into the userAnswer table. 
3. You can view the results by going into XAMPP and clicking on the admin button for mysql where you can take a look at the table.
4. Ready to be tested.

a19frebo:
Found a bug where if you just go in to a dugga a hash and password is saved to the DB. The hash and password should not be saved to the DB until the user clicks on "save".

a19gusba (approved):

a17alfja (approved):

PR #9923, opened 2021-04-09
Added functionality to create lines between entities

a19aleda:
Added a button in the side bar menu to toggle into draw lines mode which enables you to create lines from an entity to another.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -58,6 +58,7 @@ const mouseModes = {
     ENTITY: 1,
     RELATION: 2,
     ATTRIBUTE: 3,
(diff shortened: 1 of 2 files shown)


a19matma (approved):
It is now possible to click a button, select an element, select another element, and the diagram will be updated with a line between them. It doesn't seem to break any older functionality when testing, however multiple lines from the same two elements might cause the lines to "jump" around when moving the whole diagram. This doesn't come across as a big problem since multiple lines between two elements are never or rarely used. 
In whole, I would argue that the added code has solved the main issue of added functionality in a somewhat proper way.
Request approved.

Issue #9900, opened 2021-04-08
Multiple elements can be selected when holding ctrl on Windows, but not with cmd on MacOS


a19nikbo:
Expand the multiple-element-selection functionality to allow holding down the cmd button on Mac, instead of just taking ctrl into account.

HGustavs:
Since macs have both keys, perhaps we can support either key?

HGustavs:
There are still lots of modifier keys to take into account if we want to be exotic.

https://www.rcsri.org/collection/symbolics-keyboards/

a19nikbo:
Fixed

PR #9929, opened 2021-04-09
Added the "cmd" as alternative key for control.

a19frelu:
Added the same functionality to the "cmd" key for Mac users as the "ctrl" key have for the Windows users. Issue #9900


a19phith (approved):
Ser bra ut! de går att köra lenasys med mac

PR #9930, opened 2021-04-09
Update dugga.js to only show whitelight

a19danla:
Replaced the if-statements to alwats show a static white lightbulb in the navbar and set the message to "Dugga".


b19alest (approved):
What I can see it works as intended.

a19anjst (commented):
Everything looks as it should. The text dugga shows up with a white lightbuld at the top of the screen. Good job!

a19anjst (approved):

PR #9931, opened 2021-04-09
update dugga.js

a19gusba:
moved the preview of pdf to only be visible for teachers


a18osken (changes requested):
Just a quick addition to check for more files to remove the preview for, according to 
> No previews of assignments/filers for students.
> 
> Teachers still have full preview and download.
If I have interpreted HGustav's intentions correctly we should only be able to preview the filetree of a zip archive

a18osken (approved):
Yep looks good

PR #9942, opened 2021-04-09
Removing access category and group assignment from access editor

a19anjst:
Commented out the access column in the function returnedAccess in the accessed.js file. Created a text file called accessed_removedColumn.txt which describes what columns aren't used anymore by the accessed.js file.


a18willi (approved):
Everything looks good.

PR #9947, opened 2021-04-12
Added nodes to selected entity

### LINKED ISSUE #9917: Add nodes to the edges of selection box in diagram
To break it apart into smaller issues this issue is not about the resizing functionality itself, but rather a preparative addition for when we add it later. We might want to add small nodes at the edges of the selection box that will later be used for resizing. I've made a mockup of how it may look, but the nodes might benefit from being slightly bigger.

mockup2

a19chrel:
Added nodes to the selected marking, this resolves #9917 .

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -114,3 +114,6 @@
     position: relative;
     background-color: #FFF;
 }
+.selectedNode {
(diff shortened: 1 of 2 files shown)


b19phier (approved):
Ser fint ut, fixar problemet

Issue #9916, opened 2021-04-09
Create a "Copy hash to clipboard" button for hand-in receipt 


a18osken:
For some Quality of Life functionality in the receipt dialog we could add a button to copy the generated hash to the clipboard so that the user easily can paste it into Canvas without an extra whitespace or missing a character. 

Something like this:
![Copytoclipboardmock](https://user-images.githubusercontent.com/62884391/114145805-1c1ea500-9917-11eb-8241-c2623fb24f01.PNG)

PR #9950, opened 2021-04-12
G1 2021 #9916

a17alfja:


a17alfja:
we need a new pull request as we copied the wrong variable to clipboard (URL instead of Hash)

PR #9951, opened 2021-04-12
G1 2021 #9916

a17alfja:


a19arvka (approved):
Good. Nice implementation and structure inside method.

Issue #9932, opened 2021-04-09
Include KeyListerner for Zooming in or out


a19chrel:
To make the diagram easier to operate i suggest that scrollup and scrolldown should increase and decrease the value of the zoom. This make it possible to change the zoom value whitout accessing the options panel.

HGustavs:
Another key binding to add is ctrl + and ctrl - which should also zoom in and out respectively.

I suggest through a function zoom(direction) which can be called in all places, for example ctrl+ would call zoom(1) and ctrl - would call zoom(-1) this way all zooming is handled in one specific location

a19matma:
Wouldn't ctrl+ and ctrl- interfere with the option to zoom in and out of the whole browser? Or is this option not needed?
I've however fixed so that zooming in the diagram with the mousewheel now works as intended.

a19nikbo:
Fixed

PR #9956, opened 2021-04-12
Fix - zoom with mousewheel and ctrl-keys, issue #9932

a19matma:
Fixed so it's now possible to zoom in the diagram with the mouse-wheel and with ctrl+ or ctrl-.
Ctrl+ and ctrl- however interferes with the browser functionality with the same commands. I included it anyway since it was specifically asked to be implemented. A bigger function for handling all zoom-events have not been added as requested, since we already have two functions that handle this (zoomin and zoomout) and changing it could cause unnecessary problems.


a19vikth (approved):
Looks good. Zoom with mousewheel and ctrl -/+ keys works.

PR #9957, opened 2021-04-12
Update navheader.php

a19samke:
If the hostname is == to canvas.his.se this will not display the header, since we dont want it to show when embedded into canvas.


a19anjst (approved):
Looks good, everything works as it should, good job!

b19alest (approved):

PR #9962, opened 2021-04-12
Updated export popup window

b19wilfo:
Updated export popup window to stay fixed in one position while scrolling.

Coauthor- @a19danla


a18danli (approved):
Works for me!

a19samke (approved):
The layout remains in the same place, and the test code works

A19osamn (commented):
It works perfect !

PR #9964, opened 2021-04-12
updated dugga.js and  showdugga.php

a17alfja:
created function in dugga.js that enablet setting the value of a variable that represents the dugga title, this function is called in showdugga.php since this file has access to the dugga title


a19rebsa (approved):
Looks good and works as it should

a13zanal (approved):
Dugga title works fine

PR #9967, opened 2021-04-12
Cmd(Mac) + backspace = delete element

a19frelu:
"cmd" key combined with backspace on Mac is now the same as the delete button on Windows.


a19phith (approved):
Everything seems to be in order!

Issue #9945, opened 2021-04-09
Pressing escape should cancel most mouse modes in diagrams


HGustavs:
If the user is mid drag-scroll or mid-move or mid-entity-line (or most other mouse modes) and the user changes his/her mid, as one does, pushing the escape key should act as a general canceling of that mouse mode without changing the objects/scroll position/etc

PR #9968, opened 2021-04-12
Group3 2021 w15 #9945

a19aleda:
Added the cancel functionality by pressing the escape button


a19matma (approved):
Seems good. The functionality requested is implemented and works and seem to work as intended. Good job!

PR #9970, opened 2021-04-12
Update dugga.js

a17alfja:
Changed two symbols that the hash could generate (+ to -, / to _)

### DIFF
--- a/Shared/dugga.js
+++ b/Shared/dugga.js
@@ -707,7 +707,7 @@ function convertDecimalToBase64(value) {
 	}
   }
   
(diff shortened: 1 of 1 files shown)


a18osken (changes requested):
It seems keeping with the base64 alphabet is important(?), I would say we conserve the base64 alphabet and parse the generated ASCII string for `+` and `/` and switch them to `-` and `_`.

a18osken (commented):
lgtm

a18osken (approved):

Issue #9971, opened 2021-04-12
Options menu covering the ruler


a19hammi:
As seen in the picture below, our options bar currently covers the underlying ruler.
Maybe we should apply a size to the options bar relative to the canvas height?

![image](https://user-images.githubusercontent.com/81571224/114390927-2353e600-9b97-11eb-8b88-3636713f34ec.png)

HGustavs:
Also 

For all duggas, when embedded as an iframe into canvas or if the user clicks a  "full screen" button the purple top menu should also disappear.

PR #9976, opened 2021-04-12
Options menu covers the ruler

a19phith:
#9971 fixed by adding some value to top in diagram.css


a19frelu (approved):
Looking good. The option menu no longer covers the ruler.

Issue #9963, opened 2021-04-12
Keyboard shortcuts for diagrams


a19hammi:
We should introduce some form of keyboard shortcuts, for example using B for box-select and such.
We should probably take this through a design phase to decide what shortcuts for what buttons.

a19nikbo:
Fixed

PR #9981, opened 2021-04-12
Group3 2021 w15 #9963

a19steis:
These shortcuts are just a placeholder until we decide on the actual design, but the functionality should be there.


a19frelu (approved):
The shortcuts works as they is supposed to do. Looking good!

Issue #9979, opened 2021-04-12
Backspace-deletion functionality interferes with removal of characters in input field


a19nikbo:
If you select an element and press backspace it deletes the element. This is how we want it. However, if we have the input field for changing an element's name selected this functionality should be disabled. Otherwise, deleting characters is impossible.

![image](https://user-images.githubusercontent.com/81541557/114395530-c65b2e80-9b9c-11eb-87f0-8a6f0579916d.png)

HGustavs:
True!

a19nikbo:
Fixed

PR #9984, opened 2021-04-12
Bugfix when using backspace #9979

a19matma:
Fixed the issue when using backspace in the textbox would cause the whole element to be unintentionally deleted


a19steis (approved):
Good fix, when the input field is selected backspace no longer removes the element.

PR #9990, opened 2021-04-13
Update showDugga.php

a13zanal:
Removed the login restriction to upload and save a dugga, now is visible but it is not possible to save until the hash has been resolved.

Co-Authored-By: b19ponka <62886839+b19ponka@users.noreply.github.com>


a19arvka (approved):
With removed conditions in the if statement: the assignment can be saved without being logged in. Good job.

PR #9997, opened 2021-04-13
Bug fix: Deselecting a non-selected element while holding alt

a19hammi:
Very simple solution. This should prevent it from happening.


a19phith (approved):
Looks good!

Issue #9986, opened 2021-04-12
Diagram cursor modes as tool bar


HGustavs:
I think there is a previous issue for creating an illustrator / photoshop -style drawing-tool-bar in addition to the side-bar for drawing on the diagram.

Currently we have buttons in the property side-bar.

<img width="289" alt="Screenshot 2021-04-12 at 17 44 36" src="https://user-images.githubusercontent.com/6978801/114422836-ea2b6e00-9bb6-11eb-9158-bdc92ecdd84b.png">

<img width="41" alt="Screenshot 2021-04-12 at 17 46 47" src="https://user-images.githubusercontent.com/6978801/114423611-a1c08000-9bb7-11eb-8a1c-9c83669ae2da.png">

Perhaps we can move the zooming functionality there too... more commonly used features in the draw menu and object properties or more rarely used functionality in the sidebar.

Also horizontal / vertical view of toolbar using flexbox would be a good idea.

HGustavs:
First arrow is select / move and second arrow is box-select in Illustrator, is this a good way to use the icons?

a19aleda:
Perhaps the box-select icon could be a dotted/dashed square

https://static.thenounproject.com/png/1122247-200.png

a19nikbo:
Fixed

PR #9999, opened 2021-04-13
Added icons and change current buttons into the icons

a19aleda:
Added svg icons and changed the current buttons into svg icons


a19chrel (changes requested):
Currently does this not solves issue #9986. The buttons in options-panel has been changed to images instead, but should also be moved to an new panel to the left of the diagram.

a19karwa:
Closed pull request since it does not solve the linked issue. New pull request should be made to solve that issue.

Issue #9953, opened 2021-04-12
Diagram: Resize selected entity


a19chrel:
Currently when you have selected an entity, nodes will be shown in the corners of the selected marker. As shown in the picture below.

![bild](https://user-images.githubusercontent.com/62853826/114361349-533ec180-9b76-11eb-93f9-424417fe3c8b.png)

The functionallity to resize the entity when moving one of the nodes should be added.

a19nikbo:
Fixed

PR #10000, opened 2021-04-13
Group3 2021 w15 #9953

b19phier:
enable resizing of entities


a19aleda:
Currently you can resize the elements to a width of 0px, this makes them dissapear and makes it impossible to select the elements again. Maybe there should be a minimum width of maybe 10-20px?

a19aleda (changes requested):
Currently you can resize the elements to a width of 0px, this makes them dissapear and makes it impossible to select the elements again. Maybe there should be a minimum width of maybe 10-20px?

DuggaSys/diagram.js
```diff
@@ -405,6 +417,8 @@ function mup(event)
                 }
             }
             break;
+        case pointerStates.CLICKED_NODE:
```
a19aleda, line 420: Is this case needed?

a19aleda (approved):

Issue #9946, opened 2021-04-09
Attribute properties


HGustavs:
The distinct states of er attributes are mutually exclusive, an attribute can not be both a multiple value and a key at the same time for instance.

I think we can use a dropdown instead of three individual boooleans, this way we can guarantee that we dont have multiple computed attributes or some other unsupported combination.

We can use the defaults to store the list of options. The first of which we can use as the default value.

https://github.com/HGustavs/DiagramEngine/commit/b8e8f8c

Dropdowns for controlling properties can perhaps be used in other locations for other mutually exclusive properties?

a19nikbo:
Fixed

PR #10001, opened 2021-04-13
Issue-#9946

b19jenda:
Removed booleans from ER elements and added states. Now we will use state: weak instead of isWeak: true. 

Added drop down in option menu and with the drop down you can change states of ER elements.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1046,6 +1046,8 @@ function generateContextProperties()
     {
         var element = context[0];
         var state=["none","multiple","key","computed"];
+        var state2=["none","weak"];
+        var state3=["none","identifying"];
         //ID MUST START WITH "elementProperty_"!!!!!1111!!!!!1111 
         for (const property in element) {
             switch (property.toLowerCase()) {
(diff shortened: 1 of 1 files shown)


DuggaSys/diagram.js
```diff
@@ -119,17 +119,21 @@ var defaults = {
     defaultERrelation: { kind: "ERRelation", fill: "White", Stroke: "Black", width: 60, height: 60 },
     defaultERattr: { kind: "ERAttr", fill: "White", Stroke: "Black", width: 90, height: 45 }
 }
+// States used for ER-elements 
+var attrState=["normal","multiple","key","computed"];
+var entityState=["normal","weak"];
+var relationState=["normal","weak"];
```
a19hammi, line 125: Please use enums instead of strings to prevent accidental misspelling further down in code in the future.

DuggaSys/diagram.js
```diff
@@ -1137,19 +1149,41 @@ function generateContextProperties()
...
+            str += '</select>';
+        }
+        else if(element.kind=="ERRelation"){
+            str += '<select id="propertySelect">';
+                for (i = 0; i < relationState.length; i++) {
+                    str += '<option value='+relationState[i]+'>'+ relationState[i] +'</option>';
+                }
+            str += '</select>';
+        }
+        str+=`<br><br><input type="submit" value="Save" onclick="changeState();saveProperties()">`;
```
a19hammi, line 1186: A lot of copy-pasted code here. Use a single for-loop and apply the value on the string depending on object type instead.

a19hammi (changes requested):

DuggaSys/diagram.js
a19hammi: You can store the value before performing the loop and then apply the value like so:
     
     var value;
     if (element.kind == foo)
         value = bar[i]
         
Then use:

     str += '<option value=' + value + '>' + relationState[i] +'</option>';

a19hammi (approved):
LGTM!

PR #10002, opened 2021-04-13
Update fonts in sectioned Issue#9413

a18willi:
Changed fonts of the boxes "Start date" and "End date" to match the remaining boxes in sectioned.


a19danla (approved):
The font is set to Robot and it works as intended

b19wilfo (approved):
Tested it and it works, the start and end dates now have the same appearance as the rest of the text boxes.

Issue #10003, opened 2021-04-13
Conflict misssolved


a19karwa:
Accidentally removed some code from row 457,  that the programmer added while solving an issue, just add that code back so that the zoom function can work properly again...

Image of the code:
![image](https://user-images.githubusercontent.com/81557429/114547148-5ca35900-9c5e-11eb-934f-cafc2aa1a145.png)

a19nikbo:
Fixed

PR #10004, opened 2021-04-13
Fix - code added for mwheel zoom #10003

a19matma:
Added code that was accidently removed when this functionality was conflict resolved.


a19karwa (approved):
Code is now there and tested. The mouse-wheel zoom function seems to be working properly. Good job!

PR #10006, opened 2021-04-13
Allow + sign to be use in name

a19danla:
Removed urldecode from the getOP function, which purpose is to trun +sign into a blankspace.


b18phika (approved):

a18willi (approved):
The + sign shows up as described.

Issue #9998, opened 2021-04-13
Some form of feedback on current mouseMode


a19hammi:
I would love to see some form of user feedback when changing mouse mode like box-select or making edges.
Maybe changing the cursor, or showing a symbol next to the cursor? It is difficult to know if the user is in box-select when using the keyboard shortcut instead of the menu buttons.

a19nikbo:
Fixed

PR #10008, opened 2021-04-13
Changed cursor icon depending on hotkey selection #9998

a19steis:
Changed the cursor depending on which mouse mode the user is currently in.

I suggest we create an issue that implements custom made icons for the cursor for the user interaction. Mainly to make it more clear in which mode the user is currently in.


a19chrel (approved):
Looks good to me.

PR #10010, opened 2021-04-13
G2 2021 w15 issue#9832 - moved swimlanes to the top from the side

a19kimgr:


b19wilfo (approved):
Tested it and the swimlanes appear at the top as they should.

a19danla (approved):
Swimlanes appear at the top as the should.

Issue #10012, opened 2021-04-13
Change url redirection to courses so that it works with different document roots


a18osken:
Currently different local document roots can mess up the course redirection, we should update it so that it works for local environments with different document roots.

PR #10014, opened 2021-04-13
G1 2021 w15 #10012

a19simma:
Tested with different document roots.
But make sure it works if your root isn't: localhost/LenaSYS/


a18osken (approved):
different document roots works now.

PR #10016, opened 2021-04-14
Added icons and change current buttons into the icons

a19aleda:
Added svg icons and changed the current buttons into svg icons


a19steis (approved):
Looks good, the icons now visualize correct cursor modes.

PR #10019, opened 2021-04-14
Issue #9946 created drop down in option menu

b19jenda:
Added enums for states in ER elements. By clicking on an element you can change the state in the option menu. 

Use branch Group3-2021-W15-#9946


a19nikbo:
Conflict was resolved incorrectly. For more detailed changes, see https://github.com/HGustavs/LenaSYS/pull/10001

Issue #10011, opened 2021-04-13
Populate coursekeys table with the courses available on the dev branch of LenaSYS


a18osken:
It'd be nice to have a populated coursekeys table to query against.

I was thinking that we populate the table with the course names available for the development branch , found on the DuggaSys/courseed.php url. Along with a 3 letter abbreviation, Datorgrafik becoms dgr or something like that.

PR #10020, opened 2021-04-14
G1 2021 w15 #10011

a19taral:
updated database


a18osken (changes requested):
While it made me chuckle ^_^, we probably don't want ooga-booga to be a course key since it will match unrelated urls.

funny tho.

a18osken (approved):
looks good

PR #10025, opened 2021-04-14
G1 2021 10023

a19kevkh:
Assignement is now stored to database when save is clicked.

Co-authors: @B18mohah and @a16erilj


a19rebsa (approved):
It works fine

Issue #9973, opened 2021-04-12
Background grid on diagram


a19hammi:
It would be really nice to see a background grid under the elements that is properly aligned with the ruler.
Maybe this should be under a toggle as-well, or put that in another issue?

Something like this?

![image](https://user-images.githubusercontent.com/81571224/114391765-30250980-9b98-11eb-8aa5-888c3f4232cd.png)

HGustavs:
Dotted thin lines that are not visible if grid is not enabled.

a19nikbo:
Fixed

PR #10034, opened 2021-04-14
"Fixed" Enable grid function

a19phith:
#9973 
@a19frelu


a19vikth (commented):
The grid function works. But when the grid is enabled it increases the width of the page.
github screen2

a19vikth (changes requested):
Forgot to change the feedback type in the comment above

a19vikth (approved):
Looks good

Issue #10018, opened 2021-04-14
Resizeing the window after loading, makes ruler small.


a19chrel:
Currently if you open the webpage in a smaller window and resize the browser window to a bigger one. The ruler dosen't update to the new size of the browser window.

![bild](https://user-images.githubusercontent.com/62853826/114677057-376c2480-9d0a-11eb-9535-a502b94153f0.png)

a19nikbo:
Fixed

PR #10035, opened 2021-04-14
Fixed ruler resize - Issue #10018

a19steis:
Fixed the resizing issue with the ruler.


a19frelu (approved):
The ruler is resizing as expected when shifting the size of the window. 
Well done!

PR #10044, opened 2021-04-14
updated html in showdugga.php and css in style.css

a17alfja:
updated layout of receipt box by moving elements and styling


a19arvka (approved):
Nice solution. Good graphical user interface!
The div class "button-row" gets modified in the css but I noticed that the same class (before it got modified) was being used in testdugga.php. This could cause unwanted changes to testdugga.php. However, I don't think that specific section in testdugga.php gets called upon/used currently, so it might not be a problem for now.

a19rebsa (changes requested):
The buttons appear to be outside of the box when using a smaller screen (see picture). This needs to be fixed

dvlc

PR #10047, opened 2021-04-14
Update testdata.sql

a17alfja:
Removed duplicate insert


e15marol:
Looks good, there are no duplicate inserts that we could find.

b18jacha:
And we also tried a fresh install with zero errors.

e15marol (approved):
Looks good, there are no duplicate inserts that we could find.

b18jacha (approved):
Tried a fresh install with zero errors.

Issue #9995, opened 2021-04-13
Small fix: If-statement continously growing


a19hammi:
We need a better solution to the growing if-statement inside `exportElementDataToCSS`!
Currently looks like this `if (inContext && !clickedContainer && (notBoxSelection || clickedElement) && !clickedNode)`,
and who knows how long it will eventually grow if we keep going...

Me and @a19chrel have discussed it and believe that removing `deltaX` and `deltaY` from the `element.top` and `element.left` and introduce a new _specific_ delta that is _only_ for moving the element and not used in all the other modes.

This needs to be fixed **ASAP** before we start seeing an even bigger growth in the if-statement.

a19nikbo:
Fixed

PR #10052, opened 2021-04-15
Fixed if-statement

a19chrel:
Solves #9995, due to an already existing variable movingObject the solution was easy.


a19phith (approved):
Seems to solve the issue and keeps the functionallity

PR #10056, opened 2021-04-15
Simple solution to the issue

a19arvka:


a19frebo (changes requested):
Link should look like: http://localhost/LenaSYS/sh/?a=

DuggaSys should be removed from the URL

a19kevkh (changes requested):
Looks fine, but change the "http://localhost/LenaSYS/DuggaSys/sh/?a=" + hash" to "http://localhost/LenaSYS/sh/?a=" + hash".

a19kevkh (approved):

a19frebo (approved):
Looks good!

PR #10058, opened 2021-04-15
update dugga.js, showdugga.php, index.php

a19gusba:
sets the hash variable depending on url parameter


a13zanal (approved):
Works fine, gets the same hash and password when you press the save button.

b19ponka (approved):
Works great. As long as you use the provided URL, the hash stays the same

Issue #10026, opened 2021-04-14
Changing name on entities and attributes also change mouse-mode


a19chrel:
When you write in the text input to change a name of an attribute or entity, currently you also change the mouse mode. This should be disabled during input to an text-field.

https://user-images.githubusercontent.com/62853826/114687820-68e9ed80-9d14-11eb-8f96-889c59282802.mp4

a19nikbo:
Fixed

PR #10059, opened 2021-04-15
Added selection for checking if an input is active - ISSUE 10026

a19chrel:
Added a selection for checking active element for the key-listeners, if an input, textarea, select is active, none of the key-listeners is now active.

This solves #10026.


a19phith (approved):
All the key-shortcuts seems to de deactivated when typing in the text field. With other words the issue seems to be solved

PR #10065, opened 2021-04-15
Fixed weak entities

b19jenda:
Now weak entities will have double border as the other weak elements.


a19karwa (approved):
Seems to be working, good job!

PR #10066, opened 2021-04-15
Assignments localstorage

b18jacha:
Assignments without variants no longer saves to localstorage

Co-author
b18jacha
e15marol

### DIFF
--- a/DuggaSys/showDugga.php
+++ b/DuggaSys/showDugga.php
@@ -45,6 +45,7 @@
 	$readaccess=false;
 	$checklogin=false;
 	
+	$variantsize;
 	$variants=array();
 	$duggaid=getOPG('did');
 	$moment=getOPG('moment');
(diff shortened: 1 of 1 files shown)


a19frebo (changes requested):
Seems like duggas with only 1 variant is still stored in local storage. 

row 244, what is really checked here?

e15marol:
> Seems like duggas with only 1 variant is still stored in local storage.
> 
> row 244, what is really checked here?

For the first question, the variants still need to be stored if they are within the scope of the amount of assignments that exist. Meaning that assignments with variants are stored in local storage as we need to keep track of which dugga the user first gets when accessing the dugga page. It is only assignments that have variants, that are saved to local storage.

For the second question, the largest variant.quizID is checked against the dugga IDs, this is so we include all assignments that have a variant assigned.

a19frebo:
I thought that a dugga with 1 variant is a dugga with "no variants". Doesnt a dugga need a variant to even be displayed?

b18jacha:
I don't think a dugga has to have a variant but unsure. But if it doesn't have one it will not save to localstorage with value "UNK" as it did before.

a19rebsa:
This solution works fine, might create an issue later for if there is only one variant we might not want to store it in local storage.

a19frebo (approved):

a19kevkh (approved):

PR #10069, opened 2021-04-15
The ruler now live-updating

a19vikth:
The ruler is drawn while panning the container. Instead of when the panning has completed


a19karal (approved):
works as intended :)

PR #10074, opened 2021-04-15
Fixed Key Attribute Underline

a19steis:
Fixed a text underline when the attribute "key" is chosen to indicate that it is in fact, a key attribute.

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -203,5 +203,9 @@
     top: calc(50% - 4px);
     right: 0;
     cursor: col-resize;
+}
 
+.underline{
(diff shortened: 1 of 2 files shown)


a19karal (changes requested):
should be a complete line instead of what it is right now
_______  is how it should be, right now its like this _ _ _ _ _ _ _ _

also capital letters cant get an underline for some reason.

a19steis:
Tested multiple browsers and i dont see the problem.

image

Maybe you could show me a screenshot of the issue?

a19karal (approved):
works as intended :)

PR #10075, opened 2021-04-15
Added a default mouse cursor on load

a19aleda:
Since the default mouse mode is pointer we auto set the cursor to pointer at start


b19jenda (approved):

PR #10077, opened 2021-04-15
"Fixed" save the dropdown properties

a19phith:
Fixed by adding a new varible who represent the entity's state and then add a select attribute to the drop down list


a19karal (approved):
works as intended :)

Issue #10017, opened 2021-04-14
Save assignment without being logged in


a18osken:
We want to be able to save an assignment while not being logged in. 

Relates to issue #9910

Issue #9910, opened 2021-04-08
Make saved assignment relate to hash


a18osken:
We need to make it so that a saved assignment is fetched in the database through its hash instead of any log in.

I am not entirely sure of all the relations in the database but currently the `userAnswer` tables uses `uid` from the `user` table as a foreign key, which I think is what connects the answer to a specific user. This table will not be in use by any student as students no longer should have a user on LenaSys. 

This relation should probably be replaced to make it so that the answer is connected to a specific hash instead. 

Part of issue #9790

HGustavs:
I was imagining that it would be possible to just use the answer hash instead of the user id, and that that minor change would resolve the problem without any more far-reaching consequences.

a18osken:
Some clarification: We want to be able to save assignments without being logged in, essentially we should identify the hand-in with hash instead of uid.

PR #10083, opened 2021-04-15
G1 2021 #9910

a19kevkh:
Saved assigments are now related to hash. To save without being logged in, see issue #10017.

### DIFF
--- a/DuggaSys/showDugga.php
+++ b/DuggaSys/showDugga.php
@@ -35,6 +35,7 @@
 	$quizid=getOPG('did');
 	$deadline=getOPG('deadline');
 	$comments=getOPG('comments');
+	$hash=getOPG('hash');
 
 	$duggatitle="UNK";
 	$duggafile="UNK";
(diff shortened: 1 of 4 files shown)


a19simma (approved):
image

When you save an already hashed dugga, password shows as an html object. There should be some explicit message and not this. 

Also, maybe there should be a prompt to login when you save the dugga? Right now you get a hash generated and password so it looks like it worked. But the has is not saved in the database. There should be an error.

@a18osken New issues for this maybe? Unless the login check is going to be removed. But then remove this condition for saving a dugga?

a19taral (commented):
The hash is not saved in database (coursekeys) in case if the user is not logged in.

a19taral (approved):
The hash is not saved in database (coursekeys) in case if the user is not logged in.

a19kevkh:
> The hash is not saved in database (coursekeys) in case if the user is not logged in.

We are working on it on issue #10017.

a18osken:
> image
> 
> When you save an already hashed dugga, password shows as an html object. There should be some explicit error message and not this.
> 
> Also, maybe there should be a prompt to login when you save the dugga? Right now you get a hash generated and password so it looks like it worked. But the has is not saved in the database. There should be an error.
> 
> @a18osken @a19rebsa New issues for this maybe? Unless the login check is going to be removed. But then remove this condition for saving a dugga?

There are issues concerning these shortcomings which should fix the problems found here 🙏

a19rebsa (approved):
Looks good

PR #10086, opened 2021-04-15
G2 2021 w16 issue#8705

a19danla:
Made so that newly created items appears at a centered on the screen according to your viewpoint.


a18danli (approved):
Works for me, when I add new item it appears in the middle of my viewport.

a19samke (approved):
It works perfect for me !

Issue #9975, opened 2021-04-12
Select edges with mouse selection


b19phier:
One has to be able to select the edges that are created in the diagram with mouse selection to later be able to delete these edges.

HGustavs:
I have some code that I can donate.... because this is not as easy to determine 
as it seems

HGustavs:
this.linedist = function (xk,yk)
		{
				if(this.symbolkind==4){
						var x1=points[this.topLeft].x;
						var y1=points[this.topLeft].y;
						var x2=points[this.bottomRight].x;
						var y2=points[this.bottomRight].y;						
							
						var px = x2-x1;
						var py = y2-y1;
						var len = px*px + py*py;
						var u = ((xk - x1) * px + (yk - y1) * py) / len;						

						if(u > 1){
								u = 1;
						}else if(u < 0){
								u = 0;				
						}
						
						var x = x1 + u * px;
						var y = y1 + u * py;
						px = x - xk;
						py = y - yk;
						
						dst = px*px + py*py;

						return dst;

				}else{
						return -1;				
				}

		}

HGustavs:
From 2020 project allows us to click a certain distance away from the line, since the lines are perfectly thin objects it is otherwise mathematically difficult to click exactly on any line.

HGustavs:
Checking if a line crosses a selection box is also not super-easy

The most common way is to check the bounding box for containment and then check for line crossing using the normal line to line intersection

HGustavs:
<img width="658" alt="Screenshot 2021-04-12 at 17 12 56" src="https://user-images.githubusercontent.com/6978801/114417960-59529380-9bb2-11eb-8347-cabf29a6d699.png">

We have an intersection if either x1,y1 or x2,y2 is inside the box OR
if the line segment crosses either one of the 4 line segments that comprise the box

a19nikbo:
Fixed

PR #10097, opened 2021-04-15
Now possible to select lines #9975

a19matma:
Implemented functionality for clicking lines.
Note: Since the "double lines" are stored as two separate lines individually they are also treated as such when clicking on them. Future functionality could be added to make sure they share the same ID and thus treated as the "same" line.


a19matma:
Kom på att vissa variabler kunde deklarerats utanför loopen istället. Kan rätta till detta om det efterfrågas.

a19steis (approved):
The fix seems to be working intended, noticed that you could click the lines in any mouse modes.

Maybe a design decision could be made deciding in which mouse mode the line should be clickable, but this solves the immediate issue as of now. Great work!

PR #10103, opened 2021-04-16
Moving getContext

a19chrel:
Moved the getContext intro the drawElement function instead to set it through parameter to the function, because the function only uses one context.


b19jenda (approved):
Seems to work as intended

PR #10105, opened 2021-04-16
Removed unused code

a19danla:
Removed the if-else statements that were removed in issue#9863 but somehow never merged into the groups masterbranch.


b19wilfo (approved):
Testade att ta bort if-satserna och allting fungerade som det ska.

PR #10108, opened 2021-04-16
G2 2021 w16 issue#9783

a19kimgr:
Removed statiscsList that was on the left side. 
removed PHP, JS, and css code.


a18danli (approved):
Looks like it work!

a19samke (approved):
It did not work for me when I manually delete the necessary code, however when I download and run your file it works.

Issue #10101, opened 2021-04-16
Remove the functionality to send receipt to e-mail


a18osken:
You can currently enter an e-mail address to send your hand-in receipt to. To avoid storing personal data we should remove this functionality in its entirety.

PR #10109, opened 2021-04-16
G1 2021 #10101

a19frebo:
Removed the ability to send email after saving an assignment. Also removed the function from dugga.js. We also took the freedom to remove some old unnecessary outcommented code.


a19gusba (approved):
works

a17alfja (approved):
Works as intended and nice removal of old unnecessary code.

PR #10110, opened 2021-04-16
Update showDugga.php and dugga.js

a17alfja:
Fetch password to the corresponding hash in showDugga.js and use it in a function in dugga.js that changes the password value

As the hash is not currently present in the url, you can manually add it when testing.


a19frebo (approved):
Password is now correctly shown after entering an assignment through the hash.

PR #10113, opened 2021-04-16
Added maximun number of edges between elements

a19chrel:
Added a check for maximun number of edges between elements and check the following requirments:
```
- Only one edge can be drawn between two attributes
- Only one edge can be drawn between an entity and an attribute
- Only one edge can be drawn between a relationship and an attribute
- Only two edges can be drawn between an entity and a relationship
```


a19vikth (approved):
Works as intended

Issue #10039, opened 2021-04-14
revisit hash generation


a18osken:
Found a trend that new hashes always start with a 1, we should revisit the hash generation.

I was thinking we could try using the `atob()` and `btoa()` JS methods to get a base64 hash. The replacement can remain I think.

a13zanal:
We have found why hash always starts with 1 and come up with the solution by removing for (). But should check more about atob () and btoa () for a better solution.

b19ponka:
Found the cause for why new hashes always start with a 1, made an optional solution and will try to make a better solution including atob() and btoa() functions

PR #10118, opened 2021-04-16
G1 2021 #10039

b19ponka:
Updated generateHash()-function so that it does not always start with a “1”. As well as make every combination, of 6-8 character, for the hash possible


a13zanal:
Co-author

a19kevkh (approved):
Looks fine, but the chances of the hash being smaller than 8 characters is abysmal.

a19frebo (approved):
Seems to work fine, the hash is not starting with "1" everytime anymore

PR #10119, opened 2021-04-16
Added array to check if id already exists

b19jenda:


DuggaSys/diagram.js
```diff
@@ -89,20 +89,35 @@ var pointerState = pointerStates.DEFAULT;
...
 // makeRandomID - Random hex number
 //-------------------------------------------------------------------------------------------------
 
 function makeRandomID()
 {
     var str = "";
+    var checkID = true;
     var characters = 'ABCDEF0123456789';
     var charactersLength = characters.length;
+    while(checkID == true){
```
a19steis, line 103: While statement could be just "true" since the while loop runs until the return statement, so CheckID == true kind of make no sense here.

a19steis, line 103: Indentation for the whole while statement needs to be overlooked.

a19steis (changes requested):

a19steis:
Indentation for the whole while statement/function needs to be overlooked.

DuggaSys/diagram.js
a19steis: Indentation for the whole while statement/function needs to be overlooked.

a19steis (approved):
LGTM!

PR #10120, opened 2021-04-16
Added ghost elements when you are adding Entities ISSUE#10082

a19aleda:
Now when you are in entity creation mode, you have a ghost element following your cursor.


a19frelu (approved):
It does what is expected from the issue. The cursor is visible through the element thanks to the good level of opacity used. 

Looking good guys!

PR #10126, opened 2021-04-16
Solved error in diagram_IOHandler.js

a19steis:
Removed diagram_IOHandler.js from diagram.php. This removes the error when loading the page.


a19chrel (approved):
LGTM!

PR #10127, opened 2021-04-16
G1 2021 w16 #9858 rebase

### LINKED ISSUE #9858: Create dialog for log-in to assignment
A dialog for logging in to the assignment should be created as shown in the mock-up in issue #9790. This should appear when navigating to the generated url.

Part of #9790

a19simma:
Using a hash you now get a pop-up screen asking for the password. Functionality to check the password is not implement, the box is just closed and no content displayed. 

Closes #9858

### DIFF
--- a/DuggaSys/showDugga.php
+++ b/DuggaSys/showDugga.php
@@ -298,14 +298,14 @@ if($cid != "UNK") $_SESSION['courseid'] = $cid;
 				echo "</div>";
 			}
 
-			function hashPassword($password){
+			function hashPassword($password, $hash){
 				if($hash!='UNK'){
 				exit();
 				//Authentication Function
 				}
 			}
 			//Retrieved from 'password' input field
-			hashPassword($password);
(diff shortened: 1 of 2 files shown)


a17alfja (changes requested):
Works as intended but a error is thrown revolving the scope of the hash variable in hashPassword function (it is undefined). Can be solved by adding hash as a parameter.

a19gusba (changes requested):
The functionality works but the variable $hash on row 302 becomes undefined due to using global variable in a local scope if u understand php correctly (error: Uncaught SyntaxError: Invalid hexadecimal escape sequence). To solve this you can either declare hash as global inside the hashPassword function: "global $hash" or have hash as a parameter along with password, hashPassword($password, $hash);

a19simma:
Added parameter to function

a19gusba (approved):
works

a17alfja (approved):
No errors!

Issue #9753, opened 2020-05-27
Validation inside "Create user"-dialog is incomplete


b18alegu:
Inside the "Create user"-dialog the level of validation of different attributes differs vastly, and more work needs to be done.

The following attributes are properly validated:
- SSN (see #8935 for details)
- Email (see #9133 for details)

The following attributes might need more work:
- First name
- Last name
- PID

The following attributes have no validation whatsoever at the moment:
- CID
- Ny (???)
- Term

![Skärmavbild 2020-05-27 kl  14 15 59](https://user-images.githubusercontent.com/62876260/83018419-8f803680-a025-11ea-9a06-f3c1d5781ac1.png)
Example of SSN validation message. In this case the 29th of February doesn't exist since 1998 wasn't a leap year.

a20chrjo:
Approved and merged by @a20chrjo

PR #10128, opened 2021-04-16
Removed CID from "Add user"-form. Issue#10093

a18willi:
We've looked into the purpose of the CID and Term field in the "Add-user"-form. The Ny-field seemed to have already been removed in issue #9753. CID didn't seem to have a purpose so we've removed it along with it's validation and tooltip function. The Term-field as it currently stands is saved in the class-column in the User-table in the database. We have not modified the Term-field.

### DIFF
--- a/DuggaSys/accessed.js
+++ b/DuggaSys/accessed.js
@@ -429,34 +429,6 @@ function tooltipTerm()
 	}
 }
 
-//---------------------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


a19danla (approved):
CID and NY is removed and it seems to be working as intended.

b19wilfo (approved):
Tested the changes and it works.

PR #10134, opened 2021-04-19
Holding ctrl while clicking background fixed

a19steis:
The selected lines no longer deselects while holding down ctrl and clicking on the background.


a19frelu (approved):
Works as intended on Windows computers. Fine work right here! Well done.

PR #10139, opened 2021-04-19
Added an message promp and errors-messages to lines.

a19chrel:
Added an new function to display messages in the diagram, this includes success messages and error messages.


a19phith (approved):
tested the functionallity of this fix and it seems to be workning as intended

PR #10143, opened 2021-04-19
Issue #10135: Ghost busted from data

### LINKED ISSUE #10135: No sp00ks in data
The ghost elements should not be put in the data array as the data array is meant for already placed elements. They need to be broken out as it interferes with other functionality.

Who you gonna call? @a19hammi

a19hammi:
OK, on a bit more serious note. data[] should ONLY contain finished and constructed elements, and not data that is NOT supposed to exist!!!

fixes #10135


a19steis (approved):
Functionality works exactly as it did, LGTM!

PR #10152, opened 2021-04-19
G2 2021 w16 issue#9870

### LINKED ISSUE #9870: Section editor refactoring bug from project course 2020
The project course in 2020 broke the no refactoring rule in an incompatible manner.

Version 0.12 (before project 2020 version) works well evidenced by screenshot below

Screenshot 2021-04-07 at 12 22 04

With the same data the startup project 2021 version does not work

Screenshot 2021-04-07 at 12 22 18

It seems like a new incompatible column "tabs" was added to the database to store the existing tabs data in a new place.

This makes the application completely incompatible with existing installations (which is the whole purpose of LenaSYS) ...

Previously the tab count was stored in the gradesys column (since an element can not be both tabbed and have a grade at the same time). We must revert this change so that the tabs are once again read and stored in the gradesys column.

b18phika:
Updated sectionedservice.php to no longer use the "tabs" column in the table "listentries".
Fixes #9870

### DIFF
--- a/DuggaSys/sectionedservice.php
+++ b/DuggaSys/sectionedservice.php
@@ -554,7 +554,9 @@ if($gradesys=="UNK") $gradesys=0;
 		$entries=array();
 
 		if($cvisibility){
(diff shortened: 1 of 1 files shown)


a19kimgr (approved):
Tabs are showing and code is giving no errors, approved.

a19danla (approved):
Tabs are showing and everything is working

PR #10163, opened 2021-04-19
Removed SSN and PID from "Create user"-form. Issue#10149

a18willi:
Removed SSN and PID completely. Term is still used because the data entered in Term gets saved in the class-column in the user-table. We don't know if this has any functionality but we haven't removed it because  it may break something.


a19kimgr (approved):
PID and SSN are removed from create user tab. No errors on my side.

a18danli (approved):
I see no errors, looks like it worked.

PR #10165, opened 2021-04-19
Fixed zoom button placing

a19steis:
Inverted the zoom buttons and added some css.


a19phith (approved):
seems to be working as intended

PR #10170, opened 2021-04-19
Fixed: No error message for reaching line limit

a19chrel:
Moved an displayMessage to the correct place


b19phier (approved):
looks fine to me, error message is showing when reaching the line limit.

Issue #10142, opened 2021-04-19
Remove login-lock and examine score variable in generic_dugga_file_receive.js


a19rebsa:
In the template file generic_dugga_file_receive.js there is code for showing the login-lock when not logged in. This should be removed. Also the variable "score" in the saveClick() function should be examined to see if it is needed for saving the dugga, this could maybe otherwise be some type of personal information.

a19frebo:
Removed the login-lock.
Removed the variable "score" from both the global section and from the saveClick() function, since it was never used when saving the dugga.

PR #10171, opened 2021-04-19
G1 2021 #10142

a19kevkh:
Removed the login-lock.
Removed the variable "score" from both the global section and from the saveClick() function, since it was never used when saving the dugga.


a19arvka (approved):
Very nice, login-lock is not present. No conflicts found when testing against the score variable.

PR #10172, opened 2021-04-19
Grid is now enabled by default

a19aleda:
By calling enable grid in the onload function, the grid is now being drawn as default.


a19karal (approved):
works as intended :)

PR #10173, opened 2021-04-19
Checks if usergroups has the value null instead of "none" Issue#9752

b18phika:
Made small change to the if statement on line 260.  The if-statement now checks if usergroups has the value null instead of "none".

Co-author: @a18willi


a19danla (approved):

b19wilfo (approved):
Testade och det fungerar.

PR #10174, opened 2021-04-19
Fixed cmd key on MacOS when selecting lines 

a19steis:
Fixed the line selection while holding down cmd key on MacOS


a19frelu (approved):
Works just fine. Does not seem to interfere with other code functions.

Good job.

PR #10182, opened 2021-04-19
MOTD now shows for all users, not only superusers. Issue#8551

a18willi:
Set motd before if-statement so that's it's displayed for everybody not just student teachers and those with write-access

Co-Authored-By: @b18phika


a19kimgr (approved):
Works as intended, tested as both teacher and student.

a19samke (approved):
It works for me!

PR #10184, opened 2021-04-19
Fixed overlapping div in option-menu

b19jenda:
Added an ID to the div that was overlapping and then added styling


a18nadma (approved):
Approved :D

PR #10186, opened 2021-04-20
Added appropriate svg images

a19steis:
Replaced the button types with custom made svg images that should better represent the functionality.


a18nadma (approved):
Looking good ;D

PR #10187, opened 2021-04-20
Added active class for modes in toolbar

a19chrel:
Added active class for the modes in the toolbar, this class is removed when a mode is deselected and added to the element if the mode is currently in use.


a19phith (approved):
The tool-bar icons seems to work as it was intended to work!

Issue #10144, opened 2021-04-19
Remove login-lock and examine score variable in daily-minutes.js


a19rebsa:
In the template file daily-minutes.js there is code for showing the login-lock when not logged in. This should be removed. Also the variable "score" in the saveClick() function should be examined to see if it is needed for saving the dugga, this could maybe otherwise be some type of personal information.

a19kevkh:
Removed login-lock when not logged in and also removed the variable score in the saveClick() function.

a19frebo:
When we tested this we saw that the dugga is not working properly at all even before our changes. This dugga template does not seem to be complete. For example the variable elapsedTime is not defined.

PR #10189, opened 2021-04-20
Deleted login-lock and score, issue #10144

a19frebo:
Deleted login-lock and score from daily-minutes.js


a13zanal (approved):
Can't find any errors, works fine!

b19ponka (approved):
Approved. Maybe consider also removing timeUsed = Timer.score; stepsUsed = ClickCounter.score;

Issue #10145, opened 2021-04-19
Remove login-lock and examine score variable in group-assignment.js


a19rebsa:
In the template file group-assignment.js there is code for showing the login-lock when not logged in. This should be removed. Also the variable "score" in the saveClick() function should be examined to see if it is needed for saving the dugga, this could maybe otherwise be some type of personal information.

HGustavs:
We need to do some proper thinking about how group hand-ins should work

HGustavs:
Like for ER diagrams, I am thinking we should be able to share the hash and password between students, and that way more than one student can view the same assignment at the same time. But we need synchronization to support updates in the view of one student when another student changes something.

a18osken:
Once the login functionality for individual hand-ins and the saving of file submissions are solved we will delve further into group assignments and see what we can do. Would the syncing between students be a function of the diagram core or the dugga page? We can most likely implement a group value for the hash and password so that several students can share it.

PR #10192, opened 2021-04-20
Removed login-lock and score #10145

a19kevkh:
Removed login-lock and score from group-Assignment.js


a13zanal (approved):
Can't find any errors, works fine!

b19ponka (approved):
Looks good. Consider removing var elapsedTime = 0; timeUsed = Timer.score; stepsUsed = ClickCounter.score;

Issue #10091, opened 2021-04-15
Migrant variant functionality to showduggaService


a18osken:
We want to load the variant in showduggaService and communicate through ajax calls

PR #10195, opened 2021-04-20
G1 2021 w16 #10091 

e15marol:
Migrated variant functionality from showdugga to showduggaservice and removed some old functionality where dugga was saved to the database when visited.


Shared/dugga.js
```diff
@@ -565,8 +557,8 @@ function randomPassword()
 //----------------------------------------------------------------------------------
 
 function createUrl(hash) {
-	var realUrl = window.location + "?a=" + hash;
-	var localhostUrl="http://localhost/LenaSYS/sh/?a=" + hash;
+	var realUrl = window.location + "&hash=" + hash;
+	var localhostUrl="http://localhost/LenaSYS/DuggaSys/sh/?hash=" + hash;
```
a16erilj, line 561: Possible that these changes are wrong since sh/index.php does not currently get the value trough the parameter hash but rather the parameter a. So if this change is introduced then it's possible that index.php would work correctly in it's current state when redirected to by dugga.js.

e15marol: Anything related to hashes in in this pull request have been added by mistake somehow, they need to be removed/discarded before a merge is done.

a19gusba (approved):

a17alfja (approved):
works

PR #10198, opened 2021-04-20
Added background color and shadow to the ruler

b19jenda:


a18nadma (approved):
Code doesn't disrupt anything and the ruler sticks out, well done!

PR #10199, opened 2021-04-20
Swimlanes now hide after clicking the arrow once. Issue#10185

a18willi:
Updated style.css so that the correct open/close-button for the swimlanes in the course-view is showing by default.


b19wilfo (approved):
Tested and it works without issue.

a19danla (approved):
Looking good.

PR #10203, opened 2021-04-20
Update dugga.js and generic_dugga_file_receive.js

a17alfja:
moved generation of hash from when save button is clicked to when the dugga is opened. Added hash to form in generic_dugga_file_receive to enable hash storage in submission table.


a19arvka (approved):
Works. Same functionality as before but as a more logical way of generating and retrieving hash.

Issue #10191, opened 2021-04-20
Arrows hidden on moments in sectioned


a19isada:
Currently when hiding "Duggor" in a moment, the arrow disappears. There should be an arrow pointing to the right. 
How it looks when "Duggor" are hidden:
![image](https://user-images.githubusercontent.com/81559946/115356200-b4d8de80-a1bb-11eb-9760-8928f2b99da1.png)

How it looks when "Duggor" are visable: 
![image](https://user-images.githubusercontent.com/81559946/115356276-c4f0be00-a1bb-11eb-911a-469ca19177b7.png)

How the arrow should look when "Duggor" are hidden: 
![image](https://user-images.githubusercontent.com/81559946/115356457-f8cbe380-a1bb-11eb-9034-0853b55960bb.png)

a19isada:
Merged and approved by @a20chrjo

PR #10206, opened 2021-04-20
Arrows hidden on moments in sectioned

a19samke:
@a18danli (CO-WORK)
we deleted the unnecessary code,   and add a snippet of  code that makes the arrows show and hide button work

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -72,8 +72,15 @@ function hideCollapsedMenus() {
 //   arrow if it is in the arrowIcons array.
 //	 The other way around for the statistics section.
 function toggleArrows() {
-  $('.arrowComp').show();
-  $('.arrowRight').hide();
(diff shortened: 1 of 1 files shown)


a18willi (changes requested):
When using the #10191-1 branch some of the arrow funtionalities are broken. For example most of the time the arrows wont change when clicking on the purple bar. Secondly sometimes when clicking on an arrow all arrows on all the sections change aswell. This does not happen in the G2-2021-W16 weekly branch so it seems some of the code changes here have caused it.

b18phika (changes requested):
"Request changes" for the same reasons @a18willi mentioned

a19samke:
Yes we noticed it after we committed the code, for example when I click a the first arrow button then the other buttons moves. But right now we have problems how we can think, is it any of you who realize what the fault is @b18phika  @a18willi

a19isada:
This is to an old branch

PR #10207, opened 2021-04-20
Added missing else statement

a19chrel:
Added an missing else statement and moved the error to the correct location


a19nikbo (approved):
lgtm

PR #10209, opened 2021-04-20
Fix fab button overflow in mobile landscape issue#10179

b18phika:
Make fab button items horizontal in mobile landscape mode, to prevent overflow.

Co-author: @a18willi


a19kimgr (approved):
Looks good on my end

a18danli (approved):
Works, but maybe add a new issue to find better solution in the future.

PR #10213, opened 2021-04-20
Removed pointer event from toolTipText to fix the hover.

a19chrel:
Removed pointer event from toolTipText to fix the hover and removed an not hover styling in css because it is not needed due to if not hover the element will go back to default styling.


a19steis (approved):
LGTM, nice fix.

PR #10214, opened 2021-04-20
"Fixed" Make options side panel extend to the top

a19phith:
fixed by replace top 130px to 0 px, could not remove top attribute at all because it affects the height of the panel


a19matma (approved):
image

The options-panel now extends all the way to the top! Good job!

PR #10224, opened 2021-04-20
FAB sm button fixed bottom

a18danli:
Fixed spacing when FAB button is small.


a19isada:
This was only implementad at a local level and it was decided to wait with the implementation of a smaller button until later

Issue #10157, opened 2021-04-19
Line rubber banding to cursor when drawing lines


a19nikbo:
Line creation currently works as follows:

        - You click an element
        - You click another element
        - A line is created between them

How we want it to work is that when you hold down the left mouse button, a line is created between the clicked element and the cursor. If you release the mouse button over another element a line is created between them.

a19nikbo:
Fixed

PR #10229, opened 2021-04-20
Added functionality when drawing lines between elements #10157

a19matma:
When creating lines between elements, it's now possible to see a line between the selected element and the mouse cursor before selecting the next element.

Co-Author:
@a19hammi


a19chrel (changes requested):
If you use the "Escape" button on the keyboard to cancel the mouse mode action, the line does not disappear even if no first element is selected. The issue where the escape functionality was introduced is: #9945.

a19chrel (approved):
LGTM

PR #10231, opened 2021-04-20
Change names for functions and stop background code from running when it shouldn't

a19aleda:
The names are now more suitible, since they do not only enable, but they toggle.
Also added a global variable so that if the ruler is inactive the code in drawRulerBars wont run in the background.


a19matma (changes requested):
Put active in php file on row 125

PR #10235, opened 2021-04-20
change name for enableGrid and enableRuler

a19aleda:
It also added a global variable the is used to check if the ruler is active to determine wether the code for the ruler should be running or not.


a19matma (changes requested):
The values of the ruler should be updated even if its toggle off.
If the ruler is toggled off, and scrolling in the document occurs, and then the ruler is toggled back on, the old value remains. Plz fix :)

a19matma (approved):
LGTM, good job!

Issue #10240, opened 2021-04-20
State machine only store NEW values


a19hammi:
~~We need to store the previous element name that the element had before the change.
An example solution is to pass the old name as a value via the changeList parameter when using the StateFactory.~~

See https://github.com/HGustavs/LenaSYS/issues/10240#issuecomment-823418912

a19hammi:
We have two solutions actually.
Either, we move the creation of the example elements into our functions that create elements so that they are stored on load.
This way we can simply loop back through old changes and find the old name of elements.

The other way is to, as previously said, store old values next to the new values. This does mean we STORE DUBLICATE DATA which could stack up pretty bad if we're talking a theoretical hundreds of thousands diagrams stored.

HGustavs:
Since we want to be able to scrub through the diagram history... I prefer that we only store changes compared to previous version and default elements.

And each time we draw the diagram we loop through the complete history to get the "current" diagram which is the diagram that we are drawing.

This way, scrubbing (or animating likea video) is "free" since we always go through the elements from the beginning each time we update the diagram data (not necessarily each screen redraw).

HGustavs:
The use case that we are aiming for is viewing the complete drawing history of a student like a video to see in which step a mistake was made to judge the thinking of a student when modelling the diagram

a19hammi:
@a19nikbo @a19karwa 
As @HGustavs mentioned we will treat the storage a little bit differently. Creating new issue out of it.

PR #10241, opened 2021-04-20
Fixed: Undo functionality in diagram #9915

### LINKED ISSUE #9915: Undo functionality in diagram
We'll want the user to be able to go back a step when pressing ctrl + z. I'm not sure how many steps back we should save but we can say about 15 for now to not overdo it. We'll want the old state to be saved when:

  - An element is moved
  - An element is created

In the future we'll want to include when an edge is created and possibly more, but since edge creation isn't fixed yet that'll have to be a future issue.

a19hammi:
Pretty much all forms of undoing works.
Since we're missing the storage of PREVIOUS values during a value change there's no way to RESTORE values when stepBack is called.

In other words. We can now step back on all kinds of changes EXCEPT attributes.

fixes #9915


a19hammi:
Reminder for **_reviewers_**. Attribute step-back will be handled in a future issue later since we rely on #10240 to be solved first.

a19chrel (changes requested):
The undo-functionallity does not support removal of created lines while stepping back in the history. Also your if-statements and for-loop does not follow the code standard.

DuggaSys/diagram.js
```diff
@@ -127,20 +142,31 @@ class StateChange
...
             this.name = changes.name;
         }
+
         if (changes.moved) {
-            if (this.moved) this.moved.add(changes.moved);
-            else this.moved = changes;
+            if (this.moved) { 
+                this.moved.add(changes.moved);
+            }
+            else {
```
a19chrel, line 150: Both brackets should be on the same row.

a19chrel (changes requested):
Some small changes are needed to meet the code standard.

a19chrel (approved):
LGTM!

Issue #10053, opened 2021-04-15
Ensure that hash is unique


a18osken:
Since our hash need to have unique matches to redirect we must check whether the generated hash is unique. It's unlikely that it would happen but is real bad if it does.

PR #10245, opened 2021-04-21
G1 2021 w16 #10053 v4

a19arvka:
Finished, chances of generating two identical hashes are close to 0%


a17alfja (changes requested):
Stores a new entry when resubmitting via hash in url.

a17alfja:
as the password is the same for both entries, I think the generateHash on row 976 in dugga.js is run when it is not supposed to

a17alfja (approved):
works as intended. Nice work!

a19gusba (approved):

Issue #10102, opened 2021-04-16
Give localStorage items a lifespan


a18osken:
We were discussing what lifespan the localStorage items should have, we were considering 30 days or maybe the course's runtime. Either or, we do need to purge localStorage items after a while.

PR #10246, opened 2021-04-21
G1 2021 w16 #10102

b18jacha:
Created functions to set lifespan for all localstoragevariants which is used in our ajax call.


a19kevkh (changes requested):
Works, except there is a bug where we dont get a variant.

a18osken:
closing this pull request since we have a new dev branch to work towards and this solution should work in that env. as well

PR #10247, opened 2021-04-21
Make the global timer reset when a new message is pushed

a19aleda:
Added a global variable so that we can clear the timer right before we set a new one.


a19phith (approved):
good, and follows code standard

PR #10248, opened 2021-04-21
Added previous selected elements to box-select

a19chrel:
Added functionallity so if the user has selected elements before started an box-select and CTRL is pressed the elements stays selected until they are marked with the box-select.


b19jenda (approved):

PR #10252, opened 2021-04-21
fixed weird landscape scaling

a19kimgr:
swimlanes scaling might need looking into.


a19isada:
The weird landscape scaling has been fixed already, it seems as a previous issue already solved this issue so this does not need to be merged.

PR #10267, opened 2021-04-21
Arrow on moments in sectioned now always shows Issue#10191

a18willi:
Fixed wierd string that broke arrow functionality.

Co-Authored-By: @b18phika


a19samke (changes requested):
bild111

It works in some causes but not in all Moments ,check the image on above .

a18danli (changes requested):
Same for me, works in some cases only.

a19samke (approved):

a18danli (approved):
Works for me!

a19samke (approved):
It looks like it works!

PR #10270, opened 2021-04-21
Previously selected items cleared when switching to line creation mode

a19steis:
The context is now cleared when switching to line creation mode.


a19karal (approved):
Funkar som det ska :)

PR #10271, opened 2021-04-21
Fixed cmd key on MacOS

a19steis:
Fixed the line selection while holding down cmd key on MacOS.


a19frelu (changes requested):
The main target for this issue is fixed, but a new problem regarding already selected lines have occured. When multiple lines are selected and still pressing down ctrl/cmd you can deselect all but that one line you click on. Nothing should happen when you click on the already selected line, even if multiple lines are selected in the first place.

a19frelu (approved):
Works perfect, nothing to complain about!

PR #10276, opened 2021-04-21
Fixed sliding of elements

a19steis:
Fixed the issue that arose when sliding the elements out of screen.

Also removed the horizontal scroll-bar.


a19frelu (approved):
Seems to work correct. I do not see any other problems connected to this solution. 
Good job!

PR #10277, opened 2021-04-21
Fixed: Missing code #10269 

### LINKED ISSUE #10269: Some missing code from line rubber banding function
Some code is missing, making the line rubber banding to cursor functionality not work. We've located the missing code by looking at the changes in the pull request. This should be in the mouseModes.PLACING_ELEMENT: case in ddown().

image

And this should be at the top of the switch-case in mouseMode_onMouseMove(event)

image

a19hammi:
fixes #10269


a19matma (approved):
LGTM. Though maybe EDGE_CREATION should have come after PLACING_ELEMENT so it matches the order in the other mouse functions with similar switch-cases.

PR #10281, opened 2021-04-22
Fixed so makeRandomID return on first generated ID.

a19chrel:
Added an return statement so the function does not generate a longer id than necessary.


a19steis (approved):
LGTM!

PR #10287, opened 2021-04-22
fixed variant for the 6th time (today)

b18jacha:
fixed variant for the 6th time (today)


a19taral (approved):
everything is working just fine.

a19simma (approved):
Fint, fungerar. Variant kommer upp som det ska

PR #10293, opened 2021-04-22
Store lines and elements deleted together

### LINKED ISSUE #10225: Merging changes of lines and elements being deleted together
As of now, if you delete several elements that have lines then the state machine will store one **StateChange** for all elements being removed and then one for all lines being removed. We should introduce a new kind of **StateChange** with a merge flag of ELEMENT_DELETED and LINE_DELETED that is stored when **both** lines and elements are deleted!

a19hammi:
StateChangeFactory now includes a new method for storing deleted elements and lines together as a single state change.

fixes #10225


a19chrel (approved):
LGTM!

PR #10294, opened 2021-04-22
Added: Tooltips to new buttons and changed keybinds.

a19chrel:
Added tooltips to the new buttons.


a19phith (approved):
Seems to be in order, code standard correct

PR #10295, opened 2021-04-22
Restored Password fetch

a17alfja:
Added password to select in query in showDuggaservice.php and to the array at the end.  Runs setPassword with the retrieved password value in dugga.js


a19simma:
Fungerar, bra jobbat. 
image

a19simma (approved):

a19taral (approved):
it is working fine

PR #10303, opened 2021-04-22
Fixed removing elements when holding down ALT with box selection

b19jenda:
When using the box selection you can now de-select by holding down alt and selecting multiple elements.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1558,13 +1558,20 @@ function boxSelect_Update(mouseX, mouseY)
 
         if (ctrlPressed) {
             var markedEntities = getElementsInsideCoordinateBox(rect);
-
             // Remove entity from previous context is the element is marked
             previousContext = previousContext.filter(entity => !markedEntities.includes(entity));
 
             context = [];
             context = context.concat(markedEntities);
             context = context.concat(previousContext);
+        }else if (altPressed) {
+            markedEntities = getElementsInsideCoordinateBox(rect);
+            // Remove entity from previous context is the element is marked
+            previousContext = previousContext.filter(entity => !markedEntities.includes(entity));
+
+            context = [];
+            context = previousContext;
+
         }else {
             context = getElementsInsideCoordinateBox(rect);
         }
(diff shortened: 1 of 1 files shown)


a19hammi:
Don't forget to mark PR with appropriate labels! :)

DuggaSys/diagram.js
```diff
@@ -1558,13 +1558,20 @@ function boxSelect_Update(mouseX, mouseY)
...
             var markedEntities = getElementsInsideCoordinateBox(rect);
-
             // Remove entity from previous context is the element is marked
             previousContext = previousContext.filter(entity => !markedEntities.includes(entity));
 
             context = [];
             context = context.concat(markedEntities);
             context = context.concat(previousContext);
+        }else if (altPressed) {
+            markedEntities = getElementsInsideCoordinateBox(rect);
```
a19hammi, line 1568: Missing variable declaration statement. This probably creates a global variable without notifying of it.

DuggaSys/diagram.js
```diff
@@ -1558,13 +1558,20 @@ function boxSelect_Update(mouseX, mouseY)
...
         if (ctrlPressed) {
             var markedEntities = getElementsInsideCoordinateBox(rect);
-
             // Remove entity from previous context is the element is marked
             previousContext = previousContext.filter(entity => !markedEntities.includes(entity));
 
             context = [];
             context = context.concat(markedEntities);
             context = context.concat(previousContext);
+        }else if (altPressed) {
```
a19hammi, line 1567: Missing `var` declaration. This will lead to a global declaration which is unnecessary and even if we wanted to make a global we should mark them with the `global` keyword.

a19hammi (changes requested):
Code looks good but missing important keyword that leads to unnecessary global variable declaration.

DuggaSys/diagram.js
a19hammi: I pressed the wrong line. I meant the line after.

a19phith (approved):
The second commit seems to work and no other problems did accrue! good job

Issue #10275, opened 2021-04-21
Swimlanes have different sizes 


a19isada:
Swimlanes have different sizes in different courses and some of them can be hard to read. This needs to be fixed in some way to accomidate for a different number of "duggor". This is one of the examples where it is hard to read. 
![image](https://user-images.githubusercontent.com/81559946/115563714-1f6f4480-a2b8-11eb-814f-7bf6c7a32060.png)

a19isada:
Here, the swimlanes are larger than in the demo-course
![image](https://user-images.githubusercontent.com/81559946/115563988-59d8e180-a2b8-11eb-9925-98a301ada8bd.png)

PR #10308, opened 2021-04-22
Update swimlane layout

a18danli:

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -853,7 +853,6 @@ p {
 }
 
 #swimlaneSVG {
-  width: 70%;
   display: flex;
   margin: auto;
   zoom: 150%;
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -846,18 +846,27 @@ p {
 
 /*Swimlane styling*/
 #statisticsSwimlanes {
-  display: flex;
-  overflow:auto;
-  border: 2px solid darkgray; 
+  max-height: 300px;
+  overflow: auto;
+  border: 2px solid darkgray;
   background-color: var(--color-sectioned-table-lo);
 }
 
 #swimlaneSVG {
-  max-height: 300px;
-  height: auto;
-  margin-left: auto;
-  margin-right: auto;
-  margin-top: 0px;
+  width: 70%;
+  display: flex;
+  margin: auto;
+  zoom: 150%;
+}
+
+#swimlaneSVG text {
+  font-size: 12px;
+}
+
+@media only screen and (max-width: 800px) {
+  #swimlaneSVG text {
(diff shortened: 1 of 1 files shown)


b18phika (approved):

a18willi (approved):
Looks good

a20chrjo:
Changes does not appear to work in Chromium based browsers, please recheck your reviews

@b18phika @a18willi 

Swimlanes have different sizes #10275

a19isada:
I think this is caused by different lengths in weeks. The demo-course has 9 weeks but the datorgrafik has 10 weeks and therefore looks weird. The other courses do not have anything added to them and that is probably why they are to the left. I do not know if this is something we should fix or if we should just leave it as it is? what do you say @a20chrjo?

a20chrjo:
That is true for both Firefox and Chromium based browsers. Though the difference is vastly different between the two.

Swimlanes have different sizes #10275

I think this should be fixed for both browsers because it does not look nice when you have different sizes for the swimlane. Maybe with the exception for the course with less weeks. But then it should at least start at the same position to the left. If it's centered the size should be the same across the board and the course with less weeks should be resized to match the size of the courses with more weeks.

@a19isada

b18phika:
@a20chrjo I think the issue that you displayed in your last comment is at least partially due to the fact that the first and fourth screenshot counting from the top, has two scrollbars on the right. One for the whole page, and I'm guessing that the other one exists because there are too many lanes to display them all at once. If I'm not mistaken this will affect the width of the parent element and thus also the width of the element since the width of the element is set to 70%. 

image

a18willi (changes requested):
After looking into it further it seems that whenever you get the message "You either have no access or there isnt anthing undern this course" message the swimlanes gets really small, atleast in chromium based browsers.  Swimlanes size seems to only be determined by having 70% of the width as its parent object. Looking into the "statisticsSwimlanes" which seems to be the parent object containing "swimlaneSVG" may be needed.

b18phika (changes requested):
For reasons already mentioned by @a18willi

b18phika:
We (@a18willi and I) have pushed a fix to issue#10403 which will hide swimlanes when there aren't any duggor, so that part of the issue is resolved. 

The swim lanes diagram might still have different widths, due to a difference in the number of weeks, but the width of one week is now consistent, which is good. 

In our mind the swimlanes are still a bit too big, they're also not centered horizontally.  Whether or not that should be part of this issue or its own issue is a question for somebody else.

a19isada:
The swimlanes should be resized to make them a bit smaller. Swimlanes not being centered horizontally will be its own issue.

a19isada:
Old pull request. Has been renewed.

PR #10316, opened 2021-04-22
Added restriction for delElement

a19kimgr:
Fixed issue that delete button was overlapping course text


a19danla (approved):
It turns invisible while in mobile view.

b19wilfo (approved):
Works, the delete button no longer covers course-name.

PR #10323, opened 2021-04-22
Fixed keybinding in tooltip

a19steis:
Fixed the keybinding display when showing the tooltip for toggle grid and ruler.


a19karal (approved):
Funkar som det ska :)

Issue #10067, opened 2021-04-15
Fix so that the characters are inside the box, making it more readable.


a19karwa:
When putting out a relation element, the name currently goes outside the box. It should be fixed so all of the characters stays inside the rect element.

![image](https://user-images.githubusercontent.com/81557429/114849204-d282fe00-9ddf-11eb-86ed-e29e63c952f1.png)

a19hammi:
fixed in #11057

PR #10325, opened 2021-04-22
Group3 2021 w17 #10067

b19phier:


a19phith (changes requested):
image
Seems to be working with some minor problems, the text seems to stick out of the entity when alter the size of the entities. And the text seems to alter when the zoom in and out function use

a19nikbo:
I would like to add that we preferably don't want to influence the spacing between characters as it makes things look inconsistent and can be difficult to read.

a19phith (approved):
seems to be working

a19nikbo:
We'll have to put this on ice for the time being as it seems to be a difficult issue to solve in a satisfactory way. I'll close the pull request for now to avoid accidental merge

PR #10326, opened 2021-04-22
G1 2021 w17 #10102 v3

e15marol:
The variant now loads and the function to set the time to live now works, for each time the get time expire function is loaded, it checks if the variants TTL has expired or not. When its expired, it removes the key and we get a new variant.

### DIFF
--- a/DuggaSys/showDuggaservice.php
+++ b/DuggaSys/showDuggaservice.php
@@ -222,10 +222,6 @@ if($demo){
 	$query->execute();
 	$variantsize = $query->fetchColumn();
 
-	if($variantvalue == "undefined") {
-		$variantvalue = $savedvariant;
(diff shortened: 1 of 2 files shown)


b18jacha:
To test if assigned variant resets, change the value in the setExpireTime function on line 1026. (showduggaservice)

a19arvka (approved):
Works as intended. Nice solution. 
Will cause merge conflicts in dugga.js.
Please comment code in the future.

a18osken (changes requested):
Currently variant values gets set to 0 which means the assignment will not save correctly and redirection will fail

a19arvka (approved):
Works!

a18osken (approved):
This works well, nice job!

Issue #10311, opened 2021-04-22
Expand on the log-in functionality 


a18osken:
We currently require the user to submit a password every time we refresh the page. We should now make it so that we do not prompt the user for their password when they refresh the page for example.

For how long we should keep their logged-in status for an assignment is a little bit up in the air. Should we allow them to remain logged in for the duration of their browser or should we require a password after leaving the assignment for example.

Session storage is entirely fine to use since the client can't access any of the temporarily stored data

PR #10330, opened 2021-04-22
G1 2021 w17 #10311

a19simma:
2 character issue @a18osken, ty 👍 
#10311


e15marol (approved):
Works good! Simple solution.

b18jacha (approved):
Password are no longer required every refresh

Issue #10305, opened 2021-04-22
Add functionality to edit properties of lines in "properties box" 


a19nikbo:
When selecting an element you have the option to edit its properties in the properties box like this:

![image](https://user-images.githubusercontent.com/81541557/115703134-0921c100-a36a-11eb-81d8-bc3693b686a2.png)

This should be possible for lines as well. For now it's enough to make a drop-down menu where you can choose between partial / total participation (single line or double line).

a19nikbo:
Fixed

PR #10337, opened 2021-04-22
You can now change line properties in the options panel #10305

a19matma:
It's now possible to change the line from Normal to Double or vice versa in the options panel after selecting a line.

Side note: 
Warning text when selecting more than one element doesn't work correctly on mouse-down when switching from selecting an entity to a line. Further investigation required.


a19matma:
The updated state of line.kind is not pushed to statemachine the when the save-button is pressed.

a19phith (approved):
Works as intended, code follow standard! good job

PR #10338, opened 2021-04-22
Change element name and state with Enter

b19jenda:
When you select an element and re-name it or change the state you can now also save by clicking "Enter"

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -684,7 +684,7 @@ document.addEventListener('keydown', function (e)
 {
     // If the active element in DOM is not an "INPUT" "SELECT" "TEXTAREA"
     if( !/INPUT|SELECT|TEXTAREA/.test(document.activeElement.nodeName.toUpperCase())) {
-
+       
         if (e.key == keybinds.LEFT_CONTROL.key && ctrlPressed !== true) ctrlPressed = true;
         if (e.key == keybinds.ALT.key && altPressed !== true) altPressed = true;
         if (e.key == keybinds.DELETE.key && e.ctrlKey == keybinds.DELETE.ctrl) {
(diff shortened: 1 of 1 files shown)


DuggaSys/diagram.js
```diff
@@ -716,6 +716,14 @@ document.addEventListener('keydown', function (e)
             removeLines(contextLine);
             updateSelection();
         }
+       
+    }
+    // If the active element in DOM is an "INPUT" "SELECT" "TEXTAREA"
+    if ( /INPUT|SELECT|TEXTAREA/.test(document.activeElement.nodeName.toUpperCase())) {
+        if (e.key == "Enter") {
```
a19steis, line 723: Check how the other keybinds are coded.

a19steis (changes requested):
It works, but we have declared a list of keybinds " const keybinds = {...} " that we use to keybind specific keys. Its in my opinion better to be consistent (see how we coded the other keybinds in the event listener).

DuggaSys/diagram.js
```diff
@@ -716,6 +717,14 @@ document.addEventListener('keydown', function (e)
             removeLines(contextLine);
             updateSelection();
         }
+       
+    }
+    // If the active element in DOM is an "INPUT" "SELECT" "TEXTAREA"
+    if ( /INPUT|SELECT|TEXTAREA/.test(document.activeElement.nodeName.toUpperCase())) {
```
a19steis, line 723: Remove this line, its redundant because you can just put the code on row 724 under row 687 where this IF already exists.

DuggaSys/diagram.js
```diff
@@ -716,6 +717,14 @@ document.addEventListener('keydown', function (e)
...
+        if (e.key == keybinds.ENTER.key) {
```
a19steis, line 724: take this if and put it on row 720.

Change the conditional statement to (e.key == keybinds.ENTER.key && e.ctrlKey == keybinds.ENTER.ctrl)

a19steis (changes requested):
The reason we want to have "&& e.ctrlKey == keybinds.ENTER.ctrl " in the condition is so that we cant save with enter while simultaneously holding down CTRL.

a19steis (approved):
looks good!

PR #10343, opened 2021-04-23
"fixed" Accidental deletion of if-statements in removeElements

a19phith:
If statement was removed from the previous pull request and are now added in this :)


a19frelu (approved):
Looking good. Work fine. Indented correctly. Ready to be merged with the other friends of its kind.

PR #10344, opened 2021-04-23
Added custom save button

a19steis:
Custom styled the save button and added color shade to achieve some feedback.


a19matma (changes requested):
Maybe make the change in color a bit more pronounced. It's currently very subtle when hovering it and maybe some additional effects when clicking on it as well.

a19matma (approved):
Great job! Very nice, high five.

PR #10348, opened 2021-04-23
fixed

a19taral:
commented out the score variable


a19simma:
Score should be null in the database on new duggor saves.
image

a19rebsa (approved):
It works fine

a18osken (approved):
I like what you got

Issue #10146, opened 2021-04-19
Connect submissions to hash so that file submissions also are shown when revisiting a submitted hand-in


a18osken:
Currently file submissions/uploads are not saved with hash so when revisiting a file-submission assignment via hash redirection the submitted files does not come along for the ride. We should fix this.

b19ponka:
Hard to solve/test before issue #10167 (Fix the bummer issue)

a13zanal:
Need to link file submissions / uploads to hash instead of userid and to see if it works, Bummer error needs to be removed.

a13zanal:
To upload a file for submission in the database, hash needs to be added to the submission insert, in Dasty replace uid with hash and delete if login check (ha). The last thing that needs to be resolved is that the uploaded file should be displayed in previewTable.

a13zanal:
Created new branch (G1-2021-w17- # 10146-v2) because code was missing when merging into w17 branch. Commitat our previous changes as above and solved so that you get the same hash and password when uploading. What remains is to make the file appear in the "previewTable". In dugga.js there is an if statement on line 1642 which is not executed even though filez is not "undefined".

b19ponka:
Connected submissions to hash. Made some progress with showing submitted hand-ins when revisiting but not quite there yet

PR #10349, opened 2021-04-23
G1 2021 w17 #10146 v2

a13zanal:
File submissions/uploads are now saved with hash and appears in the table.


b19ponka:
Test so previous hand-ins are shown when not logged in. Before uploading files save an empty dugga and save the URL and password. Use that URL/password when uploading files.

a19rebsa (approved):
Works fine, good job!

a18osken (approved):
Yeah this works well, good job!

PR #10351, opened 2021-04-23
StateMachine initial state stored on diagram launch

### LINKED ISSUE #10334: Initial state: Premade data set state OR Empty slate?
Currently there's no way to go ALL the way back via the state machine. We need an INITIAL state that we originate from.
This state needs to be stored inside the state machine if we use a **default set of elements** when making a new diagram file.

If we instead go with a **blank slate** then don't need to store the initial state in our state machine but will need to create our example elements to perform tests on. I therefore recommend we create a function to generate the diagram that we're currently using. This function should then call our own createElement() functions so that they're recorded by the state machine correctly.

Cut short, we have two options what I know of. @HGustavs has said that both is fine so I will leave this one up for discussion in the group so that we then can decide which one is the most appropriate!

a19hammi:
fixes #10334

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -295,7 +295,7 @@ class StateChangeFactory
 
 class StateMachine
 {
-    constructor ()
+    constructor (initialData, initialLines)
     {
         /**
          * @type Array<StateChange>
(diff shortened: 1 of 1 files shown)


a19chrel (approved):
LGTM!

PR #10353, opened 2021-04-23
Added function to de-select selected lines with esc

b19jenda:
Now all lines can be de-selected when pressing esc


a19matma (approved):
LGTM! Good job!

PR #10357, opened 2021-04-23
Escape key returns to pointer state

a19steis:
Now when pressing escape from any other mode we always return to the "pointer" mode.


a19chrel (approved):
LGTM!

PR #10366, opened 2021-04-23
Added key binding for selecting all elements

a19aleda:
Added key binding for selecting all elements and add function to select all elements and lines and put them into respective context array


a19steis (approved):
Seems to do the trick, good job !

PR #10370, opened 2021-04-23
Enabled shortcuts when caps lock is on

a19steis:
The shortcuts that need to work when caps lock is on are now doing that regardless if the letter is upper or lower case.


a19frelu (approved):
This absolutely solves the issue. Good job.

Issue #10312, opened 2021-04-22
Check if there are several logins to the same hashed assignment


a18osken:
We should create some kind of token to see if there are several users currently logged in on a hashed assignment, this is to enable functionality to synchronize group assignments later.

a19gusba:
How should groups work? Should the teachers provide an url to the group with a hash or should one student create a hash and share with the others in the group?

a19gusba:
found problems when a student closes the browser in ways such as using the task manager or turning off the computer the javascript event that triggers whenever the tab closes doesn't get called when using the two methods mentioned.

PR #10371, opened 2021-04-23
G1 2021 #10312

a17alfja:


a17alfja:
The user that initially creates the dugga does not count as an active user in this solution, only using the hash to login and view the dugga will (unsure if students or teacher will create dugga).

a17alfja:
also you will need to run the create table for groupdugga in init_db.sql for it to work

a17alfja:
you can now use the file createTables.sql to create groupdugga table

a16erilj (commented):
Didn't find anything that broke during testing. So I'll give my approval that regard. However i would like some comments about the idea behind some of functions behind the code.

PR #10373, opened 2021-04-23
Fixed bug concerning arrows issue#10369

a19klafr:
Fixed the bug that made two arrows to show on initial small screens


a18danli (approved):
Works for me!

a19samke (approved):
it looks right & works for me!

Issue #10347, opened 2021-04-23
Remove the "now placing" text and the box around it


a19nikbo:
It hasn't been used in a long time, and since the cursor modes are highlighted in the toolbox now it's no longer necessary to have something like that.

![image](https://user-images.githubusercontent.com/81541557/115840950-8d348100-a41c-11eb-93ee-7c1ace5de468.png)

a19nikbo:
Fixed

PR #10375, opened 2021-04-23
Removed the "now placing" box from the options-pane.

a19frelu:
The box and text had no functionality in the current diagram and therefore it was removed.
Issue #10347


a19steis (approved):
LGTM!

PR #10376, opened 2021-04-23
Centralized course text for mobile view

a19kimgr:
fixed so course name is more centralized in mobile view.

### DIFF
--- a/DuggaSys/sectioned.php
+++ b/DuggaSys/sectioned.php
@@ -173,7 +173,7 @@
 					</ol>
 			</div>
 				<div style='flex-grow:1'>
-						<span id='course-coursename' class='nowrap ellipsis' style='margin-left: 5px;margin-right:10px;'>UNK</span>
+						<span id='course-coursename' class='nowrap ellipsis' >UNK</span>
 						<span id='course-coursecode' style='margin-right:10px;'>UNK</span>
 						<span id='course-versname' class='courseVersionField'>UNK</span>
 				</div>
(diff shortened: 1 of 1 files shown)


b18phika (changes requested):
It's impossible to center the course name exactly with hardcoded margins since the course names have different lengths. I think it'd be better to use a solution that relies on flexbox (display: flex), and apply a margin-right that equivalent to the width of the arrow on the left, to get it exactly centered.

a18willi (commented):
Looks better. But currently it looks like the text is centralized inbetween the right edge of the arrow/arrows and the right edge of the purple banner. For the text to be completely centralized the text position should not take the arrows into consideration. It looks like this problem still stands even if not in mobile view, it just gets more obvious when in mobile view. Maybe a new issue should be created to get this fixed.

a19kimgr:
The coursename is now in the middle, always, without any margins, with the exception if screen width is smaller than 200px where the text starts to disapear. However, most (if not all ) mobile devices have a screen width bigger than 200px so this should now be fixed.

b18phika:
image

As shown in the image, the text is centered between the arrows and the right side. It's however not in the center of the screen, which I believe to be the intended behavior.

a18willi (changes requested):
Requesting changes due to the same reason @b18phika stated.

b18phika (approved):

a18willi (approved):

PR #10378, opened 2021-04-23
Disabled right click when in edge creation

a19steis:
You can no longer right click an element when in edge creation mode to create lines.


a19phith (approved):
Seems to be the fix to the problem, follows code standard!

PR #10390, opened 2021-04-26
Diagram: Updated the displayMessage system to show multiplie messages

a19chrel:
Made som changes to the displayMessage system in the diagram.


a19steis (approved):
Multiple messages are shown upon actions, i like the interactive timer countdown. Good job!

PR #10392, opened 2021-04-26
Clear context on paste and add the new Elements to context

a19chrel:
Now when pasting in elements, the selected elements gets deselected, and the created elements gets added to context.


a18nadma (approved):
Code is short and doesn't collide/disrupt any other code!

PR #10395, opened 2021-04-26
Fixed error with pasting elements on each other

a19chrel:
Changed so the flag "ELEMENT_AND_LINE_CREATED" does not have the canAppendTo, this coused the problems described in the issue. Also changed some parameters for the displayMessages to the recent added enums.


b19jenda (approved):

Issue #10265, opened 2021-04-21
Fix contrast issues in sectioned page


a19isada:
If you use the lighthouse tool while inspecting the sectioned page, there is a problem with the contrast with the background and the foreground colour on most of the text. This needs to be fixed to make the page more accessible. The colours should still follow the LenaSYS colour scheme which can be found here: https://github.com/HGustavs/LenaSYS/wiki/CodeViewer

These are some of the problems: 
![image](https://user-images.githubusercontent.com/81559946/115555552-ef23a800-a2af-11eb-8dac-fd0c5258f83f.png)

PR #10396, opened 2021-04-26
Update style.css (Issue #10265)

b19alest:
Made the headers background darker and the text darker. The lighthouse tool still complains about availability but the contrast problem is gone.

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -823,7 +823,7 @@ p {
 
 #Sectionlist .section {
   color: var(--color-text-header);
-  background-color: var(--color-primary-light);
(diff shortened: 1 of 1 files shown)


a19isada (changes requested):
The colours should use the predefined variables for colours. They can be found at the top of the style.css file

b18phika (changes requested):
Lighthouse doesn't complain about contrast issues anymore. However, I'll have to agree with @a19isada, that the css-variables should be used.

a18willi (changes requested):
Better results in the lighthouse tool but as @a19isada mentioned the predefined variables should be used instead of hardcoded colorcodes

PR #10397, opened 2021-04-26
Change color scheme for toolbar

a19aleda:
Change color scheme for toolbar to match duggasys scheme better


a19matma (approved):
The new color-scheme for the toolbox looks good and as intended.

PR #10398, opened 2021-04-26
Avoiding line crossing on most frequently used browsers. Issue  #10092

### LINKED ISSUE #10092: Diagram does not avoid crossing lines automatically
Seeded version of diagram engine (from https://github.com/HGustavs/DiagramEngine/) avoids crossing lines, but the current LenaSYS code does not.

test2

a19frelu:
Now the lines between the objects and its attributes and relations will not be crossing each other, wich will make a more sleek look to the diagram.

Implemented the code from the solution of week 16 branch with the  branch of week 17 to be able to merge it with the week 17 main branch.

Issue #10092 
close #10092

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -2393,10 +2393,10 @@ function sortvectors(a, b, ends, elementid, axis)
         toElementB = data[findIndex(data, lineB.fromID)];
     }
 
-    if (navigator.userAgent.indexOf("Chrome") !== -1) {
-        sortval = 1;
-    } else {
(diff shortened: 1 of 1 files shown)


a19frelu:
Skärmavbild 2021-04-26 kl  10 01 50 (2) (kopia)

a19matma (changes requested):
Works good, but to make it cleaner "return - sortval;" should probably be written so it just returns sortval instead of - sortval.

a19matma (approved):
LGTM! Good job!

PR #10401, opened 2021-04-26
Added setCursorStyle to getData

b19jenda:
This means the cursor style will be pointer when page is loaded


a19frelu (approved):
Nothing to complain about. Muy bien.

PR #10407, opened 2021-04-26
Now hides swimlanes if there is a noaccess-message. Issue#10403

a18willi:
"StatisticsSwimalanes" is hidden if a noaccess-message is displayed.

Co-author: @b18phika

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -1164,6 +1164,7 @@ function returnedSection(data) {
       // But disable sorting if there is a #noAccessMessage
       if( $("#noAccessMessage").length) {
         $("#Sectionlistc").sortable("disable");
+        $("#statisticsSwimlanes").hide();
       }
     }
   } else {


a18danli (approved):
Looks good!

a19samke (approved):
It works but maybe this is not the best solution, When we created the issue we thought that shouldn`t see  the arrow buttons.

a19isada (changes requested):
When I visit a course that does not have any content, the week header is still there. Maybe the issue was unclear but this header should be removed when there are no "Duggor". 
image

a20chrjo:
> 
> 
> When I visit a course that does not have any content, the week header is still there. Maybe the issue was unclear but this header should be removed when there are no "Duggor".
> image

This happens when using Firefox, the changes works when using Chromium based browsers.
Though on Chromium based browser you have to click the arrow twice to show the swimlane again and this needs to be fixed as well.

b18phika:
Discarded this solution and started on a new one on a new branch.

PR #10408, opened 2021-04-26
Defocus and display upon enter

a19steis:
The input field is now defocused and a message is displayed upon hitting "enter". i.e saved.


a19frelu (approved):
Works great. Indeed a good idea to implement this type of clearification.

PR #10415, opened 2021-04-26
Make button text color darker, to improve contrast issue#10404

b18phika:
Co-Authored-By: a18willi <81558529+a18willi@users.noreply.github.com>


a19anjst (approved):
Looks good even though the change was difficult to see at first

Issue #10364, opened 2021-04-23
Generate ID's for pre-made data contents


a19hammi:
Right now we have several lines of ids that are stored in global space that are only used in a single place of our code. Not only that but we could literally just use the makeRandomID() on that one single place as-well. Unnecessary variables!

HGustavs:
These variables are necessary to know which id-s to connect the relationship ends to...

but I think we can use the history approach and explicitly store the identifiers and not generate identifiers dynamically.

a19nikbo:
Fixed

PR #10417, opened 2021-04-26
Created function for loading pre-generated objects #10364

a19matma:
onSetup function for loading in the pre-determined objects


a19hammi (approved):
LGTM!

PR #10420, opened 2021-04-26
Remove unessassary line of code that creates multiple texts

a19aleda:
remove unessassary line of code that creates multiple texts


a19steis (approved):
LGTM!

Issue #10361, opened 2021-04-23
Do not prompt for a password when refreshing the page for uploading a file.


a18osken:
Currently we will be prompted for a password when the page refreshes to display the uploaded file. There's an issue here in that we've yet to display the password to the user. If possible we should make it so that we do not prompt for the password when we refresh the page in `filerecieve_dugga.php`

a18osken:
I was considering maybe adding an element to the URL when you auto-refresh,a (0/1) that can be getOPG'd and decide whether the password prompt is triggered or not.

PR #10421, opened 2021-04-26
G1 2021 #10361

a19kevkh:
Do not prompt for a password when refreshing the page for uploading a file.


a19frebo:
Added a SESSION variable that is set to '1' when we upload and resets to '0' after.

e15marol (changes requested):
When uploading a file and you hit the refresh button, you're prompted a password.

b18jacha (changes requested):
From what we can see, after you have uploaded a file and refresh the page, it will prompt a password request

a19frebo:
The hasUploaded variable is now reset when the dugga is saved into the database

e15marol (approved):
The fixes solved the problem, good job! Works now.

b18jacha (approved):
**Very good**

> Works as intended

PR #10425, opened 2021-04-26
Fixed typo. Issue#10418

a18willi:
Replaced all mentions of "feedbacks" with "feedback". Note that there are several css classes named with "feedbacks" but we have not touched these as to not cause any unexpected errors(no refactoring rule).

Co-author: @b18phika


a19isada (approved):
Fixes the typos.

PR #10437, opened 2021-04-27
Added cardinality to the diagram

a19chrel:
Added so you can add values to the start and end to the line.


b19jenda (approved):

Issue #10377, opened 2021-04-23
Panning deselects lines but not elements


a19nikbo:
Panning should not deselect any selected elements or lines, but lines are deselected as soon as the right mouse button is clicked.

a19nikbo:
Fixed

PR #10438, opened 2021-04-27
Group3 2021 w17 #10377

a18nadma:
Selected lines no longer gets deselected when moving a line \o/

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1077,7 +1077,6 @@ function mup(event)
                         clearContextLine();
                     } else if (mouseMode == mouseModes.POINTER) {
                         updateSelection(null);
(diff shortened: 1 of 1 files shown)


a19karal (approved):
Funkar som det ska :)

a19nikbo:
This does not quite do what the issue was about. Note that it says "*Panning* deselects lines but not elements". This means that clicking on the background without moving the mouse should still deselect all lines. It's only when panning that they shouldn't be deselected.

a19karal (changes requested):
Funkar inte som det ska >:(

lines are still "selected" after you click away (background)

a19karal (approved):
Funkar som det ska :)

PR #10441, opened 2021-04-27
Fix: Prevented error on using GUI zoom buttons

### LINKED ISSUE #10439: Diagram: Zoom buttons broken
Due to an update with the change of the camera while zooming the zoom buttons in the tool-bar does not work at all. This is due to an undefined event.

bild

a19hammi:
fixes #10439


a18nadma (approved):
zoom in and zoom out buttons works well now and the code doesn't interrupt anything, well done!

PR #10444, opened 2021-04-27
"fixed" Switch out save button for line properties for the same one a…

a19phith:
Fixed by adding the button to the same class as the other save button:)


a19karal (approved):
Funkar som det ska :)

PR #10448, opened 2021-04-27
Removed line that caused statemachine to save multiple elements

a19aleda:
Removed line that caused statemachine to save multiple elements


b19phier (approved):
Looks fine and works for me!

PR #10451, opened 2021-04-27
mobile View

a19samke:
Change tabs2 to 32 px, then we comment out the (Vertical line for 2 tabs) maybe the code might be useful in the future.

@a18danli


a19klafr (approved):
No unnecessary spaces in dugga section, everything looks good!

PR #10452, opened 2021-04-27
Update displayMessage to use the passed functions instead of context

a19aleda:
Update displayMessage to use the passed functions instead of context


a19matma (changes requested):
These requested changes are still not implemented:
image

a19matma (approved):
LGTM!

Issue #10394, opened 2021-04-26
Switch out single/double line settings for a checkbox instead of a drop down menu


a19nikbo:
The drop down menu makes the user have to go through a lot of unnecessary clicks just to change between partial and total participation (single or double line). It would probably make more sense to have it be a checkbox or something similar.

![Untitled](https://user-images.githubusercontent.com/81541557/116051096-85711880-a678-11eb-91ee-e87583d833a5.png)

a19nikbo:
Fixed

PR #10453, opened 2021-04-27
Change: From Drop-down menu to radio button #10394

a19matma:
It's now possible to change the line.kind attribute with radio buttons instead of the previous drop-down menu.


a19frelu (approved):
Works good. No errors occured when tested.
Keep it up!

Issue #10434, opened 2021-04-27
Update the installer test-data inserts to comply with altered tables


a18osken:
As we've changed several tables it would be nice to have the installer up to date as well.

HGustavs:
I agree.

I suggest not doing this too often but one towards the half-time seminar and once towards the end of the course seems fine

b18jacha:
> I agree.
> 
> I suggest not doing this too often but one towards the half-time seminar and once towards the end of the course seems fine

Should we keep uid or have uid as NULL?

HGustavs:
Remind me, what does the uid do is it used in more than one location?

b18jacha:
We are working in userAnswer and Submission, It's used to link the userID

HGustavs:
If it is no longer used it is best to keep the column and set it to some value for example null, or UNK. This way if some module tries to read this information it does not crash.

PR #10458, opened 2021-04-27
G1 2021 w17 #10434

b18jacha:
We have updated userAnswer and submission in testdata.sql with new hashes, passwords etc.


e15marol:
If you want to test the new testdata you can simply check the inserts for userAnswer and see the corresponding url hashes and passwords, and copy paste the hash into the url to retrieve the saved dugga, and then the password to display it. First of all you'll have to reinstall lenasys with the new testdata, and view the data through mysql workbench.

a19kevkh (approved):
It works but some other stuff broke that isnt related to this issue. If we dont have anything in localstorage, variant doesnt get retrieved.

a19frebo (approved):
Insertion of the values works, however we found another issue that if we do not have a local storage and try to go in to a dugga through the hash we don't get a variant.

a19rebsa:
We'll create a new issue for fixing the variant problem

PR #10459, opened 2021-04-27
"fixed"Make line cardinalities zoom with the rest of the elements.

a19phith:
fixed by adding style to the text attribute to make them act the same way as the text in the entities


a19matma (approved):
LGTM!

PR #10464, opened 2021-04-27
Moving element and resizing now works

a19steis:
Moving an element and then resizing now works as intended.

Updated the code inside of the stateChange class to follow new standard using values.

No external behavior changed.


a19phith (approved):
Tested? yes
Work? yes
code standard? yes
hotel? trivago

Issue #10306, opened 2021-04-22
Make ruler follow grid


a19nikbo:
We've just made it so that the grid follows the panning when you click-drag on the background canvas. It was nice having the numbers on the ruler live update when the grid was static, but it looks a bit weird now that they're not synced. It would be better if they didn't live-update, but instead that the lines and numbers on the ruler panned left and right on the top bar, and up and down on the side bar as you panned the canvas.

a19hammi:
fixed in #10510

PR #10469, opened 2021-04-27
Group3 2021 w17 #10306

a19vikth:
Fixed ruler so that the numbers and lines moves when panned


a19nikbo:
Visually it looks really good, but it doesn't draw the ruler for coordinates below 100 on both the x and y axes.

a19vikth:
Ruler should now show negative coordinates

a19vikth:
Closed. This is the new branch https://github.com/HGustavs/LenaSYS/tree/Group3-2021-W18-%2310306

PR #10475, opened 2021-04-27
Fixed so only one fab_action function exists

b19jenda:


a19frelu (approved):
The deleted code does not make any negative changes in the functionality.
Fine work right here.

PR #10478, opened 2021-04-28
Diagram: Added snap to grid

a19chrel:
Made some changes to the grid position so it now is statis to the coordinates in the diagram. Also added the functionality to support snap to grid.


a19phith (approved):
Looks good, did not find any problem with the fix. Gj

PR #10479, opened 2021-04-28
Fix so that elements can now be resizeable on the side that a line is…

a19aleda:
Fix so that elements can now be resizeable on the side that a line is connected


a19matma (approved):
LGTM! Good job!

Issue #10466, opened 2021-04-27
If you do not have a variant item in localstorage you do not get one assigned when visiting a hash.


a18osken:
Right now, if you clear your localStorage before returning to a hash, or if you do not have one to start with, you will not have a variant and therefore the assignment will not be displayed correctly. We need to make sure that they get the saved useranswer variant .

PR #10483, opened 2021-04-28
G1 2021 W17 #10466

e15marol:
The corresponding variant to the hash URL should now load correctly, when "visiting" a hash, and the saved answer should display correctly. To check this compare the saved answer to the useranswer in the userAnswer inserts, after loading the URL address with the hashes. You could also reset the local storage and save a dugga yourself, and then use the hash and password that is generated to test this issue.

### DIFF
--- a/DuggaSys/showDuggaservice.php
+++ b/DuggaSys/showDuggaservice.php
@@ -242,7 +242,8 @@ if($demo){
 	$query->execute();
 	$variantsize = $query->fetchColumn();
 
-
+	//If the variant value is unknown (E.G: UNK) then we retrieve the variant from the variant set in useranswer 
+	//where there exists a corresponding hash, and set the resulting useranswer.variant into $variantvalue
 	if($variantvalue == "UNK") {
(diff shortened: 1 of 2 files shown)


a19rebsa (changes requested):
Now you do get a variant when local storage is empty while visiting an already submitted assignment/dugga, but you do get different variants. If I submit an assignment/dugga e.g. "biträkningsdugga" with number 25 and then empty my local storage and then go to the hash url, I get number 2. This should be fixed

a19frebo (commented):
I have the same issue, after saving an assignment and then deleting the local storage. If you then go to the hash url you have a chance to get a different variant

a19frebo (changes requested):
When you get a variant for a dugga in localstorage and enter an assignment stored in database with a different variant, the variant from local storage is loaded instead of the variant from the database.

a19rebsa:
This needs to be based of the new week 18 branch, so closing this pull request in the mean time.

PR #10484, opened 2021-04-28
SVG button added

a19steis:
Created a SVG image for the "snap to grid" button.


a19karal (approved):
funkar som det ska :)

change the keybind tho, same as toggle ruler, but im guessing u will in the future

PR #10487, opened 2021-04-28
Added: Panning with mouse3 (mousewheel)

a19chrel:
It is now possible to pan with the mousewheel pressed (mouse3)


a19karal (approved):
Funkar som det ska :)

PR #10506, opened 2021-04-28
"fixed" "Snap To Grid" button needs a keybinding

a19phith:
fixed by creating keybinding to the key s. Now works on W18 version


a19steis (approved):
Looks good!

PR #10507, opened 2021-04-28
Removed panning cursor

a19steis:
Removed the panning cursor when pressing middle mouse button


a19matma (changes requested):
It's still possible to get the bug if you do the following:
Click with the mouse-wheel-button while the cursor is in either the toolbox panel or the options panel, then move the cursor into the diagram and zoom in and out with the mouse-wheel and the bug appears again.

a19matma (approved):
LGTM! Good job Stu!

PR #10509, opened 2021-04-28
Updated onfocus function

a19steis:
Added ctrlPressed check to onfocus function.


a19frelu (approved):
Works great, good job!

PR #10510, opened 2021-04-28
Update diagram.js

a19vikth:
Fixed ruler so that the numbers and lines moves when panned and the ruler should now show negative coordinates

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -730,7 +730,7 @@ function getData()
     container = document.getElementById("container");
     onSetup();
     showdata();
-    drawRulerBars();
+    drawRulerBars(scrollx,scrolly);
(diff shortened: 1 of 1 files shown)


a19chrel (changes requested):
The number on the ruler does not match the coordinates in the diagram.

For example the position of the person-entity is (100, 100), and is drawn from the top-right corner. As shown with the red dot. This does not line up with the ruler.
bild

Also while the user is in zoomfact 0.125 the x-ruler is unreadable an solution for this can be to use vertical writing. Some CSS-Code that could be used.
`.vertical-writing
{
  writing-mode: vertical-rl;
  text-orientation: upright;
}`

a19chrel (approved):
Looks good, nice work!

PR #10512, opened 2021-04-28
Swimlanes now properly hides when there are no "duggor". Issue#10403

a18willi:
Swimlanes should now properly hide in all browsers when there are no tests or moments in a course. Swimlanes should now also stay hidden when clicking the close/open swimlanes button multiple times. 

Note: A new branch has been used after the merge. All changes in branch G2-2021-W17-ISSUE#10403 should therefore be disregarded.


a18danli (approved):
Looks like it works!

a19samke (approved):
It looks like it works!

PR #10515, opened 2021-04-29
SnapToGrid is now working on ghost elements

a19chrel:
The snap to grid functionallity does now also work with the ghost elements.


a19steis (approved):
Seems to work as intended, good job!

PR #10518, opened 2021-04-29
G2 2021 w18 issue#10494 - hamburger menu now disapears when screen becomes big again.

a19kimgr:


b18phika (approved):
Works as expected both in Chrome and Firefox.

a18willi (approved):
Works as described.

PR #10521, opened 2021-04-29
Fixed cardinality zoom and position

b19jenda:
Now the cardinality is better positioned when moving around elements and stays in place when zooming.


a19nikbo:
Note to the reviewer:

We have a separate issue regarding cardinality positioning depending on line angle etc., so that is not part of this issue.

a19aleda (approved):

PR #10522, opened 2021-04-29
Added parameter successMessage to addLine

a19chrel:
Added parameter successMessage to addLine to enable toggle if the line created message should be displayed.


a19matma (approved):
LGTM!

PR #10523, opened 2021-04-29
Analytics page and "hamburger"-button now shows properly for both teachers and superusers. Issue#10499

a18willi:
Access to analytics now checks if the user is a superuser or creator. Most teachers are creators but not superusers therefore the old code only including superusers didnt work.

### DIFF
--- a/DuggaSys/analytic.php
+++ b/DuggaSys/analytic.php
@@ -1,13 +1,18 @@
 <?php
 session_start();
-
-// Show analytics if user is superuser, otherwise redirect them to courseed.
-if(!isset($_SESSION["superuser"]) || $_SESSION["superuser"] != 1) {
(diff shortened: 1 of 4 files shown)


a18danli (approved):
Works for me!

a19samke (approved):
It works for me!
 I also wonder if the hamburger button  shoud be on the section page, when logged.in as a student ?

a19isada:
> It works for me!
> I also wonder if the hamburger button shoud be on the section page, when logged.in as a student ?

It should not, those menues are only available for teachers and superusers

a18willi:
Note: In the description of this pull request i mentioned that most teachers are marked as creators. This is not the real purpose of the creator table and therefore should not be used in this solution and should not be used to identify teachers in future solutions.

PR #10526, opened 2021-04-29
Fixed "deselect all" functionality

a19steis:
The "Escape" key now detects if anything is selected and deselects  it/them, then returns to the mouseMode it is currently in. If "Escape" is pressed again with nothing selected we return to the default mouse mode.


b19jenda (approved):
Seems to work as intended, good work!

PR #10527, opened 2021-04-29
Group3 2021 w18 10514

a18nadma:
Made the select cardinality bars better positioned and made it look prettier! Also added "none" to the empty option!

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -319,3 +319,10 @@
 #fieldsetBox {
     margin-left: 20px;
 }
+
+/*Cardinality stuff*/
+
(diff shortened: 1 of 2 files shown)


a19steis (approved):
Looks pretty, good job!

a19nikbo:
If solving conflict, keep in mind that one of the drop down menus have been removed since you should only be able to have one cardinality per line.

a19hammi:
Button moved when I pressed lol

a19karwa (approved):
LGTM! Cardinality dropdown-menu now has the options: N, 1 and none, instead of many, one and none.

PR #10528, opened 2021-04-29
Swimlane layout change

a18danli:

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -868,42 +868,49 @@ p {
 /*Swimlane styling*/
 #statisticsSwimlanes {
   max-height: 300px;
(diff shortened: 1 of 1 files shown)


a18willi (changes requested):
image
image
As seen above the spacing is not consistent between different courses

b18phika (changes requested):
The spacing at the top and bottom of swimlanes seems a bit big, and is inconsistent between courses. See @a18willi review for pictures. Tested in both Chrome and Firefox.

a18willi (approved):
Looks better now! Everything seems to be consistent between courses.

b18phika (approved):
Looks good

PR #10533, opened 2021-04-29
Fix distance between fab-button and it's menu when menu is horizontal issue#10529

b18phika:


a18danli (approved):
Looks like the distance is better than before!

a19samke (approved):
I can clearly see the difference between what it looked like before, and what it looks like now-

b18phika:
We have now pushed a fix to the issue mentioned by @a20chrjo on discord.

PR #10536, opened 2021-04-29
Fix section problems

a18danli:


a19isada:
Make sure that all the functionallity still works as it should when testing.

b18phika:
It's with great sadness that I have to deliver the news that I do in fact not own a mac, and therefore cannot test this pull request.

a18willi (approved):
Nothing seems to be broken

b18phika (approved):

a19isada:
After testing on a mac, you can visit a course page and nothing seems to be broken.

Issue #10504, opened 2021-04-28
Does the PDUGGA ajax call need an ajax call in its success function


a18osken:
![image](https://user-images.githubusercontent.com/62884391/116407826-746c0700-a832-11eb-941e-72eaf1392a62.png)
Currently we have an ajax call that fetches the variant value and then in its success function sends it in the data.

Could we check if its possible to condense this into one call and solve this variant buisness serverside?

a19arvka:
Yes, this doesn't seem to just be a lazy solution. It is necessary for the variant to be calculated first, to then be used in the next query. 
However, I have an idea to segment the functionality of the methods _getVariantValue()_ and _handleLocalStorage()_ so that "pdugga" only uses one ajax success.

PR #10537, opened 2021-04-29
"Does the PDUGGA ajax call need an ajax call in its success function" #10504 (now without merge conflict)

a19arvka:
To review:

- Make sure hash is correctly stored in your database.
- Check that the flag "ishashindb" is set to it's correct value. To do this, force the value of the hash to be one that already exists in your database, the value should be forced inside the generateHash() method like this: hash = "alreadyExistingValue"; If everything works correctly the flag will notice this non-unique hash and generate a new one (part of _handleHash()_ method).
- Make sure you get a variant when doing a dugga. Change the ms value 2592000000 to 5000 (5 seconds) on row 1165. Then clear localstorage and spam-refresh the dugga, every 5 seconds a new variant of the dugga should be loaded.


a13zanal (approved):
Looking good, well done!

b19ponka (approved):
Good job, works fine. Maybe remove console logs

PR #10538, opened 2021-04-29
Fix so lines don't get selected behind elements.

a19chrel:
Made so we don't look for lines if an element has been clicked.


a19phith (approved):
Looks good!

PR #10541, opened 2021-04-29
fixed an issue where selecting a line doesnt not deselect any selected elements

a19karal:


a19frelu (approved):
Works tremendously well. Nice!

PR #10542, opened 2021-04-29
fixed an issue where selecting an element doesnt not deselect any selected lines 

a19karal:
fixed issue 10540 - Selecting an element without holding down ctrl should deselect any selected line(s)


a19frelu (approved):
This works just fine. Good job!

PR #10544, opened 2021-04-29
General function for checking key and ctrl

a19aleda:
Put the checking of key and ctrl into a general function reducing the complexity of reading


a19steis (approved):
Great fix!, the code is much more cleaner now

Issue #10426, opened 2021-04-26
Let teachers write feedback in LenaSYS and then send it to Canvas


a18osken:
It would be nice to allow the teacher to write feedback while viewing the hand-in in lenaSYS and then have that feedback get sent to Canvas and shown as the hand-in feedback there. We could explore this possibility. It is important to not save any data in LenaSYS however.

a16erilj:
How to test the code.
Visit showdugga.php as teachers (and students?) to check if feedback area is only shown to the teachers.

How to test commenting at the current state. 
- You need to have the secret in the correct place.
- Uncomment line 39 to 45.
- (Optional: change the comment at line 41 in testpage.php).
- Run testpage.php (The script will run each time the page is loaded).
- Login as Kristina Larsson at Canvas (It's possible to masquerade as Kristina Larsson through Admin also).
- Go to http://canvas.webug.his.se/courses/1/assignments/1?module_item_id=7 to check after the comment.

a16erilj:
This assignment have been slit up to multiple smaller parts #10568, #10569, #10571 and #10578

PR #10545, opened 2021-04-29
G1 2021 w18 #10426

a16erilj:
We have created 3 whole new files in a new folder called Canvas. These files are login.php, api.php and testpage.php. 

login.php handles the secret that is required for Lenasys to communicate with Canvas.

api.php have all the functions that handles the API calls. Most of the information about what the API calls contains and do should be in this handbook https://canvas.instructure.com/doc/api/index.html

testpage.php is the place where we have toyed with the communicating with Canvas and would most likely be educational for the next group that takes on communications between LenaSYS and Canvas.

We also added a text-field for feedback in showdugga.php that should only be shown to superuser aka teachers.

### DIFF
--- a/Canvas/api.php
+++ b/Canvas/api.php
@@ -1,0 +1,91 @@
+<?php
+    include 'login.php';
+
+    //  Canvas connection handlers
+    //  Creates the request in such a way that canvas can resive it
+
+    // Puts data on to canvas
+    function putData($URL, $data){
+        $ch = curl_init($URL);
+        $headers = array(
+          "Authorization: Bearer ". $GLOBALS['token'],
+        );
+
+        curl_setopt($ch, CURLOPT_HTTPHEADER, $headers);
+        curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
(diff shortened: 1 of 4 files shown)


a16erilj:
How to test the code.
Visit showdugga.php without login but also as teachers (and students?) to check if feedback area is only shown to the teachers.

How to test commenting at the current state.

- You need to have the secret in the correct place (The place it's looking for can be found in Canvas/login.php).
- Uncomment line 39 to 45 in Canvas/testpage.php.
- (Optional: change the comment "Mera mer" at line 41 in Canvas/testpage.php).
- Run Canvas/testpage.php (The script will run each time the page is loaded).
- Login as Kristina Larsson at Canvas (It's possible to masquerade as Kristina Larsson through Admin also).
- Go to http://canvas.webug.his.se/courses/1/assignments/1?module_item_id=7 to check after the comment.

a13zanal (changes requested):
Teacher feedback in lenasys visible and send comment via php to canvas works, well done! But if one is in the "Rapport inlämning" as a teacher, the submission content disappears.

a16erilj:
The issue have been split into smaller pieces migrate the solution in to the appropriate issue.

PR #10559, opened 2021-04-30
Updated cardinality

a19chrel:
There is now only one option while changeing propertys of an line. An line can only have one cardinality and a default has been added.


a19phith (approved):
Looks good and works as intended

PR #10560, opened 2021-04-30
Changed: gridSize and added automatic scaling to the varible

a19chrel:
Changed the gridsize to the height of entitys, and added so the size is automatic scales with the varible gridSize.


a19vikth (approved):
Looks good!

PR #10562, opened 2021-04-30
Crossing lines issue. Final update to HG version.

a19frelu:
This means that the lines DO NOT cross each other with the most commonly used browsers.


a19phith (approved):
good of good

PR #10564, opened 2021-04-30
Fixed zoom position when fully zoomed

b19jenda:
Added if statement so the extra step is no longer happening when fully zoomed in.


a19frelu (approved):
Good job. Works fine and I can not find any complications  with the rest of the program.

PR #10581, opened 2021-05-03
Added comments containing parameters and return.

a19chrel:
The diagram.js has been updated with the new comments.


a19hammi:
Put two/three reviewers to go through the comments to double check them! @a18nadma

a19steis (approved):
Could not find anything that seemed wrong, maybe a19matma will find something i missed. But seems like it is correct.

a19matma (approved):
LGTM!

PR #10585, opened 2021-05-03
Added outline

a19steis:
Made the selected lines display a shadow so we can more clearly see which lines are selected.

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -320,10 +320,6 @@
     margin-left: 20px;
 }
 
-.outline {
-    filter: drop-shadow(3px 1px 3px black);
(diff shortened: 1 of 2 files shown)


a19karwa (commented):
It does make the marking of lines more clear for me, but I can not tell if it will make a difference 
for those who are colorblind, perhaps someone that is colorblind should test this aswell?

a18nadma (changes requested):
A suggestion we recieved would be to have a dot or a small square in the middle of the line as an indication that the line is being selected.

a19vikth (approved):
Looks good :)

a19hammi:
Closed until later

PR #10587, opened 2021-05-03
Update diagram.js

a19karwa:
Removed one line that was redundant and moved up a str to make it run before we write into innerHTML.


a19frelu (approved):
Awesome! 
Still does the work just fine without the unnecessary redundancy.

a19hammi:
LGTM!

Issue #10314, opened 2021-04-22
Find a way to track local changes in a group assignment


a18osken:
We should explore how easy it is to track local user activity in a hypothetical group assignment. So that later on we may synchronize several users if for example some changes have been made locally.

This is very exploratory and we must strike a balance between keeping a nice synchronous flow in the assignment while also not making the machine catch fire.

a19simma:
🔥

a19simma:
![image](https://user-images.githubusercontent.com/81557282/115845320-02a25080-a421-11eb-9441-b49d2807fac9.png)

PR #10591, opened 2021-05-03
G1 2021 w18 #10314

a19taral:


a19simma:
Press a butt in a dugga and it should console log this
image

a19simma:
issue #10314

e15marol (commented):
Local changes are tracked, but there's also a function added (addGroupButton in dugga.js line 894) that's not been used, what is its purpose? Also when saving, and then making changes to a dugga, it sometimes console logs that changes have not been made even though they have been made and are displayed through the counter.

b18jacha:
Can we test in any dugga assignment or does it need to be a group assignment?

b18jacha (commented):
Can we test in any dugga assignment or does it need to be a group assignment?

a19taral:
the add Group button is just for the testing convenience and it can be removed, and you can test for any assignment type.

a19simma:
> Also when saving, and then making changes to a dugga, it sometimes console logs that changes have not been made even though they have been made and are displayed through the counter.

It tracks 'clicks', not actual changes. Just when there's an interaction from the user in any way. Also I don't know how to reproduce this error.

a19simma:
Hopefully fixed the issue @e15marol @b18jacha pull the newest branch and check 
image

e15marol (approved):
It works now, although the counter is console logged twice now, which is a non-issue. Approved, good job! We'll have to clean up the console logs in the future as the console is getting clogged up. But for testing purposes for future functionality changes this can remain for now.

b18jacha (approved):
Looks good!

e15marol (changes requested):
Just discovered for example in biträkningsdugga 1, that if you have two or more of the bits pressed, and you save, the counter is incremented by 1 everytime you save, and while you have only 1 it displays the correct amount of "clicks" whenever you save, while none of the bits are pressed, the counter does not get displayed in the console. Not sure if this will cause issues for the future, but it should be looked over if its handled appropiately.

b18jacha (changes requested):
Also I just found that when clicking on either one of the two big "0" buttons on bitdugga 1 it counts as a change even tho we haven't changed it to anything.

a19simma:
> Just discovered for example in biträkningsdugga 1, that if you have two or more of the bits pressed, and you save, the counter is incremented by 1 everytime you save, and while you have only 1 it displays the correct amount of "clicks" whenever you save, while none of the bits are pressed, the counter does not get displayed in the console. Not sure if this will cause issues for the future, but it should be looked over if its handled appropiately.

When you click save, for whatever reason it generates every button click again, which is kinda weird. Not sure if we should fix this or leave it for another issue as the number itself isn't really important.

e15marol (approved):
Works for experimental use of tracking clicks!

b18jacha (approved):
Great start of a bigger issue.

Issue #10474, opened 2021-04-27
Returning to assignment hand-ins in Canvas


a18osken:
It is currently unclear if we are to be able to return to our assignment hand-ins in the canvas iframe or if you'd have to use the hash to be redirected to dugga instead.

HGustavs:
Im thinking that if one returns it could remember the hash in local storage together with the assignment variant?

HGustavs:
When deeper integration with canvas has been achieved I imagine that this can be handled by canvas sending some kind of identifier.

PR #10597, opened 2021-05-03
G1 2021 w18 #10474 v2

e15marol:
Made it so that you get the same dugga by storing the initial hash and reloading the dugga using the same hash from localstorage. To check, simply clear localstorage, enter the dugga assignment on canvas, and take notice of the dugga key with the hash included and save the dugga. Whereupon you will receive the hash which you can then again load the saved dugga. If you have issues loading the iframe, simply go into the HTML editor and change the filepath to your local address for biträkningsdugga.

The issue should work on both the canvas installation in the assignment and on your local installation of lenasys. Simply enter http://canvas.webug.his.se/courses/1/assignments/5 to go to the live version (which loads the dugga from your local installation).

### DIFF
--- a/Shared/dugga.js
+++ b/Shared/dugga.js
@@ -63,19 +63,26 @@ function getHash(){
 function setHash(h){
 	// Check if hash is unknown
 	if(h == "UNK"){
-		console.log("This is the generated hash: " + hash);
-		hash = generateHash();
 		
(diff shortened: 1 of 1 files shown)


b18jacha:
@HGustavs Right now it probably works as expected. However in canvas we don't have a way to go to a specific hash if we change computer. Should that be possible?

a19kevkh (approved):
Looks good.

a19frebo (approved):
Works fine!

PR #10598, opened 2021-05-03
"Fixed" Tool-tip for the option-panel

a19phith:
Fixed by adding the tool-tip panel to the option panel, both to the fab button and the panel button


a19hammi:
using old pull request to get comments back

PR #10602, opened 2021-05-03
G2 2021 w18 issue#10558 -> changed lightbulb icons

a19kimgr:
We were tasked with changing the lightbulb icons, there are multiple different icons that could  be used, look in the comments for alternatives and make a change request for the best looking one (unless the current one is best).

1 (current)
Skärmklipp

2
option1

3
option2

4
option4


a20chrjo (approved):
Nr. 1 looks the best in my opinion and the Nr. 4 comes a close second. The 4th is also a bit more visible on the site, but the file icon can be misinterpreted as a file you should upload instead of a test that should be taken (dugga).

b19alest (approved):
I agree with previous. Number 1 looks best and fits best what a dugga is.

b18phika:
Looks good

PR #10603, opened 2021-05-03
No longer possible to access student data through the accessed page. Issue#10547

a18willi:
Added a check to see if the user has a write access to the course before sending any any user data.

Co-author: @a19isada


a19kimgr (approved):
Works. student data no longer visible.

a19klafr (approved):
Seems to work as explained. No Student data is showing

Issue #10592, opened 2021-05-03
Receiptbox - Remove the receipt numbers when saving a dugga


e15marol:
We no longer need the generated receipt numbers, so these should be removed.

![Skärmklipp](https://user-images.githubusercontent.com/18677403/116863973-638f0d00-ac07-11eb-850f-eb0e05f65b7d.PNG)

HGustavs:
The exclamation point "undefined" is for when the submission date has not been exceeded? We should probably remove that part if we are submitting early?

PR #10605, opened 2021-05-03
Receiptbox - Remove the receipt numbers when saving a dugga #10592

a19arvka:
To test this change:

- Make sure the receipt dialog popup appears whenever you try to save a dugga/submission.
- Also check if it works when you use a URL with hash & password access.


a19rebsa (approved):
It works fine!

a13zanal (approved):
Looks good!

PR #10608, opened 2021-05-03
Moved the bottom buttons

a19danla:
Moved the bottom fab-button and "back to top" button 40px to the left so they wont cover the edit-buttons.

Co-authored-by: b19wilfo


a19klafr (approved):
Looks good. All important buttons can be reached in the screen, fab button donesn't cover as explained an requested.

a19kimgr (approved):
Looks good!

Issue #10574, opened 2021-05-03
Create a course that with assignments that can be used for testing (Group 1)


a16erilj:
We need a course that is linked to our Group 1 LenaSYS on the server for testing the integration between LenaSYS and Canvas.
- Create the course Webbutveckling - datorgrafik IT118G with a tag group 1.
- Create all the assignment that are in LenaSYS Webbutveckling - datorgrafik IT118G on our Canvas Webbutveckling - datorgrafik IT118G.
- Iframe the LenaSYS assignment from our [LenaSYS ](http://project-g1.webug.his.se) to Canvas assignment.
- It should be text submitted so that the student can paste the URL/hash and Password.

a19kevkh:
Created course G1- Webbutveckling - datorgrafik.

a19frebo:
Created all assignments for the course, and Iframed all assignments along with a text submission for each assignment.

PR #10611, opened 2021-05-03
G1 2021 #10574

a19frebo:
Created course G1- Webbutveckling - datorgrafik.
Created all assignments for the course, and Iframed all assignments along with a text submission for each assignment.


B18mohah:
image
images gets lost in the transition from Lenasys to Canvas.
image

a18osken (commented):
It seems that using https in the iframe link prompts a certificate check which our hosting does not have, simply removing the https should work hopefully then we won't get prompted

a19frebo:
We could not install an SSL certificate so we changed all the IFrame links to http:// instead of https://, and now the images are shown correctly aswell.

a18osken (approved):
All the previous issues have been solved, you's done it sirs

B18mohah (approved):

PR #10612, opened 2021-05-03
Update style.css (Issue #10265)

b19alest:
Changed from hardcoded colorcode to predefined variables.

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -856,7 +856,7 @@ p {
 #statisticsSwimlanes {
   display: flex;
   overflow:auto;
-  border: 2px solid darkgray;
+  border: 2px solid darkgray; 
   background-color: var(--color-sectioned-table-lo);
 }
 
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -831,7 +831,7 @@ p {
 
 #Sectionlist .section {
   color: var(--color-text-header);
-  background-color: var(--color-primary-light);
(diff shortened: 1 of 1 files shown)


a19isada (changes requested):
There is still contrast issues. Lighthouse test does not pass.

image

a18willi (changes requested):
The contrast problem described in the linked issue is still present in branch G2-2021-W18-ISSUE#10265

b19alest:
@a18willi @19isada Changed more tags because all types of duggas/moments etc didn't have the same tag that changed backgroundcolor/textcolor. Know all types are changed so all different duggas/moments/tests should have the same backgroundcolor/textcolor. Also a new item had opacity drawn down and caused contrast problems, so I took away that aswell (.hidden-tag). 

@a19anjst @b18phika but we still maybe want a new item to "stand out" from other items, can take up this on the next customer meeting maybe?

a19anjst:
Will write it down as something to consider for next customer meeting

b18phika:
@b18alest We'll bring that up on the next customer meeting. However, we think that it'd be a better idea to create a new issue for that, so I'll re-request reviews for this issue.

a19isada (approved):
Text is much clearer now. Lighthouse also passes.

a18willi (approved):
Looks great! Lighthouse isnt complainging about contrast issues anymore.

Issue #10435, opened 2021-04-27
Move highscore functionality to localstorage


a18osken:
Right now highscore is in a spot where it is impossible to make a highscore list without giving away personal information. We could however move the individual score to localstorage so that a user can view their own score on an assignment

HGustavs:
Yes this is a good idea to test.

b19ponka:
If the Hight Score is only connected to a submission through a hash, and is only compared to other High Score with the same hash (same persons score is listed). Is there still a need to store it locally? Furthermore, if the Hight Score is only displayed when a submission is passed and you can’t upload a new submission if it’s already passed. What is the need for a personal Hight Score? Some more detail about how this new Hight Score functionality is supposed to work, and for what purpose, would be much appreciated.

a19rebsa:
The information that we got from the meeting with Henrik was that it is pretty useless to have a public highscore list with the hashes, but we could fix a highscore-list with personal highscore. That would make more sense.

PR #10613, opened 2021-05-03
Score in localstorage - #10435

a13zanal:
Saves score variable for each dugga in an array in localstorage


a19rebsa:
To test this pull request you open the inspector in the browser, go to application and check your local storage. To each dugga you should see that the score is stored in an array.

e15marol (commented):
It works, however score seems arbitrary and we can't quite figure out what the score represents. Also there are some duggas that don't generate any scores, for instance 3D dugga in the demo-course. If you constantly save a dugga over and over, it generates a new highscore (different value from the previous highscore) and adds it to the array, is this intentional?

b18jacha (approved):
**It works.** 

_Would be cool to have the date saved with every highscore aswell!_

e15marol (approved):
So the highscore goes either by counting clicks or time taken and records it as a highscore value, which is what the numbers represent. The solution works, although for the future, the highscore might need a change as it doesn't serve a purpose that we can see currently. Good job!

PR #10614, opened 2021-05-03
Hamburger menu in sectioned(courses) now show for both supersusers and teachers Issue#10499v2

a18willi:
The hamburger menu now shows for both teachers and superusers while viewing a course though the sectioneditor.

Co-author: @a19isada


b19alest (approved):
Works, but hamburger menu does just work for teachers that are set as teacher on that course. So just being set as teacher on one course does not give you access to the hamburger menu for all courses. Question is if we want it this way?

a20chrjo (approved):
Works in Chrome, Edge, Firefox and Opera.

What works:
Web view & mobile view
Teacher for their specific course
Super user

PR #10620, opened 2021-05-03
Fix: Only the diagram is magnified when zooming

a19matma:
Added preventdefault() to stop the webpage from getting magnified when zooming in the diagram.
Fixed so this both works when zooming with ctrl+mwheel as well as with ctrl+"+" and ctrl+"-" on the numpad.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -894,8 +894,14 @@ document.addEventListener('keydown', function (e)
             showdata();
         }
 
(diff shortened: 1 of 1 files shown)


a19matma:
Forgot to add "-W18-" in the branch when I created it (If anyone wonders why it's missing).

a19frelu (approved):
Great stuff right here. Not an error as far as the eye can see. 
Good job!

a19hammi:
LGTM!

PR #10622, opened 2021-05-04
Updated cardenality

a19chrel:
The text is now calculated from the coordinate on the line instead of the end/start point of the line. The offset from the line is also taken into account depending of the number of connected lines to the same side of an entity.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -2728,10 +2728,19 @@ function drawLine(line, targetGhost = false)
         var posX, posY;
         var distance = Math.sqrt(Math.pow((tx - fx), 2) + Math.pow((ty - fy), 2));
 
+        // Used to tweak the cardinality position when the line gets very short.
+        var tweakOffset = 0.30; 
+
         if(findEntityFromLine(line) == -1){
-            // Set position on line for the given offset
-            posX = fx + (offsetOnLine * (tx - fx) / distance);
-            posY = fy + (offsetOnLine * (ty - fy) / distance);
+            if(offsetOnLine > distance){
+                posX = fx + (offsetOnLine * (tx - fx) / distance) * tweakOffset;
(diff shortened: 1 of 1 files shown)


a19phith (approved):
Seems to be working, good job

a19hammi (changes requested):
Not sure if this is what should be the expected behavior. I'd love if you took a second look at this. It's not a high priority issue but right now we lack much else to do.

https://user-images.githubusercontent.com/81571224/116985694-20976d00-accd-11eb-88d5-f33ff7002088.mp4

a19hammi (changes requested):
Så här ser det ut på min lokalt. Skall nog inte vara så.

https://user-images.githubusercontent.com/81571224/117012363-2a31cc80-acef-11eb-8093-3f067f6f7934.mp4

a19matma:
Will investigate this further, but yes it might be difficult to get further since what's considered acceptable behaviour is very subjective.

a19chrel:
> 
> 
> Så här ser det ut på min lokalt. Skall nog inte vara så.
> 28e820f2d19e9719db7ce54e20b5c5f6.mp4

Seems like the the line gets connected to right side of the entity instead. I would say this is an problem with the calculation of lines instead of the cardinality-draw. 
bild
But maybe the offset here could be better.

a19hammi:
> 
> 
> > Så här ser det ut på min lokalt. Skall nog inte vara så.
> > 28e820f2d19e9719db7ce54e20b5c5f6.mp4
> 
> Seems like the the line gets connected to right side of the entity instead. I would say this is an problem with the calculation of lines instead of the cardinality-draw.
> bild
> But maybe the offset here could be better.

Yep, we looked at it the other day or so. The line connects to left/right side the very PIXEL that the Refer-top intersects the Car bottom.

a19hammi:
We'll merge this for now and try finding a better solution later.

PR #10631, opened 2021-05-04
Made it more clear which lines are selected

a19nikbo:
Added a square to the middle of the selected lines and included them in the selection box calculation

Pair-programmer: @a19steis


a19karal (approved):
funkar som det ska :)

a19hammi:
As mentioned. Double lines are weird but that's because the logic for these lines are a bit broken as of now. Will be fixed in future issue!

PR #10634, opened 2021-05-04
Fix: backlayer-lineIDs value copy.

a19matma:
The IDs of the lines in the svgbacklayer should now remain unmodified after the determineLineSelect() is called.


a19steis (approved):
Seems to work fine, well done.

a19hammi:
LGTM!

Issue #10606, opened 2021-05-03
Receiptbox - The exclamation point is "undefined" if we enter a hash & password URL and press Save


a19arvka:
This Issue was found from Issue #10592, which I have just solved. 
![undefined](https://user-images.githubusercontent.com/81560260/116873615-49a9f600-ac18-11eb-89d9-aefbd46d5fd4.png)

PR #10635, opened 2021-05-04
Receiptbox - The exclamation point is "undefined" if we enter a hash & password URL and press Save #10606

a19arvka:
The value "deadline" was missing from the URL when using the hash & password URL, so I simply added the deadline variable to the URL redirection. In addition, if statement _if(comment == undefined)_ used incorrect syntax.

**To test this:**
- Enter a dugga, press Save. Does it tell you if you have missed the deadline or not? Is the message correct?
- Next, use the hash & password URL to enter the dugga again, press Save once more. Is the message still correct?


a19kevkh (approved):
GGWP

a19frebo (approved):
Looks good!

PR #10636, opened 2021-05-04
Changed the position divider

a19danla:
Changed the position divider to a lower value and removed the constant to make it scale better for longer lists.

Should work fine for 70+ items.

There is still a "bugg" where this function sometimes struggle when there are alot of newly added items. It fixes itself if you leave the site and rejoins later.

### DIFF
--- a/DuggaSys/contribution.js
+++ b/DuggaSys/contribution.js
@@ -668,6 +668,7 @@ function createTimeSheetTable(data) {
   if(tabledata.tblbody != null || tabledata.tblbody != undefined) {
     myTable.renderTable();
   }
+}
 
 function renderCell(col, celldata, cellid) {
   var str = "UNK";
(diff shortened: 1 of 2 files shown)


a19danla:
Would be nice to have more testers to see how it behaves for them. Seems to be inconsistent for us.

a19klafr (changes requested):
New items currently gets added somewhere a little bit below the showing screen, so when new item is added you don't see it until you scroll down. Would be nice if this gets changed so you see the added item right away.

a19danla:
Thats good to know it acts like that for you as well. We currently have some problem where it sometimes act how its supposed to and sometimes not. We're currently looking into this but wanted more people to test this.

a19kimgr (changes requested):
The new items are added outside the screen and its hard to see that they are added

a19klafr:
Might want to look at where the pull request is going.. currently saying master instead of G2-2021-W18

Issue #10568, opened 2021-05-03
Create feedback area for teachers in their assignments at showdugga.php


a16erilj:
Currently there are no area where teachers can send send feedback to the student whom have done the assignment. So create an feedback area in showdugga.php that only the teachers may access and also looks good on Desktop and Mobile.

Sub issue of #10426

a19rebsa:
How is this supposed to look like? Should this be in each dugga? Before the feedback was located in resulted.

a19rebsa:
Also an argument for not placing the feedback area in each handed in assignment is that this feels like an extra step for teachers when they could give feedback more easily to many assignments via resulted

a16erilj:
If you are talking about giving the same feedback to multiple  students at the same time then placing a text area for feedback resulted could be a good solution. 

Since we are talking about in individual submission so would it be nice to see the submission when commenting. So a feedback area in showdugga.php would be nice.

a19rebsa:
Ok, so for this issue we only need to the the actual feedback area for the hand-ins and not the actual functionality for the feedback to be sent to canvas right?

PR #10638, opened 2021-05-04
Create feedback area for teachers #10568

a19rebsa:
Created a feedback area that is only visible for teachers when ther are logged in and on a handed in assignment (hash is not unknown).


a13zanal:
For testing:
Save dugga, go to the hash url and log in as a teacher and the feedback box should appear.

a19gusba (approved):
works

a17alfja (approved):
works, even looks nice on mobile

PR #10641, opened 2021-05-04
The announcement icon in the navbar disappears when the width of the …

a19samke:
…viewport is less than 861px

@a18danli 

Announcement Icon is no longer  hiden!


b19alest (approved):
Works, announcement icon is now visible all the time.

a20chrjo (approved):
Works in Chrome, Edge, Firefox and Opera.

What works:
Web view & mobile view

Issue #10642, opened 2021-05-04
Replace the popup instructions in linjedugga 1 & 2


a16erilj:
Currently so is the instructions displayed in a type of popup window. Replace this with a standard instructions window.
 
![bild](https://user-images.githubusercontent.com/81558891/117005272-56494f80-ace7-11eb-8a1a-eb1145e934f2.png)

PR #10646, opened 2021-05-04
Replace popup instructions in linjeduggor #10642

a19rebsa:
We've replaced the popup-instructions with an instructions container in linjedugga 1 and 2


B18mohah (approved):
Nice

a18osken (approved):
That's a sexy accordion. This is a good assignment body standard for the youth to strive for.

Issue #10629, opened 2021-05-04
Change the title of assignments


a16erilj:
Currently so is only the mention dugga shown in the header of assignments in showdugga.php, change it so that dugga is replaced by assignment name.

![bild](https://user-images.githubusercontent.com/81558891/116983978-05c3f900-accb-11eb-9286-9bd6197f83b1.png)

PR #10653, opened 2021-05-04
Change the title of assignments #10629

a19arvka:
To test:

- Make sure every Dugga have their corresponding title once entering the dugga.
- Force the variable duggaTitle to have a faulty value, for example: _duggaTitle = "";_ or _duggaTitle = 0;_ in dugga.js, row 2009. The dugga title should then be "untitled Dugga".


a13zanal (approved):
The title has changed for both student and teacher, good!
But for teachers, the title of some of the assignments disappears.

PR #10654, opened 2021-05-04
Double lines are now included in selection box

a19nikbo:
Added check for if line is a double line and appended "-1" and "-2" to the IDs that are added to tempLines.

Co-Authored-By: a19frelu <62895155+a19frelu@users.noreply.github.com>


a19matma (approved):
LGTM! Works as intended. Original ID-values remain unaltered as well, gj!

PR #10663, opened 2021-05-05
Fix bug where ghostelement got removed when pressing esc

a19aleda:
Moved the assignment of null to ghostelement variable to only trigger if there is nothing in the context or contextLine array.


a19matma (approved):
LGTM! Bug fixed.

PR #10666, opened 2021-05-05
Elements can now be locked in place

a19steis:
If atleast one element is selected that is unlocked, then all those elements will be locked.
If all elements are locked then the elements will be unlocked.

The option is displayed when we generate the properties for one or more elements.


a19phith (approved):
Looks good, works :3

Issue #10650, opened 2021-05-04
Teachers do not get a variant when signed in


a16erilj:
Seems like teachers do not get any variants when signed in and visiting a dugga. So they are unable to get view the assignment in it's whole. Fix so that teachers are able to get an eagle-eye over their assignments. 

![bild](https://user-images.githubusercontent.com/81558891/117010823-7d0a8480-aced-11eb-8f0b-58003d6a36c1.png)

a13zanal:
Maybe found another issue when we solved this issue. We get an error alert when we enter a saved url as a teacher. @a16erilj ?

![](https://cdn.discordapp.com/attachments/837266494047780877/839416969559670794/unknown.png)

PR #10670, opened 2021-05-05
Teacher gets a variant! #10650

a13zanal:
Now a teacher gets a variant!


a18osken (approved):
Teachers now get assigned localStorage in the same way a student does, good work!

B18mohah (approved):
Nice!

Issue #10662, opened 2021-05-05
When entry in the userAnswer-table is removed, it also needs to be removed from groupdugga-table.


b19ponka:
It was discovered that entries in groupdugga-table is not being removed properly.

PR #10671, opened 2021-05-05
Added trigger

a17alfja:
Added triggers that remove entries from groupdugga whenever a matching hash entry is removed from userAnswer.


a17alfja:
You will have to run the code associated with issue #10662 in alterTables for it to work.

a19frebo (approved):
Works! When you delete a row in useranswer, the row with the corresponding hash is deleted in the groupdugga table

a19kevkh (approved):
Looks good, works fine.

Issue #10627, opened 2021-05-04
No confirmation message when saving a dugga additional times


a16erilj:
Current when going to an submission trough hash so does it allow the student to save an additional time without promoting for a confirmation about saving the new submission.

Create a confirmation window/popup that allows the student to confirm/deny saving the submission.

PR #10673, opened 2021-05-05
G1 2021 w18 #10627

e15marol:
To test this issue, simply enter the datorgrafik course and save any of the "duggor" and then save again, whereafter you will receive a popup that asks if you are sure you want to save again.

Co-Authored-By: b18jacha <62884577+b18jacha@users.noreply.github.com>


a19rebsa (approved):
It works as it should

a13zanal (approved):
Works fine!

PR #10686, opened 2021-05-05
"Delete marked items" button now hides marked items from students instead. Issue#10280

a18willi:
Can now hide multiple items at the same time using the "-" button.

Co-author: @a19isada

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -14,7 +14,7 @@ var resave = false;
 var versnme = "UNKz";
 var versnr;
 var motd;
-var deleteItemList = [];
+var hideItemList = [];
 var hasDuggs = false;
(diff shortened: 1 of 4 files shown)


a18willi:
Mentioned button
Skärmklipp1

b19alest (changes requested):
Seems to work but the popup is missleading as it asks if you want to delete the item.

a20chrjo (changes requested):
The code works. But the pop up that appears, ask for confirmation to delete instead of to hide.

b19alest (approved):
Perfect!

a20chrjo (approved):
Tested in Chrome, Edge, Firefox and Opera, in both browser mode and mobile mode.

Issue #10676, opened 2021-05-05
Maybe we should get a warning when clicking the reset button


b18jacha:
Maybe we should get a warning if we accidentally click the reset button on a dugga page?

a16erilj:
Great idea, that's something that we should do. 
I would like a solution similar to #10627

PR #10696, opened 2021-05-05
Added warnings for when clicking on reset buttons #10676

a19rebsa:
We replaced the alerts (that did not work before) with confirms that displays a warning for when clicking on reset-button in duggor. If you choose ok then everything resets and if cancel then nothing happens.


a19gusba (approved):
works

a17alfja (approved):
Works, but a solution that doesn't change every template file would have been nice. Though that is a different issue (like making templates class based).

PR #10700, opened 2021-05-05
Added div with background on the ruler

b19jenda:
Now there is a grey background in the corner of the ruler

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -190,6 +190,15 @@
     height:100%;
 }
 
+#rulerCorner {
+    position: absolute;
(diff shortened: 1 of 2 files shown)


a19phith (changes requested):
I dont see any changes, might add a screenshot to the fix because i dont see any differents

b19jenda:
ruler1
Here is how the ruler looked before in the corner.

ruler2
And here is how it should look now in the corner of the ruler, I supposed that the div would have the same grey color as the rest of the ruler.

a19phith (approved):
Now i see it! looks good

PR #10702, opened 2021-05-06
Diagram: Reduced history data size

a19chrel:
Now does the stateMachine only store the change and no flags. Also algorithms has been added to remove default values for changes. The  stepBack functioallity has also be changed so when a stepBack in the history is made, the current data is removed and replaced by looping through the history.


a19steis (approved):
Looks good, functionality seems unchanged. Good work!

a19hammi:
Nice!

PR #10703, opened 2021-05-06
"Fixed" There should be a way to see if a element is locked (visually)

a19phith:
Added a svg element to the locked function, a padlock icon will be placed on the entity if the entity is locked in place


b19jenda (approved):
Looks good!

PR #10705, opened 2021-05-06
Added btn-disable class

a17alfja:
btn-disable class prevents click and alters color of button.


a19rebsa (changes requested):
I think that the buttons should only be disabled on already handed in assignments, and not all the time for teachers

a13zanal (changes requested):
The save and reset buttons are disabled for all assignments as a teacher, maybe should only be for submitted assignments?

a13zanal (approved):
Looks good!

a19rebsa (approved):
Works great

PR #10711, opened 2021-05-06
Fix selected elements are ontop of other elements

a19aleda:
Change the z index for elements in context to be 1 and also changed the z index of ruler bars to 2, so that the selected elements doesnt overlap the ruler


a19karwa (approved):
This works, however only when less than 4 elements are stacked ontop of each other. It's more than necessary for the diagram, so we won't have to keep changing that Z-index and make our life harder than it should be.

PR #10716, opened 2021-05-06
Diagram: Support for stepForward in the history

a19chrel:
The stateMachine now support stepBack(), also some changes to the functions has been made for reusablilty.


a19vikth (approved):
Looks good!

Issue #10567, opened 2021-05-03
Force assignment in the same grouping to have unique names


a16erilj:
Currently it's possible to have the same name on assignments in the same group. If the teacher / logged in user tries to rename an assignment to the same name as another assignment throw a visual error and request a different name.

A sub issue of this is that new assignment are given the same name so try to fix this by inserting a increment to their name.
![LenaSYS](https://user-images.githubusercontent.com/81558891/116846316-5dd5ff00-abe8-11eb-950d-1198af88d5ac.png)

PR #10717, opened 2021-05-06
8 commits >:(

a19simma:
#10567


a19simma:
image
Check that you can't rename an element to a duplicate. Also so you can save without changing the name (i.e. save as itself since that's technically a duplicate)

a19taral:
image

a19gusba (approved):
works

a17alfja (approved):
works

Issue #10600, opened 2021-05-03
Individual delete button for each file uploaded in rapport inlämmning at showdugga.php


a16erilj:
There is no way to delete a individual file in "rapport inlämmning" showdugga.php. Add some buttons with the function to delete the connected file. 

This is includes individual files in zip uploads.

b19ponka:
Also look at filerecive_dugga.php

b19ponka:
Works great but, when presented to the customer, this was not a feature that they wanted

PR #10733, opened 2021-05-06
Individual delete button for each file uploaded in rapport inlämmning at showdugga.php G1 2021 #10600

a19kevkh:
Found another issue after completing this one.

That is that, if we upload a dugga, then save it and then delete it. We will be prompted for the password.


a13zanal (approved):
Works as it should, good job!

a19rebsa (approved):
Works great!

b19ponka:
Works great but, when presented to the customer, this was not a feature that they wanted

PR #10743, opened 2021-05-06
Clear contextLines at selectionbox mouse up

a19aleda:
Clear contextLines at selectionbox mouse up

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1307,6 +1307,7 @@ function mouseMode_onMouseUp(event)
 
         case mouseModes.BOX_SELECTION:
             boxSelect_End();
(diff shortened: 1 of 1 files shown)


DuggaSys/diagram.js
```diff
@@ -1307,6 +1307,7 @@ function mouseMode_onMouseUp(event)
 
         case mouseModes.BOX_SELECTION:
             boxSelect_End();
+            contextLine = [];
```
a19hammi, line 1310: Use the function call that clear lines instead!

a19hammi (changes requested):

a19karal (approved):
Funkar som det ska :)

Issue #10720, opened 2021-05-06
Login: After loading a hash url


e15marol:
When you load a hash url to login for a saved dugga, if you can't or wont enter the password and simply click away the popup, you can not from there login using the login button in the topright corner.
![Skärmklipp1](https://user-images.githubusercontent.com/18677403/117275981-e362e480-ae5e-11eb-8d29-75bf5b5e277a.PNG)
![Skärmklipp2](https://user-images.githubusercontent.com/18677403/117275991-e65dd500-ae5e-11eb-998e-38132f062ac4.PNG)

a16erilj:
Seems like something that should be checked and if it can be replicated so should it be fixed.

b18jacha:
Since the popup login menu will be removed from canvas we are trying to redirect to original dugga page instead.

PR #10744, opened 2021-05-06
G1 2021 w19 #10720

e15marol:
Since assignments are only going to be accessed through canvas in the future and not through direct urls, there should be no issues for the fix we did, in that it stores the url in local storage when you open the assignment and redirects to it if you cannot enter the correct hash or wish not to (e.g. closing the popup window).

To test this, either access a dugga through your local installation of lenasys or through the following link: http://canvas.webug.his.se/courses/1/assignments/5 and use the admin account.

If you're having issues loading the iframes, simply open up the HTML editor inside canvas and change the address to where you have your lenasys installation folder.

### DIFF
--- a/Shared/dugga.js
+++ b/Shared/dugga.js
@@ -1535,6 +1535,7 @@ function setupLoginLogoutButton(isLoggedIn){
 function showLoadDuggaPopup()
 {
 	$("#loadDuggaBox").css("display","flex");
+	localStorage.setItem("ls-redirect-last-url", document.URL);
 }
 
 function hideLoadDuggaPopup()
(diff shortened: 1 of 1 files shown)


a19kevkh (approved):
Excellent.

a19frebo (approved):
Nice solution! 👍 🚀

b18jacha:
@b19ponka @a16erilj  We just discovered a new issue with this.
- When trying to load a new hash it will always try to redirect us to previous visited dugga, no matter if we enter correct password or pressing the X button to close the popup. We need a new issue for this.

b18jacha:
> @b19ponka @a16erilj We just discovered a new issue with this.
> 
> * When trying to load a new hash it will always try to redirect us to previous visited dugga, no matter if we enter correct password or pressing the X button to close the popup. We need a new issue for this.

Created a new issue https://github.com/HGustavs/LenaSYS/issues/10788

PR #10750, opened 2021-05-06
Keyboard shortcut för zoomfactor

b19phier:
Lagt till keyboard shortcut för att komma tillbaka till zoomfactor = 1.0


a19nikbo (changes requested):
This probably won't work on Mac since it says "metal" instead of "meta" in the keybinds enum.

Just resetting the zoom factor to 1 means that it's very inconsistent where the "camera" actually goes. It should probably handled similarly to how it's handled in the zoomin() function, so that is zooms towards origo and doesn't overshoot.

It would also be good to add a @description comment to the function as with the other zoom functions.

a19nikbo (approved):
Feels much better now 👍

PR #10751, opened 2021-05-06
Fixed error with Ctrl+z

a19matma:


a19nikbo (changes requested):
When deleting all the elements in this order and pressing ctrl + z

changerequestexample

I get this error when moving the cursor

image

a19nikbo:
Also, ctrl + z after having deleted an element while moving it produces this error:

image

a19nikbo:
If you reproduce the bug in the previous comment, pressing ctrl + z while moving an element makes the selection box stay in place

image

And it gives this error

image

a19nikbo:
I'm having difficulty reproducing this error, so I might've been in the main branch when I found the bug.

I do however get this error:

issue_example

The error comes when pressing ctrl + z.

a19chrel:
> 
> 
> When deleting all the elements in this order and pressing ctrl + z
> 
> changerequestexample
> 
> I get this error when moving the cursor
> 
> image

I guess that this was indeed in the main branch, the error that was fixed in the first place was similar.

a19chrel:
> 
> 
> I'm having difficulty reproducing the error in the review, so I might've been in the main branch when I found the bug.
> 
> I do however get this error:
> 
> issue_example
> 
> The error comes when pressing ctrl + z.
This error has now been fixed, also some extra comments was also added.

a19nikbo (approved):
I can't find any more errors regarding ctrl + z. 
Well done!

PR #10773, opened 2021-05-06
Added buttons and keybinds to the functionality

a19steis:
Added buttons with SVG images to the step forward/backward functionality.

Also added tooltips to the buttons which display the keybinds.


b19jenda (changes requested):
The buttons works once but after that they stop working and errors occurs everytime you click on the buttons or using the keybinds.

b19jenda (approved):
Looks good, problem with errors is coming from another issue

PR #10774, opened 2021-05-07
included the function's path in showDugga.php

a19taral:
To test: enter to any assignment as super user, and check if the error message still occurs in the console window.


e15marol (approved):
Works, good job!

b18jacha (approved):
I approve **but**:

- Maybe the _function noUploadForTeacher()_ should be in dugga.js instead?

a19taral:
> I approve **but**:
> 
> * Maybe the _function noUploadForTeacher()_ should be in dugga.js instead?

yes but we are just intressted in Rapport inlämning assignments witch the file generic_dugga_file_receive.js and noUploadForTeacher() are created for.

PR #10778, opened 2021-05-07
Added append to history within set time limit.

a19chrel:
History entrys does now append if it is the same elements and within the timelimit.


a19phith (approved):
LGTM

a19hammi:
Need review again after merge

a19phith (approved):
new review after merge :)

PR #10780, opened 2021-05-07
Fix bug where elements could be placed with other mouse buttons than left click

a19aleda:
Fix bug where elements could be placed with other mouse buttons than left click


a19phith (approved):
lgtm!

PR #10782, opened 2021-05-07
Change back name of element at escape pressed

a19aleda:
Change back the name of a element at escape pressed, if you have not saved the value

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1122,6 +1122,14 @@ document.addEventListener('keyup', function (e)
                 displayMessage(messageTypes.SUCCESS, `Clipboard cleared.`)
             }
         }
+    } else {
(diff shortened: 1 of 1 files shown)


b19jenda (approved):
Looks good and works really well

DuggaSys/diagram.js
```diff
@@ -1122,6 +1122,14 @@ document.addEventListener('keyup', function (e)
                 displayMessage(messageTypes.SUCCESS, `Clipboard cleared.`)
             }
         }
+    } else {
+        if(document.activeElement.id == 'elementProperty_name' && isKeybindValid(e, keybinds.ESCAPE)){
+            if(context.length == 1){
+                document.activeElement.value = context[0].name;
+                document.activeElement.blur();
+                fab_action();
```
a19hammi, line 1130: While this is not wrong it should be noted that fab_action might in the future change functionality.

PR #10785, opened 2021-05-07
Added line to generate properties when everything is selected

a19aleda:
Added line to generate properties when everything is selected


a19matma (approved):
The lock property is now generated properly! LGTM!

PR #10790, opened 2021-05-07
Change position of checking ctrl down

a19aleda:
This is because once we click on the lock button, the big if statement that checks if we are currently highlighting a button element is triggered resulting in the ctrl button being skipped and not adding a element to the context even though the ctrl button is pressed


a19matma (approved):
LGTM! Works!

PR #10794, opened 2021-05-07
Now shows which date a deadline has to be set between Issue#10697

a19isada:
In the edit item menu, it now shows between which dates a deadline has to be set.
Co-Author: @a18willi


b19wilfo (approved):
Tested and it works.

PR #10796, opened 2021-05-07
Check so that only mouse button 0 can trigger double click

a19aleda:
Check so that only mouse button 0 can trigger double click


a19phith (approved):
works like a charm

Issue #10706, opened 2021-05-06
It's possible to save an empty Dugga


a16erilj:
Currently so is it possible to save a dugga without any inputs/answer. Make sure so that it's not possible to save without making atleast a change.  

For example disable the save button when there is no input/answer

PR #10797, opened 2021-05-07
G1 2021 #10706 v2

a19gusba:


a18osken (changes requested):
It is possible to press the reset button, or manually empty all values and still save the assignment

in addition there are some dev debug stuff still in code, such as blocking the groupdugga flag and some console logs.

in addition to the addition, it seems that some if/else code standards have been broken, else are to be on the same line as the bracket

DuggaSys/showDugga.php
```diff
@@ -27,6 +27,8 @@
 	include_once "../Shared/basic.php";
 	include_once "../Shared/sessions.php";
 
+	echo "<script>console.log('asdasdasdasdasdasdasdasdasd')</script>";
```
a17alfja, line 30: this should be removed

b19ponka:
Some duggas should be possible to hand in empty or unchanged (default value). So we will make a new issue for the specific duggas that needs to be altered before hand in.

B18mohah (approved):
Good job!

PR #10798, opened 2021-05-07
Change the doubleClick interval to be less

a19aleda:
So that you have less time to performe the double click event


a19matma (changes requested):
Update the comment as well!

a19matma (approved):
LGTM!

Issue #10695, opened 2021-05-05
Search after cause of visiblity error pop-window


a16erilj:
The programmers in #10656 found something called "error reading dugga visibility". Search after what causes it and fix it if possible.
![bild](https://user-images.githubusercontent.com/81558891/117154823-87478400-adbc-11eb-8368-e3e9095a58bf.png)

a19rebsa:
This error pops up when you are logged in as a teacher and you're visiting an already handed in assignment. Going to look in to what causes it.

a19rebsa:
Couldn't really solve the issue with preventing this type of "error" to occur. Showduggaservice is called an extra time when logged in as a teacher and going to an already handed in assignment. Do not know why, but the third time showduggaservice runs this problem happens. Nothing else in the dugga is affected by this error the third time tho... We did a fix in the mean time for not getting this alert.

PR #10799, opened 2021-05-07
No more error pop-window  #10695

a13zanal:
Changed debug to error_log.

For testing:
Go to a saved Url when you are logged in as a teacher and if you don't receive an alert with the text "Error reading course visibility" it should work as we planned. :)


b19ponka:
Also make sure the error_log is correctly stored

PR #10801, opened 2021-05-07
Fix so that the ruler bars move independantlly of the x and y position of the mouse positon

a19aleda:
Fix so that the ruler bars move independantlly of the x and y position of the mouse position


a19nikbo (approved):
Looks good to me. The cursor indicators no longer move to the corner. 🥇 :)

PR #10802, opened 2021-05-07
Fixed: CTRL and ALT vaibles not toggles even if input

a19chrel:
Moved out the keybinds that check if ctrl and alt is pressed outside the check for input, select, textarea check.


a19matma (approved):
Works! LGTM!

PR #10803, opened 2021-05-07
Solved "cmd"-c and "cmd"-v functionality

a19frelu:
This solution is due to MacOS overwriting the keyup event for the internal OS functions.

Co-Authored-By: Niklas Bolin <81541557+a19nikbo@users.noreply.github.com>


a19karal (changes requested):
CMD+c/v works sometimes and sometimes it doesnt

a19karal (approved):
Funkar som det ska :)

Issue #10776, opened 2021-05-07
Create a counter for the amount of times an assigment have been submitted


a16erilj:
The teacher would appreciate if they could see the amount of times an assignment have been submitted. 
So create a way to to store the amount of times the assignment have been submitted in userAnswer.

PR #10804, opened 2021-05-07
Create a counter for the amount of times an assigment have been submitted G1 2021 #10776

a19kevkh:
When you save a dugga, a record of the number of times it has been submitted will appear in the database.

To test it, just go into a dugga, save it and then check if it got a number and for dugga´s that have already been saved, check so that the number increments every time you save.


a19frebo:
Also to test it you need to run the code in alter tables that is associated with this issue

e15marol (approved):
Works great, well done!

b18jacha (approved):
This shit is epic

PR #10808, opened 2021-05-07
Added a new function for when clicking on exit button.

e15marol:
This solution works on lenasys but not inside iframe in canvas... so when testing just make sure nothing else is broken. Otherwise it should be fine to approve even though it doesn't work inside the iframe for now, since it does on the live version if you act as one of the G1 students when looking at the dugga assignments.


a19kevkh (approved):
Works as specified.

a19frebo (approved):
Works as expected! 🎉

PR #10810, opened 2021-05-07
Autoscroll to relevant swimlane#10756

a19klafr:
Please take into account that there is a bug that removes duggas from the swimlane and adds a new, empty line in the bottom.
But other than this, the autoscrolling takes you to the dugga with deadline closest to the current date. If several duggas has same deadline date you should get scrolled to the one highest up in the list.

Co-author: @a19kimgr

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -1502,27 +1502,6 @@ function drawSwimlanes() {
   }
 
 
-  var minDistance;
-  var index = -1;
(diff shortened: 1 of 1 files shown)


a19isada (approved):
Seems to work as expected but should probably be tested again after the bug has been fixed.

a18willi (approved):
This fix is currently hard to test cause of the bug described by @a19klafr. It seems to be working on the only course with different deadlines though so i'll approve it.

Issue #10779, opened 2021-05-07
sh/Index.php should redirect to 404 error page if no match occours


a16erilj:
Currently it's possible to visit index.php in sh folder directly this should not be possible since it may encourage bots to webscrape it. Make so that it instead redirects to 404 error page if not matches are found.

PR #10811, opened 2021-05-07
sh/index.php now redirects to 404 error page when no match is found #10779

a19rebsa:
Fixed so that you are redirected to 404 page when there is no match for assignment or course. So also if you try to go directly to inde.php file you should see the error page.


a13zanal:
For testing:
Go to "..../LenaSYS/sh/", now you can see 404 page instead of "no match found" 👍🏻

e15marol (approved):
Works fine, good job!

b18jacha (approved):
Works! It redirects you correctly.

PR #10814, opened 2021-05-07
Element nodes is resized depending on zoom

a19vikth:


a19matma (approved):
LGTM! Maybe you should have added a comment as to why it's 8*zoomfact, but it's ok.

a19hammi:
> 
> 
> LGTM! Maybe you should have added a comment as to why it's 8*zoomfact, but it's ok.

Yes magic numbers should be kept at a minimum.

DuggaSys/diagram.js
```diff
@@ -3536,11 +3536,19 @@ function addNodes(element)
...
     var nodes = "";
-
-    nodes += "<span class='node mr'></span>";
-    nodes += "<span class='node ml'></span>";
-
+    nodes += "<span id='mr' class='node mr'></span>";
+    nodes += "<span id='ml' class='node ml'></span>";
     elementDiv.innerHTML += nodes;
+
+    var nodeSize = 8*zoomfact;
```
a19hammi, line 3543: Please do not use magic numbers. Make a constant with the value of 8 above `nodeSize` and make a comment why it is 8.

a19hammi (changes requested):

Issue #10500, opened 2021-04-28
Create a new results table based on useranswer hashes


a18osken:
The current resulted.js/php/services table shown below 
![resulted](https://user-images.githubusercontent.com/62884391/116406183-af6d3b00-a830-11eb-84fa-905d0d369774.png)
needs to be replaced with a table showing the following:
- Hand-In Nr - Really nothing to do with the hand-in itself, just number the entries in the table
- Hand-In Hash 
- Hand-In Password - gömd som default - klicka på show för att visa.
- Submitted Date
- Grade
- Hand-In link - for the teacher to view the hand-in

a18osken:
It would also be cool to be able to sort on these things I guess, at least the submission date

a19gusba:
Got some prototypes of how the new table could look like/function

![image](https://user-images.githubusercontent.com/62884429/116882244-811e9f80-ac24-11eb-83ff-944f5794a33c.png)

a19gusba:
by clicking on the password blocker you can toggle between revealing and hiding the password

![image](https://user-images.githubusercontent.com/62884429/116882341-a6131280-ac24-11eb-8b60-1fdb046fc786.png)

a19gusba:
Have also tried to implement two sorting functions that can sort by duggaName and date in case this new table doesnt cooperate well the the already exisiting sortabletable.

![image](https://user-images.githubusercontent.com/62884429/116882535-da86ce80-ac24-11eb-8f1e-ce1485937885.png)

a19gusba:
Also tried adding basic sort/filer buttons

![image](https://user-images.githubusercontent.com/62884429/116885670-6fd79200-ac28-11eb-870f-d0548ec70bbb.png)

a19gusba:
How do you want the filter/sort to work? Drop down, multiple selection buttons, radio buttons etc. Also what other sorting options should be added

a17alfja:
![image](https://user-images.githubusercontent.com/81559713/116971040-4450b800-acb9-11eb-9755-eb551537afdb.png)

a17alfja:
we are also working on a version that uses the sortableTable class, but we have some issues with filtering that needs to be fixed.

a19gusba:
![image](https://user-images.githubusercontent.com/62884429/116986456-188bfd00-acce-11eb-9f7e-be2ffbbd708d.png)

Prototype completed. 

https://gyazo.com/ad1f921486789d010b983c2b8a6c7623

[...]

PR #10815, opened 2021-05-07
G1 2021 w19 #10500

a17alfja:


a17alfja:
link does not work atm, can be fixed later

a18osken (commented):
Just on a side note, it seems the solution reverts to the old table if you run an older php version, 7.4.16 in my case, but works fine on php version 8.0.3

However this is likely to be a local issue as I seem to be having a lot of them

a18osken (approved):
Works, nicely done!

B18mohah (approved):
Good job!

b19ponka:
I would prefer if the old code was commented out, rather than deleted, for now

Issue #10651, opened 2021-05-04
Allow teachers to switch between variants when not inside a submission


a16erilj:
Create a feature that allow teachers to switch between different variants to check if the assignment is presented correctly.

![bild](https://user-images.githubusercontent.com/81558891/117011746-6f093380-acee-11eb-9216-0e7205fc0206.png)

PR #10818, opened 2021-05-07
Allow teachers to switch between variants when not inside a submission #10651

a19arvka:
To test this: 
**As a guest:**

- Make sure no "Next variant" button is shown next to the dugga title when not logged in.

**As a teacher/superUser:**

- Enter any dugga, if there is more than one variant of that dugga, a "Next variant" button should appear next to the dugga title. 
- Clear localstorage.
- Reload dugga and enter "application", you should see a variant of a value. Ex: _{"value": "4" ...}_ 
- Press the "Next variant" button, you should be prompted to reload the page.
- See if the value in localstorage and the dugga "question" changes.

### DIFF
--- a/DuggaSys/codeviewer.js
+++ b/DuggaSys/codeviewer.js
@@ -85,7 +85,7 @@ function returned(data)
 	retData = data;
 
 	//Stores the data that was sent from Section, which contains the ordering of the examples.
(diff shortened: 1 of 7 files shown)


a13zanal (approved):
Works as decribed, good job!

a19rebsa (approved):
It works fine

PR #10819, opened 2021-05-07
Reverted Template Literals

a19danla:
Reverted the old template literals that were breaking several functions.


a18willi (approved):
Broken functionality works again. But as @a19isada said, a new issue should be created to fix the template literals without breaking any code.

PR #10827, opened 2021-05-10
Fixed A4 tooltip and keybind

a19steis:
The toggle A4 button now has a shortcut
and the tooltip now also displays this shortcut when hovering.


a19frelu (approved):
Runs like clockwork. Good job.

PR #10829, opened 2021-05-10
Added some styling so linjeduggor is centered

b18jacha:
**Added some styling so linjeduggor is centered.**

To test:
- Go to a linjedugga and resize the window, also try mobile etc. It should be in the centered.

_Co-Authored-By: e15marol <markus-olsson@outlook.com>_

### DIFF
--- a/DuggaSys/templates/dugga3.html
+++ b/DuggaSys/templates/dugga3.html
@@ -15,7 +15,7 @@
             
 		</div>
 	</div>
-<div>
+<div id="dugga3Container">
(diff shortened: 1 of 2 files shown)


e15marol:
Simple solution, same as the one used for "Transformationsdugga" which is dugga4.html and dugga4.js

a13zanal (approved):
Looks good!

a19rebsa (approved):
Looks good and works as intended!

Issue #10781, opened 2021-05-07
Exam where we should add teachers to superuser rows


a16erilj:
Currently so have we wrongly assigned some permissions to superusers only. Check what superuser permission rows that should possibly be changed into teachers rows, note them down and (change them where it's appropriate).

A tip would to check with github to see what our Group 1 have changed compared to the old LenaSYS

a19rebsa:
Where are we supposed to make these changes? What actions and parts of the site should only superusers have access to? And vice versa with teachers?

PR #10832, opened 2021-05-10
Added teacher access/functionality #10781

a19rebsa:
We fixed so that

- teachers can't upload files on submission dugga 
- feedback area is visible on handed-in assignments for teachers
- save and reset for already handed in assignments should be disabled
- teachers do not need to write in password for students handed-in assignments


a19gusba (approved):

a17alfja (approved):
The if-statement on row 205 no longer exists and shouldn't be added in the merge

Issue #10825, opened 2021-05-10
Frågedugga 1 saving the dugga unchecks the answer


a16erilj:
Frågedugga 1 when saving the dugga so is the answer unchecked. Fix this
![bild](https://user-images.githubusercontent.com/81558891/117626465-6a71ce80-b177-11eb-823d-44c1280d69fb.png)

a19gusba:
since the way this dugga is saved is quite different from other duggas we had to create a new functionality to get the old answer and display it

PR #10833, opened 2021-05-10
G1 2021 #10825

a19gusba:


a19kevkh (approved):
Mission accomplished.

a19frebo (approved):
Works as intended. The selected answer is checked when loading the dugga

Issue #10735, opened 2021-05-06
You get prompted for password when you update the "Rapport inlämning" dugga


a19frebo:
If you clear your local storage and enter the "Rapport inlämning" dugga, upload a file and then save the dugga. If you then refresh the webpage you get prompted to enter the password. 

In any other dugga you do not get prompted to enter the password when you refresh the webpage.This only happens in this dugga as i have noticed.

PR #10834, opened 2021-05-10
You get prompted for password when you update the "Rapport inlämning" dugga #10735

a19kevkh:
Fixed function that checks the latest hash visited.


e15marol (commented):
Tried with the weekly branch, after uploading a file and having cleared the localstorage, you can't save, until you refresh the page, and even then there is no prompt for a password? To even get a receipt you have to go backwards twice after uploading and saving a file, on the weekly branch.

Your branch is different in that you get a receipt upon saving the dugga, and there are no issues otherwise as far as we can see. So the issue is seemingly fixed.

b18jacha (approved):
Works!

Issue #10848, opened 2021-05-10
Cannot read property 'includes' of undefined


b19ponka:
![Screenshot 2021-05-10 145330](https://user-images.githubusercontent.com/62886839/117662292-b0428d00-b19f-11eb-9711-caaba3cb9724.png)

PR #10849, opened 2021-05-10
Cannot read property 'includes' of undefined #10848

a19frebo:
Added another condition that checks if the variable pwd is undefined and not only if it contains "undef".


a13zanal (approved):
Looks good!

PR #10866, opened 2021-05-10
Fixed correct coordinates on ruler when zooming

a19karwa:
Coordinates are now correct when zooming.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -3671,7 +3671,6 @@ function drawRulerBars(X,Y)
     //Settings - Ruler
 
     const lineRatio = 10;
-    const fullLineRatio = 10;
     
     var barY, barX = "";
     const color = "black";
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -820,8 +820,6 @@ var movingContainer = false;
 //Grid Settings
 var settings = {
     ruler: {
-        lineRatio: 10,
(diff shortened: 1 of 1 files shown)


DuggaSys/diagram.js
```diff
@@ -3663,21 +3662,27 @@ function removeNodes()
...
     //Get elements
     if(!settings.ruler.isRulerActive) return;
     
     svgX = document.getElementById("ruler-x-svg");
     svgY = document.getElementById("ruler-y-svg");
     //Settings - Ruler
 
-
+    const lineRatio = 10;
+    const fullLineRatio = 10;
```
a19steis, line 3674: "const fullLineRatio = 10;" <-- unused variable.

DuggaSys/diagram.js
```diff
@@ -820,8 +820,6 @@ var movingContainer = false;
 //Grid Settings
 var settings = {
     ruler: {
-        lineRatio: 10,
```
a19steis, line 823: This can still be used to set a const lineRatio as you did at row 3673, so it does not HAVE to be removed.

example: const lineRatio = settings.ruler.lineRatio;

Not sure if it makes a difference unless we want to keep all grid settings within "settings".

a19steis (changes requested):
Overall looks good, just some small changes.

a19hammi:
Missing issue link @a19karwa

Issue #10830, opened 2021-05-10
Make handle hash great again


b19ponka:
Handle hash function is not being used correctly

HGustavs:
If we are talking a several-character base64 hash... a colision would be very very improbable.... if it works for youtube I think we can be fairly sure it is enough for our use case. As long as the length is suitable e.g. 2 characters in base 64 may result in a colision but 6 or eight characters will most likely never happen even on youtube.

a19arvka:
I will still try to fix this. We looked up the odds, it's 1 : 5 000 000 000 to get the same hash **under the circumstance that the system and database works correctly**. However, we need to consider what could happen if there was a fault in the database. An SQL injection for example? If we don't have a second method checking the validity of a randomly generated string any server/database faults could lead to loss of data and incorrect error message being printed to a student.

a19arvka:
I managed to fix it. No functionality was changed. I just added more conditions to differentiate a hash generation collision from just re-saving a dugga.

PR #10868, opened 2021-05-10
Make handle hash great again #10830

a19arvka:
**Edited since last review!**
recursiveAjax() is now replaced with a method clearing the localstorage for that dugga, then refreshing the browser. This results in a new hash being generated and stored to the localstorage. 
Also, a student gets notified whenever he/she tries to save an already saved dugga.

To try this:

- Clear local storage, enter a dugga and press 'Save' and everything should work. Now press 'Save' again. This time a popup should appear and alert that you already have a save made on this dugga.
- Next up, don't clear the local storage, exit & enter the same dugga again. Press 'Save'. The same popup should be visable.
- **Load** an already existing dugga with URL and Password. Change something on the dugga and press 'Save'. Again, the popup should appear. 
- Make sure the popup doesn't appear when doing a dugga for the first time (when you haven't saved any changes yet).
- Unfortunately I haven't found a smart way of testing the handleHash() method changes. You can force a value of hash (a value that you already have in your database) in dugga.js, on row 89. Doing so the site will just spam-refresh since the forced value will continue to overwrite the newest generated hash. I would argue that if the site refreshes, it proves that it indeed noticed a non-unique hash. In addition, the odds of this happening without forcing a value is about 1 : 5000 000 000.

### DIFF
--- a/DuggaSys/templates/clipping_masking_dugga.js
+++ b/DuggaSys/templates/clipping_masking_dugga.js
@@ -181,7 +181,7 @@ function saveClick()
 		dataType: "json",
 		success: function(data) {
 			ishashindb = data['ishashindb'];	//Ajax call return - ishashindb == true: not unique hash, ishashindb == false: unique hash.
-			if(ishashindb==true && ishashinurl ==true){				//If the hash already exist in database.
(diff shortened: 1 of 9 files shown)


b19ponka:
Remove console.log before merge

a19gusba (changes requested):
new hash is created but a new dugga wont be loaded

a17alfja (changes requested):
When you generate a hash that already exists, it will generate a new hash but it will not re-direct you to it.

PR #10869, opened 2021-05-10
Fix: copy-paste after/before panning or moving elements

b19jenda:
Now you can copy elements and move them around without any problems or errors with ctrl C and ctrl V.


a19karwa (approved):
LGTM!

Issue #10577, opened 2021-05-03
Demo diagram


HGustavs:
I have adapted a diagram from "Fundamentals of Database Systems" which has all of the necessary properties of an ER diagram.

I suggest we add a version of this diagram as our demonstration diagram. 

So that we are sure that we can draw any ER diagram.

HGustavs:
<img width="615" alt="Screenshot 2021-05-03 at 09 41 31" src="https://user-images.githubusercontent.com/6978801/116851972-c2e32200-abf3-11eb-98d5-49037f01eef9.png">

HGustavs:
I may update this diagram slightly to add more features and variants.

HGustavs:
We can add screenshots here to showcase how close the current version of the system is to the goal diagram.

HGustavs:
It would be very interesting to know how large the json data file is for this document when the demo model has been created.

a19hammi:
fixed in #10617

PR #10870, opened 2021-05-11
Fixed demo diagram

a19steis:
The cardinalities should now be as demonstrated in issue #10577.

The cardinalities sometimes cut into the line but i think that is related to another/earlier issue.


a19frelu (approved):
Now the diagram cardinalities looks identical with the demo diagram. 
Nice effort!

PR #10872, opened 2021-05-11
Diagram: Replay-mode

a19chrel:
Added replay-mode

- By pressing reply button the tool-bar, fab button and the options panel gets hidden.
- If the user presses play from the current state and forward the diagram will be displayed.
- The delay between changes is editable.
- If the replay is on last state, if press play the replay will begin from the beginning.
- When the user is exiting the replay-mode, the diagram will be restored to the state before entering.

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -321,11 +321,11 @@
 }
 
 #zoom-message-box{
-    transform: translate(50px, 930px);
     position: absolute;
     background-color: rgba(0, 0, 0, 0.8);
     width: 75px;
-    left: 50px;
+    left: 100px;
+    bottom: 5px;
     color: white;
     border-radius: 5px;
     text-align: center;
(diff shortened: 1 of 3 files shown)


a19phith (changes requested):
image
image

seems to be working but one thing that the replay does not store is the lock in place functionallity. When the user includes a lock in place for a entity and then start the replay function. It removes the lock part of the entity. 
image

a19chrel:
> 
> 
> image
> image
> 
> seems to be working but one thing that the replay does not store is the lock in place functionallity. When the user includes a lock in place for a entity and then start the replay function. It removes the lock part of the entity.
> image

This have nothing to do with the replay-mode, this has with the stateMachine does not store the isLocked attribute of the object. It is currently the same with line cardenality. The change is not added to the stateMachine.

a19chrel:
As @a19phith found that locking is not saved and so on, an new issue #10898 has been created to look deeper into this.

PR #10873, opened 2021-05-11
Disabled panning with mouse3 while moving object

a19chrel:
It is not possible to use mouse3 now while moving a object.


a19karwa (approved):
Bracket fixed it. Now it works as intended!
👍

PR #10874, opened 2021-05-11
Fixed student superuser

a17alfja:
Students will now have the userID "student" not "1" which had superuser privileges. Also added hasAccess check in showDuggaservice and changed variable name (isSuperuser -> isTeacher)


a19frebo (approved):
Works fine!

Issue #10844, opened 2021-05-10
Save and Reset button won't be enabled in Frågedugga, Random css dugga & Random clipping dugga


a16erilj:
For some reason so does the Save and Reset buttons not become enabled in Frågedugga, Random css dugga & Random clipping dugga when the student interacts with the dugga. Checkout why and fix it if possible

![bild](https://user-images.githubusercontent.com/81558891/117656665-e8929d00-b198-11eb-86af-0d22c93e15f6.png)

a16erilj:
Not that they have become disabled. Rather that they won't become enabled after the student have interacted with the dugga. 
The buttons are disabled by default in order to stop empty submissions and should become enabled once the student / not inlogged user does a interaction with the dugga.

a16erilj:
Kinda similar to Biträkningsdugga 1 where if a button is pressed then save and reset buttons will become enabled.

a19rebsa:
Ok!

PR #10879, opened 2021-05-11
Buttons are now enabled in duggas after interacting #10844

a19rebsa:
Fixed css-dugga, clipping-dugga and frågedugga in the course "Webbutveckling - datorgrafik" so that the buttons now are enabled when the user is interacting with the dugga.


a19frebo (approved):
Works! Find the same bug as @a19kevkh, but it is not related to this issue

PR #10881, opened 2021-05-11
Made back button show again

a19isada:
Made back button show for screens less than 400 px wide. 
Co-Author: @a18willi


a19danla (approved):
Works just fine

b19wilfo (approved):
Works as intended.

PR #10891, opened 2021-05-11
Removed duplicate search icon while in mobile view. Issue#10890

a18willi:
Removes search icon from navheader when the mobile search bar is enabled in fileed.

Co-author: @a19isada


b19alest (approved):
Works fine!

Issue #10863, opened 2021-05-10
Add the collumn "Times" as in amount of times it have been submitted next to Submission date


a16erilj:
A student can turn in a submission how many times like before deadline. An the teacher would like to be able to see how many times a specific hash have been turned in. So make use of the timesSubmitted counter in userAnswer to show this.

PR #10893, opened 2021-05-11
TimesSubmitted to resulted file #10863

a13zanal:
Add timesSubmitted column in "Edit student results" table for teacher


a19rebsa:
To test:
Go to resulted.php when you're logged in as a superuser (e.g brom). You can see that a new column "Times submitted" is added. There might be many null values because the timesSubmitted column was added just yesterday in the userAnswer table. In that case try to save som duggas when you are not logged in, and the table should display values in the column.

a19arvka (approved):
**Lit**, what's the limit of the counter? 99999999?

PR #10894, opened 2021-05-11
Removed duplicate announcement button

a19isada:
Removed duplicate announcement button
Co-Author: @a18willi

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -1465,7 +1465,7 @@ div .navButt:hover {
     .access menuButton{
       width: 20px;
     }
-    #navHeading, #announcement {
+    #navHeading {
         display: none;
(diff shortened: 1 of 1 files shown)


a18danli:
Skärmavbild 2021-05-11 kl  15 10 09

You can see that it is still there if you move the screen slowly, just add a mq and display icon none and it will be fixed

a19samke (changes requested):
announsment

It looks like it still exists for me, I think you need to check more on media querys, Or you can just use JavaScript when the hamburger menu is displayed, then the announcement icon should disappear from navheader.

a18danli (approved):
Perfect!

a19samke (approved):

Issue #10871, opened 2021-05-11
Make a md guide for creating courses with assignment


a16erilj:
We would like a md guide for creating courses and assignments, if there is some functions that you find usefull (like how to clone a course) add them also to the documentation.
Create a md file named add_courses_and_assignments.md in Canvas documentation.

PR #10897, opened 2021-05-11
Make a md guide for creating courses with assignment #10871

a19kevkh:
Added "add_courses_and_assignments.md" and pictures for a description on how to add courses and assignments to canvas.


a13zanal (changes requested):
Looks good, except that this picture on point 4 does not seem to want to be included.
image

a19rebsa (changes requested):
The picture for step 4 is broken/can not be seen. When you click on it you come to a 404-page. So this should be fixed, otherwise it looks great! Well done

PR #10899, opened 2021-05-11
Lock-icon is now correct size in the hamburger menu. Issue#10887

a18willi:
Removed code that made the button 20 px wide instead of 32

CO-authos: @a19isada


b19wilfo (approved):
Looks great!

PR #10904, opened 2021-05-11
G2 2021 w19 issue#10831

a19danla:
Updated the file with template literals and row breaks to increase readability.
Hopefully it wont break the swimlane functionality this time.


a18willi (changes requested):
Arrows are hidden on moments that are collpased (see Färgduggor below)
image
All other functionality seems to work though.

a19isada (changes requested):
The arrow pointing to the right on moments is not visible when hiding duggas under a moment
image

a18willi (approved):
Looks better now!

a19isada (approved):
Works as normal now.

PR #10908, opened 2021-05-11
Fix: timelimit append on object

a19chrel:
Fix error when timelimit try to check an object.


a19matma (approved):
LGTM! Works, problem solved!

Issue #10846, opened 2021-05-10
A4 horisontal 


a19karwa:
Currently you can toggle an a4 template, perhaps it would be useful to have an horisontal a4 template aswell in the toggle?

![image](https://user-images.githubusercontent.com/81557429/117660596-a881e900-b19d-11eb-9bf2-e8bef4247b39.png)

a19steis:
Do we want a vertical A4 and a horizontal A4 sheet to both show up upon toggle?

or do we want seperate toggles for each?

a19hammi:
fixed in #10919

PR #10919, opened 2021-05-12
Added horizontal a4

a19steis:
Added a horizontal a4 on top of the veritcal when we toggle a4.

Subject to change, discussed in #10846.


a19phith (changes requested):
Looks good codewise but the messures are abit off? the horizontal a4 sheet is both to long and not wide enought

a19phith (changes requested):
Now it looks better, one more thing tho. you're sitting on a w19 branch. Would be wise to change it to a w20 branch so it wont be knas :=

a19phith (approved):
good of good

PR #10923, opened 2021-05-12
Added timeSubmitted as sortable

a17alfja:
timeSubmitted column is now sortable


a17alfja:
for the compare function to work, all null values need to be removed. (they are null because the submissions were made before the column existed)

a13zanal (approved):
Works as described!

a19rebsa (approved):
Works fine!

Issue #10851, opened 2021-05-10
Make hash in resulted.php redirect to the corresponding submission


a16erilj:
It would be nice to be able to quickly go to the dugga in question. So make the hash value in the table a link to the corresponding assignment.

![bild](https://user-images.githubusercontent.com/81558891/117663379-e8969b00-b1a0-11eb-92c8-937c09d4814e.png)

a16erilj:
A bit of warning. Currently so is the common redirection we sh/index.php?a=**HASH** broken. Feel free to treat it like it works.

PR #10924, opened 2021-05-12
Hash in resulted.php redirect to the corresponding submission #10851

a13zanal:
Added a url to the a-tag for the hash in the "Edit student results" table.

OBS. Currently, sh / index.php? A = HASH is broken, so you will not be able to get to the right page. Just check that the url is correct when you press the hash on "Edit student results".

### DIFF
--- a/DuggaSys/resulted.js
+++ b/DuggaSys/resulted.js
@@ -89,7 +89,7 @@ function renderCell(col, celldata, cellid) {
 	str = "<div class='resultTableCell'>";
 	if (col == "hash"){
 		str += "<div class='resultTableText'>";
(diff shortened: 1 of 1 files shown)


a19rebsa:
Also note that the link goes to the localhost so you need to change the filepath in the url so it's correct for your local solution

a17alfja (approved):
Link works, but we found an issue where submitted files randomly appear/disappear in rapport dugga.

a19gusba (approved):
Works! However we found a new bug when a teacher tries to visit "Rapport inlämning" assignment and the teacher can only see the files submited like 50% of the time (literally) when refreshing the page it seems to be completly random if the files are visible for the teacher or not

Issue #10865, opened 2021-05-10
Fix sh/index.php redirection 


a16erilj:
Currently so does sh/index.php redirect all calls that does not contain parameter a to 404 page. Fix it so that all parameter can be used

a16erilj:
The parameters in querstion are a & c

a should be able to be used without c
c should be able to be used without a
a && c should be able to be used together
If no a || c then send them to the error page 404

b18jacha:
> The parameters in querstion are a & c
> 
> a should be able to be used without c
> c should be able to be used without a
> a && c should be able to be used together
> If no a || c then send them to the error page 404

Thanks for clarification!

PR #10930, opened 2021-05-12
G1 2021 w19 #10865

e15marol:
We've fixed so that you can use either the hash to load the assignment or the course string. 
To test entering a course for example, simply put "http://localhost/lenasys/LenaSYS/sh/?c=Software Engineering IT301G HT15" in your URL, or to test the hash simply put http://localhost/lenasys/LenaSYS/sh/?a=<hash> with a hash being in your database, if you enter the wrong hash you get a 404 page.

Combining an assignment and course should work, but this is something that's going to be implemented in the next week's branch instead.

### DIFF
--- a/sh/index.php
+++ b/sh/index.php
@@ -58,27 +58,31 @@ function CourseAndAssignment($course, $assignment) {
 
 	global $pdo;
 
-	//Get vers.vers
+	//Get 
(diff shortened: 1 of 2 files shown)


a13zanal (approved):
Works fine, good job!

a19rebsa (approved):
Works as intended! Good job

PR #10932, opened 2021-05-12
Lines are now deselected on copy paste

a19steis:
Previous lines are now deselected when copy-pasting all elements at once


a19matma (changes requested):
Firstly, this branch should be merged into Group3-2021-W20-main and not Group3-2021-W19-main. This most likely because the pull-request was made before the weekly merge, and I was assigned as a reviewer when the weekly merge was done, (so I couldn't have done anything about it sooner). 

Secondly, when using ctrl+a and then copy paste it deselects the original and only select the new pasted elements. But when pressing ctrl+v again, it somehow adds the previous selection to the new, and thus we get a copy of both the original and the new elements and lines. Not 100% sure this is related to the issue, but original lines are indeed deselected in the diagram UI after pasting.

a19matma (approved):
LGTM!

PR #10933, opened 2021-05-12
Updated the demo diagram elements to the right sizes and positions.

a19frelu:
The demo Relations now have the correct size in the diagram and the coordinates for some elements have been changed to match the diagram template after the changes.


a19phith (approved):
Looks good

PR #10935, opened 2021-05-12
Update  The message of the day

a19samke:
The message of the day button moved to the dropdown menu in mobile vers!
@a18danli


a18danli:
Skärmavbild 2021-05-12 kl  12 28 32

In our version we didnt have the changed css for text description, but in the merge it should look like this

PR #10937, opened 2021-05-12
Blacklisted id on append and changed number check

a19chrel:
Changed the number check, before it also returned true if it was only numbers in the string. Also blacklisted id from appending because it should not be possible. And removed time when restoring an object.


a19phith (approved):
LGTM

PR #10938, opened 2021-05-12
"Fixed" Small bug: Exiting Replay-mode with Esc not always working

a19phith:
Missplaced if function from the beginning so i moved the if statement so the escape now works


a19steis (approved):
Works like a charm, good work.

PR #10940, opened 2021-05-12
Cleared the replayTimer in diagram.

a19frelu:
This fixed the problem with the replay in play mode.


a19karal (approved):
it works but there are some issues left, if u delete elements then use reply the elements will "respawn" after exiting the reply mode

a19hammi:
> 
> 
> it works but there are some issues left, if u delete elements then use reply the elements will "respawn" after exiting the reply mode

This should be a new issue

PR #10946, opened 2021-05-12
Update The message of the day and added to the hamburgerBox

a19samke:
@a18danli 
The message of the day button moved to the dropdown menu in mobile vers!

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -5138,13 +5138,13 @@ only screen and (max-device-width: 860px) {
  
   #versionCogBurger, #versionPlusBurger, #editStudentBurger,
   #testsBTNBurger, #editFilesBurger, #courseIMGBurger,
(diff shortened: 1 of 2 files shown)


a20chrjo (approved):
Works as intended, in Chrome, Edge, Firefox and Opera. Even when you refresh the page after removing the MOTD.

b19alest (approved):
Works very nice! Didn't find any issues with it.

b18phika:
Since the dropdown isn't visible to users who aren't teachers, the motd-button isn't visible to non-teachers after this change. Is this the intended behavior?

Before:
image

After:
image

a18danli:
MOTD Button should now work for all users, even guest. 

MOTD is not in the hamburgerbox in this version.

Try this on diffrent webbrowsers.

a19danla (approved):
Seems to be working as intended according to comments

PR #10951, opened 2021-05-12
Saving both copied elements and lines to localstorage

a19matma:
Saving the copy-data to localstorage now instead of saving to a local array.

Also removed default cardinality "MANY"-attribute when a new line is added in addLine(). This is because when elements are pasted, the line-cardinality should be the same as the copied elements line (if a line exists that is).


a19matma:
To test this properly, please try to hard refresh the browser as well.

a19chrel (approved):
LGTM

PR #10954, opened 2021-05-12
Added teacher access check

a17alfja:
added if statement to allow only teachers and super users to view the table in resulted.php


a17alfja:
To test,
Copy the url when you are at the table
Logout
Paste the url
Voilà, no data

a18osken (approved):
Works well, nicely done!

a18osken (commented):
The page works as intended but this error is thrown:
image
Probably since uid is not logged in, dunno if it necessarily needs to be solved however

B18mohah (approved):

PR #10956, opened 2021-05-12
Dropdown menu text is now clickable

a19isada:
Made text clickable on the hamburger menu in sectioned.


b19alest (approved):
What I can see it works good 👍

a20chrjo (approved):
Works in Chrome, Edge, Firefox and Opera

Issue #10915, opened 2021-05-11
The message of the day icon disappears when the message of the day has been clicked away


a19anjst:
The message of the day icon disappears when the message of the day has been clicked away and the page reloaded.
![Skärmbild (384)](https://user-images.githubusercontent.com/62884743/117837547-d50b3300-b279-11eb-8774-294ca97c0854.png)
![Skärmbild (385)](https://user-images.githubusercontent.com/62884743/117837550-d5a3c980-b279-11eb-8346-36df47d28286.png)

PR #10959, opened 2021-05-12
Update navheader.php Issue #10915

b19alest:
Fixed MotD icon, it no longer disappears when reloading. The problem was only for teacher/superuser.


a19samke (approved):
This works on my own environment as it should and I see no difference from the previous one!

a18danli (approved):
Looks good!

PR #10969, opened 2021-05-14
Change the swimlane text color

a19samke:
Change the swimlane text color from blue to darkblue !

@a18danli


a19isada (changes requested):
The colour is very close to black and from what I interpreted, the colour should not be black.

a18willi (changes requested):
The new color is not used anywhere else on lenasys which might make it seem out of place. But the biggest issue is that @a19anjst specifically mentioned thet the text should not be black which the new color seems to be very close to.

a19isada (approved):
Looks good. Much clearer distintion from black than before.

a18willi (approved):
Looks much better now

PR #10973, opened 2021-05-14
Added: Save, Export, Load functionality with buttons in options-panel

a19matma:
Saves and/or exports a file in JSON-format.
Import only works with JSON-files as well.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -372,9 +372,7 @@ class StateMachine
     {
         if (stateChange instanceof StateChange) {
             // Remove the history entries that are after current index
-            while(this.currentHistoryIndex + 1 != this.historyLog.length) {
-                this.historyLog.pop();
-            }
+            this.removeFutureStates();
 
             // If history is present, perform soft/hard-check
             if (this.historyLog.length > 0) {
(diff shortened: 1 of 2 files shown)


a19steis (approved):
Looks good, json files can be saved/exported and imported.

a19chrel:
Some changes need to be done, the current solution just put the loaded data into data and lines, should be added to initialState.

DuggaSys/diagram.js
```diff
@@ -476,7 +473,12 @@ class StateMachine
         // Change the sliders max to historyLogs length
         document.getElementById("replay-range").setAttribute("max", this.historyLog.length.toString());
     }
-
+    removeFutureStates(){
+        // Remove the history entries that are after current index
+        while(this.currentHistoryIndex + 1 != this.historyLog.length) {
+            this.historyLog.pop();
+        }
```
a19hammi, line 480: Wouldn't just a splice do? Same end result but doesn't need to resize the object more than once that way.

a19steis (approved):
Other than what hamlet commented, i dont see anything strange. Seems to be working fine.

DuggaSys/diagram.js
a19chrel: It is true, we just used the old code that did the removal of elements before currentIndex, do you want us to change to an splice?

a19hammi: Preferably. Less code with same result.

a19chrel: The changes are done. Can you solve the conflict and then request an review again from @a19steis

a19hammi:
Another review my boy @a19steis

a19steis (approved):
Splice change looks correct, good job ! :)

PR #10979, opened 2021-05-14
Chnaged key length to 2

b19jenda:
Now the code for handeling deleted elements and lines will now be executed. Lenght of keys when deleting elements has the length of 2 not 1.


b19jenda:
Now when deleting elements they will also be deleted in replay

a19phith (approved):
LGTM

Issue #10945, opened 2021-05-12
Deleting items then using replay mode will "respawn" the deleted items after exiting replay mode


a19karal:
Deleting items should stay deleted after exiting replay mode. no resurrections this time >:(
![33b7e0cf169b1bdf5d047fa9e56353ab](https://user-images.githubusercontent.com/62888356/117968261-486a7e80-b326-11eb-96e3-888b4c5e959c.gif)

a19chrel:
This is because currently the delete state in the stateMachine seems not to work. One reason for this can be that the new `time` key has been added to keep track of time on all changes.

At the moment if only the state object contain an array of id, the `restoreState()` function will delete the array. Because this is not the case anymore it do not work.

I would suggest that we for the delete-state uses the key deleted instead of id.

a19hammi:
fixed in #10979

PR #10988, opened 2021-05-17
Added missing changes to stateMachine

a19chrel:
Added so isLocked-attribute on elements gets saved in the stateMachine also for lines it now saves "cardenality" and "kind".


a19nikbo (approved):
Seems to works as intended and the code standard is followed. 
I'm aware that redoing the deletion of elements is not implemented in this fix, but since that is the basis of issue #10945, that's fine.

Good job!

PR #10990, opened 2021-05-17
Changed max zoom-out to 0.25 instead of 0.125

a19nikbo:
The zoomfactor for max zoom-out is now 0.25


a19karal (approved):
Funkar som det ska :)

PR #10991, opened 2021-05-17
Added check if replay-mode is active on "esc"

a19chrel:
Added check so you need to be in replay-mode to exit.


a19karal (approved):
Funkar som det ska

PR #10992, opened 2021-05-17
Added if statement to check if element you clicked on is in container

b19jenda:
Now placing new elements will only happen in container


a19karal (approved):
Funkar som det ska :)

Issue #10986, opened 2021-05-17
Minor error is thrown, noticed when reviewed issue #10954


b19ponka:
![Screenshot 2021-05-17 105343](https://user-images.githubusercontent.com/62886839/118461789-52152d00-b6fe-11eb-93e5-b1f48f9b9d60.png)

#10954

HGustavs:
I can confirm that this is a bug and needs fixing, we want the running system with as few errors as possible.

Seems like an errror mesage that was channelled, do we know the origin of original error?

a19arvka:
I found a solution, but i'm unsure why the error occured in the first place. 
Since the functionality of resulted.php have been changed recently there was an old if statement (that I created many weeks back) checking isSuperUser. This statement don't serve any purpose anymore, and i'm pretty sure we can't check _$_SESSION['uid']_ directly in _resulted.php_ because _resultedservice.php_ takes care of accessability right now. Checking accessability conditions in both _resulted.php_ and _resultedservice.php_ is likely the cause of the error.

PR #11001, opened 2021-05-17
"Minor error is thrown, noticed when reviewed issue #10954" - #10986

a19arvka:
The reason I can remove the accessibility checks "if (isSuperUser...)" is because the functionality of resulted.php have been changed. As a teacher, we no longer preview a dugga directly in resulted.php. We only enter it as a showDugga.php link. The same if statements are already implemented on showDugga.php.

To try this:
- Log in as teacher.
- Enter resulted.php ("edit student results" icon).
- Copy URL.
- Logout.
- Paste and enter the copied URL.
- Open console, make sure no SyntaxError occurs.

### DIFF
--- a/DuggaSys/resulted.php
+++ b/DuggaSys/resulted.php
@@ -133,8 +133,6 @@ pdoConnect();
 	<!-- -------------------=============####### Result Popover #######=============------------------- -->
 
 	<div id='resultpopover' class='resultPopover' style='display: none'>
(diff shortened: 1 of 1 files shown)


a19frebo (commented):
I dont get the error from resulted.php but I do get a similar error from showdugga.php. I get this after i log out as a teacher and entering the URL.
image

a19kevkh (commented):
Same errror as Fredrik gets.

image

a19kevkh (approved):
Issue found is unrelated to this one. But this one works.

a19frebo (approved):
Create a new issue for the new error that we found. This issue is however solved.

Issue #10980, opened 2021-05-17
Create a new column that conatins the times a hash have been accessed in the database


a16erilj:
It would be great to know the amount of times a specific hash have been accessed so create a new column that contains the times a hash have been accessed in the useranswer table.

a19rebsa:
Do you mean that we should add this in resulted.php?

b19ponka:
In this issue, the column refers to a column in the databas

PR #11007, opened 2021-05-17
new column in database - TimesAccessed  #10980

a13zanal:
New column in database for the number of times a dugga is accessed.


a19rebsa:
To test this pull request:
-Update userAnswer table with ALTER TABLE userAnswer ADD COLUMN timesAccessed INT(5);
-Go to a dugga e.g. bitärkningsdugga 1 and save the dugga.
-Check the userAnswer table in your database - the timesAccessed should be 0 after saving the dugga.
-Now go to incognito mode in your browser or rinse your saved webinformation and then go to the url to the saved dugga
-When you are prompted with the login-box for the dugga the timesAccessed value for that dugga should have increased with one 

(If you want to try this several times to keep increasing the value, you have to empty the webinformation or such because this value only increases when you are prompted with the login-box)

a18osken (approved):
This works, well done! 🥇 

There are some indentation issues in the chechHashPassword function from what I could see. Those could do with some fixing.

B18mohah (approved):

PR #11009, opened 2021-05-17
Disabled select all in edge creation mode

a19steis:
Disabled CTRL+A in edge creation mode as suggested in the issues comments.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1159,8 +1159,13 @@ document.addEventListener('keydown', function (e)
         }
 
         if (isKeybindValid(e, keybinds.SELECT_ALL)){
-            e.preventDefault();
-            selectAll();
+            if(mouseMode == mouseModes.EDGE_CREATION){
+                e.preventDefault();
+                return false;
+            } else {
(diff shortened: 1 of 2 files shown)


a19chrel (approved):
The solution works as it should, good job!

a19chrel (changes requested):
The code works (disable the SELECT ALL keybind) in line-creation, but the problem is still existing if you entering line-creation with selected elements, the line-creation mode trys to add a line to context[0] and context[1].

a19chrel (changes requested):
The diagram still tries to create a line when selecting all elements in an element-creation mode and then switch to line creation with the buttons. Seems to have with the "mouseup" event on the tool-bar to do.

a19steis:
This fix solves the immediate issue, but by removing the mouseup event from toolbar may introduce other issues that needs to be resolved.

a19chrel (changes requested):
While creating this new fuction `tup,` the comment for `mup` was moved above incorrect function. Also an comment for `tup` would be nice.

a19chrel (approved):
LGMT, nice work!

Issue #11013, opened 2021-05-18
Error is thrown, noticed when reviewed issue #10986


b19ponka:
![Screenshot 2021-05-18 094006](https://user-images.githubusercontent.com/62886839/118611367-0d9c9680-b7bd-11eb-86c6-6d26b19138be.png)

a19frebo:
This error is thrown because there are two SESSION variables that are not initialized, and if they are read while not initialized this error is thrown.

PR #11014, opened 2021-05-18
Error is thrown, noticed when reviewed issue #10986, #11013

a19kevkh:
To test, go into a dugga with the URL and check so that there are no syntax errors.

### DIFF
--- a/DuggaSys/showDugga.php
+++ b/DuggaSys/showDugga.php
@@ -65,6 +65,13 @@
 		$userid="UNK";
 	}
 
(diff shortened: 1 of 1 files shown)


a19rebsa (commented):
Noticed that you can not log in to some assignments, this should be looked over.

a13zanal (commented):
Unable to log in to some duggor (linjeduggorna, Transformationsduggor and Random css dugga), but Syntax error is gone.

b19ponka:
Moved to new pull #11017

Issue #10981, opened 2021-05-17
Add the amount of times a hash have been accessed in the resulted table


a16erilj:
Teacher would like to be able to see if the a specific hash have been accessed an absurd amount of times. So add a column that  shows the amount of times a hash have been accessed in the resulted table.

HGustavs:
If the number of times an assignment has been accessed is very high we can assume hacking attempts or other security issue.

PR #11015, opened 2021-05-18
Added new table column in resulted  #10981

a19rebsa:
We have added a new column in the resulted table for how many times a handed in dugga has been accessed

### DIFF
--- a/DuggaSys/resulted.js
+++ b/DuggaSys/resulted.js
@@ -52,12 +52,13 @@ function createSortableTable(data){
 			password:"Password",
 			submitted:"Submission Date",
 			timesSubmitted: "Times submitted",
+			timesAccessed: "Times accessed",
 		},
 		tblbody: data,
 		tblfoot:{}
 	};
 
-	var colOrder = ["duggaName","hash", "password", "submitted", "timesSubmitted"];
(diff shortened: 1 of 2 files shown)


a13zanal:
To test this pull request:
- Update userAnswer table with ALTER TABLE userAnswer ADD COLUMN timesAccessed INT(5);
-  log in as a teacher and go to "Edit student results" and see if "Times accessed" exists. If you only get NULL, you can follow the description below to get results, to see that it works:
     - Go to a dugga e.g. bitärkningsdugga 1 and save the dugga.
     - Check the userAnswer table in your database - the timesAccessed should be 0 after saving the dugga.
     - Now go to incognito mode in your browser or rinse your saved web information and then go to the url to the saved dugga.
     - When you are prompted with the login-box for the dugga the timesAccessed value for that dugga should have increased 
        with one.

If you want to try this several times to keep increasing the value, you have to empty the web information in your browser every time or such because if you don't then the login-box won't appear (due to it's saved in session/local storage/or something), this value only increases when you are prompted with the login-box.

a19kevkh (approved):
Works

a19frebo (approved):
Works! But you cannot sort the entries based on times accessed, but this could be another issue

PR #11016, opened 2021-05-18
Handle elements overlapping

a19aleda:
Handle elements overlapping, so that it resets the position of the element being placed over another element and triggers a display message error.


a19phith (approved):
LGTM

PR #11018, opened 2021-05-18
Fix reset on scrollZooming

a19chrel:
Fix the problem that zoomOrigo was resetted on extra zoom out.


a19karal (approved):
Funkar som det ska :)

PR #11020, opened 2021-05-18
Swimlane not show in mac safari

a18danli:
@a19samke

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -1144,15 +1144,6 @@ function returnedSection(data) {
       resave = false;
     }
 
-    if(navigator.userAgent.indexOf("Safari") != -1)
-    {
-      $("#statisticsList").hide();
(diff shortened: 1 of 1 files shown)


a18danli:
If you want to try, use the following example:
navigator.venor == "Google Inc."

DrawSwimlane will not if using safari browser and the arrows will be hidden.

a19frelu (approved):
I followed your example by changing the navigator.vendor to "Google Inc." and back to "Apple Computer, Inc." and it seems to work as intended when I tested it in the Chrome, Firefox, Opera, Brave and Safari browsers. 
The swimlanes are now visible in every tested browser but Safari.
Good job.

a19danla (approved):
Seens to be working on pc. Tried changing so it doesn't show up on Chrome and it doesn't but still shows on Firefox

b19wilfo (approved):
Tested with chrome and firefox and it works.

PR #11024, opened 2021-05-18
Added functionality to contribution tool clock. Issue#9837

b16oskni:
Updated functionalty to show Hourly activites by weeks instead of day.
Added additional input element.

co-author; @a19isada @a18willi

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ vendor/
 
 # Visual Studio 2015/2017 cache/options directory
 .vs/
+*.db
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ vendor/
 # Visual Studio 2015/2017 cache/options directory
 .vs/
 *.db
+BGHdata_2021_05.db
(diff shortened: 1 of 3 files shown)


a20chrjo (approved):

a20chrjo (changes requested):
Tested in Chrome and Firefox.

What works:
Selecting different weeks, selecting the same week.

What does not work:
Selecting just one week and you can select the weeks in inverse order i.e. first week "5" and second week "4".

b19alest (changes requested):
Don't know if this is intended or not but if only the first week is selected then the end date don't update and if only second week is selected the start date don't update.

You can also select for example week 4 as first week and week 1 as second. Which will result in nothing showing on the clock.

Would maybe be handy to be able to only chose one week?

a19isada:
If you have selected week 4 and week 3, you will now get a popup that you have made an incorrect selection. You can select one week of you choose the same week on both of them. 

"Don't know if this is intended or not but if only the first week is selected then the end date don't update and if only second week is selected the start date don't update."

This behavior is intented.

a20chrjo (approved):
It now works as intended

b19alest (approved):
Works nice! 👍

Issue #10984, opened 2021-05-17
Local storage cleanup


b19ponka:
Local storage is a bit of a mess right now, some values are stored more than once and so on.

HGustavs:
We do not want too much cruft in local storage so it is a good idea to clean this up as this may result in problems when new developers inspect the contents of local storage.

PR #11027, opened 2021-05-18
G1 2021 #10984

a19gusba:


a19arvka (approved):
All functionality remains & localstorage is clean. Great!

a18osken (changes requested):
It seems that the teacher links in resulted now randomize the variant for the teacher or something like that, you do not get the same variant anyways.

a18osken (approved):
It Twerks

a17alfja:
I found a better solution that doesn't require another ajax call.
Pls review again (sorry!)

a19arvka (approved):
Great!

a18osken (approved):
Not having to call the service script again is a really nice find!

PR #11029, opened 2021-05-18
Now the zoom-message text will be updated when redirect camera to center

b19jenda:


a19frelu (approved):
Does the job. Nice fix.

PR #11036, opened 2021-05-18
"Fixed" Be able to resize and move the A4 representation

a19phith:
Added a function to the a4 presentation so the user is now able to switch between horizontal and vertical mode and change the size of the A4 template


b19phier (changes requested):
The size slider is reseting when changing object
image
image

b19phier (approved):
looks fine to me!

PR #11037, opened 2021-05-18
G2 2021 w20 issue#10556

b19wilfo:
New items now glow if they are less than a week old. 

Co-author: @a19danla


a19isada:
Should the new items glow for everyone or just for teachers? @b19wilfo @a19danla

a18willi (commented):
Right now the new items glow red for 2-3 seconds then abruptly stop and never glows again. Is this intended or is it supposed to glow/blink any longer?

a19danla:
In the old functionality it was for everyone so we kept it that way @a19isada

b19wilfo:
> 
> 
> Right now the new items glow red for 2-3 seconds then abruptly stop and never glows again. Is this intended or is it supposed to glow/blink any longer?

The glow-time is how it was before and we didn't need to change it. We can change it if necessery.

a18willi (approved):
Since what i left in my comment earlier seemed to be intended this looks great!

a19isada (approved):
Works well and looks good!

b16oskni (approved):
Everything works as intended!

PR #11039, opened 2021-05-18
Double click on container hides options panel

a19steis:
Double clicking on the container now hides the option panel


a19matma (approved):
LGTM!

PR #11048, opened 2021-05-19
Diagram: Move element with keys

a19chrel:
It is now possible to move selected elements with the arrow keys.

**Keep in mind:** With this soultion it is not possible to move elements in snap to grid.


a19aleda (approved):
Seems to work and code looks a-o-kay

PR #11050, opened 2021-05-19
"Fixed" Weak key underlines not scaling with zoom

a19phith:
Now added  a zoomfact calculation to the dash array so it now has a fixed number of "lines"


a19nikbo (approved):
Works as intended. The underline scales with the zoom level and the number of dashes stays the same as well. 👍

Issue #10982, opened 2021-05-17
Add Student-teachers to showdugga.php


a16erilj:
Seems like we have a role called Student-teachers in LenaSYS so maybe we should add it to showdugga. Begin with give them permission to change variant but also no able to save / reset the dugga

HGustavs:
Well. 

Not necessarily "show" the dugga as showing it would be needed for editing it, but to edit the dugga (duggaed.*) information in the assignment e.g. to correct an error in an assignment text for example

PR #11053, opened 2021-05-19
Added student teacher access #10982

a19rebsa:
We've added student teacher acceess to duggaed so that student teachers now can view and change duggas in that file. We also added access to "next variant" as student teacher, and disabled "save" and "reset" buttons in duggas.

### DIFF
--- a/DuggaSys/duggaedservice.php
+++ b/DuggaSys/duggaedservice.php
@@ -51,7 +51,7 @@ logServiceEvent($log_uuid, EventTypes::ServiceServerStart, "duggaedservice.php",
 // Services
 //------------------------------------------------------------------------------------------------
 $writeaccess = false;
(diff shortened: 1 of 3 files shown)


a13zanal:
For testing:
- Log in as a student teacher (mofree)
- Check that you can view and change dugga.
- Go to a dugga and see if the "next variant" button exists.
- Also check that they should not be able to be saved or reset.

a19kevkh (approved):
Works on dugga with variants. But not with Linjedugga 1/2, Transformationsdugga 1/2, Rapport inlämning, Random CSS and the Random Clipping dugga. 

Also, found a bug with the Frågedugga 1 where if you click on the next variant button. It will work until we get this combination:
image

Then it will never switch variants after. Although this might be another issue.

a19gusba (approved):

a19taral (approved):
nice job!

PR #11054, opened 2021-05-19
"Fixed" Box surrounding selected elements not on top of other elements

a19phith:
changed the z-index for the svg overlay and the changed the order of the php file so the z-index could work


a19karwa (approved):
LGTM! box is now visible when you drag it over another element.

PR #11069, opened 2021-05-19
Removed ladExport

a17alfja:
Removed ladexport from resulted.php and the export functionality in sortableTable class


a19frebo (approved):
Nice removal of both the visuals and the functionality.

a19kevkh (approved):

PR #11071, opened 2021-05-19
Removed ssn, firstname and lastname  accessedservice.php

a19isada:
Commented out ssn, firstname and lastname getop


b19alest (approved):
Looks good, we can now maybe look into if we can remove some of the querys that uses ssn etc?

a20chrjo (approved):
The lines in the issue has been commented out. But there are more places in accessedservice.php that uses these to store data:

$ssn = $user[0];
$firstname = $user[1];
$lastname = $user[2];

So my recommendation is to make a new issue to remove these as well @a19anjst.

Issue #11056, opened 2021-05-19
The next variant gets stuck in Frågedugga 1


a16erilj:
It seems like the variant gets stuck in Frågedugga 1 after pressing the button (next variant) some times. It have probably something to do with reaching the end of the array.

This issue was found by the reviewer squad at the branch #11053

PR #11072, opened 2021-05-19
The next variant gets stuck in Frågedugga 1 #11056

a19kevkh:
Test all the dugga if they cycle through all the variants.


a13zanal (approved):
Works great!

a19rebsa (approved):
Works as it should! 👍

PR #11074, opened 2021-05-19
Removed unused code in accessed.php. Issue#11067

a18willi:
Issue only mentioned removing ssn, firstname and lastname inputs. But after looking at the code and what it was used for it seemed like the edit user menu code never was used(there werent even a way to open the menu). Hence i've removed the code for the edit user functionality in accessed.php.


a19danla (approved):
Other functions seems to be working fine.

b19wilfo (approved):
Works and as stated there doesn't seem to be a way of accessing the edit user interface so it probably was not used.

b18phika:
There's a function called "closeEditUser" (accessed.js 105-107), that doesn't do anything now that the form has been removed. That should probably be either removed or commented out. 
@a18willi

a19danla (approved):
Works as intended

b19wilfo (approved):
Works as it should.

PR #11084, opened 2021-05-20
G2 2021 w21 issue#11070

a19danla:
Commented out ssn, firstname, and lastname since they're no longer used.


a19samke (approved):
looks like it works

a18danli (approved):
Looks good

PR #11091, opened 2021-05-20
Rename MOTD.SVG to MOTD.svg

a18danli:
Change name on the svg file, sometimes it went to alt tag instead of the svg file, now the problem should be fixed.
@a19samke


a19isada (approved):
Looks good

b16oskni (approved):
Approved

PR #11093, opened 2021-05-20
Fixed incorrect y axis offset for box selection

a19nikbo:
Added an offset on the y axis for box selection of half the height of the element(s)


a19hammi:
LGTM!

Issue #10660, opened 2021-05-05
Create a pop-up window asks if the user wants to continue on saved hash or create new submission


a16erilj:
Currently so is there no list with previous visited hashes in each assignment. It would be great if a popup window that contains all locally stored hashes for the assignment pop-up window when visiting it.

Simply create a pop-up window that contains all visited hashes with dates and links for the assignment you are in.

**Updated description:** It would most likely be better with a would you like to continue on your saved issue or create a new one if a hash is already stored.

b18jacha:
Won't this let users choose the easiest variant?

a16erilj:
@b18jacha Good catch, This solution should maybe not clear the variant and refresh everything else? so that it's not possible to use create new submission for changing the variant.

a18osken:
>Good catch, This solution should maybe not clear the variant and refresh everything else? so that it's not possible to use create new submission for changing the variant.

The Reset button does this already though. I am starting to question the necessity of this issue as there's no real need for a user to have the option to delete their submission when LenaSYS is made in such a way that the latest submission is always the one that is graded. 

Furthermore, when in a canvas iframe, the ability to navigate lenaSYS should be removed. Right now we already return to a locally stored assignment if the save button has been pressed. The user experience feels a bit cluttered as well if we are to have this popup then a log-in popup for password and so on...

This option to continue or start anew feels more and more redundant as  the functionality already seem to exist.

a16erilj:
Seems like it's time to revive this issue since we are going to move load dugga button to the course meny. Which means that this solution would most likely be appropriate again.

b18jacha:
@a16erilj If we start on a new one even tho something is in localstorage. We should get same variant right?

a16erilj:
Yes you should use the same variant that is on the local storage the only thing that should change is the hash.

Here is a mockup how the popup could look like 
![bild](https://user-images.githubusercontent.com/81558891/118822027-145c0400-b8b8-11eb-8c77-39e4d662d65f.png)

PR #11094, opened 2021-05-20
G1 2021 W21 #10660

b18jacha:
Added a button to start a new submission on every dugga. 

**To test:** 
- Go to any dugga and save something, then press the **New submission** button.


e15marol:
To test the load dugga button, simply test it by loading a dugga which is the same duggatype in which you are in, so if you're inside Biträkningsdugga 1, get a hash from a saved Biträkningsdugga 1 and try and load it. It should only work for the duggas that correspond to the same dugga types. So the hash that works for loading a dugga inside Biträkningsdugga 1, should not work when trying to load it in Biträkningsdugga 2 for example.

a19gusba (changes requested):
do not use ajax calls outside the AJAXService function. Figure out what the new submission should contain, what should it do, what things should be added/removed. The variant is already saved in local storage so the item does not need to be removed and replaced. The new submission should not save so do not use the "SAVDU" opt. Why is the setPassword function called?

a19gusba:
image
why not just use the variantValue variable? What is the point of storing a variable in local storage to get in 2 lines below

e15marol:
> image
> why not just use the variantValue variable? What is the point of storing a variable in local storage to get in 2 lines below

The variantValue came out as undefined when implementing the solution, so it had to be temporarily stored and then retrieved.

a19gusba:
image
image

This creates problems since it doesnt use our "localStorageHandler()" function that is used to correctly handle values in local storage

a19gusba:
image
image

Using save dugga option when dugga should not be saved should be avoided!

a17alfja (changes requested):
When new variant when pressing save is fixed, the solution does not work for me
image
(solution in showduggaservice.php) 
I think you can reduce the number of (new) ajax calls by sending values as parameter. eg.
image

a19gusba:
> > image
> > why not just use the variantValue variable? What is the point of storing a variable in local storage to get in 2 lines below
> 
> The variantValue came out as undefined when implementing the solution, so it had to be temporarily stored and then retrieved.

image
image

Tried using console.log to check if this was the case but i would get access to the variable before and after the if statement, doesnt really see how an if statement could turn the variable to undefined unless the "confirm()" function somehow manually changed the variable

a19gusba:
does the ajax call really need to be used in "newSubmission()"?

a19gusba (changes requested):
the "New submission" button should be located on the login screen to a dugga, not the dugga itself. The function "newSubmission" still uses an ajax call

a19rebsa (approved):
It works as described!

PR #11096, opened 2021-05-20
A4 will now scale correctly when pressing the "home" key

a19nikbo:
Added an updateA4Pos() and an updateA4Size() call inside the centerCamera() function


a19karwa (approved):
LGTM!

PR #11097, opened 2021-05-20
Fixed strokewidth

a19steis:
Increased the const strokewidth by 0.5 so we can see the contour on the elements when in 0.25x zoom.


a19nikbo (approved):
Heavy construction right here. Good job! 👍

a19steis:
> Heavy construction right here. Good job! 👍

Thanks! took me forever :)

a19hammi:
Definitive promotion after this. Solid work!!

PR #11100, opened 2021-05-20
Removed white-spaces on lines drawn between elements.

a19matma:
Fixed so there are no more white-spaces on lines between elements.
Should work on both normal lines and double lines.


a19karwa:
As far as I can see, there are no more white spacings between lines and elements no matter how you drag them.

a19karwa (approved):
gg

PR #11101, opened 2021-05-20
MOTD now shows an icon in navheader if closed

a19isada:
MOTD now shows an icon in navheader if closed in courseed page. It will stay that way if you visit another part of the site and then return to the courseed page. 
Co-author: @a18willi @b16oskni

### DIFF
--- a/DuggaSys/courseed.js
+++ b/DuggaSys/courseed.js
@@ -449,11 +449,16 @@ function returnedCourse(data)
 	readonly = parseInt(data["readonly"]);
 	
 	if(motd!=="UNK"){
-		document.getElementById("servermsg").innerHTML=data["motd"];
-		document.getElementById("servermsgcontainer").style.display="flex";
-	} else {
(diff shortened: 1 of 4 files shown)


a19kimgr (approved):
Looks good and works. There is one thing that might be worth looking into, that is if the screen width is below 322px the menu gets a bit scuffed. Maybe add the button to the hamburger menu then? Dont know if this will be an issue since I dont know minimum mobile screen width.

a19klafr (changes requested):
I do not get any buttons in the navheader, or elsewhere when closing the motd on lenasys start-page (courseed.php). Tested in both chrome and firefox.

b18phika:
@a19kimgr The problem with that is that not everybody has the hamburger menu, so that's why it's not in it. We'll discuss whether or not everybody should on the meeting with the customer today. so hopefully that'll be resolved in the near future.

a19klafr (approved):
Looked for wrong item. Rechecked and everything works as requested

Issue #11082, opened 2021-05-20
Combine the title and next variant button in showdugga


a16erilj:
In order to save some space for mobile cellphone so would it be good to combine combine the title and next variant button. To something like Biträkningsdugga 1 V:1 where upon being pressed switches to the next variant. This also give the user some feedback of which variant it's on. 

![bild](https://user-images.githubusercontent.com/81558891/118936686-0a381500-b94d-11eb-9be2-0e010ddd2aa3.png)

PR #11102, opened 2021-05-20
Combine the title and next variant button in showdugga #11082

a19kevkh:
To test: check so that the title of the dugga + variant is a clickable button that gives you the next variant in both desktop and mobile mode.  (Also, remember to clear localstorage before you try it.)


a19frebo:
Looks like the branch is based off #11069 instead of the weekly branch. Should not be an issue but this is why there are more changes than what we did.

a13zanal (approved):
Works as described 👍

a19rebsa (approved):
Works fine, good job! 👍

Issue #11104, opened 2021-05-21
Make it possible to sort after Times submitted and Times accessed


a16erilj:
It seems not possible to sort after Times submitted and Times accessed. 

Make it possible after them.
![bild](https://user-images.githubusercontent.com/81558891/119097279-d3c5cd00-ba14-11eb-8f48-497325460348.png)

a18osken:
could this be because all of the values in those two columns are null?

a19rebsa:
You can sort on times submitted, but not on times accessed

PR #11112, opened 2021-05-21
Sort after Times submitted and Times accessed #11104

a13zanal:
Now it is possible to sort after Times submitted and Times accessed 👍
The sorting on Times submitted did not work as it should, solved this in the compare function so the numbers were sorted in the correct order.

Check out "Edit student results" and click on the headings for Times submitted and Times accessed and you will see masterpieces of sorting!

### DIFF
--- a/DuggaSys/resulted.js
+++ b/DuggaSys/resulted.js
@@ -205,7 +205,7 @@ function renderCell(col, celldata, cellid) {
 function renderSortOptions(col, sortKind, colname) { // Which columns and how they are sorted
     str = "";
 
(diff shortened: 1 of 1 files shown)


a19rebsa:
Also note that if you have null values in the Times submitted-, and Times accessed-columns the sorting will probably not work. In that case, truncate the userAnswer table, then save some new assignments and try to access them/submit them several times to then test out this pull request correctly.

a19kevkh (approved):
Works as stated

a19frebo (approved):
Works!

PR #11125, opened 2021-05-21
Added visual indication of national holidays in barchart. Issue#9768

a18willi:
Indication of national holidays now show in the barchart. Note that currently holidays needs to be entered manually and is probably something that should be automatized in another issue.

Co-author: @a19isada @b16oskni

### DIFF
--- a/DuggaSys/contribution.js
+++ b/DuggaSys/contribution.js
@@ -97,11 +97,31 @@ function showInfoText(object, displayText) {
   }
 }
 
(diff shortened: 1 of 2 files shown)


a19klafr (approved):
Looks good and days show as pink bars. Would be good if there was some explanation to what pink bars mean, maybe a new issue to fix this.

a18danli (approved):
Looks good for me!

Issue #11126, opened 2021-05-21
Add explanation to the legend for the pink bar indicating national holiday


b18phika:

HGustavs:
A screenshot preferably of before and after would be appreciated

PR #11127, opened 2021-05-21
Add explanation to the legend for the pink bar indicating national holiday #11126

b16oskni:
Added holiday to legend

Co-author: @a18willi @a19isada


a19danla (changes requested):
Works good in chrome but can't see any explanation for any of the bars in firefox
image

a19danla:
And in chrome 
image

b19wilfo (changes requested):
The change works on chrome and opera but does not appear to be working on firefox.

a19danla:
When i check on the weekly branch G2-2021-W21 I get the same behaviour, seems like Firefox doesn't like the way its implemented.

a19danla (approved):
It now works on Firefox.

b19wilfo (approved):
It now works on firefox as well.

Issue #11116, opened 2021-05-21
Move the ruler icon one step down


a18nadma:
Keeping the show grid and snap to grid icons next to each other would be ideal!

a19hammi:
fixed in #11129

PR #11129, opened 2021-05-21
Fixed issue #11116 :  moved the button one step down 

a19karal:
the button was successfully moved one step down.


a19matma (approved):
LGTM!

PR #11132, opened 2021-05-24
Removed fab (+) button from diagram ui

a19steis:
Removed the fab button from the diagram, also made sure to remove any unusable code. Especially in toggleReplay as suggested in the comments.


a19hammi (changes requested):
Several references to fab are still present in the code.

a19hammi (approved):
LGTM! Ready for review!

PR #11136, opened 2021-05-24
Fixed the designflaw

a19chrel:
The stateMachine does not save the copied and pasted elements in an array in one StateChange anymore, instead many states are created with the same timestamp.


a19karal (approved):
funkar som det ska :)

Issue #11087, opened 2021-05-20
Shows wrong variant when saving a dugga


b18jacha:
If we clear localstorage and go to a dugga and save it without any page refreshes it will show incorrect variant until next page refresh:

Fresh dugga:
![](https://i.gyazo.com/b8bff540e8a1490f72ffe3bf928cd228.png)

Save:
![](https://i.gyazo.com/a290cce0015b54a8eb757ed8686857df.png)

Refresh: 
![](https://i.gyazo.com/63fccde1ac1e61ee1a1ce5232385cc13.png)

b18jacha:
I think @a17alfja has a solution

PR #11141, opened 2021-05-24
G1 2021 w21 #11087

a19taral:
Added the missed parenthesis

### DIFF
--- a/DuggaSys/showDuggaservice.php
+++ b/DuggaSys/showDuggaservice.php
@@ -194,7 +194,7 @@ if(checklogin()){
 $demo=false;
 if ($cvisibility == 1 && $dvisibility == 1 && !$hr) $demo=true;
 
-if($demo || $hr && !strcmp($opt,"SAVDU")==0){
+if(($demo || $hr) && !strcmp($opt,"SAVDU")==0){
 		
 	// If selected variant is not found - pick another from working list.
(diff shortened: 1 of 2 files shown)


a19taral:
To test: check if the variant in the background changes when a hash pops up in a bit or färg dugga.

a17alfja:
image
You can remove the other part of the if-statement completely (showduggaservice.php)

a17alfja:
nvm

a19rebsa (changes requested):
When going to the hash-url you get redirected to the 404-page, so this solution breaks something.

a13zanal (changes requested):
I end up on page 404 when I enter a saved Url.

a17alfja:
I think the submission is not saved because the variant is unknown when it saves (with "SAVDU" opt parameter).

a19taral:
can you check again now

a17alfja:
I think I have found a solution
image
+
image
(in dugga.js)
The variant value wasn't set before save, so it randomized. Should hopefully work.

a19taral:
> I think I have found a solution
> image
> +
> image
> (in dugga.js)
> The variant value wasn't set before save, so it randomized. Should hopefully work.

worked like a charm. Nice work!

a13zanal (approved):
Now it works just fine! 👍🏻

a19rebsa (approved):
It works! 👍

Issue #11110, opened 2021-05-21
Update the testdata


a16erilj:
Currently when making a fresh installation of LenaSYS so are it missing some information like Times submitted and so on. Check what data is missing and fill in some appropriate data so that after a fresh install that it can be used immediately

This concerns the files/scope that we have handled and not the entire LenaSYS.

PR #11142, opened 2021-05-24
Update the testdata #11110

a19frebo:
Updated all the testdata that is inserted into the userAnswer table. Every row now has a value in the timesSubmitted and timesAccessed columns.


a19frebo:
Also fixed all the "broken" entries in the table so they should work correctly now when entering the dugga through the hash.

a19kevkh:
To test, simply truncate the Useranswer table and then run the inserts that are in the testdata.sql file in the lenasys/install/SQL folder. Then go into the edit student result page and check so that they all have the correct dugga name and the values are correct.

a19arvka (approved):
Works, all tables in resulted have a value and that value matches the one in my database after I truncate and execute testdata.sql

a13zanal (approved):
Looks good 👍🏻

PR #11143, opened 2021-05-24
Label on lines between relation and entity is possible

a19aleda:
It is now possible to add a label between relation and entity.


a19karwa (approved):
Labels works fine, good job!

PR #11144, opened 2021-05-24
Made it possible to submit and load diagram dugga

a18osken:
It is now possible to save and load diagrams. Now when the user makes any change to the diagram iframe it is possible to save the assignment to the database just as any other file submission. We skipped the inclusion of an upload field since it is strictly not necessary when we create the file ourselves from the state of the diagram.

It currently only saves the JSON file.


a18osken:
Keep in mind that while testing you need to update the database

a13zanal (approved):
Looks like everything works as described 👍🏻 Well done!

a19rebsa (approved):
It works as described, good job 👍

a19matma (approved):
Saving and loading seems to work as intended. Good job!

a19hammi:
Would love if #11153 was reviewed and merged before this one is. @a19chrel @a18osken

PR #11145, opened 2021-05-24
Position values is now without decimals

a19chrel:
As the topic said, the coordinates values is now without decimals


a19karwa (approved):
LGTM!

PR #11149, opened 2021-05-24
Circles in line graph now indicate size of commits

a19isada:
Circles in line graph now indicate size of commits in contribution. 
Co-author: @a18willi @b16oskni


a20chrjo (approved):
Works in both Chrome and Firefox

b19alest (approved):
Works good.

PR #11151, opened 2021-05-24
Added the ability to change fill and stroke color of elements

a19nikbo:
You can now change the fill and stroke colors of elements from the options panel. This has also been added to the State Machine. The undo functionality is, however, a bit broken so it will be difficult to test (hopefully fixed with pull request: #11153).

**Note that the color options are only placeholders and will most likely need to be revised for accessibility purposes.**

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -357,8 +357,52 @@
     z-index: 2;
 }
 
+
 .diagramIcons:hover .toolTipText{
     visibility: visible;
+}
+
+.colorMenu{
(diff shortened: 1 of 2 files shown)


a19nikbo:
Forgot to remove a big chunk of comments

a19nikbo:
Some changes were made to how elements are created and some old comments were left in there. They have been removed now

a19steis (approved):
Looks good, being able to change the colour of the elements is a nice addition. Good job!

PR #11153, opened 2021-05-24
Potential fix: ctrl-z bug

a19matma:
ctrl-z should work as intended now.

Co-Authored-By: Hamlet Magnusson Millqvist <81571224+a19hammi@users.noreply.github.com>

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -372,6 +372,7 @@ class StateMachine
      */
     save (stateChangeArray, changeType)
     {
+        
         if (!Array.isArray(stateChangeArray)) stateChangeArray = [stateChangeArray];
 
         for (var i = 0; i < stateChangeArray.length; i++) {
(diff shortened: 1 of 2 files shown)


a19matma:
Also purged some console.logs that were left behind

a19hammi:
Patch for pull-request #11144 
Need review from original authors. Usually I should not participate in making code but this one is an exception.

a18osken (approved):
I am not aware of what the initial bug entailed, but ctrl-z seems to work as intended now.

a19chrel:
> 
> 
> I am not aware of what the initial bug entailed, but ctrl-z seems to work as intended now.

The problem was that when an stepback (CTRL + Z) was executed, you moved multiple steps in the history. Even if you only should move one.

a19chrel:
There is one problem still, the load now gets in the wrong place, because the scrubHistory has some changes, this can easly get fixed by removing the `+ 1 ` on this line. This was an mistake by me that should get removed. I suggest that we do that in the patch for the solution.
https://github.com/HGustavs/LenaSYS/blob/daae9a670a035c7873481f36619f2b66efe57098/DuggaSys/diagram.js#L4909

a19chrel:
Also i suggest we move the replay-calculations from save() in stateMachine to toggleReplay(), because the replay was made before it was possible to import / export. And now does the values for replay mode do not update if you dont make an change.

https://github.com/HGustavs/LenaSYS/blob/4591e75e2176eaf989341de0674894fb9af39dab/DuggaSys/diagram.js#L485-L495

a19chrel (approved):
LGTM, seems to have done the trick.

PR #11156, opened 2021-05-25
"Fixed" Label string should be trimmed of trailing and heading whites…

a19phith:
Fixed by adding trim method to the labels


a19chrel (approved):
Looks good, works as intented. Nice that you change the input field as well.

PR #11162, opened 2021-05-25
"Fixed" Being able to save by pressing enter

a19phith:
fixed by adding the fucntion to the enter keybinding


a19chrel (approved):
Do what it should! Nice work!

a19chrel:
Maybe use blur on the input-field to deselect it. Like we do in the "enter on namechange".

Issue #10619, opened 2021-05-03
Lines should have different connection points on recursive relationships.


a19frelu:
It would be easier to select the different lines and it also looks a lot better if the lines have different connection points as shown under. 

How it is now.

![Skärmavbild 2021-05-03 kl  16 36 11](https://user-images.githubusercontent.com/62895155/116890658-2722d780-ac2e-11eb-9dfa-66d2151e7610.png)

What it should look like. 

<img width="485" alt="Skärmavbild 2021-05-03 kl  16 36 35" src="https://user-images.githubusercontent.com/62895155/116890723-343fc680-ac2e-11eb-9d25-494118c1b857.png">

a19matma:
True, it would look nicer, but getting this to work with our automatic repositioning on the lines when elements are moved around will be quite difficult. More discussion needed imo.

a19frelu:
This question has come up for discussion again after the request has come from the customer to implement this.

a19hammi:
fixed in #11198

PR #11163, opened 2021-05-25
Changed recursive diagrams line positioning

a19vikth:


b19jenda (approved):
Seems to work as intended!

a19hammi (changes requested):
While your calculations are VERY GOOD, the point is that the points are supposed to be separate ones. Each one should be able to use the label system later and the point is to check if "more than 1 line" is connected to an element, then use the new positioning with the lines.

This option looks really cool but it wasn't really meant for the user to manually check such a box. It's supposed to be automatic.
Screenshot 2021-05-25 151333

a19hammi:
will be published via version 2 **Group3-2021-W21-#10619-v2**

Issue #11130, opened 2021-05-21
Bug: Element missplaced during placement mode in diagram


a19hammi:
Right now the elements are not centered at the cursor along the y-axis as it once was in previous branches. The element kind of balances with its bottom on the cursor.

PR #11165, opened 2021-05-25
Group3 2021 w21 #11130

b19jenda:
Now when moving elements in snapToGrid the elements will move along with the mouse and snap to where the mouse is positioned.


a19phith (approved):
LGTM

PR #11166, opened 2021-05-25
Update sectioned.js

a18danli:
Hide the arrows if the swimlane is hidden.


a19danla (approved):
Works as intended. Tested on different browsers

b19wilfo (approved):
Works, tested on firefox and chrome.

Issue #11105, opened 2021-05-21
Highlight a row that have a extremly high number times accessed 


a16erilj:
Would be great to see if something that have been accessed an absurd amount of times. So create a function that highlights these rows.

![bild](https://user-images.githubusercontent.com/81558891/119098239-e8569500-ba15-11eb-9423-97bfb65964e5.png)

a16erilj:
A high amount of accesses could be somewhere around 20-30 according to our customer and not 200. This also it depends on the assignment type... so the options to change it should be kept for now.

PR #11167, opened 2021-05-25
Highlight a row that have a extremly high number times accessed #11105

a19kevkh:
Added search bar to highlight rows with specified number of times accessed. 

To test: simply input the value that you want to highlight in the edit student results page and then click the enable button and the rows with the same/greater values will be highlighted.


a19arvka (approved):
Works, but is the color red correct to use in regards to having any color deficiency.

a13zanal (approved):
Works great, but interesting choice of color ^^

a19rebsa (commented):
It works great, but wondering if the color is the ultimate one for this. Otherwise good job!

Issue #11157, opened 2021-05-25
Add which dates the bar chart is showcasing


a19isada:
Add the start and end date of the bar chart.
![image](https://user-images.githubusercontent.com/81559946/119464070-b0678e80-bd42-11eb-85c3-6f51485b0c1f.png)

PR #11169, opened 2021-05-25
Add which dates the bar chart is showcasing #11157

b16oskni:
Added functionalty to show from which dates the data is gathered.

co-auth: @a19isada @a18willi


b19alest (approved):
Seems to work good 👍

PR #11171, opened 2021-05-25
Now able to delete items even when multiple items are selected. Issue#11170

a18willi:
Removed broken condition for deleteitems function

Co-author: @a19isada @b16oskni


a19klafr (approved):
Works to delete one item while others are marked now!

a19kimgr (approved):

PR #11175, opened 2021-05-25
Added label changes to the state machine

a19nikbo:
Changes to the label text are now added to the state machine and can be undone/redone.


a19karal (approved):
funkar som det ska :)

PR #11181, opened 2021-05-25
Diagram-dugga removes future states before saveing.

a19chrel:
The future states are now removed before saveing the diagram-dugga.


a19phith (approved):
looks good, works as intended

PR #11182, opened 2021-05-25
Fixed default colors for pasted elements

a19nikbo:
Elements are no longer completely black when pasted


a19steis:
Default colour now works correctly when copy pasting. But dont we want the copied element to retain its colour when it is copied?

a19nikbo:
> Default colour now works correctly when copy pasting. But dont we want the copied element to retain its colour when it is copied?

True, you can make a change request and I'll fix it ^^

a19steis (changes requested):
Awaiting fix.

a19nikbo:
Changes made

Issue #11086, opened 2021-05-20
date interval don't containts the date in resulted


a16erilj:
Currently when visiting resulted so are the dates in the interval not filled in. Make it so that they are filled in when visiting the first time with the maximum interval between the submissions. Also rename them to a more fitting name so that they are not named date 1 and date 2.
![bild](https://user-images.githubusercontent.com/81558891/118943658-1b385480-b954-11eb-81ab-b21efca9196d.png)

PR #11189, opened 2021-05-26
G1 2021 w21 #11086

B18mohah:
The date range is now set aside for the earliest and most recent submitted tab


a19kevkh (approved):
Excellent

a19frebo (approved):
Works fine! 😄

PR #11196, opened 2021-05-26
Restriction for lines for Recursive relations.

a19chrel:
it is not possible to create a line from/to a relation that has an recursive relation or possible to create an recursive relation if the relation is connected to an other element.


a19matma (approved):
LGTM! Tried to break it but it seems to work as requested.
Question is if we should have a more descriptive "error"-message:
image

PR #11198, opened 2021-05-26
Fixed: Lines should have different connection points on recursive relationships. #10619

a19vikth:


b19jenda (changes requested):
There are many errors occuring when clicking the lines and changing to recursive, also there are many console.logs left

a19chrel (changes requested):
You need to look over the offsets. There is something with the connection points that isent correct.
bild

a19matma (approved):
LGTM!

b19jenda (approved):
Looks good

a19chrel (approved):
Looks good after the changes.

Issue #11203, opened 2021-05-28
Contribution stopped working after final merge because of an error in sortableTable.js


a19anjst:
Something is wrong at row 406 with checked. If checked is removed it works. Unsure why checked is used. 
![Skärmbild (441)](https://user-images.githubusercontent.com/62884743/119959872-ae9d0580-bfa4-11eb-9c29-0db1c8beb942.png)

c20adael:
Undersök detta problem. Är detta löst?

a20behta:
old issue closing as no longer relevant

PR #11204, opened 2021-07-19
Contribution stopped working after final merge because of an error in sortableTable.js #11203

b16oskni:
The contribution page couldnt load because of a non-existing html-element.

Added a statement that checks if the html-element actually exists before running.


a20behta:
@HGustavs What should we do with these requests from previous years?

a20behta:
Placed on hold untill information given about treating old pull requests.

a20behta:
Closed as old request outdated

Issue #11148, opened 2021-05-24
New items are added in the wrong place because of the position value


a19anjst:

b19wilfo:
It might be possible to move the new items after they are created, if we drag and drop new items they get the right position value.

a19danla:
When newly created items are added to a course they're placed at a position visible on screen. The value for where it should be placed is saved in the variable pos. The problem seems to be that the value pos doesn't get updated correctly before you move the item which means that if you create several new duggor in a row, without moving any of them, they will be created with the same pos-value, even if there's already a dugga with that value.

This is how the list should be: 
![image](https://user-images.githubusercontent.com/81560421/119352207-b0638200-bca1-11eb-8446-0020b4da2f73.png)

Every dugga has a unique posvalue.

This is how it looks when adding new ones:
![image](https://user-images.githubusercontent.com/81560421/119352295-cc672380-bca1-11eb-9a6f-f415a7785d85.png)

We're not 100% sure how to fix this currently as the old code just placed the new dugga at pos -1 or 100 depending on which fab-button used.

A potential solution could be to call the "move function" to move the dugga to the right position.

or

Make sure that the list containing all duggor updates correctly when adding a new dugga.

co-author: @b19wilfo

a19danla:
After some testing it seems like the pos-value wont change if you edit the dugga, it only occurs when you "physically" drag a dugga to a new place in the list. Doing so will update the pos-values for all items to their correct values.

b16oskni:
The current way to assign a position value to a new item is done through calculating a value based on the static fab-buttons relationship to the viewport. This creates unexpected behaviour when the page is scrolled to the bottom of the list but items may end up higher in the list than if you are viewing the top of the list.
![bild](https://user-images.githubusercontent.com/62884511/126352617-b975a9e7-7114-428c-bb8b-625de12f689c.png)

Since desired behaviour is not specified a temporary solution will be made where the new item is placed either at the bottom or top of the list depending on where the viewport currently is.

PR #11205, opened 2021-07-20
New items are added in the wrong place because of the position value #11148

b16oskni:
Removed unnecessary math and changed values for the position of new items so that they are listed at either the top or bottom of the list depending on the position of the viewport.


a20behta:
@HGustavs should we still treat and confirm pull requests.

Issue #11214, opened 2022-03-31
Message of the day (motd) gets overlapped by the top bar


a16linas:
When browsing a course, the message of the day is partially covered by the top bar, see attached image below

![bild](https://user-images.githubusercontent.com/37794783/161027307-b417d952-e33e-4da3-b82a-79c49001b0bf.png)

a16linas:
Solution to issue has already been found, but we're waiting until the format/structure for how the groups will handle their branches have been determined. Will create branch and push solution when that is done.

a16linas:
the #motdArea had a margin-top with -16px that caused them the overlap. Changed it to 16px and the top bar and message of the day are no longer overlapping

PR #11260, opened 2022-04-04
in the #motdArea there was a margin-top with -16px that is now changed to 16px

a16linas:
In issue #11214 the message of the day was overlapping with the top bar. The margin top has now been adjusted so they're no longer overlapping.


b20thelo:
Tested on server and approved.

PR #11261, opened 2022-04-04
added an if case that checks if versarr is not empty at showdugga.php:189

a16linas:


b20thelo:
Tested on webserver. Confirmed that it now works.

PR #11262, opened 2022-04-04
Reset button was missing from diagram

a16linas:
The function for zoom reset already existed, but there was no button for it in the diagram. Made a new button and used the fullscreen svg icon that best matched the request (youtube full screen button).


b20thelo:

PR #11270, opened 2022-04-04
9720 hidden class exists in both stylecss and codeviewercss

a20mtafa:
i fixed the issue by changing the float status to left instead of none.


a20willa:
Cannot be replicated. Closing the request.

PR #11271, opened 2022-04-04
Fixed vanishing example text in codeviewer

a18gusjo:
Added the function updateBoxes to clean up resetBoxes function and made some changes in resetBoxes. Also fixed the getTotalBoxes which always returned null.


a20willa:
Reset button does not work at all in the test server. This needs to be fixed.

a20willa:
Reset button needs to be fixed on the server.

PR #11274, opened 2022-04-04
Removed alert dialog

b20idawa:
Removed alert dialog when something has been updated


a20sarle:
Closing branch and will have a look at it next nweek since the issue and solution couldn't be replicated during test.

PR #11277, opened 2022-04-04
Confirmation when changing templates

a20joabo:
Added a confirmation when selecting a template in the code viewer to avoid unwanted overwrites.


a20willa:
This one has been checked but will be put on hold for now.

a20willa:
This does not actually solve the issue which is why it was declined. We need a system which stores the options, and altough the overwrite notifications are a good starting point, we may not accept it as it's not relevant to the issue.

PR #11278, opened 2022-04-04
Fixed incorrect double click not counting as a single click and preventing properties from appearing.

a20micro:
Updated the generateContextProperties() function to not return wasDblClicked at the start as it was always true in case of a double click, no matter where it happened, preventing the second click from acting as a single click.


b20thelo:
Tested on the webserver. Confirmed that it now works.

PR #11281, opened 2022-04-04
Text input validation fix #8528

### LINKED ISSUE #8528: Edit-windows are not consistent across the system
All Edit-windows across the system should look and behave consistently. The first screenshot shows an example from the Edit Dugga-window from the Dugga Editor. The second screenshot is from the Edit Item-window from the Section Editor. The screenshots show how a similar (or even identical) "error" is handled differently for the two windows.

**Edit Dugga/Dugga Editor.** When incorrect/not allowed characters are input the whole text field is marked with a different color and the Save-button is disabled.
image

**Edit Item/Section Editor.** When incorrect/not allowed characters are input only the border of the text field is marked with a different color and the Save-button is NOT disabled, letting the user save even if the text contains incorrect characters (this is a separate issue, #8482).
image

i19matla:
The text validation is now the same for both the different popouts.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -889,7 +889,7 @@ const zoom1_5 = 0.555;
 const zoom2 = 0.75;
 const zoom4 = 0.9375;
 const zoom0_75 = -0.775;
-const zoom0_5 = -3;
+const zoom0_5 = -3; 
 const zoom0_25 = -15.01;
 
 // Arrow drawing stuff - diagram elements and diagram lines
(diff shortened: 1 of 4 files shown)


a19meram:
Adding a change to input of validation at DuggaSys/diagram.js
DuggaSys/duggaed.js
DuggaSys/sectioned.js
DuggaSys/sectioned.php

b20thelo:
Pull request will not be accepted. Comment in original issue #8528

Issue #11184, opened 2021-05-25
White background of label not aligned properly


a19nikbo:
It's difficult to see, but if you look above the 'W' on the line you can see that the cutoff happens too far to the right.

![image](https://user-images.githubusercontent.com/81541557/119518506-dc9e0200-bd78-11eb-8415-a69e98e0b65e.png)

a20gabpa:
Found an issue where the background moves upwards when zooming out, changed background color for better visibility.

Position of background at zoom factor 4:
![WithZoomfactor4x](https://user-images.githubusercontent.com/102604680/161216882-3d349492-d2c2-4d75-9d61-454c20b7ba55.png)
Position of background at zoom factor 0.5:
![WithZoomfactor05x](https://user-images.githubusercontent.com/102604680/161216890-e12d38b3-02f2-4d0e-82f1-e9851e6aa5a1.png)

b20thelo:
Issue solved and solution pushed to master branch.

PR #11282, opened 2022-04-04
Fixed malpositioned background for labels

a20gabpa:
Issue #11184 

Updated the drawLine() function in diagram.js which handles the positioning of the white background behind the text. Implemented a way to get the width of the text through the use of canvasContext.  The width is then used to position and scale the background accordingly.


b20thelo:

b20thelo:

Issue #11240, opened 2022-04-01
Refactoring is needed in showDugga.php


c20adael:
There is a lot of commented code that needs to be checked and could be removed.

HGustavs:
Yes.

This is a task for all files not just this file specifically.

HGustavs:
Related to issue #11242

a17nikjo:
Starting to refactor the file and remove commented code that serves no purpose.

a20novra:
jag jobbar på det här @a20behta

a17nikjo:
54   - 94   related too setting up for the password promt
152 - 168 labeled as old code, primarly does console logging. - Code removed.
206 - 250 creates a login promt

a20vicad:
Some of the commented out code was seems to hold a testing purpouse and this gets to stay.
![image](https://user-images.githubusercontent.com/102599327/164203844-b303bb76-33d0-4cd4-acf7-7bde46022c51.png)

PR #11283, opened 2022-04-04
G3 2022 w1 #11240

a17nikjo:
Removed 1 of 3 sections of out commented code. The other sections are unclear if they should be removed.


c20adael (commented):
Unnecessary changes to the comments

PR #11294, opened 2022-04-04
Change of default word-list for Preview.

a20joabo:
Changed the default word-list for "Preview" to Plain text from undefined.


a20willa:
Create a new code item and choose the "preview" option on the dropdown under "Kind" to see the changes.

Issue #11251, opened 2022-04-01
Diagram research file Diagram_IOHandler.js


b19karhj:
Make a summary of what Diagram_IOHandler.js does and affect at the top of the file.

b20thelo:
Changes only apply to a comment which isn't testable. Approved for merge.

b20thelo:
Issue solved and solution pushed to master branch.

PR #11297, opened 2022-04-04
Updated the summary in diagram_IOHandler.js

a20samul:
Clarified in the top summary where the code was applied. #11251


b20thelo:
Changes only apply to a comment which isn't testable. Approved for merge.

Issue #11236, opened 2022-04-01
Clicking redirect camera to center sets zoomfactor to 1x


a20edwth:
Clicking the redirect cemare to center button allways sets the zoomfactor to 1x. This can be bothersome if the diagram is on the larger side. 
Before:
![ce99476f58b54861835c5c9bc804a770](https://user-images.githubusercontent.com/102611046/161227843-15a7ebbb-d3d4-4ab0-ad08-13ab7fe11ffb.png)
After:
![d36fa97b0c46870054b86167d2376f27](https://user-images.githubusercontent.com/102611046/161227873-c3e95d8b-907b-4875-986f-7ccfa5d4374c.png)

b20thelo:
Tested on the webserver. Confirmed that it works as it should.

b20thelo:
Closing this issue as it has been solved.

PR #11299, opened 2022-04-04
No zoom when redirect camera #11236

a20edwth:
The redirect camera to center button will not zoom in to zoomfactor 1x.


b20thelo:

PR #11303, opened 2022-04-04
FIxed the issue.

a20patpe:
The style.display i set to "block" for the copyclipboard object when the reset button is clicked. Changed so it sets it to table-cell instead. (table.cell is the default for the icon.)


a20willa:
Does not work in server side.

PR #11308, opened 2022-04-04
added functionality to addLine

b20matad:
All lines goes from EREntity to make multiple line offset better.


b20thelo:

Issue #11305, opened 2022-04-04
Clean up courseedservice.php


a20samul:
#11242  Cleaning up code in file courseedservice.php

PR #11309, opened 2022-04-04
G1 2022 w1 #11305

a20samul:
Cleaned up code in courseedservice.php


b20thelo:

PR #11322, opened 2022-04-04
made a txt file LenaSYS\AllCodeStandardIssuesInInstall.txt

a19simru:
this file contains a list of all files and folders in LenaSYS\install with a review on their compliance with the code Standard. As well as a recommendation for wheatear the code should be fixed.


a20behta:
Do not place separe text file with the issue in lenasys but rather upload them into the issue.

PR #11324, opened 2022-04-05
Fixed darkmode in code examples

b20idawa:
Fixed darkmode in code examples


a20sarle:
Solution shold work in theory but can't be replicated in webbserver. Waiting untill' next week merge and more information about if merge should be done in this case or not.

Issue #11332, opened 2022-04-05
Clean up showDuggaservice.php


a20edwth:
Remove commented code. #11242

b20thelo:
Code where tested before issue was closed and merge request approved.

PR #11334, opened 2022-04-05
Clean up of showDuggaservice.php #11332

a20edwth:
Removed comments.


b20thelo:

b20thelo:

Issue #11335, opened 2022-04-05
Clean up showdoc.php


a20edwth:
Remove commented code. https://github.com/HGustavs/LenaSYS/issues/11242

b20thelo:
Approved by testing group.

b20thelo:
Issue solved and solution pushed to master branch.

PR #11336, opened 2022-04-05
Clean up of showdoc.php #11335

a20edwth:
Removed comments.


b20thelo:
Approved by testing group.

Issue #11252, opened 2022-04-01
Diagram research file diagram_IOHandler.php


b19karhj:
Make a summary of what diagram_IOHandler.php does and affect, at the top of the document.

a20jonda:
#11296 same file issue

a20jonda:
Wrote a description in the top of the file on the current functioning stuff the file does and added small dividing descriptions for the looping and print:ing php for better readability. Then added a comment on the small and not declared javascript part which purpose is not yet understood.

b20thelo:
Pull request merge 1 was denied since it does more than what was requested and also affects code structure. Will not be included in this week's merge.  Revert structural changes of code. Ask group leader in case of questions.

PR #11341, opened 2022-04-05
G1 2022 w1 #11252

a20jonda:
#11252. Alot of features are not yet functioning and only the functioning part is in the description.


b20thelo:
Pull request denied since it does more than what was requested and also affects code structure. Will not be included in this week's merge.

PR #11346, opened 2022-04-05
Missing semicolon

a20joabo:
Added missing semicolon


a20willa:
Semicolon was missing in file. Cannot be tested in lenasys as the element does not exist.

a20willa:
Canot check if the fix works yet. Closing branch,

Issue #11342, opened 2022-04-05
Clicking the "Zoom IN" and "Zoom OUT" buttons in Diagram Dugga zooms from origo.


a20edwth:
The buttons' zoom points are in the origo(top left corner) rather than in the center of the screen.

b20thelo:
Tested on the webserver. Confirmed that it works as it should.

b20thelo:
While testing this feature we agreed that it was better than the existing solution from before, but still isn't the final solution. Needs a bit more tweaking. 

The feature uses the screen width as a coordinate instead of counting the actual coordinate in the center of the current screen.

PR #11348, opened 2022-04-05
Zoom buttons zoom from center of screen #11342

a20edwth:
The zoom buttons will now zoom from the center of the diagram.
NOTE: in some situations, mainly if the diagram is not altered, the zoom will go to the side when zooming out.
4bbcda6357b5b2667c21d722a30e8c97
After pressing "Zoom OUT"
050ca54f477dd85e8161631d36889121


b20thelo:

PR #11354, opened 2022-04-05
Removed unnecessary comments.

a20kimar:


a20behta (approved):
Looks good, no change noticed in webserver version

Issue #11315, opened 2022-04-04
Confirmation when adding an element as a teacher


a20willa:
Adding something as a teacher does not actually confirm it, which may cause confusion if the button even did something. Add some kind of notice which tells that the element has been added!
![image](https://user-images.githubusercontent.com/81626616/161606673-4c4c6fe4-10c8-4cbd-83eb-a09fe35be4c0.png)

b20idawa:
1. Log in as a teacher 

2. Choose webprogramming course

3. Make sure that the page has cleared the cache and made a hard refresh by opening the inspector and right-click on the update icon in the browser and choose "clear cache and hard refresh".
<img width="386" alt="Skärmavbild 2022-04-07 kl  11 09 58" src="https://user-images.githubusercontent.com/81676911/162164252-24ea8200-7fe9-47cf-8ffa-cd6880ff9f03.png">

4. Click on the yellow button at the bottom right of the page and add an item of your choice.
<img width="805" alt="Skärmavbild 2022-04-07 kl  11 10 57" src="https://user-images.githubusercontent.com/81676911/162164379-efde4904-4815-4e92-aca3-dcf49b19c3b6.png">

5. When a new item has been added, a notice should appear at the top of the page as shown in the picture below
<img width="1299" alt="Skärmavbild 2022-04-07 kl  11 12 20" src="https://user-images.githubusercontent.com/81676911/162164650-f2aae8d0-7b08-4e55-8ac7-ae61d6ce791e.png">
.

PR #11356, opened 2022-04-05
G2 2022 w1 #11315

b20idawa:


a20sarle:
Works locally but not yet on the server during testing, reason unclear. Will look at it again during next weeks merge.

Issue #11227, opened 2022-04-01
Superuser/teachers cannot view duggas when logged in


a97marbr:
Superuser/teachers cannot view duggas when logged in:

![image](https://user-images.githubusercontent.com/4948661/161215910-47689a2e-78be-48ad-9766-5ccb654b4c74.png)

The system gives an error as shown above and end up serving an empty dugga.

![image](https://user-images.githubusercontent.com/4948661/161216107-8f424941-a3fe-42ec-a580-771e8ba8c709.png)

a20behta:
@b20jonlu use the branch for continued development

PR #11370, opened 2022-04-05
Origin/g3 2022 w1 #11227

b20jonlu:
Fixed the issue of duggor not properly showing up. The previous code used the variable hash to select database entries which didn't work since no entry was associated with it.  Instead the variable moment is used to retrive all entries for that dugga. The solution might have solved part of the issue but further testing with a larger dataset is needed to confirm.

### DIFF
--- a/DuggaSys/showDuggaservice.php
+++ b/DuggaSys/showDuggaservice.php
@@ -315,9 +315,13 @@ if(isSuperUser($userid)){
 		unset($param);
 		if (isSuperUser($userid)){
 			if($hash!="UNK"){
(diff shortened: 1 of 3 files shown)


b20jonlu:
> Fixed the issue of duggor not properly showing up. The previous code used the variable hash to select database entries which didn't work since no entry was associated with it. Instead the variable moment is used to retrive all entries for that dugga. The solution might have solved part of the issue but further testing with a larger dataset is needed to confirm.

Forgot to mention that the solution for #11290 was merged since it solved the error in one of the files related to the issue. The compare therefore also shows those changes. The changes made to the code by me are those in the showDuggaserive.php and the console.log in dugga.js and the generic dugga js file

PR #11389, opened 2022-04-06
Tooltip functionality added to duggaed.php search bar

a20johba:
Commented out the previous search bar in duggaed.js, this can be removed when the new search bar has been tested properly. Added a new search bar by including the navheader.php search bar. Tooltip functionality added by copying hover-related functions from accessed.js to duggaed.js.


c20adael (approved):
this seems to work as intended. Next time i need a guide on how to test it.

PR #11390, opened 2022-04-06
New function updateBoxes() handles the issue

a18gusjo:
Moved out the functionallity inside the if-statements in resetBoxes. It is now excecuted everytime resetBoxes is called.
This is happening in the function uppdateBoxes which is called upon in the beginning of resetBoxes();

Testing is possible in lenaSys -->codeviewer HTML5 example 1

### DIFF
--- a/DuggaSys/codeviewer.js
+++ b/DuggaSys/codeviewer.js
@@ -3177,6 +3177,17 @@ function hideMaximizeAndResetButton() {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


a20linth:
I cannot recreate the issue (when the solution for the issue is not implemented) when testing on the LenaSYS web server. And I do not know how to determine if the solution solves the bug.

a20willa:
Cannot be replicated on webserver.

Issue #11259, opened 2022-04-01
Two finger trackpad movement in diagrams on Mac zooms crazy fast


HGustavs:
When two finger gesture moving on diagram on Mac trackpad, the diagram zooms but so fast that the zooming feature is nearly useless.

Unclear if same thing happens when one uses two finger gesture on PC

HGustavs:
Is it possible to define two finger guesture as scrolling and pinch as zoom (seems to at least partially work like that in for example google docs)

a20jonda:
Same zoom problem on Windows 11 (both scroll zooming and pinch zooming)

a20jacmi:
Scroll and pinch are working as suggested. However, there is no general way to make a distinction between the utilisation of the mousewheel or the trackpad, that will work on all browsers.

We're working on a solution that will slow down the zoom-function. It looks pretty good but we're giving it some more time to see if we can make it better.

b20thelo:
Sets a timeout that doesn't allow zoom events to trigger more than once every 75 ms. To slow down zoom on a touchpad.

PR #11391, opened 2022-04-06
Time intervall added #11259

b20matad:
Sets a timeout which doesn't allow zoom events to trigger more than once every 75 ms. To slow down zoom on a touchpad.


c20emili:
Tested the zoom with my computer trackpad and did not notice any difference in zoom speed. Did not notice any errors in the code.

a19meram:
used my computer and tested trackpack of zooming effect. Can't see any diffrent between the local version and server, the semantical version of the code have no errors

b19karhj:
Based on the test results the issue need to be corrected.

b19karhj:
The test turned out to be incorrect. The solution works and has now passed the test.

b19karhj:
Merge complete.

PR #11412, opened 2022-04-07
Scroll zoom in/out function is now mostly fixed #11058

### LINKED ISSUE #11058: Zooming while holding an element causes problems
The element slides away from the mouse pointer when zooming.

https://user-images.githubusercontent.com/62895155/118795976-70fcf600-b89b-11eb-9c09-6dc3e127838a.mov

a16linas:
Scroll will now zoom in and out and remain at the current mouse pointer coordinates. It still has some issues that occur if user moves the mouse in between zooms but it corrects itself after one zoom event. This is testable by opening any diagram and scrolling in and out, it should now remain at the same position.

What currently isn't working as intended as mentioned is alternating between move > zoom-in and move > zoom-out. However, the original issue is that when selecting a box and zooming in, the cursor moves away from the box that isn't updating it's position. Instead of updating the element position, the mouse should remain at the same place. 

If any future developer solves the issue of alternating between zoom and move, it should solve the issue about the element moving away from the cursor completely. This commit should be considered a basis for a full solution.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -872,6 +872,8 @@ var scrollx = 100;
 var scrolly = 100;
 var zoomOrigo = new Point(0, 0); // Zoom center coordinates relative to origo
 var camera = new Point(0, 0); // Relative to coordinate system origo
+var lastZoomPos = new Point(0, 0);
+var lastMousePosCoords = new Point(0, 0);
 
 // Constants
 const elementwidth = 200;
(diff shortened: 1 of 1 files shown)


a19meram:
Tested with zooming holding an element and without element. Zooming holding an element not solved yet, but some steps have been solved. Code have no errors, semantical correct!

b19karhj:
Code is correct and progress has been made. This can be merged but a new issue with the problems remaning should be created

b19karhj:
Merge completed

Issue #11325, opened 2022-04-05
Moving timer.js and clickcounter.js too dugga.js


a17nikjo:
When checking if the files timer.js and clickcounter.js existed in issue #1911, it was found out that they had already been removed. However they are still referenced by the following files. 
duggaed.php
showDugga.php
testDugga.php
validateHash.php

Check if this causes any problems and if they can just be removed / commented out.

a17nikjo:
Files appears to exist, on hold while working on issue #1911.

a17nikjo:
In issue #1911 we found out the files are still in use but could be moved therefor issue changed to moving, we still have to make sure it works afterwards to make sure backwards compatability. We also also have to make sure all places that implement timer.js and clickcounter.js implement dugga.js.

a17nikjo:
Possible sulotion of migrating files done. Needs futher testing before being merged.
timer.js and clickcounter.js content moved into dugga.js.
duggaed.php
showDugga.php
testDugga.php
validateHash.php
All implement dugga.js however still needs testing to make sure everything works after migration.

Testing currently not posssible due to lack of duggas. On hold until testing can be done.

a20vicad:
After migration, initally everything appears to function properly. Keeping an eye on it for further testing.

a20behta:
merged into master during weekly merge

PR #11414, opened 2022-04-07
G3 2022 w2 #11325

a20vicad:
Migrated files timer.js && clickcounter.js to dugga.js

Awaiting testing with duggas.


a20behta (approved):
looks good, code commented out and closed correcty. 
moved to dugga.js ,

note: includes for clickcounter and timer still exits however should not affect anything.

PR #11425, opened 2022-04-07
Removed alert dialog

b20idawa:
Removed alert dialog when a new item has been updated


a20linth:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #11426, opened 2022-04-07
fixed resize problem template 4

a18gusjo:
Fixed the problem by removing the "alignBoxesWidth" call in the resize for box2 and the event that is handled in box1.

Testing is done in javascript example 2 which uses template4 (two small boxes on top of one large box). Try to resize the boxes and see if the bug remains.


a20linth:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #11429, opened 2022-04-07
Update codeviewer.js

b20sebka:
CSS attributes are now reset to default when clicking the reset button. Should fix and white box issues that occurs when clicking the reset button.


a20willa:
Pull request against master by accident.

Issue #11359, opened 2022-04-05
File Editor Page wont load files.


a17nikjo:
Issue based on problems found when examining issue #9491.

Possible Solution:
fileed.js row 572: if statement checks if variable is not null but does not check if it is of string type.
Then converts variable to uppercase without knowing if its a string.
if (row[key] != null)  ->   if (row[key].typeOfItem == String)    is a possible solution.

fileed.php: row 67,68 ace.js is commented out and not initialized however it is used in fileed.js uses it. Enabling it seems to solve the issue? Why is this disabled? What is ace.js?

c20adael:
on the webserver it seems to load fine but i cant view fileed.php on my local server.

a19simru:
@c20adael @a20behta Simon would like to be given this If it is still available.

HGustavs:
ace is an inline code editor much like the code viewer but with editing capabilities afaik.

a20behta:
![image](https://user-images.githubusercontent.com/102468604/161975827-e0a39231-c926-4095-864b-fc2a788ea865.png)
new error found when testing after merge. 

this was created by loggin in as stei followed by uploading a dummy file 

- checking needed, does this occure with actual files. 
- would changes suggested above solve the issue. 

@a19simru branch created and ready for development

a19simru:
@a20behta very good. Please send me the branch I can't currently se it under development.

a20behta:
Branch name;
G3-2022-W2-#11359

a19simru:
@a20behta 
**question**: checking needed, does this occur with actual files.
**answer no**: actual files and links seam to be completely unfaced by the problem. Have tested various types (png,txt,js) and kinds (global, course local, version local). This only breaks dummy files.

**question**: would changes suggested above solve the issue.
**answer probably not**: Its really hard to tell because there is a bug (probably) when using localhost that caused files to not be displayed or uploaded. Meaning that I couldn't test if there was a way to solve the problem properly without being able to edit the server. However I have done some research and managed a accidental improvement. 

suggested solution involving fileed.js: This might work (couldn't check). But it does fix the problem with localhost not displaying the list of files. The list is however empty and you still can't upload new files to it (might be because localhost is not connected to the server?). It is a step in the right direction so I will make a pull request with it.

suggested solution involving fileed.php and ace.js: This does not work when I try it and it really shouldn't. Its just a echo printing out what time ace.js was last edited. There is no reason why this should effect listing files that I could think off.  

[...]

PR #11430, opened 2022-04-07
G3 2022 w2 #11359

a19simru:
Small improvement to the problem with using lenaSYS from localhost. You can now se the list (could not before) but list is empty and there is still the problem that nothing happens when you try to add files.


a20behta:
another solution suggested by CU, will use that.
use a list of red days which will be compared to the days. 
List should contain about 5 years in the past and future from current year

The result is that data needs to be updated continuously

Issue #11396, opened 2022-04-06
Diagram:  Colour issue when marking entity.


b19karhj:
When you mark an entity that has colour in it, the colour get darker, this should not happen.
Replace the red colour with purple in both BG and Stroke
![bild](https://user-images.githubusercontent.com/81624541/161992612-67f027f3-0944-4c24-85cd-bbf3c583894f.png)

b20matad:
Upon testing, objects need to be highlighted in more than just one way (The dashed selection box), in case muliple objects are selected. We will test making a dashed border for all objects when they are selected.

b20matad:
On further evaluation, a dashed-stroke is not a good way to highlight selected entities, since some entities have the option to already have a dashed border. 

We decided to change the opacity of the selcted elements instead, since it doesn¨t get mixed up with different colors or borders.

a20jacmi:
Also, me and Therese had a discussion when we tried purple as background in the elements. This made it hard to see the text inside because of such a dark color, thus the choice to go with a light blue.

b20thelo:
Removed waiting for testing as testing of pull request has been completed and the submitted solution was merged.

b20matad:
Since opacity isn't approved we have made some different suggestions for possible solutions. See the following pictures.

Brightness changed on each selected element.
<img width="347" alt="brightness" src="https://user-images.githubusercontent.com/102598238/163942519-67b78850-9015-405c-b75f-b6af22db61e9.png">

Dashed frame around each selected element.
<img width="387" alt="Rutor" src="https://user-images.githubusercontent.com/102598238/163942675-f427448e-733a-43b8-97e2-d60b239cbc3d.png">

Dashed frame with black dots around each selected element.
<img width="401" alt="RutorPlus" src="https://user-images.githubusercontent.com/102598238/163942898-1762811e-ddd7-490b-aad3-6befa5bf3f25.png">

All of the above on each selected element.
<img width="335" alt="Allt" src="https://user-images.githubusercontent.com/102598238/163943017-39d407bf-75ee-4836-91f7-12ba1d2dc7fb.png">

We have not made any changes on the marking of selected lines, since we don't consider it an issue at the moment but could be raised as a different one in the future.

b20matad:
The solution we agreed upon.

<img width="324" alt="Förslag" src="https://user-images.githubusercontent.com/102598238/163954775-46f5e2fe-6877-4ff8-9f41-6ccc94ddb856.png">

[...]

PR #11431, opened 2022-04-07
Fixed color issue #11396

a20jacmi:
We changed the possibility to chose a red color for an entity, by picking another color.

We also modified so that an entity does not change color when it's clicked on. Since the entity gets a clear box around it when it is clicked on, we see no point in changing the color as well. However, if this feature is ever asked for in the future, it is easy to just uncomment the parts mentioned in the code. This is commented in the file so that this process should be a no brainer.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -880,9 +880,9 @@ const textheight = 18;
 const strokewidth = 2.0;
 const baseline = 10;
 const avgcharwidth = 6; // <-- This variable is never used anywhere in this file. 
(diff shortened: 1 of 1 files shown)


c20emili:
Tested by clicking elements and checking the available colors and there is no darkening of the entity. the colors have been replaced with purple in stroke but lightblue in background. there's no semantic error.

b19karhj:
Color problem need to be fixed before merge.

b20thelo:
Pull request closed as more things were added to the original issue.

Issue #11357, opened 2022-04-05
Duplicate plus sign in course editor for teachers


a20willa:
The plus sign at the top may not be there as it's function is identical to the bottom one. 
![image](https://user-images.githubusercontent.com/81626616/161745327-21b1ac2e-5457-4739-b9b1-e46c6f481ab8.png)

HGustavs:
It was done this way since some people in the previous course years liked it this way. Unclear if it was students or teachers.... that is the reason for this duplicate functionality. No problem from my end, as long as both buttons run exactly the same code.

PR #11432, opened 2022-04-07
G2 2022 w2 #11357

a20mtafa:
Fixed the problem by removing some code blocks in a couple of files


a20linth:
Code refactoring - redo
Code standard - ok
Meet issue goal - yes

PR #11437, opened 2022-04-07
move Edit button code

a19xinhu:
make Edit Button can not see if you do not login
Go in a dugga and you will see 3 button which are maximize button, minimize button, reset button.
And log in with teacher account you can see 4th button: Edit button.


a20willa:
Pull request against master by mistake.

PR #11439, opened 2022-04-07
move Edit button code

a19xinhu:
make Edit Button can not see if you do not login
Go in a Dugga and you can see three Button: maximize button, minimize button, reset button.
Log in with teacher account and you can see the 4th button: Edit button.


a20linth:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #11444, opened 2022-04-07
Updated fab in style.css

a18gusjo:
Changed transition in fab-btn-list when window is <= 600px in height. The text now doesn't overlap and is obliquely downwards instead of upwards. Also moved the fab-button so the text will always be visible.

Test this by logging in as mestr, navigate to "show files" and hover over the fab button (the yellow large round one). Then make the window smaller and hover over it again.


a20linth:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #11155, opened 2021-05-24
Select & Move labels in diagram


a19hammi:
Example from draw.io on how it looks. No need for advanced calculations. Just have  a relative x and y to center of line.

https://user-images.githubusercontent.com/81571224/119369589-a139ff80-bcb4-11eb-892f-39de36562a0a.mp4

b20thelo:
The solution needs to be worked on further before being approved for merging so therefore the pull request 11363 was denied.

c20emili:
https://user-images.githubusercontent.com/81320323/162904413-242f4bb9-ed4a-4b5e-95c4-03c7960873e4.mp4

b19karhj:
The label should be able to be moved above and below the line on Y-Axis. (Atm the label can only be moved on X-Axis shown in the video).

c20emili:
removed unnessecary zoomfact that caused issues label will stay on the line even when zoomed.
https://user-images.githubusercontent.com/81320323/163206219-cfbdeec3-019a-4e9a-8937-12ff8bd6f6db.mp4

b20thelo:
Progress made. Left to finish in this issue is the possibility to deattach the label from the line and moving it within a specified distance from the line. Shown in the first video in this issue.

b20thelo:
Closing this as new issue will be opened for the remaining steps.

PR #11454, opened 2022-04-08
fix for issue #11155 Select & Move labels in diagram

c20emili:
added code and the function didClickLabel() in determineLineSelect() to be able to detect a label, changed code in mdown to determine if it was a label or a line that was clicked on, added the pointerstate CLICKED_LABEL and used it in mmoving with the function updateLabelPos() that changes how much the label should be displaced from the center of the line, added the arraylist lineLabelList and in drawLine() added storage of information about the label in lineLabelList and use it to draw the label in the correct position.


b19karhj:
Merge complete

Issue #11399, opened 2022-04-06
Multiline strings do not work correctly in the codeviewer.


a20linth:
When there is a string that has line breaks inside of it there occurs a problem with the tokenizer. The tokenizer does not include the whole string in the token string, it instead ends the string before the line break and handles the rest of the string as if they were outside of the string

 An alert is given that an unterminated string exists for both quotation marks.
![image](https://user-images.githubusercontent.com/102609651/161999310-8292ca8e-5140-43ff-9ddc-a376047c287d.png)
![image](https://user-images.githubusercontent.com/102609651/162000056-b448d6ce-5694-4bf3-91b4-72a8e17ba909.png)
![image](https://user-images.githubusercontent.com/102609651/161999540-998a55b5-f4d2-4b65-862b-90b1bd0954c2.png)
![image](https://user-images.githubusercontent.com/102609651/162000343-1a3dc3a1-fbfc-4431-a14f-922c96c8d4cc.png)
Tokens 211 to 224 should be in the same string token

a20linth:
![image](https://user-images.githubusercontent.com/102609651/162147797-78accff1-228f-4c57-9382-47d28d3d949c.png)
Example of a string token without any line breaks, works as intended.

a20linth:
Testing:
1) Login as a teacher account (username: stei password: password)
2) Enter Demo-Course
3) Create a new code dugga
4) Select a template, Kind: Code Wordlist: PHP
5) Click edit file and uppload this as a .php file (add course local file):
<!DOCTYPE html>
<html lang="en" dir="ltr">
  <head>
    <meta charset="utf-8">
    <title></title>
  </head>
  <body>
    <div>
      <p>Sample text 1</p>
    </div>
    <?php
    echo "<div>";
    echo "<p>Sample text 2</p>";
    echo "</div>";
    ?>
    <script>
    var htmlString = "<div>
    <div>
    <p>Sample text 3</p>
    </div>
    </div>";
    var htmlObject = $(htmlString);
    var multiLineString = "Line 1
    Line 2
    Line 3";
    </script>
  </body>
</html>
6) Click the edit cogwheel and select the uploaded file

Expected output:
![image](https://user-images.githubusercontent.com/102609651/162397061-10d972d5-a8e9-424b-9057-eab9beeb393c.png)
![image](https://user-images.githubusercontent.com/102609651/162397650-7b7b3a42-45d1-44a9-bf40-f4cd06d3a8fb.png)

PR #11456, opened 2022-04-08
Implemented multiline string support #11399

a20linth:
Added multiline string support and removed error message when using multiline strings.


a20mtafa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #11451, opened 2022-04-08
Box selection selects all elements touching the box, which create inconvenience.


a20jonda:
![image](https://user-images.githubusercontent.com/102598258/162381415-d941dac5-6f9d-40c1-b095-42d6a7bb0f4f.png)

a20jonda:
Alla element som är helt täckta borde bara väljas istället

a19hammi:
_Imagine this is a whispering bird_
So, the reason why it selects the lines is so that when you press the `delete` button the lines are ALSO deleted from the diagram. Otherwise, you will only delete the elements which will cause some weird shinanigans.
_Now imagine the bird flew away_

PR #11458, opened 2022-04-08
fix for issue #11451 Box selection selects all elements touching the box, which create inconvenience.

c20emili:
changed getRectFromElement() to check if the entire rectangle is within the element and lineIsInsideRect()(formerly pointIsInsideRect) to check for the entire line instead of a point. also changed the comments to clarify that its the entire rectangle/line rather than a single point.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -2512,18 +2512,18 @@ function getRectFromElement (element)
 }
 
 /**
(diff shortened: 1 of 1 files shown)


b19karhj:
@c20emili Fix so objects that are halfway selected do not get selected. Thye should only get selected when they are around 80% covered.

Issue #11440, opened 2022-04-07
Diagram: Better tooltip description for the diagram tools.


b19karhj:
Some of the descriptions of the tooltips are a bit vague and not true.
For example: The zoom in icon description is "increases size of all elements" which is not true and a bit confusing.

![bild](https://user-images.githubusercontent.com/81624541/162202086-f8bf1c33-6bd6-46d4-9187-f52982bca811.png)

Make a better tool description for:
Zoom in
Zoom out
Zoom reset
Toggle snap to grid
Toggle step forward
Toggle step backwards
Toggle replay

a20edwth:
Changed tooltip descriptions of the toolbar and added tooltip descriptions to the buttons in replay-mode.

Instructions for testing:
1) Click "Demo-Course"
2) Click "Diagram Dugga"
3) Hover over buttons in toolbar and buttons in replay-mode and read descriptions.

b20thelo:
Closing as last update from Edwin was approved and pushed.

PR #11462, opened 2022-04-08
Changed all the descriptions for the menu elements for more clarification and unified style. #11440

a20jonda:
Added parenthesis to the side to better show that the text is a further clarification from the name and changed all the texts with better wording and closer to truth explanations.


c20emili:
Looked at all descriptions and pointer, box selection ,all toggle descriptions and toggle replay are clearer. Step backward description would be better without "in history" and the rest are not clear or worse. a lot of unnecessary parentheses. The code semantics and syntax is good.

b19karhj:
Needs to be reworked.

a20jonda:
#11440

b19karhj:
Still need changes, we will go through them togheter before we do a new pull request

Issue #11442, opened 2022-04-07
Diagram:Create a working exit button for the “replay” mode.


b19karhj:
At the moment there is only one way to exit the replay mode (where you can check each step of the user's changes to the diagram), and that is by pressing the escape key.

![bild](https://user-images.githubusercontent.com/81624541/162205989-d9a5b162-ed5b-48e2-950e-5db3ff54546a.png)

Make an exit button which takes you back to the same step as escape does.

b19karhj:
Redesign the replay button as its very hard to see what it does just by looking at it.

a20samul:
Created a working exit button with intended functionality present and modified the replay-button icon.

<img width="257" alt="exitButton" src="https://user-images.githubusercontent.com/81613484/162401848-c06ef409-1161-4319-90b0-c211b6c656ff.png">

PR #11465, opened 2022-04-08
Created a working exit button for replay mode in diagram dugga #11442

a20samul:
Created a working exit button with its intended functionality for replay mode, redesigned the replay button and added title(descriptions that show when you hover the icons) to all three buttons on the panel.


b19karhj:
Made changes and went into replay and pressed the button, it took me back to the same page that escape does, and keeps the previous changes.

Issue #11467, opened 2022-04-08
Clean Navheader.php from any mentions of analytics.php.


c20adael:
Refering to the issue #11385

b20hanwa:
#11471

a20behta:
merged into weekly branch and included into main branch

PR #11471, opened 2022-04-08
G3 2022 w2 #11467

b20hanwa:
Removed code mentioning analytics.php from navheader.php


c20adael (approved):
yepp this will remove the creation of the html elements that is shown in the nav bar when you are on the analytics page. Since analytics has been removed this serves no propose.

PR #11472, opened 2022-04-08
Added a new dugga template for seminars

a20johba:
Added template files seminar_dugga.html and seminar_dugga.js. demoCourseData.sql was edited to include a Seminar Dugga in the Demo Course, note that this requires reinstallation to take effect. The Seminar Dugga page can be found by following the link on the bottom of the Demo Course page.


a20behta (approved):
Looks good, 
additions looks good 
database is updated as expected,
information is correct insertions are included correcty and expected, 
apart of sql files which run during install process.

PR #11475, opened 2022-04-08
Started adding diagram assignment function from teacher's "edit dugga" window. Status - Incomplete

a20micro:
non-functional changes regarding functionality of assigning .JSON file to a given dugga. Issue is deemed to be too big as a single issue so start of functionality has been aborted with half-complete code which will be used as a base for further smaller issues to be created.

PS: big block of code removal and addition in diagram.js was only moving it into an IF-statement which will always be true. Diagram functionality should currently function as normal.


b20thelo:
Branch should not be deleted as it is still worked on.

PR #11481, opened 2022-04-08
Select/deselect evolved. issue#11455

b20matad:
Enabled the posibility to select/deselect multiple elements when pressing "ctrl"+"mouse1". The issue concerned "box selection mode" but now works for all mouse modes.


a20samul:
Tested solution by selecting respectively unselecting multiple objects. Functionality operates as intended as objects are de-/selected while holding in ctrl + clicking on the object. Test passed.

b19karhj:
Merge complete.

PR #11483, opened 2022-04-08
Changed dugga.js to reenabled receipt popup

b20eripe:
Adjusted file dugga.js which had an fault on line 1099 resulting in the receipt popup didn't show


b20eripe:
Found an fault in data sent to ajax, restored to an older version of the line. See line 1097 - 1099


Skärmbild från 2022-04-08 15-17-49

a20behta (approved):
rolls back and returns the line which wass changed

Issue #11468, opened 2022-04-08
Dugga name missing


a97marbr:
![Screenshot 2022-04-08 at 11 21 48](https://user-images.githubusercontent.com/4948661/162406957-75cb03e8-bf52-4e92-be8d-d7b6d98a9f54.png)

When opening bit-dugga the dugga is named "undefined".

Does this problem occur in more duggas? I think an investigation could be in order.

c20adael:
Research where the dugga is suppose to get this data from and why it doesn't get it now.

a20vicad:
I can have a look at it @c20adael

a20vicad:
![image](https://user-images.githubusercontent.com/102599327/162703509-98cf17d1-d84b-4105-bf74-1946a4f91aa8.png)
The name of the dugga can now be seen in the instructions tab.
![image](https://user-images.githubusercontent.com/102599327/162703701-a2c5a9a5-8194-4cae-a3ab-6c39ec0a8721.png)
Also in the top-bar-thingy.

a20behta:
weekly merged done and included into master. tests done,

PR #11498, opened 2022-04-11
G3 2022 w2 #11468

a20vicad:
Now the title of each dugga is visible. In the case of the lower title, it was just a case of upper/lower-case characters being mixed up.


a20vicad:
To test this. Look in the different duggas available and make sure they have their right titles.

a20behta (commented):
looks good, 
does not work for every dugga however most work.

will check closer why it does not work for some

PR #11499, opened 2022-04-11
G1 2022 w2 #11396

a20jacmi:
Here is a solution. See description in the issue. I talked with Therese about the purple colour and we decided that purple was to dark for the background of the elements, since the text becomes too hard to see.


a19meram:
Tested using my laptop and home-computer, changes on the opacity of the background-color have been made and works well but the stroke-line of the border have also changed which makes it looks very border-less on zooming. 
image

b19karhj:
Merge complete.

Issue #11379, opened 2022-04-06
Adding the graphical components to the resulted.php file.


c20adael:
This is reffering to the issue #11107

c20adael:
How is this coming along? @a20kimar

a20kimar:
> How is this coming along? @a20kimar

It's going well. I just saw the message in the issue about functionality regarding resulted not being used much. I've just finished the layout and so on. The current version of lenasys has a semi-hardcoded way of handling the different checkboxes but I think I got around that fairly nicely. I'll commit and push so someone can have a look at whether this is worth continuing on or not.

a20kimar:
My solution should now be available for anyone in the issue branch. The previous method was understandably coded for one solution of one window of checkboxes and I had to make it viable for two or more. I added php elements to represent the new button and checkbox window. I then had to implement so that it functions independently from the other one, it's own css id's, classes and names and then also change or add to the javascript where needed to have this function on its own. For example would the "select all" button work on both of them at the same time while it should only work on the elements in that window. I also made sure that the elements are dynamically changed depending on the table header which was the case for the dugga filter as well.

For a more in-depth view check out the commit containing all the changes on the branch for this issue.

This is how the finished product look.

<img width="383" alt="Screen Shot 2022-04-08 at 14 34 19" src="https://user-images.githubusercontent.com/82010032/162436536-f88b18e6-0e22-48ab-a4ca-a9255021770d.png">

a20behta:
merged into master during weekly merge

PR #11502, opened 2022-04-11
G3 2022 w2 #11379

a20kimar:
Added a button to filter columns in resulted.php. This pull request does not offer any functionality when the search button is pressed. This adds a functional button beside the current existing one. The button makes a menu appear where you can check different columns in the table below on the same page and should offer functionality to filter which column the teacher would like to filter by in the future.


a20behta (approved):
changes look good 
added functionality in frontend none added in backend as discussed. 

looks good

Issue #11485, opened 2022-04-08
Feature/alert to avoid that diagram dugga is closed by mistake.


b20thelo:
When working on the diagram you don't want to lose your progress by accidentally closing down the web browser, tab, or just clicking the back arrow.  

Suggested solutions:
- An alert box that asks if you are sure if you want to close this tab and you have to either click yes or cancel.
- Some kind of feature that keeps track of when changes have been made and asks if you want to close down before saving forcing you to click "Close down without saving" or "Go back to save" or just call the save feature. 

The first step would be to add this functionality on the page where you modify the diagram.
The second step would be to add the same functionality when in the replay mode (we need to wait to add it here since we have other changes made for replay mode that needs to be merged first to avoid collisions)

b20thelo:
@HGustavs What are your thoughts on this? Do you have any suggestions for this.

a20samul:
An alert is shown before unloading the page through the actions circled in red below (see attached image).

<img width="784" alt="alert" src="https://user-images.githubusercontent.com/81613484/162900409-6f5790d7-0bdc-429a-914a-a5db328a6e89.png">

PR #11503, opened 2022-04-11
Created functionality for an alert that pops up before closing diagram dugga #11485

a20samul:
Created an alert dialog message that pops up before diagram dugga window is closed or refreshed.


a19meram:
Tested on live-server an alert pops up before closing or refreshing the diagram dugga window and works well!

b19karhj:
Merge complete.

b20thelo:
Need to test again to see that alert doesn't show up when no changes have been made.

Issue #11476, opened 2022-04-08
Cannot click the displayEditExample() button to edit the example settings while using template 10


a20linth:
Button does not work for template 10

https://user-images.githubusercontent.com/102609651/162434756-86a12de4-c691-4215-8def-bac3639da863.mp4

Button does work for template 1

https://user-images.githubusercontent.com/102609651/162434775-2e1db9df-b3e2-440d-873c-901449e2d214.mp4

Button works for templates 1-9

a20linth:
Testing:
1) Login as a teacher account (username: stei password: password)
2) Enter Demo-Course
3) Create a new code dugga
4) Select template 10 (last template), leave other options as is
5) Click on "Example Settings" (cogwheel icon in the header)
Expected results: An Edit Example window should appear.

PR #11505, opened 2022-04-11
Removed faulty code #11476

a20linth:
Removed boxid in displayEditExample() because it does not change an individual window-box, but rather the whole example. Removed an fault in displayEditExample() that causes an error to occur when using template 10


a20mtafa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #11506, opened 2022-04-11
Box selection problem.


a20jacmi:
https://user-images.githubusercontent.com/102599629/162704509-52917fbc-8db7-49e8-8700-c6a831420ff3.mov

If a group of elements are selected and you wish to use the selection box again, the functionality is broken.

Our solution is that the selection box only should select new elements, whilst deselection is done individually.

a20jacmi:
We were able to fix what we wanted. The solution is found in "diagram.js" under "boxSelect_update".

a20jacmi:
https://user-images.githubusercontent.com/102599629/162902291-a001dd95-0632-4646-a6c3-747397969cb8.mov

As you can see in this video the entities are not deselected as in the previous video, when "ctrl" is held down to select additional elements. Now, only newly selected items will be selected if "ctrl" is held down when making additional selection boxes.

PR #11507, opened 2022-04-11
Fixed issue #11506

b20matad:
Changes made in function "boxSelect_update()" in diagram.js. If marked entities were already marked, They stay in the array "previousContext" instead of being added to the array "markedEntities", which fixes the odd behaviour of the issue.


a20samul:
Tested by box-selecting multiple objects. The issue in 11506 alongside any bugs were not percieved. Test passed.

b19karhj:
Merge complete.

PR #11510, opened 2022-04-11
Added the examplename to the drop down list.

a20patpe:
The examplename used to be undefined. it should now always be the examplename.

For testing: 
1:when holding down the backward/forward the examplenames should be shown like the picture below.
image

What it looked like before for reference:
image


a20mtafa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #11514, opened 2022-04-11
Added function to remove dropdowns when user clicks outside of them.

a20joabo:
Implemented a function to remove the drop down once the user clicks outside of them.


a20mtafa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #11515, opened 2022-04-11
Dugga deadlines are now displayed the same for all screen widths.

b20hanwa:
Fixed issue#11494 by matching opacity and max-width for both .dateField in style.css.


c20adael (approved):
i have tested this and it seems to work on both mobile and pc

PR #11519, opened 2022-04-11
Template change assignment description issue#11364

a19meram:
Added a dugga description on left side of diagram


i19matla:
This comment is because we are new to this and im not tagged in the last comment to signal activity.

b19karhj:
Test failed on live server. The text on the left side did not show up

Issue #11122, opened 2021-05-21
National Holidays contribution - the view at the bottom


a19anjst:
Make a list of national holidays and if that already exist make use of it. In the week section add red days date in red. Make the weeks "smaller" depending on red days 
![Skärmbild (427)](https://user-images.githubusercontent.com/62884743/119124141-9ff8a080-ba30-11eb-84b8-901862c38755.png)

a19simru:
status update: wrote algorithms to calculate the dates of all the different holidays every year. Researched and downloaded mocha to run tests on my holiday finding. Gonna need another day of work to finnish.

a20behta:
pull request related #11522

a20behta:
@a19simru why are the following files deleted?

DuggaSys/analytics.js
DuggaSys/analytics.php
DuggaSys/analyticsService.php
Shared/css/analytics.css
Shared/navherader.php

a20behta:
@a19simru new branch created and ready for implementation we discussed

PR #11521, opened 2022-04-11
G3 2022 w2 #11122

a19simru:
@a20behta @c20adael New features added. A upgrade of the old static red day list (DuggaSYS/contibution.js) to a new dynamic red day list that updates every year. Second feature is a new column in the commit graph listing red days every week and what dates they are on.

Problems new bug created. 
Screenshot (4)
Weekly line chart and weekly bar chart stopped rendering after changes to the function getHolidays (DuggaSYS/contibution.js) old code has been left commented out to save a bug free version for the future. 

Extra problem 
I could not set the header of text of the new column of red days. so now it just says undefined. Unfortunately I once again think this requires changes to the database file.  Since I haven't been able to find any spot in the code that actually writes headers. Therefor it stands to reason that headers are pulled directly from the database.


a20behta:
this is directed straight to master, another pull request with the same branch already exists

Issue #11224, opened 2022-04-01
Section editor does not allow us to edit things in the past


HGustavs:
<img width="588" alt="Screenshot 2022-04-01 at 09 13 54" src="https://user-images.githubusercontent.com/6978801/161214007-d3085cfb-9215-4e9f-b2af-1f15be6bd5d0.png">

<img width="495" alt="Screenshot 2022-04-01 at 09 16 18" src="https://user-images.githubusercontent.com/6978801/161214333-d8c78940-fda0-4f1a-85d6-bcf7ff71b0df.png">

It seems like the editing of for example the name of an assignment is not possible until we have changed the deadline so that it matches the current course dates.

This is too restrictive, we want to be able to edit the name of the assignment first and then edit the dates. Commonly dates are the last thing that is updated.

a20linth:
The checking of a valid deadline interval is currently broken, the valid date range is from 2020-05-01 to 2020-06-29 (Date of testing is 2022-04-13). The deadline date has to be within the valid date range for the item to be updated with new information. Because a working deadline system is currently not implemented, then there should be no problem updating the item with an invalid deadline to allow the other options to be updated.

a20linth:
Another issue appeared after the weekly-merge which makes it impossible to save changes on any item in sectioned even if nothing is changed.
![image](https://user-images.githubusercontent.com/102609651/163188864-875bb83e-ebad-4355-aafe-2b1599bbde0b.png)

a20linth:
Testing:
1) Login as a teacher account (username: stei password: password)
2) Enter Demo-Course
3) Edit one of the items listed in "Tillgängliga Duggor"

Expected result:
Using invalid characters in name should not be allowed, and using an invalid deadline should be allowed

Not allowed:
![image](https://user-images.githubusercontent.com/102609651/163347161-38301349-8c7c-4b53-83dd-80f628d7a7fa.png)

Allowed:
![image](https://user-images.githubusercontent.com/102609651/163347247-a457abef-0cf3-486b-b67a-f5e5fb5292ef.png)

a20sarle:
Date picker worked as described during testing but does not allow the same functionality on the server. The functinality of the dates needs to be checked again.
![image](https://user-images.githubusercontent.com/81631818/164454282-86660d1e-519d-4e5a-8944-bca3e8096371.png)
![image](https://user-images.githubusercontent.com/81631818/164454316-b9ee7135-06f7-4d99-9d3a-f12968257552.png)

[...]

PR #11524, opened 2022-04-12
Validation corrected

i19matla:
Validation is now uniformly, issue #8528.


a20linth:
This solution does not work. I will fix this in issue #11224

Issue #11497, opened 2022-04-11
Show files page doesn't show files when working on local host


b20thelo:
Make files visible when working on localhost. 
Link: http://localhost/LenaSYS/DuggaSys/fileed.php?courseid=1894&coursename=Demo-Course&coursevers=52432

a16linas:
Found potential solution, adding a toString() to the row[key] before toUpperCase() seems to solve the issue in fileed.js line 581

b20thelo:
Closed as it was solved during last week merge. Pullrequest #11580

PR #11530, opened 2022-04-12
G1-2022-W2-#11497

a20chrka:
Should work now. Problem was at line 581, where upperCase() tried "uppercasing" a non-string element. By adding toString() before solved the issue.


b20thelo:
Tested on local host by switiching to branch and it now shows files as it should.

PR #11533, opened 2022-04-12
Added a description on diagram_IOHandler.php #11252

a20jonda:
#11252 Put a simple description of what the ui in diagram_IOHandler.php does and added some notes for future developers. diagram_IOHandler.php is far from finished so the actual planned functions is not known.


b19karhj:
Merge into repair branch.

PR #11537, opened 2022-04-12
Removed light highlight on text in dark mode.

b20fanfa:
Fixed the text where there are bright highlights in dark mode, it should not exist.


a20linth:
Cannot be tested on web server

a20willa:
Cannot be replicated on server.

Issue #11536, opened 2022-04-12
Edit button does not show in template 10


a20linth:
The edit button does not show in template 10 when being logged in as a teacher with writing access. The edit button shows in other templates

a20patpe:
The functions of adding things to the different boxes is hardcoded and template 10 is not included in this hardcoding. 
Search for: templateid == 9 in codeviewer.js. 
Below that there should be a templateid == 10. :)

a20linth:
Testing:
1) Login as a teacher account (username: stei password: password)
2) Enter Demo-Course
3) Create a new code dugga
4) Select template 10
Expected result: The edit button should be showing if you are logged in as a teacher

PR #11541, opened 2022-04-12
Gave a unique class to edit button #11536

a20linth:
Previously the edit button used the reset button class and was therefore hidden in template 10 along with the resize button where no resize capabilities should exist.


a20mtafa:
Code refactoring - OK
Code standard - OK
Meet issue goal - NO

a20linth:
The branch G2-2022-W2-#11536 is not updated to the latest commit on the branch for the web server and therefore does not work as it should when tested

PR #11544, opened 2022-04-12
Updated zoom in diagram.js (issue #11342

a20chrka:
Zooming with the zoom-buttons now zoom towards/from the true center of the screen. There are still problems problems with the scaling of the scroll variables, same as in issue #11058.


b20thelo:
A step in the right direction but still need further work so it isn't ready to merge yet. When further progress has been made a new merge request has been made.

Issue #11493, opened 2022-04-11
Downloaded markdown files does not have plus signs (+) but rather the HTML code &#43;


a97marbr:
Downloaded markdown files does not have plus signs (+) but rather the HTML code &#43; 

Potentially related to #11491

a20vicad:
![image](https://user-images.githubusercontent.com/102599327/162755387-bd0d6018-4731-4cc0-abf6-8c295ab6a6bc.png)
There is code (now commented out) that changes + to %2B (Hex-code for +)

a97marbr:
Is there any history regarding this line of code?
```
content = content.replace(/\+/g,'%2B');
```
It seems like this code has been added to solve something ... why else add this line of code only for plus signs? Can we find when the code was added?

**On a side note.** Very good to add code snippets when discussing an issue, but refrain from using images for code snippets. It is very frustrating not to be able to copy & past from the code snippets above.

a20vicad:
![image](https://user-images.githubusercontent.com/102599327/162896973-9acea120-8b69-42ac-a575-a1239facb580.png)
Yes there seems to be history.

a97marbr:
I wonder why? Will the plus sign go missing completely if the culprit line-of-code is removed? The old commit msg seem to indicate this.

a97marbr:
Yes the replacement seems straight forward. 

What I'm curious about is the result for the downloaded markdown file. Will the file contain the + sign? The original bug seem to indicate that the + sign is lost if the string replacement is not present.

So when you are testing the solution ... are you checking that the downloaded file contains the + sign?

a20vicad:
The file contains a "+"-sign but it is downloaded as a txt file.

a97marbr:
Ok then I would say that removing the string replacement is the solution.

The fact that the markdown file is downloaded as `.txt` I believe is a different issue.

a20behta:
merged in weekly merge.

PR #11547, opened 2022-04-12
Removed the lines "content = content.replace(/\+/g, '%2B'); for issue #11493

a20vicad:
In fileed.js

The only testing to be done might be to try editing markdownfiles with "+" signs and se that it works not just for me.


a20behta (approved):
Does not have major affekts on system, allow for + sign to be added just as well in both systems.

Bite DyggaSYS/fileedservice.php, unsure what the change does and how it affects the rest of the system

PR #11548, opened 2022-04-12
Implemented column filter functionality and improved the "Select all" button

a20johba:
Clicking a checkbox in the column filter previously had no effect on the resulted table. This functionality was added to hide columns that are unchecked and reveal them again when checked. The functionality of the "Select All" button was also changed slightly.

Added a function that checks what columns are filtered before updating the table. Added and rewrote functions for onclick event for the checkboxes in the dugga and column filter menu to make the "Select All" checkbox more reliable.

To test:
1. Navigate to http://127.0.0.1/LenaSYS/DuggaSys/resulted.php?courseid=2&coursename=Webbutveckling%20-%20datorgrafik&coursevers=97732
2. Open the column filter.
3. Click any checkbox and see the result in the table.
4. Click the "Select All" button with different checkboxes checked/unchecked. The button should select all if any are unchecked, or uncheck all if all are checked. Unchecking any other button after selecting all will also uncheck the "Select All" button.

### DIFF
--- a/DuggaSys/resulted.js
+++ b/DuggaSys/resulted.js
@@ -14,6 +14,7 @@ var toggleDuggaCheckAll;
 var checkboxElements;
 var searchBarElement;
 var searchDelayTimeout;
+var colOrder = ["duggaName","hash", "password","teacherVisited", "submitted", "timesSubmitted", "timesAccessed"];
 
 function searchByFilter() {
 	// Date object requires string to not apply random time zone.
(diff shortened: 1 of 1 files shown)


a20behta (approved):
Looks good. 
functions changed seem to keep functionality while streamlining and making them easier to read. 

changes done needed to be done.

a20behta:
asumed to be a lack of functionality.

Issue #11482, opened 2022-04-08
Check if a fault causes a bug.


a20patpe:
Found this line in sectioned.js:

// Default showing of gradesystem. Will show if has type "Test" or "Moment"
  if (kind != 3 || kind != 4) {
    document.querySelector("#inputwrapper-gradesystem").style.display = "none";
  } else {
    document.querySelector("#inputwrapper-gradesystem").style.display = "block";
  }

 if (kind != 3 || kind != 4) will always return true, the code was added 3 years ago and i'm unsure if the current implementation works around the fault or not.

a19simru:
@c20adael @a20behta If there is a point to this code segment then I sure cant find it. the line.
document.querySelector("#inputwrapper-gradesystem").style.display 
writes to a element in document. But the code never appears to read from document that element in document.  I haven't found any usage of the element #inputwrapper-gradesystem in the entire LenaSYS application. It's only ever referenced in the sectioned.js file. Perhaps i's read somewhere in some other way I can't think off but I cant find anything. I have tried commenting out all references to #inputwrapper-gradesystem. But I can't notice any changes at all when I run localhost. The code seams to run just the same without these unread document editions. 

conclusion: I believe #inputwrapper-gradesystem and all usages of it in sectioned.js are stubs whose usage was never actually implemented.

a20behta:
merge made in seperate file, removed if statement and else block, remaining kept the same

PR #11550, opened 2022-04-12
G3 2022 w2 #11482

a19simru:
Removed code from sectioned.js that seamed to do nothing.


a20behta:
new pull request will arrive soon with requested changes

PR #11551, opened 2022-04-12
Updated showdata() in diagram.js

a20chrka:
Ghost elements are no longer drawn.


b20thelo:
Other issues emerge so we decline the pull request and will look into it further after the weekly merge.

Issue #11477, opened 2022-04-08
Creation of file select drop-down menu connected to database


a20micro:
As part of #11401 
After the creation of a dugga of the template type "Diagram_dugga" found by clicking the plus sign under "Show tests", the teacher should be able to add an ER model. 

The ER model is exported as JSON-format and then uploaded through the "Show file" feature as JSON-format. The JSON file will be loaded when the students open up the diagram dugga. 

Under "Edit variants" a drop-down menu should be available where the teacher can select between possible .json files that exists in the database, shown on "fileed.php".

![ExampleDropDown](https://user-images.githubusercontent.com/95532115/162435699-b4eae090-ca59-452c-9761-844d2ef7ebf8.PNG)

Example of what it could look like.

b20thelo:
Comment from pull request added here as well. 

This is a work in progress for #11477, but so far it lists files from the database but it isn't doing anything with them. Next step is to filter out everything that isn't a .json file and read it's content as a new diagram dugga. It's possible to test this by editing any dugga in duggaed.php

![bild](https://user-images.githubusercontent.com/37794783/162965364-5e52eb20-c0d8-47d5-80a6-86d62d853623.png)

b20thelo:
Add an empty start option in the dropdown so that the teacher isn't forced to select a file.

b20thelo:
Closing this as the dropdown is now added and other subissues to issue 11232 includes the steps that follow after this issue.

PR #11559, opened 2022-04-12
WIP Creation of file select drop-down menu connected to database #11477

a16linas:
This is a work in progress for #11477, but so far it lists files from the database but it isn't doing anything with them. Next step is to filter out everything that isn't a .json file and read it's content as a new diagram dugga. It's possible to test this by editing any dugga in duggaed.php

bild


b19karhj:
Tested by three people, Worked but need improvments mentioned by the developer.

Issue #11480, opened 2022-04-08
Diagram module - Update toolbar to accomodate UML entity


a20gabpa:
Related to #11286 
Update the existing toolbar. The goal is to add a temporary icon for the UML entity inside the toolbar and implement a way to show that certain parts can be expanded. Something like this with addition of an small arrow inside the button to indicate more items:

<img width="158" alt="161744730-76dabdc7-399b-42b2-a27c-4d895f083055" src="https://user-images.githubusercontent.com/102604680/162439330-d1633155-f791-413f-95ab-575f265257c9.png">

![image](https://user-images.githubusercontent.com/102604680/162440509-39eda8ac-6393-4780-9b6e-f80628fe1c5a.png)

When the small arrow is clicked, a box placed right of the arrow will appear. 

Suggested steps:
- Add UML icon to toolbar
- Add ability to click on the icon (the creation of the UML entity is worked on in another issue.)
- Add arrow symbol for the ER button
- Add container / box that appears after the arrow is clicked

Issue #11445, opened 2022-04-07
Diagram module - Add UML entities


a20gabpa:
Related to #11286 
Create an UML entity. The entity should have a name and space to add multiple attributes. Something similiar to the included picture. Since an entity can have multiple attributes, the box should also be able to expand, to accomodate added attributes. 

![image](https://user-images.githubusercontent.com/102604680/162260044-9f5b82f9-452f-4f88-a4c1-fd5976b66aa6.png)

a20gabpa:
Screenshot of the UML entity with dynamic size that expands to accomodate all the attribute. In the first screenshot the UML entity has 2 attributes and the second has 4 attributes. At the moment, the attributes value cannot be changed since the functionality to modify the element in the 'Options panel' is the same as the ER element. This was not the focus of this issue and will instead be implemented in an upcoming issue regarding the 'Options panel' for UML.

![image](https://user-images.githubusercontent.com/102604680/162902633-3e898a3a-3cff-4e55-b4ed-d14f4b742494.png)

![image](https://user-images.githubusercontent.com/102604680/162902530-69d5c43a-12ee-4616-8f33-b9d3e2636539.png)
 
There is a small issue with the function 'updateCSSForAllElements'. At the moment, the function will color an element's grandchildren (overwrite current CSS) which results in the funky look for the title. This is a small issue and I will look into that.

a20gabpa:
The earlier mentioned issue where name of the UML entities gets a different look seems to have been fixed by another issue when merged. The problem does not seem exist anymore. 
![image](https://user-images.githubusercontent.com/102604680/163184333-3c839144-f9aa-4488-97e7-1072b0f2aa39.png)

PR #11560, opened 2022-04-12
Fix for issue #11480 Diagram module - Update toolbar to accomodate UML entity

c20emili:
added a press function when pressing the mose button on a plecement type opens a box with extra options and merged with the fix for issue #11445 by @a20gabpa. currently one needs to push the button after choosing to use the new placement type.


b20thelo:
Doesn't work as it should. Need further adjustment

Issue #11377, opened 2022-04-06
Add functionality to the copy link button.


c20adael:
When clicking the button that give you the link for canvas iframe, it should be copied directly to your clipbard.

Todo: get the link to to the dugga into function showCanvasLinkBox in sectioned.js

HGustavs:
Is there a mock-up of this interface?

a17nikjo:
Starting researching how to aquire the correct link to the canvas dugga.

HGustavs:
This is not trivial, since we are generating search terms... perhaps a warning of some kind e.g. color the icon if not possible.

If we have two assignments with the same name for example:

Section 1
Makeup Assignment

Section 2
Makeup Assignment

Then there would be no way to generate a copy paste icon since both those assignments have the same name....

if one assignment is called

Database Implementation

and another one

Database Design

Then one link would link to implementation and the other to design

or 

"Database Design" and "Database Implementation"

The Database part would in that case be superfluous.

a20behta:
Placing issue on hold untill @HGustavs answers questions related to link generation.

a17nikjo:
Creating a branch for this issue from branch for issue #11323 since this requires the code in that issue.

HGustavs:
It should generate an SH linl. 

```
<p><iframe src="https://dugga.iit.his.se/sh?c=forskning&amp;a=avhandling" width="800" height="1200"></iframe></p>
```

a17nikjo:
Link does currently not work since we need are waiting for issue #11512.

Current functionality can be merged while waiting for the correct link to be added.

PR #11564, opened 2022-04-12
G3 2022 w2 #11377

a17nikjo:
Link current link is not correct however all the graphical components works. And the correct function is called when button is pressed.

This can be merged and be on hold until we can create the link.


c20adael (approved):
This has been tested. Is seems to work on the local verson bit not when on the server. It is only a minor issue and can be fixed after merge

Issue #11387, opened 2022-04-06
Delete button in elements on canvas in diagram


a20edwth:
Deleting a element or line in the diagram can only be done through a keyboard press ("delete").  
Adding a delete button in the toolbar supports mouse interaction. (using the diagram without a keyboard)

a20edwth:
Suggested visual representation of the delete button in the options menu of a element:
![11475895909ea6f62c8c136a9dec1836](https://user-images.githubusercontent.com/102611046/162008645-64d7f935-856f-4cd4-a5e7-38435ff379b7.png)

b20thelo:
@HGustavs What are your opinions on the suggestion above for a physical button to delete objects and not just the delete button on the keyboard?

HGustavs:
Not a bad idea. Question is if this is the best solution.

I would like something that would not require opening the tool bar to know that the delete button is present.

How about a cross icon at the top right corner of any selection (one or more elements).


<img width="478" alt="Screenshot 2022-04-06 at 22 49 53" src="https://user-images.githubusercontent.com/6978801/162068529-d92ca0c0-21e9-44f3-b55c-8f062c2cdec8.png">

If one clicks the cross the selected items would be deleted. This way no menu needs to be opened to delete the item. We can also delete the items with minimal mouse movement.

a20edwth:
@b19karhj @b20thelo Video that shows the design and functionality of the delete button:
https://user-images.githubusercontent.com/102611046/162414596-c871de78-09ec-43e2-8244-050a8b305acd.mp4

b19karhj:
@a20edwth Looks like a very good solution.

b19karhj:
Remove the dotted frame around the object but keep the X.

b19karhj:
The X need to be larger. And the placement should be like this 
![bild](https://user-images.githubusercontent.com/81624541/164230255-343ebae3-6e13-4b10-ad87-355def6b674b.png)

b20thelo:
Closed as solution was approved by Henrik

PR #11566, opened 2022-04-12
Removed the square around the X (right branch name)#11387

a20edwth:
Removed the square around the X. (right branch name)


b20thelo:
Close as it shouldn't be included in weekly branch.

PR #11567, opened 2022-04-12
Error in fileed.js 

a20kimar:
This pull request fixes a error in fileed.js:

Ace is not defined


c20adael (commented):
Im aware that this does remove the "ace is undefined" issue but when this is done it also creates a problem that makes fileed.js fail at 581. Needs a check and a discussion before allowing this to be merged

c20adael:
we will close this for now and create an issue about the generated problem

PR #11568, opened 2022-04-12
Added additional code for the alert to only show when changes have been made in diagram dugga #11485

a20samul:
Added code for the alert to only show up if changes have been made in diagram dugga and fixed a misspelling.


b19karhj:
Already added.

PR #11574, opened 2022-04-12
G1 2022 w2 #11155

a20gabpa:
Repair merge W2


b19karhj:
When zoomed in, The label can be moved very far and out of the lines and elements position. need to be fixed.

PR #11589, opened 2022-04-13
Added functionality so that the column TeacherVisited should work.

c20adael:


c20adael (commented):
This should work BUT There is something wrong with sortableTable.js. It looks like a formatting issue.

c20adael:
@a20novra

PR #11591, opened 2022-04-13
G3 2022 w2 #11482

a19simru:
@a20behta New more conservative version that does not cut as mush code.


a20behta (approved):
tautological if statement is removed and if statement block run every time instead.
no other changes. 
nothing else seems affected

PR #11597, opened 2022-04-13
Replaced the way text is writen to clipboard.

a17nikjo:
Text is now written too clipboard using document.execCommand('copy') instead of navigator.clipboard.writeText("").


a20behta (approved):
changed work and tested. 
adds a simpel text box and removes it after copy command executed. 

tested on webserver and works as expected

Issue #11183, opened 2021-05-25
Clicking elements in line creation mode sometimes selects connected lines instead


a19nikbo:
The lines should not be clickable through an element, just like in the other cursor modes

![2](https://user-images.githubusercontent.com/81541557/119516172-ce4ee680-bd76-11eb-8d07-cf74f6dabb89.gif)

b20thelo:
![image](https://user-images.githubusercontent.com/81772705/163133513-f5202040-0882-453f-9b4a-f51da8c666ba.png)
Found possible cause of the issue. Line offset does not scale with the zoom factor. When zoomed out the lines goes into the element.

a20jacmi:
What you say is true @b20thelo, and this makes it more likely that you will hit a line when you're suppose to click on an element. However, we've found the root cause of this issue, and then the zoom factor is irrelevant, but we're waiting with the implementation until the next merge.

a20jacmi:
Solution to the problem is found in: "diagram.js" -> "function ddown(event)" -> "case mousemodes.EDGE_CREATION".

Earlier, if an element (that was not a line) was clicked it was selected on "mouse down". However, on "mouse up" a line could accidentally be selected instead which was not ideal. The "mouse up" behaviour in this particular situation is now disabled.

PR #11603, opened 2022-04-13
Fixed clicking elements in line creation mode #11183

b20matad:
Problem was that "selecting element" trigger on mouse down event, and "selecting line" on mouse up event. The changes made prevents "selecting line", if an element has been clicked in "edge_creation mode".


a19meram:
Works well without any problem.

i19matla:
Works well.

Issue #11487, opened 2022-04-10
Edit button in mobile view?


a20willa:
Does the teachers actually use the edit button in mobile view inside the codeviewer? If not, we could remove the icon, since it would be very clunky to change code in your phone.

The customer may be asked here, incase it is still needed.

b20idawa:
This is what LenaSYS looks like today in mobileview before the edit button has been removed.


<img width="292" alt="Skärmavbild 2022-04-13 kl  14 09 59" src="https://user-images.githubusercontent.com/81676911/163177991-9a2eeee0-18fa-49a1-af6b-45380b856445.png">

b20idawa:
This is what LenaSYS looks like in mobileview after the edit button has been removed. The edit button is not visible anymore in mobileview. 


<img width="292" alt="Skärmavbild 2022-04-13 kl  14 11 53" src="https://user-images.githubusercontent.com/81676911/163178509-4385c903-efe2-4abf-856f-833e4aa0ad9e.png">

b20idawa:
How to test the issue:

1. Log in as a teacher

2. Choose the webprogramming course

3. Select "PHP example 1"

4. Check that the edit button is visible

5. Then select "inspect the page" and select mobileview

6. For example, select "iPhone 12 pro"

7. Check that the edit button is no longer visible

8. The result should look like the picture below.
<img width="292" alt="Skärmavbild 2022-04-13 kl  14 11 53" src="https://user-images.githubusercontent.com/81676911/163327869-d6b5b6af-cb43-452c-b018-2113c1aabe0a.png">

a20willa:
Closing Issue as it has been solvled.

PR #11607, opened 2022-04-13
G2 2022 w3 #11487

b20idawa:


a18gusjo:
Code refactoring - ok
Code standard - ok 
Meet issue goal - yes

Issue #11598, opened 2022-04-13
Linux machines not loading contribution test database. 


a20behta:
When trying to load the contribution databases into either the webserver or a devs ubuntu machine it seems to not load anything, similar to no database. 
  * user access is given same as apache 
  * tried restarting apache
  * used the same method as on windows machines

b20eripe:
![bild](https://user-images.githubusercontent.com/102593527/163178627-40a85144-9aee-4fb2-9e58-c3b3dfae99fe.png)

Found an error with wrong slashes in the address. Windows doesn't care but Linux got mad. Adjusted the code

a20behta:
changes made into weekly branch

PR #11608, opened 2022-04-13
G3 2022 w3 #11598

b20eripe:
Fault with Linux machines solved, wrong type of slashes in address


a20behta (approved):
Changes direction of slash allowing linux to find correct position

Issue #11424, opened 2022-04-07
Add functionality to the column that should hold last visited date in resulted.php


c20adael:
The column has been added on to the site #11393. continue to add the functionality so that the last visited date is shown for each entry. 

This is part of the issue #11113

a20novra:
i researched in resulted.js, resulted.php,  and resultedservice.php. I found out that we should create a new column in userAnswer table to save teachers last visited time value. When we are done with this part then we can show it on the table as we want

a20novra:
changed SELECT query that islocated in resultedService.php to fetch data that exist in last_Time_techer_visited column (LINE 29)

a20novra:
![0](https://user-images.githubusercontent.com/81618150/162918575-23cfef3a-929c-49d8-bd8c-837e80626044.jpg)

a20behta:
merged into weekly branch looks good

PR #11609, opened 2022-04-13
G3 2022 w3 #11424

a20novra:
- Added a column in userAnswer table (alterTables.sql - Line 63-64 )
- Added last_Time_techer_visited column in select query (resultedService.php - Line 39)
- Added teacher_visited variable that include last_Time_techer_visited column data (resultedService.php - Line 63)
- Added if sats to show right data in teacher visited column(resulted.js - Line 246-249)
- Added uppdate query (index.php- Line 44-48)


a20behta:
Issue found when testing. 

image

when reinstall occured the column is not included thereby breaks the tables and cannot be merged. 
need to check again how/where to add the column to the table.

PR #11611, opened 2022-04-13
Prevent "enter" from reloading page

a20vicad:
Try to search files/duggas and press ENTER and check that the page does not reload.


a20behta (approved):
removed default on submit. 
tested and compared using webserver

PR #11612, opened 2022-04-13
Gave a unique class to edit button #11536

a20linth:
Previously the edit button used the reset button class and was therefore hidden in template 10 along with the resize button where no resize capabilities should exist.


a18gusjo:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #11615, opened 2022-04-13
fix for issue #11480 Diagram module - Update toolbar to accomodate UML entity

c20emili:
Added functions to open a box with more selection options when the mouse button is pressed for half a sekund on specific elements. 
image
image
image


a19meram:
Finds no problems after testing. The choice of uml, ER works well.

i19matla:
The issue is solved. Longpressure to get the expanding menu works fine. A finding during the testing is that the same button must be clicked to hide the expanding menu. A question to bring to the customer and perhaps it will be a new issue if the customer want to have the ability to click anywhere do hide the expanding menu.
image

Issue #11540, opened 2022-04-12
Diagram: Move grid toggle into options instead of toolbar.


b19karhj:
Atm the grid togggle is in the toolbar, It should be move into options.
![bild](https://user-images.githubusercontent.com/81624541/162929636-a32a857a-541e-4465-9150-4b62be9fa622.png)

a20samul:
- Toggle options were moved into options pane and the icons were later swapped out into buttons.

<img width="583" alt="iconsToButtons" src="https://user-images.githubusercontent.com/81613484/163261064-eb3d3db9-192e-45c1-923a-a0f930d2a320.png">

a20samul:
As the icons have been moved into the options panel it was discussed that the properties associated with the A4-template should be moved from the properties panel down to its main functionality to clarify which functions are available. 

**Suggestions:**
- Remove the slider for size change.
-  Vertical- and horizontal buttons pop up after A4-button is enabled. 
- Give the buttons better names alternatively also that the buttons will change value-name based on if it is enabled disabled + colour change based state.
- Alternatively move the A4-functionality into its own fieldset.

A mockup will be added here [under construction]

a20samul:
The design might be further altered for a nicer appearance. 
The mockup provides a visual solution for a for a unanimous look throughout the whole panel and for the functionality to be properly placed.

<img width="193" alt="optionspanel" src="https://user-images.githubusercontent.com/81613484/163945265-ddc0d3c1-7944-42c0-8451-87e65e0ee764.png">

PR #11621, opened 2022-04-13
Moved icons into options panel and changed them into buttons in diagram dugga #11540

a20samul:
Moved icons (Toggle grid, toggle snap to grid, toggle ruler, toggle A4 template) into options panel and changed them into buttons. 
Changed css for options panel and buttons to make it more presentable.

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -79,7 +79,7 @@
 #options-pane {
     position: fixed;
     top: 0px;
-    width: 280px;
+    width: 320px;
     bottom: 0px;
     z-index: 5000;
     background-color: #eb4;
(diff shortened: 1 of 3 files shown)


c20emili:
The buttons are moved and the funcionality is the same. It is hard to se what state the toggle snap to grid is because there is no visual change.

a20samul:
### For testing:
**Issue was expanded to test expanded solution**
- Name change on buttons.
- Slider removed from A4-template options.
- Properties for A4-template no longer shows up in the "properties-fieldset" and has been moved to under A4-template button and shows only when the A4-template is active i.e. Vertical- and horizontal buttons only pop up after A4-button is enabled.
- Colour changes on all toggle buttons based on if the functionality is active or not to clarify that functionality is in use (toggled on/off).

optionspanel2

a20gabpa:
Verified that the new buttons still has the same functionality. Small missalignment with CSS for lines properties.

PR #11622, opened 2022-04-13
Removed the square around the delete button #11387

a20edwth:
Removed the square around the delete button, now there is only an X.

Instructions for testing:
1) Click "Demo-Course"
2) Click "Diagram Dugga"
3) Click or box select any element/s
4) Click cross icon in the right upper corner of the selection box and the elements selected should get deleted


c20emili:
Tested looking and the square is gone. Tried the diagrams different functionalitys and they stayed the same.

Issue #11620, opened 2022-04-13
Update allowed file types to correctly reflect json files as well


a20micro:
When uploading a file a list of allowed file types is listed. This list doesn't include the latest addition of file types, .json files.
![CurrentAllowedTypes](https://user-images.githubusercontent.com/95532115/163201638-c8876675-9fd7-4852-99f6-648d80181c45.png)

All locations where such a list is shown should be updated to correctly match the file types allowed.

PR #11624, opened 2022-04-13
Added "JSON" to the end of a text string

a20micro:
Issue #11620
Same menu shows up for all types of file upload so simply adding "JSON" to the end of the <p> tag fixed updated all issue-locations.


b19karhj:
jason has been added to file suggestion. Works

PR #11625, opened 2022-04-13
fix for issue #11155 Select & Move labels in diagram

c20emili:
removed unnessecary zoomfact that caused issues

https://user-images.githubusercontent.com/81320323/163206219-cfbdeec3-019a-4e9a-8937-12ff8bd6f6db.mp4

issue number #11155


i19matla:
Works as it is supposed to.

a19meram:
tested on my computer and laptop everything seems to work well without any problems. semantic and syntax correct without flaws as I can see.

Issue #11617, opened 2022-04-13
Create icons for UML


c20emili:
The new UML buttons need icons that keeps the theme of the toolbar. 
diagram_UML_entity.svg
![image](https://user-images.githubusercontent.com/81320323/163199829-2ae0304c-fb51-43d1-8532-f9790ad8397b.png)
diagram_inheritance.svg 
![image](https://user-images.githubusercontent.com/81320323/163199891-952add29-c0af-427e-8352-c38a86d852ef.png)

b20thelo:
Icons has been approved and merged.

PR #11627, opened 2022-04-14
Created two new icons for the implementation of new diagram functionalities in diagram dugga #11617

a20samul:
Created two new icons; diagram_UML_entity.svg and diagram_inheritance.svg for the implementation of UML entities respectively incorporation of inheritance in diagram dugga. #11617


b19karhj:
Need to be pulled to @c20emili

Issue #11606, opened 2022-04-13
Error occurs when writing in the search bar on resulted.php. 


c20adael:
When using the searchbar located on the resulted.php you will get a error in the console. Why is this?
Also the search function seems to be not present. Nothing is being searched for.

HGustavs:
Search is bulit into sortabletable so it should work automatically

a20behta:
merged into weekly branch

PR #11631, opened 2022-04-14
Searching in the result editor now works #11606

a20vicad:
added .toString() to line 333 in resulted.js to make it sure it's a string that is compared to the search term. 

Testing: Enter course as superuser/teacher -> Press button "Edit student results" (paper with pencil) -> Attempt searching. (Note: ALL things can be searched for, including dates published and hashes.)


a20behta (approved):
only additon is making it to string. 

no reason it affects other parts of system

PR #11632, opened 2022-04-14
Can apply invalid deadlines, fixed validateSectName function #11224

a20linth:
Can now apply invalid deadlines when editing items in sectioned, and fixed validateSectName function for item titles.


a20muhal:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #11635, opened 2022-04-14
Added selected diagram file to be included in generated JSON params

a20micro:
Added support for selection of a file in the drop-down menu to be reflected in the automatic JSON params box on the right side of the "edit dugga" menu screen.


a20gabpa:
Works as expected, changes are shown in the params box.

Issue #11618, opened 2022-04-13
Filtering out non-json files from the "Edit Dugga - Add diagram to dugga" menu 


a20micro:
As part of issue #11401 teachers should be able to import diagrams to be shown on-load for students doing the dugga. Currently, the drop-down menu is connected to the filelink table in the database, but all files are shown.  

![AllItemsShown](https://user-images.githubusercontent.com/95532115/163200484-47b34126-5fd7-4104-81bb-fe870e5a0bd9.png)

Only .json files should be shown in this list to be selected.

a16linas:
Pull request has been accepted and branch has been merged. Closing issue

b20thelo:
Feedback from Henrik.

Wants the code to be written so that if in the future new file formats will be added that should be possible by adding another value in an array.

Example: {json, txt}

PR #11638, opened 2022-04-14
cleaned up test functions and now the dropdown box lists only json files

a16linas:
Removed function returnedSection() as I was using it to test whichever was best between that and returnedFile(). Furthermore, the dropdown now only lists json files. You can test this by uploading a global or local .json file then edit the variant of a dugga in the same course. Also added some comments describing what they do and where.

bild

### DIFF
--- a/DuggaSys/duggaed.js
+++ b/DuggaSys/duggaed.js
@@ -18,6 +18,7 @@ var typeOfItem;
 var duggaPages;
 var isClickedElementBox = false;
 var searchterm = "";
+var targetfile;
 
 function setup() {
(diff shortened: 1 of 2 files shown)


a19meram:
tested on my computer and it seems that the dropdown menu shows all types of files not just JSON files. Looks like no link was included specifying only json files and nothing else.

i19matla:
image
Only .JSON files are supposed to be shown but all files can be seen.

a16linas:
Incorrect testing by the testers, they managed to test something that wasn't on the branch. Looking at the inspector, old code is present that was removed on issue-branch #11618 prior to making the pull request. Suspect they accidentally made a local branch from whichever branch was active at the time.
bild

PR #11639, opened 2022-04-14
adaption description with canvas and drag resizer issue#11623

a19meram:
Added a resizer between the diagram and assignment description. Mock-ups available on issue#11623


a20gabpa:
Tested locally on other computer. It works but can't be verified on server due to restriction to Canvas.

PR #11640, opened 2022-04-14
Fixed error on mouseover in weekly bar chart

b20jonlu:
Fixed the error that happened on mouse over by changing the syntax of line 189 and 190 in contribution.js


a20behta (approved):
removed enter and placed into single line. 
no actual code changed

PR #11642, opened 2022-04-14
onbeforeunload added to showDugga.php

a17nikjo:
All duggas now promt a warning when leaving page if changes are made.


a20behta:
new pull with changes added

Issue #11633, opened 2022-04-14
Filter overlapp when editing student result


a20behta:
![image](https://user-images.githubusercontent.com/102468604/163347382-4e2ca20a-0e69-4ba1-ae04-8335cd0be669.png)
When logged in as a teacher and editing student result switching from one colum filter followed by pressing the second one it results in both being visable. 
tested on Webbutveckling-Datorgrafik and DemoCourse. 

both have the same issue .

when one is pressed the other one should be closed.

HGustavs:
Yes that solution is my preferred solution also.

c20adael:
@a20vicad Can you send a picture of how it looks now after merge?

c20adael:
https://github.com/HGustavs/LenaSYS/pull/11644#issue-1204343372

a20behta:
@c20adael currently our group branch looks the following vide

https://user-images.githubusercontent.com/102468604/163958770-2759eb20-2839-4db9-ba13-59aa8bd16766.mp4

o

c20adael:
@a20behta yep that looks good! also like the sound when pressing the buttons xD

a20behta:
Merged into weekly branch

PR #11644, opened 2022-04-14
Hide dropdownmenu when pressing the other button #11633

a20vicad:
Now when pressing one of the buttons in the result editor, the other dropdown will be closed, meaing that the two will no longer be open at the same time. 

image

Testing: Enter course as superuser/teacher -> Press button "Edit student results" (paper with pencil) -> Attempt opening both dropdowns by pressing both buttons.
image


a20behta (approved):
added the toggle functionality to hide the other respectively when toggled. 

looks good easy to read

PR #11647, opened 2022-04-14
Updated renderCommits in contribution.js

b20jonlu:
Updated renderCommits to take into account that the data may not have the same year as the current year(2022). Instead made it so it checks if the commit happened the same year as the first date in data. Also added some text to show the user the interval date for the commit tree.

### DIFF
--- a/DuggaSys/contribution.js
+++ b/DuggaSys/contribution.js
@@ -231,20 +231,28 @@ function renderBarDiagram(data) {
 
 function renderCommits(data) {
   
-  //creating the svg to put the commit tree in
-  var str = "<h2>Commit tree</h2>";
-  str += "<div id='innerCommitTree'>";
-  str += "<svg id='commitTree' viewBox='0 0 600 300' style='background-color:#efefef; width: 1200px; height:300px;' aria-labelledby='title desc' role='img'>";
-  
-
-  var current = new Date();
-  var currentYear = current.getFullYear();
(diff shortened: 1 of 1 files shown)


a20behta:
@b20jonlu What is the best method to test the changes and see the difference made?
looks like this is a specified period (läsperiod 2) can it be generalized to the entire year?

b20jonlu:
@a20behta There is a clear way to check the changes by going into contribution.php and looking at the commit tree with and without the changes. Without the changes nothing will be shown whereas with the changes a few dots with lines will be shown. The dates shown are static and can be changed via line 72 in contributionservice.php (haven't tested this but it should be the method to test other dates. also most dates are hard coded to be in an interval between 2019-04-01 - 2019-12-31 in the service but this should hopefuly not matter for this) . 
~~~
72 	$startweek = strtotime('2019-04-01');		
~~~
What the commit tree looks like by default
87e5b5df84a26dbed5307f5ee6a31b98

What the commit tree looks like with the changes using the big dataset with user MrRobot
63c2f746911830aba5493872c7040304


The changes are just intended to get the commit tree to work for the current year. The tree can only show a period of 10 weeks and this is hard coded in quite a lot of places so it would be somewhat difficult to change this.

a20behta (approved):
Code looks good. easy to read, nice addition to show dates for each year.

PR #11649, opened 2022-04-14
G3 2022 w3 #11122

a19simru:
Completed issue with a hard coded list of all red days from now till 2025. That list needs updating in 2026. Also the new Graph column displaying redDays works regardless of what time of the year the course is held. now. Previous problems with column title reading "Undefined" has been fixed to "red days". Previous problem with other graphs disappearing has been resolved. 

All changes are made in LenaSYS/DuggaSYS/contribution.js

### DIFF
--- a/DuggaSys/contribution.js
+++ b/DuggaSys/contribution.js
@@ -291,7 +291,12 @@ function getHolidays(date)
   var redDay4_12 = getYYYYMMDD(new Date("2023-12-24"));
   var redDay4_13 = getYYYYMMDD(new Date("2025-12-25"));
   var redDay4_14 = getYYYYMMDD(new Date("2025-12-26"));
-  holidays.push(redDay1,redDay2,redDay3,redDay4,redDay5,redDay6);
+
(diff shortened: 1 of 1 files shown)


a19simru:
This can be tested by switching to branch G3-2022-W3-#11122 and running http://localhost/LenaSYS/DuggaSys/contribution.php.

a20behta (changes requested):
DuggaSYS/contribution.js
- line 100 to 223:
   * Lines are from a previous solution and no longer relevant to the current solution. please remove
   * needs to be removed as difficulties to maintain. becomes extra code without purpose.
-  line 237 -293: the amount of red days per year differs. 
   * include all red days even if the fall on saturdays or sundays. (keeping it consistent)
   * is it possible to compress these line into arrays instead of seperate variables for each date?
line 350: why is the css changes?
line 542: same issue here, see no reason the padding should be changed. 
line 1463: think this can be a for loop instead (easier to read)
- line 1469 to 1491:
   * all these if statements seem to connect to the same add the same thing to str. can they be combined in any way?
 - line 1810 to 1841:
   * why does this exist?
   * is it relevant to your issue and if so how?
 
@a19simru check these things before we can merge.

a20behta:
Changes were not made before merge, will look at easiest way to move work to next weekly branch

PR #11660, opened 2022-04-19
Code to just copy important code

a19xinhu:
Change little code at function copyCodeToClipboard() at codeviewer.js. Make it just copy important code.


a20muhal:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #11661, opened 2022-04-19
Removed zoom buttons in mobile view

a20andlj:
Removed maximize and minimize buttons in mobile view because they did not work properly in mobile view.


a20muhal:
Code refactoring - ok
Code standard - ok
Meet issue goal - NO


REJECTED

The issue is solved staticly where if the screensize is 364px. it will disapear. Recommendation try making it dynamic.

PR #11668, opened 2022-04-19
Box around selected element: #11396

a20jacmi:
To mark the selection of elements (not lines) a dashed box is visible for each marked object. The implementation is found in "diagram.js" -> "function drawSelectionBox(str)".

Screenshot 2022-04-19 at 11 06 33


a20gabpa:
Tested and works as expected.

PR #11674, opened 2022-04-19
Make the dropbar do not disappear

a19xinhu:
Add few code to make the dropbar do not disappear when you hold 1s.


a20muhal:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #11646, opened 2022-04-14
In mobile view, when the width is too small, the timestamp will get cut off.


c20adael:
![image](https://user-images.githubusercontent.com/80856978/163378905-5b95d88b-c419-4db0-9b3e-c811e73844d5.png)

HGustavs:
Better to shorten the string to 2015-01-30 if screen is narrow. Is there a way to do this using css?

a20vicad:
Relevant information of changes and testing can be found in pull request.

a20behta:
merged into weekly branch

PR #11675, opened 2022-04-19
Minor layout changes depending on screen width #11646

a20vicad:
Now if screen width is 430 or below, the exact time of upload will not be shown. When the width is above, the time will be shown.

image

image

To test: Enter a course with available duggas, open inspect view, push the "phoneview button" image, check the different widths (remember to update the page).


a20behta (approved):
only added single if statement for width.
quick fix and easy to read

PR #11676, opened 2022-04-19
Fixed display issues in contribution.php

b20jonlu:
Fixed the problem of elements getting the wrong display style after clicking one of the buttons to change Statview. Now elements that are originaly flex will be flex after changing Statview and not display style block.


a20behta (approved):
Looks good, minor changes made. 

easy to read nicely done

PR #11680, opened 2022-04-19
Added an actionevent to the nav buttons and removed dropdown menu name.

a20patpe:
Added an event that runs when the button is clicked that offsets the button slightly, giving the user visual feedback that the button has been clicked.
Also removed the name of the Dropdown menu.

For testing:
1: In codeviewer when pressing the backwards/forwards buttons the button should offset slightly indicating it has been clicked.
image

2: When holding the button the dropdownmenue should not have a name.
image

### DIFF
--- a/DuggaSys/codeviewer.php
+++ b/DuggaSys/codeviewer.php
@@ -213,8 +213,8 @@ Testing Link:
 
         </div>
 		<!-- Dropdowns START -->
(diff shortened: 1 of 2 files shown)


a18gusjo:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #11558, opened 2022-04-12
Save functionality dugga edit isn't consistent.


b20thelo:
Found under "Tillgängliga duggor" at /DuggaSys/sectioned.php?
![image](https://user-images.githubusercontent.com/81772705/162964105-fb26c5af-de0f-4868-b75a-bff5921fc1f8.png)
Allows clicking on save and Alert is shown.

Found under "Show test" at DuggaSys/duggaed.php?
![image](https://user-images.githubusercontent.com/81772705/162964188-3650c1be-f9e3-454b-8f42-dbf0669ea8c0.png)
Save greyed out

Make sure both dialogues work as last image.

a20jonda:
found in DuggaSys/courseed.php using method in DuggaSys\courseed.js

![image](https://user-images.githubusercontent.com/102598258/163954223-4644b1fd-b7df-4375-8213-18fa09fe6bfa.png)
Save not grayed out.

a20jonda:
Functions of interest are 
- duggaed.js / validateDuggaName() 
- setioned.js / validateSectName(name)
- courseed.js / elementIsValid(element)

a20jonda:
![image](https://user-images.githubusercontent.com/102598258/164227830-0cc5504d-14a6-4059-843a-cb010c1c8c9c.png)
Regex will be changed to include dashes, otherwise Demo-Course will need to change its default name.

i19matla:
merged and done

a20jonda:
Still working on this

PR #11683, opened 2022-04-19
G1 2022 w3 #11558: Save functionality dugga edit isn't consistent.

a20jonda:
Added lines to make the 'save' button get disabled once the inputed name didn't fit the criteria. 
#11558


b20thelo:
Will close this pull request as it will is included in next weeks merge instead since further functionality needs to be added.

PR #11684, opened 2022-04-19
G1 2022 w3 #11342

a20chrka:
This is for issue #11342:

With this pull-request, there are three fixes:

**1. The zoom-buttons will now allow you to zoom towards/from the center of the screen, regardless of the position of the diagram.** 
The previous implementation was zooming towards a constant coordinate based on half the width and height of the screen (the "window" that the diagram is in), regardless of the position of the diagram. 

Consider the following example:
The size of the diagram screen is 1000 * 800.
the range of coordinates of the diagram that is currently visible is x: -1200 to 0 & y: -950 to 0.
With the old implementation, the screen would zoom towards x: 1000 / 2, y: 800 / 2, a coordinate that is not even visible on screen (because the length and height of the screen are constants, as long as you don't resize the browser and also completely unrelated to the coordinates in the diagram).
With the new implementation, the screen would zoom towards x: -600, y: -475, the coordinate that is CURRENTLY in the middle of the screen.

How to test:
(Do this on both the old and the new version to compare)
1. Move the diagram far away. 
2. Remember the approximate coordinate that is currently in the middle of the screen.
3. Zoom in or out.
4. After zooming, is the coordinate in the middle of the screen the same as before?

**2. The same fix for zooming with the buttons has also been applied to zooming with the mousewheel.**
The previous implementation had scaling problems with certain variables, making the screen jump around, seemingly at random, when zooming in or out after moving the diagram far away. The new implementation always zooms towards the mouse.

How to test:
(Do this on both the old and the new version to compare)
1. Zoom in with the mousewheel.
2. Move the mouse to a new coordinate. Remember this coordinate.
3. Zoom in again without moving the mouse.
4. After zooming, is the mouse on the same coordinate as in step 3?
5. Repeat step 1 - 4 but zoom out instead of zooming in.

**3. With the old implementation, you could zoom in and out with the zoom buttons even though minimum/maximum zoom had been reached. It would not zoom any further but it would still update variables with incorrect values, making the screen jump. This has been fixed with the new implementation.**

How to test:
(Do this on both the old and the new version to compare)
1. Zoom in until you reach zoom factor x4 (visible in bottom left corner)
2. Zoom in again. Did the screen jump?
3. Repeat with zooming out instead of zooming in.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -892,7 +892,7 @@ var scrolly = 100;
 var zoomOrigo = new Point(0, 0); // Zoom center coordinates relative to origo
 var camera = new Point(0, 0); // Relative to coordinate system origo
 var lastZoomPos = new Point(0, 0); //placeholder for the previous zoom position relative to the screen (Screen position for previous zoom)
-var lastMousePosCoords = new Point(0, 0); //placeholder for the previous mouse coordinates relative to the diagram (Coordinates for the previous zoom) 
+var lastMousePosCoords = new Point(0, 0); //placeholder for the previous mouse coordinates relative to the diagram (Coordinates for the previous zoom)
 
 var zoomAllowed = true; // Boolean value to slow down zoom on touchpad.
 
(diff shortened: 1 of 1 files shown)


a20gabpa:
Works as expected, tested on touchpad also. The zoom works great, did not encounter any jittering moving around and zooming.

PR #11687, opened 2022-04-19
dugga link now works on background element

a17nikjo:
Previusly you had to click the href link now you can click most of the row (other buttons excluded) to get to the dugga page.


a20behta (approved):
Clear and minimal changes made, easy to read.
good to use child href link, makes it easier to maintain.

PR #11689, opened 2022-04-19
Revert "Box around selected element: #11396"

b19karhj:
Reverts HGustavs/LenaSYS#11668


b19karhj:
Repair merge, it was merged into master which was not correct

PR #11691, opened 2022-04-19
Revert "Box around selected element: #11396"

b19karhj:
Reverts HGustavs/LenaSYS#11668


b19karhj:
Reverted worked, this is not needed anymore.

PR #11697, opened 2022-04-19
G3 2022 w3 #11424

a20novra:
i have changed the place of alter table userAnser code from alterTable.sql to init_db.sql line  184- 185. You can test the functionality with going to this path "LenaSYS/DuggaSys/resulted.php" when you click on a hash link it will save the current time to database and showing it on the table 
Sc11


a20behta (approved):
Looks good. 
easy to read, clear changes, minimal required, 
added to sql files for future installations

PR #11699, opened 2022-04-19
G1 2022 repair#3

b19karhj:
#11668 Was pushed to master , then reverted, but could not make a new pull request so added the code by hand into a new branch.


b19karhj:
Repair merge as #11668 was pushed into main.

Issue #11416, opened 2022-04-07
Creating login page and pop up for contribuiton


a20behta:
TLDR: Mockup implemented and created as a pop up if someone who is not logged in wants to access contribution.

#11415 required to be accepted before development can being.

Details: when a user tires to access  DuggaSys/contribution.php it requires them to be logged in. If they are not logged in the user will be promted to log in. implement that pop up based on mock up accepted.

a20behta:
related to issue #11223

c20adael:
@a20sacka When @a20caros has updated #11415 you should be good to go.

a20sacka:
Added functionality that wasnt documented:
The exit(big red) button that leaves the login screen now redirects you to previous page to avoid getting an empty contribution page.

a20sacka:
Made it so if a user isnt logged in they will get a pop up that requires them to log in and then get contribution loaded. If they are already logged in they will not get the popup and get contribution loaded directly.

Will now work on the "github user part"

a20sacka:
now working from branch G3-2022-W4-#11416 after merge https://github.com/HGustavs/LenaSYS/tree/G3-2022-W4-%2311416

Issue #11223, opened 2022-04-01
Contribution login interface


HGustavs:
We need a login interface for the contribution tool.

It should contain the following:

#1 If logged in as a teacher we see the teacher view e.g. all contributions in for a specific repository.
   * We must also have an interface for confirming contribution accounts or revoking previous confirms
#2 if **not** logged in as a teacher we see one of the following things
   * We show a "create new contribution or login" page
       * If login  is present in github data we should allow user to create a password
       * When password is created, teachers must then be able to confirm this login for viewing contribution progress
   * If previously logged in to a contribution login we show either 
         * If not previously confirmed we show "not yet confirmed" view
         * If confirmation has been revoked, we show "permission revoked"
         * If confirmation has not been revoked we show contribution view for logged in student
 
Contribution login information should be kept in normal lenasys database. No personal information should be stored, only github login name and password.

Only github logins present in github data should be available for creating logins.

Please divide into suitable sub-issues and create mock-ups for the suggested user interface

PR #11700, opened 2022-04-19
Adding functionality for creating new users from git login and password popup for contribution.php #11416

a20sacka:
Its on its good way to be completed but it is not completed as of yet, the changes made however will only affect contribution.php if you arent logged in and shouldnt break other parts of the system.

Changed resetpw.php to handle more states of "user creation"
contribution.js changed to add UI functionality for this special case of contribution login.
dugga.js changed to add functionality of special cases to login-change-password case when you are in the pending/revoked states.

One quite special thing which might cause problems for other parts of the system can be how i handled the column requestedpasswordchange in the user table; at the moment it is a tinyint when i assume it should have been a boolean, however because it is a tinyint this allows me to store **more** "requestedpasswordchange"-states in it like the states described in issue #11223 for the states **pending**, **revoked** and **accepted** currently they are represented numerically in requestedpasswordchange with **101**, **102** and **103** respectively(the **103** accepted state might not be needed).

### DIFF
--- a/DuggaSys/contribution.js
+++ b/DuggaSys/contribution.js
@@ -1570,4 +1570,29 @@ function hideTooltip() {
     }
 }
 
+function forceUserLogin()
+{
+      /* 
+      make sure the loginBox is generated before you run this function as 
+      it queries for elements that exist in the loginbox and changes their properties
+      */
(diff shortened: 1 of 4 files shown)


a20behta (approved):
looks good, seems incomplete and some else if statements need follow up with else aswell.
approved as nothing should break anything else on the site, allow for continuation on same part for next weeks branch

PR #11709, opened 2022-04-20
CHanged the position of the textbox to static.

a20patpe:
Changed the position to static from absolute.
This makes it so that the textwrapper will take up all the space available to it instead of taking up all the space it needs to display the text.

For testing: 
1: Go into "Javascript example 1" (Or any example that has a significant amount of text in it.)
2:Go into inspector(Ctrl+Shift+i).
3: Activate select element(Ctrl+Shift+C)
4: When checking the "Normtextwrapper" it should cover the entire box.
image
5: When selecting text you should be able to do it far away from the text.
image

### DIFF
--- a/Shared/css/codeviewer.css
+++ b/Shared/css/codeviewer.css
@@ -341,7 +341,7 @@
 
 .normtextwrapper{
     /* Width of codeborder */
-    position: absolute;
+    position: static;
     top: 0px;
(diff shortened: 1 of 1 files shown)


a20muhal:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #11710, opened 2022-04-20
href now found by searching all children

a17nikjo:
you can now click on the background elements for duggor at webbprogrammering aswell.


a20behta (approved):
looks good, 
annoying solution but only one for unknown/varying id and class names

Issue #11512, opened 2022-04-11
sh - shorter url-links and Canvas search function integration with LenaSYS should be name independent


a20sarle:
This issue relates to both #10566 and #11219 where information about how hotlinks, search and LenaSYS integration with Canvas can be found. The previous documentations should describe how iframes are chosen and shown to be displayed in Canvas' own iframe.

Today there's most likely a long link-name used to choose which iframe to be shown in Canvas' own iframe on the course and assignment pages. The goal is to make this link as short as possible. Todays' solution and hotlink is most likely based upon the course name which results in varying issues.

What happens if a course and/or src/file in iframe change name? Can Canvas' iframe still find the right examples and courses in LenaSYS? (Better to link / use id that's not based on name).

Imagine if Canvas iframe has linked a course with the name "database" and it changes name to "data systems". In this case all the previous connections should still be the same and the iframe on Canvas should continously link to the same source. If a new course named "database systems" is added, the "link-search"/src for the course "databse" should still be linked exactely as before.

Once an "LenaSYS-iframe-source" has been found/connected with an iframe in Canvas the link should be constant and never changed even if the course name is changed or multiple courses with similar names has been created.

a20sarle:
The src-link is related to the serach query in the iframe on Canvas.
![image](https://user-images.githubusercontent.com/81631818/162740219-8bb56e51-d698-490b-9432-8ff9e60e4b39.png)

a17nikjo:
Related to issue #11377 
Some comments from HGustav can be found here.

a17nikjo:
in /sh/index.php we create links starting with " header("Location: /DuggaSys/", however these links do not work when testing locally. A fix is adding .. infront liko so  "header("Location: ../DuggaSys/". This allows for testing on localhost.

a20linth:
The URL shortener currently looks like this:
<p><iframe src="https://dugga.iit.his.se/sh?c=Demo-Course;a=daily-minutes" width="800" height="1200"></iframe></p>

but if course id(cid) and assignment id(id) are instead used, the url shortener could look like this:
<p><iframe src="https://dugga.iit.his.se/sh?c=1894;a=20" width="800" height="1200"></iframe></p>

Courses (course) have an unique id (cid) for every course that cannot be modified.
![image](https://user-images.githubusercontent.com/102609651/163979105-e0f52bfb-fafd-4d71-b3b1-d15299c72e68.png)

[...]

PR #11711, opened 2022-04-20
Fixed URL parameter getters and headers

a20linth:
Fixed SH URL parameter getters and gave all headers in sh/index.php a relative path #11512


a18gusjo:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

a20linth:
This solution did not meet customer requirements

PR #11714, opened 2022-04-20
Removed zoom buttons in mobile view

a20andlj:
Removed maximize and minimize buttons in mobile view in media query. And increased the pixel width from 364 px to 480 px.


a20muhal:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #11222, opened 2022-04-01
Too restrictive dialog when creating new course version


a97marbr:
![image](https://user-images.githubusercontent.com/4948661/161211619-d66fa4fe-e2b0-4017-bc6c-e5f69df41e21.png)

This dialog is too restrictive. We must be able to have version id longer than 6 characters and I'm not sure why we don't allow characters other than digits? THe following should be ok I think

- 789231311
- spring-3214
- 123123-123123

Version name also needs to be less restrictive. I think the following would be OK version names

- VT-22
- VT-22lp2
- Spring

a20novra:
jag want to work with this @c20adael  @a20behta

a20behta:
@a20novra Looks good can begin work

a20novra:
![1](https://user-images.githubusercontent.com/81618150/161734104-88615325-a203-4a70-a48a-b2088fa638cd.jpg)
![2](https://user-images.githubusercontent.com/81618150/161734123-8ef96495-4070-4fcd-8901-ae6e138f13f4.jpg)
![3](https://user-images.githubusercontent.com/81618150/161734148-c0017964-e339-42d4-97af-955e6f05b31f.jpg)


When i try to change Version Name Regexp i get this error, but i can't find header.php file. (OBS! it create the new version anyway but i cant understand why we get error)

c20adael:
The file seems to be named navheader.php. You should investigate that file since it exists in the folder that is specified. @a20novra

c20adael:
i dont know why an "n" is missing from the filepath

a20behta:
@a20novra New branch created from new weekly branch. 
can continue developement from here

a20novra:
I think this issue should be divided into twp parts. The first part should be about RegEx, and the other part should be about why when we try to copy content from exsiting course we get error massage .

PR #11716, opened 2022-04-20
G3 2022 w4 #11222

a20novra:
I changed regex on file sectioned.js Line 2409 and Line 2444. 
Select query changed in anheader.php, to fetch start date of course too(Line 127). Year value is equl to year that course starts(Line 136-137).
newCourse

 You can test it by creating new course with information below or other information that you want, but be carefull that copy content from should be none, otherwise you get error. There is other issue for solving this problem. 
testnew


c20adael (approved):
This works as intended for now

Issue #11549, opened 2022-04-12
Diagram: Ghost element should not exist after you place an element on the diagram


b19karhj:
When you selected an element to paste in the diagram, it appears as ghost before and after you have painted it out. 
![bild](https://user-images.githubusercontent.com/81624541/162951554-67c72a15-36a9-43e7-ab6f-468bb21a5154.png)
The ghost element to the right should not be there.

b20thelo:
Also when holding down the mouse button on the canvas and pulling to the side the canvas should be moved. No new element should be added.

b20thelo:
Disabling ghost entities also removed lines when being drawn therefore this issue needs another solution and previous pull request were denied.

PR #11717, opened 2022-04-20
Solution for issue #11549

a20chrka:
Issue: #11549

Changed the opacity of the ghost element from 0.5 to 0. In other words, the ghost element is still there, but it is invisible. The "ghost lines" should still be visible. It seems to me that the ghost elements are not only visible features but also have a functional purpose. I didn't find any other solution without breaking parts of the system.

Test by selecting one of each element and see if the ghost appears before placing it. Only the lines should be visible when attaching it to the first object.


b20matad:
Test passed! no problems found.

i19matla:
Works fine

PR #11718, opened 2022-04-20
Changed code for LenaSYS title dissappearing so that it works correctly#11641

b20hanwa:
The code for LenaSYS title dissappearing is now fixed and works correctly.


a20behta (approved):
Clean css code change, max size added

Issue #11701, opened 2022-04-20
Codeboxes in codeviewer -> HTML_Ex1 are not alligned correctly by default


a18gusjo:
The codeboxes in codeviewer are not alligned correctly until the reset button is clicked.
![2022-04-19 21-37-06](https://user-images.githubusercontent.com/102605031/164084105-3e9cfb67-7b08-44fd-a0b3-a672a0e322a8.gif)

HGustavs:
This does not have anything to do with the closed accordions?

One of the most notable changes after clicking reset is that accordions are expanded.

a18gusjo:
> 

I found where the issue lies by using breakpoints when clicking the HTML_Ex1 link:

https://user-images.githubusercontent.com/102605031/164222200-c808d407-084c-40a5-8946-ae9e72411c9b.mov

Removing the function call "alignBoxesHeight3stack(boxValArray, 2, 3, 4);" inside the if-statement for templateid = 6 seems to solve the problem.

a18gusjo:
Me and Mohammed tested the solution on the webserver and it didn't work. I will investigate further.

PR #11719, opened 2022-04-20
G2 2022 w4 #11701

a18gusjo:
How to test:
Open HTML5_Example 1(templateid=6) and make sure it doesn't look like this: 

Then create a new code example with template 7 and test the same thing.


a20muhal:
REJECTED 

Removes all the boxes.

a18gusjo:
Seems like HTML5_Example 1 is not "templateid=6" but "templateid=8". We need to test this again cause it might work.

Issue #11708, opened 2022-04-20
Edit button still not showing in template 10


a20willa:
Another issue (#11607) accidently broke a fix in #11536. The code need to be adjusted to fix the issue.

HGustavs:
Screenshot please to confirm that this is indeed a bug

a20willa:
The edit button (among other icons) are missing in the below image
![image](https://user-images.githubusercontent.com/81626616/164219558-e9cde900-d51d-4d81-b671-182ae8357704.png)

PR #11720, opened 2022-04-20
Fixed hideMaximizeAndResetButton() #11708

a20linth:
Removed hide editbtn to allow editbtn to be shown in template 10


a18gusjo:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #11721, opened 2022-04-20
Changed functionality for selectionbox

i19matla:
While marking inside the ER-diagram only the rectangle with solid lines can be seen, and after letting the mousbutton go the dashed selectionbox appears.


a20chrka:
The dashed selection box no longer appears and therefore the test has passed. However, some design choices needs to be reviewed. When using the selection tool with this implementation, lines will be marked (before releasing the mouse button) while other entities is not marked.

a20samul:
Individual elements are not marked i.e. when less than two objects are selected. See image below.
doesntshowoneismarked

PR #11722, opened 2022-04-20
Created a hover function that display a div when hovering over the commit links

a20caros:


b20jonlu:
From what i can see all commits enable the same <div id="commitDiv> which works for the commits in the first week but not the ones further down. such as those for week 4.
theseCommits

There is a second div but it is not enabled by the commits in week 4 so i am not sure what it does.
secondCommits

a20caros:
image

The problem is solved by placing a div in front of the table. The div has a fixed position which means that it follows the side
@b20jonlu

c20adael (approved):
It works as intended

PR #11733, opened 2022-04-20
Added a toggle icon in main header

b20fanfa:
Now there is a toggle/icon in the main header. The next step is to implement the function, so when you click dark or light mode is activated.


a20andlj:
Code refactoring - ok
Code standard - ok
Meet issue goal - no
Comment: The toggle icon is on the wrong side in the menu.

Issue #11600, opened 2022-04-13
Error checking for relations


a20edwth:
This feature checks the diagram for errors and outlines them in red(stroke set to red). Errors could be an entity not having a primary key or foregin key, no cardinality set or wrong relation type(double or single lines). This feature should also be able to be turned on or off

This issue is a sub-issue of issue #11233

a20edwth:
### Error checking mockup

Suggestions for how errors in the diagram will be represented to the user. 
“Suggestion 1” shows the user which element/s the error affects, but doesn’t clearly show what the cause is. 
“Suggestion 2” shows the cause of the error more clearly.

**List of errors:**

1. Wrong key
2. No key
3. More than one key
4. Cardinality missing
5. Wrong relation lines(double or single lines)
6. Wrong relation type

**Wrong key:** 
Suggestion 1
![error_bild1](https://user-images.githubusercontent.com/102611046/163376467-3723e4ae-b5ac-4ed2-9b8d-c2b1e6a5aaae.png)


Suggestion 2
![error_bild2](https://user-images.githubusercontent.com/102611046/163376526-515b38a0-a7cb-4e01-92a4-e16d63abf2de.png)



**No key:**
Suggestion 1
![error_bild3](https://user-images.githubusercontent.com/102611046/163376906-0eb59c09-c18a-4a70-9789-484ccfbac60b.png)


Suggestion 2
![error_bild4](https://user-images.githubusercontent.com/102611046/163376921-2a2534d0-ee9a-41b9-84cf-f0e2e8aa6391.png)



**More than one key:**
Suggestion 1
![error_bild5](https://user-images.githubusercontent.com/102611046/163376953-720acdd3-ff95-418b-b8d2-8a6448248230.png)


Suggestion 2
![error_bild6](https://user-images.githubusercontent.com/102611046/163376971-03b6379d-b9f0-4714-94dd-beb2eacf23f2.png)



**Cardinality missing:**
Suggestion 1
![error_bild8](https://user-images.githubusercontent.com/102611046/163376994-9af13b36-95e4-487b-94a0-51e193501e0d.png)


Suggestion 2
![error_bild7](https://user-images.githubusercontent.com/102611046/163377012-007ff9e0-8c00-4be0-9195-71bea9d97825.png)



**Wrong relation lines(double or single lines):**
Suggestion 1
[...]

PR #11743, opened 2022-04-21
Error checking for ER diagrams

a20edwth:
Error checking for ER diagrams, errors will be outlined in red. #11600

Instructions for testing:
Note: testing requires knowledge in ER diagrams.

1) Click "Demo-Course"
2) Click "Diagram Dugga"
3) Change the diagram to test all problems listed below:

1. Wrong line type, **correct**:  single line for strong entities and double line for weak entities
2. Wrong key type, **correct**: strong key for stong entities and weak key for weak entities
3. More or less than one key 
4. More or less than 2 lines to entities from a relation
5. Attributes with same name on the same entity
6. No cardinality set (Cardinality is the "N" or "1" situated on a line from an entity to a relation)
7. Wrong relation type, **correct**: weak relation must have a line to a weak entity and strong relation must have lines to two stong entities

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1094,7 +1094,7 @@ function onSetup()
         { id: makeRandomID(), fromID: EMPLOYEE_ID, toID: SUPERVISION_ID, kind: "Normal", cardinality: "MANY" },
         { id: makeRandomID(), fromID: EMPLOYEE_ID, toID: SUPERVISION_ID, kind: "Normal", cardinality: "ONE"},
         { id: makeRandomID(), fromID: EMPLOYEE_ID, toID: DEPENDENTS_OF_ID, kind: "Normal", cardinality: "ONE" },
-        { id: makeRandomID(), fromID: EMPLOYEE_ID, toID: WORKS_FOR_ID, kind: "Normal", cardinality: "MANY"},
+        { id: makeRandomID(), fromID: EMPLOYEE_ID, toID: WORKS_FOR_ID, kind: "Double", cardinality: "MANY"},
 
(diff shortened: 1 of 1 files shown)


c20emili:
the error detecting system works for the faults mentioned in the first comment but a weak entity are not allowed to have anything but weak relations which is incorrect.
image
and doesn't can't there be a double line on a strong relation to show a forced relation?

c20emili:
This fault are not entirely correct: "1.Wrong line type, correct: single line for strong entities and double line for weak entities"
Strong entitys **can** have double lines to show a forced relation. a weak entity can have different types of relations as long as it has one weak relation to a different entity.

a20edwth:
Added a commit where enities can now have a forced relation and weak entities can have different relations as long as it has one weak relation.

Testing follows the same instructions.

c20emili:
Test failed. a weak entity can be weak to another weak entity which this does not allow (se picture below).
image
and a weak entity needs a weak relation which this does not reflect (se picture below).
image

a20edwth:
The entity gets an error since a weak entity needs a weak key which it does not have, and a weak entity cant have a strong key thus an error for the strong key.

a20samul:
1. Error doesn't disappear if you accidently delete the line and replace it with the same line again.
errordoesntdissappear

2. The weak entity has a weak key but the error does not disappear from employee element.
weakkey

3. It would perhaps be more correct if the line going from employee to the relation was coloured red rather than the diamond.
lineshouldberedinstead

4. The behaviour is not consistent, "EMPLOYEE" doesn't become red whereas "DEPENDANT" does when same error is made.
bdale

The behaviour is not consistent through out the functionality. This issue needs improvement.

Tested on three different laptops.

Issue #11727, opened 2022-04-20
Icons disapear too fast when hovering over plus sign


a20willa:
Hovering over the plus sign in the dugga list while logged in as a teacher, the icons will disappear very fast, and thus you sometimes can't click them in time. We should add more time to click them.

a18gusjo:
This is how it looks at the webserver today:

https://user-images.githubusercontent.com/102605031/164246887-6d858f6b-4638-4aaf-93ca-9b4379d34dc0.mov

PR #11746, opened 2022-04-21
G2-2022-24-#11727

a18gusjo:
Updated the distance between the FAB icon (yellow icon with a + inside at the bottom of the page) and the content that is displayed when hovering over FAB.

**For testing:**
1. Switch to the group branch.
2. Login as stei
3. Go to: [this_link](http://project-g2.webug.his.se/DuggaSys/sectioned.php?courseid=1&coursename=Webbprogrammering&coursevers=45656)
4. Try to create a new example with the yellow icon at the bottom of the page.
5. Repeat the process with the branch: G2-2022-W4-#11727
6. See if the items disapear when you try to hover over them. If not, the bug is fixed.

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -6665,7 +6665,7 @@ only screen and (max-device-width: 320px) {
   width: 100%;
   list-style: none;
   text-align: center;
-  bottom: 90px;
+  bottom: 80px;
 }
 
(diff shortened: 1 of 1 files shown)


a20andlj:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #11643, opened 2022-04-14
Teachers should be able to select which type of diagram the student are able to draw in diagram dugga's.


c20emili:
The teachers should be able to choose which type or types of diagram such as ER or UML or both the dugga contains. Depending on the diagram types the dugga contains the student should have access to different tools in the toolbar.

HGustavs:
I suggest that this is done using a setting in the test json string.

PR #11747, opened 2022-04-21
fix for issue #11643 Teachers should be able to select which type of diagram the student are able to draw in diagram dugga's.

c20emili:
changed diagram.js mainly in showDiagramTypes(), diagram.php, duggaed.php, duggaed.js mainly in createJSONString(formData) and diagram_dugga.js. added checkboxes to the duggaed interface and added that the diagram types changes depending on what checkboxes was checked when creating the dugga variant.
image
image
image
image
test by:
-logging in as toddler
-go to demo courses 
-click show test 
-click the pen on the same line as Diagram Dugga 
-select an enabled existing variant by pressing the cogwheel on the same line
-change the checkboxes to select the types allowed
-press update
-log out
-go to diagram dugga
-look at the availible entitys
-repeat to test different types

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1161,7 +1161,8 @@ function getData()
 }
 
 function showDiagramTypes(){
-    if(!!diagramType.ER && !!diagramType.UML){
+    console.log(diagramType);
+    /*if(!!diagramType.ER && !!diagramType.UML){
         document.getElementById("elementPlacement4").classList.add("hiddenPlacementType");
     }
     else if(!diagramType.ER && !!diagramType.UML){
@@ -1174,7 +1175,7 @@ function showDiagramTypes(){
     }
     else{
         alert("NONE");
-    }
+    }*/
 }
 
 /**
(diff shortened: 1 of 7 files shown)


a20gabpa:
I have tested the solution and it works. The teacher can chose one or two of the diagram types and is reflected in the diagram dugga as a student. 
Some comments:
- It is possible to place other diagram types using the keybindings. For example only allowing UML, I could place ER entities after pressing '3'.
- As a teacher, when editing the diagram dugga, after clicking the cogwheel on a enabled variant, initially the values on the checkboxes does not show the saved configuration in the .json param. (saved from earlier) 
-  It is possisble to deselect every diagram type. If all is deselected, both ER and UML are still accessable in the diagram dugga as a student.

c20emili:
Added as a teacher, when editing the diagram dugga, after clicking the cogwheel on a enabled variant, initially the values on the checkboxes does show the saved configuration in the .json param. (saved from earlier)

a20gabpa:
The value of the checkbutton are updated according to the json params. Works as expected

PR #11752, opened 2022-04-21
Delete button resized and repositioned

a20edwth:
Moved the position of the delete buton and made it bigger. #11387

Instructions for testing:
1) Click "Demo-Course"
2) Click "Diagram Dugga"
3) Click or box select any element/s
4) Click cross icon in the right upper corner of the selection box and the elements selected should get deleted

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -5486,21 +5486,21 @@ function drawSelectionBox(str)
 
         //Determine size and position of delete button
         if (highX - lowX + 10 > highY - lowY + 10) {
-            deleteBtnSize = (highY - lowY + 10) / 4;
+            deleteBtnSize = (highY - lowY + 10) / 3;
         }
         else {
-            deleteBtnSize = (highX - lowX + 10) / 4;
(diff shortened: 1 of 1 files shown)


c20emili:
works when using box select and on lines but not on elements when pressing outside the elements selection box.

a20edwth:
Made new commit that makes it possible to delete single elements. Same testing instructions applies.

a20gabpa:
Test failed. It works when only using the pointer tool to click on the delete button. However when using the box selection tool to select a single element, and then clicking the delete button (X), does not work. If I however switch back to the pointer mode with the element already selected (with box selection) it works. It also work when multiple elements are selected and on single lines using box selection. 

image

Works to click on X:

image


Does not work to click on X:

image

a20edwth:
Fixed the problem above

a20samul:
Tested by pressing delete on single and multiple elements both using the pointer and the box selection tool which lead to element/s and line/s deleteing. Delete button is resized and repositioned. Functionality works as intended, test passed.

b20thelo:
Wait with pull request as issue #11396 might include this solution too. In case it does this pull request will be closed.

PR #11753, opened 2022-04-21
Logout redirects to current page

a20muhal:
Logout will be now redirected to the current page, while also loggin the user out.

testing

1. Log in to any user.
2. Go to any page. 
3. Log out.
4. If you get redirected to courseed.php, this has failed. But if you're on the same site where logout request was issued it passed.


a20andlj:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #11667, opened 2022-04-19
Moving several selected objects in Diagram


b20matad:
When several elements are selected you should not need to hold ctrl to move them all, also you should be able to click and hold anywhere within the selection box to move all objects.

PR #11755, opened 2022-04-21
Fixed issue #11667. Moving elements inside "selection box"

b20matad:
Move multiple objects by pressing down mouse inside selection box or an element inside the selection box. Works for 
the mousemodes: "POINTER" and "SELECTION_BOX".

All changes were made to "diagram.js" in the following methods: "mdown()", "ddown()", "mup()", "drawSelectionBox()".


c20emili:
all elements selected gets moved and if one element gets to close to a different element when moving the slected elements the element that gets to close get automatically moved and not all the selected elements.

Issue #11748, opened 2022-04-21
Check if possible to create default toolbar in diagram dugga to show teachers when they are logged in.


c20emili:
currently the toolbar for in diagram dugga depends on the version of the dugga the student have and the teachers don't have that when they're logged in so check if its possible to create a default toolbar setting so the teachers dont have a weird toolbar. connected to issue #11643. 
Teachers toolbar
![image](https://user-images.githubusercontent.com/81320323/164451808-ae79ce25-546c-4bcf-8725-24a3f2bf18c4.png)
Student toolbar
![image](https://user-images.githubusercontent.com/81320323/164451918-f05a12f6-7c92-48df-abd1-9db4d919c4e9.png)

HGustavs:
Im thinking we go by a selection in the menu whether we are making an ER/UML/Either etc type diagram and the menu depends on this choice. Or am I missing something. I dont understand what the difference is exactly.

i19matla:
Issue merged

PR #11756, opened 2022-04-21
fix for issue #11748 Check if possible to create default toolbar in diagram dugga to show teachers when they are logged in.

c20emili:
changed diagram_dugga.
Student version
image
Teacher version
image
test by:
-logging in as toddler
-go to demo courses
-click show test
-click the pen on the same line as Diagram Dugga
-select an enabled existing variant by pressing the cogwheel on the same line
-change the checkboxes to select the types allowed
-press update
-log out
-go to diagram dugga
-look at the availible entitys
-log in as teacher and look at the icons for entitys and relations
-repeat to test different types

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -961,7 +961,7 @@ var movingObject = false;
 var movingContainer = false;
 
 //setting the base values for the allowed diagramtypes
-var diagramType = {ER:true,UML:false};
+var diagramType = {ER:false,UML:false};
 
 //Grid Settings
 var settings = {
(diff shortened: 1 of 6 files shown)


a20edwth:
Followed testing instructions. Everything worked as intended.

a16linas:
Added functionality where the ER and UML checkboxes are now hidden when the dugga is not of type diagram. This is testable by editing variants of diagram duggor and non-diagram duggor

bild

a20samul:
Tested on three different laptops. 
Looked through all created variants of duggor both diagram_dugga and duggor which are not of type diagram. 
Checkboxes for duggor of variant diagram_dugga have the checkboxes for both UML and ER present and the duggor that are not of variant diagram_dugga do not have the checkboxes as intended. Test passed.

PR #11757, opened 2022-04-22
Clean comment

a19xinhu:
Leave a space for comments. Capitalize except program name.
Modified notation symbol. (/* or // )

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -24,8 +24,8 @@ let width = screen.width;
 var menuState = {
   idCounter: 0,
   /* Used to give elements unique ids. This might? brake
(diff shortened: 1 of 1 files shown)


a20andlj:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Comment: 
For the next time. Do not remove comment signs like in this example (pictures) because the first part is code and the second part is actually a description. If someone should uncomment the code then the compiler will think that the description is also code.

a20andlj:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Comment: 
For the next time. Do not remove comment signs like in this example (pictures) because the first part is code and the second part is actually a description. If someone should uncomment the code then the compiler will think that the description is also code.

Issue #11490, opened 2022-04-10
Copying lines


a20willa:
Perhaps copying a line would be a good feature, regarding codeviewer? The line numbers on the side could for example be used as a "copy" toggle for that specific line. 

I do believe this would not be too hard to implement, as the lines currently are divided into divs anyways. We could simply take it's contents and copy it. This is kinda similar to the way github does it:

![image](https://user-images.githubusercontent.com/81626616/162617557-288c1940-36dd-4f15-8226-06d3a5bbf45f.png)

a20sarle:
Shoud it be implemented?

a20sarle:
Customer wants to prevent students from copying large amounts of code. The code lines which are in focus / the most important should be easily copied while all other code should be hard to copy. If students want to copy all code to laborate with there should be a button to copy all code but otherwise LenaSYS should not allow copying of code (except certain chosen lines of code).

a19xinhu:
Find a example which can copy code. Click Copy Button. It should just copy the copy with class "impo".

a20sarle:
If the copy icon is clicked only code with class "impo" is copied, as described above. However, students can still manually select all code to copy all the lines which should be prevented and only allowed using a download button. Keeping issue open.

a19xinhu:
I add code to make the code can not copy when you not login or student. Only can copy the "impo" code by click CopyCode button. But when you login as teacher, you can copy code by Ctrl+C+V.

a19xinhu:
My Code works on my localhost, but it does not wok on the server.

a20willa:
Source: https://web.dev/async-clipboard/

It seems that the clipboard API will not work because the test servers does not use an *encrypted connection (https)* but a *unencrypted one (http)*. Thus, the code I made will not work as it's dependent on the API, and instead the program will use the default `execCommand` function instead.  

However, the *execCommand` has proven to be problematic too, as revealed by testing. It seems that you cannot simply copy a variable in the way I tried. 

``` 
document.execCommand("Copy", false, lines)
```

Thus, I will try to go around this problem by using [a workaround](https://www.sharmaprakash.com.np/javascript/copying-value-from-variable-to-clipboard/#:~:text=In%20javascript%2C%20copying%20a%20value,with%20a%20bit%20of%20workaround.)

PR #11759, opened 2022-04-22
Make copy impossible for student

a19xinhu:
These code make student or not log in can not copy code, only can use Copy Button to copy "impo" code.  While login as teacher, you can copy as usual.

### DIFF
--- a/DuggaSys/codeviewer.js
+++ b/DuggaSys/codeviewer.js
@@ -402,7 +402,7 @@ function returned(data)
 		createBlocks(ranges[boxid], boxid);
 	}
 
-	//hides maximize button if not supported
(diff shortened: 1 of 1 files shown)


b20fanfa:
Code refactoring - No
Code standard - ok
Meet issue goal - No

Comments: 
1. When you are logging in as a teacher you should always be able to copy ALL code, no matter if you press copy icon or select and copy. 
2. When logged as a student, pressing copy icon, only the FIRST important row was copied, NOT ALL. 
3. When doing an issue, stick to it and do not clean code/refactoring. That is beyond the issue and does not need to be done.

b20fanfa:
Reference to #11490.

Issue #11737, opened 2022-04-21
Diagram module - Inheritance for UML elements


a20gabpa:
Related to #11248
Add the ability to represent inheritance in UML diagrams. Inheritance is represented using a non-filled or filled triangle. Filled is overlapping inheritance and non-filled is disjoint inheritance. 

Something similiar to the picture:

![Skärmbild 2022-04-21 094652](https://user-images.githubusercontent.com/102604680/164405690-9794c8e8-c908-4124-92a4-dd9b1e0ba4d4.png)

a20gabpa:
The added inheritance icon, can be either be filled or non-filled:

![image](https://user-images.githubusercontent.com/102604680/164453189-e1270efc-492e-46c6-a14f-f43462e417e2.png)

The type of inheritance can be changed in the options panel, at the moment 'BG color', and 'name'  is not used and might be hidden in the updated options panel:

![image](https://user-images.githubusercontent.com/102604680/164453242-33d79864-445c-4222-a926-1102c877a3c0.png)

Using example with the inheritance icon together with UML entities:

![image](https://user-images.githubusercontent.com/102604680/164460817-bad3bd4b-b1b1-4bcf-be52-72fba5eb82a5.png)

There is however a small problem. Since the new element is smaller compared to the other, the line's connection seems to move around when zooming. Could be something that the line offset should be multiplied with the zoomfactor. At 1x zoom, it looks good.

![image](https://user-images.githubusercontent.com/102604680/164460939-432dac6b-d89a-4982-9a09-b1af071f5ab9.png)

When zooming in, the line moves out of the box. 

![image](https://user-images.githubusercontent.com/102604680/164460970-8f4598c1-a39d-4566-92a7-a457472a75cb.png)

When zooming out, the lines length increases and it goes through the icon and reaches the other side:

![image](https://user-images.githubusercontent.com/102604680/164461050-c4158de0-f5de-41dc-969d-cc6baf601e7a.png)

Should this be a seperate issue? It is somewhat related to issue #11183

PR #11763, opened 2022-04-22
Implemented inheritance element for UML

a20gabpa:
Related to issue #11737 
I have implemented an element for inheritance in UML. The icon for placing the element is found in the relation menu:

image

Things that could be tested:
- Create, move and delete inheritance element (also test with "x" in corner of selection)
- Using box-selection to mark elements
- Move with "Snap to Grid" enabled
- Connect with lines to different elements
- Check if possible to connect two inheritance elements together
- Verify that it is not possible to adjust the size of the element (no blue boxes)
- Change the type of inheritance (found in the options panel, in a dropdown)
- Check that is not possible to change the color of the background or stroke on the element.

Issue that is visual: when connected with lines, the lines connected on the left and right side, moves when zooming in and out. Might be a problem with scaling line offset when zooming out.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -777,6 +777,7 @@ const elementTypes = {
     ERAttr: 2,
     Ghost: 3,
     UMLEntity: 4,       //<-- UML functionality
+    UMLInheritance: 5,     //<-- UML functionality
 };
 
 /**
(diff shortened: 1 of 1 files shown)


a20gabpa:
How did you change the stroke color? I cannot replicate the problem on my machine?

PR #11765, opened 2022-04-22
Removed margin and padding right to center the list

a20caros:


a20behta (approved):
Removed margins in discussion.

Issue #11762, opened 2022-04-22
Poop emoji in Diagram Dugga


a20behta:
When opening the yellow bar in th ediagram dugga there seems to be a poop emoji which replaces the arrow to close it
![image](https://user-images.githubusercontent.com/102468604/164674795-2024a882-9b94-43f0-997b-d8773aa5314e.png)

it only appears when the yellow bar is open but not when closed. dissapears quickly when closing again. 

https://user-images.githubusercontent.com/102468604/164676049-fa733501-d75e-4229-bce8-61a0d9eb58f5.mp4

video shows when it appears and dissapears

a20behta:
merged into weekly branch. closing issue

PR #11770, opened 2022-04-22
Removed emoji from Diagram Dugga #11762

b20hanwa:
Removed the emoji in Diagram Dugga on optmarker in both diagram.php and diagram.js.


a20behta (approved):
removed reference to picture, quick and easy fix

Issue #11666, opened 2022-04-19
Create dynamic loading, of initial diagram, in digram.js


a20micro:
As part of issue #11401, students should be able to see an assigned diagram on load if the dugga is of type "diagram dugga". Currently, the initial diagram is hard-coded in diagram.js

![CurrentData](https://user-images.githubusercontent.com/95532115/163965455-69b22a5c-86c8-4665-abec-7ba9ac64274c.png)

To create a dynamic initial diagram, the diagram.js file should have conditional loading of demoData and only if there's no assigned .json diagram file. To comply with previous issues, this should be done via passing a string as an argument, which would be the .json file's filename/filepath.

a16linas:
A fix has been made for this issue since when the ER/UML feature was implemented, the fetching of diagram json files got some bugs, pull request #11923 can fix this

i19matla:
merged and done

Issue #11664, opened 2022-04-19
Fetching variant data, from database, in showdugga.php and extract the file name.


a20micro:
The initial diagram, when opening "showdugga.php", should be the assigned .json diagram file, selected via the "edit dugga" menu.
Currently the initial digram data is hard-coded in diagram.js
![CurrentData](https://user-images.githubusercontent.com/95532115/163961916-6ac11b1d-6613-4c8f-a2eb-6834abaee282.png)

For students to view different initial diagrams, the first step is extracting the file to be loaded from the variant table. (in example below, diagram(10).json)

![jsonParamsDB](https://user-images.githubusercontent.com/95532115/163963227-daa9e48b-03bf-4765-90cd-adddd5c76ae6.png)

To do this, the data must be fetched within the scope of diagram.js and the text string must be spliced to extract the file name.

a16linas:
ShowDugga.php now outputs the filename and filepath to the attached json diagram. Will add onto this on issue #11666 where we implement the file content into the diagram

a20micro:
Functionality updated as Javascript doesn't allow dynamic file fetching and requires user input to create file objects (safety reasons). Functionality moved to showDugga.php and rather than returning the file name, returns the entire file content as a text string.

a16linas:
We would like this issue removed since we've now merge it into #11666 and will continue working there

PR #11771, opened 2022-04-22
Showdugga.php now outputs the filename and respective filepath for json diagram #11664

a16linas:
Added functionality where showdugga.php now lists the filename and filepath for the selected json diagram in variant editor. You can test this by selecting a json diagram and creating a new variant on a diagram dugga, then view said dugga from sectioned and look at the console output, you should see a json filename and filepath. Will add functionality where we read its content to generate a new diagram in #11666 
bild

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -990,7 +990,7 @@ var settings = {
 // Demo data - read / write from service later on
 var data = [];
 var lines = [];
-var cid = 0;
+var diagramToLoad = "";
 
 // Ghost element is used for placing new elements. DO NOT PLACE GHOST ELEMENTS IN DATA ARRAY UNTILL IT IS PRESSED!
 var ghostElement = null;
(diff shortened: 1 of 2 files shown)


a20edwth:
Followed test instruction. Worked as intended.

a20micro:
Additional functionality has been added. As Javascript doesn't support opening local files without user input (safety reasons, meaning file-path as a string can't be used), approach had to be changed and functionality has been moved to showDugga.php as php allows dynamic fetching of file content. 

For testing, rather than filename and filepath, entire content of .json file will be shown as console output.

a20gabpa:
Test failed. When selecting a global file, "Untitled-1.json" or "testfile.json", I get the following alert message. It seems that the code can not find the file.

image

However, it seems to work if a file is in course local instead. I have created a .json-file "test.json" and when selecting this file instead, it prints out the correct content.

image

image

a16linas:
Merged #11664 to #11666 so we're removing this pull request

Issue #11628, opened 2022-04-14
Contribution, add database information to mockup.


a20behta:
related to #11593 

the hover functionality is added to be able to hover over a commit will start functioning. 
need to import the database information based on the commits to display when hovering occurs. 

take data from database.
filter correct changes based on where hover occurs. 
display specified lines of code.

b20eripe:
We are creating a new array which is filled by code rows selected by the commit reference

a20caros:
<img width="811" alt="image" src="https://user-images.githubusercontent.com/81676918/163380485-5a6a3510-e934-4edf-9927-187c91448218.png">

b20eripe:
This query gives a rowk that is twice as large as it should be, debugging continues

c20adael:
here is a potential debuggung function https://stackoverflow.com/questions/4323411/how-can-i-write-to-the-console-in-php

b20eripe:
![bild](https://user-images.githubusercontent.com/102593527/164018421-db65b18c-f2ce-48e6-a790-73b8be0749ec.png)

for-loop to print out the array with the results from the query

a20caros:
<img width="552" alt="image" src="https://user-images.githubusercontent.com/81676918/164021194-f2064a03-8f2d-42bd-b5f7-dd8738c2ec14.png">
An onclick on the link for a commit

b20eripe:
This can possibly be solved using an new PHP containing the SQL queries and then use ajax to access updated data from the contribution.js.

c20adael:
We are not following the standard by doing this. What can be done is calling contributionservice.php again.

b20jonlu:
So i tried to solve the problem by going the route of storing the data about all the commits for a github user during the 70 day interval when calling the contributionservice.php by adding this code.
~~~
//Commit changes
	$datefrom = $startweek;
	$datefrom=date('Y-m-d', $datefrom);
	$dateto=strtotime("+10 week",strtotime($datefrom));
	$dateto=date('Y-m-d', $dateto);

	$commitchanges=array();
	$query = $log_db->prepare('SELECT cid FROM commitgit WHERE author=:gituser AND thedate>:datefrom AND thedate<:dateto');
	$query->bindParam(':gituser', $gituser);
	$query->bindParam(':datefrom', $datefrom);
	$query->bindParam(':dateto', $dateto );
	if(!$query->execute()) {
		$error=$query->errorInfo();
[...]

PR #11772, opened 2022-04-22
G3 2022 w4 #11628

b20jonlu:
Partial solution for https://github.com/HGustavs/LenaSYS/issues/11628
The basic functionality for showing commit changes when hovering over a commit link is now in place and it is possible to view what lines were changed by a certain commit. More work is still needed in order to implement the feature in a more refined state. 
Image of what the window looks like when hovering over the commit link. 
Screenshot (15)

### DIFF
--- a/DuggaSys/contribution.js
+++ b/DuggaSys/contribution.js
@@ -23,6 +23,8 @@ var firstSelWeek;
 var secondSelWeek;
 var updateShowAct = true;
 
+var commitChangeArray = [];
+
 //sorting for multiple views
 //Restores all views when pressing the All button
 function restoreStatView() {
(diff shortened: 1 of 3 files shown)


a20behta:
current implementation seems to remove filter buttons in header, other views of contribution and pushes the pop up on top of the other contribution information

https://user-images.githubusercontent.com/102468604/164717634-baca7f1b-324b-4f32-a5d5-7e8aca190a42.mp4

video above shows issue clearer

b20jonlu:
@a20behta i don't think it was this branch which caused this issue since the hover element is the one that is inside our weekly branch #merge4. My implementation is responsible for the text inside the window, i only changed the function called by the on mouseover and added overflow protection for the text which shouldn't be the cause. From testing the weekly branch i think this issue might also be there so if you could also test it that would be helpful

b20jonlu:
oh you might have tested the wrong branch
there is a w3 version that was made by eric i am on the w4 version

a20behta:
issue seems to only occure on the webserver and not locally

PR #11776, opened 2022-04-23
G3 2022 w4 #11112 Fixed some more of the load time issue

a19simru:
I have improved the speed at witch the file loads when using larger databases. What I have changed is the way that the for($i=0;$i<70;$i++){} loop in contributionservice.php works.  This loop makes 4 queries that all count the number of rows in that database that match with the query parameters. This would mean that we have 70 iterations of the loop times 4 queries equalling 280 queries total. But the way that the queries are done actually uses two queries per row count. Meaning that we actually have a total of 560 queries. This was because we used to use the syntax fetchAll() after every query as a part of the counting process. fetchAll() is a query in and of itself. I have changed this to need only the expected 280 by replacing fetchAll().  Meaning that we now have 280 queries less than previously. I also needed to change the way that some functions in contribution.js accessed this data. Since it used to be stored in a superfluously large array. Now its stored as a more practical int. 

To test that the site still works after changes just switch to branch G3-2022-W4-#11112 and go to http://localhost/LenaSYS/DuggaSys/contribution.php.

To test that load time is faster use that software that b20jonlu used to measure load difference while using the largest database we have.

### DIFF
--- a/DuggaSys/contribution.js
+++ b/DuggaSys/contribution.js
@@ -152,6 +152,7 @@ function renderBarDiagram(data) {
   var maxDayCount = 0;
   for (var i = 0; i < 7 * numOfWeeks; i++) {
     var day = data['count'][dateString];
+    //console.log(data['count']);
     var commits = parseInt(day["commits"][0][0]);
     var events = parseInt(day["events"][0][0]);
     var comments = parseInt(day["comments"][0][0]);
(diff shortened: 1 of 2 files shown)


a20behta:
Currently looks like the changes increased the load time not decreased it. 

our weekly merge performance using the built in functionality 
image

compared to the branch trying to merge. 
Screenshot 2022-04-25 132526

this was tested on the webserver while logged in as stei/teacher. 

merge request denied as does not solve the issue. 

@a19simru check why less queries cause a longer load time and why the _fetchall()_ function was there to begin with

a19simru:
The longer load time was as we have discussed coincidence. You notice no difference in performance if you test enough times. After research I have found that the fetchColumn() syntax is roughly 10 times more more effective than fetchAll() syntax. But the overall change is unnoticeable. Because the query's from fetchAll() are not nearly as big a time drain as the four remaining queries in the for loop. It is a improvement in both speed and code complexity. But it could defiantly  have been a bigger improvement.

PR #11783, opened 2022-04-25
G3 2022 w4 #11628

b20jonlu:
Fixed the issue in pull request https://github.com/HGustavs/LenaSYS/pull/11772


a20behta (approved):
Looks good no real issues found. 
can look at optimizing contribution.js, seems to be three nested loops and unsure if that is needed. 

Believe there is nothing that can be done but worth checking out.

PR #11785, opened 2022-04-25
G2 2022 w4 #11490 Copy Code

a19xinhu:
Make teacher copy what they want. Make student can only copy first row important code.


b20fanfa:
Reference to #11759

When we tested the first pull request, we failed to see the new changes. We commented on what was wrong. The developer made a new commit. After talking to the developer what would be involved and would not be involved, an additional commitment was made.

The problem is when we test from the "test server" we can not see the changes. But in the developer's environment (local server) it worked. We decided not to approve the pull request.

b20fanfa:
Reference to #11490.

Issue #11688, opened 2022-04-19
Diagram: Should not be possible to draw lines from two entity elements to the same attribute element.


b20thelo:
![image](https://user-images.githubusercontent.com/81772705/164014018-cb534e0e-705e-45a8-b645-ff336d87c966.png)

Error found in ER model: It should not be possible to draw lines from two entity elements to the same attribute element as it goes against the ER rules.

Other errors that go against ER rules should also be eliminated. If a new error is found add a description of it as a new comment and fix it as well.

b20thelo:
This issue is kind of related to issue #11479 (solved with pullrequest #11576).

i19matla:
Found a separate problem related to this, when there is a chain of interconnected attributes. A new issue will be created to solve that problem.

i19matla:
Merged and done

PR #11791, opened 2022-04-25
Attribute lines to entities and/or relations Issue #11688

i19matla:
Attributes connected to a entity or relation are now blocked from being connected to another entity or relation in diagram.js in the functions addLine and onSetup.

During testing: This issue solves blocking connections between attributes to entities or relations, not the chain of connection that can be created according to er rules between attributes.

A new issue will be created to solve rules of connection between attributes, from attribute to attribute and block connections in that chain to a relation or entity.

image


a20gabpa:
Test passed. It works as expected. I have tested the different types of attribrutes and all have the same behavior. @c20emili

PR #11792, opened 2022-04-25
Made default value unselectable

a20joabo:
Made default value unselectable in file drop-downs.


a20muhal:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #11793, opened 2022-04-25
Repaired broken string generation

a17nikjo:
Code was mixing up " and `, leading to variables being included in the string.


a20behta (approved):
looks good.
very basic fix nothing to comment

Issue #11614, opened 2022-04-13
Hide diagram functionality in non-diagram duggor.


a20micro:
As the teachers should be able to import initial diagram for duggor of type diagrams (#11401), a drop-down menu connected to the database was added to the "Edit Dugga" screen. 
![image](https://user-images.githubusercontent.com/81772705/164428600-926b2a92-3f8c-4b62-8566-e5a07a490dd1.png)

This "Add diagram to dugga", "Diagram types allowed", "Error check" segment is supposed to only show up for duggor of type diagram.

i19matla:
Merged and done

PR #11796, opened 2022-04-25
Dropdown boxes are now hidden when selected dugga is not diagram #11614

a16linas:
The dropdown box for selecting json diagrams is now hidden when the selected dugga is not of diagram type as per stated in issue #11614 ,

bild


a20samul:
Tested on three laptops. 
The dropdown box for selecting json diagrams is now hidden when the selected dugga is not of diagram type. Test passed.

PR #11800, opened 2022-04-25
G2 2022 w4 #11512 v2

a20linth:
New SH URL handler. Not complete, still needs handling for changed course/assignment names, changed course/assignment id, deleted course/assignment and prevention of creating multiple SHs for the same URL.


a20andlj:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #11113, opened 2021-05-21
Collect the information when the last time a teacher visited the submission


a16erilj:
We want the information when was the last time a teacher visited a submission. This should then be displayed in resulted.

A small example
![bild](https://user-images.githubusercontent.com/81558891/119113502-72f2c080-ba25-11eb-9955-8cf86076905f.png)

c20adael:
This issue will get split in to smaller parts. One visual, one database part and one JS part.

a20behta:
implemented and works as expected 
placed in weekly merge

PR #11801, opened 2022-04-25
The latest version of the red day list in contributions

a19simru:
I have already explained what this issue is on several occasions and its the same as always. The only change is that we push into a array rather than make a bunch of variables and the push those variables into a array on lines 118 - 182, also i updated the red day list to contain all holidays that can be red days even if they are on a week end as requested.

To test that this works change branch to G3-2022-W4-#11113 and go to http://localhost/LenaSYS/DuggaSys/contribution.php where you will se a new column in the contribution table.


a20behta (approved):
Looks good. 
easy to read. 
shame the red days have to be placed like they are but nothing to do about it

Issue #11799, opened 2022-04-25
Message of the day margin


b20thelo:
![image](https://user-images.githubusercontent.com/81772705/165084154-e6c04f0f-ed2f-4fb4-83b1-20dcac514a36.png)
Remove white space over message of the day.

PR #11802, opened 2022-04-25
Removed the top margin from the message of the day

a20samul:
Removed the top margin from the message of the day so no more white space is visible above the message of the day banner. #11799 

beforeafter


a20jacmi:
Looks good.

Issue #7658, opened 2019-05-28
Suggestion: Make it possible to check off assignment criterias


a17alian:
It would be nice to have checkboxes next to the criterias inorder to be able to check them off. This makes it possible for the student to keep track of which criterias that have been met eg G, VG etc. 

This screenshot below is from the real dugga. The dots here could instead be checkboxes that the students could check off when they have tried to incorporate the criteria.
![image](https://user-images.githubusercontent.com/37792256/58486731-b41da600-8166-11e9-863a-d604375c8fee.png)
This would make it possible for the teachers to see which grade the student has tried to go for and the student can make sure they have met all the criterias they wanted.

This could maybe be achieved using the markdown in dugga.

a17pioja:
Suggestion for how to make the list item change style using jquery:
```
if($(ID OF THE CHECKBOX).prop("checked") == true){
		$( "#ID OF THE LIST ITEM" ).css( "text-decoration", "line-through" );
	}
	else if($(this).prop("checked") == false){
		$( "#ID OF THE LIST ITEM" ).css( "text-decoration", " none " );
	}
```

a17pioja:
Design suggestion: 
When creating an assignment the super user has an add list button that adds a new list
The list can be extended with +1 list item when pressing the "+" button
The "+" character under the last list item adds 1 list item to the list
The "-" character next to each item removes the list item that it's next to

A checked item gets line-through styling

**SUPER USER VIEW**
![sketch 1](https://user-images.githubusercontent.com/49141519/78773064-f4f36780-7992-11ea-874f-46a2161d6e4b.png)
**STUDENT VIEW**
![sketch 2](https://user-images.githubusercontent.com/49141519/78773072-f6bd2b00-7992-11ea-935e-77b780adc25b.png)

b18veran:
@HGustavs & @a97marbr

a19xinhu:
I started trying to add code in showdugga.php, but I couldn't verify the effect of the code I added in localhost lenasys.

b20idawa:
My idea of a solution is to implement code so that LenaSYS can convert the checkboxes in the markdown file to html tags. Checkboxes must thus be supported on LenaSYS and must therefore be implemented. I think this need to implements in showdoc.php file

b20hanwa:
Closed with pull request #11809

PR #11809, opened 2022-04-25
Make it possible to check off assignment criterias #7658

a20novra:
I changed three file: 
First one is group-assignment.html which added example html tags and content. 
The second file is dugga.css, which added storke class with som css in that. 
The third one is group-assignment.js that added a function to add and remove a class name. 

You can test it by going to demo course > group assignment. Try to check a criteria, and see the results.


a20behta:
@a20novra noticed issue with solution and will fix it before making a new one

Issue #11630, opened 2022-04-14
Add the functionality so that the user can use the new buttons on contribution.php


c20adael:
the issue related #11592

b20eripe:
Related issue #11736 is solved, used for testing of this issue

b20eripe:
![bild](https://user-images.githubusercontent.com/102593527/164468757-202938c8-0f95-48c9-b4b4-3c43c0ce2f57.png)

Dynamic indexing of different course year and all db files in those directories. 1 array with all relevant years and an 2D-array containing strings of all databases with their year.

These 2 will be used as data in two selections of which course to view.

b20eripe:
![bild](https://user-images.githubusercontent.com/102593527/164705758-ee801694-4b53-4742-919b-dd738caf47a8.png)

Added dynamic loading of choices of year for the course selection

b20eripe:
![bild](https://user-images.githubusercontent.com/102593527/164707252-75a476a6-582b-401a-950f-15624835bca7.png)
Improved the indexing of years to only include the year and not the filepath

a20behta:
merged into weekly branch. closing issue

PR #11824, opened 2022-04-26
G3 2022 w4 #11630

b20eripe:
Added all required functionality


b20eripe:
Test folders is possible to get from me via USB memory. Important that all files are put in directory 'contributionDBs' that should be placed directly in htdocs. 'contributionDBs' should contain folders named by the year and those folders should contain all db files you want to use. It loads dynamically as long as the path names are correct, no specific requirements for the db-files.

b20eripe:
bild

bild

c20adael (approved):
Looks like it works. It has been teste locally.

Issue #11728, opened 2022-04-20
Diagram: Remove or fix?


b20thelo:
![image](https://user-images.githubusercontent.com/81772705/164246705-353f8cbe-8dbb-45b8-b34e-1029faa750d0.png)

Remove the submission section or fix so that the minusbutton is presented in a better way.

HGustavs:
Low priority.

Do not remove functioinality but feel free to make this prettier.

But not a priority since this is not meant to be used on mobile platforms anyhow.

PR #11828, opened 2022-04-26
Added code so that the button does not flex-wrap and so that the window is appropriate for smaller screens.

a20samul:
Added code so that the button does not flex-wrap and so that the row doesn't overflow. I also made the "Answer" box smaller alongside margin corrections on correct elements so that the elements would fit properly on smaller screens.

fix

Compare with #11728.


a20samul:
fix2

Padding was off if "instruction file" and "Search in table" is compared. fixed the margin.

PR #11829, opened 2022-04-26
added alt tags too images

a17nikjo:
Lighthouse no longer complains about missing alt tags on said page.


a20behta (approved):
alt tags added to images/icons required

looks good and minimal changes

PR #11830, opened 2022-04-26
Added labels to input elements missing them

b20jonlu:
Added titles and placeholders to inputs missing them. Used lighthouse dev tool in order to audit webpages and fixed forms that were missing labels. For input types checkbox and dates i used titles so when the user hovers over the input they get a description of it, for input types text i used placeholder to display text if the input is empty. For resulted.php it was possible to use the label element since they were already in place but in other places it would have been difficult to do so without altering the look of the webpage significantly.

There is most certainly inputs still missing labels on some webpages. I did not fix forms on webpages such as the showdugga.php pages. I also noticed other issues in the lighthouse tool but i did not alter them since they are not a part of this issue. Since i did all my testing logged in as Toddler it is possible that some elements are only visable to students were not found.

The webpages i tested extensively:

- sectioned.php
- fileed.php
- resulted.php
- duggaed.php
- accessed.php
- codeviewer.php
- courseed.php

### DIFF
--- a/DuggaSys/codeviewer.php
+++ b/DuggaSys/codeviewer.php
@@ -229,7 +229,7 @@ Testing Link:
 							<td>Kind:</td>
 						</tr>
 						<tr>
-							<td><input class='form-control textinput' type='text' id='boxtitle' value='Title' /></td>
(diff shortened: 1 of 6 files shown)


a20behta (approved):
only added titles at places required. 
used placeholder for situations where the text should not be visible looks good

PR #11831, opened 2022-04-26
Simple design and functionality for relative deadlines

a20kimar:
Added a simple design (subject to change in the future)
Added functionality to take the values of the design and input it into the database
Added functionality so that when the page loads the database values is used in the settings menu for each dugga
Added a column for the table quiz in the database

This is a simple implementation of the functionality to store and fetch relative deadlines. The relative deadlines does nothing as of right now and the design and value of the relative deadlines are subject to change. It is implemented in such a way that the relative deadlines can be stored as we choose to and up until the point where we translate these deadlines into real dates can the design be changed freely without interfering with the functionality of storing them in the database.

To test:
First you need to re-install lenasys OR manually add a column with this below:

`ALTER TABLE quiz ADD relativedeadline varchar(10);`

Next go into any course with a dugga(demo-course for example), login as a teacher and click on the cogwheel next to a dugga.

Change the values of the relative deadline and press save.

Reload the page and press the cogwheel again to check if the relative deadline is properly stored and fetched.

If you want to doublecheck the database then login to your database and run this:

`SELECT relativedeadline FROM quiz;`

If it returns a table showing a column filled with NULL(or the value you picked in the above steps) then it works. If you get an error there's something wrong somewhere.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -181,7 +181,7 @@ function selectItem(lid, entryname, kind, evisible, elink, moment, gradesys, hig
   }
   if(relativedeadline !== undefined) {
     var splitdeadline = relativedeadline.split(":");
-
+    // relativedeadline -> week:weekday:hour:minute
     $("#relativedeadlinehours").html(makeoptions(splitdeadline[2],hourArrOptions,hourArrValue));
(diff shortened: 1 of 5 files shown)


a20behta (approved):
Looks good, 
clean minimal changes. 

unsure if adding to init_db is a good idea however should be fine.

Issue #11813, opened 2022-04-26
Mark a line when a label is clicked.


a20jacmi:
When a line-label is clicked the line should still be selected on mouse-up.

a20jacmi:
This was a pretty easy fix in: "diagram.js" -> "mup()" -> (case) "CLICKED_LABEL". This case was previously empty.

PR #11832, opened 2022-04-26
Select line if label is clicked #11813

b20matad:
If line label is clicked, the entire line is selected on mouse up.


a20samul:
Tested on two laptops. The whole line is selected when the label on top of it is pressed on mouse up. Test passed.

PR #11833, opened 2022-04-26
assignment description with a toggle final one issue#11623

a19meram:
Changed the description style to a toggle instead of a drag line.


a20samul:
Tested on two laptops. 
The description is changed into a toggle version.
Test passed.

PR #11834, opened 2022-04-26
Gather dugga instructions from .md file#11773

a20vicad:
There are 3 "testfiles" that make it possible for teachers to create duggas that get their instructions from a markdown file. This is in preperation for it to be possible to edit their instructions.

These three files are the "temporary testfiles" (they're under templates)
image

When creating a new dugga using this "model" you open the dugga editor and add a new dugga by pressing the yellow "+"-button and under "template" you choose "duggaTest". Then under the "start page" of the course and add a new test. Under "link" you choose the dugga you just made and press "save". That is all, thank you.


c20adael (approved):
This works as it is supposed to. The dugga in question does load its instructions from a md file.

Issue #11229, opened 2022-04-01
Accessing links to external services such as youtube or github not working from canvas-integrated course page


a97marbr:
Accessing links to external services such as youtube or github not working from canvas-integrated course page.

Proposed solution: Add an option that allows links to be opened as new tab/page. If the option to open link as a new tab/page is used the link should be presented in a different way than other links in LenaSYS.

Provide mockup example on how the design differs between:
- a regular link
- a link that opens in a new tab

a20samul:
Added pdf file with suggestions on displaying self referencing- and external links.
[Mockup links.pdf](https://github.com/HGustavs/LenaSYS/files/8417054/Mockup.links.pdf)

b20thelo:
Feedback from Henrik.

Henrik wants Youtube and Github links to open up in new tabs by default, so when a link is created it recognizes that the link includes those domains. He also wants a button further to the right side for all links that open in a new tab.

a20chrka:
As mentioned, this issue has been resolved in issue #11836. Therefore this issue is now closed.

PR #11836, opened 2022-04-26
Github and Youtube links now open in new tab and all objects have new tab button #11229

a16linas:
All duggor now have a button that will open the link in a new tab but Github and Youtube links will always open in new tab if you directly clicked on the dugga row. Also changed the Show Canvas Link button to no longer just show "This is a test link" but now it displays the link for the dugga.


a20samul:
Tested on two laptops. 
If a YouTube or GitHub link is created, it is automatically opened in a new tab.
All duggor are opened in the same window unless the link-button is pressed then they also have the option to open in a new tab. Functionality works as intended. Test passed.

PR #11841, opened 2022-04-26
Made the sidebar only show for superusers

b20jonlu:
The sidebar for accepting users is no longer made in the contribution.php file. Instead the sidebar is made in contribution.js. Now when a guest visits contribution.php the sidebar will not be shown. After a user has loged in if they are a superuser the sidebar will be made and shown.

The sidebar still needs to exists in contribution.php but it will just be an empty hidden div untill the function in contribution.js is called. 
Contribution.php login
noside

Contribution.php as Toddler
sideyo


a20behta (approved):
Looks good. easy to follow and understand changes made

PR #11846, opened 2022-04-27
G3 2022 w4 #7658

a20novra:
I changed three file: 
First one is group-assignment.html which added example html tags and content. 
The second file is dugga.css, which added storke class with som css in that. 
The third one is group-assignment.js that added a function to add and remove a class name. 

You can test it by going to demo course > new assignment. Try to check a criteria, and see the results.


c20adael (approved):
This works but only on this dugga with this prewritten description. in the future we have to make this global if a text is tagged with a certain id or class.

Issue #11795, opened 2022-04-25
Connections in a chain between attributes


i19matla:
If attributes are connected in a chain, the attributes not connected to a antity or relation is able to connect to a entity or relation. See picture.

![image](https://user-images.githubusercontent.com/81630339/165067093-0bc6cdc5-33ff-4d92-ab03-5e221c08fa75.png)

i19matla:
Continuation from issue #11688

i19matla:
merged and done

PR #11849, opened 2022-04-27
G1 2022 w4 #11795

i19matla:
Issue #11795 Rules according to ER diagram and element connections is implemented.
Test connection between attributes and attributes, attributes to entities and attributes to relations. Test so blockings works for these attributes when they are connected.


a20gabpa:
Test passed. Three users have tried with different attributes and it was not possible to connect a attribute directly to an entity if it was already connected to an attribute.

Issue #11822, opened 2022-04-26
Redirect camera to center


b20thelo:
Redirect camera to center will instead be named Reset view. When clicked all elements should be viewed. 

![image](https://user-images.githubusercontent.com/81772705/165255041-c49a7bd6-9509-4795-aa8b-0a775119a7bf.png)

a20edwth:
Changed redirect camera to center so the zoomfactor is set appropriate to the size of the diagram.

Instructions for testing:
1) Click "Demo-Course"
2) Click "Diagram Dugga"
3) Change the size of the diagram and click Redirect camera to center.

i19matla:
Because of the label close, i close this issue now. Can´t see it is merged.

PR #11851, opened 2022-04-27
Redirect camera to center changed

a20edwth:
Changed redirect camera to center so the zoomfactor is set appropriate to the size of the diagram. #11822

Instructions for testing:
1) Click "Demo-Course"
2) Click "Diagram Dugga"
3) Change the size of the diagram and click Redirect camera to center.


b20thelo:
Tested on two different computers and the functionality works as it should now. Clicked on different things to see if we could make the function break and it still worked as it should.

PR #11852, opened 2022-04-27
Fixed onclick function in contribution.php

b20jonlu:
Fixed the onclick function for the `rankTable` and the `ghContribTable`. The previous version used the same `renderSortOptions` as `myTable` which meant that the function was trying to sort on the wrong table. Also had to add the `compare` function which is called to sort the table by `sortableTable.js` since it was missing from `contribution.js`. The compare works for sorting ints and strings but not for arrays so i added the functionality of handling those arrays to the compare function.

Sorting `rankTable` by Kind 
sortKinddown

Sorting `rankTable` by Number 
sortNumDown

Sorting `ghContribTable` by GitHub Contributions 
ghContribTable

### DIFF
--- a/DuggaSys/contribution.js
+++ b/DuggaSys/contribution.js
@@ -961,6 +961,78 @@ function renderAllRankSortOptions(col, status, colname) {
   return str;
 }
 
+function renderRankSortOptions(col, status, colname) {
+  str = "";
+  if (status == -1) {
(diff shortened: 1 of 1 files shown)


c20adael (approved):
It works as intended but with one small issue. in the second table there is a number not sorting correctly.

PR #11853, opened 2022-04-27
Hide course dropdown menu until a year is selected

a20johba:
Creating a multi-level dropdown menu was too complicated and I was unable to find a solution in time, so I made a quick fix instead where the course dropdown is hidden until a year is selected.

To test:
1. Go to http://127.0.0.1/LenaSYS/DuggaSys/contribution.php
2. Verify that there is no visible "Choose Course" button (to the right of "Choose Year")
3. Select a year in the "Choose Year" dropdown menu
4. Verify that the "Choose Course" dropdown menu is now visible

### DIFF
--- a/DuggaSys/contribution.js
+++ b/DuggaSys/contribution.js
@@ -1060,13 +1060,8 @@ function courseSelection(elem){
   var dropdown = document.getElementById('courseBtn');
   dropdown.options.length=0;
 
-  // Toggle visibility
-  if(elem.value == "ChooseY") {
-    dropdown.style.visibility = "hidden";
-  }
-  else{
-    dropdown.style.visibility = "visible";
-  }
+  // Set visibility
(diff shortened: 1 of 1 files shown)


CanFight:
Year dropdown did not work for me since i dont have the required files, however if a year is manually added to it and selected the "choose course" button appears as it should.

However the user experience with the implementation is kinda weird since there is nothing that tells me to choose a year if i want to choose a course. Maybe instead of hiding the button for choose course it could be disabled and tell you to pick a year if you click on it while grayed out?

a17nikjo:
Year dropdown did not work for me since i dont have the required files, however if a year is manually added to it and selected the "choose course" button appears as it should.

However the user experience with the implementation is kinda weird since there is nothing that tells me to choose a year if i want to choose a course. Maybe instead of hiding the button for choose course it could be disabled and tell you to pick a year if you click on it while grayed out?

Edit: commented with wrong git account.

PR #11854, opened 2022-04-27
Added gridview for all templates. And added recomendations to how resizing could look. 

a20patpe:
Added a grid to the #div2 box so that elements can be placed within them.
image

The template boxes are now slotted into the grid as follows:
image

Since the old way of resizing and positioning the boxes used percentages to slot/resize the boxes in the #div2 container the functions for resizing, minimizing and maximizing the boxes have to be altered to fit the new way of resizing.

The implementation of Minimizing and maximizing should not cause an issue: simply change the old function but with measurements shown as pixels instead of percentages.

There are alot of functions for resizing boxes that used percentages which have been removed. 
Since resizing boxes was important we have added an initiall draft of how resizing would look like (found in the resizeBoxes funciton) in its place, however it is not currently fully implemented ( this would have to be separated into smaller issues since fully implementing it all at once isn't great from a version controll point of view.


b20sebka:
for testing:
1. open every template
2. open inspekt element and open the grid for said template
3. check that all boxes are within their grid

how to open grid in inspekt:
bild

a20andlj:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes
Comment: Does not work ok on the Firefox browser.

a20sarle:
Will be merged. Should be a new issue regarding FIrefox browser. Works on other browsers.

Issue #11744, opened 2022-04-21
Table definitions in the instructions


a20edwth:
The teacher should have the option to create table definitions for diagram assignments that will be shown in the instructions.

This issue is sub-issue 1 of issue https://github.com/HGustavs/LenaSYS/issues/11233

Example of table definition:
![17674fc4431f8524aa1f748f4f9cef1c](https://user-images.githubusercontent.com/102611046/164436707-f44e999c-5880-4b3a-b11c-4f75869da957.png)

Instruction field:
![68463b078579b16a5473a1fcf9f90aca](https://user-images.githubusercontent.com/102611046/164434857-fa6655f4-452b-4ae2-ab5d-063527e0f715.png)

This feature should be implemented in the **Variants for: Diagram Dugga**:
![a2e0398d095d49e3a86a087d0b835e10](https://user-images.githubusercontent.com/102611046/164439474-c5d589a2-631d-48c9-a632-8102315ffdac.png)

To reach this window from Demo-Course click on the "Show tests" icon:
![ee27b8063901db28d2b8c6c34b33e930](https://user-images.githubusercontent.com/102611046/164439857-fdd7f3f9-e08f-4137-ad49-e2f1387fc713.png)

Then from this window click on "Edit Variants" for Diagram Dugga:
![d3ec8313fe0a5ccd0aa021258847c9ed](https://user-images.githubusercontent.com/102611046/164440341-40b625c6-9366-4d8c-963d-02811e74ac3f.png)

a20chrka:
Main issue: 
- #11233 

Sub-issues:

- #11909 
- #11915 
- #11916

a20chrka:
All current sub-issues for this are completed. However, this issue will remain open as there may still be hidden problems or ER-rules that has been overlooked. 

Also note that this issue should serve as the main-issue for future issues regarding UML table definitions, which is yet to be implemented (and possibly IE as well).

Issue #11233, opened 2022-04-01
Side by side database table diagram / relation definigtion view in diagram editor


HGustavs:
We want concurrent database definition generation for ER and UML diagrams.

if we have an ER diagram, we would want to see table definitions concurrently generated in some kind of a pop up view or side view.

<img width="831" alt="Screenshot 2022-04-01 at 10 16 36" src="https://user-images.githubusercontent.com/6978801/161223868-5ae7ed8c-f129-4779-aa05-0a06292a74fb.png">

The question is what the interface should look like, if this should be a pop-over or a side view or something else entirely.

We should also be able to disable this feature so that for a certain assignment it should NOT be possible to see the relation definitions.

HGustavs:
Errors should also be shown if for example no weak relation is given to a weak entity, the table definition can not be generated for this diagram.

<img width="845" alt="Screenshot 2022-04-01 at 10 13 26" src="https://user-images.githubusercontent.com/6978801/161223330-dd2e5f05-c7f6-4332-8551-89229ac0d366.png">

In this case we have three errors, key missing in person entity, cardinality incomplete for relation and weak entity without defining relationship.

Until these are completed the diagram is not correct

It should be possible to disable checking of the diagram for a dugga.

No definitions should be generated until the whole diagram is correct

HGustavs:
Divide into separate issues and make mock-ups accordingly.

HGustavs:
It would be nice to have an interface that allows us to see the kind of error when hovering the diagram in some way.

b20thelo:
This issue is connected with issue #11248

i19matla:
[issue#11233 mockup.pdf](https://github.com/HGustavs/LenaSYS/files/8416942/issue.11233.mockup.pdf)

Mockup for diskussion

a19meram:
file research needed to find a link between submission code-block and duggaed.php.. Are submission code disabled?

a20edwth:
This issue consists of 3 separate functionalities/features and should therefore be divided into 3 sub-issues one for each feature.
1. Concurrent database definition generation(table definitions)
2. Error checking for relations
3. Interface for checking information regarding relational errors(sub-issue 2)

 **Concurrent database definition generation(relation definitions)**
In this sub-issue relation definitions for the entities should be generated concurrently.

The relation definitions should look something like the image the customer posted:

[...]

PR #11858, opened 2022-04-27
Foundation for dynamic/automatic ER-Table generation

a20chrka:
Issue: #11233
Sub-issue: #11744 

I have added a new button (that is currently using the same icon as another button) in the diagram toolbar. This button toggles the generation of the ER-table. If it is toggled ON, "Properties" in the option panel will display the ER-Table for the current diagram. 

The ER-Table is generated by a single function that returns the entire ER-Table in the form of a string. Therefore, it is safe to call the function wherever it is needed, as long as the code in the function is not edited.

This pull request DOES NOT feature:
* Recognition of keys
* Recognition of foreign keys (through relations)
* Removal of parent attribute (if an attribute has other attributes connected to it)
* Recognition of strong/weak elements

How to test:
1. Press the Options side-panel.
2. Press the "Toggle ER-Table" button.
3. Note that "Properties" in Options is updated with the ER-Table.
4. Try adding new attributes and entities. Connect them randomly.
5. See if the ER-Table is updated. (If not, try clicking on a random element)
Note: Not all functionality is implemented and this is just the foundation for ER-Table generation.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -780,6 +780,18 @@ const elementTypes = {
 };
 
 /**
+ * @description Same as const elementTypes, but uses their names instead of numbers.
+ * @see generateErTableString() For comparing elements with this enum.
+ */
+const elementTypesNames = {
+    EREntity: "EREntity",
+    ERRelation: "ERRelation",
+    ERAttr: "ERAttr",
+    Ghost: "Ghost",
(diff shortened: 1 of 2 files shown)


a20samul:
The new toggle button enables/disables the ER properties in the options panel and updates as more objects are connected to eachother in a manner described in the comment above. Test passed.

PR #11862, opened 2022-04-27
Fixed selection colors issue #11396

b20matad:
Purple and pink color scheme for selected objects. 
Default color changed to white for objects.
Default color set to black for lines.
Updated colors that can be set to different elements.


b19karhj:
Talk with them tomorrow

Issue #11453, opened 2022-04-08
Ruler and A4 template does not match well/ruler does not show a known unit of length.


a20jonda:
![image](https://user-images.githubusercontent.com/102598258/162387291-5c82296d-9d1b-47d5-949b-d3a711ee844e.png)

a20jonda:
set ruler to millimeter

a20jonda:
The A4 measurements will also be changed to match ruler.

b20thelo:
The grid should be aligned to cm marks.

PR #11863, opened 2022-04-27
G1 2022 w4 #11453: Ruler and A4 template does not match well/ruler does not show a known unit of length.

a20jonda:
#11453
Made the ruler measure in CM and made the A4 template match.

What was not completed:
The grid was able to match, but problems occurred in the snap to grid function. So the grid is not in 1CM measurements


a20jonda:
image

a20jonda:
Note:
Realscale was not possible in javascript so the scale is ~10% off. Cannot be solved without forcing the user to measure their monitor.

a20jonda:
image
image

a20jonda:
image

Issue #11745, opened 2022-04-21
Turn on/off error checking


a20edwth:
The teacher should have the option to turn on or off error checking for diagram assignments.

This issue is a sub-issue of issue https://github.com/HGustavs/LenaSYS/issues/11233

This feature should be implemented in the **Variants for: Diagram Dugga**:
![a2e0398d095d49e3a86a087d0b835e10](https://user-images.githubusercontent.com/102611046/164439474-c5d589a2-631d-48c9-a632-8102315ffdac.png)

To reach this window from Demo-Course click on the "Show tests" icon:
![ee27b8063901db28d2b8c6c34b33e930](https://user-images.githubusercontent.com/102611046/164439857-fdd7f3f9-e08f-4137-ad49-e2f1387fc713.png)

Then from this window click on "Edit Variants" for Diagram Dugga:
![d3ec8313fe0a5ccd0aa021258847c9ed](https://user-images.githubusercontent.com/102611046/164440341-40b625c6-9366-4d8c-963d-02811e74ac3f.png)

a20edwth:
Issue #11600 has to be merged before beginning this issue

PR #11864, opened 2022-04-27
fix for issue #11745 Turn on/off error checking

c20emili:
Added a checkbox when creating a variant to change if the error checking is active or not. changes in diagram_dugga.js, duggaed.php and duggaed.js.
test by logging in as teacher and select diagram dugga and selecting the active variant and checking unchecking the checkbox. then as a student checking in diagram dugga and looking if the error checking is active.


b20thelo:
Test passed and conflict removed together with pullrequest creator

Issue #11124, opened 2021-05-21
Investigate what makes the contribution part load so slowly


a19anjst:
For example if issues are shown to be the slow part, you can try to comment out certain code rows to see where it is slow and see what you can achieve. Don't commit any code to the project where you have commented out code that is being used.

a19samke:
![Contribution_Network](https://user-images.githubusercontent.com/62887067/119468117-739d9680-bd46-11eb-85a1-119fb0e5cbfa.jpg)





Why contribution is slow to load, I can imagine because there is too much data to be loaded

a18danli:
Shouldn't be because of the commit-tree, the contrubution was already slow before this was added. So should be because of the amount of data or how the contribution is constructed.

a20behta:
Placed on Hold untill test data for database is avaliable. 
Waiting for Group webserver is avaliable for testing
Current information in pdf

[contributionservice issue.pdf](https://github.com/HGustavs/LenaSYS/files/8409169/contributionservice.issue.pdf)

b20jonlu:
Update report. 
2022-04-06 i managed to get contribution.php to load by replacing the /htdocs/BGHdata_2021_05.db file and managed to test the issue and confirm it still existed. This also allowed me to confirm my suspicion of the root cause of the issue. The performance issues are caused by a high amount of database requests. contributionservice.php has about 340+ sql querries and 280 of them are caused by the for($i = 0; $i < 70; $i++) loop, and 50 by the do while($weekno < 11) loop. 
2022-04-07 i have started changing the requests to be less wasteful to try to improve the performance of contributionservice.php

b20jonlu:
The performance issue is mainly caused by the code execution for the $count variable. Around 3/4 of the time it takes to load contributionservice.php is taken up by $count. My attempts to improve the code have lead me to confirm that the queries are responsible but i haven't found a solution that keeps the json format the same as the original which means the code didn't work with the related modules.
@a20behta 
@c20adael 
[Report contributionservice issue.pdf](https://github.com/HGustavs/LenaSYS/files/8451658/Report.contributionservice.issue.pdf)

b20jonlu:
[...]

PR #11895, opened 2022-04-28
G3 2022 w5 #00001 five times faster load speed on the file contributionservice.php 

a19simru:
@a20behta @c20adael I used ORDER BY in the queries to make it possible to move the queries out of the for loop. The code runs significantly faster as a result.
Screenshot (8)
before changes on the HGustav page.

Screenshot (7)
after changes on the HGustav page.

Please note that I am using a mutated database so load times are longer than they would be with the normal database.

To test that it work's just open my branch on your machine and compare it to the branch for merge 5.


a20behta:
related to issue #11124

PR #11897, opened 2022-04-28
Holy poop emoji is back!

c20emili:
Holy poop emoji is back! "Operation shitstorm is a success!!"- @b19karhj


b19karhj:
Poop emoji, pointer is back and tooltip is back. The war has been prevented.

PR #11900, opened 2022-04-28
Added a button without function in dugga viewer

a20vicad:
Literally just that.

image
A picture of the button that does nothing.


a20vicad (commented):
Now it looks good. Button in the right spot.

PR #11901, opened 2022-04-28
Fixed the problem

a20mtafa:
i found the opacity which where hidden before

For testing:
go to the home page when not logged in


a20willa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

a20muhal:
Do not have inline css
Next time branch name should be according to standard, this is also for the pull request name. 
Also please read the issue description again.

PR #11910, opened 2022-04-28
Variant status are no longer "Undefined"

a16linas:
When viewing variants in localhost they're no longer listed as undefined
bild


a16linas:
Switched base as per request of new group leaders

b20thelo:
Tested on two different computers. Both switch over to the branch and tried using localhost. Undefined was removed and instead, it said enable/disable as it was meant to do.

i19matla (approved):
Tested, works fine.

PR #11911, opened 2022-04-28
Fixed ghost on inheritance and remove stroke color option from options panel

a20gabpa:
Things I have done:

- Updated the UML inheritance to remove the ghost (made it transparent)
- Removed the stroke color option in the options panel for all elements
- Removed the background color options in the options panel for UML inheritance element

Things to test:

- Check if the option to change the color of the stroke is not present in the options panel. Try with one element selected and multiple elements selected.
- Check if a ghost appears when placing the UML inheritance icon.
- With a UML inheritance icon selected, the options for background and stroke color should not appear in the options panel.


b20thelo:
Test passed. Will be included in the presentation branch.

PR #11917, opened 2022-04-28
added so the label follows the line even when far from center for issue #11155

c20emili:
added so the label follows the line even when far from center.
test by creating a label and moving it from the center and then moving the line.

https://user-images.githubusercontent.com/81320323/165773067-8bc5b344-e549-494d-8c78-1deea2487dd3.mp4


c20emili:
https://user-images.githubusercontent.com/81320323/165778887-28c55d06-a45b-45aa-ad42-ed9bba8dcff3.mp4

PR #11918, opened 2022-04-28
Displays a description of the dropdown menus and of relative when you hover over them

a20caros:
…ent relative which prints the title when hovering over the element


b20hanwa:
Works as intended when tested.

Don't know if this should be merged, sence the relative deadline is still being worked on. @a20vicad

PR #11919, opened 2022-04-28
Update diagram.js

a20jonda:
#11453

Updated to include grid in 1CM. Changed the snap to grid function to work with it.
Updated the ruler to be more fancier
image

Note: Pressing 'A' shows the A4 template and pressing 'S' enables the snap to grid.


b19karhj:
Tested and worked as intended.

Issue #11920, opened 2022-04-28
ER and UML checkboxes functionality needs rework


a16linas:
Currently when allowing both ER and UML diagram type in variant editor, the diagram doesn't get rendered until you zoom in/out and then it renders the default diagram template. However, if only one of them is selected, the diagram is rendered as expected. Also, having none of them selected shouldn't be allowed.

![bild](https://user-images.githubusercontent.com/37794783/165778029-7950bd3f-af64-4d90-9093-3f3efba9008d.png)

![bild](https://user-images.githubusercontent.com/37794783/165777710-13f36646-fb29-4b16-80b8-c723799e1e93.png)

c20emili:
The problem is not in the ER/UML types. The problem lies in that the json string changes length and currently the way the file is found is by counting characters. So if anything other than the diagram type change the file does not get found because the name becomes wrong.

PR #11923, opened 2022-04-29
possible fix for the file fetching problem

c20emili:
possible fix for the file fetching problem in issue #11920
changed the string split to fetching from an array with key.This branch is based of the presentation branch. Changed in showdugga.php at rows 73-79 and duggaed.js.
For testing:
- Start by uploading a diagram file if you haven't before.
- to upload a file:
- - log in as toddler
- - go to demo courses
- - click the electrical plug in the nav bar
- - press the yellow plus and select where you want your file (anywhere should work)
- - select your file
- - if you don't have a diagram json file:
- - 1. go to diagram dugga and change the diagram
- - 2. open options 
- - 3. click export
- - click the upload button
- change the diagram file of the currently enabled variant.
- to change variants:
- - log in as toddler 
- - go to demo courses 
- - press the paper icon in the nav bar
- - press the pen on the same row as the diagram dugga
- - if there is a enabled variant 
- - 1. Click the cogwheel on the same row as the variant
- - 2. Change diagram file(if you only have one diagram file you need to change something else or it won't update)
- - 3. Click update
- - if there isn't a enabled variant 
- - 1. Change diagram file(if you only have one diagram file you need to change something else or it won't update)
- - 2. Click create
- - 3. Click the cogwheel on the same row as the variant
- - 4. Click enable
- go to diagram dugga and se that the diagram is the same as the one in your file

If test passed speak with @c20emili when resolving conflicts.

### DIFF
--- a/DuggaSys/duggaed.js
+++ b/DuggaSys/duggaed.js
@@ -476,7 +476,7 @@ function createJSONString(formData) {
 	return JSON.stringify({
 		"type":formData[0].value,
 		"filelink":formData[1].value,
-		"diagram File":$("#file option:selected").text(),
+		"diagram_File":$("#file option:selected").text(),
 		"diagram_type":{ER:document.getElementById("ER").checked,UML:document.getElementById("UML").checked}, //<-- UML functionality
 		"extraparam":$('#extraparam').val(),
 		"submissions":submission_types,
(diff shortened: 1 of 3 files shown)


c20emili:
The checkboxes can still be both deselected because it was not the main issue so that functionality got overlooked.

a20chrka:
Test passed, you can now have both ER and UML "checked-in" and the diagram will still be loaded.

Issue #11922, opened 2022-04-28
Fix so that the A4-functionality doesn’t toggle on when everything in the diagram is selected.


a20samul:
Right now the A4-functionality toggles on when pressing ctrl+A which is the same command for selecting everything in the diagram.
Change the keybind for the A4-functionality.

PR #11927, opened 2022-04-29
G1 2022 w5 #11922

a20samul:
ctrl+a for select all
ctrl+p for a4 template


a20chrka:
Test passed. The A4-grid no longer appears when using short-command "ctrl + A"

PR #11937, opened 2022-04-29
Removed duplicate event.

a20patpe:
There were two instances of the template resizing. this caused the bug.
For testing:
1: enter the codeviewer.
2: enter codeexample "Javascript example 1"
3: The box in the bottom right should no longer have an opacity style applied to it.


a20willa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #11816, opened 2022-04-26
Make ghost icon disabled if no check marks are checked


a20willa:
By checking the check boxes in the dugga viewer, we can hide selected duggas if the ghost icon is pressed. However, the ghost icon can be pressed even when nothing is selected. This makes no sense, and the button should be disabled if this is true. This should also be visually shown in some way, perhaps modifying the icon? 
![image](https://user-images.githubusercontent.com/81626616/165251309-9afc0f0a-ac62-4240-ae03-46c71305806a.png)

b20sebka:
Was not closed after merge.

PR #11959, opened 2022-04-29
Ghost button disabled when no checkboxes is clicked #11816

a20andlj:


a20willa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #11885, opened 2022-04-28
Ruler disappearing when pressing escape


a20edwth:
After deactivating and activating the ruler either by clicking the "Ruler" button in the options panel two times or by clicking "t" on the keyboard two times, when pressing "escape" the ruler will remain active but will disappear from the screen.

Picture shows how the ruler is active but no ruler is on the screen:
![74d05aef2829a5b2327f6b59e0cdd653](https://user-images.githubusercontent.com/102611046/165745464-0aacf8cd-3d9b-492a-aefd-c3a74523aa6a.png)

a20edwth:
What hid the ruler was that for some reason when the "display" css for the ruler got changed, display would always get changed to 'none' when pressing escape. All sorts of "solutions" to solve this problem involving the display property was tried. But in the end, hiding the ruler by changing its position and not touching to display property proved to be the easiest and effective.

a20edwth:
Fixed the problem where after the ruler has been toggled off and on pressing escape hides the ruler while the ruler is still active.

Instructions for testing:

Click "Demo-Course"
Click "Diagram Dugga"
Open "Options" (Yellow bar to the left of the diagram)
Click "Ruler" x2
Press ESC/Escape on the keyboard (The ruler souldn't disappear)

PR #11963, opened 2022-04-29
Fixed escape disabling ruler

a20edwth:
Fixed the problem where after the ruler has been toggled off and on pressing escape hides the ruler while the ruler is still active. #11885 

Instructions for testing:
1) Click "Demo-Course"
2) Click "Diagram Dugga"
3) Open "Options" (Yellow bar to the left of the diagram)
4) Click "Ruler" x2
5) Press ESC/Escape on the keyboard (The ruler souldn't disappear)


a20chrka:
Test failed, for now. Multiple developers has worked on the ruler. We will wait with approving this until all pull requests regarding the ruler are active.

PR #11969, opened 2022-05-02
Add target _blank to github link

b20idawa:
Added target _blank to github link


a20willa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #11970, opened 2022-05-02
Changed the icon ID

b20fanfa:
Change the ID of the icon. The other icons are black from the beginning, while the new one is white. Needed to add ID in whiteTheme.css and blackThem.css to make the button in the opposite color.


a20willa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #11974, opened 2022-05-02
Deleted white background

b20idawa:
Deleted white background when pressing the gear icon


a20willa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #11975, opened 2022-05-02
error check is now hidden in non diagram duggor

a16linas:
Error check is now hidden if the dugga is not of type diagram
bild

bild


c20emili:
Test passed. error check hidden on other duggas

Issue #11973, opened 2022-05-02
The distance the label gets displaced from the line is incorrect


c20emili:
Currently the label gets displaced to far from the line.
![image](https://user-images.githubusercontent.com/81320323/166201778-be87fb21-1fcd-46fa-b130-56b56c5477cb.png)

c20emili:
changed displacement distance.  
![image](https://user-images.githubusercontent.com/81320323/166202577-2a9ef6cf-bfd6-4376-b0d5-ac9143cc7119.png)
#11979

PR #11979, opened 2022-05-02
Fix for issue #11973 The distance the label gets displaced from the line is incorrect

c20emili:
changed displacement distance.  
image


b20thelo:
Checked on multiple computers and approved by Christian. Feedback from Henrik solved.

Issue #11978, opened 2022-05-02
Thickness of the theme toggle button


b20sebka:
change the thickness of the dark/light theme toggle button to match the others in the nav bar.
![bild](https://user-images.githubusercontent.com/102599780/166202954-db021396-09a2-4f2d-98bb-1f3b6e225d6f.png)

b20sebka:
This was not closed when pull request was merged.

Issue #11929, opened 2022-04-29
Home icon


a20muhal:
home icon is too small, make it bigger
![image](https://user-images.githubusercontent.com/102594578/165936409-18085c03-2b7d-4423-b919-32d0ceb09759.png)

a20sarle:
House is bigger but margin does not follow, likely because of uneven values of original svg image element #theme-toggle. Reference #11715. The original img seem to have extra margin / padding. Should be solved before adjusting the margin of the house, reference to ---.
![image](https://user-images.githubusercontent.com/81631818/166200989-2c83fda0-091d-4fe7-87a4-b2420cc20a1b.png)
![image](https://user-images.githubusercontent.com/81631818/166198810-ec29abb0-cacf-4be9-8e87-9eacf45ffe45.png)
![image](https://user-images.githubusercontent.com/81631818/166199250-95824474-e515-4673-9b49-69af715ab9d7.png)

a20sarle:
Continue to work on issue once issue Home icon #11929 and Thickness of the theme toggle button #11978 has been merged to be able to see what margin adjustements are needed.

a20sarle:
There were no other margin adjustments needed after merge.

a20sarle:
> Continue to work on issue once issue Home icon #11929 and Thickness of the theme toggle button #11978 has been merged to be able to see what margin adjustements are needed.

There were no other margin adjustments needed after merge.

a20sarle:
Issue was written over when all the different groups merged their codes. Needs to be implemented again.

PR #11988, opened 2022-05-02
Home icon #11929

a20sarle:
Semi finished. Icon is enlarged but margin should be fixed but to continue #11978 needs to be fixed first.


a20willa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #11986, opened 2022-05-02
Change font color of marked elements to white in diagrams


b20matad:
Black font color is hard to read on selected elements.
<img width="200" alt="image" src="https://user-images.githubusercontent.com/102598238/166205220-d95b2aac-f9ee-4802-a75c-4ac46cf6e806.png">
Change font color of selected elements to white.
<img width="161" alt="image" src="https://user-images.githubusercontent.com/102598238/166205424-f6be73a4-7ff4-4ffd-b282-c1421dcc5a61.png">

b20matad:
Added functionality to diagram.js in method "updateCSSForAllElements()"
<img width="349" alt="image" src="https://user-images.githubusercontent.com/102598238/166209154-f54701cd-793f-49dd-8713-e2034ce13d7e.png">

PR #11992, opened 2022-05-02
Selected objects get white text. Issue #11986.

a20jacmi:
When multiple objects are selected they get a dark violet color. To make the text more readable the font color is changed to white, from black.


a20chrka:
Test passed, looks a lot better and the text is readable when the element is selected.

PR #11994, opened 2022-05-02
Added overflow handler to div2 container.

a20patpe:
Should a box overflow it will now be hidden instead of adding a scrollbar to the div2 container.
For testing:
1: Php example2 should no longer have a scrollbar on the div2 container.


a20willa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12024, opened 2022-05-02
Added an ID in both whiteTheme.css and blackTheme.css files.

b20fanfa:
The "announcementBoxOverlay" ID had no dark mode styling. Therefore, id was added to both whiteTheme.css and blackTheme.css and changed to the correct color.


a20willa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12028, opened 2022-05-02
Cancel Button position

a19xinhu:
change the padding value, make cancel button on the same line with save button.
image

image


a20willa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #12025, opened 2022-05-02
MOTD overlaps with title on frontpage


a20vicad:
![image](https://user-images.githubusercontent.com/102599327/166229043-89d056cf-aeb0-4c08-ad1f-0bf3cf8fdf56.png)
Like this, this is wrong and should be changed.

PR #12030, opened 2022-05-02
MOTD overlaps with title on frontpage #12025

a20novra:
Added a margin botton in curseed.js LINE 417.


a20novra:
Added matgin-top not matgin-bottom .

Issue #11996, opened 2022-05-02
Create a new icon for the new functionality ER-table in diagram dugga


a20samul:
A new _ER-table_ functionality has been created, which is shown in options pane, but lacks an icon. It temporarily has the same icon as the functionality for _Enter replay mode_
- Create a new icon that corresponds to the functionality.
<img width="31" alt="newiconertable" src="https://user-images.githubusercontent.com/81613484/166211628-fad7689c-4b2f-4daa-b5b7-17cdfbaae3e3.png">

PR #12032, opened 2022-05-02
Created an icon for the ER-table functionality #11996.

a20samul:
Created an icon for the ER-table functionality and switched it and incorporated it into the toolbar. 
To test, it should look like:
ERtable
#11996


c20emili:
test passed. functionality works and looks good

PR #12033, opened 2022-05-02
Added style margin-top

a20novra:
Added a margin botton in curseed.js LINE 417.


a20novra:
Added matgin-top not matgin-bottom

a20novra:
@a20vicad

PR #12034, opened 2022-05-02
Added purple color background to the checkboxes.

b20fanfa:
I needed to change the code a bit, there was a syntax error.


a20willa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #11987, opened 2022-05-02
Creating a illegal line prevents some lines to be made


a20edwth:
When creating a line that is illegal (induces a error message) or when deleting certain lines, other lines that should be possible becomes illegal (induces a error message). 

An example of this error occurring is when creating a line from a entity, then deleting that line and then try to create the same line. A error message will appear:

https://user-images.githubusercontent.com/102611046/166205597-6e854a48-c4a6-489c-ad51-d0be05d3a083.mp4

More research to the differenet combinations that causes this problem has to be made.

a20edwth:
When trying to create a line between two entities created by the user there will be a error message, as it should. And when trying to create a line to a relation it will work as intended:

https://user-images.githubusercontent.com/102611046/166208846-b6113d02-2723-4a2e-9f0e-fc9cb594b542.mp4


When trying to create a line between two entities one created by the user and the other by the diagram, there will be a error message, as it should. And when trying to create a line to a relation it will not work as intended (an error message appears):

https://user-images.githubusercontent.com/102611046/166208999-8e999a77-a3f2-436e-a2ed-77b500cfcf60.mp4

a20edwth:
When creating a line from an attribute to a relation and deleting that line, it will not be possible to create a line to any relation, there being an error message:

https://user-images.githubusercontent.com/102611046/166213571-81de2e82-a75f-4f66-9785-43c16b8c7670.mp4


When creating a line from a relation to an attribute and deleting that line, it will not hinder any line to that attribute.

https://user-images.githubusercontent.com/102611046/166213790-884aa3ab-4533-46a0-bf9e-3e81cad480de.mp4

a20edwth:
The biggest factor to this error was that the array allAttrToEntityRelations didn't get cleard correctly when deleting a line. As well as some other factors.

Instructions for testing:
1) Click "Demo-Course"
2) Click "Diagram Dugga"
3) Test by creating lines as shown in the videos located in the issue thread.

PR #12035, opened 2022-05-02
Fixed incorrect error messages

a20edwth:
The biggest factor to this error was that the array allAttrToEntityRelations didn't get cleard correctly when deleting a line. As well as some other factors. #11987

Instructions for testing:
1) Click "Demo-Course"
2) Click "Diagram Dugga"
3) Test by creating lines as shown in the videos located in the issue thread.


c20emili:
test passed but some bugs need improvement. composite attributes on relations work depending on the direction of the line drawn.

PR #12037, opened 2022-05-02
12016 the save button is wrong placed

a20mtafa:
For testing:
1. log in as name: stei password: password
2. click on the settings in one of the courses 
3. the save buttons position is now changed


a20willa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #12014, opened 2022-05-02
Changing template will not load when pressing save


b20idawa:
When you change the template, the template will not load until you refresh the page. A better solution is to load the new template immediately when you press save

https://user-images.githubusercontent.com/81676911/166215892-6af11c53-6b09-436e-8859-3ffcd4380984.mov

a20linth:
When recreating this issue, a fatal error occurs when trying to change template.
![image](https://user-images.githubusercontent.com/102609651/166235138-af9630d9-c975-4d45-aa78-c93f2f3f4a37.png)

a20linth:
Testing:
1) Login as a teacher account (username: stei password: password)
2) Enter Webbprogrammering course
3) Open code dugga JavaScript Example 1
4) Click on Choose template and pick template 8

Excpected result: Code example should update to newly chosen template

b20sebka:
Was not closed after merge.

PR #12056, opened 2022-05-03
Fixed fatal error #12014

a20linth:
fixed the Issue in codeviewerService.php where an fatal error occurs and was ignored for some reason.


a20willa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12066, opened 2022-05-03
Resizing boxes in template 6.

a20patpe:
Reszing the boxes in template 6 is now possible.
For testing:
1: Create a new code example.
2: set its template to template 6. (the 6th from the left)
3: You should be able to resize the boxes.


a20willa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12067, opened 2022-05-03
added more account request bar click functionality

b20jonlu:
- Fixed the issue of the side bar changing to display:none; when pressing escape by changing the style.display to "inline-block". 
- Made it possible to open and close the side bar using escape. 
- Also made it possible to open account request by clicking the hide-accountRequest-pane instead of only the span.
- And made it so clicking on anything not belonging to accountRequest-pane will close the menu


a20vicad (commented):
Exactly what I wanted, good.

PR #12070, opened 2022-05-03
Changed "New Quote" to "New message"

b20idawa:
Changed "New Quote" to "New message"


a20willa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12071, opened 2022-05-03
Made columns sizes remain consistent when searching. Changed code related to the alternating row colors.

a20johba:
Fixed the issue of columns changing sizes when searching by hiding filtered columns rather than removing them. Fixed a subsequent issue caused by this change that alternates the row colors with javascript code directly rather than using code in style.css. This appears to work well for duggaed.php, fileed.php, and resulted.php, but further testing may be necessary to prevent unintended bugs.

To test:
1. Go to any table that can be filtered by typing in the search bar (ex. duggaed.php)
2. Enter a search term. The column sizes should remain the same when the table updates. Note that the table width itself may change size if a scroll bar appears/disappears in the browser window as a result of the change in size of the table. This could be a separate issue but may be unavoidable.
3. Look at the row colors. The colors should alternate between grey and white.


a20vicad:
image
All entries are white in this branch. Can you check this out?

PR #12073, opened 2022-05-03
Make searching icon bigger

a19xinhu:
change the height for search button to make it looks bigger.
1 login as a teacher 
2 navbar "show test" button
3 compare the "searching" size (from 18px to 25px)


a20willa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12077, opened 2022-05-03
Added some code that goes trough the array and checks the correct boxes

c20adael:
This does solve the problem where the sub-boxes did not get checked when pressing the section box. 
To test this you need to a course and try to mark the section box. 

https://user-images.githubusercontent.com/80856978/166443507-f141009c-b688-4492-986a-0b76730711da.mp4


a20vicad (commented):
Did testing, it works well.

Issue #12011, opened 2022-05-02
Can not open examples as designer


b20sebka:
When logged in as designer you can not open any examples.

https://user-images.githubusercontent.com/102599780/166215287-db89e26a-344d-41f0-bc33-5906316d6210.mp4

a20linth:
Testing:

1) Login as a designer account (username: mestr password: password)
2) Enter Webbprogrammering course
3) Open any of the listed code dugga

Excpected result: The code example content should be visible and the user should have write access

b20sebka:
Was not closed after merge.

PR #12079, opened 2022-05-03
Fixed navheader access #12011

a20linth:
The navigation header recognizes write access. And fileedButton is only updated if it exists.


a20willa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #12039, opened 2022-05-02
Grid dashes on the background "boxes" on the grid change when zooming 


a20samul:
The grid dashes on the background boxes change depending on zoom level leading to a strange look when zoomed all the way out (compare images below). 
- Fix so that the the dashes are consistent.

<img width="756" alt="grid1" src="https://user-images.githubusercontent.com/81613484/166247598-87dec640-529d-4800-8bb3-78aebbe828bb.png">
<img width="748" alt="grid2" src="https://user-images.githubusercontent.com/81613484/166247603-4077c27d-bf42-4d4f-b80d-2c6c16677cb0.png">

<img width="740" alt="grid3" src="https://user-images.githubusercontent.com/81613484/166248594-20f9bee8-cda3-4912-a894-850020947c2b.png">

a20chrka:
Test failed. This is what it looks like on my machine:
Old version:
![image](https://user-images.githubusercontent.com/102599740/166469400-18ca78c3-0972-476e-aed9-2da7e093e0d2.png)

New version:
![image](https://user-images.githubusercontent.com/102599740/166469760-583bc8a8-a35f-491a-83ff-c2abb6b4827d.png)

Also, note that different zoom levels in the browser affect how the lines are drawn:
![image](https://user-images.githubusercontent.com/102599740/166470341-5e4a986f-571a-4cd1-90f6-719bd9e966d2.png)

All of them are different, but none of them are aligned properly

PR #12080, opened 2022-05-03
fixed the dashes to a consistent style issue #12039

a19meram:
The background is changed on the zoom out. Se figure 1 and 2

- figure 1 old style
image

- figure 2 new style
image


a20chrka:
Test failed. This is what it looks like on my machine:
Old version:
image

New version:
image

Also, note that different zoom levels in the browser affect how the lines are drawn:
image

All of them are different, but none of them are aligned properly

Issue #11960, opened 2022-04-29
Undo function 


b20sebka:
An undo function when choosing the examples would be good if someone were to accidentaly delete an example and wanted it back.

a20andlj:
Testing:
1. Login as a teacher.
2. Delete an example
3. Press Yes - a recycle icon should appear besides the + and the ghost icons.
4. Press the recycle icon
5. The example should reappear

Testing 2:
1. Delete another example
2. Press yes
3. Wait for 60 seconds - then the recycle icon should disappear and the example should still be deleted.

b20sebka:
Was not closed after merge.

PR #12081, opened 2022-05-03
Undo button if accidental deletion of an example #11960

a20andlj:
I implemented a button so if you accidentally delete an example you get 60 seconds to regret the action.


a20willa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12082, opened 2022-05-03
New feature Relative deadlines

a20kimar:
## Relative Deadlines

The relative deadline calculates a deadline from the course start up until the amount of time you pick. So if the course starts at 2020-05-01 and you pick 2 weeks as the relative deadline then this dugga will have its deadline set to 2020-05-01 plus 14 days. This is, in general, what relative deadlines do.

The relative deadlines are chosen through the menu that appears when you press the cogwheel(logged in as a teacher) on a dugga or quiz. You'll then have the option to choose a time, a amount, and a type to store in the database for this particular dugga.

- Time = Hour and minute of the day
- Amount = Amount of the type
- Type = Days, Weeks or Months

Currently, what happens behind the scenes is that you'll have the option to choose absolute deadlines(as it was before) if you want by ticking a checkbox next to the absolute deadline date picker as the picture below shows. Doing this, this will then override the relative deadline. The relative deadline will still be stored and saved but the absolute deadline will be used as it was prior to the implementation of relative deadlines.

image

When the site loads, a check will be made whether the relative deadline and the absolute deadline are the same or not. If they are the same, the websites assume that the relative should be used and shows a unchecked absolute deadline. If these two dates differ, then it will show as the above picture show. When a teacher then clicks save, the website will check whether the absolute deadline checkbox is checked, if it isn't then it will override the deadline datepicker date with the previously calculated relative deadline and then store that data in the database. This way, we can assure ourselves with the fact that relative deadlines will never cause any issue with how the deadlines are being used in any other part of the website. All we're essentially doing is changing what deadline should be stored prior to storing it. This also gives the flexibility for relative deadlines to be stored in a json format as a string and retrieved and read after the fetching has happened, so we could potentially store many other variables to do with relative deadlines that has not to do with a date. 

Keep in mind that currently, all relative deadlines will be NULL once you re-install the database and I choose to automatically pick absolute deadline if the relative deadline is NULL and therefore when you click on the cogwheel for the first time for each dugga, it will always pick absolute deadline. Reasoning behind this is that I didn't want to disrupt any ongoing deadlines by overriding the current deadline. Once you uncheck the absolute deadline checkbox and save it, it should pick the relative deadline over the absolute one.

If we're ever storing deadlines outside of the sectioned.js file then maybe there would be some issues but as far as I've seen, there is no ajax requests besides the one I've been working with that handles deadlines and that one only points to sectionedservice.php which is the file that handles all SQL requests to the database.

Just to emphasize:
These changes *do not* interfere with the current logic with comparing absolute deadline validation for it being within the course dates or not. 

They *do not* interfere with how the deadlines are being stored in the database.

They *do not* interfere with how the deadline are being handled subsequent any deadline being fetched from the database.

### Tests
In order to test this, I would simply play around with deadlines and see which one is being picked after saving a dugga.
Here's a few tests that can be made:

**Remember to re-install the DB by running install.php if you didn't do that since last merge.**

- Open any dugga through the cogwheel menu and change the relative deadline. Press save and open again(can be done without refreshing the site) and check whether your change persisted or not.
- Open any dugga through the cogwheel menu and change both relative and absolute deadline by checking the absolute checkbox. Then proceed to save and open the same dugga again to see whether the absolute deadline checkbox is checked or not and if the dates are the dates you previously picked.
- Open any dugga through the cogwheel menu and do the same as the previous step but this time before saving uncheck the absolute deadline checkbox and then check again whether the absolute or relative is being used. The one being used will be the one displayed to the left of the cogwheel menu button. The picture below shows where this is.
image

I'm sure there are many other ways to test this, but this is the simple way or figuring out whether the data is being stored *and* if it's being displayed correctly.

### Known issues
- When copying a course the old absolute deadlines will still be prioritized
- There is no "preview text" yet as requested in the original issue request
- The absolute deadlines checkbox positioning behaves strange on smaller devices

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -141,10 +141,10 @@ function selectItem(lid, entryname, kind, evisible, elink, moment, gradesys, hig
   var hourArrValue=[0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,23];
   var minuteArrOptions=["00","05","10","15","20","25","30","35","40","45","50","55"];
   var minuteArrValue=[0,5,10,15,20,25,30,35,40,45,50,55];
-  var weekArrOptions=["1","2","3","4","5","6","7","8","9","10","11","12"];
-  var weekArrValue=[1,2,3,4,5,6,7,8,9,10,11,12];
-  var weekdayArrOptions=["Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday", "Sunday"];
-  var weekdayArrValue=[1,2,3,4,5,6,7];
+  var amountArrOptions=["1","2","3","4","5","6","7","8","9","10","11","12","13","14","15","16","17","18","19","20","21","22","23", "24"];
+  var amountArrValue=[1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,23,24];
+  var typeArrOptions=["Days", "Weeks", "Months"];
+  var typeArrValue=[1,2,3];
 
   nameSet = false;
   if (entryname == "undefined") entryname = "New Header";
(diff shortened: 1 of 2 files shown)


b20hanwa:
Can't get the deadlines to work when tested, and the save button seams to have broken. We will have to continue with this in next merge. All code changes in this pull request is saved in screenshots.

Issue #12051, opened 2022-05-03
There is no "Back to top" - button on pages other than main course pages


a20vicad:
It would be usefull to have the same type of "back to top" - button that exists on the main page for courses on other "tall" pages such as the fileed or resulted as well.

PR #12085, opened 2022-05-03
There is no "Back to top" - button on pages other than main course pages #12051

a20novra:
The code below added to resulted, fileed, and duggaed .php/.js


a20vicad (commented):
This is an adequate solution. Good, I like.

Issue #11875, opened 2022-04-28
Including the new testing-page while installing locally only works on windows machines


a18gusjo:
Reference to issue #11402 where an alternative that includes a new testing page is provided while installing LenaSYS locally.
Including the new testing-page only works on windows machines whilst Mac users have problems including this.

![image](https://user-images.githubusercontent.com/102605031/165728107-96e4fb12-bd47-4f98-8e4c-23ca204bec22.png)

a18gusjo:
![image](https://user-images.githubusercontent.com/102605031/166209902-40dc38f7-cc92-4a2b-8962-1c8285077795.png)
This error message appears when navigating to the tesing-course from Mac OS.

PR #12087, opened 2022-05-03
Including the new testing-page while installing locally only works on windows machines #11875

a18gusjo:
The installing script of is LenaSYS now working on Mac while running it from Safari or Firefox. The problem will probably remain while using Chrome though.

For testing: 
Run the install script on Safari, Chrome and Firefox on MacOS and see if this error message will occur when navigating to the tesing course:
https://user-images.githubusercontent.com/102605031/166209902-40dc38f7-cc92-4a2b-8962-1c8285077795.png 

Issue:
https://github.com/HGustavs/LenaSYS/issues/11875


a20willa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12088, opened 2022-05-03
Created a hover function on some parts in the hamburger menu that´s missing the hover function

a20caros:


a20vicad (commented):
Can't see any problems with this. It looks good.

Issue #12064, opened 2022-05-03
double code from last merge in diagram.js


b20matad:
from last merge there are double code in method "mdown()" in diagram.js.

Code 1
<img width="567" alt="mdown1" src="https://user-images.githubusercontent.com/102598238/166427974-2a78e957-cd67-4314-896c-88a488845b5c.png">

Code 2
<img width="908" alt="mdown2" src="https://user-images.githubusercontent.com/102598238/166427978-e78acce5-1715-4422-83ca-82a8a7194336.png">

These two solution needs to be combined in to one.

PR #12094, opened 2022-05-03
Removed and fixed double code in mdown #12064

a20jacmi:
This mainly regarded the functionality around the delete button and the feature to move a selection box.

To test this: Try to use these features in all mouse modes. Create new objects and try on them as well. Press the delete button when it's on top of another object.


a20gabpa:
Test passed. It works well with pointer, box selection, entity, relation and attribute. It does not work as expected with the line tool selected. Also tested with the delete key on the keyboard and the same behaviour appears. Works on all mouse modes except line tool. With the line tool you can not click the delete button of the element or press the delete key.

PR #12096, opened 2022-05-03
11934 dugga instructions editable

a20behta:
Added functions for editing dugga. 

currently only locally and current active html not in session or persistant in server. 

unsure if html should be accessable and editable by user/teacher. 

currently should not affect system


a20behta:
testing:
Logg in as a teacher. go to any dugga and press the new edit instructions button. 
make some changes to the pop up and press save. 

changes should appear and replace the dugga instructions

PR #12122, opened 2022-05-04
Fixed bug caused by weekly merge

b20jonlu:
Fixed the incorrect behavior of the checkboxes. The code was in the wrong place.


a20vicad (commented):
This has been tested and works like it used to. Which is good.

Issue #12119, opened 2022-05-04
Create a new default variant with diagram file and instructions added for diagram dugga 


c20emili:
Currently the default diagram dugga variant does not have any files selected which creates errors. A new default needs to be created and an diagram file added to the install file to be selected in the variant.

PR #12123, opened 2022-05-04
Added new default to the install file for diagram and a diagram file

c20emili:
#12119 
Added new default to the install file for diagram and a diagram file.
Test by re-installing the database and opening diagram dugga.


i19matla:
Requires Lena SYS to be reinstalled.

Issue #12084, opened 2022-05-03
Highlight functionality not working


a20linth:
The highlighting functionality has stopped working when hovering mouse over html tags. It works as intended for squirrely brackets.

https://user-images.githubusercontent.com/102609651/166447574-d54cdbc9-b888-44e6-9662-bc76dafa10de.mp4

https://user-images.githubusercontent.com/102609651/166447623-0b5d8b77-1f0b-4878-a7ee-8cd9bb643e19.mp4

a20linth:
The class attribute ‘normtext’ should be associated with every div container representing a line of code in the codeviewer that is not marked as important. This class is currently missing, appending the class again fixes the highlighting issue.

An error is thrown by the highlight function for void elements because they do not have closing tags, the highlight function should therefore not be appended to void elements because it needs a start and end tag to work properly.

a20linth:
Testing:

1) Enter Webbprogrammering course
2) Open HTML5 Example 1
3) Hover mouse over brackets and HTML tags: 
brackets { } and ( ) - should be highlighted
HTML, HEADER, TITLE, BODY, H1, P and SCRIPT - should be highlighted
LINK - should not be highlighted (void elements do not have ending tags)

PR #12124, opened 2022-05-04
Fixed highlighting html tags #12084

a20linth:
HTML tags are now highlighted when hovering mouse over them. Removed highlightHtml function from void elements to avoid an annoying error message.


a20linth:
Testing:

1) Enter Webbprogrammering course
2) Open HTML5 Example 1
3) Hover mouse over brackets and HTML tags: 
brackets { } and ( ) - should be highlighted
HTML, HEADER, TITLE, BODY, H1, P and SCRIPT - should be highlighted
LINK - should not be highlighted (void elements do not have ending tags)

b20idawa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12128, opened 2022-05-04
fixed the dashes to a consistent style issue #12039

a19meram:


a20chrka:
Test failed. This is what it looks like on my machine:
image

Note that the lines are not aligned with the ruler and that the origo lines are pretty much indistinguishable from the other lines.
I think it has to do with how it is rendered on different machines with different screens.

PR #12129, opened 2022-05-04
Removed the opacity 

a20mtafa:
i found the code where the opacity made the text color brighter and removed it so that the text color is darker and that there isn't any inline css also


b20idawa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12139, opened 2022-05-04
Added margin to whiteTheme

b20idawa:
Added margin to whiteTheme for the content


b20idawa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #12104, opened 2022-05-03
You can't press the "X" on selected elements (when you try to remove it) when placing lines.


a20chrka:
When the linemode is used and you have selected the first element it should be connected to, you cannot press the "X" on the element nor press the delete button on the keyboard.

![image](https://user-images.githubusercontent.com/102599740/166473391-c38bd9d6-d551-4329-91e0-c8362e3e44af.png)

b20matad:
After enabling the possibility to delete objects, in line creation mode, we've decided this is not a good solution; since when you mark an element in line creation mode a ghost line is created from this element. We will instead remove the delete button while in line creation mode.

b20matad:
How our solution looks and behaves.





https://user-images.githubusercontent.com/102598238/166882881-9a54615b-d04a-465f-b24d-952d84d18b57.mov

PR #12160, opened 2022-05-05
Removed delete button from edge creation. Issue #12104

a20jacmi:
To test this you go to edge creation and draw a line between objects. No box with delete button should appear when objects are selected. Switch between other modes to see the difference.


a20chrka:
Test passed. The X no longer appears and it does not seem to break functionality.

Issue #12137, opened 2022-05-04
Display default instructions if no file found for the diagram dugga.


c20emili:
Put a check for no file found when fetching instruction file that gets a default text for diagram dugga.

PR #12164, opened 2022-05-05
Fix for issue #12137 Display default instructions if no file found for the diagram dugga.

c20emili:
#12137
If no file enabled, user logged in as teacher or there is no file a default instruction is shown on the left of the diagram.
Test by having no variants for diagram dugga, disabling all variants and logging in as teacher and looking at diagram dugga.

### DIFF
--- a/DuggaSys/showDugga.php
+++ b/DuggaSys/showDugga.php
@@ -56,7 +56,7 @@
 	#vars for handling fetching of diagram variant file name
 	$variantParams = "UNK";
 	$filePath ="";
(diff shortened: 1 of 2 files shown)


a16linas:
[TEST FAILED]

The default instruction does not appear when opening the diagram dugga as a student, or at least when user is not logged in. Tried this with a freshly installed database with testdata variant and a newly created variant, same result. 
bild

a20chrka:
Test passed, now works as intended.

Issue #12089, opened 2022-05-03
Diagram is not visible until you scroll


a20chrka:
Diagram does not load properly, causing a visible bug if no specific diagram file is used.
![image](https://user-images.githubusercontent.com/102599740/166450572-0adee741-94ea-4fd1-82d7-01dbd99dcec3.png)

After you scroll, the diagram appears, however.

PR #12165, opened 2022-05-05
Fix for issue #12089 Diagram is not visible until you scroll 

c20emili:
#12089
If no file enabled, or there is no file a default diagram is now shown.
Test by having no variants for diagram dugga, disabling all variants and logging in as teacher and looking at diagram dugga.


a20chrka:
Test passed. Default diagram is loaded when no diagram file is chosen.

PR #12167, opened 2022-05-05
Fixed the return button interactive area

a20johba:
...except for 1 pixel around the image. I have no idea why.

To test:
1. Go to the demo course page.
2. Attempt to click the return button outside of the image area. The button should not activate (except within 1 pixel)
3. Click the return button. It should work as intended.


a20vicad (commented):
This looks better. Some different things related to navbar needs to be fixed but that's a later issue.

PR #12169, opened 2022-05-05
duggaTitle argument missing or invalid.

a17nikjo:
DuggaTitle was missing or invalid in the as an argument when calling the funciton displayDuggaStatus(), from the dugga some of the dugga templates in their returnedDugga() function.

This still only works aslong as the dugga is loaded correctly since it requires data from an json file coming from an AJAX thing.

To test open a dugga while not logged in, not in demo course. Dugga Title Should now appear at the top if the dugga was loaded correctly.


a20vicad (commented):
This is good but I will open another issue to see if there is a way to make it work for teachers aswell, since it really ought to.

Issue #12127, opened 2022-05-04
You can't close the options menu unless you select an element first


a20chrka:
You can always open it, however. But for some reason you can't press the options button again until an element has been selected.

a20jacmi:
Now the menu can be closed/opened in any situation. We've also added functionality to close/open the menu by double clicking in the container, while in pointer-mode.

Changes are made in "diagram.js" for the following functions:
toggleOptionsPane()
mdown()

PR #12170, opened 2022-05-05
Fixed option pane issue #12127

b20matad:
Now you can close the options menu even if an element isn't selected. 
Also enabled the ability to double click the canvas in pointer mouse mode to show the option menu.

To test: Try opening and closing the option menu in different mouse modes; with one, many and zero selected elements.
Also open and close the option menu by double clicking the canvas in pointer mouse mode.


a20chrka:
Test Passed, no problems with option panel now.

PR #12171, opened 2022-05-05
added navbar to contribution page

a19simru:
The bar has been added however three new issues are needed because three features don't work. I don't think I am ideal for any of them. Rather they should be given to the people who made the features in question.

1. Dark mode does nothing.
2. Message of the day does nothing.
3. The account request pane covers some of the left navbar and some of the user account button. You can still use the user account button but it looks ugly
Screenshot (9)
.


a20vicad (commented):
Very good. I'll add the relevant issues that arose.

PR #12172, opened 2022-05-05
Fixed the placement of the FAB button

c20adael:
Changed the class to that it can be applied on multiple pages to form a uniform result. 
To test this: 
Visually check the yellow button on all of the following pages: duggaed, accessed, courseed and sectioned.

If any pages that aren't uniform are found please document these.

### DIFF
--- a/DuggaSys/accessed.php
+++ b/DuggaSys/accessed.php
@@ -123,7 +123,7 @@
 
 	<!-- Add User Dialog END -->
 
-	<div class="fixed-action-button" id="fabButtonAcc">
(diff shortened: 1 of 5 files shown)


b20hanwa:
The fab button looks uniform on most pages. Courseed is the exeption, where the fab button is placed lower than on duggaed, accessed and sectioned. @c20adael

b20hanwa (changes requested):
The fab button on courseed.php is placed lower than on other pages.

b20hanwa (approved):
The fab button is now uniform on all pages affected!

PR #12190, opened 2022-05-05
Fixed the navheader issue

b20jonlu:
Fixed the incorrect navheader. now it looks the same as it did before the issue.
How it looked like before the issue:
server_before

How it looks after the fix:
local_now

The inline-block caused the empty space and the changes to the if statement fixed the problem with having two bars


a20vicad (commented):
Now it looks like it's supposed to again

PR #12191, opened 2022-05-05
Fixed the background color in dark mode in file "fileed.php".

b20fanfa:
To resolve this issue, it needed to comment out a code (from the "sortableTable.js" file). Because it converted the color. It prevents us from adding the right dark mode color. A class was created instead called "changeColorInDarkModeTable". Which is on both blackTheme.css and style.css


b20idawa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12193, opened 2022-05-05
Implemented teacher & user login for contribution.

a20sacka:
Fixed so you can log in with the users from the "USER" table and not only "git_users" table

Added so if you exist on the "USER" table but dont have superuser status youre not allowed to login, this can be changed so all users on the "USER" table are allowed to log in with more priviliges than a normal git-student user. This is how it looks if you attempt to log in with a non super user
image

If you now try to log in with a user that exists on the "git_user" table youre allowed to but your stored session variables dont get stored in the same variables as a teacher/super/one who logged in with someone who exists on the "user" table. The rest of the contribution file will need to take this into account for what is shown; Following is an example of where the user who logged in with the "git_user" table gets their session variables stored
image

Currently the check for viewing contribution content is done with 2 separate checks, in the future we might want to combine these into one dedicated check for viewing contribution data
image

----------------------------------------------- Testing ----------------------------------------------
make sure you run install.php so you have the new git_user table

Check if you can create a new git_user
check if you can log in with a user without login privs an example can be teacher2
check if you can log in with a newly created git_user
try to type in wrong password for user/git_user

### DIFF
--- a/DuggaSys/contribution.js
+++ b/DuggaSys/contribution.js
@@ -2221,12 +2221,13 @@ function resetForceLogin()
 
   let userExists_Git = null; // if it exists in the git data
   let userExists_Lenasys = null; // if it exists in the lenasys data
+  let userStatus_Lenasys = null; // if it is a super/teache
  
   function checkIfGitUserExists(username, _callback) // checks if user exists in the git data and or the lenasys data
 {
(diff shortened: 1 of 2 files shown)


a20vicad (commented):
This does do the thing it's supposed to do. Ready for the next step.

Issue #12036, opened 2022-05-02
Error when elements have the same name


a20edwth:
There should be an error when there exists entities with the same name, as well as when there is an attribute connected to an attribute with the same name. 

![image](https://user-images.githubusercontent.com/102611046/166243873-ef0cabed-4893-4478-995a-a486070738e1.png)

a20edwth:
Additional requirement: It should not be possible to have relations with the same name, unless they have the same properties.

This is not fine:
![3df552b2846ade8e394005c8b7ea744c](https://user-images.githubusercontent.com/102611046/166428435-10d6f35b-d8e9-4290-a265-c5dd00350581.png)

This is fine:
![c8d67f162e6cec0cbaa7b28faf67aa86](https://user-images.githubusercontent.com/102611046/166428445-d5b5194b-8a4f-4cb1-8a73-a16a1bd2a142.png)

a20edwth:
Error for when attributes on an entity have the same name, when entities have the same name and when relations have the same name but not the same properties.

Instructions for testing:
1) Click "Demo-Course"
2) Click "Diagram Dugga"
3) Test by creating lines as shown in the pictures located in the issue thread.

a20edwth:
Additional requirement: Attributes connected to an attribute that is connected to a relation should be recognized when checking if relations with the same name have the same properties.

Picture clarification: 
![779421916a870c62a675896c05679267](https://user-images.githubusercontent.com/102611046/166896573-ec3ee4ea-47fe-4dbb-a856-2002e2914e56.png)

PR #12195, opened 2022-05-05
Error for elements with same name

a20edwth:
Error for when attributes on an entity have the same name, when entities have the same name and when relations have the same name but not the same properties. https://github.com/HGustavs/LenaSYS/issues/12036

Instructions for testing:
1) Click "Demo-Course"
2) Click "Diagram Dugga"
3) Test by creating elements and lines as shown in the pictures located in the issue thread.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -6239,6 +6239,14 @@ function checkElementError(element)
 
     // Error checking for ER Entities
     if (element.kind == "EREntity") {
+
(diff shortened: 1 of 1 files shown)


a16linas:
Tested by adding and connecting attribute to already existing attribute with the same name and now they're displayed as red to indicate an error, also tried adding a ER entity and naming it identical to another entity and as expected it also shows red even though they're not connected
bild

Issue #12143, opened 2022-05-04
Move load, save and load dugga button further down on page.


b20thelo:
During a meeting with Henrik he expressed that he felt like the Save, Reset and Load Dugga buttons was too close to the diagram and he asked for this to be corrected.

Hint: Even though the diagram has its own CSS file called diagram.css this is handled by style.css. Henrik approved making changes to style.css if needed to correct this.

PR #12197, opened 2022-05-05
Move load, save and load dugga button further down on page. #12143

b20thelo:
- Added a margin at the top of the submitButtonTable to increase the distance between the buttons and the diagram dugga iframe as Henrik requested.

Result should look as the image below.

**Notice:** Solution had to be implemented in style.css which affects the whole LenaSYS page.  Therefore I've also checked multiple other dugga pages to see how the change affected these pages and my conclusion is that the change is so minimal that it doesn't affect them in a bad.


a20chrka:
Test passed, looks good.

PR #12199, opened 2022-05-05
fix: Ruler and A4 template does not match well/ruler does not show a known unit of length. #11453

a20jonda:
https://github.com/HGustavs/LenaSYS/issues/11453 .quick fix on misalignment on negative side.

image


a20chrka:
Test passed. Looks good, negative side is now aligned with the lines.

PR #12200, opened 2022-05-05
Added a drag and drop feature for code windows

a20willa:
It is now possible to drag and drop files into code viewer, and they will be automatically uploaded and added to the code window.


b20idawa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12204, opened 2022-05-06
Variants are no longer Undefined by default on localhost

a16linas:
When testing variants on local webserver, they're always "Undefined". It works fine on the production but can prove troublesome when testing variants locally. Now they're "Enabled" and Disabled"
bild


a20chrka:
Test passed, variant is no longer undefined as default.

Issue #12180, opened 2022-05-05
Behaviour of selection box in different scenarios.


a20jacmi:
When the mouse hover over the options pane and comes back to the container the selection box disappears. It reemerges when the container is clicked.

https://user-images.githubusercontent.com/102599629/166900406-25035f59-49ee-44a9-a583-a6ce86ea3299.mov

When edge creation is chosen the selection box disappears, but this doesn't happen when "placing an element" is chosen.

https://user-images.githubusercontent.com/102599629/166900421-29bd82a6-d642-41f6-8150-df62b24fcd40.mov

b20matad:
Problems solved by:
Options panel don't call "mdown()" on mouse down anymore.
All selected object are deselected if switching to "PLACING_ELEMENT" mousemode.

PR #12206, opened 2022-05-06
Behaviour of selection box in different scenarios #12180

a20jacmi:
There are two videos in the issue that clearly outlines the problem. Do as in the videos and notice a different outcome.


a20chrka:
Test passed, works as intended.

Issue #11916, opened 2022-04-28
Remove parent attribute from the list of attributes connected to an entity when generating the ER-Table.


a20chrka:
Sub-issue to: issue #11744

When an attribute (the parent) has other attributes connected to it (children), the parent attribute should not be stored in the ER-Table.
Consider the following example for the demo diagram dugga:
![image](https://user-images.githubusercontent.com/102599740/165770304-9445a495-472f-4873-87de-53eefb208453.png)
In this case, attribute "Name" is the parent and "F Name", "Initial" and "L Name" are the children.

However, "Name" is still added when the ER-Table is generated, even though it should not: 
![image](https://user-images.githubusercontent.com/102599740/165771097-f8c29afa-3f1a-4fcd-b777-722556c86213.png)

All functionality regarding the generation of the ER-Table string is found in lenasys/DuggaSys/diagram.js, in the function "generateErTableString()".

Issue #11915, opened 2022-04-28
ER-Table generation needs to consider foreign keys for each entity in the diagram.


a20chrka:
Sub-issue to: issue #11744

In ER, there are different rules for how you should handle relations between entities when creating an ER-Table based on an ER-model/diagram, specifically how to handle "1-1", "1-N" and "N-M" relations. Look up the rules for how to handle these, and implement functionality accordingly. (This may be of use: https://www.gatevidyalay.com/er-diagrams-to-tables/)

All code regarding the generation of the ER-table string is found in lenasys/DuggaSys/diagram.js, in the function "generateErTableString()".

I suspect that this issue may need further splitting, since each type of relation are resolved differently.

a20gabpa:
The solution has been merged.

Issue #11909, opened 2022-04-28
ER-Table generation needs to recognize key attributes. 


a20chrka:
Sub-issue to: issue #11744

Currently, there is nothing in the ER-Table generation algorithm that takes what "type" an attribute is into consideration. More specifically, if an attribute is a key or not. 

Take the following entity from the demo diagram dugga:
![image](https://user-images.githubusercontent.com/102599740/165751552-9ecb8909-6c5c-4162-881b-8bd846f965fa.png)

This is what the generated ER-Table entry looks like for entity "EMPLOYEE":
![image](https://user-images.githubusercontent.com/102599740/165755116-0857fb05-2370-4a24-9365-7cdf4a92f736.png)
Note that the primary key "Ssn" is not underlined even though it should be.

**Where to start:**
All functionality regarding the generation of the ER-Table String is found in lenasys/DuggaSys/diagram.js, specifically the function called "generateErTableString()". In this function, look for the following piece of code:
![image](https://user-images.githubusercontent.com/102599740/165760955-5368daed-63db-4b54-b3ca-429ecf5e3106.png)

Each element in currentEntityAttrList[] is a copy of each attribute that is connected to the entity that is currently being checked.
Therefore, it should be possible to check each element in currentEntityAttrList[] if they are a key attribute. All keys should have their ".name" underlined when pushed to stringList[].

PR #12207, opened 2022-05-06
Improvement for issue #11744 Table definitions created automatically when creating ER or UML diagram.

a20gabpa:
-  Improvement for issue #11744 
Related to issues:
- #11909 
- #11915
- #11916 

 What has been changed, diagram.js and mainly in function generateErTableString(), generateContextProperties() and some minor CSS changes.

- Start by selecting an ER element.
- Open the options panel
- Press the ER-table button

The options panel should change. Note weak entities is not implemented meaning that weak entities are not shown in the ER-table

For issue 11909 (recognize key attributes):
- In the textbox, key from every entity should be underlined. For example in the standard demo-data, the key SSN in Employee should be underlined. Also try adding, changing and removing key-attriubutes to entities.

For issue 11915 (consider foreign keys):

- In the textbox, foreign keys should be overlined. The foreign keys is implemented for 1-to-1, 1-to-N and N-to-M relation and is not implemented for weak (upcoming issue). Ideas on what to test, try creating new relation between entities, edit relation, check if the foreign keys are updated. 
- Also note, that the renaming of foreignKeys does not change per relation. For example, in demo-data, Employee will have departmentNumber from the relation 'WORKS_FOR' and departmentNumber from the relation 'MANAGES'.

For issue 11916 (removed parent attribute):

- In the textbox, the parent attribute should not be added to the ER-table. For example, in the demo-data 'Name' has three children, 'F Name',  'Intial', 'L Name'. The ER-table should only show 'F Name',  'Intial', 'L Name'. The parent 'Name' does not show. Ideas on what to test, creating more attributes with children, remove parent and connect children to entity.


a20chrka:
Test failed.
When selecting a key-attribute, bringing out options, pressing the ER-Table-button and pressing "delete" on the keyboard, the element is not removed from the screen.

a20chrka:
Test passed with latest commit.

PR #12209, opened 2022-05-06
Double click UMLRelations fixed #12203

### LINKED ISSUE #12203: Double-click on UML-inheritance element does not work.
With all other elements, if you double-click on them, the option-panel will pop out. This does not work with the UML-inheritance element.

The UML-inheritance element looks like this and can found if if you hold down the mouse-button on the ER-Relation symbol:
image

Note: You must be logged in as a teacher (the Toddler account) to access the feature in the default diagram.

b20matad:
Now when UML relations are double clicked the options menu will show.
To test: Try double clicking UML relations, both disjuct and overlapping, try the other elments to.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -4523,6 +4523,17 @@ function generateContextProperties()
 
               //If UML inheritance
               else if (element.kind = 'UMLRelation') {
+                //ID MUST START WITH "elementProperty_"!!!!!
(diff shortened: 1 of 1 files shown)


a20chrka:
Test passed. Double-clickling now works for UML-relatiions, both overlapping and disjoint.

i19matla:
Works for ER to.

Issue #12196, opened 2022-05-05
Copy and pasted elements cover messages


a20edwth:
When copied elements are pasted they will hide messages before they were pasted behind them. (including the "elemetns pasted" message)

Video showing this issue:

https://user-images.githubusercontent.com/102611046/166942222-24eeccdf-9a9f-4f22-a95e-a52caea24776.mp4

a20edwth:
Made it so messages will not hide behind pasted elements.

Instructions for testing:
1) Click "Demo-Course"
2) Click "Diagram Dugga"
3) "Copy"(ctrl + c) any element/s
4) "Paste"(ctrl + v) anywhere on the screen
5) Drag the pasted element/s towards the message in the down left corner of the diagram (if the elements are behind the message, it is correct)

a20chrka:
Test passed. Messages are now drawn in front of all elements in the diagram.

PR #12212, opened 2022-05-06
Pasted elements do not cover messages

a20edwth:
Made it so messages will not hide behind pasted elements. #12196 

Instructions for testing:
1) Click "Demo-Course"
2) Click "Diagram Dugga"
3) "Copy"(ctrl + c) any element/s
4) "Paste"(ctrl + v) anywhere on the screen
5) Drag the pasted element/s towards the message in the down left corner of the diagram (if the elements are behind the message, it is correct)

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -287,7 +287,7 @@
     left: 100px;
     bottom: 20px;
     position: absolute;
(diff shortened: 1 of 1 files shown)


a20chrka:
Test passed. Messages are now drawn in front of elements in diagram.

PR #12221, opened 2022-05-06
Background elements now respond correctly.

a17nikjo:
Background elements of home button, darkmode button and back button are now clickable.


b20hanwa (approved):
Looks good and works as it should when tested.

Issue #11993, opened 2022-05-02
File with same name can be uploaded


a20samul:
- As per now a file with the same name can be uploaded in edit file, which may cause confusion.  Fix this issue.

<img width="960" alt="helloWorldfile" src="https://user-images.githubusercontent.com/81613484/166210596-9b2b7406-7b89-4256-a2ca-a8cbf2a7460e.png">

PR #12222, opened 2022-05-06
File with same name and filetype will not be uploaded  #11993

b19karhj:
Uploading a single or multiple files to edit files should where there already is a file with the same filename and filetype should no longer be possible.

Test:
Go into edit files.
Upload  file/files with the same filename and same filetype: Alert should be displayed.
Upload file/files with same filename but different file type: Alert should not be displayed and file should be uploaded.


a20chrka:
Test passed. Prompts correctly when uploading a file with identical filename as another file.

Issue #12201, opened 2022-05-06
Empty dropdown option in "Add diagram to dugga". 


b20thelo:
Request from @HGustavs.

Add a new **empty** option in the dropdown "Add diagram to dugga". At the moment there is no empty option which causes the following problems. 

- If there is only one diagram JSON file available in the dropdown it looks as if it is selected. However, it hasn't yet been added to the parameter section. 
In order to add the select file in the parameter section, you need to switch from one option to another, as that is when the filename is added(not possible to switch with only one option).
- If you don't want the selected file to be added in the parameter section, but you've changed the option in the dropdown, you manually have to delete the file name in the parameter section.
A teacher wants to be able to create a diagram without a selected file as it isn't always needed for the task.

Sections connected to this issue:
<img src="https://user-images.githubusercontent.com/81772705/167071525-388a2a6e-850e-473d-a22e-aec1f86f6576.png"/>

PR #12223, opened 2022-05-06
Empty dropdown option in "Add diagram to dugga". #12201

b20thelo:
#12201 
**What has been done**
- Made adjustment to the feature that creates dropdowns so an empty option is shown. Since this solution is global and can be used by other dropdowns as well, if the if-statement is correct, we named it "Nothing selected" instead of something related to an empty canvas or demo data.
- Before what was shown in the dropdown was added in "diagram_File":"", which caused issues when we named the empty option "Nothing selected". This was solved by adding a value which for "Nothing selected" is empty and for the JSON files is the file name. 

**How to check solution**
Go to Edit dugga and select the diagram. There try to switch between options in the dropdown and check so that it updates as it should in the parameter section. 

There might still be an issue with what is shown in the diagram when you select the option "Nothing selected". Emil works on a solution for that in another issue, but the intention is that the hardcoded test data should be shown then.

### DIFF
--- a/DuggaSys/duggaed.js
+++ b/DuggaSys/duggaed.js
@@ -286,7 +286,9 @@ function selectVariant(vid, el) {
   				}
 				else if(result =="errorActive"){
 					document.getElementById("errorActive").checked = obj[result];
-				}
+				}else if(result == "diagram_File"){
+                    document.getElementById('file').value = obj[result];
+                }
   			}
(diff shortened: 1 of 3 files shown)


a20chrka:
Test passed. Params for "Diagram_file:" is empty when selecting "Nothing selected".

i19matla:
What is mentioned in the initial description as "options" above refers to: "Add diagram to dugga"

PR #12225, opened 2022-05-06
Added dugga title for superusers

b20jonlu:
Added the duggatitles for superusers. Previously the variable was only assigned if the user was a guest.

### DIFF
--- a/DuggaSys/showDuggaservice.php
+++ b/DuggaSys/showDuggaservice.php
@@ -322,6 +322,15 @@ if(isSuperUser($userid)){
 					$newcoursevers=$row['vers'];
 					$newduggaid=$row['quiz'];
 				}
+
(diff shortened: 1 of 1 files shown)


b20hanwa (changes requested):
The Dugga title works on some duggas, but not on all. For example; Clipping maskin dugga and Bit-dugga works but not 3D-dugga. I don't understand why this is though.

b20jonlu:
How 3d dugga and clipping maskin dugga looked before changes
duggatitle_just_didnot_work

The reason why 3d-dugga doesn't work is that it didn't work before either. it seems the code to render the instructions is tied to each duggas javascript and some of them don't work it would seem.

a20vicad (commented):
The fact that no title is shown for 3D-dugga is no major issue. This looks good in other cases.

PR #12231, opened 2022-05-06
Template 9 reszising now functional.

a20patpe:
It should now be possible to resize the boxes in template 9.
I will add that the increased (human)complexity is due to there being 4 boxes stacked on top of each other in this template.
If there is a better way to do this i implore the person reading to do that.
For testing:
1: Create a new code example.
2: Enter the code example.
3: Set it's template to template 9(9th from the left.)
4: resizing the different boxes should now work.


b20idawa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12234, opened 2022-05-06
Issue with scrolling in commit tree and weakly bar chart fixed #12086

### LINKED ISSUE #12086: Commit tree and Weekly bar chart "scroll" differently. In the contribution page
The Commit tree chart's scroll function is built into the chart. But on the Weekly bar chart overflows over the whole <div>

We want the Weekly bar chart to be the same as the Commit tree chart.
image

b20eripe:
Issue with scrolling solved and I also fixed the sizing of the diagrams to be the same when zoomed in.

### DIFF
--- a/DuggaSys/contribution.js
+++ b/DuggaSys/contribution.js
@@ -234,9 +234,10 @@ function renderBarDiagram(data) {
 
   // Renders the diagram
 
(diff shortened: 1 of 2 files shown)


a20vicad (commented):
Good, now they work the same way

Issue #12218, opened 2022-05-06
Verify password and username are valid on contribution.php


a20vicad:
Verify frontend & backend that entered password and username has valid chars and are valid in contribution.

#11416

a20behta:
related to #11416

a20behta:
issue note:
add vertifications or password in js before sending to php.

* no space
* 64 char max
* alphanumerical only. 
* no special characters.

PR #12240, opened 2022-05-06
Solved validation for username and password on contribution.php #12218

a20behta:
both backend and frontend will check using regex and 2 size more or equal 8 and less than or equal 64

note: only done for creating account. not checking. 

testing frontend: 
  change to branch on webserver  
  be logged out and go to to http://project- 
  g3.webug.his.se/DuggaSys/contribution.php

  try a password less than 8 characters 
  try a password more than 64 characters
  try a password with special characters. 

testing Backend
  
comment out else if statement in contribution.js
                            line 2323 - 2335

  try a password less than 8 characters 
  try a password more than 64 characters
  try a password with special characters.

### DIFF
--- a/DuggaSys/contribution.js
+++ b/DuggaSys/contribution.js
@@ -2320,7 +2320,7 @@ function resetForceLogin()
 		        $("input#password").removeClass("loginFail");
             $("input#password1").removeClass("loginFail");
 			  	}, 2000);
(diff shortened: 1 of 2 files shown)


a20vicad (commented):
Now it is much clearer for the user, good work.

Issue #12216, opened 2022-05-06
Keybindings in tooltip has disappeared


a20edwth:
Keybindings used to appear in the tooltip for buttons like bellow:

https://user-images.githubusercontent.com/102611046/167102610-c15ee5b3-3c8f-439f-bbbc-12eed1f27f40.mp4

As of now these keys have disappeared and only "Keybinding: " can be seen:

https://user-images.githubusercontent.com/102611046/167102767-5a7da40e-9975-4fda-b4cc-9944640600d4.mp4

PR #12243, opened 2022-05-06
Fixed keybindings and other related problems

a20edwth:
The cause to this issue was that the onload function in diagram.php **getData()** did not get properly executed. This was also the cause to other proplems with the diagram, such as the diagram being blank as a initial state and the grid not being turned on at the start. All of these things were done in **getData()**. The reason for why **getData()** did not execute, was the call for function **getVariantParam()** which is not defined. 

Picture of the error message containing this information:
unknown

To solve this issue a condition for executing **getVariantParam()** was created that checks if **getVariantParam()** is **not** undefined. #12216 

Instructions for testing:
1) Click "Demo-Course"
2) Click "Diagram Dugga"
3) Hover over buttons and see if keybindings are being shown

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -7360,6 +7360,7 @@ async function loadDiagram(file = null, shouldDisplayMessage = true)
 
 function fetchDiagramFileContentOnLoad()
 {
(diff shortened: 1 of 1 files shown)


a20chrka:
Already fixed in #12165

Issue #12135, opened 2022-05-04
The ability to move elements and moments should be limited


a20vicad:
The moments and elements should only be movable if the user drags in a specific spot like this:
![image](https://user-images.githubusercontent.com/102599327/166696611-1a9f27fd-122a-45b3-8562-0904273192fb.png)

![image](https://user-images.githubusercontent.com/102599327/166696191-fa138b14-8f08-49ac-8134-e328ae0cc61e.png)

a20caros:
<img width="944" alt="image" src="https://user-images.githubusercontent.com/81676918/167153787-a0417e6d-c533-42d8-a611-ae779d30fadd.png">
The issue is resolved by placing a button on the side of the table. Only when that button is visible can the element be moved. To solve the problem that sections and moments that are the headings should not be able to move when they are closed, a code has been added to tell when the element is closed or open. When it is open, the button appears and you can move the element and when it is closed, it is not possible to move the element.


To test the issue
1. choose a course
2. check that the buttons are located on the side and that you can only drag the element when you move the element with the button
3. check that the buttons on selection and moments (rurbikes) are only visible when the element is open and you can not move the element when it is closed

PR #12245, opened 2022-05-06
The ability to move elements and moments should be limited  #12135

a20novra:
The ability to move elements and moments is limited.


a20vicad (commented):
Interesting solution. I'll make another issue for the more visual part.

Issue #12177, opened 2022-05-05
Add colorblind-friendly colors for elements in the diagram.


a20chrka:
This issue includes research on what colors to use for colorblind users. If I recall correctly, there are different types of colorblindness that needs to be considered.

Until the customer says otherwise, simply add the colorblind-friendly colors to the list shown below:
![image](https://user-images.githubusercontent.com/102599740/166898463-2a68cf4e-5f22-4ea0-afb9-e16a55450e88.png)

a20samul:
Added the colour blind-friendly version of already pre-existing colours. Colours were extracted from the IBM colour blind palette. (Note that the text-color should be changed to white on colours black and newPink)

<img width="464" alt="colorBlindSafePalette" src="https://user-images.githubusercontent.com/81613484/167109388-fdca2fd8-036e-4861-bcbd-c6081d208427.png">

PR #12246, opened 2022-05-06
Updated the color selector to make it colour blind friendly #12177

a20samul:
Updated the color selector to make it colour blind friendly #12177

To test:
1. Select an element
2. Open the option panel
3. Try all the colours
Edit: The colour black was removed.

colorBlindSafePalette


a20chrka:
Test passed, looks good.

i19matla:
Creates a new issue due to discoveries during the test after this merge. The colorpalette is still left out if the optionspanel is sent in without collapsing the colorpalette first.

Issue #12181, opened 2022-05-05
Min-height on variants for diagram dugga


a20samul:
When uploading several submission types the diagram-variants-window (see image) the window for the uploaded diagrams decreases. 
- Set a min-height for the uploaded diagrams window so scrolling isn't frustrating.
- Also as a bonus: delete the top-margin that is set on _Submission types_  div.

<img width="891" alt="minimiwindowandmargini" src="https://user-images.githubusercontent.com/81613484/166900835-bbc2d28d-4291-4b21-9faf-37d9733fb927.png">

PR #12248, opened 2022-05-06
Fixed so "Uploaded diagram"-window doesn't shrink anymore so scrolling isn't as frustrating #12181

a20samul:
This pull request contains:
- #12181
- Set a minimum height on uploaded diagram window so it doesn't shrink when adding more submisson types leading to less  frustration when scrolling.
- Also removed the margin that was between "Diagram types allowed" and "Submission types".

Test by:
1. Log in as teacher
2. Go to "Edit dugga"
3. Click the pen on Diagram dugga 
4. Add submission types by pressing the (+)-button
5. Make sure that uploaded diagrams window doesn't shrink. 
6. Make sure there is no margin between "Diagram types allowed" and "Submission types".
*Note that it does make a little "jump" when the first submission type is added, the reason for that is the scroll being added to the side.

### After:
doesntshrink

### Before:
image

### DIFF
--- a/DuggaSys/duggaed.php
+++ b/DuggaSys/duggaed.php
@@ -209,7 +209,7 @@ session_start();
                       <!-- Submissions for dugga -->
                   <div>
                     <div id="duggaSubmissionForm">
-                      <fieldset style="width:90%; margin-top:5%;">
+                      <fieldset style="width:90%;">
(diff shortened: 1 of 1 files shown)


a20chrka:
Test passed, looks good. Definitely less annoying to scroll among variants now.

PR #12249, opened 2022-05-06
Implemented zooming function

a20joabo:
The Zoom In & zoom out - text buttons now make text smaller and larger. 
The reset button resets the font size to the default value of 9px.
There are limits in place to avoid the text overlapping with itself. 

This issue can be tested by: 

- Open a code example, preferably one with files ready. 
    - e.g Webbprogrammering -> JavaScript Example 2.
 - Pressing the Zoom in button will make text larger. Pressing the Zoom out button will text smaller. Pressing the reset button will make the text return to the default size of 9px.
image
- The user will not be allowed to zoom in so much that the text overlaps

### DIFF
--- a/DuggaSys/codeviewer.js
+++ b/DuggaSys/codeviewer.js
@@ -2656,12 +2656,20 @@ function getTotalBoxes(template) {
 
 function zoomText(boxid, increment)
 {
+	//Upper & lower limit of how small text can get.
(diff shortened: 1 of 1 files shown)


a19xinhu:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12253, opened 2022-05-06
Made all dates dynamic rather than static in the contributionsservice.php file

a19simru:
If/when we need to look at contributions data other than the late spring course 2019 this will now be much easier. On line 63 and 64 are two strings that represent the start date and end date of our data collection. When we want another course date we just need to edit these two strings and all sql queries will change search dates accordingly.


a20vicad (commented):
More dynamic now that it's possible to change start and end -date in a single place

PR #12255, opened 2022-05-09
Made the hamburgermenu icon bigger 

b20idawa:


b20idawa:
How to test the issue:

1. Log in as a teacher
2. Select the webprogamming course
3. Minimize the window so the hamburgermenu icon appear
4. Check that the hamburgermenu icon is bigger 
5. The result should look like the picture below
Skärmavbild 2022-05-09 kl  08 48 00

a19xinhu:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #12244, opened 2022-05-06
Closing the options panel makes the down-arrow disappear


a20edwth:
When closing the options panel it says "Options" instead of "▼Options" as it should.

Video showing this:

https://user-images.githubusercontent.com/102611046/167149364-7791e252-6627-4c79-af43-fa5f3f0b34c1.mp4

a20edwth:
Made it so when closing the options panel the arrow wont disappear.

Instructions for testing:
1) Click "Demo-Course"
2) Click "Diagram Dugga"
3) Close the options panel (if there is an arrow, it is correct).

PR #12269, opened 2022-05-09
Arrow does not disappear

a20edwth:
Made it so when closing the options panel the arrow wont disappear. #12244 

Instructions for testing:
1) Click "Demo-Course"
2) Click "Diagram Dugga"
3) Close the options panel (if there is an arrow, it is correct).


a20chrka:
Test passed, the arrow no longer dissapears after closing options.

PR #12270, opened 2022-05-09
G3 2022 w6 00001

a19simru:
Same change but navbar can't be seen if you are not logged in.


a20vicad (changes requested):
image
The navbar now looks like this in course main-pages

a19simru:
I don't precisely know why this works or what caused the problem in the first place but now there is no problem.

PR #12281, opened 2022-05-09
Added the right color when the page is in dark mode

b20fanfa:
Has the right color in the frame (in dark mode)


b20idawa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #12140, opened 2022-05-04
Option panel open and close functionality


b20thelo:
Make it possible to click the whole option panel to open and close it, not just on the name "Option" or the arrow/poop emoji. 

![image](https://user-images.githubusercontent.com/81772705/166806275-82bbc5af-e496-4405-a023-e25c0d04a908.png)

PR #12282, opened 2022-05-09
Option panel open and close functionality #12140

b20thelo:
Issue #12140  (Please note that I missed # in the branch name when you switch to the branch). 

**What has been done**
- Added so that the minimized option panel can be clicked anywhere to view the option panel.
- Adjusted the tooltip so that it still is aligned as it should.

**How to test**
In the diagram dugga check if it works to open/close the option panel by clicking anywhere on the marked part in the image below.

In my test, this still worked when resizing the page using the browser zoom.


a20chrka:
Test passed

i19matla:
Works on the server to.

PR #12293, opened 2022-05-09
Added database entries to duggor without useranswers

b20jonlu:
Added new rows to the variant and useranswer tables in order to fix the alert when viewing duggor when logged in as a superuser. Now every dugga in Demo-Course and Webbutveckling should have the data needed so that the alert won't show up. Also added some placeholder values for the param and variantanswer for the variant table so the duggor in Demo-Course aren't all blank.

**In order to test the changes you need to run install.php to update the database otherwise nothing will change!**


a20vicad (commented):
Looks like you got all duggas that needed.

Issue #12277, opened 2022-05-09
Create working keybinds for new functionalites in diagram dugga


a20samul:
### Add working keybinds to these functionalities:
- Enter replay mode
- Toggle ER-table
- <s>Toggle Error Check</s> Not yet merged/implemented

The three last buttons without keybind:
<img width="167" alt="addKeybinds" src="https://user-images.githubusercontent.com/81613484/167380513-9fd78ec6-aa42-430e-8f08-101a1f18ea90.png">

PR #12304, opened 2022-05-09
Created working keybinds for two functionalities #12277

a20samul:
- #12277
### To test:
**Make sure that you have pressed inside the diagram iframe first**
1. **Hover over the functionalities "Replay mode" and "ER-table". Make sure that the keybinds are present.**
keybinds

2. **Try the replay keybind by pressing "r". Make sure that the replay mode is enabled/disabled.** 

3. **Try the ER-toggle button by pressing "e".** 
    - Make sure that an ER-entity is **selected** and the option panel open. 
    - Note: that the option panel doesn't automatically open since you won't be able to toggle between the table and the other functions if that were the case.


i19matla:
Works as intended on the server.

Issue #12158, opened 2022-05-05
ER-Table generation: Allow the user to decide which candidate-key to be used as primary key.


a20chrka:
Main issue: #11744 

If there are multiple candidate-keys connected to the same element (see image below), only one should be generated in the ER-Table.
![image](https://user-images.githubusercontent.com/102599740/166882951-dabff59d-74eb-4edb-b8ad-c2cd0dc494b6.png)

 Needless to say but in reality (when creating the table by hand), the creator of the table decides which candidate key becomes the primary key. It should be the same for the automatic ER-Table generation. 

Until the customer says otherwise, create a checkbox for candidate key attributes in Options -> properties that represents if the key is chosen as primary key or not. Preferably, the checkbox should appear to the right of the "Variant-dropdown" (marked in red below), only when "key" is chosen:
![image](https://user-images.githubusercontent.com/102599740/166882609-8ef24d04-b056-42a6-a090-26587aa350bb.png)

Alternatively, add another "Variant", in the dropdown list:
![image](https://user-images.githubusercontent.com/102599740/166883331-4ff1a387-73bc-45e7-b6b7-0b8e7977eac0.png)

a20gabpa:
The solution has been merged.

PR #12307, opened 2022-05-09
Added ability to choose between candidate and primary keys #12158

a20gabpa:
Related to main-issue #11744 
and sub-issue #12158 

The user can now edit the type of keys on attributes. Note, the previously type namned 'key' is now split into two categories and are called 'candidate' and 'primary'. If an attribute is primary, only the primary keys will have underlines in the ER-table. If no primary keys are present all candidate keys are underlined instead. 

Also note that multiple primary keys and / or candidate keys can exist on the same attribute, this it to allow the user to learn by misstakes and not stop possible misstakes.

To test:

- Try adding more primary keys to an entity. Check if only primary keys are underlined in ER-table.
- Try adding more candidate keys to an entity and one or more primary keys. Check that only the primary key is underlined in the ER-table.
- Try changing the primary to a candidate. Check that all candidate are now underlined in the ER-table.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -3949,251 +3949,253 @@ function generateErTableString()
 
     //Iterate through all relations
     for (var i = 0; i < ERRelationData.length; i++) {        
-        //If it is a weak relation
-        if (ERRelationData[i][0].state == 'weak') {
-            //
-        }
-        else {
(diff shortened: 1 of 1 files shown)


a20chrka:
Test passed, works as intended.

i19matla:
Works as intended on server.

Issue #12299, opened 2022-05-09
Improve visuals of the "move element" area


a20vicad:
Related to #12135
![image](https://user-images.githubusercontent.com/102599327/167402844-165a50a4-2f90-443e-966b-b3fa9548f9b2.png)

PR #12308, opened 2022-05-09
G3 2022 w6 #12299

a20novra:
I canged the photo, and added some css.


a20vicad (commented):
This looks much better.

PR #12310, opened 2022-05-09
Modified the circle diagram to that multiple polygons are shown with the different types

c20adael:
To test this you need to go to the contribution page. Then you need to select a person from the database. After that you need to scoll down to the circle diagram and test its functionallity. 

BE AWARE: Some of the elements will be blocked behind other elements. This is intentional as of now.

image


b20hanwa (approved):
Works well when tested, good job!

PR #12313, opened 2022-05-09
Updated commitDiv to not show empty lines

b20jonlu:
Changed it so that empty lines of code no longer show in the list. Instead empty lines are counted and the total amount is displayed at the end of the section.

Before
commitdiv_blank_rows

After
commitdiv_empty_count


b20hanwa (approved):
Works well when tested!

b20hanwa:
Works as it is supposed to when tested, good job!

PR #12317, opened 2022-05-09
Prevented deleted headers from reappearing

a20johba:
All headers (and other items) should now delete properly. Changed the behavior slightly so that all items are deleted at the same time, and the timer refreshes every time an item is set to be deleted. This means that if you press the delete button on one item, wait 50 seconds, then press to delete another, the timer will be reset back to 60 seconds. After 60 seconds, both the items are deleted at the same time.

To test:
1. Create a couple of headers.
2. Delete as many as you want.
3. 60 seconds after the last item was set to be deleted, all items should delete. The table will briefly reload when this happens. However, there is no message. You can see when this happens by looking in the console log.
4. Only the headers you set to delete should be gone. The deleted headers should still be gone after refreshing the page or after adding new headers.
5. (optional) Add a header, delete it, and click the undo button. The header should reappear.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -560,6 +560,7 @@ function deleteItem(item_lid = null) {
    }, 60000);
 }
 
+// Permanently delete elements
 function deleteAll()
 {
   for(var i = delArr.length-1; i >= 0; --i){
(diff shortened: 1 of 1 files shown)


a20vicad:
At the moment, objects will return even if doing a small update like moving an element. This seems odd since it's not a restart of the page.

Issue #12239, opened 2022-05-06
Hide the button to turn on of error checking when error checking is not allowed


c20emili:
Subissue for #12236 
If error checking is not allowed there should be no button to turn it on.

PR #12319, opened 2022-05-09
Fix for issue #12239 Hide the button to turn on of error checking when error checking is not allowed

c20emili:
Added so the techer can choose if the error checking button is shown.
For testing:
When logged in as teacher, no variant is selected, no variant exists or error check button is deaktivated it should look like below
image
if error check button is deactivated it should look like below
image
- #12239


a20chrka:
Test failed, does not work properly on the web server, The button does not appear even though I'm logged in as teacher.

a20chrka:
Test passed. Teachers now has access to error-checking at all times.

Issue #12202, opened 2022-05-06
Edit course


b20sebka:
![image](https://user-images.githubusercontent.com/102599780/167079740-36c014fc-eec3-4c95-9d4e-f8d4f8e7a675.png)
When trying to edit the current course it shows an error in "version name" even though it is correct according to the standards.

PR #12322, opened 2022-05-09
Fixed error in "Edit course" version name #12202

a20andlj:
Before when you edited the current course it showed an error in "version name" even though it was correct according to the standards. This is fixed for both when you add new courses and when you edit existing courses.


a20andlj:
Test 1:
1. Go to the demo course logged in as a teacher
2. press the gearwheel
3. try different types of letters and numbers in the version form square to see if the border turns redd or green. Correct i 2 capital letters and to numbers.

Test 2
1. Go to the + sign next to the gearwheel and perform the same things as in test 1: point 3.

b20idawa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12323, opened 2022-05-09
Revert "File with same name and filetype will not be uploaded  #11993"

i19matla:
Reverts HGustavs/LenaSYS#12222

Found some faulties.


i19matla:
Need improvement.

PR #12326, opened 2022-05-10
change text color in dark mode on page Testing-course

a20sarle:
Text was almost in the same color as background before but should now be white, solution was to remove the color from #Sectionlist to make the text inherit the color from #content.

Testing:
1. go to Testing-Course. It should look like something below before the solution was implemented.
image
2. switch to this branch and go to Testing-course. The text should now look like below.
image


b20idawa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12342, opened 2022-05-10
changed the colours of rows of the code examples

a20sarle:
changed the dark colour #242729 to #303030 and #2d3133 to #343434.

Testing:
Go to course webprogrammering.

1. before switching to branch the background color of the rows should look something like this:
image

2. after switching to the branch it should look something like this:
image


b20idawa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12344, opened 2022-05-10
Auto logout of all instances

a20willa:
If logging out of one instance of LenaSYS, all other instances are auto logged out too.


b20idawa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12345, opened 2022-05-10
Set the font'size of the element in question to the  same as the others

a20patpe:
The font size of the shaderprogramming duggas is now the same as the other duuggas.
For testing:
1: Enter webprogrammering.
2: Reduce the screen size so that mobile view is activated.
image
3:It should look like this: 
image
Not like this
image


b20idawa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12351, opened 2022-05-10
Checkboxes should now skip invisible headers

a20johba:
Added an if-statement to skip headers with style.display set to "none".

To test:
1. Go to Demo Course
2. Create a new header
3. Delete the header
4. Click the checkbox on the top of the table to select all
5. All checkboxes down to the next visible header should be selected. Any deleted (hidden) headers should be skipped.


b20hanwa (approved):
Works as intended when tested!

PR #12355, opened 2022-05-10
made the icons white on the purple background

a20sarle:
Added filter to all the icons with id "dorf" to make them white. The "link icon" will be changed in another branch where and since the styling was moved to a new file.


b20idawa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #12316, opened 2022-05-09
The "pop-out" arrow for the instructions on the left side of the diagram needs to be updated.


a20chrka:
The arrow in question is the grey one directly to the left of the vertical purple bar with all the buttons for tools/mouse-modes:
![image](https://user-images.githubusercontent.com/102599740/167412248-3e538851-35e9-4499-b156-23ea37b914d2.png)

The following needs to be fixed:

- Animation when pressing the button. Currently it snaps to the new position instantly when pressed. It should follow the purple-bar.

- It seems to have a fixed position on the webpage. When scrolling up and down, it seems to be stuck on screen. This can also be seen when zooming out in the web-browser:
![diagraminstructionarrow](https://user-images.githubusercontent.com/102599740/167416435-c90d34e9-ac3e-4159-9635-e0c0c8d21917.PNG)
The arrow should always be between the diagram and the instructions-panel.

- The arrow is currently overlapping the scrollbar for the instructions:
![image](https://user-images.githubusercontent.com/102599740/167417030-9085c869-23c2-41b6-8890-c5985fe94b58.png)
Add an offset between the diagram and the instructions, so it no longer overlaps. 

- Flip the arrow when instruction panel is closed. This is what it currently looks like when the instruction panel is closed:
![image](https://user-images.githubusercontent.com/102599740/167417777-6ee31d70-d404-47fb-b1b1-581d815336a0.png)

PR #12357, opened 2022-05-10
Fixed the arrow functionality in diagram dugga #12316

a20samul:
**To make it easier have a duplicated tab opened of the g1-project webserver with and without the branch to preview the changes more clearly.**
### This pull request contains:
- Fixed so the arrow doesn't overlap the scrollbar
- Arrow flips when description is opened/closed.
- Animation is smoother as the arrow now follows the open/close animation of the description.
- Before the arrow scrolled outside of the screen when zoomed in/out but is now positioned absloute in one position in the middle.

## To test:

1. Make sure that the arrow doesn't overlap the scrollbar:
overlap

2. Make sure that the arrow flips when opened/closed.
openedclosed

3. When opening and closing make sure that the arrow doesn't jump from one place to another but rather follows along the animation when opening and closing the description.

4. Zoom in/out (outside the diagram, making the whole window smaller/bigger). Make sure that the arrow is positioned in the same position at all time (in the middle).
zoomed

### DIFF
--- a/DuggaSys/templates/diagram_dugga.html
+++ b/DuggaSys/templates/diagram_dugga.html
@@ -46,7 +46,7 @@
         
         #close_open {
             width: 23px;
-            position: fixed;
+            position: absolute;
             top: 520px;
             left: 272px;
             border-top: 14px solid transparent;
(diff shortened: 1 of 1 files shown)


a20chrka:
Test passed. Nice, looks great!

PR #12361, opened 2022-05-10
Changed the opacity to 65% 

a20mtafa:
I changed the opacity to 65 % on the elements text so that there will be more soothing color to the subheadings.


a19xinhu:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Remember to change branch name next time.

PR #12362, opened 2022-05-10
Buttons nolonger created on headers

a17nikjo:
both when logged in as teacher and as guest the buttons dont show up on headers/moments anymore.
image


b20hanwa (approved):
Works good when tested!

PR #12363, opened 2022-05-10
Removed one onclick that was called twice

b20eripe:
Removed an onclick function that was called twice. Tested on Chrome and Firefox and seems to work as intended.


a20vicad (commented):
Didn't take a long time to test or to read changes. Very appreciated.

PR #12366, opened 2022-05-10
Added home, return and darkmode to burger menu.

a18gusjo:
When resizing the window, to very small the icons seem to move around a bit. To solve this I added home, return and darkmode to the teacher burger menu when the screen is smaller than 320px in witdth. The problem is that you need to a teacher to have access to this menu. I therefor added a new burger menu that contains home, return and darkmode. This only appear if you are not a teacher and the screensize is less then 320 px.

**For testing:**
**1.**  Log in as stei and navigate to webbprogrammering.
**2.**  Open the inspect window (ctrl+shift+i) and make the inspect window very small.
image
**3.**  Open the hamburger menu which should now contain the home, return and and darkmode functions.
image
**4.**  Next, log out and see if another hamburger menu should appear. 
It should look like this:
image

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -21,7 +21,16 @@ var compareWeek = -604800000;
 let width = screen.width;
 var time;
 var lid;
+/*navburger*/
+function navBurgerChange(operation = 'click') {
 
+  var x = document.getElementById("navBurgerBox");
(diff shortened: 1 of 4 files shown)


b20idawa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12368, opened 2022-05-10
Moved contribution functions into dedicated file, cleaned up ajaxservice from previous contrib

a20sacka:
contribution_loginbox_service.php is the new service for the loginbox,

all javascript functions for contribution_loginbox.php specific functionality is moved into contribution_loginbox.js.

condensed several Ajaxservice methods that were for the old contribution_Service into one supported by the javascript function CONT_LOGINBOX_SERVICE_RETURN. This is instead of moving them into separate ajax calls.

changed the functions in contribution_loginbox_service.php to work with CONT_LOGINBOX_SERVICE_RETURN.

---------------------- Testing -----

Make sure it works like it did before everything was moved


a20vicad:
Things work like they did before. Which is mostly correct.

PR #12379, opened 2022-05-11
i have modified the baseline of the different polygons

c20adael:
This will make them differ from each other so that all of the rectangles are able to use its hover function. same goes for the polygons. 

This can be tested by going to the contribution page and selecting a username. then scroll down to the circle diagram and view the difference.


a20vicad (commented):
Now the diagram is much clearer.

PR #12396, opened 2022-05-11
changed regex to match eachother and use same syntax

a20behta:
matched the regex for both php and js files. 

reinstall lenasys (Lenasys/install/install.php)
once done try creating a user (Eg proci) with a password containing åäö.
it should result in a warning popping up


a20vicad (commented):
They do be matching

Issue #12399, opened 2022-05-11
Replay button in Replay mode.


a20jacmi:
In replay mode: When the play button is pressed the sequence starts from the initial state. In contrast, when the replay button is utilised it does not go back to the first state.

The replay button should start from the first state.

a20jacmi:
The line "onclick="stateMachine.replay(0)" in diagram.php was changed to "onclick="stateMachine.replay(-1)"; where "-1" is the initial state, which the replay button should always go to when pressed.

PR #12410, opened 2022-05-11
Fixed replay button & removed launch.json #12399

b20matad:
The replay button now always starts from beginning, before fix it started the replay on the second step. Also removed the launch.json file.

To test: Make some changes to a diagram, enter replay mode, and press the replay button (not play button) several times to see if it always starts from the initial state.


a20chrka:
test passed, replay now starts from step one (initial state of diagram)

i19matla:
Works fine

Issue #11605, opened 2022-04-13
Look if edit window can be in a single one instead of two


a20willa:
Pressing the edit button currently causes it to appear in another window. Can the second winndow be removed?

a20joabo:
The issue occurs when pressing the pencil button inside of the code editor. Pressing the edit button opens another page within a iframe. The section that should be changed is located in **codeViewer.php**.

The section responsible for the button is:
 `<td class="butto2 editbtn" onclick="showIframe(&quot;1&quot;,&quot;undefined&quot;);"><img id="dorf" title="Edit file" class="markdownIcon" src="../Shared/icons/newMarkdown.svg">`

Current functionality seems to be temporary since it does not match rest of the design of the code-viewer & is prone to failure. _Example_: pressing the back button in the new window takes the user back to the main page inside of the window. 

The functionality should be re-implemented to match the design of the code viewer and to meet requirements set out by the customer. 

This issue could be split up into multiple parts.

![image](https://user-images.githubusercontent.com/102584401/165935919-00b11270-94ec-49ce-a90c-62c169620b7b.png)
After pressing the back button within the window:
![image](https://user-images.githubusercontent.com/102584401/165935993-fb44300c-97da-4030-8484-891eb9096b05.png)

a20joabo:
![editbutton](https://user-images.githubusercontent.com/102584401/165942337-c14180a8-32b6-40e1-a22a-b6306e68905a.png)

The image shows a rough sketch of how the page should look after pressing the _Edit file_ button.

a20linth:
The preview window can be fetched from fileed.php using the fetch API and replacing the empty preview window in codeviewer.php with the preview window from fileed.php. Javascript functions from fileed.js associated with the preview window can then be accessed with getScript JQuery request.

PR #12411, opened 2022-05-11
Preview window is now shown as a single window #11605

a20linth:
The preview window is now fetched directly from fileed and display as a single window instead of being a iframe window of the fileed page.

Testing:
2) Enter Webbprogrammering
3) Open PHP Example 1
4) Click on "Edit file"
5) The preview window from fileed should be displayed as a single window
6) The preview window should behave like other windows
7) The other windows should also behave the same as before
image

### DIFF
--- a/DuggaSys/codeviewer.js
+++ b/DuggaSys/codeviewer.js
@@ -5078,13 +5078,40 @@ function showBox(id) {
 	});
 }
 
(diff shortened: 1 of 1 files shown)


a20sarle:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12414, opened 2022-05-11
made the home button bigger in original file to match the implementation of the other icons

a20sarle:
made the house / home button bigger in the global navigation bar, changes were made in the original file to match the other icon implementations.

Testing:

Go to any page where the global navigation bar can be seen.

Home button looked like this before:
image

Should on this branch be larger and look something like this:
image


a20joabo:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12415, opened 2022-05-11
Fix for issue #12119 and the error that come on certain duggas because of missing variables

c20emili:
Changed the incorrect if conditionand updated  the install sql.
test by going do different duggas and se if you get 
image
If test passed speak with the different groups so the get the solution


b20matad:
diagram.json file did not exist after install correctly in the database.
And variant.json string need to be updated to include gfile-link.

i19matla:
Works fine.

PR #12417, opened 2022-05-11
Removed any changes to positioning of elements i BlackTheme

a20patpe:
BlackTheme.css should now have no mentions whatsoever of anything to do with the Position, Widht, Height, Margins, Paddings, top,left,right,bottom.
Since it is only concerned with the color of the Page.
For testing:
1: Navigate through lenasys.
2: Constantly switch between dark and light mode.
3: No element should change position at all.(Meaning that blackTheme.css does not change the position of anything.)

### DIFF
--- a/Shared/css/blackTheme.css
+++ b/Shared/css/blackTheme.css
@@ -75,8 +75,7 @@
 }
 
 #picktemplate img{
-	width: 140px;
-	height: 140px;
+
 }
 
 #picktemplate h1{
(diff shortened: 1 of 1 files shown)


a20sarle:
Need to be tested in a certain way. Wait to test untill' monday.

a20sarle:
Tested by group leaders last week. Changing to "waiting for merge".

PR #12418, opened 2022-05-11
Made text the same nuance as the rest text and icons

a20sarle:
Made the text the same nuance as the rest text and nuance in dark mode in by changing the percentage of the inverted filter.

Testing:

1. Go to dark mode
2. Go to course Demo-Course

Bafore the text looked something like this:
image

On this branch it should look something like this:
image


a20joabo:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12421, opened 2022-05-11
Added a dark mode variant for drag and drop immage

a20willa:


a20joabo:

Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12422, opened 2022-05-11
Removed MessageOfTheDAyButton

a20patpe:
The removal of this was not included in the week 7 merge. Removing it again.
For testing:
1: Enter a course.
2: The message of the day should still pop up but there should not be a button.
image


a20sarle:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12426, opened 2022-05-11
removed function onlload

a20willa:


a20sarle:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #12095, opened 2022-05-03
Using "angle-brackets" (<) in a line's label causes problems


a20chrka:
![image](https://user-images.githubusercontent.com/102599740/166455022-215ff76f-f76d-463b-ba2f-629376b4e516.png)
Note that the label is named "<< included >>", but only "<" appears on the selected line in the diagram.

Issue #12424, opened 2022-05-11
A new functionality has been added create new icon for it


a20samul:
A new functionality has been created refer to issue #12095, create an icon for it.

PR #12430, opened 2022-05-11
fixed "angel-brackets" with a esc method issue #12095

a19meram:
Testning:

- go to diagram dugga.

- Click on a double line and type <&# 60 ; Include>> or something else then include.

- After that, all the brackets shall be visible with the text.


b20matad:
Need branch G1-2022-W7-#12424 to be merged before testing can be done.

Issue #12420, opened 2022-05-11
Navbar breaks when resizing window


b20sebka:
![image](https://user-images.githubusercontent.com/102599780/167859481-ea92764b-a31f-4147-8861-3e302e975e86.png)
Navbar breaks just before the hamburger menu is created.

PR #12431, opened 2022-05-11
Fixed navheader and burgermenu ranges #12420

a20linth:
"access" div box is now the correct size and burger menu is now displayed in other browser widths.

Testing:
1) Logged in as stei
2) Enter Demo-Course
3) When resizing the window width the icons in the navheader should not go outside of the navheader and appear under it.

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -2656,7 +2656,7 @@ button.option {
  left: auto;
 }
 
-@media only screen and (max-width: 420px) {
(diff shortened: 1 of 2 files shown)


a20sarle:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Meet issue goal, but would be nice it the icons would stop "jumping"... probably because the "log in / out " button gets larger.

Innan:
image

Efter:
image

PR #12432, opened 2022-05-11
Created an icon for the new angle-brackets functionality

a20samul:
created an icon for the new functionality, refer to #12095 and #12424.
This should be merged into #12095 or tested after #12095. 
To test: the icon should look like this <>


a20chrka:
Can only be tested in by merging.

Issue #12423, opened 2022-05-11
Tool tip not positioned correctly


b20thelo:
After the latest merge, the positioning of the tooltips is off. Fix so that tooltip at both the options panel and icon toolbar is positioned correctly.

![image](https://user-images.githubusercontent.com/81772705/167864070-833edd70-9863-422f-b267-df471527f4d2.png)

PR #12434, opened 2022-05-11
Tool tip not positioned correctly #12423

b20thelo:
Issue #12423

**What has been done**
- Used already existing ID for tooltip in the options panel to give different styling for it.
- Corrected the changed styling that caused the issue after the weekly merge for tooltip in the toolbar.

**How to test**
Hover over the toolbar and options panel to see that the tooltip is positioned correctly. 

**What it should look like**
image
image

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -337,10 +337,9 @@
 
 /*tooltip*/
 
-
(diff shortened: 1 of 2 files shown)


a20chrka:
Test passed, tooltips are now drawn correctly

PR #12435, opened 2022-05-12
changed press duration for the code example list to show when press on forward or backward button

a20sarle:
One can press down the forward or backward button in the global navigation to show a list of the previous and next code examples. To make the list show quicker the press duration was changed from 1000ms to 300ms.

Testing:

2. Klick one of the code examples
3. Press the forward or backward button in the global navigation bar 
image

Before it should take lolnger time for the list to show whereas the displaying of list should feel more responsive on this branch. The change is minimal though.
image


a20joabo:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #12391, opened 2022-05-11
Relation elements turns black when trying to draw lines.


a20chrka:
This is what it looks like:
![image](https://user-images.githubusercontent.com/102599740/167842455-0ddb289d-5779-4c3b-984f-252aff8a1d9e.png)

After placing the line and moving the diagram, they are turned back to normal.

a20jacmi:
This was just a typo in "diagram.js" under this function:

<img width="630" alt="1" src="https://user-images.githubusercontent.com/102599629/168026350-e3d27f35-0eff-472f-b060-5e2c79077061.png">

Quotation marks around "fill" and "class" were messed up:
<img width="1401" alt="2" src="https://user-images.githubusercontent.com/102599629/168026354-d8248b42-1221-4ef2-aacb-9d34b805e5cb.png">

PR #12449, opened 2022-05-12
ERRelations no longer turn black in EDGE_CREATION #12391

b20matad:
Problem was a typo how ERRelations are drawn in function drawElement().
To test: Draw a line from different elements, ERRelations shouldn't turn black anymore...


a20chrka:
Test passed, relations no longer turn black.

i19matla:
Works fine

Issue #12413, opened 2022-05-11
User gets prompted for "OK" several times if no default general information file exists


a16linas:
Related to #12251 

When opening a diagram with only default files, the site will give error that it cannot read file in a directory and prompts user to press OK several times until the diagram is displayed. Problem seems to be at: https://github.com/HGustavs/LenaSYS/blob/1172c541dc21c27f57d8c5ccacb8e5222f0774ea/DuggaSys/showDugga.php#L84-L124.

![bild](https://user-images.githubusercontent.com/37794783/167850813-d72ab0c5-7ab1-4571-8ed9-54b6a48a8b03.png)

Think it only needs an exception handling

PR #12450, opened 2022-05-12
User no longer gets prompted to press OK multiple times when opening dugga

a16linas:
Added some exception handlings so it won't try to fetch any files for default variants. This wouldn't have been a problem on the production, I suspect, but proved difficult for testing. Now when opening any dugga regardless of variant parameters, it should not give an error for undefined variables.


a20chrka:
Test passed, it works better. There are still related issues but this is definitely an improvement.

i19matla:
Something has happend along the way, worked during the earlier testing but not now. Not merged to the weekbranch and not as the stand alone G1-2022-W7-#12413 branch either right now and there were no conflicts. Can´t see right now why this happend.

Issue #12386, opened 2022-05-11
Background color change


b20sebka:
![image](https://user-images.githubusercontent.com/102599780/167841430-8715d973-5bde-46a1-a402-7cc6617597f7.png)
Code has a different background color than the box

PR #12452, opened 2022-05-12
Fixed consistent darkmode background color #12386

a20linth:
The background color in darkmode for codeviewer text and textwrapper are now the same shade.

Testing:
1) Enter Webbprogrammering
2) Open any code example
3) Enter darkmode
The codeviewer should look like this:
image

And not like this:
image


a20sarle:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12460, opened 2022-05-12
Teacher can now select instruction and information files from a dropdown box

a16linas:
Instead of having to manually enter the filename, they can now select them from a dropdown based on whichever file extension they want
bild


a20jacmi:
This seems to work, although the number of test files were limited.

Issue #12409, opened 2022-05-11
Remove the "Check" frame from the toolbar when logged out (or logged in as student) AND when error check is disabled


a20chrka:
This is what it currently looks like when logged out and error check is disabled:
![image](https://user-images.githubusercontent.com/102599740/167846433-169e9b28-2aff-411b-b77c-5c953f870475.png)

Remove the "Check" and it's box.

a20edwth:
The error check field gets hidden when error checking is turned of. 

Instructions for testing:
1) Click "Demo-Course"
2) Click "Diagram Dugga"
3) Turn off error checking with a teacher through dugga variants
4) Check if the error checking field disappears when error checking is turned off.

PR #12463, opened 2022-05-12
The error check field gets hidden

a20edwth:
The error check field gets hidden when error checking is turned of. #12409 

Instructions for testing:
1) Click "Demo-Course"
2) Click "Diagram Dugga"
3) Turn off error checking with a teacher through dugga variants
4) Check if the error checking field disappears when error checking is turned off.


a20chrka:
Test passed, box for error-checking button is no longer visible if logged out and error chech is disabled in variant.

i19matla:
I do not delete the branch since there is a subbranch to this.

Issue #12142, opened 2022-05-04
Lines between ER and UML elements


b20thelo:
Since the UML functionality has been added it is possible to draw lines between the different types of diagrams. Here are a few examples. How should these be handled?

ER entity to UML entity
<img src="https://user-images.githubusercontent.com/81772705/166809425-42312608-76e3-484f-8e78-b600433ed0b9.png" height="150">

UML entity to ER relation
<img src="https://user-images.githubusercontent.com/81772705/166810167-956f7814-e603-4c1b-aeda-cbdbae491020.png" height="150">

UML entity to ER attribute
<img src="https://user-images.githubusercontent.com/81772705/166810319-bbd41dfa-1b9f-415c-a472-230f3155d87c.png" height="150">

ER entity to UML inheritance
<img src="https://user-images.githubusercontent.com/81772705/166810458-f44fbd43-d30f-4856-9c2e-24c0a31d11f9.png" height="150">

@HGustavs Do you have an any feedback regarding this? Should this be blocked?

b20matad:
Added a check to the function "addLine()" in diagram.js to detect if the two elements the line is added between has the same type, i.e UML or ER.

PR #12477, opened 2022-05-12
Lines between UML- and ER-elements: Issue #12142.

a20jacmi:
It should not be possible to draw lines between UML- and ER-elements, thus this functionality is now restricted. To test this, simply try to draw a line between these types and notice the error message in the bottom left corner.


a20chrka:
On hold. Might have been some misscommunication, we need to check with the customer first.

a20jacmi:
This was done in a new pull request on the next merge.

PR #12478, opened 2022-05-12
Added safeguard for the file fetching in showDugga

b20eripe:
Added safeguards for the file fetching to prevent the possibility to fetch an directory. Tested in Chrome and Firefox on an Linux machine


a20vicad (commented):
I can't say that I see the issue being solved yet as the prompts still appear.

a20vicad (changes requested):
I can't say that I see the issue being solved yet as the prompts still appear.

PR #12479, opened 2022-05-12
Fixed generic_dugga_file_receive for superusers.

b20jonlu:
Fixed the issue of text files being saved as json. Also changed the location of new submissions from htdocs/submissions to duggasys/submissions.
Updated demo-course sql to use unique hash and password.
Made it possible to view all submissions when logged in as superuser if none was selected.

Creating submission as guest - Generic Dugga File Receive

https://user-images.githubusercontent.com/102533453/168076917-144991e8-afbd-4244-9023-c132778eb3d3.mp4

Viewing all submissions as SuperUser - Generic Dugga File Receive
viewing site as super

Viewing specific submission as SuperUser - Generic Dugga File Receive

https://user-images.githubusercontent.com/102533453/168076929-fc854632-947f-4192-aa3c-38dd88d28d51.mp4

Viewing all submission as SuperUser - Rapport inlämning
viewing site as super_raport

### DIFF
--- a/DuggaSys/filereceive_dugga.php
+++ b/DuggaSys/filereceive_dugga.php
@@ -338,7 +338,7 @@ logServiceEvent($log_uuid, EventTypes::ServiceServerStart, "filereceive_dugga.ph
 				$seq++;
 
 				$filepath=$submissionpath;
-				if ($fieldkind = 4) {
+				if ($fieldkind == 4) {
 					$extension = "json";
 					$mime = "json";
 				}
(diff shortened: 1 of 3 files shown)


b20jonlu:
This code does not work on the web-server. The issue appears to be the caused by the code line `$zip = new ZipArchive;` in the `function processDuggaFiles()` in `showDuggaservice.php `. 

The localhost version works
localhost

The project-g3.webug version has a status 500 error
project-g3

There is another place where ZipArchive is used in the code and that is in the file downloadzip.php which also doesn't work on the web-server. The problem appears to be with the new ZipArchive line and not one of its functions which is where i will start searching for a solution

a20vicad (commented):
We need a solution for the zipArchive issue for the webserver but this is a good start.

PR #12484, opened 2022-05-12
Temp merge highlighting

c20adael:
This tempmerge is a combined project from @a20caros , @a20novra and @c20adael. What has been done is that there is highlighting when you create any type of test or dugga/heading/test on a course site.

Also highlighting has been added to the front page so that when you update a course or create a new one it will show witch one it is. 

How to test: Go to any course and create something from the fabbtn and check if it gets highlighted red.

### DIFF
--- a/DuggaSys/courseed.js
+++ b/DuggaSys/courseed.js
@@ -10,6 +10,7 @@ var versions;
 var entries;
 var motd;
 var readonly;
+var updateCourseName = false;
 
 $(document).ready(function(){
     $('#startdate').datepicker({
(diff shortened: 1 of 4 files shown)


a20caros:
How to test: 
1. on the first page 
2. login to a super user 
3. click on the cogwheel  and change the course name
image
4. check if the updated course name is highlighted in red 

How to test:
1. on the first page 
2. login to a super user 
3. click on the fabbtn and create a new course 
image
4. check if the new course is highlighted in red

Issue #12481, opened 2022-05-12
Lines between UML-classes.


a20jacmi:
Right now it's not possible to draw lines between UML-entities.

a20jacmi:
Now it is possible to do this. An exception was added in "diagram.js" under the function "addLine()" in the following line (this is the code before our implementation).:

<img width="1019" alt="Screenshot 2022-05-13 at 09 00 37" src="https://user-images.githubusercontent.com/102599629/168229800-696d1180-6720-4d61-ae53-d51c11c4817a.png">

PR #12492, opened 2022-05-13
Lines between UMLEntities #12481

b20matad:
Added exception so lines can be drawn between two UMLEntities.

To Test: Try to draw lines between two UMLEntities - This should be possible. 
Also try to draw lines between other elements of the same kind - This should not be possible, except for ERAttributes.


a20chrka:
Test passed, but new related issues will be created.

PR #12493, opened 2022-05-13
Added a burger menu for the "eye" and "ghost buttons"

a18gusjo:
Added a burger menu for the eye and ghost button that will be displayed when the width of the window is <=800.
The tab-button is not included because it will be removed next merge. 

**For Testing:**
**1**. Login as Stei and navigate to webbprogrammering.
image

**2**. Resize the window to 800 px width or enter mobile view. (a  burger menu should appear)
image

**3**. Test the functionality of the buttons by clicking the checkbox of a few code examples and then clicking the "ghost-button". Repeat the process but unhide the code examples with the "eye-button" instead.
image


a20joabo:
Code refactoring - ok
Code standard - no
Meet issue goal - no

Code standard is not met since there are seemingly two pairs of buttons - one pair visible normally and another pair only in use once the window gets small enough that the hamburger menu appears.

A solution following the code standard would be moving the buttons inside of the hamburger menu instead.

a20muhal:
Not relevant anymore, closing!

Issue #12458, opened 2022-05-12
Fix error handling in showDugga.php to prevent error alerts when loading an existing diagram file.


a20micro:
Loading a diagram file generates a bunch of error alerts the user has to cycle through before the page loads the actual diagram. 
As of issue #12413 (not merged with master) this behaviour is fixed when it comes to loading a diagram dugga without an assigned diagram.
![ForNewIssue](https://user-images.githubusercontent.com/95532115/168041917-65e558e6-b173-4a91-aa12-a638acbbd02c.png)
This behaviour shouldn't happen in at all and all errors should be handled correctly, without stalling the page for the user.

PR #12494, opened 2022-05-13
Fixed multiple variables generating error alerts when opening diagram dugga #12458

a20micro:
Added proper error handling for variables regarding fetching of diagram .json files, instruction files and general error information. By adding any type of variant to a diagram dugga, there should be no error message popup, and if the .json file is undeclared, should load in the standard demo-data as a diagram.

To Test:
1. open diagram dugga.

Note: To completely test, all combinations of variant params should be checked. However, this would take a long time and could instead be done with two tests, one where each field contains something, and one where all fields are empty.


a20chrka:
Test passed, no error messages appear when loading diagram.

PR #12499, opened 2022-05-13
Drag icon no longer shown when not logged in

a20willa:
For testing:
1.  Go to corseed.php and click on any course
2. Login and check if drag icons on the most left is visible
3. Logout and see if they are not shown anymore


a20sarle:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12505, opened 2022-05-13
Height of logged in button is  now 30px in mobile view. 

a18gusjo:
Height of logged in button is 30px in mobile view instead of 50px.

**For Testing:**
1. Be logged in and enter mobile view. 
2. Check whether the height of the login button is larger than the height of the nav bar.
It should look like this:
image


a20sarle:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #12462, opened 2022-05-12
Colours not working properly for all elements.


b20thelo:
<h2>ER elements not coloured correctly</h2>

**Derived attribute**
<img src="https://user-images.githubusercontent.com/81772705/168041497-1f3edee7-9e3c-4fe1-be2d-770a8c8f2a51.png" width="200"/>
Inner and outer lines doesn't look the same.

**Weak relation**
<img src="https://user-images.githubusercontent.com/81772705/168041767-e39fb8b2-d6e6-4ee0-ae6d-6e5a150d19c2.png" width="200"/>
Lines look very dark

**Relation**
<img src="https://user-images.githubusercontent.com/81772705/168042334-8d8224ad-0ad2-45fa-8c0c-cfe24ee12bc9.png" width="200"/>
Lines look a bit dark.

Lines
<img src="https://user-images.githubusercontent.com/81772705/168042506-f8d5211f-4318-4d78-880d-fe521e0970c5.png" width="200"/>
Different colours on lines and entity

**Attributes**
<img src="https://user-images.githubusercontent.com/81772705/168037566-1081cbdb-3193-4b4a-960d-478da57bd91d.png" width="200"/>
Line around attribute isn't as thick as other elements.

b20thelo:
What has been made

- Fixed typo error which caused outer line on normal attributes and multivalued attribute to look thinner.
- Removed stroke from element values as the value is no longer in use. (Might cause programmers confusion in the future if it stayed).
- Changed black stroke colour to websafe dark colour.
- Changed so that errorReset doesn't use by default a black, but instead gets stroke colour from the variable list.
- Fixed so that the weak key also changes to white when multiple objects are selected. Also fixed so that it changes back when unselected.

PR #12506, opened 2022-05-13
Colours not working properly for all elements. #12462

b20thelo:
#12462

**What has been made**
- Fixed typo error which caused outer line on normal attributes and multivalued attribute to look thinner.
- Removed stroke from element values as the value is no longer in use. (Might cause programmers confusion in the future if it stayed).
- Changed black stroke colour to websafe dark colour. 
- Changed so that errorReset doesn't use by default a black, but instead gets stroke colour from the variable list.  
- Fixed so that the weak key also changes to white when multiple objects are selected. Also fixed so that it changes back when unselected. 

**How to test**
Open the diagram and make sure you have an attribute with a weak key. For example Name in image below:

Then select this attribute together with at least one more attribute. The dashed underline should change to white together with the text, then back to black when being unselected. 

Before and after for attribute stroke thickness.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -8253,6 +8253,7 @@ function updateCSSForAllElements()
             var useDelta = (inContext && movingObject);
             var fillColor;
             var fontColor;
+            var weakKeyUnderline;
             if (data[i].isLocked) useDelta = false;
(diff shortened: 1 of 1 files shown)


a20chrka:
Test passed, looks good!

PR #12507, opened 2022-05-13
Greyed out zooming buttons when no longer clickable. 

a20joabo:
The appearance of the zooming buttons now behave intuitive.

The Zoom In & zoom out buttons now grey out when they are no longer clickable. 
Pressing the opposite button of a greyed out one will make the other clickable. 

The reset button will make both buttons press-able again.

Buttons legend

**This issue can be tested by:**

1. Open a code example, preferably one with files ready.
        e.g Webbprogrammering -> JavaScript Example 2.

2. Pressing the zoom in button will make the text larger up until the max font size has been reached.
3. Whenever the max font size has been reached the zoom in button now gets greyed out.
4. Press the opposite zoom button, Zoom in button is now clickable.
5. Max zoom again.
6. Press reset button and the button gets clickable again.
7. Repeat steps 2-6 for the Zoom out button.

### DIFF
--- a/DuggaSys/codeviewer.js
+++ b/DuggaSys/codeviewer.js
@@ -2807,15 +2807,23 @@ function zoomText(boxid, increment)
 	}
 	
 	else { 
-	
-		var zoomButton = document.querySelector('#box'+boxid+'wrapper #zoomIn');
 		
-		zoomButton.style.opacity = "0.4";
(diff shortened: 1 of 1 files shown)


a20sarle:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #12442, opened 2022-05-12
(Mobile view?) Copy to clipboard does nothing


b20sebka:
![image](https://user-images.githubusercontent.com/102599780/168018941-c3198003-2d3d-4b4b-8a7d-2a9897d3d0b5.png)
When clicking "copy to clipboard" the yellow overlay comes up but nothing is copied

a20linth:
It is the same in desktop view

a20linth:
When the "Copy to clipboard" button is pressed on ios or android the selected text in the code viewer is deselected before copyCodeToClipboard() is executed. The function copies the empty selection followed by the deselection. This does not happen on desktop because when "Copy to clipboard" is pressed the selection is not deselected.

PR #12511, opened 2022-05-13
Added mobile support for copy button #12442

a20linth:
Added a 0.5 second delay to the selectioncahnge event because when mobile users have selected code and press the cody button their selection is deselected before the copy command is executed. With the slight delay that will not be a problem. Also added clipboard  API support to replace depricated method execCommand().

Testing:
2) Enter Webbprogrammering
3) Open JavaScript Example 1
4) Select a code paragraph and click the "Copy to clipboard" button. The selected code should have been copied to the clipboard
5)  Click the "Copy to clipboard" button with no selection. If there are rows marked as "important" that should be copied, if not, the clipboard should be unchanged.
6) Now also test this on a mobile device (Not mobile view on your desktop). This should work the same on both iOS and Android.


a20sarle:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12512, opened 2022-05-13
Page needs to open in new external tab

a20sarle:
Solution was to add the attribute `target` with value `_blank`.

Testing:
1. Log in as Stei (password is "password")
2. Go to course "Webbprogrammering"
3. Click the button which has been circled in the screenshot below
image

Before: Link opened in the same window
image

After: Link should open in new browser tab keeping the previous tab next to it
image


a20joabo:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12513, opened 2022-05-13
added a function to sectioned.js to check if user is inlogged

a20novra:
added a function to sectioned.js to check if user is inlogged. there is an if sats that show the buttons if user is inlogged.


a20vicad (commented):
Good, now it's only visible for logged in. Real machine

PR #12519, opened 2022-05-13
Added resizing support for browsers other than chrome. And added a function for hiding the text in boxes.

a20patpe:
Resizing should now work for all browsers. 
For testing: 
1: Create a code example for each template.
image
2: Go into template 1 and pick template 1.
3: resizing should work as expecet in all browsers.
4: resizing the widht of a box to as small as it can be should hide the text.
image
5: Press the forward button and repeat for all templates.
image

Known bug: 
The initial resizing of the height of the stacked boxes in templates 6,7,9 is a bit "jumpy". 
This is due to a rounding error with the initial size of the boxes. Once the initial resizing of the box has occured there should not be a problem with resizing the boxes in that template.

### DIFF
--- a/DuggaSys/codeviewer.js
+++ b/DuggaSys/codeviewer.js
@@ -3952,13 +3952,14 @@ function resizeBoxes(parent, templateId)
 	{
 		$('#box1wrapper').resizable({
 			resize: function( event, ui ) {
-				remaining = ($(parent).width()) - $('#box1wrapper').width();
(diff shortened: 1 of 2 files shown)


a20sarle:
Code refactoring - ?
Code standard - ok
Meet issue goal - yes

a20muhal:
Code deletion is too extensive. It's hard to determine if faults will occur because of this. 

Group lead requests sit down with development and test for faults.

PR #12523, opened 2022-05-13
Fixed so that overlay works over linnes

a20willa:
For testing: 
1. Go into a code example
2. hover over a example with some code
3. Check that overlay works


a20joabo:

Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #12525, opened 2022-05-13
Error check can still be turned on through keybind when hidden


a20samul:
### The error check can still be turned on by using its keybind _"h"_ on the keyboard. 
- Remove the keybind for the error checking functionality as it should mainly be reserved for teachers and not always attainable by students unless purposely made so.

<img width="947" alt="h" src="https://user-images.githubusercontent.com/81613484/168319094-201cacac-2e65-4621-bb4e-ba75da7dfa78.png">

a20samul:
Issue expanded to that the keybind should be there but only for the teachers.

a20samul:
Issue expanded to that the keybind should be there but only for the teachers.

a20samul:
## Issue expanded to that the keybind should be present but only for teachers.

PR #12529, opened 2022-05-16
Students no longer have access to error check when logged out through keyboard #12525

a20samul:
Removed the keybind for the error check functionality so that logged out students don't always have access to it unless purposely given to them, as it is mainly reserved for teachers.

### To test:

1. Go to diagram dugga
2. **Log out** if not already done so
3. Press the _Ssn_ attribute 
4. Open Option panel
5. Change the variant from candidate to _weakKey_ and press save.
6. Press "h" on your keyboard and make sure that the boxes don't turn red.
7. Log in (teacher mode)
8. Repeat steps 3-6 and make sure that the boxes turn red.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -757,7 +757,6 @@ const keybinds = {
         TOGGLE_KEYBINDLIST: {key: "F1", ctrl: false},
         TOGGLE_REPLAY_MODE: {key: "r", ctrl: false},
         TOGGLE_ER_TABLE: {key: "e", ctrl: false},
-        TOGGLE_ERROR_CHECK:  {key: "h", ctrl: false},
 };
 
 /** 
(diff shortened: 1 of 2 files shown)


a20chrka:
Need improvement. When you press "H" now, it NEVER works, even when you are logged in as a teacher (when error-checking should always be available). If the button is visible, pressing H should work.

a20samul:
Same testing conditions apply: 

## To test:
1. Go to diagram dugga and **log in**!
3. Press the Ssn attribute that is connected to the "Employee" entity.
4. Open Option panel
5. Change the variant from candidate to weakKey and press save.
6. Press "h" on your keyboard and make sure that the boxes don't turn red.
7. Log out and repeat steps 2-5 but this time make sure that the boxes do turn red.

a20chrka:
Test passed, pressing H now works when logged in as teacher.

Issue #12495, opened 2022-05-13
Purple banner (at the top of the screen, with the home button) is not loaded when loading up a diagram through hash


a20chrka:
This is what it looks like:
![image](https://user-images.githubusercontent.com/102599740/168235239-2346ba39-b0f5-4f48-b4f9-8a6b663e9da2.png)

This is caused by the last few letters of the URL (&embed):
![image](https://user-images.githubusercontent.com/102599740/168235448-a1e0ac51-67e2-464a-af86-e928684b1fe2.png)

Make the URL load without "&embed" and the purple banner should appear.

PR #12532, opened 2022-05-16
Changed redirect when loading dugga via hash so nav-bar still shows #12495

a20micro:
#12495 

Removed &embed from header in index.php which determined url. Now simply redirects to page normally.


a20chrka:
Test passed, purple banner now appears when loading the diagram through hashing.

a20chrka:
Need improvement. I can't even open "Edit Variants" on Diagram Dugga. 
Given error code:
image

Issue #12520, opened 2022-05-13
Delay on tooltip.


a20jacmi:
When hovering over the toolbar a tooltip is shown instantly which is disturbing. A delay is quite common to solve this and could be set to avoid this odd behaviour.

b20matad:
Added transition-delay to "tooltiptext" in diagram.css

PR #12534, opened 2022-05-16
Delay for tooltip instructions. Issue #12520.

a20jacmi:
Hover over some tool for more than 1200 ms to make the tooltip visible. Before this implementation the tooltip popped up instantly.


a20chrka:
Need improvement, feels a bit too slow. Decrease the time for it to pop up and it should be fine. Try 600 to 800 ms.

a20chrka:
Test passed after fix.

PR #12544, opened 2022-05-16
Added highlighting to edited elements on the sectioned site.

c20adael:
This can be tested by going to the sectioned page and pressing the cogwheel. Then modify the name and press the save button. DO NOT press enter to save because there is something wrong with that save method.
image


a20vicad (changes requested):
Page seems to update on some changes and does not "flash" red when this happens

a20vicad (commented):
The changes look good, now it works.

PR #12550, opened 2022-05-16
Change the height

a19xinhu:
Change the height to make the number is aligned to code.

1. Go into a example at course Webbprogramming.
2. Check the code is aligned to the number.


a20joabo:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12552, opened 2022-05-16
remoed extra bottom margin on plus button

a20sarle:
Testing:

1. Go to any course 
2. Log in as stei (password: password)
3. Open inspector
4. Click the "select element tool" in inspector
image
5. Hover the yellow plus

Before it looked like this:
image

On this branch it should look like this:
image


a20joabo:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12524, opened 2022-05-13
Made the navheader table-layout fixed #12287

a20linth:
Gave navheader table a fixed table-layout to prevent the navbar to go outside the window. Quite a few problems occured when doing that shift that had to be taken care of.

Testing:
1) Enter Webbprogrammering
2) Open JavaScript Example 1
The navbar should not break when resizing the window width.
3) Login as teacher acount stei
The navbar should not break when resizing the window width.

Issue #12412, opened 2022-05-11
Icons in mobile view


b20sebka:
![image](https://user-images.githubusercontent.com/102599780/167850217-d3c6bb07-1b3e-48b7-a7ba-2c99ee10f22f.png)
Using a browsers mobile view option in inspect element you can see that many icons are incredibly displaced. Picture shows a samsung galaxy s8+

PR #12555, opened 2022-05-16
Fixed course lable icons #12412

a20linth:
Course lable icons and text adjust to screen width to fit the content on the screen (icons in the nav header are fixed in pull request #12524)

Testing:
1) Enter Testing-Course (Demo-course does not work on this branch)
2) Resize the screen width
The course label should not break when resizing the window width.
3) Login as teacher acount stei and resize the screen width
The navbar should not break when resizing the window width.


a20sarle:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #12252, opened 2022-05-06
Possibility to add name/description to dugga variants


b20thelo:
Add a new field under "Dugga edit" where all properties for the dugga are set. The teacher can then add a short text(restrict the length to maybe 100-200 characters) that describes the dugga variant(should be available for all dugga types, not just diagram). 
The text input should be restricted in length and it should be shown in a new column in the variant list section:
![image](https://user-images.githubusercontent.com/81772705/167212614-f3893cd4-1d35-4ff5-8ea2-772fc5110163.png)

Purpose of feature: 
- If the teacher started to create a diagram version and wants to close down but not discard the changes they can save the variant with a comment and come back later to finish it.
- The teacher can write a comment to themselves to remember what the variant includes. 

This field will also be good during the project as we can create different tests for the diagram and name them with different test names to remember what the test is for.
<hr>
Asked Markus if this feature would be liked by teachers and he told me to do ahead and create an issue for it.

b19karhj:
William is added to the team. This is not a one man job.

Issue #52, opened 2014-04-04
missing buttons in codeviewer


e12johka:
when entering first examle in every section, the "skip backward" button is lost.

PR #12558, opened 2022-05-16
 Possibility to add name/description to dugga variants #12252 

b19karhj:


a20willa:
For testing:
1. Go into demo course
2. Edit any file
3. Add a note in the note field
4. Create and or update the dugga
5. Check to see that the note appears in the left column

b19karhj:
Seems that the server had to be updated. It now works

a20chrka:
Test failed. Too many letters in "Note" causes a few problems...
image

I can no longer access variant #52 because the "Note" pushed both the cogwheel and the "delete" button off screen.

b19karhj:
Now the note colum has a max width.  
bild

Issue #12533, opened 2022-05-16
Clean up showDugga.php


a20chrka:
A lot of commented code, remove it if its not needed.

PR #12559, opened 2022-05-16
Cleaned up  showDugga.php #12533

a20micro:
Cleaned up messy code, old comments and non-descriptive comments while adding new comments in showDugga.php #12533


a20chrka:
Test passed, nothing seems to break after clean-up.

Issue #12503, opened 2022-05-13
Users should not be able to login before being accepted


a20vicad:
Git_Users should not be able to login to contribution during the time between creating an account and getting it accepted. I.e when their account_status is 101(pending) or 102(denied)
Related to #12404

PR #12561, opened 2022-05-16
Added block from logging in when git_user is pending or revoked

a20sacka:
Added so you are blocked from logging in when you are pending or revoked.
relates to issue https://github.com/HGustavs/LenaSYS/issues/12503

when status_Account is 101
image

when status_Account is 102
image

------------------------------Testing----------------------------------
Make sure you cant login with a pending/revoked user, test by manually changing account_status to 101 and or 102 in git_user table


a20vicad (commented):
Acceptable solution

PR #12564, opened 2022-05-16
The buttons are now in order

a20mtafa:
The buttons are now in the order as the groupleader had requested. i did this by changing the positions of some code blocks for the buttons. it took a while because they were wierdly placed


a20mtafa:
For testing:
1. log in Username: stei Password: password
2. go to webbprogrammering
now the buttons should be in the right order

a20sarle:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #12522, opened 2022-05-13
The "cogwheel" buttons in sections can not be "tabed" to


a20vicad:
![image](https://user-images.githubusercontent.com/102599327/168296529-77fd7489-b43e-42e9-a65e-0909bca2d357.png)
These are the cogwheels in question.

PR #12565, opened 2022-05-16
Did cog whell clickable #12522

a20novra:
COMMENT out an unnecessary callning. Som changes in dugga.js to click on cog wheel image when someone klick on enter btn.


a20vicad (commented):
Good, next thing to add an issue for is to directly tab into the popup

PR #12567, opened 2022-05-16
Fixed popover buttons not being clickable

b20jonlu:
Made the popover the same as the version used for dugga2. This fixes the issue of certain buttons being unclickable because the submitButtonTable was over them.


a20vicad (commented):
I think these where all the duggas in need of this change.

PR #12569, opened 2022-05-16
Fixed problem with the resetButton in mobileview 

b20idawa:
Fixed problem with the resetButton in mobileview when touch the element


a20sarle:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #12545, opened 2022-05-16
Lines to/from inheritance triangle.


a20jacmi:
The lines looks weird, in some positions, when attached to an inheritance triangle. A suggested solution would be to attach them to the center of the triangle or close to it. The goal is to avoid a gap between the element and the line.

PR #12570, opened 2022-05-16
Fixed lines for UMLRelations #12545

b20matad:
Lines now connect to the center of the inheritance element. 
To test: Draw lines to and from inheritance; zoom in and out; check that there is no whitespace between line and element.


i19matla:
Works fine after merge

Issue #12358, opened 2022-05-10
ER-Table generation: Creation of weak keys.


c20emili:
**Main issue** #11744
The ER table need to be configured for weak entitys by creating composite primary keys by taking parts from the entity/entitys they're weak towards.

c20emili:
Merged in master

PR #12573, opened 2022-05-17
Implemented creation of weak keys.

a20gabpa:
Related to issue #12358 

What should be tested:

- Create a diagram that has a weak entity conected to strong / normal entity through a weak relation. Remember also to check that the line's cardinality is correct (1 on strong / normal side and N on weak side). After open the ER-table. In there the weak entity should have a underlined keys that consists of its of own weak key and a foreign key from the strong / normal entity.
- Try with different cases like when a weak entity is weak to another weak entity and also if the entity has two identifying relation to two strong / normal entities.
- Test if the program breaks if an random element is removed while the ER - table is open.

Note, if some odd behaviour is found, try to click on every element in the faulty relation and click on save. It seems that some elements does not save the corect values when a new relation is created. If this does not fix it, then some fault was found.


a20chrka:
Test failed, this is what the diagram looks like:
image

This is what the ER-table looks like:
image

There is no weak key.

Issue #12574, opened 2022-05-17
Color selector has different titles when selecting one/multiple objects


a20samul:
The color selector has different titles when selecting one- vs multiple objects. Change so that both titles say _"Color"_ to give it a more consistent look.

<img width="573" alt="bg" src="https://user-images.githubusercontent.com/81613484/168741620-7021622e-6bd0-4d8f-8108-8109c85ffa8b.png">

PR #12575, opened 2022-05-17
Changed the title for a more consistent look #12574

a20samul:
- #12574 
- The color selector had different titles when selecting one- vs multiple objects. Changed so that both titles now say "Color" to give it a more consistent look.

### To test:
1. Go to diagram dugga
2. Select one object 
3. Open the option panel, make sure the color alternatives title says _"Color"_.
4. Now select multiple objects and open the option panel, , make sure the color alternatives title says _"Color"_ instead of _"BG Color"_.

should **not** look like this:
image

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -5068,7 +5068,7 @@ function generateContextProperties()
               menuSet[i].classList.remove('options-fieldset-show');  
           }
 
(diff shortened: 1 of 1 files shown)


a20chrka:
Test passed, looks good.

i19matla:
Works fine after merge

Issue #12549, opened 2022-05-16
Rename UML-Entity to "UML Class"


a20chrka:
This is what the button for UML-entity says:
![image](https://user-images.githubusercontent.com/102599740/168554635-32fc981a-0c1e-4f99-ae6a-d7760c37cb64.png)

While the element itself says "Class":
![image](https://user-images.githubusercontent.com/102599740/168554930-506716f2-4c1c-4161-b9e9-b5f7adab9b5a.png)

Rename the name of the button and the tooltip so it represents what it actually is (and matches the element).

PR #12581, opened 2022-05-17
Rename UML-Entity to "UML Class" #12549

b20thelo:
#12549

**What has been done**
Updated so that it says UML class instead of UML entity

**How to test**
Just hover over the entity/class button and check that the tooltip text is correct for the UML class.


a20chrka:
Test passed, looks good.

Issue #12584, opened 2022-05-17
Tooltip missing for the option panel


a20samul:
Tooltip keybinding: 'O' missing from option panel, fix this:

<img width="176" alt="tooltip" src="https://user-images.githubusercontent.com/81613484/168759016-04f7c87c-bde7-43c2-87fc-73ca66a38323.png">

PR #12585, opened 2022-05-17
Fixed so that the keybinding is visible in the tooltip #12584

a20samul:
- #12584
- Fixed so that the keybinding: "O" shows in the tooltip for the options panel

### Before:

optionbefore

### After: 

option after


a20chrka:
Test passed. Keybind appears.

Issue #12537, opened 2022-05-16
Massive bug when removing attributes/functions from an UML-entity.


a20chrka:
This is what it looks like:
![image](https://user-images.githubusercontent.com/102599740/168544852-04015342-e059-4a69-8bda-0d4123cfe5f9.png)

To reach this error-state, do the following:
1. Place a new UML-entity in the diagram.
2. Remove all attributes and/or all functions in the UML-entity.
3. Click "Save" on the UML-entity
4. Move the diagram or zoom in/out. It should look like the image above.

Note that you can't stop moving the diagram afterwards.

a20gabpa:
Tip: At the moment, the UML has two lists, one for attributes and one for functions. The problem is that the UML class is built on the values in these lists, if these are empty, neither of the other boxes are drawn / created. The updateCSSForAllElements function tries to access its child but is encounter with undefined, and thus breaking the program. One possible way to fix the problem could be to handle NULL or empty list cases correctly.

PR #12591, opened 2022-05-17
UML-entity unchanged when removing attributes/functions. Issue #12537.

a20jacmi:
Create an Entity of UML type and remove all attributes/functions, and notice that there is no weird behaviour anymore.


a20chrka:
Test passed, looks good.

i19matla:
Seems to work fine

PR #12592, opened 2022-05-17
Important code lines is only cpoied

a20willa:
For testing: 
1. When logged in as a student
  1.1. Go into any code example
  1.2. Make sure it has some code
  1.3. Try to manually copy code - **You should not be able to select it**
  1.4. Make some lines of code "important" 
  1.5. Press the copy icon - **Only important code should be copied**

2. When logged in as a teacher
  1.1. Go into any code example
  1.2. Make sure it has some code
  1.3. Try to manually copy code - **You should be able to select it**
  1.4. Make some lines of code "important" 
  1.5. Press the copy icon - **All code should be copied**

### DIFF
--- a/DuggaSys/codeviewer.js
+++ b/DuggaSys/codeviewer.js
@@ -394,7 +394,11 @@ function returned(data)
 	// Add all drop zones
 	if (retData['writeaccess'] == "w") {
 		initFileDropZones();
-	}	
+	} else {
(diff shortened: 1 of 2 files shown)


a20joabo:
Code refactoring - ok
Code standard - ok
Meet issue goal - no

Not allowing student to mark text works, but the copying of important lines using the Copy button does not work.

PR #12594, opened 2022-05-17
Added press effect on copy button in mobile view 

b20idawa:
Added press effect on copy button in mobile view


b20idawa:
How to test the issue:

1. Select the webprogramming course
2. Select PHP Example 1
3. Select mobile view and iPhone 12 Pro for example
4. Press the copy button and check that the button has an effect when you press the button

a20joabo:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12597, opened 2022-05-17
G3 2022 w7 #000001

a19simru:
@a20vicad If you press the cog wheel next to a course on the main page (courseed.php) and mark visibility as deleted the course will now be scrubbed from memory and removed from the main page. All duggas, code examples, etc will also be scrubbed from memory if they belonged to that course. Because of the amount of information and work (including grades) that can be deleted very easily with three mousse clicks this seams kind of dangerous. Especially since every logged in user currently has editing privileges on courses. But the customers did specify that they wanted everything to be quick and easy for the user.

To test switch to my branch and  make a course in the main page then delete it whit the method above.

Warning I had a lot off memory leak problems when implementing this. Currently I "think" all problems are resolved. But be aware that this has been a big issue.

### DIFF
--- a/DuggaSys/courseedservice.php
+++ b/DuggaSys/courseedservice.php
@@ -530,11 +530,39 @@ foreach($queryz->fetchAll(PDO::FETCH_ASSOC) as $row){
 	$userCourse[$row['cid']] = $row['access'];
 }
 
-//Delete Courses that have been marked as deleted.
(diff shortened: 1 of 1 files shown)


a20vicad (commented):
In the end of the document it seems like you just changed indentation. Avoid this in the future

PR #12604, opened 2022-05-17
Fixed tab priority to go to the menu that is open 

b20jonlu:


a20vicad (commented):
This works just like I wanted it to. Very good and appreciated.

PR #12606, opened 2022-05-17
Added a fast install script that dumps the database and reinitializes it.

a20patpe:


a18gusjo:
For testing:
**OBS test locally only!
Be aware: Database will be dumped. Records insterted after installation will be removed.**

Step 1:
	Enter any course and create new code examples etc.
image

Step 2: 
	Open the install_light.php located at LenaSYS/install/ and run it.
image

Step 3: 
	Open the course you choose.
Step 4:  
	Everthing created after installation should now have been removed.

a20patpe:
If you hard drive is too slow, the php script might time out, to fix this change the max_execution_time variable in php.ini 
image
To something bigger than 120 seconds.

Issue #11604, opened 2022-04-13
Dugga, changes made by student should be saved in localstorage.


b19karhj:
When a student make changes in a dugga and managed to leave the dugga, The changes the student made should be saved in localstorage so the student can go back to the dugga with the changes remaning. There should also be a button where you can clearthe localstorage.

a17nikjo:
Storing and loading the information is no problem, the bigger problem seems to be what to store. The best solution would be able to store any dugga. Storing the html code of the document and loading it works fine. The problem lies with restoring js and php related things.

Storing input values might be a better approach, this is however hard since all duggas are unique and avoiding a hard coded per dugga solution is preferable. Researching if we can record user input and then replay them to restore previous instance that way.

PR #12607, opened 2022-05-17
Stored the current diagram in localstorage for issue #11604

c20emili:
- #11604 
Saved the current dugga in local storage so changes remain when returning.
added a button to refresh the diagram above redo
worked in diagram.js, diagram_dugga.js and duggaed.js.
For testing:
1 Change the diagram
2 Leave the page
3 Return (The diagram should be as you left it)
Pressing the refresh button or changing/deleting the variant clears the active diagram


a20chrka:
Test failed. The same diagram is still present even though its variant is deleted.
Also, it should be possible to do a hard-refresh to restore the diagram, it will get old real fast when you have to make a new variant everytime you test something and need a clean slate.

c20emili:
> Test failed. The same diagram is still present even though its variant is deleted. Also, it should be possible to do a hard-refresh to restore the diagram, it will get old real fast when you have to make a new variant everytime you test something and need a clean slate.

A hard-refresh cant clear localstorage without defeating the point of localstorage wich is that it saves on the device so the site can be closed and reopened and the values are saved. Would a button or a keybinding to refresh the diagram suffice?
@a20chrka

Issue #12539, opened 2022-05-16
Add a new color that is completely black for elements in the diagram.


a20chrka:
The color should be added among the other colors:
![image](https://user-images.githubusercontent.com/102599740/168547216-31528911-65c6-477a-887c-e32b28e336b7.png)

Since the color will be black, the font/text color should be white.

a20samul:
Note that a white text should also be implemented on the dark pink colour #DC267F according to material design.

PR #12611, opened 2022-05-17
added a new black color to the circle issue #12539

a19meram:


a20chrka:
Test failed. Underlines for weak keys does not change to white:
image
Same attribute but black:
image

i19matla:
Works fine

Issue #12601, opened 2022-05-17
Tabbed items with purple background have tabs with white background


a20linth:
When an item in sectioned is tabbed it is given tabs that have white/black backgrounds, this is also the case for items that should have purple backgrounds. 
![image](https://user-images.githubusercontent.com/102609651/168784476-d1b7cb9f-aa65-4e43-93ef-ee5b3bb5fbbd.png)
![image](https://user-images.githubusercontent.com/102609651/168784585-79b00e37-b1af-4118-b743-ae2bbd9b92ff.png)

This issue is made more  apparent by #12598.
![image](https://user-images.githubusercontent.com/102609651/168785569-983152cd-3b45-4281-9ac7-ffb9a913578e.png)

PR #12614, opened 2022-05-17
Corrected tab elements class list #12601

a20linth:
Corrected tab elements class list so that the correct styling comes from css, removed hardcoded styling.

Testing:
2) Enter Webbprogrammering
3) Create all item types, give random tabs to the new items
Before:
image
After:
image


a20joabo:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12620, opened 2022-05-17
Added new icons to tabbable list

b20jonlu:
Added the new icons to the tabbable list in order to make sure they are disabled when a menu is open.


b20hanwa (approved):
Works as intended when tested!

Issue #12632, opened 2022-05-18
The "pop-out" button functionality has dissapeared.


a20chrka:
For UML and ER-entity, buttons are now placed separately, instead of being "popped out" when holding down the mouse on "Entity (same goes for ER-relation and UML-inheritance):
![image](https://user-images.githubusercontent.com/102599740/169007599-8457ea1e-4253-4ad9-aa17-a25b3e01ff39.png)

Bring back the "pop-out".

Issue #12141, opened 2022-05-04
Dugga variant preview doesn't work as it should


b20thelo:
When you create a diagram dugga variant and you want to preview it to see what it would look like for the student, it doesn't work as it should.

Preview button found here
![image](https://user-images.githubusercontent.com/81772705/166806633-8ca1b4fa-a497-49d0-a067-ec9baf973e46.png)

A few things that need to be looked at are:
- Even though the diagram variant in this example was set to only use ER (UML disabled) both functionalities are available in the preview (only in preview, not when you access the dugga as a student). It doesn't read the parameters correctly but instead show teachermode as the teacher is logged in.
- The hide/show button for the description is misplaced (only in preview, not when you access the dugga as a student). As a result this functionality doesn't work as it should there.
![image](https://user-images.githubusercontent.com/81772705/166807475-3bdae18e-df70-4b83-a4cc-7af5b56ba3ae.png)

The preview functionality should instead show the parameters generated in the param section. Parameters are shown here in the variant list:
![image](https://user-images.githubusercontent.com/81772705/166807008-1062c70c-433c-4cbf-a189-ffd7b54588f0.png)

b20thelo:
The "hide/show button" was fixed in another issue. Not merged yet

a20samul:
### Preview issue with the arrow fixed. refer to:
- #12408

<img width="678" alt="preview" src="https://user-images.githubusercontent.com/81613484/168063202-91dffbd8-a84d-43ed-8ae5-2865893a45cb.png">

c20emili:
Part of the reason why the preview does not work is that the functionality that fetches the diagram and instructions files lays in showdugga.php rather than diagram.php. To solve this they need to be moved and the variabels $cid, $vers and $quizid needs to be fetched differently. A different part of why the preview does not work is that it does not execute the template(diagram_dugga).js file for which i don't know a solution right now.

c20emili:
There is a solution for the issue but since it needs to move a significant part of the code that is in a wrong file (diagram specifik code that is in a global file for all duggas), the solution will have to wait after merge to ensure no one is changing the code that will be moved during the move.

PR #12642, opened 2022-05-18
Fix for issue #12141, added a missing semicolon that fixes issue #12632 and updated install file to fix issue #12659

c20emili:
#12141  #12642 #12659 
Moved the diagram specific functions in showdugga.php to diagram.php, stored the variables in the url in javascript variables to be used in diagram_dugga.html for both duggaed.php and showdugga.php, moved where the diagram.js and diagram_dugga.js finds the moved functions ,updated the showfacit in  diagram_dugga.js and added so the getVariantPreview fetches the duggaed.js after the templete.js because they got functions with the same name.Added a missing semicolon in showduggaservice.php. updated the install file.
To test:
Open diagram dugga and see that it works as after the latest merge.
Log in as toddler 
Go to show tests(paper in the middle of the navbar)
Select diagram dugga by clicking the pen
Press the play button on the variant you want to preview(if there is none create one)
The preview should work as if you are a student with that variant.
Check the funcionality of the preview dugga is everything as the variant states


a20chrka:
Test passed, seems to work fine.

Issue #12610, opened 2022-05-17
The dropdowns for filelink and gfilelink does not change to the selected variant when changed before.


c20emili:
When selecting a available variant the dropdowns for filelink and gFilelink does not update properly. The options are not correct evven when the type is correct(se figure 1).
![image](https://user-images.githubusercontent.com/81320323/168805307-2323c899-1cb2-4dca-aa61-da5aea8c9cb1.png)

PR #12643, opened 2022-05-18
The dropdowns for filelink and gfilelink does not change to the selected variant when changed before. #12610 

b19karhj:
Put updateInstuction and Updateinfromation in setVariant to get info from left field. It now works as intended.


a20chrka:
Test passed.

Issue #12633, opened 2022-05-18
Error-check: Marks attributes connected to relations.


a20chrka:
This is what it looks like (standard diagram):
![image](https://user-images.githubusercontent.com/102599740/169016322-4a65b62f-3cc2-46e3-a53a-d444b5eba762.png)

They should not be marked with red (when error-check is enabled) since this is allowed according to ER.

PR #12647, opened 2022-05-18
Attribute on relation no longer results in error

a20edwth:
Attribute on relation no longer results in error. #12633 

Instructions for testing:
1) Click "Demo-Course"
2) Click "Diagram Dugga"
3) Click "Toggle error check"
4) Observe if attributes on relations are outlined in red


a20chrka:
Test passed, attributes connected to relations are no longer marked with red.

i19matla:
Works fine

PR #12652, opened 2022-05-18
made code lines inherit background color from .box

a20sarle:
Code lines should inherit color from .box since the background should follow if the background color changes. Solved by setting background-color: transparent and through removad or redundant css code lines.

Testing (repeat the same teps for light and dark theme):
2. Open inspect
3. Click the select tool
image
4. Select a code line 
image
5. Find element .box in the css style
image
6. Change the background color to any color

Before: not all elements were affected by the color change.
image

After: every codeviewer element (description and code)  on the page shuold get the the same background color.
image

### DIFF
--- a/Shared/css/blackTheme.css
+++ b/Shared/css/blackTheme.css
@@ -114,7 +114,7 @@
     font-family: Arial,Sans-Serif;
     font-size: 14px;
     outline: 0 none;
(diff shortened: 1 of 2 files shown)


a18gusjo:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12657, opened 2022-05-18
Fixed colors of icons in both lightmode and darkmode

b20idawa:
Fixed colors of icons in both lightmode and darkmode


a18gusjo:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes
I will have to check with the group management before I give this a pass because courseed does not work as expected.
image
This is very wierd since the the files changed shouldn't affect courseed.
@a20muhal

a20mtafa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

There has been problems before with this. test it before you merge it

PR #12661, opened 2022-05-18
You can now copy only important code

a20willa:
Lines are now copied using another function instead, as the old version did not work well when selecting specific elements in a class.


a18gusjo:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12662, opened 2022-05-18
made the code lines numbers background inherit from .codeborder

a20sarle:
Made the code lines numbers background inherit from `.codeborder` through removing redundant css line.

Testing (follow the same steps both for light and dark mode:
2. Click any of the code examples
3. Open inspect and click the select tool
image
4. Select the code bar 
image
5. Find the class `codeborder` in inspect
image
6. Change color of class codeborder to any color

Before:
- before changing color
image

- after changing color
image

After: All lines should get the color chosen except the ones that were marked with another color since before
- before changing color
image

- after changing color
image

### DIFF
--- a/Shared/css/codeviewer.css
+++ b/Shared/css/codeviewer.css
@@ -217,7 +217,6 @@
     padding-left: 4px;
     padding-right: 6px;
     width: 22px;
-	background-color: #614875;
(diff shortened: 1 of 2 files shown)


a18gusjo:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #12638, opened 2022-05-18
Fix certain variants for diagram dugga not being deletable


a20micro:
Certain variants currently can't be deleted because of SQL related issues. Research the cause of this issue and make it so teachers can delete all variants.

![CantDelete](https://user-images.githubusercontent.com/95532115/169019070-1822df9e-0186-45fa-abb0-55f3befd283f.png)

b20jonlu:
This issue is probably related to this one https://github.com/HGustavs/LenaSYS/issues/12617

a20micro:
should potentially be solved by deleting all rows in useranswer which uses the corresponding variant ID as a foreign key. This happens in duggaedservice.php 

![MbyErrorSolution](https://user-images.githubusercontent.com/95532115/169042607-3b96cd4b-7483-4233-9703-8ad49ed2945c.png)

PR #12663, opened 2022-05-18
It's now possible to delete variants with assigned user answers #12638

a20micro:
Made it so deleting a variant also deletes all stored user answers that correlates to said variant.


i19matla:
Works fine after merge

Issue #12639, opened 2022-05-18
Dem-course items start with 2 tabs


b20sebka:
In demo-course items start with 2 tabs

PR #12666, opened 2022-05-18
testdata of listentries are created with no tabs #12639

a20linth:
Demo-Course and Webbutveckling - datorgrafik testdata is created witth 0 tabs instead of 2.

Testing:
1) Run install script (install/install.php)
2) Enter Demo-course or Webbutveckling - datorgrafik
3) The initial listentires should start with 0 tabs (previously started with 2 tabs).
Expected result:
image


a18gusjo:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #12655, opened 2022-05-18
Remove double drag-button


a20caros:
<img width="760" alt="image" src="https://user-images.githubusercontent.com/81676918/169036898-9a185ce2-8b70-4734-9e9a-31c0ab5374c1.png">

PR #12675, opened 2022-05-18
removed image tag #12655

a20novra:


b20hanwa (changes requested):
Code should not be commented out. Only remove or change code.

b20hanwa (approved):
Now the code looks better!

Issue #12644, opened 2022-05-18
Notes is not cleared when deselecting row


b19karhj:
When you unmark the row in varient, notes do not clear from field.
![bild](https://user-images.githubusercontent.com/81624541/169029215-a1a5c7d8-82e6-4e7a-8d6d-3e50e71d1f48.png)

a20chrka:
Test passed, works fine.

PR #12681, opened 2022-05-18
Notes is not cleared when deselecting row #12644 

b19karhj:
If nothing is selected in variant. Note-box is cleared


a20chrka:
Test passed, works fine.

PR #12690, opened 2022-05-18
Potential patch for problem with ER table

a20gabpa:
#12358 
Apperently there is a problem where the ER-table generation does not work on newly created entities.

What has been done:
- Updated the default for entities, relations and attributes by adding a normal state.

What can be tested:

- Place new entities and new attributes and see if they appear in the table. 

At the moment:

image

How it should look:

image

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1043,9 +1043,9 @@ var ghostLine = null;
  */
 var defaults = {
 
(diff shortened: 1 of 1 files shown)


a20chrka:
Test passed, ER-Table entry appears for new entities.

i19matla:
Works as description tells, need to be followed up so it works with the initial loaded data to.

Issue #12546, opened 2022-05-16
UML-Entities saves everytime a function/attribute is added to their lists.


a20chrka:
This is what it looks like when I rapidly add new attributes:
![image](https://user-images.githubusercontent.com/102599740/168552289-6c44070e-e30e-4f7f-8188-df3e61e22e89.png)

Note that the UML-entity is not saved when attributes/functions are removed. 
The UML-entity should only be saved when "Save" is pressed.

PR #12691, opened 2022-05-18
G1-2022-w8-#12546: UML-Entities saves everytime a function/attribute is added to their lists.

a20jonda:
#12546
Enter:ing not will not save in textarea


a20micro:
Test passed on remote, no longer shows the save pop-up when adding new row in UML element.

i19matla:
works

PR #12698, opened 2022-05-19
Made it so loading diagram from hash doesn't remove nav-bar

a20micro:
removed &embed attribute from redirect link


a20chrka:
Test passed, banner appears again.

Issue #12162, opened 2022-05-05
ER-Table generation: Consider multi-valued attributes.


a20chrka:
Main issue: #11744 

Consider the following entity:
![image](https://user-images.githubusercontent.com/102599740/166885792-b59f6a4b-b57d-4d77-8bc4-6f3677bfbcbf.png)

The entry for this entity and its multi-valued attribute should look something like this:
![image](https://user-images.githubusercontent.com/102599740/166887085-5cae3bce-ff65-41ec-a632-89f373bb56c4.png)

PR #12701, opened 2022-05-19
Added multi-valued attributes for Issue #12162

c20emili:
Added multi-valued attributes in the ER table. #12162 
Worked in diagram.js
For testing:
1. Have a diagram with an ER entity that has a multiple attribute.
2. Select the ER entity
3. Open the options panel
4. Press E or the ER-Table button
5. look at the table at the bottom there should be the multi-valued attributes in the form of Attribute(EntityKeys,Attribute)

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -4784,7 +4784,7 @@ function generateErTableString()
             }
         }
     }
-    //Just for testing
+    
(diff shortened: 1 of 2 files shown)


a20chrka:
Test passed

i19matla:
Works fine

Issue #12649, opened 2022-05-18
The refresh diagram button needs a new icon and maybe a new tooltip/name


c20emili:
The current icon for the refresh diagram button(top button in history) is the same as the button for ER table and it needs it's own icon. The tooltip and name may need an update since they were rushed but discuss with manegement about that before changes.
![image](https://user-images.githubusercontent.com/81320323/169031625-6c3616b5-be8d-4ee8-8673-5c2997bd227f.png)

PR #12703, opened 2022-05-19
 The refresh diagram button needs a new icon and maybe a new tooltip/name #12649 

b19karhj:
New button/name/tooltip added for diagram reset.
bild


a20micro:
Test passed on remote. Button exists and does what it's supposed to.

PR #12706, opened 2022-05-19
removed unnecessary css code to make #NewTabLink follow light and dar…

a20sarle:
Simplipied the complexity by reusing css code and changing the color set in the original svg file.

Testing:
1. Log in as teacher (stei: password)
2. Go to course Webbprogrammering
3. Check that the link-icon is purple in light mode ( and that it's continously white / light gray-ish in dark mode)

Before (in light mode)
image
After (in light mode)
image

Should look like this in dark mode
image

### DIFF
--- a/Shared/css/blackTheme.css
+++ b/Shared/css/blackTheme.css
@@ -43,25 +43,11 @@
         --color-red-hover: #ce3d3d;
     }
 
-    #dorf {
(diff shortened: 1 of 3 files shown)


a20mtafa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #12705, opened 2022-05-19
Diagram: If no diagram file is not set, the reset button does not reset to default state.


b19karhj:
If no diagram file is not set, the reset button does not reset to default state.

PR #12716, opened 2022-05-19
Diagram: If no diagram file is not set, the reset button does not reset to default state #12705

b19karhj:
Reset will now work when no diagram is set in variant


a20micro:
Test passed on remote. Refresh button in toolbar properly resets the diagram to the initial diagram.

i19matla:
Works on weekbranch to

PR #12717, opened 2022-05-19
Fixed so that editing in mobile view works

a20willa:
For testing:
1. Use a mobile phone
2. Login as a teacher
3. Go into a code example
4. Press the edit button 
5. Go into landscape mode


a20mtafa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12721, opened 2022-05-19
Diagram: Add confrim and localstorage clear to orginal reset button in lower left corner.

b19karhj:
Outer reset button now have confirm and also clear localstorage


b19karhj:
To test. Change position of an element, press lower left reset button (you have to scroll down as its outside the diagram) It should now give a confirm option and clear local storage.

a20chrka:
Test passed.

Issue #12673, opened 2022-05-18
add multiple items when creating items


b20sebka:
Customer wants a functionality to add multiple items at once in as few clicks as possible

![image](https://user-images.githubusercontent.com/102599780/169051252-5cfdf42c-9e6c-4e54-9d1d-d6ff878c6117.png)
Customers example was when you press the button 5 times and then add a heading it should add 5 headings

PR #12722, opened 2022-05-19
Added possibility to create multiple items at the same time #12673

a20linth:
Clicking the bottom right plus icon will increase the number of items to be created, when an item type is clicked. Default number of items to be created is 1.

Testing:
2) Enter Demo-Course
3) Click on the plus sign icon in the bottom right 7 times
4) Create a new code example with the bottom right button.
5) 8 new code examples should be created and console output should say:
image
6) Create a new test dugga
7) only one new test dugga should be created and output should say:
image


a18gusjo:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #12578, opened 2022-05-17
Undo deletion button is not working


b20sebka:
https://user-images.githubusercontent.com/102599780/168745469-1e5e9d72-1e60-43c8-852e-944f710e1933.mp4

Undo deleting button is not working

PR #12726, opened 2022-05-19
Made the undo delete button work again. #12578

a20andlj:
Fixed so the undo delete example button works again.
Also fixed the position of the button  so it sticks to the + ghost and eye buttons.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -699,7 +699,7 @@ function deleteAll()
 
 // Cancel deletion
 function cancelDelete() {
-  clearTimeout(time);
+  clearTimeout(delTimer);
   var deletedElements = document.querySelectorAll(".deleted")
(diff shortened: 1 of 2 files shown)


a20andlj:
Test1:
2. Create a new code example
3. Delete the example
4. Press Yes in the confirmbox
5. Press the recycle button to se if the deleted item returns.
6. If so - test ok

Test2:
1. Delete one of the examples
2. Press Yes in the confirmbox
3. Wait for 60 sec until the recycle button disappears - then refresh the page. If the deleted item does not return. Test ok

Test3:
1. Delete another example
2. Press Yes in the confirmbox
3. Wait for 60 sec until the recycle button disappears
4. Create a new example - if only one new example appears the test is ok.

a20mtafa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12728, opened 2022-05-19
Changed from static colors to root

a20sarle:
Changed static color to root. Should be tested through version testing (check with group leader).

Did not go through the whole style.css file as it was too unclear where the different sections (duggasys, contribution and diagram) started and ended.

### DIFF
--- a/Shared/css/blackTheme.css
+++ b/Shared/css/blackTheme.css
@@ -580,7 +580,7 @@ body {
 
 /* Style for "pane-container text" START */
 .pane{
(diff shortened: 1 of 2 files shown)


a20sarle:
Needed to add / change some things before merge to avoide collide.

a20sarle:
Needed to add / change some things before merge to avoide collide.

a20sarle:
Something happened with the branch. Creating a new one to avoid unwanted collides.

PR #12731, opened 2022-05-19
Mobile view for larger mobile devices.

a20patpe:
Larger mobile devices such as an ipad är will now be in mobile view.
Added a clease where the orientation is portrait.
(portrait meaning the widht(px) of the screen is smaller than the height(px) of the screen.)

For testing:
1: Press ctrl+shift+i
2: Press ctrl+shift+m
(might be different for apple users)
3:Change your view to the ipad air(Or any large device where the widht of the screen is smaller than the height.).
image
4: Enter an example.
image
5: The example should be shown in mobile view(Where the boxes are all stacked ontop of eachother).

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -892,7 +892,7 @@ p {
   font-size: 10px;
 }
 
-@media only screen and (max-width: 800px) {
+@media only screen and (max-width: 800px), only screen and (orientation:portrait) {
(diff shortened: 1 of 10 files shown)


a18gusjo:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12732, opened 2022-05-19
Fixed the primary color in ddark mode

a20willa:
For testing:
1. Switch between dark and light mode and check that the purpple color is not the same


a18gusjo:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12737, opened 2022-05-19
Created new SVG dugga

b20jonlu:
Created the new SVG dugga to look like the one in the mockup, also added some things to the backend in order to have a foundation to work on. Most of the javascript and html is a combination of parts from the other duggas and the code in https://github.com/HGustavs/SVG-Demo

! Don't forget to run install.php before testing.

Landing page view
SVG_dugga

It is possible to create svg objects by writing code in the right side editor
example_svg_drawing

Dropdown menu to select a different object
svg_dropdown

This is just a backbone that can be used to build out the functionality further. One of the things missing is save and load functionality, as of now every object in the dropdown is coded into the javascript file and it is not possible to save the object.


a20vicad (commented):
Looks good, based on the provided SVG-demo code

PR #12740, opened 2022-05-19
Child attributes of primary/candidate key are now underlined as one key object

a16linas:
If an attribute that's a primary or candidate key has child attributes, they're now printed as one underlined key object

bild

You can test this by having a primary or candidate attribute have child attributes, they should then be printed in the ER table as one key object. And if a primary and candidate exist, the candidate and its child attributes will not be printed at all


i19matla:
Works fine

Issue #12669, opened 2022-05-18
Refresh screen breaks UML-Class


b20matad:
After refreshing the page, the functions jump up as attributes in the UML-class. This is only the case after altering the default text in the boxes. (See the video)

https://user-images.githubusercontent.com/102598238/169048491-46163530-ab2e-4337-8377-296984189843.mp4

b20matad:
Problem solved, functions were overwriting attributes in the method "saveProperties()", which updates the statemachine. These changes were made in the file "diagram.js".

PR #12741, opened 2022-05-19
Refresh screen breaks UML-Class #12669

a20jacmi:
This was probably just a typo that created a lot of problems. To test this you should create a UML-Class; add some attribute and functions; use the undo and redo tool; save and load a diagram, with and without history; refresh the page. If the UML-Class remain the same without any odd behaviour your golden.


a20chrka:
Test passed.

i19matla:
Works great

PR #12743, opened 2022-05-19
Changed color from purple to gray

b20fanfa:
Needed to add new color on two "color" variables: --color-text-primary and --color-primary-2.


b20fanfa:
### For testing:

1. Log in as teacher. 
Should look like this: 
Skärmavbild 2022-05-19 kl  16 07 26

Issue #12710, opened 2022-05-19
Diagram: Variant: Disable, update and enable,


b19karhj:
when nothing is selected the buttons still can be pressed.

![bild](https://user-images.githubusercontent.com/81624541/169250814-acf1638c-8c4c-4c63-bef5-cc97c7859d06.png)

PR #12744, opened 2022-05-19
 Diagram: Variant: Disable, update and enable, #12710 

b19karhj:
Test:
Go to show test, click the pen on diagram dugga. Check if update/disable/enable buttons are disabled.
Click one of the cogweheels for a diagram, check if disabled or enable are disabled depending on the status.

Click update or enable / or disable.

the buttons should show as disabled again.


a20chrka:
Test passed, buttons are now greyed out when no variant is being edited.

Issue #12724, opened 2022-05-19
Error check: Primary keys are not recognized as a key


a20edwth:
Primary keys are not recognized as a key, and will result in a error for the entity.

Entity with a candidate key (recognized as a key):
![40f1dd303cfd427656d790d400a90808](https://user-images.githubusercontent.com/102611046/169276482-2a6a2461-99c5-4793-8d44-2919273b99e8.png)

Entity with a primary key (not recognized as a key):
![83df16a0e186e6f8219a3a03d652517e](https://user-images.githubusercontent.com/102611046/169276949-4de4ce13-a5eb-4fbe-9660-c5dbafb665b6.png)

It should also be possible for a Entity to have more than one candidate key, which currently results in error:
![a2a43981bf3ee20199e92fe1ada75def](https://user-images.githubusercontent.com/102611046/169277261-12995d27-10ca-47b0-a33e-6d657fd06a42.png)

a20edwth:
Primary keys are recognized and there can be more than one candidate key.

Instructions for testing:
1) Click "Demo-Course"
2) Click "Diagram Dugga"
3) Change a candidate key to a primary key (possible to have candidate keys togheter with one primary key)
4) Create 2 candidate keys on the same entity

PR #12747, opened 2022-05-19
Primary keys are recognized

a20edwth:
Primary keys are recognized and there can be more than one candidate key. #12724

Instructions for testing:
1) Click "Demo-Course"
2) Click "Diagram Dugga"
3) Change a candidate key to a primary key (possible to have candidate keys togheter with one primary key)
4) Create 2 candidate keys on the same entity


a16linas:
Test passed

i19matla:
Works fine

Issue #12749, opened 2022-05-19
Diagram: Reset does not work (wrong function name in one place)


b19karhj:
The reset does not happen, one function name is incorrect.

PR #12750, opened 2022-05-19
Diagram: Reset does not work (wrong function name in one place) #12749 

b19karhj:
Quick fix (called an old function name (now updated)).

Test : Move element or add element, click the reset icon and after pressing ok, the diagram should now be in default position


i19matla:
Works fine

PR #12751, opened 2022-05-19
I have Probably fixed the problem with deleted elements being unhidden after moving another element.

a19simru:
I made it so all elements listed as deleted will be hidden once more after moving a element. Be aware that I have not been able to test this as I have discovered a weird bug on my machine that makes me unable to se the sectioned.php pages in localhost. I don't know why but it seams to date back several merges. I can se sectioned.php on branch G3-2022-#merge4. But whilst I have not been able to test my solution I believe it should work.

To test switch to my branch and try to run secioned.php as a superuser and delete a element. Then move another element and check that my element does not disappear.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -1277,12 +1277,12 @@ function returnedSection(data) {
           if (retdata['writeaccess']) {
           if (itemKind === 3) {
             if(isLoggedIn){
-              str += "<td  class='LightBox" + hideState + "'>";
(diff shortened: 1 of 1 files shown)


b20hanwa (commented):
Deleted items are now hidden when moving an element. I don't understand why deleted items should come back at all when you move an element though. I know that this has to do with being able to undo when an element is deleted. But is there a solution for not having deleted elements reapear when doing changes within 60 seconds?

PR #12759, opened 2022-05-20
added header and password is now hashed

a20willa:
For testing:
1. Go into this link: [localhost]/LenaSYS/install/install_light.php
2. The title should now be `Light LenaSYS Installer` and the password should be hashed


a20mtafa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12769, opened 2022-05-20
Change width to suit mobil view

a19xinhu:
1. Check you use a iPhone5 or use mobil mode iPhone at PC.
2. Compare the css between PC and mobil view. Is it different now.


a20mtafa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #12456, opened 2022-05-12
Instruction and information files currently needs to be manually written instead of selecting from a list


a16linas:
When editing variants on a dugga, to select information and instruction files, the teacher needs to manually write the filename instead of simply choosing from a list of available files.
![bild](https://user-images.githubusercontent.com/37794783/168031321-329f52e8-a271-42a3-b808-02645f27cc9b.png)

a16linas:
Issue is back at [merge 8](https://github.com/HGustavs/LenaSYS/tree/G1-2022-%23merge8)

PR #12770, opened 2022-05-20
Instruction and General information files are now visible

a16linas:
Issue #12456 was present again in merge 8, the files weren't loaded into the dropdown box.
Ajax call was only made if the dugga was of diagram type, now its done when opening the variant editor.

bild


a20micro:
Test passed on remote. Files now properly shows up without having to load a diagram beforehand.

i19matla:
Works fine

Issue #12603, opened 2022-05-17
Add IE-Inheritance


a20chrka:
See issue #12318 for more info on IE

Add IE-inheritance to the diagram:
![image](https://user-images.githubusercontent.com/102599740/168784901-5a4166a9-99ba-4bc8-b2d0-9955885634b6.png)

This element should pretty much have the same functionality as UML-Inheritance, but look differently of course:
![image](https://user-images.githubusercontent.com/102599740/168785376-d31243c4-ef00-44ad-8ee0-d72ac9fea8f6.png)

Just as UML-Inheritance, IE-Inheritance can be either overlapping or disjoint (see top image).

The button for IE-Inheritance should added to the "pop-out" that already has ER-Relation and UML-Inheritance:
![image](https://user-images.githubusercontent.com/102599740/168785727-93e07498-a811-47f3-ac5e-711ffe166a1c.png)

PR #12772, opened 2022-05-20
Added IE inheritance #12603

a20samul:
### This pull request contains:
- #12603 
- Added IE inheritence to toolbar and diagram 
- Created 2 icons: for IE inheritance and IE entity.

ie

> _Image 1: Demonstrates the changes made._

### How to test:
1. Press the half circle icon on the toolbar and place the IE inheritence on the diagram.
2. Press on the element and open the options panel
3. Change between the inheritance: overlapping/disjoint.
4. Try switching between elements: IE/ER/UML.


a20chrka:
Fixed in another pull-request

PR #12787, opened 2022-05-20
Change the background in "Diagram_Dugga"

b20fanfa:
The background has a darker background.


a20mtafa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12788, opened 2022-05-20
Added save and load functionality to SVG dugga

b20jonlu:
Added the save button functionality to store the content from the editors into the userAnswer table and also for that data to be retrieved later by a superUser.
Also changed how showDuggaservice loads data from userAnswer. previously it always used moment as a key but now if the hash matches it will use that instead. This should work better with load dugga functionality. Had to add unset($grade) because it prevented submissions since grade was "UNK" by default which isset and greater than 1 making submissions impossible.

Video demonstration of what is now possible.
https://user-images.githubusercontent.com/102533453/169524415-7fdc04b9-d9d8-4642-b0c9-c0bccf1cdfb9.mp4

EDIT: for best effects run install.php before testing


a20vicad (commented):
Great, now it works to the same level as the rest of the duggas.

PR #12791, opened 2022-05-20
Added a check to make sure that only tables with footers will have the sum/average row

a20johba:
The sum/average row was rarely used and often appeared as a blank white row. This is now removed from most tables except those that explicitly set the hasFooter variable to true, such as the result table in contribution. Other tables may need to be updated if they wish to use this feature.

To test:
1. Go to duggaed/resulted/fileed
2. Confirm that the table does not have an empty row on the bottom
3. Go to contribution
4. Confirm that the top-most table (above the weekly bar chart) has a bottom row for average/sum.

Note that the table on the bottom of the contribution page has values summed up on the bottom row, but does not make use of the footer. This was the case before my changes as well and it may need to be updated to be in line with the other tables.


a20vicad (commented):
This works well for the current version and should keep working

PR #12796, opened 2022-05-20
Revert "It's now possible to delete variants with assigned user answers #12638"

i19matla:
Reverts HGustavs/LenaSYS#12663

Testing lost functionality


i19matla:
This pull request did not affect the problem we searced for. Tested before revert, what it should do works and it do not break anything.

PR #12797, opened 2022-05-20
Added a dark mode vairant to icon

a20willa:
For testing:
1. Go into darkmode in the homepage and check that the icon is changed


a20mtafa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

PR #12798, opened 2022-05-20
Revert "Revert "Diagram: Add confrim and localstorage clear to orginal reset button in lower left corner.""

i19matla:
Reverts HGustavs/LenaSYS#12794

This did not affect the problem we searched for.


i19matla:
Tested works.

Issue #12702, opened 2022-05-19
When loading in diagrams from hash you don't get the right diagram because of local storage


c20emili:
Make it so that when loading in a diagram with hash it gets prioritised over a locally stored diagram.

c20emili:
Localstorage is not working as intended as of last merge. Check if this is still an issue after localstorage is fixed.

PR #12806, opened 2022-05-20
Fix for issue #12702

c20emili:
#12702
Stores if coming from hash in session and use it to determine if should load in from localstorage or not.
For testing:
Save a diagram 
Save the link and password
Change the diagram
Open the hashlink with the password
Se if the diagram is the is the one you saved 
if it is the one before the changes


i19matla:
Solved in another issue from a other group that breaked functionality for us.

PR #12807, opened 2022-05-20
Increased size on button and made it non visible in mobile view

a20andlj:
Increased the size to match the other icons in the navbar.
And made the button not visible in mobile view.

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -1357,6 +1357,16 @@ header td {
   height: 32px;
 }
 
+#downloadButt {
+  margin-top: 2px;
+  height: 40px;
+}
+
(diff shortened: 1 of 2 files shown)


a20andlj:
Test1:
1. Go to Course Webbprogrammering
2. Klick the contact icon
3. Check if the size of the download button size is ok compared to the other button in the navbar menu.
(Note: The search field and the search icon cover the download button - but that is another issue. You can hide it in the inspector if you want to see better.)

Test2:
1. Resize the page to mobile view (under 480px) to see if the download button disappears. 
2. If it does - test ok.

a18gusjo:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #12693, opened 2022-05-19
Tools/functionalities not visible on smaller screens


a20samul:
As more functionalities are added to the toolbar some functions are not visible on smaller screens and thus become hidden since the toolbar is not scrollable. This might be an issue to consider for future developers.

<img width="233" alt="toolbar" src="https://user-images.githubusercontent.com/81613484/169218293-7f2cde61-4559-43df-9d63-2c14e18749ae.png">

> _Image 1: Demonstrates how rest of the functionalities are hidden, such as the include-button and there is not way to get to it by mousepress._

b19karhj:
If I dont solve this issue on Monday morning. 
Overflow-y will not work as it disables the function in the picture.
![bild](https://user-images.githubusercontent.com/81624541/169564674-0831ad39-d95b-4119-9595-56fa32965f0b.png)

A suggesten would be to move the the function in the picture out of the toolbar (but this is more work and will take a bit of time)

b19karhj:
Fixed so you can now scroll and use the pop out boxes, which you can see in the comment above. Problem is tooltip text is not working. The source for that problem is how the tooltips are fetched and displayed, and needs reworking.
To  see how the scroll works for the moment, search for "document.getElementById("diagram-toolbar").style.overflow = "scroll";" in diagram.js and diagram.css

PR #12808, opened 2022-05-20
ools/functionalities not visible on smaller screens #12693

b19karhj:
You can now scroll the the toolbar with middle mouse, so if you are on a smaller screen you get access to all items.

Test: click on diagram and scroll the toolbar with middle mouse.


b19karhj:
Overflow-y destroys the function of switching between uml and er 
bild
Whats in the picture will not work.

Issue #11766, opened 2022-04-22
Diagram Module - Research on UML lines


a20gabpa:
The purpose is to perform research on how lines and relations are displayed, and how they work in UML diagrams to prepare a foundation for discussion with the "customer". A sub-purpose is to research what kind of relations should be implemented in diagram dugga.

UML has a wide variety of lines and arrows to represent relation between elements. shown in following figure. Research on which lines are necessary to implement and how they possible can be edited in the diagram. For example, should it be a normal line where the cardinality / type can be changed through the options panel?

![image](https://user-images.githubusercontent.com/102604680/164687918-9acf4285-df95-4641-8bda-1e27ccd9b662.png)

Research the possibility of making the lines "smarter". At the moment when connecting UML elements with lines, it chose the shortest path with sometimes leads to the lines going behind other elements. What if it is possible to make the lines "smarter" so when a UML is selected, the line will leave the UML element from the top if the connection is placed over the element? Or something similiar to the third picture where the lines are created by horizontal or vertical segements?

![image](https://user-images.githubusercontent.com/102604680/164689598-5189fc1d-815c-4c4f-a2e4-414c4a6e4910.png)

![image](https://user-images.githubusercontent.com/102604680/164694472-3e71bdd3-9ee2-4ad1-a094-f269ea3f8ec9.png)

HGustavs:
These symbols are partially decoration on the inheritance symbol and partly decoration on association symbols

I would say that this is a variant on the EER Information Engineering symbol (crows foot notation)

HGustavs:
So we could show a crows foot instead of writing 1:N or 1..* or a diamond symbol or an arrow....

those are all variations on how we decorate the ends of associations.

a20gabpa:
My suggestion is that UML lines should be drawn with only 90-degree angles. One of the reasons is that if the lines are angled, any symbols on the edges need to be redrawn to follow the same angle. This can lead to more complicated code as, for example, a diamond has to be redrawn at a 30-degree angle. Another reason is to be able to minimize lines appearing behind other elements when they are close to each other, as shown in the original comment.

What I have in mind is that the lines used for UML elements, instead consist of several line segments in different lengths that together form the connection between the elements. The image below is a quick sketch of a UML diagram.

![image](https://user-images.githubusercontent.com/102604680/166228012-4d61f238-f144-4e62-a6ce-ab258081287b.png)

In this case, the line could be divided into three separate segments (marked with the colors red, green, yellow).

[...]

PR #12810, opened 2022-05-22
Straight lines and UML cardinality

a20gabpa:
Related to #11766 
Note, if this pull request passes the test, close #12811  since this request also includes the same code. And if this request don't pass the test, try #12811  where only straight lines are implemented.

What has been changed:
- Updated how the lines are drawn. UML elements use a special kind of line (polyline) that uses only straight line segments. 
- Implemented the ability to type/write cardinality on UML lines (at start and at end of line)
- Updated the options panel (generateContexProperties) to remove the ER kinds from UML lines and remove UML cardinality on ER lines.

What can be tested:

- Test by using the line tool on UML elements (class or inheritance). The ghost of the line should not be a single diagonal line, instead it should be a line consisting of three line segments.
- Test by using the line tool on ER elements (entity, relation, attribute). The ghost of the line should be a diagonal line (as before)
- Test to connect a UML class to an UML class. The new type of line should be used, similiar to the picture:

image

- Test selecting the new UML lines, it should work by clicking any of its three line segments. Also check if the selection box appears.
- Select an ER line and check that the options panel's content is correct depending on the selection. Selected UML lines should also have two extra textfield for its cardinality.
- Add cardinality to the UML line, test with one field and with both fields. (Note that the start-and end side depends on the order the elements are seleted when creating the line)
- Try labels and test to remove the cardinality.

What does not work:
- Cardinality with the UML inheritance symbol results in the text being behind the element (Still works on class-side). This is due to the line's connection being in the middle of the element and thus resulting in the offset being to small to compensate for this. However  I think that it should not be possible to have cardinality on inheritance.
- Clicking save on line from ER entity to ER attribute. The error is seen in the console. (This error is present on merge8-branch and thus not fault from the code)

What limitations are they:
- At the moment, the solution always use a line with three segments meaning that some types of placement where one or two segments would be enough or even better, is not implemented. 

Before:

image

After:

image

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -2695,14 +2695,19 @@ function changeLineProperties()
     var radio1  = document.getElementById("lineRadio1");
     var radio2 = document.getElementById("lineRadio2");
     var label = document.getElementById("lineLabel");
+    var startLabel = document.getElementById("lineStartLabel");
+    var endLabel = document.getElementById("lineEndLabel");
     var line = contextLine[0];
-
-    if(radio1.checked && line.kind != radio1.value) {
-        line.kind = radio1.value;
-        stateMachine.save(StateChangeFactory.ElementAttributesChanged(contextLine[0].id, { kind: radio1.value }), StateChange.ChangeTypes.ELEMENT_ATTRIBUTE_CHANGED);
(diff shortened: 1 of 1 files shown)


a20chrka:
Test passed, looks good.

i19matla:
Works fine!

PR #12811, opened 2022-05-22
Updated the line tool to use line segments (polyline) for UML elements

a20gabpa:
Related to #11766 
Note, review this only if #12810 fails.

What has been changed:
- Updated how the lines are drawn. UML elements use a special kind of line (polyline) that uses only straight line segments. 

What can be tested:
- Test by using the line tool on UML elements (class or inheritance). The ghost of the line should not be a single diagonal line, instead it should be a line consisting of three line segments.
- Test by using the line tool on ER elements (entity, relation, attribute). The ghost of the line should be a diagonal line (as before)
- Test to connect a UML class to an UML class. The new type of line should be used, similiar to the picture:

image

- Test selecting the new UML lines, it should work by clicking any of its three line segments. Also check if the selection box appears.

What does not work:
- Clicking save on line from ER entity to ER attribute. The error is seen in the console. (This error is present on merge8-branch and thus not fault from the code)

What limitations are they:
- At the moment, the solution always use a line with three segments meaning that some types of placement where one or two segments would be enough or even better, is not implemented. 

Before:

image

After:

image

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -2053,7 +2053,6 @@ function determineLineSelect(mouseX, mouseY)
     }
     
     for(var i = 0; i < allLines.length; i++) {
-        
         // Copy the IDs.
         bLayerLineIDs[i] = allLines[i].id;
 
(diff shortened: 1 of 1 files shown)


i19matla:
Works fine

Issue #12778, opened 2022-05-20
Diagram: Finding out what causes hard refresh to not work with dark mode, hash and other things


b19karhj:
Finding out what causes hard refresh to not work with dark mode, hash and other things

PR #12820, opened 2022-05-23
Fixed functionality for hash, once again #12778

a20micro:


a20micro:
Issue not exactly corresponding with added functionality, however as there's like 3 hours left it's considered be of enough importance that the hash functionality is correct so it's sent in on the branch where the main research was done.

a20chrka:
Test passed!

PR #12830, opened 2022-05-23
style.css continued changing from static css colors to ones in root

a20sarle:
Changed static css color values to ones used in root.

Should be tested usin versiontesting (ask group leader how it's done).


a20muhal:
Tested for version error, Passed!

Issue #12804, opened 2022-05-20
Error check: Relations attribute types


a20edwth:
ER Relations can have all the types except keys.

So these should be allowed:
![25d7d8597e5e8438d5039cbce777ca4f](https://user-images.githubusercontent.com/102611046/169534063-044416b4-1a66-4b0f-88bb-ee7e126fb564.png)

a20edwth:
ER Relation attributes can now have every type except key types.

Instructions for testing:
1) Click "Demo-Course"
2) Click "Diagram Dugga"
3) Change type of an attribute on a relation (non-keys: correct, keys: error)

PR #12834, opened 2022-05-23
Relation attributes all types except keys

a20edwth:
ER Relation attributes can now have every type except key types. #12804 

Instructions for testing:
1) Click "Demo-Course"
2) Click "Diagram Dugga"
3) Change type of an attribute on a relation (non-keys: correct, keys: error)


i19matla:
Tested and works nice

PR #12839, opened 2022-05-23
Moved the search icon to the right place

a20mtafa:
I moved the search icon to the right place by using Padding-right


a18gusjo:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #12838, opened 2022-05-23
Diagram: Update demo data


b19karhj:
Update demo data so its the latest version

PR #12842, opened 2022-05-23
 Diagram: Update demo data #12838 

b19karhj:
Updated to new demo data, So its fresh and the people who take over do not have to do it.
Test: go into diagram dugga, if a diagram loads. it works.

Reinstall project before testing


i19matla:
Works

Issue #12817, opened 2022-05-23
Create the option to choose UML aggression and composition or IE optionality and cardinality at the end of lines


c20emili:
There's symbols at the end of lines that users supposed to be able to pick in both UML and IE. Add the functionality to draw the select symbols for both UML and IE.
UML:
![image](https://user-images.githubusercontent.com/81320323/169765420-edc6b03d-1a00-4108-8989-5fa1c8cf4e83.png)

IE:
![image](https://user-images.githubusercontent.com/81320323/169765487-d559feb3-855b-4257-898d-7364a9d66f24.png)

c20emili:
Options added but needs improvement

PR #12843, opened 2022-05-23
Added symbols to end of lines for UML and IE

c20emili:
Added dropdowns to the lines between UML and IE entities/classes where you can select the type of symbol to the end of the line and a radio button to be able to select a dashed line.
#12817 
For testing: 
Create multiple UML or IE entities/classes.
Select the line tool and draw a line between them
select the lina and open options
add cardinalitys and symbols and se if it works
if the entities/classes are to close together the symbol does not look the best
UML should work but IE is not tested

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -866,7 +866,8 @@ const relationState = {
  */
 const lineKind = {
     NORMAL: "Normal",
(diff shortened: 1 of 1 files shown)


i19matla:
Works fine, tested on merge to be tested with other functionality

PR #12846, opened 2022-05-23
Revert "Revert " Diagram: Variant: Disable, update and enable, #12710 ""

i19matla:
Reverts HGustavs/LenaSYS#12844
Seems like the problem comes from another merge from another group. Adds back this pull request.


i19matla:
Works

PR #12848, opened 2022-05-23
Changed the background to darker

b20fanfa:
Changed the background to darker so that it would fit in dark mode. Then you can see the text.


b20fanfa:
### For testing

1. Log in as teacher.
2. Go to "sectioned", at the top it should be there.
It should look like this in dark mode:
Skärmavbild 2022-05-23 kl  11 51 23

a20mtafa:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #12851, opened 2022-05-23
Make lines draw correctly from IE-inheritance


a20samul:
<img width="170" alt="IE-inheritance box" src="https://user-images.githubusercontent.com/81613484/169800446-0f36d8a8-fe3a-47df-93e8-b9a2e7d6c442.png">

Fix height on the box.

PR #12853, opened 2022-05-23
Lines draw correctly from IE-inheritance #12851

a20samul:
- #12851

Lines now draw correctly from IE-inheritance.

How to test:
1. Place an IE-entity on the diagram
2. Place and IE-inheritance on the diagram
3. Draw a line between them, manke sure there isn't any space between the line and entity.
4. The fact that the inheritance turns black when drawing a line between ie-entities will be fixed in another issue.

lines


a16linas:
Test passed

i19matla:
Works fine

PR #12854, opened 2022-05-23
Added error promt if query fails

a17nikjo:
When deleting undeletable things in duggaed it now promts an error.


a20vicad (commented):
Good, now it also handles other potential problems.

PR #12865, opened 2022-05-23
Dark mode variant added to diaggam

a20willa:
For testing:
1. Go into a diagram dugga
2. Check that the description field is changing in dark mode


a18gusjo:
Code refactoring - ok
Code standard - ok
Meet issue goal - yes

Issue #12467, opened 2022-05-12
Implement contribution overview


a20vicad:
We would like an overview mode for contributions. It should be low-level-of-detail (Glanceable) in style with a tag cloud (pic. related)
![image](https://user-images.githubusercontent.com/102599327/168066000-1729f2bd-4892-4a7d-927d-f284540e2797.png)
Read, it should show perhaps in which file most changes have been made and in what language.

Related to #12356 & #12353

PR #12869, opened 2022-05-23
G3 2022 w8 #12467

a20behta:
Displays all files from contribution database into contributino file. 

displayed based on procent of blame usage. 

testing. 
go to contribution
logg in and pick a user.
scroll down and watch chaos. 

NOTE
functionality not completed and needs some more work.  
specifically placing them correctly according to issue


a20vicad (commented):
Good, this is a good start and the next group will still have some things to do.

Issue #12872, opened 2022-05-23
Diagram icons needs update


b20thelo:
The Latest CSS update caused some issues in how icons are shown.

![image](https://user-images.githubusercontent.com/81772705/169864557-23e0b516-475a-4773-83d5-cb7d52ff4297.png)
Background of the active icon appears in front of the sub toolbar arrow. 

![image](https://user-images.githubusercontent.com/81772705/169864916-3d77694c-abfb-412a-bd35-617f17c1d3e1.png)
Icons for the sub toolbar have rounded corners while the main toolbar doesn't.

![image](https://user-images.githubusercontent.com/81772705/169865105-602376c0-7550-47a2-90a5-02de7b681c25.png)
In the sub toolbar, the icons are placed too close to each other. It is hard to see that they are separate icons.

PR #12873, opened 2022-05-23
 Diagram icons needs update #12872 

b19karhj:
Fixed css issues that was not noticed "gone" in todays test.


i19matla:
Tested and works

PR #12874, opened 2022-05-23
Suggested changes have been applied and tested.

a19simru:
I have tried testing why it was set up this way. It seams so deliberately done. But I can't figure it out. Everything works if substr(1) is removed. 

to test my change switch to my branch and try to use a issue link in contributions.


b20hanwa (approved):
Works as it should when tested. Good job!

PR #12983, opened 2023-04-03
G2 2023 v0 issue#12968

d21erisu:


b21kurar:
Please add .txt file with a description of the model

d21erisu:
Vi tolkade det som att att diagramservice.php hämtar data från github gällande tex kommentarer, issues och tillagd kod under en vecka. Datan hämtas och presenteras. Vi tror att de andra filerna inte gör just detta. Funktionaliteten i diagram.php var inte "samma" som i diagramservice.php. Lite oklart...

a21ponkr:
The text document doesn't include all the requested input/output parameters specified in the issue. Otherwise it looks correct.

a21ponkr (changes requested):
The text document doesn't include all the requested input/output parameters specified in the issue. Otherwise it looks correct.

a21jeaha (changes requested):
i agree with @a21ponkr assessment of the current state of the issue.

d21erisu:
We fixed it! Please review again ;)

a21jeaha (changes requested):
test reviewed by me and @a21ponkr.

the input parameter log_uuid is missing, check row 33. I interpreted it as an input parameter.

a21jeaha (approved):
good work, no further issues found

a21jeaha:
i just realized something shouldn't the database query's count as input output as well ?

c21alest:
I would argue that queries aren't anything that the service receives upon start or sends to the caller. It uses the input data described to collect data from db (external communication in the middle of the process is not what I would describe as either input or ouput data, more like manipulating data) and this data is stored in a variabel query which is then inserted into the output array.

b21kurar:
I agree @c21alest , this seems to be an appropriate interpretation for the purpose of this model. If needed in the future, then we could add query as a separate parameter, as per @a21jeaha 's comment.

PR #12987, opened 2023-04-03
Prevents multiple clicks when toggling instructions.

b21phiro:


b21eriho:
The implementation does indeed prevent the behaviour of repeated expanding and compressing of the info-box when clicked faster than the panel opens and closes. This box or panel is located at the top of the page in diagram dugga

PR #12994, opened 2023-04-03
Removed inline styling and added padding

b21eddka:
Removed the inline styling for "assignment_discrb" and added said padding.


a21aldsa:
The issue seems to be solved, the inline styling is removed and the padding is added. The code has also been transferred from diagram_dugga.html to diagram.css. **However**, it seems you have changed the indentation of the whole file. If i recall correctly, this is not allowed.

b21oscpe:
By request of @c21adano & @a21aldsa this pull-request is closed

PR #12998, opened 2023-04-03
Change text to white when uml diagrams are black

a21aldsa:


d21matbo:
Issue appears to already be solved from #issue12982 and merged with more functionality

PR #13008, opened 2023-04-03
Change color from black for readability when an entity is pink

a21aldsa:


a19clasv:
Fixed! And you also added in a change in color for when it's black.

a19clasv (commented):
Looks good! Pun intended.

Issue #12882, opened 2022-05-24
Store the diagram in localstorage to save current changes if the student if accidentely leaves the page not working at the moment.


c20emili:
There is a function to store the diagram in localstorage so the  student can resume on the same diagram they left if they leave the page. It was supposed to be moved so it did not happen everytime the mouse moves but only when leaving the page but part of the code went missing during a merge and now it does not work as at all since it never stores the changes. Missing code to store the diagram when leaving the page from dugga.js in the function addAlertOnUnload():
``` 
// For the diagram dugga there is a diffrent condition
else if(document.getElementById("diagram-iframe")){
    if (document.getElementById("diagram-iframe").contentWindow.stateMachine.historyLog.length != 0) {
        document.getElementById("diagram-iframe").contentWindow.storeDiagramInLocalStorage();// storing the diagram in localstorage
    }
}
```
Currently in mmoving(event) for diagram.js on row 2500 (bottom of the function) there is an outcommented function storeDiagramInLocalStorage() that  should be removed if localstorage gets stored when leaving page.

PR #13009, opened 2023-04-03
Store and retrieve diagram from localstorage

a21oligu:
Changes to a diagram is now saved to localstorage when a user leaves the page. When a user opens a dugga, the diagram from localstorage is retrieved and loaded. Solved issue #12882


a21aldsa:
localstorage works fine on diagram dugga, good job. 

It is noteworthy though that the colors do not get saved while everything else does, i think this should be a new issue.

Issue #12984, opened 2023-04-03
Changing entity type allowes you to connect entities of diffrent types.


c21adano:
Here two er entities are connect:
![ertoer](https://user-images.githubusercontent.com/102599332/229482323-0419ac22-ae30-4e19-9cea-597a0a8e321e.PNG)

But if one of the er entities is changed to uml, they willl still be connected:
![umltoer](https://user-images.githubusercontent.com/102599332/229482340-1a54151c-c4a6-4fbc-80f9-c7c4b242b4a9.PNG)

Two diffrent entity types should not be able to connect with each other.

PR #13011, opened 2023-04-04
Prevents element to change type if lines are attached.

b21phiro:
Notera: Detta är egentligen för issue nummer #12984, men råkade skriva fel tidigare.


b21eddka:
Everything looks good and is working properly, however according to the code standard you are not supposed to use box comments inside functions. I still approve of this pull request.

b21eddka (approved):

Issue #12954, opened 2023-04-02
Improve aggression, composition, optionality and cardinality at the end of lines.


c21adano:
When selecting an icon to appear at the end of lines a save must be performed before the icon appears. Improve the function so it updates when an icon type is selected.

PR #13012, opened 2023-04-04
Improve aggression, composition, optionality and cardinality at the end of lines. Issue #12954

b21willu:
Added onchange="changeLineProperties"  to both select dropdowns. This will call the changeLineProperties() function whenever an icon is selected.


a21aldsa:
it works

PR #13015, opened 2023-04-04
G2 2023 v0 issue#12962

c21axepe:
Me and a21oscgu have worked on issue #12962:
https://github.com/HGustavs/LenaSYS/issues/12962

UML class diagram with short description and Input and output i separate txt file created


a20fanma:
It looks good at my knowledge. The only thing I want to point-out is, in class diagram, coursename is missing "c". It says "oursename" instead of "coursename"

a21jeaha (changes requested):
@a20fanma 
It looks good at my knowledge. The only thing I want to point-out is, in class diagram, coursename is missing "c". It says "oursename" instead of "coursename"

PR #13019, opened 2023-04-04
issue#12885

a19clasv:
Ready for testing


b21phiro (changes requested):
The issue states "Try and do this in the main css file (...)". The developer have only made changes in the darkmodeToggle.js file and diagram.php, but not in asked file. 

The changes in the file "DuggaSys/darkmodeToggle.js" does not change the behaviour from the previous code.

a19clasv:
I reverted the changes in the file "DuggaSys/darkmodeToggle.js", but "blackTheme.css" already seems to have some stuff for the diagram and I also didn't even touch "diagram.css".

d21matbo:
As written above a "refresh" is required for the diagram to change color, otherwise it gets the job done!

PR #13022, opened 2023-04-04
G2 2023 v0 issue#12966

b21ricni:
https://github.com/HGustavs/LenaSYS/blob/master/DuggaSys/pushnotifications.php

Created class diagram and textfile on pushnotifications.php issue#12966


a20fanma:
It looks good at my knowledge. I do not think bindParam() should be in class-diagram. bindParam is use to bind a variable in the file with a variable in the database.

d21erisu:
Removed bindParam()

Issue #12999, opened 2023-04-03
condens line 646 with 650


c21adano:
in diagram.php:
line 646 has a check if the ID variable is set. but nothing is performed in the if statement. this can probably be condensed so that the check is performed on line 650.

PR #13028, opened 2023-04-04
G1 2023 v0 #12999

b21matfi:
Commented out code that was not used in diagram.php line 649 so that it can be used in the future.


b21oscpe:
#12999

b21phiro (approved):
The code indeed does nothing, and is now not running.

Note: Could've been removed entirely instead of commented out.

PR #13029, opened 2023-04-04
Diagram: Last two tooltips goes beyond bottom of the screen.

a21aldsa:


a21aldsa:
They were all pushed down to accommodate for the top tool tip spilling out over the top. So I ended up pushing them all up (making it so the bottom two don't spill over downward) and then giving the top tool tip an ID and giving that ID its own top value in css.

This seemed more elegant than adding IDs or a class to just the bottom two tool tips or something along those lines, even though my solution did affect the css of all the tool tips, instead of just the ones regarding the issue.

Issue #12957, opened 2023-04-02
CTRL z reverts all colour changes.


c21adano:
CTRL z will revert all colour changes first simultaneously. It dose this even if other changes happend after the last colour change.

b21eriho:
The actual issue at hand is the fact that colors are not appended to the undo-redo-structure at all. This isn't handled in the same way as other attributes since those are input fields so the color works differently

c21adano:
are you able to add colours to the undo-redo-structure or is it not possible due to them working differently?

b21eriho:
Only saw this now after already fixing this, but yes the fix was to add the color-change-event to the undo-redo-structure correctly, it was a single line that had been missed or accidentally removed

PR #13031, opened 2023-04-04
G1 2023 v0 #12957

b21eriho:
color-changing events have been added into the undo-redo-structure, a single line was missing for this. Beware this interacts with another bug meaning the first two events performed will be undone as one so when testing perform more than two changes and perform color-changes after those first two as well.


DuggaSys/diagram.js
c21axepe, line 1: Looks good!

c21axepe:
Reviewed and done

PR #13033, opened 2023-04-04
G2 2023 v0 issue#12964

a21jeaha:
model and description ready and pending review


a21jeaha:
sorry my bad, ill keep that in mind for next time..

backend-models/resultedservice.txt
c21alest, line 1: Maybe vers should be coursevers instead as it is the PHP variabel

c21alest (changes requested):
To my knowledge json_encode and array_push aren't any methods instead they are standard PHP functions used to manipulate data. Otherwise it looks good! But I'm unsure if the classes resulted.php and resulted.js are needed. My understanding is that we are only suppose to model the service (PHP backend), so I don't know if this is good (more detail) or bad because it's information we don't need. Maybe that's something for @b21kurar and @a19hakja to decide?

backend-models/resultedservice.txt
a21jeaha: i don't quite follow, is it the diagram, or the in the txt file?

a21jeaha:
your completely right, that was a misunderstanding from my part. I have taken it with the higher ups, we wont remove the already supplied work as it might be needed at a later time.

a21jeaha:
json_encode and array_push removed from diagram
input: vers renamed to coursevers in txt file

a21ponkr (approved):
Looks good

PR #13034, opened 2023-04-04
G2 2023 v0 issue#12960

b21hamwe:
issue#12965 from yesterday follows here as well since an un-noted problem with github desktop occured. 
Not the best solution, but issue-responsible (b21kurar) approved of this solution.

For issue#12960
I would still recommend the testing individual to look out for redundant variables in the diagram model, although i've tried my very best for this not to occur.


backend-models/codeviewerService.php.svg
a21jeaha, line 1: some duplicates and missing variables that needs fixing :)

dubbles
-------

sectionName
courseId x 2
exampleName 
currentId
word
wordList


missing
-------

codeDir
box

a21jeaha (changes requested):
variables need adjusting, iv supplied a list of found issues :) 

dubbles
-------

sectionName
courseId x 2
exampleName 
currentId
word
wordList


missing
-------

codeDir
box

a19hakja:
Good job, looks good to me aswell. Merging.

b21kurar:
Messy merging, let's not do this again

Issue #13032, opened 2023-04-04
Database code in diagram.php


HGustavs:
It seems like we have back-end stuff in diagram.php this should for architectural reasons be in diagramservice.php

Document which back-end stuff ended up in diagram.php and if it can be moved to diagramservice.php

Especially since diagramservice needs to be moved to microservices.

Same thing for iohandler should this be moved to sevices or not? Is that file a second diagramservice.php or something else. This is relevant for microservice migration.

c21adano:
diagramservice.php right now only tracks a users activity on git hub. it fetches relevant data from the database for what the user has done for the current week and then prints out a json string containing that data. I believe we could repurpose the diagramservice.php file so it contains all the database code from diagram.php and returns json objects for querys made to the database that diagram.php then can use.
Although you have to be careful and also check diagramIoHandler.php because it seems that it is sending a request to diagram.php to create a new directory, Although i am unable to track down where this gets done in diagram.php.

c21adano:
Group 1 and 2 leaders have talked about who should move the backend code, and we agreed that group 1 can do it

Issue #12995, opened 2023-04-03
database request result in sql errors


c21adano:
in diagram.php:
As things stand now a connection is made to the database but there is no check on whether the connection succeeded or failed. As a result when attempts are made to perform a request to the database you will get sql errors.  rework this so that no requests are performed if the database is not connected.

PR #13041, opened 2023-04-04
Issue#12995

a21nibah:
just a clerification. I saw that i wrote that this issue will cause errors. it's not that it will cause errors but more like that it could cause errors. so i have fixed so pdo connect  and the diagram files now handle a fialed connection more gracefully.


a21nibah:
no it is fine. it is just i had to change the rights to read and write so i could work via ssh on the server. i can provide a list of files that were changed if needed

b21oscpe:
"ah yes. det där var bara för att jag behövde ändra rättigheter på allt till execute så jag kunnde jobba mot servernn. de filer som ändrats finns i commit" - @a21nibah

b21oscpe:
The discussion about the "745 files" can be ignored in testing, the problem is resolved

c21adano:
Related: #13032 
I think its best to temporarily close this pull request, due to issue#13032 which is about moving the database code out of diagram.php and the numerous file changes done. The file changes themself are not harmful but may "overwrite" someone else work.
This issue(#12995) will be reviseted at a later date and its branch will be keeped for that reason.

Issue #13014, opened 2023-04-04
Stop it from changeing text color when writing


b21sebgr:
Investigate and fix the changeing of color in the box where the github name should be written in (check correctnes in both dark and light mode) currently the text goes almost invisible in darkmode

Bug found in DuggaSys/contribution.php

![image](https://user-images.githubusercontent.com/102599253/229743232-593f0c76-9487-4ca1-90db-9242c6ac5b7c.png)

![image](https://user-images.githubusercontent.com/102599253/229743402-163446c9-f9f3-47b9-adcf-ef783d1f10df.png)

c21sebar:
Can this also be related to previous work of students with GitHub integration #13010 ?

PR #13054, opened 2023-04-05
Darktheme inputs text color and login form background color

b21leowa:
Solved why the input field text was black on black background. This fix ruined login form because it had inline css background color white so it was white text on white background. Therefore I commented out the inline css in dugga.js

Issue #13014


b21leowa:
För att testa min kod, gå in på följande: 
https://cms.webug.se/root/G3/students/b21leowa/LenaSYS/DuggaSys/courseed.php

a21lukpo (approved):
Everything works correct. Found similar issue on different places. Creates new issue for this.

Issue #12884, opened 2022-05-24
Diagram: Remove inline styling in diagram_dugga.html and move to diagram.css


b19karhj:
At the moment uploaded files text are hidden a bit.
![image](https://user-images.githubusercontent.com/81624541/170030458-6e582210-f0e6-4c3b-a612-72549ebf7f12.png)

Remove inline styling on row 157 in diagram_dugga.html

Add `padding: padding: 5px; padding-left: 25px; ` to class assignment_discrb in diagram.css

PR #13075, opened 2023-04-05
Removed inline styling and added padding. Issue #12884

b21eddka:
Removed inline styling from assignment_discrb aswell as added padding in diagram dugga. Fixed issue number #12884.


c21theti:
Looks good, padding is added.

PR #13089, opened 2023-04-05
G3 2023 v1 issue#13064

c21rebja:
Removed the yes/no functionality added in the last merge. Also changed the visual validation to use borders instead of background colors to make it easier to take care of the issue with the invisible text.


a21henhe:
Looks good

b21sebgr (approved):
This seems to be working correctly now

PR #13091, opened 2023-04-05
#12963 model contribution_loginbox_service

a21ponkr:
Added class diagram containing attributes and operations. Also added text containing input and outputs.


d21erisu (changes requested):
row 144 has  "$git_userrows"   you have "gituserrows" in the model, rename it to match :)
Otherwise it looks good!

d21erisu (approved):
Great!

b21kurar:
issue #12963

Issue #13024, opened 2023-04-04
Implement GitHub Markdown Link style


c21sebar:
Implement GitHub Markdown style to work with LeanaSys Markdown style

a21rammo:
#13004 This issue might contain useful infomation. To test if the implementation of Github Markdown works for Lenasys markdown. You will need to go into the course (the one i'm using is demo course).  First click the "cable" symbol:

![image](https://user-images.githubusercontent.com/102600501/230047401-fec201e8-ebe4-4f4c-917c-5c5ae92d4787.png)

**Click the "+" symbol**

![image](https://user-images.githubusercontent.com/102600501/230047606-3a9d9e3b-5d0b-4552-b37e-870558f093d0.png)


**Add dummy empty file:**

***The dummy file must be a .md file!***



Go and try to edit the new md file and you should see a markdown and to the right of it is a markdown preview:
![image](https://user-images.githubusercontent.com/102600501/230048463-2a9457f4-1006-4133-8db7-fa200173d785.png)

d21emijo:
from #13004 switched $1 to $2 and $2 to $1 as noted in https://github.com/HGustavs/LenaSYS/issues/13004#issuecomment-1496045933
Somehow this change parses everything correctly this way and not just the first letter.

d21emijo:
Potential issues if using characters such as []() in the link
or trying to implement github images as they use ![Alt img text](Link to image) as noted in https://github.com/HGustavs/LenaSYS/issues/13004#issuecomment-1497297557

PR #13104, opened 2023-04-05
try to switch $2 and $1

d21emijo:
issue #13024 ready to be tested and merged @a21rammo


c21rebja (approved):
Looks good!
image

But you have to use https:// before the link, otherwise it just sends you to "https://cms.webug.se/root/G3/students/c21rebja/LenaSYS/DuggaSys/youtube.com". The old markdown works this way too, but it could be improved as another issue.

PR #13107, opened 2023-04-05
g3-2023-v1-issue#13106

a21hugni:
Test if "Read Only" is visible in darkmode while logged in: DuggaSys/courseed.php 

image


g21emmka (approved):
Looks good!

Issue #13051, opened 2023-04-05
Change lines to white in darkmode


c21adano:
Line currently appear as black in darkmode and are hard to see, they should be white for visibility.

![linesindarkmode](https://user-images.githubusercontent.com/102599332/230044457-3a8d1729-ac38-43ae-ad87-a60d0a0c6bbe.PNG)

PR #13108, opened 2023-04-05
G1 2023 v1 #13051

b21matfi:
Added classes where the lines where created and then added some css code inte the blackTheme.css to change the lines when its dark mode.


b21oscpe:
#13051

Issue #13074, opened 2023-04-05
Add a button in the menu for placing statediagram states.


b21oscpe:
![issue-state](https://user-images.githubusercontent.com/102578165/230064589-9f659ca6-d182-4415-bd71-4f8d2a86f174.png)

Add a button for state diagram entity

b21eriho:
In the system, there is functionality to setup which diagram types may be used in the diagram, I am making the assumption that this is not part of this issue and my role is simply to add the button and the logic for when it should be displayed is another issue? Is this correct?

b21eriho:
Further, should I create the icon? and if so is there a standard way to do this so that it matches the already existing icons?

c21adano:
"should I create the icon?" - yes you should create the button, but you can reuse the IE entity image for the time being.

 "my role is simply to add the button and the logic for when it should be displayed is another issue" another person is making the state entity and all that is required right now is a button which can later be linked with state entity.

b21oscpe:
Exactly, all that is required from this issue is a way of creating a state entity, the rest is handled in later issues

b21eriho:
As in the button that later will add a state entity, or a button that *does* add a state entity, because the latter would require implementing a state entity in entirety in this issue

c21adano:
the button that later will add a state entity

b21eriho:
So, I've added the button to the toolbar, however the system has a way of dynamically hiding the appropriate buttons on the example diagram page based on the setup of that dugga. This does not work with the new button since that would require giving it an id that corresponds to the action of creating a state entity, which obviously isn't implemented yet. The result of this is that said button is displayed when it should be minimised along with a few other visual and functional bugs that depend on the supporting systems for the toolbar.

On top of this the new amount of buttons pushes another toolbar entry off the screen, this entry was simply information regarding special characters so could be relocated reasonably in theory since it isn't a tool and as such needen't be in the toolbar.

I assume I should simply describe this in my pull request because the system does not behave correctly at this point so testing will otherwise return with "there are bugs" even though the bugs need to be fixed in the "backend" of the system and as such do not belong to this issue

c21adano:
Sounds good.

This issue is related to issue#13057 were the the entity will be added. Other issues will be dealt with later.

b21oscpe:
"A merge should never make the system worse", send me a message on discord so we can discuss the problem

PR #13112, opened 2023-04-05
G1 2023 v1 #13074 Add a button in the menu for placing statediagram states.

b21eriho:
The button is now there, however it's sometimes there when it would usually be hidden by the system. Selecting it also in some cases doesn't allow you to swap back due to a bug. However all these things are caused by the backend not being implemented yet. Also the icon for a statediagram state is at this point the same as for IE-entity

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -737,7 +737,8 @@ const keybinds = {
         PLACE_UMLENTITY: {key: "6", ctrl: false},       //<-- UML functionality
         EDGE_CREATION: {key: "7", ctrl: false},
         PLACE_IEENTITY: {key: "8", ctrl: false},       //<-- IE functionality
-        IE_INHERITANCE: {key: "9", ctrl: false},  //<-- IE inheritance functionality
+        IE_INHERITANCE: { key: "9", ctrl: false },  //<-- IE inheritance functionality
+        PLACE_SDENTITY: { key: "1", ctrl: true },   //<-- SD functionality
         ZOOM_IN: {key: "+", ctrl: true, meta: true},
         ZOOM_OUT: {key: "-", ctrl: true, meta: true},
         ZOOM_RESET: {key: "0", ctrl: true, meta: true},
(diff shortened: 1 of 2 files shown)


b21eriho:
I can't tell how the tooltip is different from your screenshot, in what way is it inconsistent?

a21aldsa:
I have attached a screenshot of the tool tip for ER. Note how it is much cleaner. I understand why you have done it this way but I strongly believe there are ways to rewrite this to look more elegant. Perhaps just add a space between state and diagram since these are two separate words?

image

b21eriho:
Wording updated

b21eriho:
I have implemented the button, and it now gets hidden like it should in the demo diagram and is shown appropriately in dugga.php. However the system will print endless errors when the state diagram tool is selected and if one attempts to place a state the system breaks and no other tools can be selected or used.

This issue is however caused by the values for the state entity are undefined and the system can thus not read them. However the same attribute of the button that allows it to be hidden, causes the system to attempt to read these null values so I cannot bypass this issue. 

Summary: The button is there and is hidden when it should, however selecting it will cause errors and using it will break the system until a reload. Do I create a pull-request with this, since the solution to this issue is to implement the state entity which isn't a part of this issue.

c21adano:
for testing just assign another diagram typ to the button. The state diagram entity will be added in another issue.

b21eriho:
That will not work, in that case the button will be hidden in the same way that that other diagram type, the system is very interlocked in this case

c21adano:
hmm then i guess it should be fine to just test it right now, and fix the issues when we add the state entity to the system.

a19clasv (approved):
There is a state diagram state button, but there are other issues as mentioned in the previous comments. However, c21adano said to fix the issues when we add the state entity to the system, so it's approved.

b21eriho:
These conflicts are indeed problematic, since both these developments and the ones they are conflicting with seem to have added things starting in the same spots and thus these changes should realistically be added in sequence, not be picked between

b21oscpe:
Pray

PR #13113, opened 2023-04-05
courseedService-issue#13090

b21hamwe:
Documented functionality in courseedService file and searched for potential microservices

### DIFF
--- a/backend-models/courseedService_microservices.txt
+++ b/backend-models/courseedService_microservices.txt
@@ -1,0 +1,79 @@
+courseedService_microservices
+
+
+Functionality : 
+
+Row 40 :If isset($_SESSION)
+Checks what kind of user it is, meaning, student or teacher, if none, get guest. 
+You get username based on your, "uid".
+Name:userCheck
+
+Row 49 : Gets the username based on "uid" for logging.
+Name : getUsername
+
+Row 54 : If "uid" exists, set username. 
+Name : setUser
+
+Row 67 : checklogin(). Checks what kind of user it is based on "uid". Micro
+Name : checklogin
+
+Row 72 : If the user is a teacher, theres an option to delete or create data in course table in the database. 
+Name : delCreateCourse 
+
+Row 85 : If the previous binding of parameter in query returned false, an error message presents.
+errorInfo() functions gets called. Micro
+Name : error
+
+Row 96 : Gets the username based on uid, was previously just set.
+If none, error message.
+Name : getUsername/getuid
+
+Row 103 : Fetches and prints all courses bound to the user, printed as "rows".
+Name : coursePrint
+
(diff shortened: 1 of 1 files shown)


backend-models/courseedService_microservices.txt
```diff
@@ -0,0 +1,79 @@
...
+You get username based on your, "uid".
+Name:userCheck
+
+Row 49 : Gets the username based on "uid" for logging.
+Name : getUsername
+
+Row 54 : If "uid" exists, set username. 
+Name : setUser
+
+Row 67 : checklogin(). Checks what kind of user it is based on "uid". Micro
```
a21jeaha, line 17: doesn't it just check if its true?
this is the function being accessed.

function checklogin()
{
      // If neither session nor post return not logged in
      if(array_key_exists('loginname', $_SESSION)){
        return true;
      } else if(failedLoginCount($_SERVER['REMOTE_ADDR']) < 100 && array_key_exists('username', $_COOKIE) && array_key_exists('password', $_COOKIE)) {
        return login($_COOKIE['username'], $_COOKIE['password'], false);
      } else {
        return false;
      }
}

you might be mixing this up with the logging that also takes place in the file :)

a21jeaha (changes requested):
I'm sorry to say that this needs thorough re-work. 

a clear list on what can become a micro-service is needed, the information can be found here and there but you have to search for it at the moment. The document concentrates to much on details rather than giving information on what actual services are provided. 
I would suggest separating the detailed descriptions, services provided and possible micro-services.

your missing actual services, examples of which are the creation of a new course, at row 70 - 111, creation of a new version of a course, at row 114 - 149.

as such, you need to find the functionality/services that the file provides. Knowing WHAT the file does helps you distinguish what services could become their own micro-service. It would be nice if each separate service had a start and end row, to make it easier do distinguish were code related to the service starts and ends, as of now we only get the start and have to find the end ourselves.


If it helps I have supplied links below of what I hope is the current documentation regarding the tables that are used. It might help you distinguish the services might be aiming to do. 


https://github.com/HGustavs/LenaSYS/wiki/Database
https://user-images.githubusercontent.com/43996354/81922100-eb859c80-95db-11ea-85e3-2bd4c5685426.png

backend-models/courseedService_microservices.txt
b21kurar: image
This is a separate function in another file (sessions.php). It is described there, we can skip it for this issue.

a21jeaha (changes requested):
the main functionality still has not been added. 

row 78 creates a new course 
row 114 Add new version of a course 
row 150 update information of the course version
row 171 make another version of the course the current active version 
row 179 copys a course version somehow?  

these are then ones i have found, i have not checked the duplication part

a20fanma:
Everythings looks good

PR #13119, opened 2023-04-06
Icon selection not saving right 

a21aldsa:
Updated what's saved to the state machine for startIcon for UML lines. There seems to have been a mistake in the code. It erroneously saved the value for startIcon under the name endLabel.


b21willu:
Looks good!

Issue #13068, opened 2023-04-05
Stop it from changing text color when creating a new course


a21lukpo:
Investigate and fix the changing of color in the box where a new course is created, name should be written in (check correctness in both dark and light mode) currently the text has the same color as the background so its invisible in darkmode.

Bug found in DuggaSys/courseed.php when logged in

![image](https://user-images.githubusercontent.com/102577654/230053228-8b202b96-ebb1-4424-a306-571b8d96b6ee.png)
![image](https://user-images.githubusercontent.com/102577654/230053288-6fad0a58-760c-4162-9b60-10689693a0a1.png)

c21sebar:
Same problem as in #13014 , maybe this problem exist on more text-labels in LenaSys? 
Seems to be an overall issue to all of LenaSys with darkmode support.

c21rebja:
When we merged issue #12991 yesterday this wasn't a problem. The fix in issue #13014 most likely caused this problem.

PR #13134, opened 2023-04-06
G3 2023 v1 issue#13068

c21rebja:
"Stop it from changing text color when creating a new course #13068"

Changed so that the input boxes turn red using classes instead of just setting the style specifically for that element. This makes it possible to use it on all inputs, no matter which theme is currently active. We tested this on the create course form, but it should work just as well on the others.

image


e21krida:
Tested, seems to work flawlessly on the webserver. 👍

a21olljo (approved):
Worked great on the webbsever for both the dark and light mode, code also looked good

PR #13136, opened 2023-04-06
G3 2023 v1 issue#13056

b21sebgr:
Added connection from frontend to backend for git url both when creating course and updating a course


e21krida (approved):
Ser bra ut, verkar inte vara några noterbara problem alls. 👍

Issue #13053, opened 2023-04-05
The ruler should also be darkmode once darkmode is activated


b21oscpe:
The ruler surrounding the drawing-area should also be darkmode once darkmode is initialized.

PR #13138, opened 2023-04-06
The ruler should also be darkmode once darkmode is activated. Issue #13053

b21willu:


c21theti:
Looks good. Ruler is in darkmode

c21theti:
But i could not change to lightmode because the darkmode toggle was not there

PR #13139, opened 2023-04-06
v2-2023-v1#issue12961 complete

b21tilbe:
A UML diagram has been created plus a textfile containing input/output/parameters


b21antbr (changes requested):
Most of it looks good! However, 
-duggainfo should be stated as duggainfo[] in uml, since it's an array. 
-entry[] missing from uml. 
-unset() missing from uml. 
Input output in txt looks good.

Issue #12975, opened 2023-04-03
Er entity with black background.


c21adano:
Er entity with black background changes colour when something else is interacted with.

before interacting with anything else:
![erwhitetext](https://user-images.githubusercontent.com/102599332/229470725-447f9136-7450-4832-a43b-b5abb041f5b9.PNG)

When line is selected the er entities change text colour to black:
![erblacktext](https://user-images.githubusercontent.com/102599332/229470735-24bef8b0-c156-49f0-b85a-87e0685ca68d.PNG)

a21aldsa:
I couldn't solve this issue but my findings are:

It seems that only the text becomes black,  this also applies for IE and UML too. My best guess is that updateCSSForAllElements gets overwritten or simply not called, since this has checks in place for checking if the background is black or pink, and if this is the case; the font color gets set to white. 

Whenever any other tool is used, even if its just panning with the selection tool, the text becomes white again. Maybe a function for updating some graphics is called here?

Issue #13052, opened 2023-04-05
Box outline hard to see in darkmode


c21adano:
Boxes have a dark gray outline which can be hard to see in darkmode and should be changed to a white colour.

![boxoutline](https://user-images.githubusercontent.com/102599332/230045187-0ac2de14-0e48-4028-808c-362071a92afb.PNG)

PR #13140, opened 2023-04-06
Ready for testing!

a19clasv:
Fixed issue #12975 by copying element updating in function updateCSSForAllElements() for when the mouse is not in EDGE_CREATION mode, but changed it so the copy goes off in EDGE_CREATION mode and removed the highlighting in the copy.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -7871,6 +7871,7 @@ function drawLine(line, targetGhost = false)
 
     }
 
+    updateCSSForAllElements();
     return str;
 }
 /**
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -10868,6 +10868,78 @@ function updateCSSForAllElements()
                         }
                     }
                 }
+            } else if (mouseMode == mouseModes.EDGE_CREATION) {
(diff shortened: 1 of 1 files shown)


a19clasv:
Also slightly refactored parts of the function by turning some repeated parts into inner functions to make it more readable and easier to modify in the future.

b21oscpe:
For tester:
This branch has merge-conflicts with g1-2023-v1 however if it works @c21adano and I will look into it and resolve these.

c21adano:
Related: #13052.
This pull request is trying to insert 10 rows of code after line 10894 and #13052 have already inserted about 40 rows after line 10893.

underlineconflict
The row of code underlined should be the only bit that causes the conflict. 84 to 92 are functions that a19clasv wrote, while line 94 holds a function call to check if border should be darkmode or lightmode. They should not clash in functionality:
fontContrast changes text colour while toggleBorderOfElements changes border colour.

The code should not need any change and only the conflict boarder should be removed the resolve the conflict. further testing may be required but from an eagle eye view of the code, nothing should break and the two codes should be able to coexist.
startandmidconflict

b21matfi (approved):
Everything works fine, issue fixed

Issue #13116, opened 2023-04-05
Changing properties (Name and Variant) on ER-Attribute does not work


d21matbo:
When placing down an ER-Attribute it recieves the standard properties

> Name: Attribute
> Variant: normal

These values can be edited through the menu, but when pressing Save nothing happens.

It appears that this issue only affects the ER-Attributes

![image](https://user-images.githubusercontent.com/102599161/230122264-2c572041-f881-41e6-b12b-abf615f9e13d.png)

PR #13144, opened 2023-04-06
Changing properties on ER-Attribute does not work #13116

a21oligu:
Made it so that newType is assigned with optional values. If both "typeSelect" and "propertySelect" are null, newType is instead set to undefined. Solves issue #13116 

Changing color while editing an ER attribute will reset the name and variant to thier default values in the options-pane. Could potentially be an issue.


d21matbo (approved):
Issue resolved, looks good!

Issue #13118, opened 2023-04-06
Icons not saved in options.


c21adano:
In the option panel the selected icons wont appear in the icon box after reload, they will still appear on the line but it will say that it dose not have any icon selected:
Triangle is selected for both sides:
![iconselected](https://user-images.githubusercontent.com/102599332/230069831-93b684d2-c18c-4d98-88e7-fe5a9ee41c8d.PNG)

After reload it wont say which icons are selected and selecting a new one will remove the other one.
![iconsafterreload](https://user-images.githubusercontent.com/102599332/230069850-07201bdd-754d-4e8c-8657-3054308aef2f.PNG)

Issue #13078, opened 2023-04-05
Icon selection not saving right


c21adano:
It seems like the icon selected on a line dont save correctly.

This can happen when a line has two icons and the whole diagram is reloaded: 
Before:
![image](https://user-images.githubusercontent.com/102599332/230069797-0209b1b4-53ad-4881-906c-1654041e554f.png)
After:
![image (1)](https://user-images.githubusercontent.com/102599332/230069817-459ef262-7227-47c0-b7ca-4fbcc9c9e92a.png)

In the option panel the selected icons wont appear in the icon box after reload, they will still appear on the line but it will say that it dose not have any icon selected:
Triangle is selected for both sides:
![iconselected](https://user-images.githubusercontent.com/102599332/230069831-93b684d2-c18c-4d98-88e7-fe5a9ee41c8d.PNG)

After reload it wont say which icons are selected and selecting a new one will remove the other one.
![iconsafterreload](https://user-images.githubusercontent.com/102599332/230069850-07201bdd-754d-4e8c-8657-3054308aef2f.PNG)

a21aldsa:
Firstly, i believe these to be two separate issues. The menu not stating the icons selected on a line is not just on reload but also when simply deselecting the line and then going back to it.

Secondly, I can't seem to replicate the issue with a reload making the lines bugged, since I can't see how to save a diagram in digram.php so that it doesn't just reset completely upon reload.

b21oscpe:
Are you testing on the correct branch? Diagrams should be saved in localstorage now if I'm not mistaken.

a21aldsa:
I believe I am, I will checkout g1-2023-v1 on the server just to double check. If it works there that means I had the wrong branch

a21aldsa:
Still no localstorage on diagram.php. Am I testing the right file?

c21adano:
The first half of the issue can only be recreated in the dugga version of the site, while the second appears in both.

This issue will be split into two smaller issues, aldsa will still be focusing on the first half of this issue. I will move the 2nd half to its own issue.

PR #13149, opened 2023-04-06
Icons not saved in options.

a21aldsa:
The previous author thought that icon could be used in an if case with startIcon/endIcon, this is not the case as there are several edge cases like icon being ZERO_ONE and startIcon / endIcon being 0-1. Or icon being WHITEDIAMOND while startIcon/endIcon are White_Diamond. Or icon being FORCEDONE and start/endicon being 1! 

This complete inconsistency of naming has made this issue very bothersome. The best way i found to solve it was to firstly use toUpperCase() on startIcon/endIcon, this solved certain icons like Arrow or Triangle. But the ones that could not be compared as easily I found it best to give them their own If statements; this does inherently lead to some bad smell and the possibilty for refactoring is massive. I strongly believe that if another issue comes up where the root of it lies in these inconsistent naming schemes; a new issue should be created for refactoring these to be consistent and, along side that, this solution should be refactored too.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -6281,34 +6281,39 @@ function generateContextProperties()
             //iterate trough all icons associated with IE. add these icons to the drop down
             //if the line in context has one of these lines in the starting position, just like for UML, it is automatically selected
             Object.keys(IELineIcons).forEach(icon => {
-                //this only really covers WEAK, since the rest have a inconsistent naming scheme, like ONE_MANY; its also reffered to as 1-M
-                //This means we have to manually check these and others like them
-                if (contextLine[0].startIcon != undefined && contextLine[0].startIcon.toUpperCase() == icon){
-                    str += `<option value='${IELineIcons[icon]}' selected>${IELineIcons[icon]}</option>`;
-                }
-                //icon can be ZERO_MANY while start icon can be 0-M.
-                else if (contextLine[0].startIcon != undefined && (contextLine[0].startIcon.toUpperCase() == "0-M") && (icon == "ZERO_MANY")) {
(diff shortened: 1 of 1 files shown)


a19clasv:
This is still an issue. It seems like if a line is drawn from entity a to entity b and the line tries to save an icon at a, which is on the left part in the options menu, the icon is instead stored in End cardinality which is shown when refreshing the page.

a19clasv (changes requested):
It needs to store icons as icons. The icon in the left part of the options menu is instead stored as a an end cardinality, which is shown when refreshing the page.

a21aldsa:
> It needs to store icons as icons. The icon in the left part of the options menu is instead stored as a an end cardinality, which is shown when refreshing the page.

This is referring to a previous issue of mine: issue #13078. This was an issue regarding the state machine and how the value of the startIcon was saved as an endLabel. 

I fixed this issue and it got merged, but not before I was assigned this new issue: #13118. Therefore, the changes made in that issue is not reflected in this branch.

If I'm not mistaken, this was an issue about the drop down menus not reflecting the current icon on the lines.

a19clasv:
I checked g1-2023-v1 and it had the same problem even though you said it got fixed and merged

a21aldsa:
> I checked g1-2023-v1 and it had the same problem even though you said it got fixed and merged

I tried replicating what you reported, with little to no avail. Please double check that you are on the right branch, fetched and pulled. If you still are finding that the icons on the lines are disappearing and are instead shown as text on said lines; that would be a rather big problem.

a19clasv:
It works now when I clicked on courseed.php instead of just refreshing, though I think I did that when checking before this branch got merged. It's at least fixed now.

PR #13153, opened 2023-04-10
Fixed so the input fields are correct.

b21leowa:
The input fields are now in the correct color depending on the theme. As seen below.

image

image

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -25,7 +25,7 @@ var lid;
 var collectedLid = [];
 var updatedLidsection;
 var numberOfItems;
-
+var backgroundColorTheme;
 var isLoggedIn = false;
 
 function IsLoggedIn(bool){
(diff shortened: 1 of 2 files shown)


b21sebgr (changes requested):
code looks good for the most part, good solution but you missed one field as seen on the screenshot:
image

the text in that field should be white and the background should be black

b21leowa:
> code looks good for the most part, good solution but you missed one field as seen on the screenshot: image
> 
> the text in that field should be white and the background should be black

Do you mean the Version ID: field?
I can fix the box but what color should it be?

Right now it looks like this for me: 
image
image

b21sebgr:
> 

yes i meant the "version ID" field and the "change this to default version" checkbox, the checkbox should not get a white background when pressed but rather stay the same background but with a white checkmark if possible (in darkmode)

PR #13157, opened 2023-04-11
Added State diagram start and stop.

b21phiro:


b21phiro:
NOTE: Found bug with text, dissapears when selecting a tool. But comes back when element is placed.

c21adano:
> NOTE: Found bug with text, dissapears when selecting a tool. But comes back when element is placed.

This issue is know and have been merged earlier today.

b21oscpe:
Changes have been made to the original issue

Issue #13000, opened 2023-04-03
Model service distribution in dugga.js


b21kurar:
dugga.js does a lot of things.

We need a model of the distribution to the services.
What parameters is sent to which services?

Related to issue #12920 

This issue could be split into many smaller models/issues that can then be consolidated to a larger model.

b21kurar:
This would mean model the parameters sent to what service, in lines 1016 to 1260

PR #13160, opened 2023-04-11
#13000 Model service distribution in dugga.js

a21ponkr:
Worked on issue #13000 where I modelled which file the parameters/services point to.


c21alest (changes requested):
There is two dugga.js - file classes. The access one should be renamed to dugga.js - access. Otherwise good!

a21ponkr:
Fixed, please review again! :)

a19hakja:
Nice, merging!

Issue #13050, opened 2023-04-05
State diagram start and stop


c21adano:
Add state diagram start and stop to diagram.php, they should function and have a visible contrast in darkmode.

![statediagramstartandstop](https://user-images.githubusercontent.com/102599332/230042585-8ebd500b-4542-407b-8227-3f7bc4be539f.PNG)

PR #13161, opened 2023-04-11
 State diagram start and stop #13050 

b21phiro:


d21matbo (changes requested):
The UML-State entities keeps their size when the zoom on the canvas changes, this does not appear to be correct since the other entities have their sizes change with the canvas.

The newly added buttons for Start and End-state claims to have the same keybind as the line tool

b21phiro:
NOTE: The keybinding is assigned, however it seems like there's an issue with keymodes not unique to this issue.

d21matbo (approved):
Entities now behaves like the others, looks good

Issue #13124, opened 2023-04-06
UML inheritance appears slightly above where it should 


b21oscpe:
https://user-images.githubusercontent.com/102578165/230330281-291cc32a-a60d-4397-8670-d0c1ae575e81.mp4

This issue is similar to / the same as #13083

b21oscpe:
This isssue seems to occur only for some, investigate if this is unique to different computer/screens or if its a bug in the code.

PR #13167, opened 2023-04-11
 UML inheritance appears slightly above where it should #13124 

a21aldsa:
added some margin-top to the div encapsulating the UML inheritence element


b21phiro:
Looks fine to me.

Issue #13151, opened 2023-04-06
Divide "checkElementError()" into smaller functions


b21oscpe:
Divide the function mentioned in #12924 into smaller, more maintainable and readable functions.

PR #13168, opened 2023-04-11
Refactor checkElementError() #13151

a21oligu:
Refactored the function _checkElementError()_ by dividing it further into more maintainable, traceable functions: _checkEREntityErrors()_, _checkERRealtionErrors()_, _checkERAttributeErrors()_ and _checkLineErrors()_.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -8601,7 +8601,6 @@ function checkEREntityErrors(element) {
     var primaryCount;
     var strongEntity;
     var weakrelation;
-    var lineQuantity;
 
     // Checks for entities with the same name
     for (var i = 0; i < data.length; i++) {
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -8190,6 +8190,7 @@ function drawElement(element, ghosted = false)
     var xAnchor = tooBig ? margin : hboxw;
     var vAlignment = tooBig ? "left" : "middle";
 
+    //console.log(123123123)
     if (errorActive) {
         // Checking for errors regarding ER Entities
         checkElementError(element);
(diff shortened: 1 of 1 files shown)


b21eriho:
What do you mean Adam, that you have decided against these changes or are you telling me to go through it before saying that it has passed testing (which is obvious)

c21adano:
there are over 1400 changed lines in this pull request

c21adano:
just want to understand why that is before a merge is done

b21eriho:
Yes of course... I do not understand your point, of course I would not simply say to merge it without understanding the changes... Are you saying you are also reviewing it?

a21oligu:
I divided the function checkElementError() into smaller functions. The checkElementError consisted of all the logic for the different types of tests, although only one was used at a time. I divided these sections into separate functions by moving the sections to individual functions containing the code only for the test that is to be run.

b21eriho:
Thank you, that is very helpful

c21adano:
no just panicked a little, large changes like this one which changes 1400 lines need to be investigated as to way so many changes happend. 

It seems like part of the code has been auto indented:
autodent

This is the cause of my concern, we are not allowed to merge things that auto indent large parts of the kode.

b21eriho:
Rather, it seems since a large function was split into multiples the indentation had to be changed since the code is no longer inside an if-statement

a21oligu:
> Rather, it seems since a large function was split into multiples the indentation had to be changed since the code is no longer inside an if-statement

@b21eriho is right, splitting the function meant that the if-statements was removed, thus the indentation was changed.

c21adano:
ok, but wait with the merge for now. local storage seems broken and we dont wanna merge until we find why.

b21eriho:
I am still in the middle of the code-review and naturally only leader and issue-responsible may merge so this will not be merged without your approval

b21eriho:
I have looked through through the code as an overview and the indentation and comments seem to work correctly. However this branch has a strange bug compared to the group's current stable branch, when selecting the tool to create an entity and then trying to switch back to the selection tool to pan around the system breaks and no tool can be selected nor used

c21adano:
Dose the site look like this? There has been some problems with the local storage, where the site breaks if no local storage is currently saved.
eretst


Otherwise its sounds good. Oscard will merge later.

b21eriho:
image
It looks like this, in my case I added a few entities and then swapped back to another tool, and now the system is unresponsive.

b21eriho (approved):

PR #13169, opened 2023-04-11
Changed the color of the delete button for entities in darkmode

c21theti:


a21aldsa (approved):
the X is now easily visible in darkmode. the issue is resolved

PR #13170, opened 2023-04-11
Fixed red background in add popup

a21lukpo:
There is now no red background, only a red border that shows if the input is incorrect, also changed background color when input is correct since this was originally white and the input text couldn't be seen. This can be tested on my server version https://cms.webug.se/root/G3/students/a21lukpo/LenaSYS/DuggaSys/courseed.php


g21emmka (approved):
These changes only works on _edit course version_ and _new course version,_ on these two places it looks good and works as intended.

a21rammo:
@g21emmka the branch has conflicts FYI

g21emmka:
@a21rammo I know I have talked to @c21sebar about this, he will look into this. That's why i just left a comment about the changes that has been made on this issue.

a21rammo:
Yeah i found out that this PR is trying to change the variabel name.style.backgroundcolor. There is one PR that was merged 18 hours ago that also changed this variabel. @b21leowa and @a21lukpo will needs to sort out which is correct way of the input fields.


image

Issue #13172, opened 2023-04-11
Make a check for cssUrl to not be null


b21oscpe:
The variable "cssUrl" being null causes the page to malfunction, make a guard to prevent this.

HGustavs:
Where is this and what error does it result in?

PR #13176, opened 2023-04-11
Make a check for cssUrl to not be null #13172

a21aldsa:
added a if statement checking if localstorage isnt null


b21phiro:
Yes

b21phiro (approved):
yes

PR #13187, opened 2023-04-11
G3 2023 v1 issue#13137

c21rebja:
Test the validation for both the new course and edit course. You should not be able to create/edit the course unless all input fields are valid. Leaving the field for the github URL empty also counts as it being valid, since it's optional.

Invalid
image

Valid
image

Also valid
image

( Co-creator @g21emmka )


d21emijo (approved):
working as intended
image

image

hint needs revising in the future

Issue #13069, opened 2023-04-05
Add support for GitHub in Moments in LenaSys


c21sebar:
A course page should not pull all data from github, some folders on github should not automaticly become a task/moment/heading in LenaSys, only selected folder should be imported.

Add support to point to a folder in the github repo for the course page in the "moment". 
Can we add support in the existing "moment" type or do we need to add another "GitHub" type?
![image](https://user-images.githubusercontent.com/102595788/230053507-b34898eb-ac44-4669-a073-f0baf2f3c625.png)
![image](https://user-images.githubusercontent.com/102595788/230054862-5095e7cf-0dfa-4c98-b970-f75ff75e0e23.png)

This would be a requirement for the traversal inorder to know what filesystem to start to loop thorugh and the download and compare if changes has been made.

d21emijo:
![image](https://user-images.githubusercontent.com/102797583/231194697-1101f6e8-9259-4c3c-b726-ef65e966fade.png)
added a github icon

in the hidden section somehow github icon is slightly visable?
´´´
onclick='confirmBox(\"openConfirmBox\", this);'>`;
´´´
new onclick needs to be made for github support

d21emijo:
This button will act as a "setting/control" panel for the GitHub implementation, for future implementation the functionality of the click needs to be changed, as it now it only works in the same way as the trashcan.

PR #13188, opened 2023-04-11
G3 2023 v1 13069 add support for GitHub in moments in lenasys

d21emijo:
For issue #13069 
Added a button to hold information about the GitHub implementation, pointers to the folders, files in the code example etc.
Needs to implement functionalty of the button itself as it now only works in the same way as the trash can.
The button will act as a control/setting for the Github.


c21rebja (approved):
The button with the symbol is present everywhere, and works to click on even if it's just trashcan-functionalities. Looks good.

g21emmka:
Looks good!

PR #13224, opened 2023-04-12
#12961 Model showDugga service fixed

b21tilbe:
UML diagram fixed plus textfile with output/input/parameter


a21oscgu (changes requested):
The function "unset" is still missing among functions. I'm actually not sure if it has to be included.

There are still some duplicates found in outputs ("variantvalue", "variantsize", "savedanswer" and "highscoremode"). Is it supposed to be like this?

PR #13226, opened 2023-04-12
Added black arrow.

b21phiro:


b21eriho (changes requested):
image
This works well as long as the black_triangle is set on the first of the two ends on the line, however if the second icon is set to be the black_triangle it does not show (see image below).

image

b21eriho (approved):
Arrow now works on both ends of the line

Issue #13166, opened 2023-04-11
Buttons in the buttons-panel are hardcoded


b21oscpe:
Write an algorithm to achieve the same thing but in a more maintainable way.

HGustavs:
Screenshot showing the problem please. This issue is unclear with only textual representation.

b21oscpe:
The issue was brought up by and directly assigned to one of the developers within the group.

The issue regards the way buttons are structured/added and needs to be fixed for maintainability purposes. The issue was found in a previous issue concerned with adding a button to the buttons-panel, which proved difficult.

PR #13227, opened 2023-04-12
G1 2023 v2 #13166

b21eriho:
Refactor the functionality for deciding which entity- and relation-buttons should be shown in a manner that prevents as much repeated code and shortens the function "showEntityTypes" by approx. 200 lines


b21oscpe:
old branch

Issue #13216, opened 2023-04-12
State chart lines should only be drawn between state chart elements


b21oscpe:
It is currently possible to draw lines between state chart elements and elements of other types. Make sure it is only possible to draw lines between state chart elements.

PR #13231, opened 2023-04-12
Lines only between state chart elements #13216

a21oligu:
Removed a condition from if-statement in function addLine(). The condition caused the program to ignore the "type to type" check if _fromElement_ was of type UML_STATE.


a21aldsa (changes requested):
This only works one way: I can't draw a line between, for example, a UML element to a state element, which is good. But I can draw the line perfectly fine if I just start on the state element instead and run it to the UML element. 

I do not believe this issue is resolved.

a21aldsa (approved):
I made a mistake, it seems i was on the wrong branch. I did a sanity check and double checked my tests. This issue is resolved, it works perfectly fine. Good job

PR #13249, opened 2023-04-12
G3 2023 v2 issue#13248 - Translate ordinary GitHub URL into GitHub API URL

e21krida:
I've implemented a translation of an URL with the syntax of **_"https://github.com/user/repo"_** into an API-URL with the syntax **_"https://api.github.com/repos/user/repo/contents"_**, simply to make it easier to use in the future, so the translation of the URL doesn't have to be done manually. This because the API-URL has specific syntax that the current PHP script requires to function properly.

To test this, simply run the script with a link to any appropriately-sized (relatively small) repository, which in my testing was "https://github.com/e21krida/Webbprogrammering-Examples", put the link as the value of the variable " $url" found on line 36 in the code, for example:

    // Here you paste the appropriate link for the given repo that you wish to inspect and traverse.
    $url = 'https://github.com/e21krida/Webbprogrammering-Examples';

Relevant files: recursivetesting/FetchGithubRepo.php


a21lukpo (approved):
Everything works correctly :)

Issue #13058, opened 2023-04-05
Log service usage/distribution from dugga.js


b21kurar:
To write tests and get an understanding of how/what/when different services are used, logging code should be created, and called through the AJAXService() function in dugga.js

It may be advisable to write the code directly into the AJAXService function, or maybe a separate file called by the function. However it should be easy to comment out, as it should **not** be part of production/user usage. This is because constantly logging every action will slow the system down. As such it is advisable that the code should be easy to reactivate from dugga.js, since logging will be done for most of our work with microservices.

This should save the logs to a file, and then be followed by repeated usage of LenaSYS. 
Important to ensure that the same actions result in the same output to the log.

b21kurar:
Been doing some investigation on dugga.js and what is sent when using the system. 
kind=PDUGGA seems to do a lot of work, it appears to be the parameters that decides what dugga is opened.

b21kurar:
In dugga.js there appears to already exist a lot of console.log statements, these could be used to verify tests.

I have also created a branch where a log is printed out for each AJAXService call with the parameters.
g2-2023-v1-log_investigation

b21kurar:
The parameters in AJAXService(opt, apara, kind)

opt: Probably option, usually GET, POST or others like them
apara: array of parameters
kind: kind of LenaSYS page, such as: PDUGGA, SECTION

c21alest:
All logged data displayed in the console as shown in this comment

> Here is an example of apara values for PHP example dugga:
> [![apara.png](https://camo.githubusercontent.com/08484f340bfba8a02d1b6368a0a6303c95ea5ba37b595c96f08e758a727fc36d/68747470733a2f2f692e706f7374696d672e63632f6e7a30684d3778562f61706172612e706e67)](https://postimg.cc/wyRpG13n)

will be sent to a log.js or log.php (according to issue #13121 ) that will display this data in an easy way and also generate a csv for more in depth logging information.

c21alest:
Here is a graphical explanation of this logging solution that also tries to describe #13121 
[![dugga-js-logging-comonents-and-communication.png](https://i.postimg.cc/QMsHRbmV/dugga-js-logging-comonents-and-communication.png)](https://postimg.cc/DSjvG1tK)

b21kurar:
Discussions have been made with @c21alest and @b21hamwe, these are the results.

[...]

PR #13253, opened 2023-04-13
G2 2023 v2 issue #13058 logging to JSON files

c21alest:
Related to issue #13058

dugga.js was modified to send AJAXService data to logging.php which then stores all this data as logs.
There are two different log files, one to store only the latest log (latestlog.json) and another one log.json that stores all logs.
This log file can be deleted to make a new one.


b21hamwe (approved):
Looks alright

PR #13255, opened 2023-04-13
Added accessedservice.php_functionalrequirements.txt

c21axepe:
Added the text doc that describes the functional requirements for accessedservice.php


b21antbr:
The mindset of specifying the functional requirements is to look for where user interaction happens.

We have used if and else if checks to explain specific parts of the file and what the part actually does and how it relates to user interaction.

The code function is a bit simplified, but make sure to look for safety measures from the system such as case sensitivity or similiar protections and how these can be tested.

For each of these methods test examples have been explained as well.

d21erisu (changes requested):
I want a clear difference between Functional and Non-Functional requirements. 
Otherwise awesome!

b21ricni (changes requested):
Mixed Functional requirements and Non-func, would be nice with a fine line which is what. For exempel row 35 is functional req and 36 is non func in the txt document

Nice with the test possibilities

d21erisu (approved):
Great!

b21ricni (approved):
Great x 3

d21erisu:
There are 2 versions, both should be "saved".

Issue #13206, opened 2023-04-12
Database implementation for fetching github repository (meta-data)


a21rammo:
Add database-implementation for the github integration for the fetched data. This issue has a connection with when a teacher is fetching a repository. This issue means that with the fetched data from previous issue #13001 , there has to be an insertion into the database. This means that this issue also has the task to add extra column in correct table in the mysql database related with the fetched data to handle the insertion correct.

Add db-implementation for all columns that are fetched, check with https://github.com/HGustavs/LenaSYS/issues/13072 and https://github.com/HGustavs/LenaSYS/issues/13071 for more information on how to support them.

See FetchGithubRepo.php for more information on what data the fetch get.
https://cms.webug.se/LenaSYS/recursivetesting/FetchGithubRepo.php

https://github.com/HGustavs/LenaSYS/issues/13001

PR #13256, opened 2023-04-13
G3-2023-v2-#13206-database-implementation-for-fetching-GitHub-repository-meta-data

a21hugni:
Created a new table in init_db.sql.


a21rammo:
@a21henhe dubbelkolla databas om tabellen har skapats.

a21henhe:
> @a21henhe dubbelkolla databasen om tabellen har skapats.

Precis, ingen förändring sker i databasen bara av att byta till branchen med rätt kod i "init_db.sql" utan man behöver på något sätt köra det scriptet, det vill säga genom att installera om databasen via install.php. Först då kommer tabellen att skapas. Det innebär att webbservern kommer behöva installera om samt alla som hostar lenasys lokalt i utvecklingen kommer också behöva installera om för att få in den här förändringen. 

I övrigt ser tabellen bra ut och kan mergas enligt mig @a21rammo

a21rammo:
The pull request was reopened because it inserts GitHub metadata into a new database, a SQLite. The reason for this change is to ensure that if the metadata database crashes, it won't cause problem to other parts of the project that use the main database. This was suggested by @HGustavs for security reasons. @a21henhe you will need to review it again.

a21henhe:
This seems to work fine at least on the webserver, although when testing locally I received the following error message upon every reload (worked fine the first time entering LenaSys): 
image

This did not occur on the web server but I would maybe suggest adding "IF NOT EXISTS" to the query of creating the table gitRepos in basic.php and I would also like to fix the indendation of the same query: 
image


But the meta-database creation does work!

e21krida:
@a21henhe Good catch, we forgot to re-implement that part after troubleshooting the creation last week. I'll push with that fix now!

Issue #13035, opened 2023-04-04
CTRL z removing the first to changes simultaneously.


c21adano:
CTRL z removes the first to changes done simultaneously instead of individually. This means that when three changes have been done clicking CTRL z twice will remove all three changes.

PR #13259, opened 2023-04-13
G1 2023 v2 #13035

b21matfi:
i removed the do while loop, that was causing the issue to CTRL Z removed the two changes at the same time. Removes the first the changes ever made


d21matbo (approved):
Changes works! Looks good

Issue #13219, opened 2023-04-12
"You cant place elements too close together." but you can spawn elements inside each other.


c21adano:
If you try to place two elements ontop of each other you will get the error messages "You cant place elements too close together"
![tooclose](https://user-images.githubusercontent.com/102599332/231445018-04856e9f-689c-413c-bac3-b9feba1b8996.PNG)

But error does not seem to work for when you spawn elements ontop of each other:
![UMLontopofeachother](https://user-images.githubusercontent.com/102599332/231445164-4e29e5d0-9b4f-4c42-a879-8740401d2f1b.PNG)

Make it so elements cant be put ontop of each other.

PR #13260, opened 2023-04-13
Overlapping elements when creating #13219

a21oligu:
Changed how the _entityIsOverlapping()_ function works, updated the function-call throughout diagram.js  and changed where system-error messages for collision are generated. Solves issue #13219


b21matfi (approved):
Nice, can not place elements on top of eachother anymore.

PR #13261, opened 2023-04-13
Created profileservice.php_functional-requirements

c21axepe:
Added a txt doc that describes functional requirements and inputs/outputs


d21erisu:
There are 2 documents to review, i added a version that follows the template but the other one is in a way more detailed, so should be kept

a21oscgu (approved):
Looks good!

Issue #13221, opened 2023-04-12
Pop out function disabled in dugga diagram.


c21adano:
There used to a pop out function when holding down left click on entities that would reveal all enabled entities.

But the function was disabled in the latest weekly merge and now show all elements in a row:
![duggapopoutremoved](https://user-images.githubusercontent.com/102599332/231462184-620d931e-13c4-43eb-b2ea-6f04b66ea1b9.PNG)

The small triangel in the corner indicated which elements were part of a pop out.

PR #13263, opened 2023-04-13
G1 2023 v2 #13221

b21phiro:
Pop out function disabled in dugga diagram. #13221.

NOTE: Accidently committed for black arrow.


d21matbo (changes requested):
This does not exactly solve the issue...

For starters, diagram_dugga previously worked by getting information from the JSON string in the database table 'variant', where it only loads the notations used for the current dugga, i.e. ER and UML. Now everything is loaded which is not correct.

If we overlook the above part the following things does not work 
1. The changes both affect diagram.php and diagramdugga
2. All notations show untill one specific is selected from the pop-out
3. The state-start and state-stop are always visible
4. After selecting the state-start/stop from the pop-out it is not possible to change back
5. It is possible to have multiple pop-out windows show at the same time (see figure)
image

d21matbo:
This is the JSON part previously used to dynamically create the pop-out windows for diagram_dugga; {"type":"md","filelink":"HTML_Ex1.txt","gType":"md","gFilelink":"HTML_Ex1.txt","diagram_File":"example_ER.json","diagram_type":{"ER":true,"UML":true},"extraparam":"","notes":"","submissions":[{"type":"pdf","fieldname":"","instruction":""}],"errorActive":true}

a21oligu:
I've done some changes that should be reviewed @d21matbo.

d21matbo (changes requested):
#### From previous review:
1. The changes both affect diagram.php and diagramdugga (Might acually be better like this)
2. ~~All notations show untill one specific is selected from the pop-out~~
3. The state-start and state-stop are always visible
4. After selecting the state-start/stop from the pop-out it is not possible to change back
5. ~~It is possible to have multiple pop-out windows show at the same time~~

#### New: 
6. The Placements of the pop-outs doesn't match with the buttons (depends on screen size)

Points 3 and 4 needs fixing, 1 and 6 are less important

Issue #13211, opened 2023-04-12
Change entity outline depending on colour and mode.


c21adano:
currently Er entities has support for changing outline to white when darkmode is activated, the other two diagram types should also have this functionality.

![blackwithwhiteconners](https://user-images.githubusercontent.com/102599332/231430480-bfb47aea-8618-40f3-a2fa-9471240b5e01.PNG)

But there is also an issue with the Er white outline change:
![whitewithwhiteconners](https://user-images.githubusercontent.com/102599332/231430496-62acae49-db20-4599-be7e-467b25ca7a95.PNG)

The outline changes colour based on if the dugga site is in darkmode or lightmode, but the diagram part of the site has its own darkmode button that changes only the diagram canvas to darkmode.
![darkmodesiwths](https://user-images.githubusercontent.com/102599332/231430704-aa9ee847-03db-40be-bfbe-c29f137593dd.PNG)

Link the outline switch to the diagram canvas button instead of the dugga darkmode button.
![switchtorightdarkmode](https://user-images.githubusercontent.com/102599332/231430722-5fa97e7f-37e7-4a50-bb21-39abdbde94d2.PNG)

PR #13265, opened 2023-04-13
Change entity outline depending on colour and mode. #13211

a19clasv:
Added class 'text' to IE and UML parts that have a stroke color that can possibly be toggled and made the borders change as soon as dark mode is toggled in the diagram.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -4203,7 +4203,7 @@ function toggleDarkmode()
         stylesheet.href = "../Shared/css/blackTheme.css";
         localStorage.setItem('diagramTheme',stylesheet.href)
     }
-    updateCSSForAllElements()
+    toggleBorderOfElements();
 }
 
 /**
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -4203,7 +4203,6 @@ function toggleDarkmode()
         stylesheet.href = "../Shared/css/blackTheme.css";
         localStorage.setItem('diagramTheme',stylesheet.href)
     }
-    toggleBorderOfElements();
 }
 
 /**
@@ -11391,6 +11390,7 @@ function updateCSSForAllElements()
         //If more than one element is marked.
         return inContext && context.length > 1 || inContext && context.length > 0 && contextLine.length > 0;
     }
+    toggleBorderOfElements();
 }
 /**
  * @description toggles the border of all elements to white or gray; depending on current theme.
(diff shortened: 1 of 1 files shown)


a21aldsa (commented):
Sure, the issue is resolved. But the issue itself is nonsensical, why would anyone want white borders on a white element? This defeats the purpose of borders, I have attached screenshots that show what I mean. @c21adano @b21oscpe 

image

image

c21adano:
The outline should be white in darkmode if the entity itself has a black colour selected. Its seems i described the first part of the issue poorly and i apologize. If the entity has a white colour it should have the normal outline.

a19clasv:
I fixed it

DuggaSys/diagram.js
```diff
@@ -11410,21 +11411,23 @@ function toggleBorderOfElements() {
...
                 //assign their current stroke color to a variable.
-                let strokeColor = text.getAttribute('stroke');
-                //if the element has a stroke which has the color #383737: set it to white.
-                //this is because we dont want to affect the strokes that are null or other colors.
-                if (strokeColor == '#383737') {
+              let strokeColor = text.getAttribute('stroke');
+                let fillColor = text.getAttribute('fill');
+                //if the element has a stroke which has the color #383737 and its fill isn't white: set it to white.
+                //this is because we dont want to affect the strokes that are null or other colors and have a contrasting border.
+                if (strokeColor == '#383737' && fillColor != '#ffffff') {
```
a21aldsa, line 11418: if its grey, its automatically not white. that means this check is extraneous, or am I misunderstanding something?

a21aldsa (changes requested):
While yes, the border becomes white when the color of the element is set to anything other than white and the theme is dark; you forgot to account for other element colors such as green or pink. Your changes now makes it very difficult to see the border on these elements.

image

This will lead to some users having a hard time differentiating between, for example, the different fields in a UML diagram.

Albeit, this might very well be a separate issue in and of itself. If this is the case I am ok with calling this issue resolved.

a21aldsa:
Also, while it is very easy to see the inside border color of a white element on darkmode now, it is very hard to see the outside border. I believe this should also be resolved or taken up in another issue.

c21adano:
> Also, while it is very easy to see the inside border color of a white element on darkmode now, it is very hard to see the outside border. I believe this should also be resolved or taken up in another issue.
> 
> image

This could be an issue in the future but doesn't need an immediate fix.

a19clasv:
Just give me another issue or expand this issue, stating which border colours should be according to which element colours and which mode. Otherwise this issue is resolved.

b21oscpe:
Discussion provides future issues, the original issue is resolved

PR #13266, opened 2023-04-13
added an error catch

b21matfi:
Checked diagram.php if there are function that can be moved to diagramservices.php but one was to fetch the data from diagram services and the other to get the error message. In diagramservices i added and error handler where it was missing. By using getMessage(); which returns the description of the error.


b21eddka:
I approve of this pull request.
image

Issue #13236, opened 2023-04-12
Document functional requirements for courseedservice


b21kurar:
Related to issue #13208

PR #13271, opened 2023-04-13
G2 2023 v2 issue #13236 courseedservice.php functional requirements

c21alest:
Related to #13236


a20fanma:
There is a template called "FRD_Mall" which show how to represent functional requirements.

a20fanma:
Everythings looks good

PR #13272, opened 2023-04-13
G3 2023 v2 issue#13222

g21emmka:
Fixed so that the github url is inserted into the database when creating or editing course. 

Test this by creating a new course and see if the github url is visibile in the database column, also check if the url changes when you insert a different url in editing course. 

( Co-author @c21rebja )

### DIFF
--- a/DuggaSys/courseed.js
+++ b/DuggaSys/courseed.js
@@ -80,10 +80,8 @@ function createNewCourse()
 	var courseGitURL = $("#ncoursegit-url").val();
 	$("#newCourse").css("display", "none");
 	//$("#overlay").css("display", "none");
-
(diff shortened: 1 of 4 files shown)


b21leowa:
When I added a link to the a new course it added the GitURL into the database, it works. But when i added a link the result was not as I expected. The result is right now:

image
This is maybe another issue though.

It also worked with editing the course and the result is the same as above.

c21rebja:
Yep, the issue is just to upload the link. We saw that the formatting of the link was wrong, but that's an issue for later. :) @b21leowa 

We talked about this with @c21sebar. @a21rammo could you create an issue for this if there isn't one already?

b21leowa (approved):
I approve these changes :+1:

PR #13273, opened 2023-04-13
Made text white for ER-table when in darkmode, Issue#13257

b21eddka:
When you toggle ER-table (can be accessed through key "E") on an ER element the text appears as white when in darkmode.


d21matbo (approved):
Changes works! Looks good

Issue #13196, opened 2023-04-12
Markdown php vs js versions


HGustavs:
Are we still using two different markdown versions, one in php and one in javascript?

In the case that we still have two implementations, do we have a testing case (preferably automated) to test if outputs of each are identical?

b21geoyo:
I was under the impression that there was a file named "markdown.php" in the repository, but after searching through GitHub history, including commits and branches, as well as using Git log commands, it seems like the file doesn't exist, or at least not anymore. However, I did find another PHP file named "showdoc.php" that seems to be similar to the "markdown.js" file. I'm not entirely sure what the "showdoc.php" file is used for. I will look into it further.

b21geoyo:
In the commit [Removed duplicate functions](https://github.com/HGustavs/LenaSYS/commit/101798f05cb4748d2b05338df8be8196c67467a0), I removed the unnecessary duplicate Markdown functions in "showdoc.php" since the file already had a script tag link to "markdown.js". I tested the file to ensure it worked as expected without the duplicate functions. As for the questions asked earlier, it appears we had two implementations of Markdown, one in PHP and one in JavaScript. However, with the removal of the duplicate functions and use of only "markdown.js", we now only have one implementation of Markdown, which eliminates the need for automated testing to verify output consistency between PHP and JS versions.

c21sebar:
> Are we still using two different markdown versions, one in php and one in javascript?
There are some support or duplication of functions from markdown.js in showdoc.php but theese functions aren't exactly the same and seems to never be used. The new implemented GitHub style of markdown has been implemented only in markdown.js aswell.

> In the case that we still have two implementations, do we have a testing case (preferably automated) to test if outputs of each are identical?
No need for a testing case if only one of the two are used, right now it looks and seems like only markdown.js is used to display markdown files and never showdoc.php. So no need for an automated test to check if both php and javascript outputs the same.

Issue https://github.com/HGustavs/LenaSYS/issues/13326 will cleanup showdoc.php and keep only the javascript version.

PR #13275, opened 2023-04-13
G3 2023 v2 issue#13196-issue#13326

b21geoyo:
I removed the unnecessary duplicate Markdown functions in "showdoc.php" since the file already had a script tag link to "markdown.js". I tested the file to ensure it worked as expected without the duplicate functions. As for the questions asked earlier, it appears we had two implementations of Markdown, one in PHP and one in JavaScript. However, with the removal of the duplicate functions and use of only "markdown.js", we now only have one implementation of Markdown, which eliminates the need for automated testing to verify output consistency between PHP and JS versions.

You can test showdoc.php via the link below, make sure you are logged in LenaSYS: 

https://cms.webug.se/root/G3/students/b21geoyo/LenaSYS/DuggaSys/showdoc.php?cid=1885&fname=Test.md


a21olljo:
@b21geoyo Har du kommit åt showdoc.php i tidigare iterationer av LenaSYS? Testar för och se hur showdoc.php såg ut tidigare och kommer inte åt och ser filen på någon annan branch än din, andra versioner ger bara "Bummer! You have reached a non-navigable link!" https://cms.webug.se/root/G3/LenaSYS/DuggaSys/showdoc.php

a21olljo:
Just wondering if you have answered the questions asked in the issue #13196? The questions where:

"Are we still using two different markdown versions, one in php and one in javascript?"
This one is sort of answered with you finding the "showdoc.php" and confirming it handels markdown. But only partially because you found redundant functions in it and "markdown.js"? 

"In the case that we still have two implementations, do we have a testing case (preferably automated) to test if outputs of each are identical?"
Here you made manual testing, giving the outcome that some of the outputs are identical, why you removed the identical parts? But that is not really how a "testing case" is done (https://en.wikipedia.org/wiki/Test_case) After looking at the codes I agree with you that there is some redundancy but it is probably there for a reason which would become a different issue. 

Then I understand why you wanted to indent the code, but it makes the jobb of testing and knowing what has been change a pain, right now it says that 500 lines where removed and 200 added.
Removing code was also not a part of the issue, removing over 300 lines seems a bit much for a issue that were suppose to investigate if we had two file and test there relation.  

Not trying to be rude but understand what has been done.

c21sebar:
After comparing this branch with our current master and our group branch nothing seems to be different or have stopped working.

The php markdown support in showdoc.php seems to not be used only the javascript file. There would be no reason to duplicate the javascript version for php when/if the php version is never used.
And the implementation of the GitHub markdown style has been made in markdown.js and not in showdoc.php?
Is my understanding of this correct? @HGustavs @b21leowa @a21henhe 

If so we can merge this and remove the comment that says "Identical php code exists in showdoc any changes must be propagated" since showdoc isn't used?

Since issue https://github.com/HGustavs/LenaSYS/issues/13196 is not about removing the php version a new issue should be created and linked to this pull request **if** we land on deleting the php version that isn't in use.

And https://github.com/HGustavs/LenaSYS/issues/13196 should be answered in that issue as comments.

b21leowa (approved):
:+1: With the latest update from @b21geoyo, the website works as intended and there is no problem to be found for me. Please verify and test as well @a21olljo

Issue #13264, opened 2023-04-13
Add JSON files to the project


b21oscpe:
Files such as the json-file found in #13213 and the like should be added to the project so that they can be referenced when we load in a diagram through the dropdown menu.

Add the available files, (at the time of writing this issue: UML, IE), to the project at a suitable place

Files can be found in their respective issue

Issue #13212, opened 2023-04-12
Create mock-up diagram: ER


b21oscpe:
This issue is related to #13204 

NOTE: This issue regards creating an ER model.

We want to be able to load a mock-up of a selected modeling language from a dropdown menu.

The purpose of this is to showcase the elements included in the modeling language as well as simplify development testing.

Instructions:

1. Create an example model in diagram.php with the modeling tools available. Make sure to use as many tools as possible to showcase all functionality. (You are allowed to look at models online for reference but keep in mind that these may have copyright so copying is not allowed.)
2. Export the model through **Options -> Export**.
3. Leave the output as a comment to this issue.

The exported file/diagram will be used as "value" for the dropdown later.

c21theti:
[ERdiagramexample.zip](https://github.com/HGustavs/LenaSYS/files/11232486/ERdiagramexample.zip)
![image](https://user-images.githubusercontent.com/102578329/232037525-2ce79fd7-beec-49ef-b351-67da6986ba26.png)
Diagram is finished and contains all elements included for ER diagram

c21theti:
![image](https://user-images.githubusercontent.com/102578329/232039495-c3c40511-8fd2-47fd-a29e-8492fb9d27d1.png)
[exampleERdiagram.zip](https://github.com/HGustavs/LenaSYS/files/11232599/exampleERdiagram.zip)
ER mock-up diagram is finished with all ER elements included

PR #13284, opened 2023-04-14
 Add JSON files to the project #13264 

a21aldsa:
Uploaded a couple JSON files holding diagrams to the server. Created a function called loadMockupDiagram which takes the parameter to determine the type of mockup that should be loaded, IE or UML (when future mockups are created, they can be easily added). And then fetches this file from the server, and calls loadDiagram with that json file as parameter.


b21matfi (approved):
checked by calling the function  loadMockupDiagram with parameters 0 and 1, it loaded different diagrams.

b21oscpe:
Putting this merge on hold as we discuss the posibilities of the files being stored on the server instead

a21nibah:
This is an elegant solution. 2 things. Please restructure your parameters so it is not dependent on numbers but rather are more clear. The er type is missing.

a21nibah:
actually now that i think about it  I believe it would be better if you just had the path as a parameter to make this reusable and then you can specify the path you want instead

a21aldsa:
> This is an elegant solution. 2 things. Please restructure your parameters so it is not dependent on numbers but rather are more clear. The er type is missing.

The er type is missing since the issue for creating the ER json is not completed, #13212.

a21aldsa:
> actually now that i think about it I believe it would be better if you just had the path as a parameter to make this reusable and then you can specify the path you want instead

I will implement this

a21aldsa:
> actually now that i think about it I believe it would be better if you just had the path as a parameter to make this reusable and then you can specify the path you want instead

I have now implemented this feedback. The function now wants a path to a json file as parameter, if the parameter is faulty the .catch of the fetch will catch this.

b21matfi (approved):
testing with loadMockupDiagram("JSON/UMLDiagramMockup.json") works but merge is still going to be put on hold untill all the files are done

a21aldsa:
> testing with loadMockupDiagram("JSON/UMLDiagramMockup.json") works but merge is still going to be put on hold untill all the files are done

The ER diagram just got completed, in issue #13212 I have added it. Note that #13212 is untested so if that test comes back negative this commit will have to be reverted. I do not recommend merging just yet.

b21oscpe:
I would suggest you test to see if the JSON file provided in #13212 is loadable and if it is we can merge

b21matfi:
im going to test it today

b21matfi:
can you make another request so that i can approve or comment?

a21aldsa:
> I would suggest you test to see if the JSON file provided in #13212 is loadable and if it is we can merge

It loads just fine, although it is worth noting that the ER diagram and the IE are way off center.

a21aldsa:
The ER diagram in #13212 has been updated, I will make a new commit for this new, updated, file.

b21matfi:
a21aldsa do you want me to wait with the testing until you are done?

a21aldsa:
> a21aldsa do you want me to wait with the testing until you are done?

It has now been updated to the new ER model.

b21matfi (approved):
All of them are loadable, Works good. Nice Work. Uses the file path as intended

PR #13295, opened 2023-04-14
Check if emojis works in a markdown file. 

b21leowa:
Go to this page and create a markdown file by adding a dummy file and add .md in the end of the file.

The try to do emojis by typing the following:  :D  :) ;) :( :'( :P :/ :o <3 (Y) (N)

It should look like something like this: 

image

### DIFF
--- a/Shared/markdown.js
+++ b/Shared/markdown.js
@@ -431,17 +431,17 @@ function markdownBlock(inString)
 
     //Markdown smileys
     //Supported: :D :) ;) :( :'( :P :/ :o <3 (Y) (N)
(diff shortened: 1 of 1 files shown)


d21emijo (commented):
# with changes it works as below
With theese changes you dont need a whitespace to create a smily, will this be an issue that one will create unintentional smilys?
image

# current way smilys work below
in the example below is how its currently works where you need a whitespace before and after to create a smily
image

a21rammo:
Looks good, if there is no issues with merging you can click the Merge PR button @d21emijo

c21sebar:
We keep the current emoji support with whitespaces before ":D" to not cause issues in text while write for example "1<3" .

Issue #13203, opened 2023-04-12
Cardinalities need improvement in darkmode


c21adano:
Cardinalities currently has a white box around them in darkmode while also being black numbers.
![cardinalitiydarkmode](https://user-images.githubusercontent.com/102599332/231420915-c22bc354-c154-41ce-adf2-22ae5598c917.PNG)

We want them to be white numbers without a black background.

PR #13297, opened 2023-04-14
G1 2023 v2 #13203 cardinalities in dark-mode

b21eriho:
Cardinality labels on lines are now white on a black background in darkmode as they should instead of black on white background like in lightmode.


b21matfi (approved):
Looks good, dark back ground and white text in darkmode and the other way around in light mode

Issue #13282, opened 2023-04-14
icons in darkmode


c21adano:
Currently alot of the icons that can be on lines dont have darkmode versions. Seem like triangle arrow is the only one that does and weak doesn't need one.
![iconslightmode](https://user-images.githubusercontent.com/102599332/231985750-291791eb-dd76-4863-81a7-69df8af8c247.PNG)

![iconsindarkmode](https://user-images.githubusercontent.com/102599332/231985764-61680d97-0a8d-4a5b-9ca8-9427f6d80534.PNG)

Issue #13209, opened 2023-04-12
add a blackarrow


c21adano:
there is currently a triangle arrow icon that can be used for class, IE and state diagrams. We want a black version of the triangle arrow.

It should look something like this:
![blacktrianglearrowforstate](https://user-images.githubusercontent.com/102599332/231426548-4a148a5f-7259-49a5-89bd-9de926b62606.PNG)

How the white arrow looks in darkmode is the way the black version should look in darkmode. So reuse the darkmode version from the white arrow.
![blackarrow](https://user-images.githubusercontent.com/102599332/231424454-0503a901-2982-4a19-8c54-df72361cf7a0.PNG)

This is how the white arrow looks in lightmode, the black version should just be black inside the triangle.
![whitearrow](https://user-images.githubusercontent.com/102599332/231424470-a84f4a17-d4fc-4f0f-8f89-910c9d6d85ae.PNG)

b21oscpe:
b21phiro cannot continue this branch as of now, it needs to be complemented.

PR #13298, opened 2023-04-14
G1 2023 v2 #13282

b21matfi:
Used css to fix the issue

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -4205,13 +4205,11 @@ function toggleDarkmode()
     if(stylesheet.href.includes('blackTheme')){
         // if it's dark -> go light
         stylesheet.href = "../Shared/css/style.css";
-        lineColor = '#000000';
         localStorage.setItem('diagramTheme',stylesheet.href)
     } else if(stylesheet.href.includes('style')) {
         // if it's light -> go dark
         stylesheet.href = "../Shared/css/blackTheme.css";
         localStorage.setItem('diagramTheme',stylesheet.href)
-        lineColor = '#FFFFFF';
     }
 }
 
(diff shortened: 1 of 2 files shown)


b21eriho:
Which icons have you fixed? All but the triangle arrow and weak?

c21adano:
I fixed the conflict on line 1101, it was caused by an extra space from issue#13226. Please test again to see that nothing broke.

b21matfi:
all should be fixed that were not fixed before

b21eriho (changes requested):
All looks good apart from the black triangle which looks like the normal triangle
image

c21adano:
> All looks good apart from the black triangle which looks like the normal triangle image

this is caused by issue#13226 adding a black triangle. It doesn't have to be fixed in this issue.

b21eriho:
Now this clearly overlaps with issue #13209 and how this should be handled I do not know

c21adano:
> Now this clearly overlaps with issue #13209 and how this should be handled I do not know

#13226 and #13209 are the same issue, #13226 is the merged pull request

b21eriho:
Yes, in either case the black arrow does not look right though I think it looked right previously since issue #13226 was merged, so this seems to me like a conflict-solving issue

b21eriho:
I missed you comment with the image, in that case I'll approve my review

b21eriho (approved):

c21adano:
to clarify #13226/#13209 was the issue that this issue #13282/#13298 had a conflict with. After resolving the conflict the black arrow changes were merge into this issue. Therefor i rule that it would be unfair to push improvements that can be done to #13226/#13209 onto b21matfi.

I will now merge this into g1-2023-v2, then an issue regarding the black arrow will be made.

PR #13302, opened 2023-04-14
created FR contribution_loginbox_service

a20fanma:
created functional requirements for contribution_loginbox_service.php


c21alest (changes requested):
FR2: I would argue that there is functionality that only allows a git user creation if user don't already exist. As of now a user can be created even if it exists.

Can't see any requirements for _requestContributionUserLogin_ line 303
And also no logout requirements line 457.

c21alest:
Next time click the Re-request review instead of messaging me to review this because otherwise I can't request new changes @a21oscgu @a20fanma. But it now looks okay. I would have liked to see more information in the Where, While... fields but this seems enough to describe the FR.

b21kurar (approved):

c21alest (approved):

Issue #13290, opened 2023-04-14
Reset view button breaks the ruler.


c21adano:
Pressing the reset view button breaks the ruler:
![camrabutton](https://user-images.githubusercontent.com/102599332/231993789-dbb7cb2e-cb58-4e93-96e5-fd2815a70fd8.PNG)
![ruler](https://user-images.githubusercontent.com/102599332/231993870-829a95fe-62d4-4757-a256-92913112c5a7.PNG)

PR #13303, opened 2023-04-14
G1 2023 v2 #13290

d21matbo:
Identified that the issue only occurs when no elements are placed on the canvas. 
Added a check that stops the function from executing if no elements are placed


b21oscpe:
#13290

b21matfi (approved):
pressing the reset button does not break the ruler anymore, everything seems to be working fine

Issue #13287, opened 2023-04-14
Include button opens the er-table function


c21adano:
There is an include button on the left side menu:
![includeopensertable](https://user-images.githubusercontent.com/102599332/231990406-febf457a-88b5-4e72-8b8e-16622b0dd786.PNG)

When the button is pressed it opens the er-table function that is also linked to the er-table button:
![ertablefromincludebutton](https://user-images.githubusercontent.com/102599332/231990419-4d5aa159-5509-41b3-8eda-b2fa932977e1.PNG)

The include button is suppose to open some for of special characters menu. Im unsure of if there is support or not for that in the code right now, if there is no support repport it. Otherwise if there is support for it link it to the button.

a21aldsa:
> When the button is pressed it opens the er-table function
I cannot replicate this on my machine. I've tried Firefox and Chrome. This does not directly affect my ability to solve this issue, but it is still noteworthy.

c21adano:
You must have highlighted an er element for the table to open up, it will appear in the option field

a21aldsa:
I see it now.

After some digging I have found 0 indication of other intended functionality for the include button. Perhaps the intention was simply a tool tip telling the user how to use special characters?

Aside from that, its current functionality,  toggleErTable(), does not make sense to me. I do not see a point in displaying the er table in a "string format", especially when the user can't even copy this text or otherwise use it?

My best guess is that this is trying to be a relational data model, see the attached screenshot.

![image](https://user-images.githubusercontent.com/102581306/232011065-40ddb979-5e2a-43a2-8a4c-f738b56b703a.png)

But, if this theory is true, this means this functionality is incredibly unfinished since even this simple model is not represented at all like it should be, see the attached screenshot below.

![image](https://user-images.githubusercontent.com/102581306/232011342-b68b31a9-0b60-46f9-b552-da7825e60782.png)

My final assessment is that this button should be given a better purpose or simply removed. I also strongly feel that toggleErTable() and subsequently, generateErTableString(), should be looked at more closely to determine exactly what the intended functionality is, if it turns out the idea was to have a way to generate relational data models from ER diagrams, this functionlity needs to be improved. Otherwise I see no reason to keep them.

c21adano:
The er table will be developed further, there have already been some issues regarding it but more will follow. For the moment you could disable the include button, we can developed in the future but right now its of no concern.

a21aldsa:
[...]

PR #13304, opened 2023-04-14
Include button opens the er-table function #13287 

a21aldsa:
commented out the entire fieldset for the include button


b21eddka:
I approve of this pull request, the button does not exist anymore and is instead commented out for future development.
image

Issue #13300, opened 2023-04-14
Black and white triangle arrows look the same in darkmode. 


c21adano:
![232022261-39edf07f-7bf0-4bef-a573-adbe11e98a7f](https://user-images.githubusercontent.com/102599332/232025434-6c7ade3e-aa39-46e0-ba2d-796490be5b6d.png)

Both triangle arrows need to be distinct in both darkmode and lightmode, but right now both of them are white in darkmode. The black version should have a black filling in darkmode so you can still see that it is the black version.

PR #13307, opened 2023-04-14
G1 2023 v2 #13300

b21matfi:
Changed css for the black triangel, had the same as the normal one


c21theti:
Looks good, black triangle is fixed and has black color with white outline. 
image
though the arrow has some small space between the entity and the line that could be fixed

b21matfi:
That the small space of the entity is how it is created on on the line for the triangle. I think that the problem is in the **zoomfact** variable and how it is created.

b21matfi:
it can also be that it is not overlapping to the other entity. connected right where the other entity ends

PR #13310, opened 2023-04-14
Create FRD accessedservice.md

b21ricni:
Remade accessedservice.php.pdf to a md file
Left the pdf file in, was unsure if to delete it or keep it


a21oscgu (approved):
Looks good!

PR #13311, opened 2023-04-14
Create FRD showDuggaservice.md

b21ricni:
Replacement for FRD_showDUGGAservice.php.PDF


b21ricni:
future issue is to clean up PDF files, since testers need them to check everything transferred correctly

b21tilbe (approved):
Everything seems to be in order. Approved

PR #13318, opened 2023-04-17
Create highscoreservice.php.md

a21oscgu:


a19hakja:
Usually you'd provide a short description with your pull request, but it's not a big deal here. Assigning a reviewer.

a20fanma (changes requested):
Add a test proposal, to make sure the function works. For example, "try to answear a dugga, to see if the teacher is able to view the score"

Issue #13308, opened 2023-04-14
Arrow not scaling right when zooming in and out


c21adano:
If you have two arrow icons on a line one of them gets distorted when zooming in or out.
![arrowzoomedin](https://user-images.githubusercontent.com/102599332/232056875-f773a0e1-7dc6-435c-b3f3-a43ff6f8f159.PNG)
![zoomedout](https://user-images.githubusercontent.com/102599332/232056926-4576a4c3-a122-424b-9dea-7ce02f0de21b.PNG)

PR #13320, opened 2023-04-17
G1 2023 v2 #13308

b21matfi:
Added zoomfact to the end arrow, the arrow did not have zoomfact and thats why it was not working as intended


a21aldsa (approved):
Looks good, a line with arrows on both sides is now consistent and and both arrows scale properly, even when zoomed all the way out/in. I believe this issue is resolved.

PR #13322, opened 2023-04-17
Made repo unique in database

a21lukpo:
Modified the database to make repoSHA  unique to avoid duplicated inserts.

Check in server cmd by using "cd /var/www/html/root/G3/students/e21krida/githubMetadata" then "sqlite3 metadata2.db" and finally "select repoName from gitRepos ;"

Now there should only be one repo with the files instead of multiple repos with the same files

### DIFF
--- a/Shared/basic.php
+++ b/Shared/basic.php
@@ -252,7 +252,7 @@ $sql2 = '
     repoURL VARCHAR(255), 
     repoFileType VARCHAR(50), 
     repoDownloadURL VARCHAR(255), 
-    repoSHA VARCHAR(255), 
+    repoSHA VARCHAR(255) UNIQUE, 
     repoPath VARCHAR(255) 
 	);
 '; 


a21rammo:
@HGustavs Where can i find the credentials or information to connect to the sqlite database through datagrip/mysqlworkbench?
```
 $log_db->exec($sql);
try {
	$metadata_db = new PDO('sqlite:../../githubMetadata/metadata'.$metadataDbVersion.'.db');
} catch (PDOException $e) {
	echo "Failed to connect to the database";
	throw $e;
} 

$sql2 = '
	CREATE TABLE IF NOT EXISTS gitRepos ( 
		repoID INTEGER PRIMARY KEY AUTOINCREMENT,
		repoName VARCHAR(50), 
    repoURL VARCHAR(255), 
    repoFileType VARCHAR(50), 
    repoDownloadURL VARCHAR(255), 
    repoSHA VARCHAR(255) UNIQUE, 
    repoPath VARCHAR(255) 
	);
'; 
$metadata_db->exec($sql2);```

a21olljo:
Looks good and works, a simple and elegant solution to make sure that there only exists one version of a repo,  just wondering if you have taken into consideration multiple versions of the same repo, a new version of a repo also means a new sha code by the hash algorithm. 

Maybe using the path as unique will guarantee no duplicates?

a21olljo:
Looks good and works to make sure that there only exists one version of a repo, just wondering if you have taken into consideration multiple versions of the same repo, a new version of a repo also means a new sha code by the hash algorithm.

e21krida:
> 

We have not, this will have to be handled by a future issue. This is simply to stop duplicated SHA-files from existing, if we want multiple duplicates of the exact same repository then this will require additional functionality implemented through a seperate issue. But if they're not unique, it will cause some issues with using the table to begin with as we then have no unique value outside of the auto_incremented ID, making it a whole lot harder to work with the database to begin with.

a21olljo:
Okej, if its only temporary and to ensure that no duplicate entries of the same sha will occur. Thought that the issue was about stopping all duplicate entries of the same file, despite iteration. But if the exchange of old file and new will be done later in another issue that's fine.

PR #13325, opened 2023-04-17
Functional Requirements pdf#13238

b21tilbe:
Deleted the old functional recuirements pdf and added a new finished version.


a21jeaha:
@b21tilbe where can i find the added file, i cannot seem to find it..

a21jeaha (changes requested):
the file should be located in the right folder

Might be a nitpick but the header should be in English 

Functional requirements look ok.
Test Proposal look ok.

FR1 and FR2

don't they both need the same thing. What I mean with that is that, FR1 still needs an "opt" parameter and FR2 needs permission to manage quizzes. 


FR3
the text feels a bit of, i have supplied some feedback bellow.

Shall: shall be used to control access to quiz
While: While a user is authentication and is authorized, they can manage quiz's
Where: this one feels a bit off, shouldn't it be more in line with. "users with right credentials can access..... and so on? 
If Then: If user is authenticated but not authorized, then certain parts of.....

RF4 
shouldn't there be another FR, retrieving and print information?

I'm not so sure about the last one but that's just my opinion.

b21tilbe:
Was not sure if a forth FR was needed for displaying output. But give me a heads up and i can add it later if you want to.

Issue #13309, opened 2023-04-14
Lable in darkmode is hard to see


c21adano:
When you lable a line in darkmode the text appears black on black background
![image (3)](https://user-images.githubusercontent.com/102599332/232057376-2ad784f3-7f00-42f3-8e66-db8de2a41b04.png)

PR #13327, opened 2023-04-17
G1 2023 v2 #13309

d21matbo:
Changed the color of the text to use its css style


c21theti:
Looks good. Label is now visible and in white color

Issue #13281, opened 2023-04-14
Change the REGEXP for Github URL


g21emmka:
Change the regexp for the github url by removing the last part (.git), per the customer´s request.

PR #13332, opened 2023-04-17
Deleted .git from regex

g21emmka:
Issue #13281 
Removed the ".git" part from the regex on the git input fields. Test this by entering for example "https://github.com/HGustavs/Webbprogrammering-Examples" into the input field on both create and edit course. 

image

( Co-creator @c21rebja )

### DIFF
--- a/DuggaSys/courseed.js
+++ b/DuggaSys/courseed.js
@@ -562,7 +562,7 @@ function setActiveCodes() {
 const regex = {
 	coursename: /^[A-ZÅÄÖa-zåäö]+( ?(- ?)?[A-ZÅÄÖa-zåäö]+)*$/,
 	coursecode: /^[a-zA-Z]{2}\d{3}[a-zA-Z]{1}$/,
-	courseGitURL: /^(https?:\/\/)?(github)(\.com\/)([\w-]*)\/([\w-]*)\/*?$/
+	courseGitURL: /^(https?:\/\/)?(github)(\.com\/)([\w-]*\/)([\w-]+)$/
 };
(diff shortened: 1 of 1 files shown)


b21geoyo (changes requested):
I tested the regex for the GitHub repository URL on the input field. Unfortunately, the regex in the input field is not specific enough and could potentially match URLs that are not actually GitHub repository URLs. For example, the regex could mistakenly allow URLs like the ones in the images below, which are not valid GitHub repository URLs.

- The URL "https://www.github.com" is not a repo URL because it lacks the necessary information to identify a specific GitHub repository.

   image

- The URL "https://github.com/HGustavs" is not a valid GitHub repository URL, as it is missing the repository name component.

  image

- The URL "github.com/my user/test repo" is not a valid GitHub repo URL due to spaces in the username and repository name. Also it does not include "https://" in the URL.

   image
   
In conclusion, the regex used for validating GitHub repository URLs has some issues that need to be addressed. One of the main problems is that it's not specific enough and matches anything that has "github.com" in it, including non-repository URLs.

c21rebja:
@b21geoyo It's ready for review again. Having "https://" in the url is optional since it isn't necessary to have to reach the page. Also, when you copy the link from the searchbar, "https://" is excluded from the link. The other changes are implemented.

/ @c21rebja & @g21emmka

b21geoyo (approved):
The previous issues with the GitHub repo URL regex are now fixed. I did a git pull on my own server to test the new version due to the old regex version still present on your servers.

PR #13335, opened 2023-04-17
overlapping elements are not saved in historyLog anymore. Issue#13306

b21eddka:
When creating elements that overlap already existing elements, it is not saved in historyLog. As a result Ctrl + Z works as intended.


d21matbo (approved):
Looks good! Ctrl+Z does indeed work.

Issue #13215, opened 2023-04-12
Two of the same entity type in diagram.js and diagram.php


c21adano:
Related: #13074 and #13050
Last version two state diagram issues were made to implement state diagram functions, an oversight was made and the two codes slightly conflicted with each other. Both issues created their own entity type for state diagrams:
![twoofthesameentitytype](https://user-images.githubusercontent.com/102599332/231435734-6f589013-fc8e-42d8-a72c-1f7f2724797c.PNG)
They should be merged into one entity type that will represent state diagrams from now on. UML_STATE should be merge into the sd entity type and everything that uses the UML_STATE type should be move to use the sd type instead.

![stateconflict](https://user-images.githubusercontent.com/102599332/231437288-3dfe9472-9dd1-477a-8c8e-9eade48b1b0b.PNG)

Buttons linked to the sd and UML_STATE types also conflict:
SDState and UMLInitialstate also use the same number which causes conflicts, change the numbers so all button can exist on the site at the same time.
![sdandumlinitialstateusesamenuber](https://user-images.githubusercontent.com/102599332/231437266-ca858272-8ae5-4d9a-9ed1-b4c8460313c9.PNG)

The button for state diagram linked to the sd entity type has been commented out for now due to this and should be uncommented when the number changes. 
after the number is changed the setElementPlacementType of each related button should also be changed to match their new number.
![sdcommentedout](https://user-images.githubusercontent.com/102599332/231437355-369a1658-f1bf-41cd-af7f-860d867c0ce0.PNG)

a21nibah:
but if umlInitialState is part of the code that was conflicted which was indicated from the comments in the java script code does it not make sence to just remove those buttons instead? that is what i have done. that is assigned sd state 8 and removed the uml initial state and uml final state. Why do the buttonss need to coexist?

a21nibah:
Oh I see. Misunderstood the comments in the code.

c21adano:
The uml Initial state and uml final states are the start and end point of a state diagram they look distinct and need for a state diagram. The SD entity type was linked with a button that is suppose to later be linked with the state diagram entity when that is implemented, the code was commented out due to not being linked to anything at that moment and uml initial state and uml final states were left alone due to working.

Its true that initial and final states need improvment but that is a later issue, as of now the code need to be refactored so UML_STATE entity type dose not conflict with the SD entity type that we much rather use.
[...]

PR #13338, opened 2023-04-17
G1 2023 v2 #13215

a21nibah:
made the needed code changes from issue 13215. the button should  work now and not have any conflicts when you press the state diagram button.


b21eriho (changes requested):
On line 786 of diagram.js the commented out code says SDState: 8, however the new cod reads State: 8. Further down in the code (line 804) it is reffered to as SDState again though, this has to be made uniform.

Further on in diagram.php on lines 401 & 409 (buttons for the uml start and end states) the parameters for "elementPlacement()" & "setElementPLacementType" should be changed from 8 and 9 respectively to be 9 and 10 in accordance with the new numerical value of those states

a21nibah:
the required changes have now been implemented.

b21eriho (changes requested):
The parameter to the setElementPlacementType function on lines 395 & 403 still need to be updated, otherwise it looks good

Issue #13289, opened 2023-04-14
Delay slider stuck while playing


c21adano:
In replay mode while playing a replay the delay slider that dictates how fast the replay should play cant be moved.
If the delay slider is set to 1 second it will be stuck on 1 second unless you pause the replay and changes.

![delaystuckwhileplaying](https://user-images.githubusercontent.com/102599332/231993065-c1557441-0350-40d5-9ef9-0a9b72f47356.PNG)

PR #13340, opened 2023-04-17
Delay slider stuck while playing #13289

a19clasv:
In replay mode while playing a replay the delay slider that dictates how fast the replay should play can now be moved during replay and change the speed accordingly.


b21matfi (approved):
Looks good, replay is not stuck can change while playing and before, works fine

Issue #13315, opened 2023-04-17
Arrow doesnt seem to fully connect


b21oscpe:
Some arrows seem to not fully connect and leave some space between the arrow and the entity. The issue was found #13307 here

![232051437-8d7a1466-4dd1-46fa-a90c-4fedc24f720e (1)](https://user-images.githubusercontent.com/102578165/232430344-9ff28335-07a3-43b3-8175-c82f8c93f481.png)

PR #13341, opened 2023-04-17
Arrow doesnt seem to fully connect #13315

b21willu:


b21matfi (approved):
Looks good, it is connected now to the element.

PR #13347, opened 2023-04-18
G3 2023 v2 issue#13262

g21emmka:
To reach the numbers representing the latest commit, we had to reach the link that it was connected to. This wasn't easy, since the element neither had classes or an id that made it easy to identify and reach. Becasue of this the easiset solution (that we could come up with) to reach the link was to use DomDocument. This is an extension and needed to be installed on the server for it to work properly. 

If the structure of the html on github had been better, we wouldn't have had to use DOM, but we found it necessary in order to resolve the issue.

When testing you should see several numbers, which respresent the lastest commit thas has been made. 

This have been implemented in the file recursivetesting -> getLatestCommit.php

( Co-creator - @c21rebja )

### DIFF
--- a/recursivetesting/getLatestCommit.php
+++ b/recursivetesting/getLatestCommit.php
@@ -5,27 +5,37 @@
 	<meta http-equiv="X-UA-Compatible" content="IE=edge">
 	<meta name="viewport" content="width=device-width, initial-scale=1.0">
 	<title>Document</title>
-
-	<script>
-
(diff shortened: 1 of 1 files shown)


d21emijo (approved):
# Working as intended

image
By adding that text to the github.com/hgustavs/lenasys/commit/ + ### text 
will get you to the latest commit on the master branch

## only takes latest commit to master as seen in picture below in account

image

Issue #13057, opened 2023-04-05
State diagram: states shape


b21oscpe:
The box symbolizing a state should be implemented. Preferably by reusing code from similar modelling techniques. 

The shape should be placeable on the drawing-area. 

EDIT (2023-04-14): 
The button for state entities is currently disabled so we want you to draw a state entity on the board and make it have a similar code-format to the other entites that already exists. This means that it does not have to be placeable just yet. We want the state entity that you make to be linkable to a button in a way that the other entities work.

c21adano:
IE entities have a very similar appearance as to what state entities should have, but they should still be distinct from each other, this can be done by giving state entities rounded conners.

Like other entities state entities should only be allowed to be connected with other state entities.


![statebox](https://user-images.githubusercontent.com/102599332/230048960-359c1ce6-f13b-4be3-ba87-fa59c7bfdd42.PNG)

b21oscpe:
This comes after adding the button.

b21eriho:
I would like to double-check, should I in this issue handle the error-checking for drawing a line between entities of another type and the one I am implementing here. Mainly asking since that seems like a reasonable place to split this issue and I'm not certain I will have the time to implement that type of error-checking before the merge

c21adano:
The error checking can become its own issue later on.

PR #13349, opened 2023-04-18
G1 2023 v2 #13057 state diagram state

b21eriho:
Added a state that gets preloaded, it cannot be edited yet though if copy-and-pasted lines can be drawn between states. The rounded corners are not quite right but it's the best I can come up with for now, they should maybe be done using an svg-path instead of a rect.


b21eriho:
On lines 785 and 801 this issue will conflict with issue #13215 , the changes from that issue are better and will not break anything in this issue and as such they should be used in the merge

b21matfi (commented):
looking at the issue it looks like it should be a button for the state and then the "drawing"part but i don't see the button is it something i missed?

b21eriho:
https://github.com/HGustavs/LenaSYS/issues/13057

I can't say I entirely understand what you mean, the issue is to create the shape

b21matfi:
i was just unsure if the issue was to create the shape or the button. Got a little confused when reading the issue.
i will review the issue

b21eriho:
The button was temporarily removed during a merge-conflict but is being re-added in another issue that is currently in the works

b21matfi:
alright perfect, thank you and it looks good

b21matfi (approved):
The shape looks good.

c21adano:
Related: #13338 
 I fix the small conflict, nothing major. It was with the elementtype number that were conflicting, i went with the numbers used by #13338 due to it changing the numbers in the diagaram.php file too.

PR #13350, opened 2023-04-18
FetchGitHubRepo now uses the URL dynamically sent in when creating/editing course

a21henhe:
FetchGitHubRepo does no longer use the static repo-url and instead uses the link the user puts in when creating/editing a course. To test this, create a new course, have inspector tools open, go in to Network, click the file "fetchGithubRepo.php" and then click preview. This might take some seconds to load but the result should be as the picture below (but you should have some colored sections of the files in the githubrepo, my limit to fetch from github without token ran out for the hour): 

image

Co-authored by b21leowa


e21krida (approved):
Prövat med flera olika repos, fungerar jättebra 👍🏻

PR #13351, opened 2023-04-18
Functional Requirements for duggaedservice

b21tilbe:
Created an updated version of the functional Requirement pdf file according to the previous review.

I was not sure if a forth FR for output was necessary. If you want me to update the pdf. I can add it afterwards if you want to.


a19hakja:
When checking out this branch, I can see that there are two .pdf  files (one in the backend-models folder and one in the functional requirements folder) which isn't supposed to be the case.

b21tilbe:
The one in the backend-models folder is the correct one. I'll delete the other one

b21tilbe:
I can't seem to find the pdf-file in the backend-model map you were talking about. Only the one in Functional Requirement.

a19hakja:
image

This pull request is from branch [g2-2023-v2#13238](https://github.com/HGustavs/LenaSYS/tree/g2-2023-v2%2313238), where this file lies @ LenaSYS/backend-models/

b21tilbe:
That's weird. It is not visible to me at all

a19hakja:
Somehow it's gone now, assigning a reviewer

a21jeaha (approved):
no more complaints from my side, good work :)

Issue #13345, opened 2023-04-18
If you select an element before placing another you may place it under the first


b21oscpe:
![40914eb44575db529239e07b93e56301](https://user-images.githubusercontent.com/102578165/232721436-22820a1f-06ac-4566-9b40-d0cfe6700060.png)

As shown on the picture the unselected entity has been placed under the selected entity. We don'nt know what causes this.

Investigate and attempt to solve.

PR #13354, opened 2023-04-18
G1 2023 V2 #13345

d21matbo:
Solved the problem of overlapping the selected element by removing a continue in the loop for exactly that element. It is no longer possible to place a element under the selected one.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1081,6 +1081,8 @@ var data = []; // List of all elements in diagram
 var lines = []; // List of all lines in diagram
 var errorData = []; // List of all elements with an error in diagram
 var UMLHeight = []; // List with UML Entities' real height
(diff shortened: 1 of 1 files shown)


a21aldsa (changes requested):
This issue is only solved when trying to place an element behind a UML element, For all other elements it is still very much possible to overlap by first selecting an element and then clicking just outside the element. This means you can freely overlap any element behind any element except UML, this issue is therefore not solved.

image

a21aldsa:
Here's an example of me placing an IE element behind a ER, by selecting it first.

image

d21matbo:
These changes should also stop IE Entites from overlapping when no or another element is selected

a21aldsa (approved):
Looks good to me; it seems impossible to spawn elements behind other elements, no matter the type or where you click. Good job.

PR #13356, opened 2023-04-18
G3 2023 v2 issue#13337

a21lukpo:
We changed the local path that the files are saved in to be dynamic instead of static.

Relevant files recursivetesting/FetchGithubRepo.php


b21sebgr (changes requested):
To avoid crashes in the future the splitting and handling of the URL it should be handled inside of the function instead of depend on a global variable that could be incorrect

e21krida:
@b21sebgr This has been updated, it's now done as a parameter instead. Notice that we had to make a new branch to stop all the merge-conflicts that would appear otherwise

Issue #13357, opened 2023-04-18
Research togglePlacementType in diagram.js


c21adano:
togglePlacementType is causing problems with the state entity button.

![toggleplacenttype](https://user-images.githubusercontent.com/102599332/232763238-c6e5b03b-7160-4ec6-be16-42a5ae79abb8.PNG)

PR #13360, opened 2023-04-18
g1-2023-v2-#13357 Add SD to togglePlacementType()

b21eriho:
Before this issue clicking the button for placing SD states would break the system, after this issue the system will not break in this case, however it still breaks if you attempt to place an SD state. Meaning you can select said button and then select another without it breaking as long as you do not click in the diagram area


b21eddka:
The SD state button works as described, it only breaks the system when trying to place it in the diagram area. Meaning that the button itself does not work, however does not break the system in other circumstances. If this is deemed as a successful solution i approve of the merge.

PR #13362, opened 2023-04-18
G2 2023 v2 issue#13331

c21axepe:
The serviceLogEvent outputted unix time instead of datetime.
I solved this by converting unix to datetime in the log.php file.
Before:
image
After:
image


b21hamwe (approved):
Well done, gold star.

Issue #13283, opened 2023-04-14
Mock-up for state entity options-panel


b21oscpe:
When you click on for example an ER-entity and then navigate to the options-panel, this is what you see:
![8d263de2842bad9ecbe7ea5b239184b4 (1)](https://user-images.githubusercontent.com/102578165/231986295-05d371e7-1474-4ffa-b18e-d6f5e73e905b.png)

We want something similar to be shown when you click on a state entity

b21oscpe:
Currently this issue relates to the STATE ENTITY, we want to be able to change the text of the entity, note that the "do:" text should remain unchanged.

PR #13375, opened 2023-04-18
G1 2023 v2 #13283

b21matfi:
Fixed so the SD State button works(typo), and fixed so that you can change name to the state name and add attributes in the optionpane.


b21eriho (approved):
All that is set out to fix works

Issue #13333, opened 2023-04-17
Inconsistencies with selected lines highlight in both light and darkmode


d21matbo:
When selecting a line to change its properties the colour changes to visualise that it is selected.
However there appears to be some inconsistencies with this effect.

### For lightmode everything except the number for the cardinality and Icons:Triangle is changed
![image](https://user-images.githubusercontent.com/102599161/232478701-82dec100-029a-419f-afd1-cd0c60c14e18.png)
![image](https://user-images.githubusercontent.com/102599161/232480087-db8d6aae-5e47-40c5-9c8d-23c9c60bc737.png)

### For darkmode only the label is changed
![image](https://user-images.githubusercontent.com/102599161/232480482-bf36b99e-0797-4e02-aea8-74a2de53289e.png)

b21oscpe:
Everything selected should be highlighted purple

PR #13376, opened 2023-04-18
 Inconsistencies with selected lines highlight in both light and darkmode #13333 

a21aldsa:
Removed the CSS code for setting the line color in darkmode. It overwrote the selection color. This fixes the issue but this made the lines black in darkmode though, so to fix that I: 

Added a function called isDarkTheme for returning true/false if the current theme is dark mode or not and then used this function for setting the current linecolor in DrawLine. Then called upon showdata() on the bottom of toggleDarkMode() which redraws everything and therefore makes the lines automatically the correct color when switching themes.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -7446,7 +7446,12 @@ function drawLine(line, targetGhost = false)
     else{
         var strokeDash="0";
     }
-    var lineColor = '#000000';
+    if (isDarkTheme) {
+        var lineColor = '#FFFFFF';
+    } else {
+        var lineColor = '#000000';
+    }
+    
 
     if(contextLine.includes(line)){
         lineColor = selectedColor;
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -11104,7 +11104,7 @@ function drawSelectionBox(str)
         selectionBoxHighY = highY + 5;
 
         // Selection container of selected elements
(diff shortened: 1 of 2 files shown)


b21eriho (changes requested):
These changes have indeed added more elements being shown as selected, however the cardinalities are still shown as nonselected.
image

Also this branch has faults in darkmode in general
image

in comparison to the current stable branch:
image

However this last problem may not be related to this issue, though I still felt it was relevant

b21eriho:
Right, presuming this is true I'd approve this issue

b21oscpe:
"Looks good👍"

c21adano:
I resolved the two conflicts, showdata() and toggleBorderOfElements() were overlapping and The new if (isDarkmode) was overlayed ontop of ineColor, i was unsure if ineColor was a typo of if it was suppose to be its own function call so i only commented it out for now.

a21aldsa:
> I resolved the two conflicts, showdata() and toggleBorderOfElements() were overlapping and The new if (isDarkmode) was overlayed ontop of ineColor, i was unsure if ineColor was a typo of if it was suppose to be its own function call so i only commented it out for now.

I reckon that should say lineColor, are you sure you didnt accidentally remove something?

b21oscpe:
8c7020d7f45337a16eb1dc8340860767

Merge should have added the correct code and look as described by developer

Issue #13223, opened 2023-04-12
GitHub Button clickable functions, show screen with information/settings


c21sebar:
For Issue https://github.com/HGustavs/LenaSYS/issues/13431
From #13220 
Add so that the GitHub button is clickable and show a screen

For the Moment-Github button
Github-Folder link:                "link"-text input OR dropdown from the CourseRepo
Filepath:                                 "text that show filepath on webserver"
Order of items                       "file that controls the order of the examples"

For the course-example Github button
Template                          template X
number of files                 number of files

For each file in the code-example
Name:                                "Name.type"
GithubUrl                            "GitHubDownloadUrl"

d21emijo:
added a screen to github moments

![image](https://user-images.githubusercontent.com/102797583/232798608-5e2dcedf-a109-4194-be1c-0946e3686d57.png)

d21emijo:
for 
"
For the course-example Github button
Template template X
"
something similar to 
![image](https://user-images.githubusercontent.com/102797583/233069522-714e3db8-a9b3-4cf6-912c-0b110b98616d.png)
from duggasys

PR #13378, opened 2023-04-18
G3-2023-v2-issue#13223-GitHub-screen

d21emijo:
from issue #13223 added a new "screen" for github interactions


c21sebar (approved):
This works, nothing breaks. A screen with 4 input fields appears only on the moment type in a course, this is as intended. Both for moments that already exist and new moments added. For future issues a submit button should be added with functionallity to save the inputs should be added.

PR #13383, opened 2023-04-19
G3 2023 v2 issue#13337 v2

e21krida:
We changed the local path that the files are saved in to be dynamic instead of static.

Relevant files recursivetesting/FetchGithubRepo.php

Worked in conjunction with @a21lukpo 
(Had to create new branch since everybody was working on the same file and a whole lot of merge-conflicts appeared otherwise)


b21sebgr (approved):
at least it is working

Issue #13387, opened 2023-04-19
Initial and final states break the system


b21oscpe:
@c21adano said it had something to do with the refactoring of UML_STATES that happened before the merge
(The problems arise when you try to interact with an initial or final state)

![54d43954f6d399b2af0161b4ebfe538b](https://user-images.githubusercontent.com/102578165/233018598-5a204c27-6286-4e91-a032-bdaf3027fa6d.png)

b21oscpe:
Using the initial states and connecting them to an entitity also breaks the proportions in a way that makes it so that state entities dont scale with the zoom variable.

b21oscpe:
This issue has been extended to refactor some parts that required refactoring in order to make the neccessary changes, testers beware O_o

PR #13403, opened 2023-04-19
 Initial and final states break the system #13387 

a21aldsa:
While trying to fix this issue, which I thought had something to do with generateContextProperties, I tried going through the code in generateContextProperties and I couldn't read it properly, it was had such horrible indentation that it was next to impossible to know which bracket belonged to which block. After consulting with the issue leader I got permission to rewrite the part pertaining to SDState; I made it more correct indentation-wise and also added a check for if the element is of the type SD before it checks any "kind". This was done more so for ease of expansion, so more kinds can be added more easily, but somehow this fixed the issue entirely. I am not sure why, but my best guess is that the horrible indentation made a bracket not be included or perhaps the whole block was included in the block above it (IE) which then lead to the element that was having its context generated got its kind overwritten from, for example, UMLInitialState to SDState.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -8913,6 +8913,7 @@ function drawElement(element, ghosted = false)
 
     // Check if element is SDState
     else if (element.kind == "SDState") {
+        elemAttri = element.attributes.length;
         //div to encapuslate SD element
         str += `<div id='${element.id}'	class='element' onmousedown='ddown(event);' onmouseenter='mouseEnter();' onmouseleave='mouseLeave()';' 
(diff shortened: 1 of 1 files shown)


b21eriho (changes requested):
This change has solved the issues presented in issue #13387, however I reckon the old (now commented-out) code should be entirely removed. Otherwise good.

a21aldsa:
> This change has solved the issues presented in issue #13387, however I reckon the old (now commented-out) code should be entirely removed. Otherwise good.

the commented out code has been removed

b21eriho (approved):
Looks good 👍

PR #13408, opened 2023-04-19
G1 2023 v3 #13166 refactor hiding of submenu-buttons

b21eriho:
The code for only showing the currently selected element for diagram entities or relations was previously over 300 lines long and unnecessarily repetitive. After this implementation it is instead a simple algorithm that's about 85 lines long. Most commits are in another branch however, namely branch g1-2023-v2-#13166 , since they were migrated to this version of the stable branch.


b21eriho:
For tester, no behavior should change since this is just a refactor of old code

a21aldsa (approved):
the functionality is the same as it was. Good job.

PR #13415, opened 2023-04-19
Added try-catch to SDState.

b21phiro:


b21oscpe:
link the issue that relates to this PR

DuggaSys/diagram.js
```diff
@@ -8893,56 +8893,66 @@ function drawElement(element, ghosted = false)
...
     // Check if element is SDState
     else if (element.kind == "SDState") {
-        elemAttri = element.attributes.length;
-        //div to encapuslate SD element
-        str += `<div id='${element.id}'	class='element' onmousedown='ddown(event);' onmouseenter='mouseEnter();' onmouseleave='mouseLeave()';' 
+        try {
+
+            elemAttri = (element.attributes.length ?? null);
+
+            if (!elemAttri || elemAttri == null) throw Error(`Element attribute for ${element.name} is missing or is null!`);
```
a21nibah, line 8900: please change this line so that the trhow error is below the if statement

a21nibah (changes requested):
someone else will need to test this because  the way it is now the code seems to reacts to mouse events which is something i can not use.

d21matbo (changes requested):
Something goes very wrong here.

In the image below we can see that the state entity goes invisible when the attributes field is left empty. In addition every pixel the mouse moves this error message is displayed and multiple other errors are thrown in the console

image

b21oscpe:
What's the status on this PR?

b21phiro:
Cancelled due to error check/alert allready in place.

Issue #13414, opened 2023-04-19
Create column/table in SQLLite for commit/link


c21sebar:
Create column/table in SQLLite for commit/link for issue https://github.com/HGustavs/LenaSYS/issues/13413

PR #13430, opened 2023-04-19
G3 2023 V3 13414 Create column/table in SQLLite for commit/link

c21rebja:
Issue #13414 

Added another table in the SQL lite database, containing the columns githubURL and commitID.

image

This can be seen in basic.php and in CMD:
- Enter students/"studentid"/githubMetadata
- Write sqlite3 metadata2.db
- Write .schema

The tables should show up.

(co-creator @g21emmka)

### DIFF
--- a/Shared/basic.php
+++ b/Shared/basic.php
@@ -256,7 +256,7 @@ $sql2 = '
 		repoPath VARCHAR(255) 
 	);
 
-	CREATE TABLE latestCommit (
+	CREATE TABLE IF NOT EXISTS latestCommit (
 		githubURL VARCHAR(255) PRIMARY KEY,
 		commitID VARCHAR(50)
 	);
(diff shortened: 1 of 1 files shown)


a21rammo:
Please attach issue in the development option bar. So it automatically closes issue when PR is closed.

a21henhe (changes requested):
Seems to be working fine on the webserver. Although when testing locally, it results in error messages on each refresh, since it tries to recreate the table each time: 
image

I would suggest creating the table using: "CREATE TABLE IF NOT EXISTS...." just like the previous tables created in the file are using.
Otherwise it looks good :) @c21rebja

c21rebja:
@a21henhe If not exists added! :)

a21henhe (approved):
Looks good :)

Issue #13390, opened 2023-04-19
Changing type on IE or Class entities allows them to connect with ohter types


c21adano:
If you connect two IE or Class entities then change the type of one of them, the connection will stay and allow for two diffrent kinds of diagrams to be connected.

There is and error message that plays when you try to change ER while connected, reuse that error message here.

![changingtypealowstocennect](https://user-images.githubusercontent.com/102599332/233031935-8ccc644a-8eca-4b13-afa0-10f632a704c7.PNG)

PR #13432, opened 2023-04-19
G1 2023 v3 #13390

b21matfi:
Fixed issue where you can change type when connected, if its UML and ER and displays a message.


c21theti (approved):
Looks good. Error mesage is added when you try to change type when connected if the type is wrong

PR #13433, opened 2023-04-19
Diagrams have now changed positions so they don't load in at different positions

c21theti:


d21matbo (approved):
Looks good! The diagrams now shares the same positions.

Issue #13353, opened 2023-04-18
Problems when moving multiple entities (with box selection) when some entities are overlapping


d21matbo:
### Moving a cluster of entities has problems while one or more of them overlap at the new position.

The following sequence of images shows how one attribute "2" has its position reset due to overlapping with entity "B", while the rest of the cluster does not.

![image](https://user-images.githubusercontent.com/102599161/232748431-dbec9b4b-3299-453d-ac1b-18100e9a8dec.png)
![image](https://user-images.githubusercontent.com/102599161/232749002-6ee532d4-dbce-4dd9-87a4-455ec71ac4dd.png)
![image](https://user-images.githubusercontent.com/102599161/232753649-9b69bac8-345c-43dd-a786-ac3be7980691.png)

This does not look like intended behaviour as it is possible to really mess up the diagram when only parts of it gets reset.

b21oscpe:
dont forget to link your issue in your pull-request

PR #13435, opened 2023-04-19
Problems when moving multiple entities (with box selection) when some entities are overlapping #13353

a19clasv:
All of the moving entities get reset when overlapping and they don't get reset because they overlap the original position of a moving entity


b21oscpe:
link your issue please

b21eriho (approved):
Looks good 👍

Issue #13366, opened 2023-04-18
State entity outline should be fixed for lightmode


b21oscpe:
![7e0a0e3bb63fac1d1d4155b97f33dbbf](https://user-images.githubusercontent.com/102578165/232776466-524000ff-b5bf-46df-9efe-17f57cd5f825.png)

The corners of the state entity doesn't have enough contrast to the background, this should be fixed.

PR #13436, opened 2023-04-20
 State entity outline should be fixed for lightmode #13366 

a21aldsa:
It was not just lightmode, the corers were rounded in a way that cut off the borders no matter the theme. I fixed this by first removing the CSS border radius since that would never have worked and then i replaced the svg <rect> with a path that had arcs on the appropriate corners. I chose this over rx on the rects because these are two rects on top of eachother which would mean 8 corners are rounded when it really should only look like 4. I guess I could've used a clip for rx but path seems better and more proper.


b21eriho (approved):
Looks good 👍

Issue #13438, opened 2023-04-20
Add a button for Test-case generation


b21oscpe:
This issue needs to be handled AFTER #13423 

We want something similar to this to be added to the toolbelt:
![298c61a15cf7f2ad681d76110ea249d3](https://user-images.githubusercontent.com/102578165/233305787-badd016c-3ee5-4832-b10a-8ae0aef0d7af.png)

As a part of this issue you need to design the button, come to me or @c21adano with your suggestions

PR #13444, opened 2023-04-20
G1 2023 v3 #13438 test-case generation button

b21eriho:
The button has been added, not much more to it


a21aldsa (approved):
There is a button, good job.

PR #13466, opened 2023-04-20
Test-file for fileedservice

b21tilbe:
Created a step by step guide to test out the code from fileedservice.php.

### DIFF
--- a/backend-models/service-tests/fileedservice-tests.txt
+++ b/backend-models/service-tests/fileedservice-tests.txt
@@ -1,35 +1,50 @@
 Fileedservice.php
 
 All tests
-----------------------
+----------
 
-- Test to check if the variable "$userid" has been set by session or not:
-	1. Set up the database and necessary databases.
-	2. Start by calling the function "session_start()".
-	3. Set "$_SESSION['uid']" to "null".
-	4. Call "fileedservice.php" with correct values for all parameters.
-	5. Check what the value is for "$userid". It should be 1.
-	6. Check what the value is for "$debug". It should be NONE.
+- Test 1: Test the user's login:
+	1. Start up the session.
+	2. Set the user ID in the session.
+	3. Call the function "checklogin()".
+	4. For the test to pass, the returning value should be true.
 
-- Test if "$kind" is'nt 2 and that the user has access to the course.
-	1. Arrange the database and it's appropriate tables.
-	2. Call the function "session_start()".
-	3. Set "$_SESSION['uid']" to a value that has access to the course. Be sure to also set "$_GET['kind']" to a value that's not equivalent to 2.
-	4. Call "fileedservice.php" with correct values for all parameters.
-	5. Check what the value is for "$hasAccess". It should be true.
-	6. Check what the value is for "$debug". It should be NONE.
+- Test 2: Test to get a username from the database:
+	1. Start up the session.
+	2. Set the user ID in the session.
+	3. Call the function "checklogin()".
+	4. Call the function "getUsername()".
+	5. For the test to pass, the returning value of the function "getUsername()" should be the user's username.
 
-- Test if "$kind" is 2 and if the user is a superuser.
-	1. Arrange the database and it's appropriate tables.
-	2. Start by calling the function "session_start()".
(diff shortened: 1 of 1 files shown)


a19hakja:
Link to issue pls

b21kurar:
This does not contain any values, as such no tests can be written from this. No pre-requisites to reach code to be tested, no specification of what code is tested.

Please take a look at this file, this follows the template in a correct manner: https://github.com/HGustavs/LenaSYS/blob/e3146f5c3192f2495a03c7d6f3deeeaf274535f5/backend-models/service-tests/duggaedservice-tests.txt

backend-models/service-tests/fileedservice-tests.txt
```diff
@@ -121,14 +121,41 @@ if ($result && $_SESSION['uid'] == 1) {
 Test the user's access to a course (line 39)
 --------------------------------------------------------------------------------------------------------------------------------------------
 
+*********
+* pre-req:
```
a21jeaha, line 125: pre-req is what is needed before the service can execute.

in this case the pre req for "DELFILE" (which is the first service used) is that 
if (checklogin() && $hasAccess) = IS TRUE

what values these need to have is what is part of the pre-req 
image
you need to specify what values they must have for the specific service test.
everything that is

backend-models/service-tests/fileedservice-tests.txt
```diff
@@ -121,14 +121,41 @@ if ($result && $_SESSION['uid'] == 1) {
...
-	2. Set the user ID in the session.
-	3. Call the function "hasAccess()". Try different access levels.
-	4. Examine the returning values.
-	5. For the test to pass, the returning value should be false if the user does not have access to the course and true if the user has.
+
+*********
+* Start up the session:
+*********
+
+session_start();
```
a21jeaha, line 141: just copying in session_start();  is not helpful

image

what value does the user id need to have to be able to execute the service ?
this depending on what privileges are needed (probably something to do with $hasAccess).

if anything this is part of the pre-req

backend-models/service-tests/fileedservice-tests.txt
```diff
@@ -121,14 +121,41 @@ if ($result && $_SESSION['uid'] == 1) {
...
+
+
+*********
+* Test access:
```
a21jeaha, line 145: as i stated before this is not a service that was suppose to be tested. as such i'm having a hard time giving useful information directly associated to this commit.

Here you have a complete example of a service test that might be useful. this follows the new standard that is expected.

duggaedservice.php

---------------------------------------------------------------------------------------------------------------------
- Create an assignment (line 57):
---------------------------------------------------------------------------------------------------------------------

*********
* pre-req:
********* 

    { (checklogin) IS TRUE && 
    (hasAccess($userid, $cid, 'w') || isSuperUser($userid) || hasAccess($userid, $cid, 'st')) } IS TRUE


*******
* send parameters inputs to duggaedservice.php 
******
send
{
    $opt = "SAVDUGGA";
    $qid = null; or $qid = “UNK”;
    $cid = 1885;
    $userid = 2;        // this is automatically added depending on what session is active (if any), we want the value to be 2
    $courserver = 1337;

} 

*****
* Get newly creaded Assignment ID
*****
    
    SQL-query: SELECT MAX(id) FROM quiz;     // save to variable $id this value will identify the last assignment that was created (it has the highest ID number) 

*****
* Gather service output
*****

    Save the values of the echoed array as a JSON (at the end of the file. echo json_encode($array);)
    this is the expected output for the micro service.  

*****
* remove inserted row from DB, using qid 
*****

    SQL-query: DELETE FROM quiz WHERE id = $id;

a21jeaha:
@b21tilbe your not suppose to close the issue just because it wasn't finished before the merge.. what you should do is delete the old branch and create a new one. 

click on "branches" 
image
click on "yours"
image
then just click on the trashcan on the old branch. 

this will remove the branch from the issue as well, from there just create a new branch by clicking here 
image
give it a proper name and make sure that you change the branch source to the newest group branch.

an issue is only to be closed when it is completed or if the group leaders decide that the issue no longer is required!

Issue #13421, opened 2023-04-19
Roadmap: Refreshknapp i code-example


a21rammo:
* Refreshbutton for code-example - #13162 
* Fetch the relevant data from SQlite and MySQL - #13443 see issue https://github.com/HGustavs/LenaSYS/issues/13348 and https://github.com/HGustavs/LenaSYS/issues/13439
* Fetch files that are in the code-example - #13001 
* Structure the SQLite database https://github.com/HGustavs/LenaSYS/issues/13478

fetchGitHub script should solve these issues 
* https://github.com/HGustavs/LenaSYS/issues/13348
* https://github.com/HGustavs/LenaSYS/issues/13439
* Update database with the new relevant files that are fetched - #13451

PR #13468, opened 2023-04-20
SQLite modification

b21sebgr:
For issue https://github.com/HGustavs/LenaSYS/issues/13478 so we can continue in https://github.com/HGustavs/LenaSYS/issues/13421
Modified the sqlite database to better handle what we currently want to use it for

### DIFF
--- a/DuggaSys/sectionedservice.php
+++ b/DuggaSys/sectionedservice.php
@@ -9,6 +9,7 @@ date_default_timezone_set("Europe/Stockholm");
 // Include basic application services!
 include_once "../Shared/sessions.php";
 include_once "../Shared/basic.php";
+include_once "../recursivetesting/FetchGithubRepo.php";
 
 // Connect to database and start session
 pdoConnect();
--- a/DuggaSys/sectionedservice.php
+++ b/DuggaSys/sectionedservice.php
@@ -471,23 +471,28 @@ if($gradesys=="UNK") $gradesys=0;
 							$this->open("../../githubMetadata/metadata2.db");
 						}
 					}
-					$query = $pdo->prepare("SELECT filename FROM box WHERE exampleid=:exampleid;");
-					$query->bindParam(":exampleid", $exampleid);
-					$query->execute();
+					$query1 = $pdo->prepare("SELECT filename FROM box WHERE exampleid=:exampleid;");
+					$query1->bindParam(":exampleid", $exampleid);
+					$query1->execute();
+					$query2 = $pdo->prepare("SELECT cid FROM codeexample WHERE exampleid=:exampleid;");
+					$query2->bindParam(":exampleid", $exampleid);
+					$query2->execute();
 					$files = array();
-					foreach($query->fetchAll() as $row) {
-						array_push($files, $row['filename']);
+					$row2 = $query2->fetchAll();
+					$cid = $row2['cid'];
+					foreach($query1->fetchAll() as $row1) {
+						array_push($files, $row1['filename']);
 					}
 					$gdb = new githubDB();
 					$downloads = array();
 					foreach($files as $file) {
-						$que = $gdb->query("SELECT downloadURL FROM gitRepos WHERE fileName=".$file.";");
+						$que = $gdb->query("SELECT downloadURL FROM gitFiles WHERE cid=".$cid." AND fileName=".$file.";");
 						while($row = $que->fetchArray(SQLITE3_ASSOC) ) {
 							array_push($downloads, $row['downloadURL']);
 						}
 					}
 					$gdb->close();
-					//13179 här anropas uppdateringsfunktionen
+					//TODO rest från 13179, här anropas uppdateringsfunktionen
 				}
 			}
 		}
--- a/DuggaSys/sectionedservice.php
+++ b/DuggaSys/sectionedservice.php
@@ -63,7 +63,7 @@ $feedbackenabled =getOP('feedback');
 $feedbackquestion =getOP('feedbackquestion');
 $motd=getOP('motd');
 $tabs=getOP('tabs');
-$exampelid=getOP('exampelid');
+$exampelid=getOP('exampleid');
 
 $visbile = 0;
 $avgfeedbackscore = 0;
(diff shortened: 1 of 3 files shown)


a21rammo:
You seem to have a merge conflict with @e21krida. Talk with each other and figure out which of your codes should remain and be merged to the weekly branch. 

image

e21krida:
@b21sebgr Why the modification of repoID INTEGER PRIMARY KEY AUTOINCREMENT -> cid INTEGER, or am i reading it wrong? Also, @a21hugni developed most of that table, so pinging him in for comment.

a21rammo:
The picture before was current code. Here is screenshot of what sebastians code is (the incoming code): 
image

b21sebgr:
> You seem to have a merge conflict with @e21krida. Talk with each other and figure out which of your codes should remain and be merged to the weekly branch.
> 
> image

my job was to change this exact code that now have a merge conflict so my suggestion is that my branch overrides @e21krida:s changes

e21krida:
That doesn't necessarily answer my question, why the removal of the autoincremental feature? Why rename it? And why isn't the other columns necessary data?

a21hugni:
> > You seem to have a merge conflict with @e21krida. Talk with each other and figure out which of your codes should remain and be merged to the weekly branch.
> > image
> 
> my job was to change this exact code that now have a merge conflict so my suggestion is that my branch overrides @e21krida:s changes

If your changes overrides the existing database and that same database is being used somewhere then everything goes to the pipe turn.

b21sebgr:
> > > You seem to have a merge conflict with @e21krida. Talk with each other and figure out which of your codes should remain and be merged to the weekly branch.
> > > image
> > 
> > 
> > my job was to change this exact code that now have a merge conflict so my suggestion is that my branch overrides @e21krida:s changes
> 
> If your changes overrides the existing database and that same database is being used somewhere then everything goes to the pipe turn.

that is totally correct statement, thats why this is how it should have been designed in the first place but since it wasnt and that made it impossible to work with when others started using it i now have made a change to make it usable, all code that already is using this will have to change but that isnt very much since it still early in develompent for this database, the question should be why this change is done since groupleader already requested it, the question should rather be about if it is done in a correct way

e21krida:
>if you had read the code you would probably relize

There is no need to be passive aggressive, the merge-conflict presented looks like the following.
```
$sql2 = '
	CREATE TABLE IF NOT EXISTS gitRepos ( 
<<<<<<< g3-2023-v3-issue#13421
	cid INTEGER,
	repoName VARCHAR(50), 
        repoURL VARCHAR(255), 
	lastCommit VARCHAR(50),
	PRIMARY KEY (cid)
=======
	repoID INTEGER PRIMARY KEY AUTOINCREMENT,
	repoName VARCHAR(50), 
	repoURL VARCHAR(255), 
	repoFileType VARCHAR(50), 
	repoDownloadURL VARCHAR(255), 
	repoSHA VARCHAR(255) UNIQUE, 
	repoPath VARCHAR(255) 
	);
``````
This is the presented code changes within the conflict, i can very clearly see in the commits that you've created a new table, i was simply trying to delve into your reasoning before you seemed to get upset. If you wanted to highlight a change and a choice to split the table into two different tables, then inform us of that without expecting us to read into the entirety of what you've done. I can't comment on the efficiency of the previous "repoID" as i wasn't the creator of it, but throwing out statements regarding it's validity in the database with a negative stance without asking first seems quite contra-productive to the goal of upholding solid communication within the group. There might very well have been a reason as to why the "repoID" existed with an auto incrementing value for future work. 

Just give us some information as to why you've done what you've done instead of assuming that we're trying to devalue your work or downright attack it, that's not the intention here at all. I asked questions because i'm curious, simple as.

**Comment by e21krida** [comment]
> Once again im sorry if i seemed attacked as it was not my feeling and i will try and keep this in mind for future writing.

No need to apologise! Just wanted to make sure there was no hard feelings. You might very well have improved upon it, remember i was only included in the issue that created this table to bugtest after the creation as we couldn't get it to work after @a21hugni implemented the actual code at hand. All i can ask for is a comment by @c21sebar as i want to remember a conversation between him and @a21hugni as to why the auto-incrementing value was there to begin with. This will also induce an error in FetchGithubRepo.php given that it's trying to feed into columns that now doesn't exist. So we need to get @a21rammo to open an issue to fix this real quick after this has been merged. 

Also, did you test if for example "PRIMARY KEY (cid)" works in sqlite? I want to remember overhearing from @a21hugni that it doesn't work, but that you have to implement it as a command on the creation of a specific column. But i could be mistaken as well. I'm going to assume that you tested it but just making sure!

**Comment by c21sebar** [comment]
The change to the SQLite db was needed as we continue to develop and put things together and we noticed that the first structure didnt work to communicate with the rest of the system. We work incrementally and implement and fix things as we go, the final, at the end of the project, implementation/version of the code we merge will not always be what we initially wrote. Yes now we need to go backwards and fix the querys to the SQLite db. This can be done in a later stage in issue https://github.com/HGustavs/LenaSYS/issues/13447 where we cleanup that the php, i can add it there as another point to fix, we do these fixes after we have solved the other branches in that file to avoid more merge conflicts.

**Comment by b21sebgr** [comment]
I have now looked up primary key in sqlite and both internet and my tests points to it actually working in sqlite too, however if it shows to fail in the future here are a stackoverflow that might give some options to how to solve the problem https://stackoverflow.com/questions/734689/sqlite-primary-key-on-multiple-columns

**Comment by e21krida** [comment]
>I have now looked up primary key in sqlite and both internet and my tests points to it actually working in sqlite too.

Well brilliant, the more you know! If that was all then i'd say we just continue on with your changes @b21sebgr

File: DuggaSys/sectionedservice.php [hunk 1]
```diff
@@ -9,6 +9,7 @@
 // Include basic application services!
 include_once "../Shared/sessions.php";
 include_once "../Shared/basic.php";
+include_once "../recursivetesting/FetchGithubRepo.php";
```
a21olljo, line 12: Is FetchGithubRepo.php used anywhere?

a21rammo:
I talked with group leader (@c21sebar), about this pull request. I rather not to merge an incomplete branch into the wekkly branch. Instead, I want to ensure that it's fully functional so that other team members can continue their work without any disruptions. This is why we've created a separate branch, which will be merged with the current branch. Once we've updated and tested the queries and insertions to work with the new database table, we'll merge this combined branch back into the weekly branch. This is in order to prevents any unnecessary obstacles for the team.

a21rammo:
New branch was created from the weekly branch as source. Name of the new branch is: [g3-2023-v3-fix-queries-for-new-db](https://github.com/HGustavs/LenaSYS/tree/g3-2023-v3-fix-queries-for-new-db)

a21olljo:
Merge conflicts has been resolved and the code looks good. However because it is no complete solution why I cant really do extensive testing, but from what could be done the result looked good. This is also the reason why the branch isent merged into the weekly branch, and instead g3-2023-v3-fix-queries-for-new-db https://github.com/HGustavs/LenaSYS/tree/g3-2023-v3-fix-queries-for-new-db. 

Continuation will be done within the issue #13542

PR #13473, opened 2023-04-20
Overlapping with SD states is no longer possible. Issue#13381

b21eddka:
It is no longer possible to overlap the SD states when placing or moving them.


a21aldsa (approved):
Works, good job.

PR #13474, opened 2023-04-20
added sorting to log.php G2 2023 v3 issue#13329

a21oscgu:
- I have added sorting functionality for log.php.

- I have added default sorting set on time (timestamp) from descending order.

image

- If you click on the button "Set to default(timestamp, descending order)", then order and sort will be set on default (default is time in descending order).

- I have added styling so the tables becomes easier to read.

- I have updated the `<th>` tags. They are now clickable links:

image

- If you click on one of the links (and then select the table in the dropdown again), then the sort variable will switch from ascending to descending or vice versa. Which means that the content is sorted in ascending/descending order.

### DIFF
--- a/Shared/log.php
+++ b/Shared/log.php
@@ -148,7 +148,7 @@
                 $order = "id";
                 $sort = "ASC";
 
-                $userLogEntriesSql = $log_db->query('SELECT * FROM userLogEntries ORDER BY '{$order}''{$sort}';'); // ORDER BY $order $sort
+                $userLogEntriesSql = $log_db->query('SELECT * FROM userLogEntries ORDER BY $order $sort;'); // ORDER BY $order $sort
                 $userLogEntriesResults = $userLogEntriesSql->fetchAll(PDO::FETCH_ASSOC);
                 $sort == 'DESC' ? $sort = 'ASC' : $sort = 'DESC';
                 echo"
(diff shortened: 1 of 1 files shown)


c21alest (changes requested):
When you click on a title to change sorting the page reloads and the chosen table resets. I would like to see a fix for this :)

a21oscgu:
> When you click on a title to change sorting the page reloads and the chosen table resets. I would like to see a fix for this :)

Is it possible to accept the current version and add a new issue for this problem @b21kurar?

b21kurar:
> > When you click on a title to change sorting the page reloads and the chosen table resets. I would like to see a fix for this :)
> 
> Is it possible to accept the current version and add a new issue for this problem @b21kurar?

If it still fulfills the issue at hand (that means that @c21alest approves of the functionality/code standard/other relevant stuff...), then I say yes we can add it as a bugfix issue.

c21alest:
> > > When you click on a title to change sorting the page reloads and the chosen table resets. I would like to see a fix for this :)
> > 
> > 
> > Is it possible to accept the current version and add a new issue for this problem @b21kurar?
> 
> If it still fulfills the issue at hand (that means that @c21alest approves of the functionality/code standard/other relevant stuff...), then I say yes we can add it as a bugfix issue.

I approve!

c21alest (approved):

a19hakja:
Well done, it's a bit difficult to trace your steps, I'd like a bit more intuitive titles on your commits. But otherwise great work.

I'll set up another issue for the bug fix

PR #13476, opened 2023-04-20
G2 2023 v3 issue#13343 - Write tests for services in duggaedservice

a21jeaha:
issue#13343 is finished, needs to be reviewed. 

I have created a text-file which outlines what is to be tested and what values that are used for said tests.
These values are used among other things for SQL-queries and need to be tested to make sure that all queries behave as intended and that LenaSYS still functions with the added or removed data. To give an idea of what to search for, I experienced a bug where and INSERT made it impossible to edit an added "momen", this was a result of faulty values in the INSERT.

could not open edit
image

### DIFF
--- a/backend-models/service-tests/duggaedservice-tests.txt
+++ b/backend-models/service-tests/duggaedservice-tests.txt
@@ -2,8 +2,9 @@
 duggaedservice.php
 
 ---------------------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


b21hamwe (approved):
Well made document, the few questionmarks that's been taken up to discussion has been
cleared. 
Good job!

Issue #13437, opened 2023-04-20
Implement a way to output nodes and edges


b21oscpe:
State diagrams may now be placed on the drawing area and lines can be drawn between them. Similar to ER table generation we want to be able to generate data from our models. To start with, find a way to output the amount of nodes and edges in a state diagram model.

PR #13477, opened 2023-04-20
G1 2023 v3 #13437

a21nibah:
Made it so you can see how nodes and edges are connected. the output shows up in The console when you click the test case button.
There is also an alert that pops up that tells you the amount of lines and nodes.
Send me a dm on discord if you want to use the diagram I tested against.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -4014,7 +4014,7 @@ if(data[i].kind==elementTypesNames.SDState)
     }
 }
     }
-    output+="line counter is "+lineCounter+"and state counter is"+elementCounter;
+    output+="line counter is "+lineCounter+" and state counter is "+elementCounter;
     return output;
 }
 /**
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -3975,7 +3975,7 @@ function generateOutput()
 if(data[i].kind==elementTypesNames.SDState)
 {
     if(data[i].id==lines[j].fromID)
-    console.log("output: "+lines[i].fromID);
+    alert("output: "+lines[i].fromID);
     }
 }
     }
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -3964,22 +3964,45 @@ function boxSelect_Draw(str)
 //#endregion =====================================================================================
 //#region ================================ GUI                  ==================================
 /**
- * @description hides or shows the diagram type dropdown 
+ * @description prints out info about state diagrams
  */
-function generateOutput()
+function printStateDiagram()
 {
+    var lineCounter=0;
+    var elementCounter=0;
+    //had to make a separate loop to count the lines otherwise if it is put in the other loop the counter behaves in a strange way. 
+    for(var i=0; i<lines.length; i++)
+    {
+        if(lines[i].type==entityType.SD)
+    lineCounter++; 
+    }
     for (var i=0; i<data.length; i++)
     {
+        if(data[i].kind==elementTypesNames.SDState)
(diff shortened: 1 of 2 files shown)


a19clasv (changes requested):
The output doesn't show up in the console and the alert doesn't pop up. What does show up is an error message like this if I click on a node.
Error
diagram.js:8919 Uncaught TypeError: Cannot read properties of undefined (reading 'length')
    at drawElement (diagram.js:8919:40)
    at showdata (diagram.js:11782:16)
    at updateSelection (diagram.js:3311:13)
    at ddown (diagram.js:2172:17)
    at HTMLDivElement.onmousedown (diagram.php:1:1)

a21nibah:
did you click on the test cases button?

a19clasv:
It seems to be working on their machine if they use the enter key to do an onclick(). There seems to be something wrong with the button itself. I get this error message.
image
Uncaught ReferenceError: printStateDiagram is not defined
    at HTMLDivElement.onclick (diagram.php:486:93)

It would be best if the one who made the button took a look at it again.

a21nibah:
Clerification. It works if you move the onclick to legend and i click on it. But this is because of the way screen readers parse tags so as long as there is text on the tag i can click on it and if there is an onclick event bound to it it will work.
We have tested the function it  self though and it works so there is something bwith the button that is broken.


It seems to be working on their machine if they use the enter key to do an onclick(). There seems to be something wrong with the button itself. I get this error message.
[image]<https://user-images.githubusercontent.com/57549915/233404857-6d08dc5a-5311-43c6-87e5-5d7707bc7517.png>
Uncaught ReferenceError: printStateDiagram is not defined
at HTMLDivElement.onclick (diagram.php:486:93)

It would be best if the one who made the button took a look at it again.

b21oscpe:
@b21eriho

b21eriho:
I don't know what is wrong, the button works in my folder of the server. Pressing it prints a message about the amount of states and lines but nothing is shown in the options-pane. What is it I am meant to check?

b21oscpe:
If the button is working fine without any issues I suppose it's all good? I am wondering what caused the error for @a19clasv

a21aldsa (changes requested):
The part that counts the lines and states works fine. But the part that says what lines come from what and goes to what is not correct, I made a diagram that went as follows: start to A to B to C to end. Here is what showed up in the console after using the test case button: 

line goes from A
line goes to A
line goes to B
line goes from B
line goes to C
line goes from C

This is not correct.

a21aldsa:
And for reference, I also could not replicate the error talked about previously.

a21nibah:
After talking to issue responsible I have added a comment in the function stating that the output is not correct regarding connections. Since the issue only was to print the Count of states and edges this is something extra that goes beyond the scope of the issue.
Can someone test and see if the output works now if you click the test case button? It should work the same as generating the er table. I can not see any output but that is also true for the er table.

a19clasv:
It works for me now that I closed all the windows connected to the server and reopened diagram.php. It seems there was an issue with the server and I wasn't the only one that had problems with it. I checked the test case button after that and it works, but there's 2 spaces missing. An example output is "line counter is 4and state counter is3". There's a space missing after the 4 and one before the 3 in the example.

a21aldsa:
I tested it again and, yes, the output is placed in the same spot as for generate er text. The state and line count are correct. The connections are still not working but that was apparently concluded to be beyond the scope of this issue and will therefore be the subject of a future issue. With that I conclude this issue as solved and ready for merge. Good job.

a21nibah:
It should now be fixed.

Issue #13410, opened 2023-04-19
Frontend error-handler for github integration


a21rammo:
Connected to issue: #13194

PR #13479, opened 2023-04-20
G3 2023 v3 issue#13411

a21henhe:
The script fetchGitHubRepo.php now sends http-error codes and error messages that can be dealt and showcased by the frontend (which will be done in issue #13410 ). To test if this works, you need to intenionally break the code by setting the variable "$json" to false and test that error message, and then setting the variable "$data" to false and test that error message. These will be showcased in an alert when creating a course with a github-url or editing a course with a github-url. When no error occurs, the console.log in the "success"-part of the ajax-query in the file courseed.js should be fired and should put a response in the console.

Me and b21leowa have tested both these faults on the webserver since we need to push faulty code to the webserver in order to test and it seems to be working. You can also check that the returned object looks fine by clicking on the tab "Network -> fetchGithubRepo.php -> preview" in the developer tools. 

Note: even when the faults are generated, the user is still being prompted with the alert that the course has been updated : this will be dealt with in the above mentioned issue)

Co-authored by b21leowa

### DIFF
--- a/DuggaSys/courseed.js
+++ b/DuggaSys/courseed.js
@@ -98,13 +98,11 @@ function fetchGitHubRepo(gitHubURL)
 			type: "POST",
 			data: {'githubURL':regexURL, 'action':'getNewCourseGitHub'},
 			success: function(response) { 
-				console.log(response)
+				console.log(response);
 			},
 			error: function(data){
-				console.log(data)
 				switch(data.status){
 					case 422:
-						console.log("hallå")
 						alert(data.responseJSON.message);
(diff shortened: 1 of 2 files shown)


b21leowa:
My website https://cms.webug.se/root/G3/students/b21leowa/LenaSYS/DuggaSys/courseed.php has a specific error for testing. The error should message that the JSON is wrong. ($JSON = false)

a21lukpo (approved):
Tested on server and locally, worked both times.

Issue #13442, opened 2023-04-20
State diagram lines needs improvements


c21adano:
Currently when you create a line between to state elements a line without icons is made. State diagram are supposed to have a black arrow on one end of the line, currently to the black arrow has to be selected in the options. We want a black arrow to automatically be placed on the end point of every new line.

The first selected element is the start point and the second element selected while creating a line is the end point.
![basesdline](https://user-images.githubusercontent.com/102599332/233319246-2d3d722b-9c46-4b32-8abe-c3fc62bcce0b.PNG)
![sdlineimprovementneeded](https://user-images.githubusercontent.com/102599332/233319276-eb35ce25-993e-457f-bd0f-c2f3c46c40d8.PNG)

d21matbo:
What behaviour is preferred?
1. The line **only** gets the arrow when it is connected with two State entities
2. The line has the arrow all the time, this includes while it is being drawn

c21adano:
option 2 would be better, it would show the user were the arrow will be placed.

PR #13480, opened 2023-04-20
G1 2023 v3 #13442

d21matbo:
Drawing a line between two state entities should now show a arrow on the end of the line
The arrow is also set when the line is created

It is still possible to remove or add additional arrows on the line once it is created, if this behaviour is unwanted this could potentially be raised as a new issue.


a21aldsa (approved):
Works; the arrow is there automatically as soon as the line is begun to be drawn. Good job.

Issue #13465, opened 2023-04-20
"do:" and "exit:"


b21oscpe:
We want to seperate "do:" from "exit:" with two different textfields. Note that "exit:" does not have to be there and should not be there from the start, like "do:"

![image](https://user-images.githubusercontent.com/102578165/233346361-ff8c541f-c952-40c9-bb5b-f53c1bb872df.png)

![func](https://user-images.githubusercontent.com/102578165/233346377-d8570625-8d01-4d21-a97a-7f3de871a2d7.png)

PR #13483, opened 2023-04-20
G1 2023 v3 #13465

b21matfi:
has exit: and do: they are not in the same box could not fix it


b21oscpe:
Similar to the notation below, there should not be a line between "do" and "exit", this will be a future issue. We have divided "do" and "exit" into seperate input fields to reduce the possibility of incorrect user input.
image

Issue #13470, opened 2023-04-20
Lines should be selectable with box select


c21adano:
Related: #13440
"Most lines of the type UML, and even SD, cannot be marked with the selection tool."
"they should work like ER, as in, become selected when the selection tool is used."

a21aldsa:
My findings are that: 

It seems this issue is in how the box selction tool works. For example, it uses functions like intersectsBox that assumes lines have a x1, x2, y1, y2 attribute. Which is simply not the case for any UML or SD line. Marking any line with the box selection tool only works for ER.

I am not sure what the way forward is here, maybe write/rewrite functions for box select? Or add these attributes into all the lines?

PR #13489, opened 2023-04-21
G1 2023 v3 #13470

b21eriho:
Fixed box-select on the lines with corners, did this by adding the attributes that the box-select looks for, this probably isn't the best way to do it, however it does work.


b21oscpe:
dont forget the label and to link your issue in the future

b21eriho:
I created this branch from the issue, I presumed they would then auto-link, do they not?

Issue #13330, opened 2023-04-17
Present log feature: Search bar


b21kurar:
A search bar would be good to add at the top, which can filter out the results based on the query.

A "simple" solution could be just find if the query occurs in any row, and present them in the results.
![image](https://user-images.githubusercontent.com/129070844/232477539-b097686d-878d-41a6-a57c-642550f15b14.png)

a21ponkr:
![hesten](https://user-images.githubusercontent.com/102595853/232795723-f8438482-a570-4130-9d5f-0c8eb38d64c8.png)

Looks like this as of now, definitely needs fixing.

a21ponkr:
Searching for example "Fileed" works now, however it still shows up as two tables.
![hesten2](https://user-images.githubusercontent.com/102595853/232802725-db45e117-6ebe-474c-90d0-cf02b243000d.png)

a21ponkr:
Search randomly stopped working without having done any changes since yesterday

b21kurar:
> Is it needed to make the search bar in Javascript?

I didn't see this comment, but no it should not. We try to keep the code standard of the other services. We will change this logging system to follow a microservice architecture soon

a21ponkr:
Then I need to rethink my solution

a21ponkr:
So, I made it in Javascript anyway because I could not find a decent solution for it to be in only PHP, however if someone else finds a way they can comment it here, thanks.

a21ponkr:
![good enough](https://user-images.githubusercontent.com/102595853/233471829-7a3621d4-89ae-490d-9d72-09854337ded0.png)

a21ponkr:
Fixed column title not showing up when i search

PR #13500, opened 2023-04-21
G2 2023 v3 issue#13330 v3

a21ponkr:
Created a search functionality for log.php in issue #13330

### DIFF
--- a/Shared/log.php
+++ b/Shared/log.php
@@ -22,6 +22,7 @@
         <?php
             try {
 	            $log_db = new PDO('sqlite:../../log/loglena6.db');
+                var_dump($log_db);
             } catch (PDOException $e) {
 	            echo "Failed to connect to the database";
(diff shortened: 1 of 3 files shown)


a21ponkr:
this is in need of a review

b21kurar:
This pull request also makes changes to an unrelated file, this shouldn't be part of the PR

a21ponkr:
I have put the javascript portion to another file called logSearch.js
I don't know if I can do anything about the readme.md part since it was not my commit

a19hakja:
I'll set this up for review and if we need to convert to php we'll make an additional issue

a21jeaha (approved):
from what i could gather the search function does what was asked. 
The only problem i could find is that this is based on the old log.php. I cannot say if this will work with how that file functions now.

Issue #13461, opened 2023-04-20
Arrow should be filled


b21oscpe:
According to the notation for state diagrams the arrow should be filled, which means it should be completely white in darkmode. Rather than "transparent" or black and outlined.

![5f1e6b68e6f21c399bd698a7c613d2d1](https://user-images.githubusercontent.com/102578165/233339488-28f7393c-b977-4ec8-8e4d-59bb911a0b29.png)

c21adano:
I think the state diagarm reuses the the black arrow from the other diagarms changing it may affect them. so i propose another solution, in darkmode the state diagrams switch from the black triangle arrow to the white triangle arrow.

PR #13503, opened 2023-04-21
Change arrow to filled white #13461

c21theti:


b21matfi (approved):
Looks good! arrow is fully filled white

Issue #13495, opened 2023-04-21
Accessibility for buttons in the buttons-menu to the left need optimization


b21oscpe:
Currently we have some issues with how the buttons work with screenreaders, the following "pattern" to create a button is as following:

```
fieldset (Holds one or more buttons)
    legend (Which holds the title of the button)
    div (container for button, currently holds the onClick)
       img (SVG)
       span (Tooltip)


    ...etc
``` 

With the onClick on the "div" tag it is pretty difficult to interact with for screenreaders, a temporary solution to this is to put it on legend however this makes it so that the button itself is not clickable. This needs to be solved in order for our buttons to reach an acceptable level of accessibility.

PR #13506, opened 2023-04-21
 Accessibility for buttons in the buttons-menu to the left need optimization #13495 

a21aldsa:
Added correct alt to all img in the menu, this makes navigation with a screen reader possible. Also added aria-hidden="true" to the legends on the menu since these legends just make navigating a menu like this one with a screen reader harder. 

There are still many accessibility problems on this site; including but not limited to: the options panel being readable even when its  supposed to be hidden, the drop downs in the menu not being accessible via a keyboard and also all tool tips not being available on keyboard focus but only on mouse over (direct violation of ARIA). These are major issues for accessibility but they are sadly not covered by the issue I was assigned.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -4241,10 +4241,10 @@ function setReplayRunning(state)
     var stateSlider = document.getElementById("replay-range");
 
     if (state){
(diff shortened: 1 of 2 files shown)


b21eriho (approved):
I've tested the changes with a screen-reader and what was set out to achieve has been achieved as far as I can tell. However I naturally use the screen-reader differently than someone who is hard of seeing so there may be something I've missed. Overall good job

a21nibah:
I tested this while it was being developed and everything looks good. After the conflict is resolved  you can merge.

Issue #13486, opened 2023-04-21
Zoom-containers are of different dimensions


b21oscpe:
The containers / Icons(?) are not the same size 

Move the zoom-container to that its content is within the container of the one to the left, all zoom related buttons should be adjacent

![3687712871ba77db1b340f7333fee8cd](https://user-images.githubusercontent.com/102578165/233586407-4cebbad9-e75b-4b51-a8c0-312e27e5dec0.png)

PR #13507, opened 2023-04-21
G1 2023 v3 #13486

b21matfi:
Added the zoom indicator to the zoom container and changed some css

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -599,6 +599,12 @@
     height: 27px;
     width: 27px;
 }
+#zoom-message {
+    height: 27px;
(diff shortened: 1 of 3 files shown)


a19clasv (changes requested):
They're supposed to be to the left of the screen, not the right. "Move the zoom-container to that its content is within the container of the one to the left". The empty magnifying glass and the zoom multiplier are also supposed to be together, with the magnifying glass being on the right of the multiplier. Like this; 
image
not this;
image

a19clasv:
It's fixed

b21eriho (approved):
Looks good, functions as it should

Issue #13499, opened 2023-04-21
Change the variable names of a and b in sortvectors.


c21adano:
The sortvectors function has two variable named a and b whos names should be change to better fit their purpose

![aandbnames](https://user-images.githubusercontent.com/102599332/233627101-5a43ae43-abdd-4f26-a99a-bb4b828e68c2.PNG)

c21adano:
the a and b variable are also used in the sortElementAssociations function.

PR #13513, opened 2023-04-21
Change the variable names of a and b in sortvectors. #13499

a19clasv:
I was going to rename a and b as well as the other variables like toElementA and by, but I couldn't rename them nor find their definitions. Therefore I renamed a to currentElementIdA and b to compareElementIdB so that the other variables would still be correct.


a21aldsa (approved):
Its good, issue resolved. You should remove the, now faulty, TODO comment before merging though.

Issue #13493, opened 2023-04-21
data_returned function in diagram.js is never used.


c21adano:
data_returned is a function in diagram.js that is never used.

First confirme that i didnt overlook it being used somewhere, then delete the code if it truely isnt used

a21aldsa:
data_returned has 0 references and 0 implementations across all files. In fact, it is not mentioned more than once across all files, just when the function is created. This is safe to delete. I will create a PR

PR #13516, opened 2023-04-24
 data_returned function in diagram.js is never used. #13493 

a21aldsa:
removed the function after making sure it had no references, calls, implementations or mentions.


d21matbo (approved):
I was also unable to find usage of the function. Looks good!

Issue #13497, opened 2023-04-21
Investigate getlines ToDo comment in diagram.js


c21adano:
There is a comment next to the getLines function, investigate if the comment makes any reasonable sense.

![getlines](https://user-images.githubusercontent.com/102599332/233625695-e2952b26-690f-4351-9325-c1fc44784e2e.PNG)

a21aldsa:
After doing some digging around I have concluded that the comment has a point. The function returning a readily available variable makes it rather useless, and to strengthen that; the function is only called once. Should I rework this function to return a specific line via a parameter or should I remove this function and change the one place where its called to just use the variable straight up?

c21adano:
I think the rework could be healthy for the system, we already had a similar rework a couple of weeks ago regarding the colour check that done every constantly.

A rework will make the code easier to understand and work with.

PR #13520, opened 2023-04-24
 Investigate getlines ToDo comment in diagram.js #13497 

a21aldsa:
removed get lines since its only purpose was returning a readily available, global, variable and it was only called once in the whole codebase. I changed the code where it was called to just use the global variable "lines" and then removed the, now unused, function.


b21matfi (approved):
Nice work! testing to see if it broke the diagram which it did not.

Issue #13462, opened 2023-04-20
Arrows are either too big or lines have to be adjusted


b21oscpe:
![463ad3925e937926826a9e4d4891243b](https://user-images.githubusercontent.com/102578165/233341145-8cd04f5c-fc68-4fd6-86ac-99d5caaecb21.png)

The arrow is turning in an unwanted way. If possible, allow for diagonal lines but only when it is needed (Say for example that the last kink is too close to the entity, which results in the image above.)

Alternative solution would be to make the arrows smaller or adjust where the kink should be placed.

(FYI kink is the word HGustav uses for "turns")

b21oscpe:
note that this issue, which might relate, has been finished and merged #13442

PR #13522, opened 2023-04-24
 Arrows are either too big or lines have to be adjusted #13462 

b21phiro:
Issue: https://github.com/HGustavs/LenaSYS/issues/13462

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -7460,9 +7460,8 @@ function addLine(fromElement, toElement, kind, stateMachineShouldSave = true, su
  * @param {int} concideredNearValue
  * @return {boolean} result
  */
-function isClose(fromX, toX, fromY, toY) {
-
-    const concideredNearValue = 600,
+function isClose(fromX, toX, fromY, toY, zoom = 1) {
+    const concideredNearValue = 600 * zoom,
           deltaX = toX - fromX,
           deltaY = toY - fromY;
 
@@ -7540,7 +7539,8 @@ function drawLine(line, targetGhost = false)
         (fx + x1Offset),
         (tx + x2Offset),
         (fy + y1Offset),
-        (ty + y2Offset)
+        (ty + y2Offset),
+        zoomfact
     );
 
     // Collect coordinates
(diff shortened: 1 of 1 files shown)


b21phiro:
NOTE: The arrows do NOT rotate when diagonal.

a21aldsa (changes requested):
While yes, you can make diagonal lines between state elements now. And yes, as previously noted, the arrow is still bugged. This PR can't be approved, in my opinion, because this has negatively affected other parts of the system; the lines between IE and UML. These should not be able to go diagonal. The attached screenshot shows two IE elements with a wrong line and, subsequently, bugged icons.

image

a21aldsa:
Also, the lines feel very "floaty", as in, they dont snap to left, right, up, down, and now, corners, like i would imagine they should. But perhaps this on purpose as a part of a design choice you have made?

c21adano:
This pr has conflict with the stable branch, the first two conflict only occupy the same rows and can coexist, but the conflicts at 7919 and 8152 have real conflicts with the stable branch.

So the change that happend on the stable branch added -sd at the end of the class, while this pr has changed som of the numbers on "points". Im unsure if the stable version points numbers should be used or this pr points numbers. 

I therefor ask for help regarding distinguish which changes should be keept or changed to resolve the conflict. Of course the conflict comes after making the requested changes.

a21aldsa:
Oh and I think you should remove the log message that is still in the code before merging.

b21phiro:
Fixed so only SD elements go diagonally.

c21adano:
I resolved the 4 merge conflicts, the first conflict was due to two functions being created on the same line, the 2nd conflict was on ~7683 were this issue split remade and if into an else if, while another issue wrote more code on it, keept the else if with the new code.
3rd and 4th conflict was due to the size change on arrows, another issue added a new sd class to those arrows, i keept the size change and added back the sd class.
This will need a new round of testing to check for problems.

a21aldsa (changes requested):
The changes requested have been implemented. But, I noticed something that I did not before. The lines change drastically depending on current zoom level. Here's a screenshot showing the line as it is supposed to be, this is at 1x zoom:

image

And here is one, of the exact same line and location, just slight more zoomed in, at 1.5x:

image

I do not know if this is within the scope of this issue, perhaps it is a new issue but that is not for me to decide.

a21aldsa (approved):
still the same

Issue #13518, opened 2023-04-24
State diagram black colour in darkmode needs improvement.


c21adano:
When a state diagram state is made black in darkmode both the state text and border is hard to see.
![stateblack](https://user-images.githubusercontent.com/102599332/233934876-f56b3b38-55f5-4703-8e99-36cc2e9973be.PNG)

Other diagram typs have a white border and white text while being colour in darkmode.
![classblack](https://user-images.githubusercontent.com/102599332/233934900-857c79ce-40a6-4ae2-9b9a-e199d6f40368.PNG)

Change it so state diagram state also gets a white border with white text in darkmode while being black.

PR #13523, opened 2023-04-24
G1 2023 v3 #13518 darkmode black entity

b21eriho:
There was a style-tag and a few missing classes that prevented the automated systems from kicking in when the color of an entity was set to black. That has now been fixed


b21willu:
Looks good! It now gets a white border with white text in darkmode while being black.

PR #13534, opened 2023-04-24
G2 2023 v3 issue#13524 - Update tests for duggaedservice, using JSON response as expected output

a21jeaha:
I have removed the SQL-querys that were used for comparisons, same goes for the comparison it self.
They are now replaced by the saving of the output generated by the service file.

### DIFF
--- a/DuggaSys/duggaed.js
+++ b/DuggaSys/duggaed.js
@@ -647,9 +647,10 @@ function showVariantEditor() {
 }
 
 function returnedFile(data){
+	console.log(data);
 	retdata = data;
 	filearray = [];
-
+	
 	/*Because the 'filename' value is just raw JSON, it needs to be parsed. You can test this by printing retdata in the
 	console while in the variant editor. After parsing they're placed in array which then gets filtered for JSON files*/
 	for (var i = 0; i < retdata['entries'].length; i++){
--- a/DuggaSys/duggaed.js
+++ b/DuggaSys/duggaed.js
@@ -647,7 +647,6 @@ function showVariantEditor() {
 }
 
 function returnedFile(data){
-	console.log("hejhopp"+data);
 	retdata = data;
 	filearray = [];
 	
--- a/DuggaSys/duggaed.js
+++ b/DuggaSys/duggaed.js
@@ -647,7 +647,7 @@ function showVariantEditor() {
 }
 
 function returnedFile(data){
-	console.log(data);
+	console.log("hejhopp"+data);
 	retdata = data;
 	filearray = [];
 	
--- a/backend-models/service-tests/duggaedservice-tests-SQL-query-version.txt
+++ b/backend-models/service-tests/duggaedservice-tests-SQL-query-version.txt
@@ -12,6 +12,8 @@ duggaedservice.php
     { (checklogin) IS TRUE && 
     (hasAccess($userid, $cid, 'w') || isSuperUser($userid) || hasAccess($userid, $cid, 'st')) } IS TRUE
 
+*checking current number of quizzes 	
+    SQL-query: SELECT count(id) FROM quiz;      // counts the number of quizzes, save it to $preCount 
 
 *******
 * send parameters inputs to duggaedservice.php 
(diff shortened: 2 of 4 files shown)


c21axepe:
För att kolla send output som du skrev tex:
send
{
    $opt = "SAVDUGGA";
    $qid = null; or $qid = “UNK”;
    $cid = 1885;
    $userid = 2;        // this is automatically added depending on what session is active (if any), we want the value to be 2
    $courserver = 1337;

} 
Har du kollat outputs eller har du bara som det som kan vara i koden?

a21jeaha:
In the first draft of the document, a comparison was made on the total number of entries in a specific table before and after the service was executed. But after information received yesterday we now only collect the output generated by the service.
As of now there is no implemented solution to record that output. as such i haven't specified how it is to be collected. only that its suppose to be collected.

DuggaSys/duggaed.js
c21axepe, line 1: Changes by mistake?

DuggaSys/duggaed.php
```diff
@@ -1,3 +1,4 @@
+
```
c21axepe, line 1: Changes by mistake?

DuggaSys/duggaed.js
```diff
@@ -649,7 +649,7 @@ function showVariantEditor() {
 function returnedFile(data){
 	retdata = data;
 	filearray = [];
-
+
```
c21axepe, line 652: Changes by mistake?

DuggaSys/duggaed.php
```diff
@@ -9,6 +10,7 @@
 <script>var cid = <?php echo $cid ?>,vers = <?php echo $vers ?>;</script>
 <?php
 ?>
+
```
c21axepe, line 13: Changes by mistake?

backend-models/service-tests/duggaedservice-tests.txt
c21axepe, line 1: for create an assignment, template missing?
coursevers instead of courserver at row 25
add release, deadline, qstart to send at 97-108
change disable to disabled at row 236
change answare to answer at row 238
agreeing with the question mark at 340, why is the DELETE used on userAnswer

c21axepe (changes requested):
Overall good, minor changes requested!

backend-models/service-tests/duggaedservice-tests.txt
b21antbr: Both of us have been reviewing this together. To add on this out of curiosity, row 316 what is the purpose of $variantID, what is the purpose of $variant at 432? Do these new variables have to be php variables?

backend-models/service-tests/duggaedservice-tests-SQL-query-version.txt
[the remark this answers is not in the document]
a21jeaha: yes it is. seeing as future testing of the service is a possibility i did not remove it outright.
instead just changed the name of it.

DuggaSys/duggaed.js
a21jeaha: yes, i was working on another issue and forgot to change brash ;)

backend-models/service-tests/duggaedservice-tests.txt
a21jeaha: "agreeing with the question mark at 340, why is the DELETE used on userAnswer"

it removes the answer given for a specific variant (a kind of test)

a21jeaha: when a new "variant" is created it automatically receives a primary key value ("vid" in the table variant) an auto incremented value. we save that ID to     $variantID. 
that value is needed as an input to know which row is to be changed.  

its almost the same situation with $variant, and no they dont NEED to be php variables, but it makes it easier to distinguish them as variables in the doc.

a21jeaha:
changes implemented

c21axepe (approved):
Looks good!

b21antbr (approved):
Looks good, well done!

PR #13537, opened 2023-04-24
G3 2023 v3 issue#13418 connect the link in getlatestcommit with link from mysql

c21rebja:
With a github url we're getting the connected course id from the MySQL database, then getting the latest commit from the same URL. This will later be inserted into the SQLite database, but so far this is a standalone function that isn't connected to the rest of LenaSYS. 

To test this, create a new course with the link: "https://github.com/HGustavs/Webbprogrammering-Examples". Then go into the getLatestCommit file (you can find this under recursive testing) and compare if the course id (from mysql), url and latest commit are correct. If multiple courses share the same url, multiple ids will show up.

image

co-author @g21emmka

### DIFF
--- a/recursivetesting/getLatestCommit.php
+++ b/recursivetesting/getLatestCommit.php
@@ -15,10 +15,11 @@
 	pdoConnect();
 	session_start();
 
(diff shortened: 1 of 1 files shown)


b21leowa (approved):
Works locally and on the webbserver. I Approve these changes 👍

PR #13538, opened 2023-04-24
Created $info variable.txt

b21hamwe:
Created txt. file with an example of how we could solve the $info variable


a19hakja:
For clarification, this issue was at first interpreted wrongly. a txt. file, as this pull request provides wasn't what was requested. Thereby, after some discussion this pull request was closed

d21erisu:
According to the developer he misinterpreted the issue and requested this pull-request to be closed. So i closed it.

Issue #13543, opened 2023-04-24
Add borders around boxes on the profile page


c21rebja:
It's hard to see the edges on the boxes on the profile page with darkmode on. Add a white border to these. 
To make it coherent with both modes, add a dark border to the styling in light mode. 

![image](https://user-images.githubusercontent.com/102578746/234012546-e6686488-be70-4990-a886-6a93ae6cb34d.png)

co-author @g21emmka

PR #13544, opened 2023-04-24
adding borders to boxes on profile page

c21rebja:
#13543 

Added borders on both light and dark mode on profile page.

image

image

co-author @g21emmka


b21geoyo (approved):
Borders around the boxes for both dark and light mode have been implemented and the issue is now resolved.

Issue #13382, opened 2023-04-18
Cardinalities aren't highlighted when selected


b21oscpe:
This issue is related to #13376, read here for further information

(Cardinalities & Icons)

![232813041-5d5a9e0c-cbd6-46ce-9461-0fb852cae3bd](https://user-images.githubusercontent.com/102578165/232817986-b31d8086-d512-435f-b62e-6793609c0da1.png)

PR #13546, opened 2023-04-24
Highlighting cardinality and icons #13382

a21oligu:
Fixed highlighting for cardinality and icons when selecting line


a21aldsa (approved):
Works fine, good job.

PR #13551, opened 2023-04-24
G2 2023 v3 issue#13393

a20fanma:
Created a text-file of tests for accessedservice

### DIFF
--- a/backend-models/service-tests/accessedservice-tests.txt
+++ b/backend-models/service-tests/accessedservice-tests.txt
@@ -46,11 +46,15 @@ Check the test result if a user was added with username or ssn, by using SQL-que
 ---------------------- 
 
 Test 3: try to update a created user_course by changing the access
+Row: 83 - 99 and 258 - 278
+Pre-req: { ($uid, "UNK") }
 To make sure the updating the user_course works and how the database will respond. 
 It can be done with SQL-query to create a user_course: 
    SQL-query: INSERT INTO user_course (uid, cid, access, vers, vershistory) 
       VALUES (:uid, :cid, ‘R’, :vers, ‘ ’) ON DUPLICATE KEY UPDATE vers=:avers, vershistory=CONCAT(vershistory, CONCAT(:bvers, ‘, ‘))
 It should display an error message if the insert of a user to the user_course into the database was unable.
+
+Pre-req: { ($opt, "UPDATE")==0 }
 After the user_course was created, then make an update of the user_course by using SQL-query: 
   SQL-query: UPDATE user_course SET access=:access WHERE uid=:uid AND cid=:cid
 By updating the access of the user_course. The values of access, uid and cid should be checked. 
--- a/backend-models/service-tests/accessedservice-tests.txt
+++ b/backend-models/service-tests/accessedservice-tests.txt
@@ -21,3 +21,16 @@ Then make an update by useing SQL-query.
 To update the username of the user. The values to be checked are username and uid.
 If the update was unable to succeed, it should display an error message which says 
 that the user does not exist in the database.
+
+---------------------
+
+Test 2: try to add two users with same ssn or username
+By adding two users with same ssn or username to see how the database will react.
+This can be done with the SQL-query to insert a user.
+   SQL-query: INSERT INTO user( username, email, firstname, lastname, ssn, password, addedtime, class) 
+       VALUES ( :username, :email, :firstname, :lastname, :ssn, :password, now(), :class)
+Both the username and ssn values should be checked, to avoid duplcation. If there are two or more
+users with the same ssn or username, it should display an error message. Which the database will indicates 
+there is a duplicate ssn or username.
+
+---------------------- 
--- a/backend-models/service-tests/accessedservice-tests.txt
+++ b/backend-models/service-tests/accessedservice-tests.txt
@@ -34,3 +34,13 @@ users with the same ssn or username, it should display an error message. Which t
 there is a duplicate ssn or username.
 
 ---------------------- 
+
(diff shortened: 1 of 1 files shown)


backend-models/service-tests/accessedservice-tests.txt
```diff
@@ -0,0 +1,87 @@
...
+
+********
+* pre-req
+********
+
+(checklogin() is TRUE && (hasAccess($userid, $cid, 'w') || isSuperUser($userid,) is TRUE))
+
+--------------------
+
+Test 1: try to update a created user by changing the username
```
b21kurar, line 14: To be able to call this a test case we need actual values, not variable names. I would recommend to look at how this looks. https://github.com/HGustavs/LenaSYS/blob/26a2a6c97af16d1be7871ebf481d2a253056c7eb/backend-models/service-tests/duggaedservice-tests.txt#L23 

Here there are actual values prepared to be used for the test case

backend-models/service-tests/accessedservice-tests.txt
```diff
@@ -0,0 +1,129 @@
+
+accessedservice.php
+
+-------------------
+
+********
+* pre-req
+********
+
+(checklogin() is TRUE && (hasAccess($userid, $cid, 'w') || isSuperUser($userid,) is TRUE))
```
a21jeaha, line 10: To make sure added text arent confused with code i write it with only caps and each statement on a different row.
as such.
(checklogin() IS TRUE 
&& 
(hasAccess($userid, $cid, 'w') || isSuperUser($userid,) )) IS TRUE

backend-models/service-tests/accessedservice-tests.txt
```diff
@@ -0,0 +1,129 @@
...
+
+--------------------
+
+Test 1: try to update a created user by changing the username
+Row: 62 - 80 and 220 - 238
+Send {
+   $opt = "ADDUSER"
```
a21jeaha, line 17: $opt = "ADDUSER"   should it not be "UPDATE"?
image

backend-models/service-tests/accessedservice-tests.txt
```diff
@@ -0,0 +1,129 @@
...
+********
+* pre-req
+********
+
+(checklogin() is TRUE && (hasAccess($userid, $cid, 'w') || isSuperUser($userid,) is TRUE))
+
+--------------------
+
+Test 1: try to update a created user by changing the username
+Row: 62 - 80 and 220 - 238
```
a21jeaha, line 15: only one service at a time. if the same service is used at another point in the file, then write the related test information there again.

backend-models/service-tests/accessedservice-tests.txt
```diff
@@ -0,0 +1,129 @@
...
+Send {
+   $opt = "ADDUSER"
+   $uid = "UNK"
+   $regstatus = "UNK"
+}
```
a21jeaha, line 20: Where in the text does pre-req end and sending of input start?

backend-models/service-tests/accessedservice-tests.txt
```diff
@@ -0,0 +1,129 @@
...
+To make sure the update of a user works and what response it will get form the database.
```
a21jeaha, line 21: as of yesterday we no longer compare the results. we send in exampledata to execute a service, and collect the output generated by the service. 

this is the information relevant in this file.
image

backend-models/service-tests/accessedservice-tests.txt
```diff
@@ -0,0 +1,129 @@
...
+It can be done by creating a user with SQL-query.
+   SQL-query: insert into user( username, email, firstname, lastname, ssn, password, addedtime, class) 
+       VALUES ( :username, :email, :firstname, :lastname, :ssn, :password, now(), :class)
+
+Send {
```
a21jeaha, line 26: first of, where are the proper variable values, have you actually tried inserting the values bellow in a query?
you need to have proper tested values that
1. make sense in the context
2. is a valid input

you are aware that these are if else statements? 

image

only one of these can execute at a given time. which means that you ether need to make a test for each individual if else statements.
The alternative is to specify what variable values each variant of the test will use, the person writing the test is not suppose to have to figure this out.

backend-models/service-tests/accessedservice-tests.txt
```diff
@@ -0,0 +1,129 @@
...
+   $prop = "class"
+}
+Then make an update by useing SQL-query.
+   SQL-query: UPDATE user SET username=:username WHERE uid=:uid
+To update the username of the user. The values to be checked are username and uid.
+If the update was unable to succeed, it should display an error message which says 
+that the user does not exist in the database.
+Compare the result to make sure the test worked, by using SQL-query.
+   SQL-query: SELECT uid FROM user WHERE ssn=:ssn 
+
```
a21jeaha, line 41: you missed this entire section 

image

the same as the privies comment apply.

a21jeaha (changes requested):
The document needs to be redone from the ground up...

i have commented quite a bit, but you need to follow the established standard.. 
the link below is what is expected. 

https://github.com/HGustavs/LenaSYS/blob/20e001c53daea1505f107e0b18a7405df9e08b46/backend-models/service-tests/duggaedservice-tests.txt 

the values sent as input need to be tested to make sure that they are valid, so you need to have a mysql client installed to try out inserts, make sure that they work and verify changes in LenaSYS.

a20fanma:
I think I will give up on this issue. I have tried to finish the issue but it needs to redo. Because I have misunderstood how I was suppose to write these tests, and I do not think it will get any better if I continue.

a20fanma:
But I will try to finish this issue

Issue #13517, opened 2023-04-24
The output for what is shown in the model is wrong


b21oscpe:
We want the output to be something similar to: 

INITIAL STATE GOES TO STATE 1
STATE 1 GOES TO STATE 2
STATE 2 GOES TO STATE 3
STATE 3 GOES TO FINAL STATE

![2c009810a6b358391698591eb8605293](https://user-images.githubusercontent.com/102578165/233928959-e408bad1-f214-4909-a92c-8147c40b55f4.png)

PR #13558, opened 2023-04-25
G1 2023 v3 #13517

d21matbo:
Remade the function that generates the state diagram info so that its output matches the desired output listed in #13517.


b21eriho (approved):
Looks good 👍

Issue #13539, opened 2023-04-24
Make text scalable in states


b21oscpe:
When a width limit is reached it should be stretched vertically to make more space

![4f15c512b0cc23177e10502e98ceb38f](https://user-images.githubusercontent.com/102578165/234005759-ba5e1308-26f4-4035-93e2-ec52651f9827.png)

PR #13560, opened 2023-04-25
When a width limit is reached it stretches vertically to make more space

b21phiro:
Issue: https://github.com/HGustavs/LenaSYS/issues/13539


a21aldsa:
I believe IE, UML and ER should have this functionality as well, perhaps for future issues? @b21oscpe

b21oscpe:
> I believe IE, UML and ER should have this functionality as well, perhaps for future issues? @b21oscpe

For sure

PR #13566, opened 2023-04-25
G3 2023 v3 issue#13410

b21leowa:
Both of us have tested on the webserver with $json and $data set to false as per the commits. You can test this as well, by changing these variables to false in the webserver: do sudo -i , then go into your studentfolder, switch to the correct branch, go into the recursivetesting folder and then nano into the file FetchGitHubRepo.php and set these variables to false before each of their if-statements (but only one at a time). That way, you can test that these error message occur at the right time. When done, just do "git restore . " on the webserver

If the link is correct and both JSON and data is true, the course will be added. But there is a slight chance that the course won't be published because of the API has to many request recently. To solve it you need to add your API token on the authorization in the fetchGithubRepo.php. You can do it on the server by nano the file and add it. But we prefer to add it only on the local machine so you won't have your API token in clear text on the webbserver.

When fetching a githubrepo fails, the request is canceled and the user is prompted why with an alert. But if there is no alert when trying to publish/update a course you'll need to clear cookies and refresh. 

Also test by: 
- [ ] Adding a new course without putting in a githublink 
- [ ] Edit a course but dont put a githublink

Co-author: @a21henhe

### DIFF
--- a/DuggaSys/courseed.js
+++ b/DuggaSys/courseed.js
@@ -39,24 +39,24 @@ function updateCourse()
 	var courseid = "C"+cid;
 	// Show dialog
 	$("#editCourse").css("display", "none");
+	
 	if(courseGitURL) {
-		dataCheck = fetchGitHubRepo(courseGitURL);
-		if(dataCheck) {
+		if(fetchGitHubRepo(courseGitURL)) {
 			$("#overlay").css("display", "none");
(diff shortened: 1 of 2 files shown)


d21emijo:
## only data = false
image
this gives no alerts/errors, and doesnt create a new course

## only data = false
image
this gives the alerts in order
image
image
and creates no new course

## only JSON = false
image
gives the alerts in order 
image
image

## only JSON = false
image
gives no alerts/errors and doesnt create a new course

PR #13570, opened 2023-04-25
G3 2023 v3 Implemented the ability to delete files from the github folder

a21olljo:
It is now possible to use lenasys "file editor" to delete files located inside of the github folder found under the course specific folder.


a21olljo:
After conversing with @c21sebar we realised that the instance of the server that the branch were on missed some sqlite functionalities, which where solved with a restart. Testing of the PR can now continue.

g21emmka (approved):

c21rebja (approved):
It looks good. Created a course with a single file in the github repo, it showed up in the courses/1/github folder. Went into the file editor for course 1 on duggasys, found the file and deleted it. It also disappeared from the courses folder. So everything works like it should. :)

All files are saved in the same place (under course 1), but that is an issue for later.

g21emmka (approved):
I did the same steps when I was testing as @c21rebja. 
I was able to delete the file I added to the course and the file disappeared from the github folder and also from the folder on the course site. 
All files were saved in the same folder (1), but this might be an issue for later. 
Everything else looked good and I was able to delete the file that I added.

Issue #13569, opened 2023-04-25
Test-case button crashes the system and doesnt work


b21oscpe:
When attempting to press the "test case" button it does not even load the output field anymore. On top of that it crashes the system in a way that makes placing entities impossible. I think this issue stems from an incorrect solution to one or more merge-conflicts.

Investigate what is wrong and attempt to fix it.

Related issues that may help: 

#13517 
#13525 

![e69245bf795bdd2ad458eb4317eaa1c1](https://user-images.githubusercontent.com/102578165/234275349-ba61d81e-a8e0-4751-9ca5-9828acbe17e0.png)

PR #13571, opened 2023-04-25
Test-case button crashes the system and doesnt work #13569

a21aldsa:
The function was rewritten in a way that stopped using lineCounter and opted instead to use stateLines.length, this person also removed lineCounter but did not remove it from output. This meant that it tried to use an undefined variable as output which broke the system. This is now fixed and the test case button now works as intended.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -4026,7 +4026,7 @@ function generateStateDiagramInfo()
             }
         }
     }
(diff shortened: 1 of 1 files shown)


d21matbo (approved):
The "problematic" lines appears to have been introduced by a merge conflict in #13533 They are now removed and the system works correctly.

Issue #13588, opened 2023-04-26
Remove double line option from state diagram lines


c21adano:
There is an option on state diagram line to change the line kind to be double, but this doesnt work for state diagram lines.
![doubleline](https://user-images.githubusercontent.com/102599332/234541517-22bb4fe3-4ef5-4bc8-8943-326a0a9cf280.PNG)

b21matfi:
Should i remove all the options? or only the double? ¨

c21adano:
Remove all option is fine, removing double would only leave the normal kind anyway so it may be better to just remove all options.

PR #13590, opened 2023-04-26
G1 2023 v4 #13588

b21matfi:
Created an if statement so that SD does not get any line kind


c21theti (approved):
Looks good. Double line option is removed

PR #13595, opened 2023-04-26
G2 2023 v4 issue#13491

b21hamwe:
Done researching and making changes in-order to label the $info variable so that we in our log.php file can see what all the other variables in $info carry in form of data. 
This new "styling" do not change the earlier tables of data, but every new log will be printed out in this new format.

### DIFF
--- a/DuggaSys/accessedservice.php
+++ b/DuggaSys/accessedservice.php
@@ -44,7 +44,7 @@ $access = false;
 $debug="NONE!";
 
 $log_uuid = getOP('log_uuid');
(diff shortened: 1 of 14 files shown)


b21tilbe (approved):
The code snippets are syntactically correct. It seems to work correctly with JSON too. Approved

b21hamwe:
? JSON ?
The solution doesn't send any JSON data ? 

What do you mean with, "seems to work correctly with JSON too." ?

b21tilbe:
Sorry, has been reviewing wrong file.

b21hamwe:
d21erisu, ready to merge.

Issue #13420, opened 2023-04-19
Save last commit when course is created


c21sebar:
For issue https://github.com/HGustavs/LenaSYS/issues/13413

Save last commit when course is created / a course is updated with a github repo url

c21rebja:
The branch connected to this issue has been abandoned, because of complications. Parts of the code is instead split up used in different issues connected to issue #13413 . 

(co-author @g21emmka )

Issue #13419, opened 2023-04-19
Get previous commit from SQLLite


c21sebar:
For issue: https://github.com/HGustavs/LenaSYS/issues/13413
Get previous commit from SQLLite to be used when comparing to the newly fetched

PR #13600, opened 2023-04-26
G3 2023 v4 issue#13419 get previous commit from sqllite

g21emmka:
issue #13419, issue #13420 is also resolved in this pull req.

Added sqlite connection to getLatestCommit. Saved the data (cid, url, latest commit nr) into sqlite db and then makes a select to bring the latest commit nr back. This is in preparation for a later issue. 

When testting you should see the following things: 
image

Don't mind the error, it exists because the course already has a row in the database, so the course id isn't unique. This will not be an issue once the information in the database gets removed when a course is deleted.

Test this by going to getLatestCommit.php. For the code to work you need to have an entry in the MySQL database that matches the url "https://github.com/c21sebar/test" since it's hardcoded.

co-author @c21rebja

### DIFF
--- a/recursivetesting/getLatestCommit.php
+++ b/recursivetesting/getLatestCommit.php
@@ -16,7 +16,7 @@
 	global $pdo;
 	$pdolite = new PDO('sqlite:../../githubMetadata/metadata2.db');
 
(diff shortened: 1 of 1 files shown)


a21olljo:
Looks good, works as intended and the data is stored in both MySQL database and SQLite database and it is demonstrated that information can be retrieved from respective table.

Issue #13567, opened 2023-04-25
Add icon images to repository


c21adano:
Related: #13529
Issue#13529 created icons for a dropdown rework, the icons need to be added to the project files. They currently wont have any use but will be used in the future for a icon dropdown rework.

Name them something along the lines of "iconDropDown_Arrow", or a better name if you can think of one. 

The icons can be found as a comment on issue#13529, they currently have alot of transparent background which should not mater.

a19clasv:
I didn't make it in time for the deadline, so I made the rest of the icons instead and renamed the ones from issue #13529.
![iconDropDown_None](https://user-images.githubusercontent.com/57549915/234287201-5d076124-ab0c-480f-bbc2-d33114e751b3.svg)
![iconDropDown_Weak](https://user-images.githubusercontent.com/57549915/234287209-0586dd4c-809c-4be2-90a8-a4f37409128f.svg)
![iconDropDown_M](https://user-images.githubusercontent.com/57549915/234287224-84b776e6-b05a-42b5-ae15-da7577bc878b.svg)
![iconDropDown_1M](https://user-images.githubusercontent.com/57549915/234287231-b93d830e-1d79-4931-a328-a8af8b5cf8df.svg)
![iconDropDown_1!](https://user-images.githubusercontent.com/57549915/234287242-88ae45ba-b558-423e-9b3f-c0d91a24e7af.svg)
![iconDropDown_1](https://user-images.githubusercontent.com/57549915/234287264-e62999bf-410c-439a-87fa-a162beb6a4f8.svg)
![iconDropDown_1](https://user-images.githubusercontent.com/57549915/234287329-e77ced83-bb1c-436f-b97e-87aa551b5201.svg)
![iconDropDown_01](https://user-images.githubusercontent.com/57549915/234287467-54350ca3-fd9c-4c5a-b03b-369558a489db.svg)
![iconDropDown_0M](https://user-images.githubusercontent.com/57549915/234287533-ed156e88-4279-4997-a4b0-a5aba91521e5.svg)
![iconDropDown_Arrow](https://user-images.githubusercontent.com/57549915/234291853-a1094713-dc69-41e4-8987-11d15b88b3e0.svg)
![iconDropDown_BlackDiamond](https://user-images.githubusercontent.com/57549915/234292012-6564939e-0aa9-4ac7-a07f-e924804bdfaf.svg)
[...]

PR #13603, opened 2023-04-26
Add icon images to repository #13567

a19clasv:
Added the images to the repository


a21aldsa (changes requested):
The viewbox for these icons is way too big compared to the size of the icons and the icons themselves are not placed at 0.0, I believe this will cause issues when these icons are used later.

Please move all the icons to 0.0, and shrink the viewbox so it fits snug.

c21adano:
> The viewbox for these icons is way too big compared to the size of the icons and the icons themselves are not placed at 0.0, I believe this will cause issues when these icons are used later.
> 
> Please move all the icons to 0.0, and shrink the viewbox so it fits snug.

that creation of the icons was another issue. If they need to be fix it will be done at a later time.

a21aldsa:
> > The viewbox for these icons is way too big compared to the size of the icons and the icons themselves are not placed at 0.0, I believe this will cause issues when these icons are used later.
> > Please move all the icons to 0.0, and shrink the viewbox so it fits snug.
> 
> that creation of the icons was another issue. If they need to be fix it will be done at a later time.

fair enough, ill approve the PR

a21aldsa (approved):

Issue #13583, opened 2023-04-26
Create activation-button according to #13577


b21oscpe:
See #13577

Issue #13577, opened 2023-04-26
Create a button for sequence diagrams 


b21oscpe:
The sequence diagram has several buttons to be implemented in order to make it functional: 

A "lifeline" which consists of an actor and a dotted line.

"Activation"

"Object" 

"Option loop or alternative"

This totals to 4 buttons based on the minimal notation that we currently have #13535 

Make a suggestion of where these buttons should be placed and/or alternative buttons. (For example: Would you like to combine two of the buttons into the same button or would you want there to be a 5th button?)

Comment your proposed design and then implement it when it has been accepted

![0d5ebddb454683a913139af5bf7dfd4a](https://user-images.githubusercontent.com/102578165/234532837-25ea9f03-8372-48f2-9d9e-ae98b8997c8d.png)

c21adano:
I would suggest only adding the object entity to start with, then add the others after the entity type has been made with the implementation of the object entity.

c21adano:
This issue is no longer about adding all button, just the Object button to get a start for sequence diagrams. The other button are their own issue that will link back to this one later on.

Issue #13535, opened 2023-04-24
Minimal model for Sequence Diagram


c21adano:
Related: #13521 & #12922
Make a minimal model with the notation that we aim to support, and also publish this here in this issue and also as part of the repository (standardized image format such as svg or png)

a21aldsa:
[sequenceDiagram.pdf](https://github.com/HGustavs/LenaSYS/files/11319965/sequenceDiagram.pdf)

![sequenceDiagram](https://user-images.githubusercontent.com/102581306/234208114-4b3def50-2590-4c4f-ba57-0d6e8e629e2f.png)

[sequenceKey.pdf](https://github.com/HGustavs/LenaSYS/files/11319968/sequenceKey.pdf)

![sequenceDiagramKey](https://user-images.githubusercontent.com/102581306/234208163-bdf80fcd-4d1e-4518-a087-23b12419a2a1.png)

PR #13604, opened 2023-04-26
Create activation-button according to #13577 #13583

a19clasv:
Made the button. The actual functions for it can be added at a later time.


a21aldsa (changes requested):
An activation, according to the minimal model created in #13535 has rounded corners, I believe the icon should reflect this.

a19clasv:
It looks rounded now

c21adano:
I fixed the conflict </fieldset> was conflicting with </fieldset>, so there should be no problem.

a21aldsa (changes requested):
> It looks rounded now

image

No it doesn't. It looks beveled; blurry, almost pill shaped and out of place. Please just make it rounded, like the minimal model in #13535 dictates. Use rx or ry if you do not want to use a vector drawing program.

c21adano:
> 

But isnt that how activation entity looks? or am i missing something here?

a21aldsa:
> > 
> 
> But isnt that how activation entity looks? or am i missing something here?

an activation, according to the minimal model, looks like this: 

image

It is a rectangle with rounded corners.

a19clasv:
I did true rounded corners before I made it beveled, but it was impossible to tell it was rounded. Therefore I beveled it so you could tell its edges weren't 90 degrees.

a21aldsa:
> I did true rounded corners before I made it beveled, but it was impossible to tell it was rounded. Therefore I beveled it so you could tell its edges weren't 90 degrees.

Here is a similar icon that was made for another issue, looks rounded to me

image

c21adano:
I talked with @b21oscpe: Personally i would not care, but it's rounded in the notation and it would be bad practise not to follow it.

a19clasv:
It's rounded now

a21aldsa (changes requested):
The top and bottom border are way too thick. Also, I couldnt help but notice that your svg is very convoluted with a complex transformation matrix: `transform="matrix(0 0.896024-.718979 0 15.40289 2.130293)"`

Here is a picture of your icon, note the borders at the top and bottom: 

image


Try something like this instead, I believe you will find it a little easier to read and that it does not have the thick top and bottom borders. Bear in mind this is a rough coded sketch and I have not yet tested it on lenasys. 

`
<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" version="1.1">
    <rect style="fill:none; stroke:#ffffff; stroke-width:2;" width="8" height="20" x="8" y="1" rx="2"/>
</svg>
`

a21aldsa:
remove the width and height for the actual svg tag if it does not work properly

a21aldsa:
> It seems to work properly

glad to hear it, ill test again

a21aldsa (approved):

a19clasv:
I didn't remove the height and width

a21aldsa:
> I didn't remove the height and width

I can see that, I can also see that it looks good in the menu. Good job

PR #13505, opened 2023-04-21
Create Update-profileservice-tests.txt

b21hamwe:
Written a txt test file for profileservice.php

PR #13615, opened 2023-04-27
Create profileservice-tests.txt.txt

b21hamwe:
Changes made to the document.

### DIFF
--- a/backend-models/service-tests/profileservice-tests.txt.txt
+++ b/backend-models/service-tests/profileservice-tests.txt.txt
@@ -1,0 +1,232 @@
+profileservice-tests.txt
+
+
+Before tests start, create 2 dummy accounts, one as teacher, another as student. Examples for names, 
+name = TestSTUD, PASSW = testpasswS
+name = TestTEACH, PASSW = testpasswT
+
+
+
+--------------------------------------------
+Row : 46 Check if it blocks teacher
+--------------------------------------------
+---------------------
+Name : teacherBlockerTEST
+---------------------
+
+*********
+* pre-req:
+********* 
+pdoConnect();
+session_start();
+
+if(isset($_SESSION['uid'])){
+	$userid=$_SESSION['uid'];
+}else{
+	$userid="1";		
+} 
+
+*********
+End pre-
+********* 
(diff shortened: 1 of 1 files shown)


a19hakja:
Accidental PR overload, Here's the original one

https://github.com/HGustavs/LenaSYS/pull/13505

Closing.

a19hakja:
Nevermind, we decided to go with this pull request since it's issue branch is taken from v4 and not v3 like that last one was. My b

backend-models/service-tests/profileservice-tests.txt.txt
```diff
@@ -0,0 +1,232 @@
...
+	$userid=$_SESSION['uid'];
+}else{
+	$userid="1";		
+} 
+
+*********
+End pre-
+********* 
+
+if(checklogin()) {
```
a19hakja, line 33: These are supposed to be independent of language, which this isn't, it has alot of copy-pasted code from the file which it also isn't really supposed to have. What we need really, is to provide values for the file to be able to perform intended function. And the pre-requisites needed to get there.

I'm comparing with the accepted test-file we have so far: https://github.com/HGustavs/LenaSYS/blob/master/backend-models/service-tests/duggaedservice-tests.txt

backend-models/service-tests/profileservice-tests.txt.txt
```diff
@@ -0,0 +1,232 @@
+profileservice-tests.txt
+
+
+Before tests start, create 2 dummy accounts, one as teacher, another as student. Examples for names,
```
a19hakja, line 4: Do we need to create new accounts every time we test? Isn't creating a teacher dummy account the same as fulfilling this kind of pre-req?

{ (checklogin) IS TRUE && 
    (hasAccess($userid, $cid, 'w') || isSuperUser($userid) || hasAccess($userid, $cid, 'st')) } IS TRUE

a19hakja (changes requested):
I sort of understand what we have here, but it isn't what we were going for. Maybe you can consult with @a21jeaha who made the first test-value document successfully.

Issue #13562, opened 2023-04-25
Fix text-containers for other diagram types aswell


b21oscpe:
This issue relates to #13560 and the same concept should be applied to other diagram types.

PR #13616, opened 2023-04-27
G1 2023 v4 #13562 text-breaks in entites

b21eriho:
I've copied Philip's function that he made in issue #13539 to UML and IE entities. Since these entities are left-aligned instead of centered I've given it some more room before folding because I thought it looked incorrect otherwise.


b21matfi (changes requested):
Looks like all the attributes are fixed but the names go out, for an easy fix you can just add a new variable named elemName = element.name.length and then just add what you have done with the function. (this is an idea not sure if it works)

b21matfi (approved):
Looks good, Good job!

Issue #13591, opened 2023-04-26
dropdown for diagonal state diagram lines.


c21adano:
State diagrams currently switch between diagonal or kinked lines based on distance between each element, it would be better if there was an option to choose which line you want.

PR #13618, opened 2023-04-27
G1 2023 v4 #13591

d21matbo:
It is now possible to change the line type using the options panel on the right.

The previous bahaviour (line type based on distance) is kept while placing the element and once it is placed the type can only be changed by using the options panel.

Options-panel when a line is selected now has a new dropdown for line type


a21aldsa (approved):
Works

a21aldsa (changes requested):
You cant edit lines, for example, change from strait to segmented.

d21matbo:
The newest commit should solve the problem that got merged from stable

a21aldsa (approved):

a21aldsa:
Works fine now

PR #13619, opened 2023-04-27
G3 2023 v4 issue#13447

e21krida:
What has been done is a general cleanup of the code within FetchGithubRepo.php. 

1) The individual parts like writing to databases have been separated out into it's own functions
2) The code that handles the downloading and writing of files and directories onto the web server has been separated out into it's own function
3) All styling has been removed.
4) Some parts of the code has been simplified to potentially save a few milliseconds, as the script takes it sweet time currently.

This means that instead of having one massive "bfs" function, we now have the functions "insertToFileLink", "insertToMetaData" & "downloadToWebServer". 

To test this, create a new course with a GitHub repository, preferably a smaller one and afterwards analyze that the information was downloaded properly into the "courses" directory and that all information was properly fed into the MySQL database "fileLink" and the SQLite database "metadata2.db" which is found in the main folder after choosing the student folder.

### DIFF
--- a/recursivetesting/FetchGithubRepo.php
+++ b/recursivetesting/FetchGithubRepo.php
@@ -9,23 +9,26 @@ if(isset($_POST['action']))
 {
     if($_POST['action'] == 'getNewCourseGitHub') 
     {
-        $url = ($_POST['githubURL']);
-        // Dismantles the $url into an array of each word, separated by a slash
-        $urlParts = explode('/', $url);
-        // In normal GitHub Repo URL:s, the username is the third object separated by a slash
-        $username = $urlParts[3];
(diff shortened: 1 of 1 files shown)


e21krida:
> Should the information that was downloaded end up in the "courses" directory or in "courses/1/Github" directory?

Yes! the "1" will represent course id, it's currently hard coded as feeding the course id into the script is currently not an option. But as far as i know, this will represent the final structure on the web server. It wasn't my implementation, but @a21olljo who implemented that functionality.

g21emmka (approved):
When creating a new course the files from the github url can be found under "courses/1/Github", all information can also be found in MySQL database table "filelink" and the SQLite database.

So it looks good!

c21rebja (approved):
Looks good!

Issue #13613, opened 2023-04-27
Superstate button needs functionality


b21oscpe:
The superstate button is the state with two states inside of it

![cf720433eeb9eba172bc07b7d6828c53](https://user-images.githubusercontent.com/102578165/234825413-9a99651f-ce2e-4ef2-babc-aed30f6f4fbc.png)

It should place a white box that holds an editable text at the top, keep it according to the notation #13535

PR #13620, opened 2023-04-27
G1 2023 v4 #13613

b21matfi:
I added functionality to the button, created an svg for the SuperState. This can be changed.


b21matfi:
To be clear, you can place it and change the name, you can not do more then that

d21matbo (changes requested):
image

The image above illustrates that the newly added superstate does not work well when we zoom.

Another issue I found was that the superstate currently uses the same collision detection as the rest of the elements, this makes it impossible to place any element inside the superstate (which I believe we want to be able to do)

b21matfi:
I have fixed so that you can zoom in and out, but the rest should be done in another issue. For example to be able to put other elements inside it

d21matbo (approved):
The newly added changes fixes the issue with the zoom, great job!

PR #13623, opened 2023-04-27
Line always connects to arrow for SD states

b21eddka:
When adding lines between SD states the line will always connect to the arrow.


b21matfi (approved):
Looks good! good job!

PR #13628, opened 2023-04-27
G3 2023 v4 issue#13573

a21olljo:
Added file services to files from Github, this PR involves checking that the files can be accessed wherever they are used, and verify that nothing has been missed or done incorrectly. 

issue#13573

### DIFF
--- a/DuggaSys/fileedservice.php
+++ b/DuggaSys/fileedservice.php
@@ -123,7 +123,6 @@ if (checklogin() && $hasAccess) {
 				if ($kind == 2) {
 					$currcwd .= "/courses/global/" . $filename;
 				} else if ($kind == 3) {
-                    $debug = "PATH: ". $path;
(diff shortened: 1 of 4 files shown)


a21rammo (commented):
```php  } else if ($kind == 3) {
            $currcwd .= "/courses/" . $cid . "/" . $filename;
            if ($path == null)
                $currcwd .= "/courses/" . $cid . "/" . $filename;
            else 
                $currcwd .= "/courses/" . $cid . "/Github/" . $path;
```

You added an extra things to handle when path is null. The difference is that where the file saves, if path null it saves to courses/cid/filename otherwise its courses/cid/Github/path. I think you did this to handle a wider range of scenarios, kinda a error handler.

a21rammo:
If nothing else, issueansvarig can merge it.

PR #13630, opened 2023-04-27
G3 2023 v4 issue#13621

e21krida:
The actual error message that appears to the left of the input-field is the first child element of "messageElement", the previous developer of the error message for a course code already in use had implemented the error message as innerHTML of "messageElement", not of "messageElement.firstChild". This meant that the error message didn't disappear after the value of the input-field had been corrected and simply emptying the innerHTML would cause child nodes to be removed as well, wreaking havoc on the rest of the code. I have corrected this by applying the error message to "messageElement.firstChild". 
I have not been able to locate if this solution affects anything else after rigorous bug-testing. 

I also took the time to position the error message "more" in line with respective input-field after checking consistency of miss-alignment over a few different screens. But this will have to be done in a better fashion if it ever becomes an issue again. 

This is what it should look like now:
bugfix

And this is what it looked like before:
imagebug

To test this, create a new course with a specific course-code, then attempt to create another one with the same course-code. The error message should appear in the same fashion as the others do, for example with wrong syntax. The error message should also disappear after correct course code has written into the input-field. 

Relevant file: courseed.php

### DIFF
--- a/DuggaSys/courseed.js
+++ b/DuggaSys/courseed.js
@@ -651,6 +651,8 @@ function elementIsValid(element) {
 				console.log("debug2");
 					messageElement.innerHTML = `${element.value} is already in use. Choose another.`;
 					return false;
+				} else {
+					console.log("debug3");
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


a21hugni (approved):
Looks good!

Issue #13532, opened 2023-04-24
Create testing API


b21kurar:
Implement the functionality for basic testing.

https://github.com/HGustavs/LenaSYS/issues/13062#issuecomment-1520057987

I believe for our current testing needs we might only need the following functionality:

* `assert `- make sure that two values are equal, we will only test with JSON format for now. 
* `login(user, pwd)` - login as a user, is needed for pre-requisites of tests. Some functionality is only allowed by certain users
* `send(array, service)`- send values via POST to a service
* `readOutput()`- read the HTTP JSON response

b21kurar:
To be clear, for `readOutput()` we want these values to be extracted:
![image](https://user-images.githubusercontent.com/129070844/234219955-b993ace8-4e9f-43b6-b124-b31c6cfb1275.png)

I am not sure how this works, but all services end with a `echo json_encode($array)` that are then read by the browser in the response tab, do you know if there is any way to access this?
![image](https://user-images.githubusercontent.com/129070844/234220496-22d15b3b-8fe8-48a8-8fe6-3206536c9e54.png)

Otherwise it may be better to update the logging database and passing this JSON data to it, letting `readOutput` accessing it from there instead.

c21alest:
After some research regarding how the API will communicate the conclusion is that further actions are needed to proceed, that I can't approve of my own. My understanding is that this API is a communication between two PHP files, the service to test and the test PHP file. There are two ways of communicating between the PHP files with a REST API. The first one that doesn't require any modules are to stream context to the test PHP file and then use file_get_contents to read the response. But this is not a great way as it prints all data from the PHP test files including comments made in the code. The other way is to use php-curl, this gives the functionality of a standard HTTP request call. But php-curl is not installed on the server. So my question to you @b21kurar is, will it be possible to install php-curl?

c21alest:
The API now works to make an assert equal test but I'm not sure what the `login(user, pwd)` should do. Am I suppose to implement a way that the test logs in to LenaSYS with account and do something, if so what? Do you know the answer for this @b21kurar?

b21kurar:
Example:
```
// logging in
login("user", "pwd");

// prepare and send values to service
 $arrWithValues = // add an array of values
send($arrWithValues, "someservice.php");

$expectedOut = "{/*add many values*/}";
$actualOut = readOutput();

assert($expectedOut, $actualOut);

```

c21alest:
[...]

Issue #13062, opened 2023-04-05
Investigate a suitable testing system


b21kurar:
Related to issue #13059 
We need to decide on how a reasonable testing system will look and be used.

A suggestion could be a simple test.php that collects tests from separate files (e.g microservice1test.php, microservice2test.php ,etc...) and then outputs results showing which tests passed/failed sorted by which microservice

E.g
Microservice 1
test 1 pass
test 2 fail

Microservice 2
test 1 fail
test 2 pass

etc...

(microservices should have a name that represents their function, please don't just call them microservices1)

b21kurar:
I believe this could be a scalable solution, since new microservices could easily be added as new test files, and the main test.php can search for all test files in a test folder.

b21kurar:
Could however use some consideration on microservices calling other microservices, I expect it to be easy to add tests to the already existing microservices tests, and it should be easy to locate them based on coherent naming of microservices.

d21erisu:
I think having a specific file per microservice with the sole purpose of testing that specific service seems like a great idea. If the files have appropriate names it will be easy to see what services have tests and to modify them. Also easy to add and remove tests if necessary.

HGustavs:
What standard testing tools are there that are applicable and do these tools have standard data files for defining tests of microservices?

Perhaps we can use data format from commercial/open source testing tools, in order to be close to "real" testing environments? In that case we can at some point integrate with external testing tools. Currently the plan is to be independent lenasys testing tools, but in the future we may want to be inclusive?

d21erisu:
I'm thinking that some form of black box testing might be good as well? To see that the correct microservice is used. But at the same time we want to test that the microservice itself does what is expected.

b21kurar:
> I'm thinking that some form of black box testing might be good as well? To see that the correct microservice is used. But at the same time we want to test that the microservice itself does what is expected.

The black box testing would probably just cover the current service right? As I see it, that black box test would test multiple microservices at once. As such we need to then find a suitable way to test individual microservices. This is however just an assumption from me.

d21erisu:
Using already existing testing tools (commercial/open-source) seems like a good idea. 

The testing tools: JUnit, Selenium and TestNG are recommended so i will read more about them

d21erisu:
Also what types of tests are we interested in, unit tests seems good, treating each microservice as a unit and testing each one of them. This is white box and made by the developer. Are we more interested in making automated tests?
[...]

PR #13631, opened 2023-04-27
G2 2023 v4 issue #13532 Created test API

c21alest:
Related to #13532 

Created test.php file in /Shared folder which helps testing services.
This includes 3 tests.
* Assert test (Compare two values)
* Login test - Check if login credentials are valid
* Call service - With the help of parameters like opt some service functionality can be executed

Session.php have been modified to allow checkLogin() with POST

----------------------------------------------------------
## Example on how to run this test from another file:

include_once "../Shared/test.php";

$testData = array(
    'value1' => 'test',
    'value2' => 'test',
    'service' => 'https://cms.webug.se/root/G2/students/c21alest/LenaSYS/DuggaSys/courseedservice.php',
    'serviceData' => serialize(array( // Data that service needs to execute function
        'opt' => 'NEW',
        'username' => 'usr',
        'password' => 'pass',
    )),
    'test1Debug' => false, // If true more information of test will be displayed
    'test2Debug' => true,
    'test3Debug' => true,
);

testHandler($testData, false); // 2nd argument true = prettyprint (HTML) false = raw JSON, no debug mode is avalible when using prettyprint

----------------------------------------------------------

### To test
* I have not yet checked that a service call makes any changes to the database, just that the service marks it as a success.
* Check that different login types are working (student, teacher etc) and that a student can't perform any activities that requires a teacher or admin.
* Test that JSON format can be compared in the assert test.
* Is it clear how to use the test or is any further documentation needed?
* Session.php has been modified so further testing might be needed to make sure that this won't brake anything


b21kurar:
I think it would be more preferable to let the testdesigner decide when to call the different functions (such as `assertEqualsTest`), that way it would land more in line to the discussions made in issue #13062 . That would make the `testHandler` unnecessary as of now. 

I think that a good way of doing this is letting the test designer send an array of all relevant values to each function and then letting them decide the order of execution.

And also, future issue, adding an ignore list for the response JSON.

b21kurar:
An issue I am encountering is trying to create a new course via a test case. It doesn't seem to work. I have even added new courses by hand and can't get the test case page to show the new courses
image
image

b21kurar:
The testcase used:
image

b21kurar:
I realize that it may be because I am creating entries in my own database, while `courseedservice` only checks the master database...

b21kurar:
That means that for testing, we need a way to redirect which database is to be used

b21kurar:
Since the PDO connection is made via a global variable, maybe it could be as easy as modifying it? 
image
I believe however that this could be a separate issue

b21kurar:
Wait, nevermind, those values are set in `coursesyspw.php` and should connect to the correct database. I am unsure of how it is set up

c21alest:
https://cms.webug.se/root/G2/students/b21kurar/LenaSYS/DuggaSys/courseedservice.php returns the new courses you created

b21kurar:
> 

This is my result when viewing that page: 
image
I search for `test-` but no matches are found. 

Can't find it in the response tab either:
image

They are however existing in the database:
image

c21alest:
> This is my result when viewing that page: image I search for `test-` but no matches are found.
> 
> Can't find it in the response tab either: image
> 
> They are however existing in the database: image

Created and Activeversion is null, can this be the problem? I'll try to reset my LenaSys and create my own courses.

b21kurar:
> > This is my result when viewing that page: image I search for `test-` but no matches are found.
> > Can't find it in the response tab either: image
> > They are however existing in the database: image
> 
> Created and Activeversion is null, can this be the problem? I'll try to reset my LenaSys and create my own courses.
They are now available in the service, the session might have logged me out without me knowing. However still not available in the test case. Will be trying some stuff!
image

Issue #13627, opened 2023-04-27
Unable to place other elements within the superstate


b21oscpe:
The superstate is implemented but it is not completely finished, for example it is not possible to place states within the superstate, other functionalities such as connecting a state entity to the superstate might be absent aswell. Investigate what other functionalities are missing and report them back as a comment.

a21aldsa:
Coloring a superstate does nothing. 

![image](https://user-images.githubusercontent.com/102581306/235106394-9282737f-a354-4466-af93-8253b950069f.png)

PR #13634, opened 2023-04-28
 Unable to place other elements within the superstate #13627 

a21aldsa:
Made it so that you can overlap any element with a superstate, thereby making it so that you can place any element withing a super state. **Note** that I did not manage to implement so that states can be spawned inside the superstate directly since the selection box got in the way and I couldn't figure out a way around this, but I did make it so that you can spawn states elsewhere, normally, and then simply drag them into the superstate.


b21matfi (approved):
You can not place as described in the comment but you can drag elements into the super state

looks good!

Issue #13450, opened 2023-04-20
Create a dropdown list to choose a folder to refresh/update/download based on the metadata from the course-GitHub link 


c21sebar:
For issue https://github.com/HGustavs/LenaSYS/issues/13431
Create a dropdown list to choose a folder to refresh/update/download based on the metadata from the course-GitHub link

d21emijo:
![image](https://user-images.githubusercontent.com/102797583/234008591-7aeca3a6-f948-4751-bffe-59635d23c129.png)
make insert before test

d21emijo:
![image](https://user-images.githubusercontent.com/102797583/234015518-809e99b1-55ef-4f36-905b-e43f53ea7d59.png)

only use the directorys like examples, from "repoFileType" in the "gitRepos" table

d21emijo:
## way to get data in to the db
![image](https://user-images.githubusercontent.com/102797583/234594184-0a6a98ad-a117-4579-a48e-d78d586af536.png)
When creating a course and using a github link,  the data will be saved on the sqlite db in the table "gitFiles"

## the data will be saved in the "gitFiles" tables
```
CREATE TABLE gitFiles (
                cid INTEGER,
                fileName VARCHAR(50),
    fileType VARCHAR(50),
                fileURL VARCHAR(255),
    downloadURL VARCHAR(255),
    fileSHA VARCHAR(255),
    filePath VARCHAR(255),
                PRIMARY KEY (cid, fileName)
        );
```
here we want to use the "fileName" to display all the diffrent options, but only show the options where "fileType" is the "dir" type,
![image](https://user-images.githubusercontent.com/102797583/234596153-5dc2f317-e3f9-44c3-82dc-6787ef568bd5.png)

d21emijo:
### TODO
+ the dropdown currently doesnt do anything and has no option value
```
                    foreach($rows as $row){
                    echo "<option value=''>" .$row['fileName']. "</option>";
                    }
```
+ The table gitFiles can't use the same names as the cid and filename together makes the primary key

d21emijo:
# Test
1. Get data into the sqliteDB by creating a course or by doing an insert
2. the gitHub moments screen will have the dropdown for that course will fetch all the directorys and print them out as option
![image](https://user-images.githubusercontent.com/102797583/235114838-36afbc52-9d3c-4735-a70f-55db31b6aa7e.png)
[...]

PR #13635, opened 2023-04-28
G3-2023-v4-13450-create-a-dropdown-list 

d21emijo:
issue #13450  Added a dropdown with options from the sqlite database for that specific course.
what has been changed and can be viewed in the https://github.com/HGustavs/LenaSYS/issues/13450#issuecomment-1527320173

### DIFF
--- a/DuggaSys/sectioned.php
+++ b/DuggaSys/sectioned.php
@@ -662,7 +662,9 @@
 		</div>
 		<div>
 			<?php
-	testingtesting();
+		
+	echo testingtesting();
+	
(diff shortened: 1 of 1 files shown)


c21sebar (approved):
I can select directories from the github repo only in the course that is selected, since the fetch github.php is hardcoded to add all files to cid 1 I tested by changing in the fetchgithub.php file to other cid and the dropdown works as intended. Only the directories from the repo in the sqlite db with the same cid as the course is able to be selected.

PR #13636, opened 2023-05-02
G2 2023 v4 issue#13629

b21hamwe:
Fixed the codeviewerService file, $info variables string


a19hakja (approved):
image

The strings added can be seen in the presentation, these were previously only "UNK" all over.

Issue #13643, opened 2023-05-02
ZOOM buttons borders are visible in both light and darkmode*


d21matbo:
The borders used to show highlight for zoom-buttons is always visible, this is more of a problem with lightmode but it can also be a problem with darkmode (if something is behind it*). 

![image](https://user-images.githubusercontent.com/102599161/235625960-f4d6140d-cca1-4c1d-a914-976e8444a9ce.png)
![image](https://user-images.githubusercontent.com/102599161/235626237-7c574e42-c6c8-4245-8574-325a7a3ed323.png)

If it is possible make the border the same colour as the rest of the button

b21eriho:
The way I interpret your message is that the background of the icons is transparent black as well as the background of the entire area is transparent black and they therefore add together, is this correct?

b21eriho:
My mistake, I've reread your initial message and I'd say remove the border and see how it looks because I see no reason for it to be there

PR #13655, opened 2023-05-02
 ZOOM buttons borders are visible in both light and darkmode* #13643 

a21aldsa:
After consulting with the group leader, as can be seen in #13643 , I removed the border.


b21matfi (approved):
Looks good! borders are removed

PR #13657, opened 2023-05-02
Remove consolelog #13657

b21matfi:
removed the console log("Moving Obejct") when there is no display that show that you are moving an object


a21aldsa (approved):
Log message is gone, good job.

Issue #13656, opened 2023-05-02
Options-pane symbol incorrect when opened by double-click


b21eriho:
When double-clicking on an entity to open the option-pane the icon is shown as this
![image](https://user-images.githubusercontent.com/129077727/235637612-872f592a-bc45-49d3-b4da-8e626b6545c9.png)

However when opening it by clicking it on the right hand side it shows this icon 
![image](https://user-images.githubusercontent.com/129077727/235637717-d49ac69c-4eef-4b4d-bc50-97775161d34c.png)

We want it to always show the second icon

PR #13661, opened 2023-05-02
Options-pane symbol incorrect when opened by double-click #13656

b21oscpe:
Changed the unicode for ">Options" once an entity or line is double-clicked


c21adano (approved):
The arrow appears correctly when double clicking an entity now.

PR #13664, opened 2023-05-02
G3 2023 v4 issue#13453

a21henhe:
Clicking refreshbutton on a moment now runs the method "refreshCodeExample" on each code example within a certain moment.

This might be troublesome to test since moments are very bugged. Sometimes the codeexamples "moment" does not update properly and you might have to drag around the moments a bit and update them to apply the moments id to the code examples. You can always console.log the object "retdata" right within inspector tools to be able to see the status of all code examples within entries (see picture below)

image

Co-authored by b21leowa


a21lukpo (approved):
Works correctly according to me and @a21hugni

Issue #13649, opened 2023-05-02
It is possible to draw SD-lines to an initial state


b21eriho:
As of now it's possible to draw a line from a state to an initial state, this should not be allowed
![image](https://user-images.githubusercontent.com/129077727/235634672-b55fcccb-c044-4085-82c9-7506e774bcfc.png)

b21phiro:
To clarify: Lines can and should go from initial-state to state. Not vice-versa.

PR #13665, opened 2023-05-02
G1 2023 v4 #13649

b21matfi:
Added an if statement to check if the line is drawn to the UML Initial State


a21aldsa (approved):
I would personally remove the log message before merging but the issue is resolved, looks good!

b21phiro (changes requested):
Yes I agree, please remove the console.log at line "7482".

b21matfi:
Fixed, the log message is removed

Issue #13663, opened 2023-05-02
Fix visual bug induced in #13643


b21eriho:
When solving issue #13643 the border around the zoom-buttons was removed causing the buttons to now shift slightly to the bottom right when hovered. This should not happen. 

May be solvable by reverting the changes made in #13643 and instead changing the border color to be entirely transparent.

PR #13666, opened 2023-05-02
 Fix visual bug induced in #13643 #13663 

a21aldsa:
added a fully transparent border


b21matfi (approved):
Nice the icons are back in the right place, and the border is transparent

good job!

Issue #13633, opened 2023-04-28
Setting the course id (cid) after the course 


a21olljo:
Right now course id 1 is hardcoded into the fetchgithubrepo.php:s bfs function, this variable should be retrieved from the course that the github repo belongs to. 

![Screenshot 2023-04-28 110126](https://user-images.githubusercontent.com/102599614/235105151-b2facb0f-6acb-47d2-82fc-aa74d411edd3.png)

PR #13667, opened 2023-05-02
G3 2023 v4 issue#13633 setting the course id cid after the course

c21rebja:
When creating a new course, the files for that course are saved in a folder in courses. The name of the folder should be the ID of the course (courses/"course id"/Github/"files").

For example:
image

Test this by creating a new course and checking whether the correct files have appeared in the correct folder. To double check, you can test to see if the folder name matches the course id in the MySQL database.

co-author @g21emmka

### DIFF
--- a/recursivetesting/FetchGithubRepo.php
+++ b/recursivetesting/FetchGithubRepo.php
@@ -4,8 +4,6 @@ include_once "../Shared/basic.php";
 session_start();
 pdoConnect(); // Connect to database and start session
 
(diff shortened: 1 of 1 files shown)


c21rebja:
Because of the order the functions are called in, this can't be properly tested. The code works as long as there is an available course id. We will close this pull req for now and use this function at a later point. We will then refer back to issue #13633.  

Co-author @g21emmka

PR #13669, opened 2023-05-02
 changed diagram.js so that arrow labels are printed

a21nibah:
the labels should now be correctly displayed.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -4080,14 +4080,7 @@ const stateLinesLabels=[];
 
         // Add any connected entity to the output string, and if it has not been "seen" it is added to the queue.
         for (let i = 0; i < connections.length; i++) {
-        if(i<stateLinesLabels.length)
-        {
-        output += `<p>"${head[ENTITY].name}" goes to "${connections[i][ENTITY].name} ${stateLinesLabels[i]}"</p>`;
-            }
-        else
-        {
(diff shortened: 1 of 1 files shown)


a21aldsa (changes requested):
For any transition, it will just output ("undefined"). Perhaps this is because the if and else statements you added are missing their brackets? On that note, it is also does not have correct indentation.

a21nibah:
Ok. So there is some strange syncing issues going on.  The way that the output should be displayed with labels according to the group leader will not work. The problem is that something is happening where the output is printed out Before all lines are actually added I Believe. It is kind of hard to pin down the issue. Could we have the labels as a separate thing instead? First we list the states and then we list the labels? Or some other solution because right now i am not sure where the delay is going on because when i try to loop through the list the labels are showing up correctly. Just that loop is something strange where the output is printed.

a21nibah:
Update. I Believe i have tracked down the issue. I Think  what is happening is the length of Connections is less than the labels and that is what is making the loop not work. Still. As the code is written now there is no way to really write out the label for each state like the group leader wants. The quickest way to implement it is just to list the labels after the states.

b21eriho:
In that case list the labels after the states for now

a21nibah:
The labels are now listed correctly.

a21aldsa (approved):
works as described in the comments on this PR. good job.

PR #13673, opened 2023-05-02
G2 2023 v4 issue#13396 Added courseedservice-tests.txt

a21oscgu:
Added test for courseedservice.php

### DIFF
--- a/backend-models/service-tests/courseedservice-tests.txt
+++ b/backend-models/service-tests/courseedservice-tests.txt
@@ -1,0 +1,199 @@
+
+courseedservice.php
+
+---------------------------------------------------------------------------------------------------------------------
+- New Course (line 79)
+---------------------------------------------------------------------------------------------------------------------
+
+*********
+* pre-req:
+********* 
+
+{(checklogin) && ($ha)} IS TRUE
+
+*********
+* send parameters inputs to courseedservice.php 
+*********
+send
+{
+    $userid
+    $coursecode
+    $coursename
+    $courseGitURL
+    $description
+}
+
+*********
+* Gather service output:
+*********
+Save the values of the echoed array as a JSON (at the end of the file. echo json_encode($array);)
+this is the expected output for the micro service.
+
+---------------------------------------------------------------------------------------------------------------------
+- New Vers (line 116)
+---------------------------------------------------------------------------------------------------------------------
+
+*********
+* pre-req:
+********* 
+
+{(checklogin) && ($ha)} IS TRUE
+
+*********
+* send parameters inputs to courseedservice.php 
+*********
+send
+{
+    $cid
+    $coursecode
+    $versid
+    $versname
+    $coursename
+    $coursenamealt
+    $motd
+    $startdate
+    $enddate
+}
+
+*********
+* Gather service output:
+*********
+Save the values of the echoed array as a JSON (at the end of the file. echo json_encode($array);)
+this is the expected output for the micro service.
+
+---------------------------------------------------------------------------------------------------------------------
+- Update vers (line 152)
+---------------------------------------------------------------------------------------------------------------------
+
+*********
+* pre-req:
+********* 
+
+{(checklogin) && ($ha)} IS TRUE
+
+*********
+* send parameters inputs to courseedservice.php 
+*********
+send
+{
+    $courseid
+    $coursecode
+    $versid
+    $versname
+}
+
+*********
+* Gather service output:
+*********
+Save the values of the echoed array as a JSON (at the end of the file. echo json_encode($array);)
+this is the expected output for the micro service.
+
+---------------------------------------------------------------------------------------------------------------------
+- Update course (line 173)
+---------------------------------------------------------------------------------------------------------------------
+
+*********
+* pre-req:
+********* 
+
+{(checklogin) && ($ha)} IS TRUE
+
+*********
+* send parameters inputs to courseedservice.php 
+*********
+send
+{
+    $courseid
+    $versid
+}
+
+*********
+* Gather service output:
+*********
+Save the values of the echoed array as a JSON (at the end of the file. echo json_encode($array);)
+this is the expected output for the micro service.
+
+---------------------------------------------------------------------------------------------------------------------
+- Insert vers (line 181)
+---------------------------------------------------------------------------------------------------------------------
+
+*********
+* pre-req:
+********* 
+
+{(checklogin) && ($ha)} IS TRUE
+
+*********
+* send parameters inputs to courseedservice.php 
+*********
+send
+{
+    $cid
+    $coursecode
+    $versid
+    $versname
+    $coursename
+    $coursenamealt
+    $motd
+    $startdate
+    $enddate
+}
+
+*********
+* Gather service output:
+*********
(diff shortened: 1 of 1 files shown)


backend-models/service-tests/courseedservice-tests.txt
```diff
@@ -0,0 +1,258 @@
...
+
+{(checklogin) && ($ha)} IS TRUE
+
+*********
+* send parameters inputs to courseedservice.php 
+*********
+send
+{
+    $opt = "NEW";
+    $userid = "UNK"; or $userid = !NULL;
```
a21jeaha, line 20: you need to have real values for each variable, it has to be something that can be entered by the person writing the test program. That person should not have to figure out what value CAN be entered, that is the whole point of this document.

a21jeaha: this of course does not include the last variable, i know that in that case no values are presented at the time of writing this

backend-models/service-tests/courseedservice-tests.txt
```diff
@@ -0,0 +1,258 @@
...
+    $coursecode = "G1337";
+    $coursename = "Testing-Course";
+    $courseGitURL = NULL; // There are no values in the database
```
a21jeaha, line 23: i had problems doing inserts into the table with the INSERT string found in the service file. 
might have to do with the fact that the names of the colums are different. 
image
image

there is of coarse nothing you can do about it (or rather should not do), but it can be a good idea to document it.

backend-models/service-tests/courseedservice-tests.txt
```diff
@@ -0,0 +1,258 @@
...
+{(checklogin) && ($ha)} IS TRUE
+
+*********
+* send parameters inputs to courseedservice.php 
+*********
+send
+{
+    $opt = "NEW";
+    $userid = "UNK"; or $userid = !NULL;
+    $coursecode = "G1337";
```
a21jeaha, line 21: This value cannot be used. I receive this error when i try to make an insert.
image

The entered value has to be unique in the table "course", this is indicated by this line in mySQL.
image
In other words you need to make sure that the value does not already exist in the table.

backend-models/service-tests/courseedservice-tests.txt
```diff
@@ -0,0 +1,258 @@
...
+*********
+* send parameters inputs to courseedservice.php 
+*********
+send
+{
+    $opt = "NEWVRS";
+    $cid = 1885;
+    $coursecode = "G1337";
+    $versid = 1337;
+    $versname = NULL; // There are no value in the database
```
a21jeaha, line 51: it has to have a value
image

backend-models/service-tests/courseedservice-tests.txt
```diff
@@ -0,0 +1,258 @@
...
+
+{(checklogin) && ($ha)} IS TRUE
+
+*********
+* send parameters inputs to courseedservice.php 
+*********
+send
+{
+    $opt = "NEWVRS";
+    $cid = 1885;
```
a21jeaha, line 48: image
image
image

backend-models/service-tests/courseedservice-tests.txt
```diff
@@ -0,0 +1,258 @@
...
+    $versid = 1337;
+}
+
+*********
+* Gather service output:
+*********
+Save the values of the echoed array as a JSON (at the end of the file. echo json_encode($array);)
+this is the expected output for the micro service.
+
```
a21jeaha, line 90: the insert that was created for this test should at the end of the test have been made into the active course version. the active version needs to be changed back to its original value, and the insert conducted needs to be removed.

backend-models/service-tests/courseedservice-tests.txt
```diff
@@ -0,0 +1,258 @@
...
+*********
+* send parameters inputs to courseedservice.php 
+*********
+send
+{
+    $opt = "UPDATEVRS";
+    $courseid = 1885;
+    $coursecode = "G1337";
+    $versid = 1337;
+    $versname = NULL; // There are no value in the database
```
a21jeaha, line 110: Cannot have the value NULL 
image
image

backend-models/service-tests/courseedservice-tests.txt
```diff
@@ -0,0 +1,258 @@
...
+}
+
+*********
+* Gather service output:
+*********
+Save the values of the echoed array as a JSON (at the end of the file. echo json_encode($array);)
+this is the expected output for the micro service.
+
```
a21jeaha, line 118: don't forget to revert to the old version name

backend-models/service-tests/courseedservice-tests.txt
```diff
@@ -0,0 +1,258 @@
...
+---------------------------------------------------------------------------------------------------------------------
+- Update vers (line 152)
```
a21jeaha, line 92: a new version of this test is needed as well, the other version is to handle when 
image

backend-models/service-tests/courseedservice-tests.txt
```diff
@@ -0,0 +1,258 @@
...
+    $courseid = 1885;
+    $versid = 1337;
+}
+
+*********
+* Gather service output:
+*********
+Save the values of the echoed array as a JSON (at the end of the file. echo json_encode($array);)
+this is the expected output for the micro service.
+
```
a21jeaha, line 144: dont forget to revert values

backend-models/service-tests/courseedservice-tests.txt
```diff
@@ -0,0 +1,258 @@
...
+
```
a21jeaha, line 204: dont forget to revert the database values

backend-models/service-tests/courseedservice-tests.txt
```diff
@@ -0,0 +1,258 @@
...
+* send parameters inputs to courseedservice.php 
+*********
+send
+{
+    $opt = "UPDATE";
+    $cid = 1885;
+    $coursename = "Testing-Course";
+    $visibility = 1;
+    $coursecode = "G1337";
+    $courseGitURL = NULL; // There are no value in the database
```
a21jeaha, line 225: cannot have the value NULL.
image
image

backend-models/service-tests/courseedservice-tests.txt
```diff
@@ -0,0 +1,258 @@
...
+}
+
+*********
+* Gather service output:
+*********
+Save the values of the echoed array as a JSON (at the end of the file. echo json_encode($array);)
+this is the expected output for the micro service.
+
```
a21jeaha, line 233: don't forget to revert the database to its original state

backend-models/service-tests/courseedservice-tests.txt
```diff
@@ -0,0 +1,258 @@
...
+* pre-req:
+********* 
+
+{(checklogin) && ($ha)} IS TRUE
+
+*********
+* send parameters inputs to courseedservice.php 
+*********
+send
+{
```
a21jeaha, line 248: values?

I am well aware that the database is empty. but try to supply real values.

these should be enough.
 $motd = "this is a test message of the day";
 $readonly = 0;

Issue #13549, opened 2023-04-24
Code examples should automatically be created and named after the files in the folder


c21sebar:
Code examples should automatically be created and named after the files in the folder

c21sebar:
After https://github.com/HGustavs/LenaSYS/pull/14023 was merged we had a meeting where we learned that the files for each moment is not stored in a folder anymore but all files for all code-examples are store at the same place. Meaning our solution doesnt work and we have to redo the function to suit the https://github.com/LenaSYS/Webbprogrammering-Examples/tree/master/Examples%20Html moment. We were told we are not allowed to rename the files or store the files in separate folders. This repo/link is not to be altered with and is how it should look. This link contains all the files for the "Examples html" moment and we now need to sort out wich files are related by looking at the names of the files and run the function that create code-examples on these files.

PR #13674, opened 2023-05-02
G3 2023 v4 issue#13549

b21sebgr:
https://github.com/HGustavs/LenaSYS/issues/13549


a21lukpo (approved):
Looks like it could work in the future, also shouldn't break anything at this point so it's ok!

Issue #13640, opened 2023-05-02
Lines cannot be deleted while "Line tool" is active


d21matbo:
When a line is selected on the canvas it is not possible to delete it while the line creation tool is active

![image](https://user-images.githubusercontent.com/102599161/235623799-51a5ed12-ee9c-4c15-8918-6a2f19021707.png)

For context:
Changing the active tool to either "Pointer" or "Box Select" while the line is still selected, the line can be deleted.

PR #13685, opened 2023-05-03
G1 2023 v4 #13640

d21matbo:
Added functionallity to remove lines while the Line creation tool is active,

It should now be possible to remove the lines with the red 'X' on the screen or 'Delete' & 'Backspace' buttons on your keyboard


a21oligu (approved):
Works as expected, good job!

Issue #13641, opened 2023-05-02
Tooltip for ZOOM buttons is hidden behind the ruler


d21matbo:
The tooltip that is generated while hovering over the zoom-buttons is somewhat hidden behind the ruler

![image](https://user-images.githubusercontent.com/102599161/235625355-1dceaed7-1b20-4edd-bada-967f80adc293.png)

PR #13686, opened 2023-05-03
Tooltip partially hidden #13641

a21oligu:
Tooltips for zoom are no longer partially hidden underneath the ruler. Solves #13641


b21eddka (approved):
After fixing minor issue everything works fine. Approved.

PR #13699, opened 2023-05-03
Added missing content to codeviewerService.md

c21axepe:
Added the missing content as requested!

### DIFF
--- a/backend-models/functional-requirements/FRD_codeviewerService.md
+++ b/backend-models/functional-requirements/FRD_codeviewerService.md
@@ -6,30 +6,30 @@
 
 | FN No. | Test Proposal: |
 | --- | --- |
-| FR1 |  |
-| FR2 |  |
-| FR3 |  |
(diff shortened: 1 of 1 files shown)


backend-models/functional-requirements/FRD_codeviewerService.md
b21hamwe, line 1: Smaller fix for Test proposal, FR1 : Isn't really a test in my eyes, add log and compare with database in order too see that the data is correct for a teacher. 
This test right now only checks if a teacher can get access to the codeexamples, but it doesn't look at what happens behind the curtains.

Also, FR2 is missing some information. "Where the user has access (writeaccess = w "

b21hamwe (changes requested):

backend-models/functional-requirements/FRD_codeviewerService.md
c21axepe: 1. We added more information about what happens "behind the curtains"
2. We wrote || instead of or thats why information was cut

b21hamwe (approved):
Seal of approval, good job!

Issue #13682, opened 2023-05-03
Prevent lines going FROM UML-final state


b21phiro:
This is bad. Remove this "feature" please.

![bild](https://user-images.githubusercontent.com/45258330/235892862-edb2b2eb-19bc-4d36-a126-cb9ef54afc5b.png)

PR #13707, opened 2023-05-03
G1 2023 v4 #13682

b21matfi:
created so that you can not draw lines from UML Final State


b21matfi:
will close the PR because i used old version and redo

PR #13710, opened 2023-05-03
 changed the output in diagram.js so that when a line label is printe…

a21nibah:
seems some other changes were commited too. seems to be changes from the latest g1-2023-v5 but there is nothing that should impact. anyway the output should now be correct.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -903,8 +903,7 @@ const relationState = {
 const lineKind = {
     NORMAL: "Normal",
     DOUBLE: "Double",
-    DASHED: "Dashed",
-    RECURSIVE: "Recursive"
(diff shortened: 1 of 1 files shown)


a21oligu:
Seems like some code that was removed in a previous merge is now being added back. Could you pull the current g1-2023-v5 to your branch to make sure it is up to date, and that the merge wont override current group branch?

b21eriho:
Yeah there is something weird going on, on line 2481 you have added back a debug-message for when an object is moved that was removed in another issue according to git

a21nibah:
Do you have any suggestions for how to fix? Maybe something is going on with v5. Everything is up to date so and branch was created from v5 so there must be someone else who merged something back in to v5.

Issue #13680, opened 2023-05-03
Toggle load button as well as dropdown


b21eriho:
When example diagrams are to be loaded the dropdown is toggled but the "load"-button is always there, make it be toggled along with the dropdown. 
![image](https://user-images.githubusercontent.com/129077727/235888713-1cfbb9fd-bda8-43df-a8da-efbd852d16f4.png)

b21oscpe:
The issue is fixed and I've taken some freedom in styling the button so that it follows the general design decisions made for buttons of similar kind within the options-pane. I was wondering if I am allowed to extend my issue to make the "Null" option clean the board? And perhaps also rename it from "Null"...

![964e174d1800b0d3d8f07e8053e44027](https://user-images.githubusercontent.com/102578165/235932552-c6ebd9bc-80b2-43b7-a127-da43ec148a13.png)

PR #13712, opened 2023-05-03
Toggle load button as well as dropdown #13680

b21oscpe:
The "load" button is now toggled instead of fixed. It has also been styled according to the design of the options-pane

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -123,6 +123,11 @@
     display: none;
 }
 
+#diagramTypeDropdown{
+    width: 100%;
+    margin-bottom: 10px;
+}
+
 .hide {
     display: none;
 }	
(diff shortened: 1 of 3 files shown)


a21nibah (changes requested):
can you please change this so it uses display instead of the way you are doing it now? something similar to the dropdown button. The button hides for a normal user but it is still there for a screen reader. Or just set hidden=true which would work as well and would be less code to write I think. But display set to None I believe is better and is how elements are hidden in modern web development.

b21eriho:
Test out using hidden=true, better availability would be preferred

b21oscpe:
b7a3de2f389fcbadf16fe3a8f37ca61c
Found this after a quick search, but I will look into using "hidden=true", although I think it this solution might end up weird visually

b21eriho:
If it doesn't work visually we'll have to reassess

Issue #13711, opened 2023-05-03
Check if TODO is fullfilled for the function "updateSelection"


b21phiro:
```
/**
 * @description Updates the current selection of elements depending on what buttons are down. Context array may have the new element added or removed from the context array, have the context array replaced with only the new element or simply have the array emptied.
 * @param {Object} ctxelement Element that has was clicked or null. A null value will DESELECT all elements, emptying the entire context array.
 */
function updateSelection(ctxelement) **// TODO : Default null value since we use it for deselection?**
{
(...)
}
```

There's a old TODO in the code. Check if the default value checks out with the todo. 

If default value is NULL, remove TODO comment.

If not, make sure to add default NULL values from those functions who uses the ```updateSelection``` function.

b21matfi:
When selecting an element it shows the element when not selecting anything it comes out as null 
![image](https://user-images.githubusercontent.com/102650424/235935424-3c5568f1-6f9b-4079-a138-5625fe4908af.png)

PR #13716, opened 2023-05-03
G1 2023 v5 #13711

b21matfi:
checked if the default is null with console log and then after i removed the TODO

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -3092,7 +3092,6 @@ function updateSelectedLine(selectedLine)
  */
 function updateSelection(ctxelement)
 {
-    console.log(ctxelement);
     // If CTRL is pressed and an element is selected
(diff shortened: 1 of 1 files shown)


b21oscpe (approved):
There is a call of this method in the code that sends no parameter, thought it was worth to mention though im not sure if its relevant, other than that it looks fine

`function checkDeleteBtn(){
    if (deleteBtnX != 0) {
        if (lastMousePos.x > deleteBtnX && lastMousePos.x < (deleteBtnX + deleteBtnSize) && lastMousePos.y > deleteBtnY && lastMousePos.y < (deleteBtnY + deleteBtnSize)) {
            if (context.length > 0) {
                removeElements(context);
            }
            if (contextLine.length > 0) {
                 removeLines(contextLine);
            }            
            
            updateSelection();
            return true
        }
        return false
    }
    return false
}`

PR #13717, opened 2023-05-03
added test proposals to courseedservice

b21antbr:
Co-authored-by: c21axepe c21axepe@student.his.se

### DIFF
--- a/backend-models/functional-requirements/FRD_courseedservice_functionalrequirments.md
+++ b/backend-models/functional-requirements/FRD_courseedservice_functionalrequirments.md
@@ -7,19 +7,19 @@
  | FR5 | Copy a course version |
  | FR6 | Update name and or visibility of course |
  | FR7 | Set the message of the day |
- | FR8 | Set courseed to read only |
+ | FR8 | Set course to read only |
  | FR9 | Retrive course information |
  
  | FN No: | Test Proposal: |
  | --- | --- |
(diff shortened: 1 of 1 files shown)


a19hakja:
I'm a bit confused. Not saying anything is wrong, I'm just wondering if we really need this? Since we're already writing test descriptions with proper values for each file, before we even do the real tests. And if we do in fact need this and this is what was asked for, it isn't something I would look for in a file that has to do with functional requirements. you feel me?

I might not be in charge anymore so it isn't really up to me. I'll check in with management to make sure that this is what they want, aswell as hold out for an answer from you before

d21erisu:
> I'm a bit confused. Not saying anything is wrong, I'm just wondering if we really need this? Since we're already writing test descriptions with proper values for each file, before we even do the real tests. And if we do in fact need this and this is what was asked for, it isn't something I would look for in a file that has to do with functional requirements. you feel me?
> 
> I might not be in charge anymore so it isn't really up to me. I'll check in with management to make sure that this is what they want, aswell as hold out for an answer from you before

It's not wrong, just too close to reality. We want as fuzzy abstract proposals as possible if that makes sense. Just a clue on how to test it, not an actual way to test.

a19hakja (changes requested):
Yeah i felt it being a bit fuzzy too, try to aim for something more like this https://github.com/HGustavs/LenaSYS/blob/master/backend-models/functional-requirements/FRD_Fileedservice.md

a19hakja (approved):
aight this looks good, i'll approve

Issue #13702, opened 2023-05-03
Fix keybinding functionality for UML-final state


b21phiro:
![bild](https://user-images.githubusercontent.com/45258330/235922740-23240628-9f48-4c3c-a6dd-423b36b68619.png)

Key is assigned to final state elements, but does not work. Add functionality so you can place elements using key-bind. Change key if needed, but remember to update the tooltip if key is changed.

PR #13720, opened 2023-05-03
Fix keybinding functionality for UML-final state #13702

b21eddka:
Fixed typo which made certain elements' mouse mode undefined. It is now possible to place elements with thier corresponding keybind. Changed keybind to UML-final state from ctrl + < to f.


a21oligu (approved):
Looks good

Issue #13687, opened 2023-05-03
Add more descriptive text for tooltips


b21phiro:
Right now the descriptions for tooltips are very minimal. For example. Add more descriptive text for each tooltip. Use text from wikipedia or whatever. For example, for ER Entity:

```An entity–relationship model (or ER model) describes interrelated things of interest in a specific domain of knowledge. A basic ER model is composed of```.

This image as a suggestion.
![bild](https://user-images.githubusercontent.com/45258330/235896782-f6e0ecf4-bfd7-47fe-9c2f-6762920fafa6.png)

For those elements that are too hard to describe, skip them. Make sure the text is not too long.

My suggestion to solve is:
- ```<p>``` Tags in diagram.php file inside each ```toolTipText``` elements.
- Write the CSS in diagram CSS, PLEASE do not use inline!
- The class for given text should be descriptvie, do not call it for example ".a".

b21phiro:
As much detailed descriptions as you can would be prefered. However i do understand that some can be more complex. The idea is just that the user get a General deacription of what for example and UML Diagram is, rather then what it does.

Not standardize, mostly unique. So hardcode dercriptions in to the html i suggest

c21adano:
As an example:
State diagram state
A state diagram state is a representation of a status a process can have.
Each state represents a unique status that a process can have.

IE entity
Each entity represents an object along with its attributes.
Each entity is represented by a table with a field that shows attributes.

ER entity
Each entity represents an object which is a representation of concepts or data.
The entity only holds the name of the object and if it depends on another object.

UML Class entity
Each class entity represents its own class along with the attributes and operations held within the class.

Would these description be good enough or should they be more descriptive?

c21adano:
Nevermind it seems they are all linked and when one is selected they all give the selected message.

c21adano:
There are none element function that have tool tips, should i add more description to those as well? like er-table and testcase.

b21phiro:
If you feel like it, yes.

PR #13721, opened 2023-05-03
G1 2023 v5 #13687

c21adano:
Added descriptions to diffrent element and functions.

Added descriptions to:
All entity types,
ER relation,
UML inheritance,
IE inheritance,
ER attribute,
Inital state,
Final state,
Lifeline,
SuperState,
Sequence object,
Sequance activation,
Error toggle

The descriptions briefly the function of each element/function. Note that the pop out boxes have diffrent descriptions based on which element is selected.

### DIFF
--- a/DuggaSys/diagram.php
+++ b/DuggaSys/diagram.php
@@ -287,25 +287,30 @@
                             <div class="ERButton placementTypeBoxIcons" onclick='togglePlacementType(0,0); setElementPlacementType(0); setMouseMode(2);'>
                                 <img src="../Shared/icons/diagram_entity.svg" alt="ER entity"/>
                                 <span class="placementTypeToolTipText"><b>ER entity</b><br>
-                                    <p>Change to ER entity is selected</p>
+                                    <p>Change to ER entity</p>
+                                    <p>Each entity represents an object which is a representation of concepts or data.</p>
+                                    <p>The entity only holds the name of the object and if it depends on another object.</p>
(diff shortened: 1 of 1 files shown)


a19clasv (changes requested):
Sequence activation could use a better description. Here's my suggestion. "Represents that an object is active (or instantiated) during an interaction between two objects, with the length indication the duration.". There are some grammatical errors as well. Here are suggestions for how to fix them.
ER relation: "Add an ER relation", not "Add a ER relation". 
Attribute: "Each attribute represents different characteristics of an entity", not "Each attribute represents diffrent characteristics of an entity". Attribute should probably be named ER attribute as well, but that's not part this issue.
UML inital state: "The initial state represents the start of a process.", not "The initial state represent the start of a process.".
UML final state: "The final state represents where a process ends.", not The final state represent where a process ends.".
UML super state: "A state that can contain substates.", not "A state that can conatins substates."
Be consistent with punctuation, eg. add a period in "Each attribute represents diffrent characteristics of an entity" or change the other ones.

a19clasv (changes requested):
UML super state: "A state that can contain substates.", not "A state that can conatin substates."
Sequence activation: "Represents that an object is active during an interaction, with the length indicating the duration.", not "Represents that an object is active during an interaction, with the length indication the duration.". That one's on me, my suggestion was wrong.

a19clasv (approved):

PR #13726, opened 2023-05-04
G2 2023 v5 issue#13434 - Transition log.php to a microservice architecture

### LINKED ISSUE #13434: Transition log.php to a microservice architecture
This is a service in the system, it should follow the same standard, but still be written in a microservice architecture.

a21jeaha:
I have changed the name of the file so that it follows the standard. (logSearch.js is now named log.js).
And added a document placed here.
/shared/service documentation/Communication used in the old style service files.pdf
That describes my view of how data is sent between the three file service structure (.php .js service.php) 

close #13434


c21alest (approved):
Nice work!

Issue #13722, opened 2023-05-03
Remove unused function "diagramToScreenPosition"


b21phiro:
This function is never used yet there is a duplicate function with same name. Either remove one or both. Both if nothing breaks.

```
/**
 * @description Converts a coordinate on the canvas into a pixel position on the screen.
 * @param {Number} coordX Coordinate position in the x-axis.
 * @param {Number} coordY Coordinate position in the y-axis.
 * @returns {Point} Point containing the calculated screen position.
 * @depricated TODO : Needs to be updated
 * @see screenToDiagramCoordinates() For converting the other way.
 */
function diagramToScreenPosition(coordX, coordY)
{
    console.warn("diagramToScreenPosition() is depricated. It should be updated to use new screenToDiagramCoordinates() algorithm reversed.");
    return new Point(
        Math.round((coordX + scrollx) / zoomfact + 0),
        Math.round((coordY + scrolly) / zoomfact + 0)
    );
}
```

```
/**
 * @description Converst a position in screen pixels into coordinates of the array.
 * @param {Number} mouseX Pixel position in the x-axis.
 * @param {Number} mouseY Pixel position in the y-axis.
 * @returns {Point} Point containing the calculated coordinates.
 * @see diagramToScreenPosition() For converting the other way.
 */
function screenToDiagramCoordinates(mouseX,mouseY)
{
    // I guess this should be something that could be calculated with an expression but after 2 days we still cannot figure it out.
    // These are the constant values that the expression should spit out anyway. If you add more zoom levels please do not come to us.
    // We're tired.

    // We found out that the relation between 0.125 -> 4 and 0.36->-64 looks like an X^2 equation.
    var zoomX = 0;

    // ZOOM IN
    if (zoomfact == 1.25) zoomX = zoom1_25;
    if (zoomfact == 1.5) zoomX = zoom1_5;
    if (zoomfact == 2) zoomX = zoom2;
    if (zoomfact == 4) zoomX = zoom4;

    // ZOOM OUT
    if (zoomfact == 0.75) zoomX = zoom0_75;
    if (zoomfact == 0.5) zoomX = zoom0_5;
    if (zoomfact == 0.25) zoomX = zoom0_25;

    return new Point(Math.round( ((mouseX - 0) / zoomfact - scrollx) + zoomX * scrollx + 2 + zoomOrigo.x), // the 2 makes mouse hover over container
[...]

PR #13739, opened 2023-05-04
G1 2023 v5 #13722

c21adano:
Removed the diagramToScreenPosition function in diagram.js due to not being used, but keept screenToDiagramCoordinates due to it being used a bit in digram.js.


b21matfi (approved):
Nice work!

PR #13740, opened 2023-05-04
G2 2023 v5 issue#13400

b21ricni:
Rewrote sectionedservice-test


a21ponkr:
@b21ricni on line 189 it says "$armin[] = 1 //Can be 0-2" should that be changed to admin or is armin correct? Otherwise I think it looks good and you've resolved everything JP requested

a21ponkr (approved):
looks good

Issue #13691, opened 2023-05-03
Continue work with recursive state-arrows


b21eriho:
Continue work on and fix the last issues with #13526 including system-break when box-selecting and arrow placement

PR #13762, opened 2023-05-04
Continue work with recursive state-arrows #13691

b21willu:
Fixed positions for the recursive line and fixed the problem with box selection. 
To add recursive relation, draw the line from an SD element and release outside the element and it should attach itself to the same element.


b21willu:
https://github.com/HGustavs/LenaSYS/issues/13691

b21oscpe (approved):
Looks good! Works as intended, however I would imagine that the selection for a line should cover the entire line (The purple box should go around the entire recursion and the X should be in the top right.)

This could be a future issue
321108222fb9ab424115ea8b43c55163

PR #13763, opened 2023-05-04
 fixed the output of diagram.js for state line labels. 

a21nibah:
fixed so that the output now looks better and unlabled lines are printed as unlabeled. Also the strange changes that were there yesterday are no longer there.


b21oscpe (approved):
Output works as described
54bc1d23671abe724b070e6f5b2b80ff

PR #13766, opened 2023-05-04
Clean function "setCursorStyles"

b21eddka:
Removed all instances of cursorStyle/setCursorStyles and replaced it with containerStyle/setContainerStyle.


b21matfi (approved):
Great Job!

PR #13767, opened 2023-05-04
Changed resetDiagram() to load in empty diagram

b21oscpe:
It is now possible to reset the diagram from the reset-button aswell as the "Empty board"-option in the dropdown. There is a lot of outcommented code in resetDiagram() and arguably it serves no purpose anymore as it just calls another function.

The state machine value outputs correctly in both cases after a reset.
e73900f284e6acffde9bab27249bd3ce


b21matfi (approved):
Great job! you can reset the diagram by using the reset button.

Issue #13737, opened 2023-05-04
Black UML initial and final states hard to see in darkmode


b21eriho:
Note: to be done after #13688 to avoid merge conflicts

If the uml initial and final states are colored black they become very had to see in darkmode. This needs to be resolved, perhaps using a white outline or in some other way.

PR #13772, opened 2023-05-04
G1 2023 v5 #13737

b21matfi:
I changed so that the states are white in darkmode and white in dark mode


a21aldsa (changes requested):
When in darkmode; this overrides the colors set by the user. I have added two screenshots of the same elements in light and darkmode.

image

image

b21matfi:
Fixed the issue where the white overrites the chosen color. You can not change to black in darkmode because the solution checks the  #color. Talked with the issue leader and that will be a issue in the future.

a21aldsa (approved):
The issues I would point out are set to become issues of their own, based on the comments in this PR. So, with that in mind, I consider this PR approved. Good job.

Issue #13684, opened 2023-05-03
Create the functionality for sequence lifeline


b21eriho:
Using the button "sequence lifeline" (actor icon) should create an entity that consists of an actor icon with a dashed line extending below it according the notation in #13535. The line length should be changeable.

PR #13773, opened 2023-05-04
 Create the functionality for sequence lifeline #13684 

a21aldsa:
Created a new svg for actor and its lifeline. Also created vertical resizing via reuse of existing code for nodes and horizontal resize.


b21oscpe (changes requested):
After placing a lifetime it's not possible to stretch it vertically or horizontally even if it's selected, you must first select the selection tool.
861c5a1a46bd467de8b36f994c4a3b71

Furthermore, in my personal opinion it is important to seperate the SVGs for Actor and Lifetime to support toggling between Actor/Object, for example within the options-pane. (I have not checked if they are seperate SVG's, just thought i should mention.)

a21aldsa:
> After placing a lifetime it's not possible to stretch it vertically or horizontally even if it's selected, you must first select the selection tool. 861c5a1a46bd467de8b36f994c4a3b71
> 
> Furthermore, in my personal opinion it is important to seperate the SVGs for Actor and Lifetime to support toggling between Actor/Object, for example within the options-pane. (I have not checked if they are seperate SVG's, just thought i should mention.)


I'm not sure what you mean by "separate", theyre separate paths, but they are not in separate svg tags, is this what you mean? Based on the issue I believed the actor and its lifeline should spawn together and move together and that sequence objects would be implemented later with their own lifeline. 

And the resizing nodes not being visible on selection is a general design choice for all elements; not something i added. I think @b21eriho should decide if this needs to be changed and if so, I believe this to be a separate issue since it affects all elements.

b21oscpe:
The idea of a toggled actor to reduce the amount of buttons is something I've discussed with @HGustavs

a21aldsa:
> What I meant is that it should be possible to toggle the Actor on/off to represent either an object or actor for the purpose of future development. I do realise that it is not part of this issue although I would like to know that this implementation supports that idea

Technically not right now but it would not be hard to implement, just group the actor body and its head, since these are separate from the actual lifeline and then create the svg for the object in another group. Then using JS you can have one of these set to not be visible and toggle that back and forth. The resizing should work just fine, as far as I can see.

a21aldsa:
You might have to rename the element though since its currently just called sequenceActor and might also have to fiddle with the height of it since sequence objects are wider than they are long

b21eriho:
If the issue that is causing this to be under "changes requested" is simply that the selection tool is needed to rescale the entity then that can be ignored as it is not an issue introduced in this issue. If there are other concerns please elaborate om them @b21oscpe

b21oscpe (approved):
Nope, in that case it looks good, it was mainly the resizing thing, the actor/object thing was mainly just to make sure

Issue #13688, opened 2023-05-03
Better "default-color" for inital-state and final-state.


b21phiro:
![bild](https://user-images.githubusercontent.com/45258330/235897913-da2570c0-0d93-476c-9ec5-adad5abe78df.png)

This blue is nice but should not be as default. Make the default color into a color that can be picked using the options screen.

b21phiro:
Change the color to one of the defaults that exist as option. If you feel like it, you can add colors. Just make sure the default colors are selectable.

PR #13775, opened 2023-05-04
Added blue color to option bar and changed initial state and final state to black color.  #13688

c21theti:


a21aldsa (approved):
Personally, I am not huge on the pure blue being selectable in the options panel. I feel like this color makes all text just harder to read, no matter the text color and also looks fairly outdated. 

With that said, the issue is definitely resolved since the states are no longer blue. Good job.

PR #13777, opened 2023-05-04
G2 2023 v5 issue#13768 Updated API test

c21alest:
After discussion with @a21jeaha, @d21erisu and @a19hakja some changes was needed to API.

* Updated assertEqual() test to compare expected-output from JSON to the actual output from callService() test.
* Added filter option to exclude JSON output from callService, used in assertEqual() (assertEqual() is not test 3)
* Fixed pretty print mode

### DIFF
--- a/Shared/test.php
+++ b/Shared/test.php
@@ -15,7 +15,7 @@ $testData = array(
         'username' => 'usr',
         'password' => 'pass',
     )),
-    'filter-output' => serialize(array( // Filter what output to use in assert test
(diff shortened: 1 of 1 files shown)


b21kurar:
Tested for NEW and UPDATE in `courseedservice.php` and the test works as intended.

However when modifying `filter-output` by removing `'motd'` it will not pass the test, this is because it will still compare for both `'debug'` and `'motd'`
image

b21kurar (approved):
After testing `courseedservice.php` with both NEW and UPDATE I approve this PR.

PR #13779, opened 2023-05-04
G3 2023 v5 issue#13764

b21sebgr:
The file should now be removed from sqlite when it is removed from the course

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -525,7 +525,6 @@ function confirmBox(operation, item = null) {
   }
   //ändra 
   else if (operation == "openGitHubTemplate") {
-    console.log("testworkornah?");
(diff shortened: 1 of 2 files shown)


c21sebar (changes requested):
Files removed from fileupload page doesnt get removed in the sqlite db. When deleting an item in listentries the files connected to the entrie doesnt get deleted and still stays in the MySQL db, SQLite db and in the webbserver. 

Maybe Im wrong here but listentires in the DB is what is shown in the coursepage, and the different code-examples, headings , their position and so on. This is in my opinion not the place to remove from the sqlite db. This deleted button simply delete the list item and not the connected files if there are any in the item.

The place where we delete files are from the fileupload page called fileeditor, the icon that looks like an "eluttag".

PR #13780, opened 2023-05-04
Fixed so that lines save properly

c21adano:
The save function for sd lines was a copy of the endIcon save, but not all endIcon mentions were replaced with innerType or lineType. innerType and lineType was also a bit mixed upp.


c21theti (approved):
Looks good. Lines now save properly

Issue #13742, opened 2023-05-04
Remove unused code in diagram.php


b21phiro:
There's an entire section in the ```diagram.php``` that is commented and should be removed. Remove it.
```<!-- <fieldset>
            <legend>Zoom</legend>
            <div class="diagramIcons" onclick='zoomin();'>
                <img src="../Shared/icons/diagram_zoomin.svg"/>
                <span class="toolTipText"><b>Zoom IN</b><br>
                    <p>Zoom in on viewed area</p><br>
                    <p id="tooltip-ZOOM_IN" class="key_tooltip">Keybinding:</p>
                </span>
            </div>
            <div class="diagramIcons" onclick='zoomout();'>
                <img src="../Shared/icons/diagram_zoomout.svg"/>
                <span class="toolTipText"><b>Zoom OUT</b><br>
                    <p>Zoom out on viewed area</p><br>
                    <p id="tooltip-ZOOM_OUT" class="key_tooltip">Keybinding:</p>
                </span>
            </div>
            <div class="diagramIcons" onclick="zoomreset()">
                <img src="../Shared/icons/diagram_zoomratio1to1.svg"/>
                <span class="toolTipText"><b>Zoom RESET</b><br>
                    <p>Reset the zoom to 1x</p><br>
                    <p id="tooltip-ZOOM_RESET" class="key_tooltip">Keybinding:</p>
                </span>
            </div>
        </fieldset> -->```

PR #13792, opened 2023-05-05
Removed unused code #13742

a21oligu:
Removed a section of code which was commented and unused. Solves issue #13742


a21aldsa (approved):
Looks good!

Issue #13786, opened 2023-05-05
Replace variable names a and b for function sortElementAssociations


b21phiro:
## Description
In the file ```diagram.js``` the function ```sortElementAssociations(element: object)``` sorts the associations for each side of an element as described in the comment. 

It has a bunch logical operations going on, in which it uses callback-functions with parameter names that are bit unclear.

```
function sortElementAssociations(element)
{
    // Only sort if size of list is >= 2
    // h : Replace variable names a and b
    if (element.top.length > 1) element.top.sort(function (a, b) { return sortvectors(a, b, element.top, element.id, 2) });
    if (element.bottom.length > 1) element.bottom.sort(function (a, b) { return sortvectors(a, b, element.bottom, element.id, 3) });
    if (element.left.length > 1) element.left.sort(function (a, b) { return sortvectors(a, b, element.left, element.id, 0) });
    if (element.right.length > 1) element.right.sort(function (a, b) { return sortvectors(a, b, element.right, element.id, 1) });
}
```

## Your job

1. Try to get a grasp what the parameters ```a``` and ```b``` are suppose to be.
2. Give them a better descriptive name.

## Caution
Do not change the functionality of the function. Input and output should be the same as of right now. Make sure it returns the same values as before. Only change the names to something more suitable for the parameters.

PR #13794, opened 2023-05-05
 Replace variable names a and b for function sortElementAssociations #13786 

a21aldsa:
following the given names for the params for sortvectors, which sortElementAssociations calls. I figured currentElementID and compareElementID were suitable variable names instead of 'a' and 'b'


a21oligu (approved):
Looks good!

Issue #13783, opened 2023-05-05
Same variable name as outer scope?


b21phiro:
## Description
In the file ```diagram.js``` there's seems to be an issue where an instance of the variable ```var element``` seems to allready exist, or the very least, same naming as a global variable outside the scope. 

## Your job.
Try to investigate if this is the case, and if the declaration of this variable at line 7079 (at this time) interfere with the global one. 

If there is an issue, rename the local variable. Make sure that the output is the same as before. And remove the todo comment.

Else, just remove the TODO comment.

PR #13795, opened 2023-05-05
Variables should use local scope #13783

a21oligu:
Updated some "element" variables to use `const` instead of `var`. Solves issue #13783


a21nibah:
are you sure the variable should be a constant?

a21oligu:
> are you sure the variable should be a constant?

As an element is an object, const is fine for local scope. The other option would be to use let, which is prohibited by the project's code standard.

a21nibah (approved):
just wanted to make sure since if something is a constant the variable can not be changed later. Everything looks fine.

Issue #13769, opened 2023-05-04
Create dynamic function for fetch of diagrams.


b21phiro:
## Description
In the ```diagram.php``` file, the function ``` fetchDiagram() ``` have a ajax call inside using JQuery. 
```
$.ajax({
    async: false,
    method: "GET",
    url: `diagramservice.php?courseid=${courseid}&did=${did}`,
}).done((res) => {
    console.log(res)
    response = res;
}).error((req, status, err) => {
    console.error(err);
});
```

Create a function that takes ```courseID``` and ```did``` as parameters and return a response. Make sure you get same output as the old strip of code. Make sure to use block-comment for function, and use sensible naming for variables and function. 

If you need to break up into multiple functions, do that.

## Please
(Avoid JQuery).

## Caution!!!
Do **not change the code** inside the ```fetchDiagram()```. Make a function with the same functionality as the ajax call, but as a dynamic function.

## Bonus
If you want plus points from me personally, use try-catch for error handling.

## Tip
Test the new function by calling it using hardcoded values as parameters. The return of the function should be the same as previous AJAX script.

Use JS fetch API:
https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch

PR #13799, opened 2023-05-05
G1 2023 v5 #13769

b21matfi:
created a new function when fetching the courseId and did, gives the same response as before.


c21adano (approved):
Works!

PR #13805, opened 2023-05-05
Sqlite metadatadb is now deleted and recreated when running install script

a21henhe:
Previously a user would manually have to delete the metadata-database in order to update it to its latest changes. Now, a user only needs to run the installscript and it deletes and creates the metadata-database automatically.

Note: This issue causes a bug which has been separated to a subissue: #13806
It causes error-alerts indicating that some files already exists and therefore they are not created again.

Co-Authored by b21leowa


a21rammo:
Lets not merge code that contains known bug. Its better to resolve the bug in the subissue and ensure that no error alerts popup before merging changes into weekly branch. The way I understood this bug comes with the new code from this PR.

a21henhe:
@a21rammo We have now solved the found bug in issue #13806 and you may now review the code

a21rammo:
> @a21rammo We have now solved the found bug in issue #13806 and you may now review the code

tested and works, good job.

a21rammo (approved):

PR #13808, opened 2023-05-05
added pop out for state elements

c21adano:
added a pop out for initial, final and super state. I also changes the id of sequance element to not conflict with state elements.


a21aldsa (changes requested):
The highlight of the selected option does not work:

image

As you can see, the initial state is still highlighted even though i have selected super state. Compare this to previously existing pop outs:

image

Note how the IE in this pop out is highlighted to reflect my current option.

a21aldsa (approved):
works

Issue #13735, opened 2023-05-04
[NOTE: #13734 must be done] Translate PHP into native JS in diagram.php.


b21phiro:
# Note
Must be done AFTER issue https://github.com/HGustavs/LenaSYS/issues/13734. 

# Goal
The function ```fetchDiagram()``` ```inside diagram.php``` has PHP inside JavaScript. 
```
<?php 
                if (isset($cid) && $cid != "UNK") {
                    echo "const courseid = '$cid';";
                } else if (isset($_GET["folder"])) {
                    $folder = $_GET["folder"];
                    echo "const courseid = '$folder';";
                } else {
                    echo "const courseid = '1894';";
                }

                if (isset($quizid) && $quizid != "UNK") {
                    echo "const did = '$quizid';";
                } else if (isset($_GET["id"])) {
                    $id = $_GET["id"];
                    echo "const did = '$id';";
                } else {
                    echo "const did = '21';";
                }
            ?>
```
Translate the code from PHP into JavaScript.

## Background
This is to clean up the code inside diagram.php.

PR #13809, opened 2023-05-05
G1 2023 v5 #13735

a21nibah:
the php code has now been removed from diagram.php as according to the issue. Do to scoping issues with modules after talking to issue responsible we decided to just include the module functionality in diagram.php and remove search.js. There are also some issues with the output  that will give errors but those are other issues that will be fixed. the output that you get with the php code there and without the php code is the same.


b21oscpe (approved):
I noticed that you put the wrong issue-# ID which complicated testing a bit hehe

looks good, seems to work as normal. Tried both diagram.php and navigating from courseed.php
(Other than the 500 error)

PR #13797, opened 2023-05-05
G3 2023 v5 issue#13668

c21rebja:
issue #13668 

"The bfs and fetchgithubrepo.php functions has been redesigned and the functions has been implemented under the refresh button on the navigation meny at the top.

The bfs function now look at an option parameter to be able to download and insert to sqlite or just insert to sqlite. We also changed so that when creating a new course all the files from the repo are not created on the webserver, the metadata from the files are added to the sqlite db. We tested that the files can be created with the DOWNLOAD parameter and it works.

The refresh button in the navbar is set to the REFRESH parameter to fetch the metadata from the githubrepo only if there has been a new commit to the repo."

To test this, create a course with a github repo you can edit (preferrably your own). After creating a new course, go to the course page and press the refresh button in the top navbar. It should say the course is up to date. Then open the sqlite database in CMD and "select * from gitRepos;" This will show you the latest commit for that repo.  Now go to the repo you have on your github and make some changes (so that the commit nr will change), then press the refresh button again. This time it should say the course has been updated. Then check the sqlite database again to see if the commit number for the specific course has changed.

co-author @g21emmka

PR #13812, opened 2023-05-05
G3 2023 v5 issue#13732

a21lukpo:
Added a new column to the file editor. There should now be a column that is called "Type".  The data inside should be either "Manual" or "Github". This is decided by the method used when adding files. If you manually upload files the type is "Manual" and if you upload a course from github the files added should have the type "Github". We also added 2 radio buttons that will be used to filter files (they are not working currently, will fix the filtering in future issue).

Relevant files to test: fileed.php

Co-authored by @e21krida and @a21hugni

### DIFF
--- a/DuggaSys/fileed.js
+++ b/DuggaSys/fileed.js
@@ -379,9 +379,9 @@ function renderCell(col, celldata, cellid) {
     } else if (col == "kind") {
         str += "<span>" + convertFileKind(celldata) + "</span>";
     } else if (col == "type") {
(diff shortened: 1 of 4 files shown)


c21rebja (approved):
Looks good. When adding a new course the files from that link appear with the type github. 
image
Since this branch doesn't have the fix for course id's yet (PR #13797) and the files all still appear in course folder 1, it might be worth testing these two issues together when both are merged to make sure everything works together.

Issue #13727, opened 2023-05-04
Create the functionality for sequence object


b21eriho:
Should be done after issue #13684 

The button for creating a sequence diagram object should be implemented. It should be made according to the notation found in issue #13535 . Meaning a label with a dashed line extending below it. Please use a similar method to the one utilized in issue #13684 so that other items may later be able to interact with both of these elements.

PR #13817, opened 2023-05-08
 Create the functionality for sequence object #13727 

a21aldsa:
Renamed actor to actorOrObject, gave this a new attribute called actorOrObject which will determine which svg is drawn for that specific element, actor or object. Created new svg for object and set it behind an if statement checking the aforementioned actorOrObject attribute. Added the functionality of switching between these two via two buttons in the options pane via a function I created called toggleActorOrbject.


a21aldsa:
I decided to add the functionality to the options pane as to not need a button for objects in the menu. As discussed here: #13773

b21matfi (approved):
GoodJob! looks good

Issue #13709, opened 2023-05-03
Add a pop-out for sequence buttons


b21eriho:
The buttons for sequence lifeline, sequence object and sequence activation should function as a pop-out like there is now for entities and relations. The button to be selected when first booting the system should be the lifeline.

PR #13818, opened 2023-05-08
Add a pop-out for sequence buttons #13709

b21oscpe:
The buttons now work as intended however they place super states and they are not always placing things correctly. Other than that, the pop-out is there and works as it should.

Because I thought i needed to create a diagramType SE (Sequence) it has been created as diagramType.SE at line 1067. This may conflict.


b21eriho:
This has been moved to the next stable-branch due to conflicts close to deadline that are too much work to fix today.

Issue #13417, opened 2023-04-19
The toggle error-button does not give any feedback


b21oscpe:
![2c94dea66696ce3347f406b4ba20f900](https://user-images.githubusercontent.com/102578165/233085303-e461a840-9b6f-4ff9-ba97-6d04ad79eb0f.png)

The button should either send a system message, like the ones you get when you successfully connect a line between two entities, or have some sort of container showing if it is ON or OFF.

b21sarbo:
![image](https://user-images.githubusercontent.com/102578759/234779618-e4e4ff10-126b-46f0-8843-f6f4241d1404.png)

"Like the ones you get when you successfully connect a line between two entities". The green box.

b21sarbo:
![image](https://user-images.githubusercontent.com/102578759/234780074-ce1e85c1-5921-4835-87dc-ee4c9e058a55.png)
![image](https://user-images.githubusercontent.com/102578759/234780123-7fd3d5f4-6070-441e-adb6-e6088abb0d4c.png)
![image](https://user-images.githubusercontent.com/102578759/234780326-cf8afbca-9524-4b48-93a4-6d0e54488d2e.png)

Picture 1: The check tool is off.
Picture 2: The check tool is on.
Picture 3: The user hovers over the check tool.

b21sarbo:
![image](https://user-images.githubusercontent.com/102578759/236768383-729697c1-4592-48c8-90d7-0bab4b8c0c30.png)

I have added a green box showing whether the tool is on or off.

b21eriho:
Looks good, go ahead and make a pull request

PR #13819, opened 2023-05-08
g1-2023-v5-#13417

b21sarbo:
The issue has been fixed, I have added green boxes showing whether the Error Check tool is on or off when toggled.


c21theti (approved):
Looks good! The toggle-Error button  is now giving feedback

Issue #13728, opened 2023-05-04
Add sequence activation to the system


b21eriho:
Should be done after issue #13684

In accordance with the notation found in issue #13535 we want a box-like element that can be positioned along the dotted line in the lifeline element made in issue #13684 and that can also be of varying height. There should be able to be multiple activations on one line and the activation-boxes should snap to the middle of the line horizontally. 

This element should ideally be placeable using the "sequence activation"-button, if this cannot be achieved within the timeframe that will become a later issue instead.

PR #13824, opened 2023-05-08
Activation element for sequence diagrams #13728

a21oligu:
Added activation element for sequence diagram. Can be created by clicking the button in the toolbar. 

Is missing the "snap" functionality where an element should snap to the middle of the timeline, however one can place an activation element on top of a timeline.

Solves issue #13728


b21phiro:
## Note
Developer wanted to put the snapping feature in future issue due to time. I granted him permission to do so.

Issue #12860, opened 2022-05-23
Remove label and cardinality from IE elements.


b20thelo:
Make if-statement so that  UML label and cardinality isn't available from IE elements.

PR #13829, opened 2023-05-08
Prevent label and cardinalities on IE lines #12860

a21oligu:
It is no longer possible to add and render labels to IE lines as they will be ignored. Solves issue #12860


b21willu:
Looks good! Not possible to add and render labels or cardinalities.

PR #13831, opened 2023-05-08
added a function that only retrieve lines from svgbacklayer

c21adano:
added a function that retrieve lines from svgbacklayer then checks if any lines were retrieved, if none were return empty array.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -2238,19 +2238,39 @@ function mouseOverSelection(mouseX, mouseY){
 }
 
 /**
- * @description Calculates if any line is present on x/y position in pixels.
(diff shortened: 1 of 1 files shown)


a21oligu (changes requested):
I would consider renaming to _backLayerLines()_ for better readability, or even a more descriptive name such as _getLinesFromBackLayer()_.

Also, the `var lines = bLayerLines` at line 2273 is not calling the function but is instead referencing the function, thus the actual value returned from a proper function call is not assigned.

The solution could use an array function such as `map` to make the code even more readable.

As of now, the code is broken and the line selection is not working at all.

a21oligu (approved):

a21oligu:
The fixes seems to solve the previously described issues. Good job!

PR #13836, opened 2023-05-08
It is now possible to filter github and manual files

a21lukpo:
We added functionality so you now can filter between github files and manual files. This is done by clicking on the "Manual" or "Github" radio buttons or clicking on the "Type" heading. This can be easily tested on this link since we have uploaded github files to this course. 

Link: https://cms.webug.se/root/G3/students/e21krida/LenaSYS/DuggaSys/fileed.php?courseid=1&coursename=Webbprogrammering&coursevers=45656

Relevant issue: Issue#13731

image

### DIFF
--- a/DuggaSys/fileed.js
+++ b/DuggaSys/fileed.js
@@ -668,7 +668,7 @@ function compare(a, b) {
 				var tempB = a;
 		}
 	
-    if (col == "filename") {
+    if (col == "filename" || col == "type") {
         tempA = JSON.parse(tempA);
         tempB = JSON.parse(tempB);
         tempA = tempA.shortfilename.toUpperCase();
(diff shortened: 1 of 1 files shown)


b21geoyo:
I tested the filtering options for manual and github. I manually added a file and it showed up when filtered it to manual.

I also tested the github filtering option, and it showed the github files that have been previously added.
However I had some issues adding github files, therefore I made an INSERT INTO the gitFiles table in the SQLite database and the row I added did NOT show up in the list of files.

a21olljo:
After that b21geoyo had issues with testing and adding new course to then se if what was downloaded could be sorted, I tested it and found that everything worked as it should. After speaking with b21geoyo we approve this merge request.

PR #13837, opened 2023-05-08
G2 2023 v5 issue#13791

a21ponkr:
converted showDuggaservice-test.txt to markdown


b21hamwe (changes requested):
you've missed Segment which is an output in the info variable.
Add segment as output and more that you've missed.

PR #13841, opened 2023-05-08
G2 2023 v5 issue#13393

c21axepe:
Created a txt document with tests for all services in accessedservice.php


b21antbr:
I co-authored on this!

b21antbr:
Added tests for accessedservice.php as file accessedservice-test.txt (since no description was provided)

a19hakja (changes requested):
The values looks good to me, since the rest of these test-cases are in the process of being converted to .md, i talked to the devs about doing that with this one aswell, in this pull request, which we agreed on would be a good thing to do. Therefore I'll request changes, with those changes being the conversion to .md.
Good work!

PR #13842, opened 2023-05-08
G2 2023 v5 issue#13820 - Change fileedservice-tests.txt to MD and add expected output

a21oscgu:
- Convert to Markdown .md
- Add expected output
- Make sure to delete old .txt and replace with new .md


c21axepe (approved):
The txt file is converted into md and the expected output is added
Great job!

Issue #12883, opened 2022-05-24
IE-inheritance bugs out when changing between disjoint and overlapping


a20samul:
For future developers:
We managed to implement IE in the diagram, however a bug has snuck its way through.
When placing many IE-inheritances (the half-circle on the expand-subtoolbar) on the diagram and changing the inheritance of the previous one to overlapping (by selecting one inheritance and then opening the options panel to the right and switching from disjoint to overlapping throught the dropbox) the latter ones dissappear.

PR #13843, opened 2023-05-08
G1 2023 v5 #12883

b21matfi:
Removed the code that was causing the bug, you can now switch to overlapping and move it around with out it breaking the diagram.


a21aldsa (approved):
Solid work, good job

Issue #12889, opened 2022-05-24
Hiding instructions need improvements.


b19karhj:
When you hide the instructions, the instructions field turns white.
![image](https://user-images.githubusercontent.com/81624541/170040907-0ee94867-6d93-497f-bacd-7570ce0c9e09.png)

It should not turn white during the animation. It should work the same way as when you hide general information

PR #13844, opened 2023-05-08
G1 2023 v5 #12889

c21adano:
fixed the left side description on the dugga site to not vanish when the closing animation plays.


b21matfi (approved):
image
Good job it looks mych better now!

Issue #12694, opened 2022-05-19
ER to UML, UML to ER options panel dropdown issue


a20samul:
- When switching from ER to UML in the options panel using the dropdown  and pressing the enter-key the element changes but the property panel stays the same until save button is manually clicked. This change should include both to change simultaneously.
- When changing back from UML to ER, the property fieldset never changes back even if you press, save so you have to exit out of the options panel and then re-choose the element and reopen the options panel, which can be a hassle and hard for new users to know.

This might be somthing for future developers to consider.

<img width="283" alt="ertouml" src="https://user-images.githubusercontent.com/81613484/169219997-9ea2589d-72fe-4f04-beb6-dcee68216039.png">
<img width="273" alt="umltoer" src="https://user-images.githubusercontent.com/81613484/169220000-8c24a414-4519-4005-beaa-257e0c7e0df1.png">

b21oscpe:
I have a question regarding this issue as it seems like it has been changed a bit since it was created.
![fd8ebac17f991dcf71c6168eaeca815d](https://user-images.githubusercontent.com/102578165/236803377-df3b9126-2bdf-4fb5-9dd2-b92ed871a0ba.png)
This code (which i assume was added later on) suggests that UML and ER entities may not be changed back and forth as they "should not be able to connect". See error message:
![Skärmklipp](https://user-images.githubusercontent.com/102578165/236803593-a0f8d9fc-a7c6-4989-952c-294fe6c7fdf6.PNG)

Is this something I should revert for this issue? It is possible to change from ER to UML however it does not work the other way around

PR #13864, opened 2023-05-08
ER to UML, UML to ER options panel dropdown issue #12694

b21oscpe:
It is now possible to switch types with enter and the save button with the options-pane acting accordning to the switch. 

An another note:
The type-changing is not 100% accurate and desirable yet (Because UML classes cant count lines for some reason). This makes it so that it is possible to change from UML - UML to UML - ER, between relations

Image below shows that the system seems to think there are 0 lines connected to the UML element:
7d8560efd4160df8975b00990d27f54a

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -2895,14 +2895,12 @@ function changeState()
     }
 
     else if(element.type=='SD') {
-        // State diagrams don't have properties, so this feels unneccessary (?)
-        /*
+
         if (element.kind != 'IEEntity') {
             var property = document.getElementById("propertySelect").value;
             element.state = property;
             stateMachine.save(StateChangeFactory.ElementAttributesChanged(element.id, { state: property }), StateChange.ChangeTypes.ELEMENT_ATTRIBUTE_CHANGED);
         }
-        */
 
         //Check if type has been changed
         if (oldType != newType) {
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -2832,7 +2832,7 @@ function changeState()
     }
 
     else if (element.type == 'ER') {
-
+        
         //If not attribute, also save the current type and check if kind also should be updated
         if (element.kind != 'ERAttr') {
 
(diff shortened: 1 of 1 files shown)


b21oscpe:
Original issue: #12694

a21aldsa (changes requested):
When switching anything to SD, everything breaks. Things also break when switching from IE to UML. Also you cannot switch SD to anything, it just says "successfully saved" without any visible changes.

b21oscpe:
SD worked like this from the start,

I looked into it now and found that this is because it was added recently and hasnt been configured properly within the stateChanged() aswell as many other areas of the code

b21oscpe:
SD should now work as intended, however, lines aren't counted correctly for UML & SD, this is another issue though

a21aldsa (changes requested):
As previously noted, switching to UML can break things but it seems that this will become a new issue, judging by the comments on this PR. 

Something I would like to see fixed though is the rules for switching type while an element is connected. It is currently possible to change any UML, IE or state to anything, even if it already has connections:

image

b21oscpe:
In what way does UML break?

Yes as previously mentioned the lines function is broken

a21aldsa:
> In what way does UML break?
> 
> Yes as previously mentioned the lines function is broken

We had a private conversation about this but the summary is that we made sure I was on the right branch and I showed him that placing text on elements, like functions or name. I also showed him how you can create non allowed connections like IE-state. And how changing IE, or state to UML breaks everything.

a21aldsa:
> In what way does UML break?
> 
> Yes as previously mentioned the lines function is broken

We had a private conversation about this but the summary is that we made sure I was on the right branch and I showed him that placing text on elements, like functions or name. I also showed him how you can create non allowed connections like IE-state. And how changing IE, or state to UML breaks everything.

b21eriho:
What does that mean for this issue?

a21aldsa:
> What does that mean for this issue?

Its essentially just a testing I did and then reported my findings verbally while we both looked at the same screen. This does not affect the issue more than any other testing. The comment is just to make the test documented.

a21aldsa (approved):
Since this issue only covers switching element types via the drop down panel, and this now working perfectly fine; I will consider this issue to be resolved. All issues that were discovered during this (like the ability to make illegal connections to other elements by switching the type) have been made into new issues, as far as I am aware. Therefore this issue is resolved.

PR #13866, opened 2023-05-08
added a button for sequance option loop

c21adano:
added a button for sequance option loop

### DIFF
--- a/DuggaSys/diagram.php
+++ b/DuggaSys/diagram.php
@@ -558,6 +558,14 @@ function search(key)
                         <p id="tooltip-STATE_SEQUENCE" class="key_tooltip">Keybinding:</p>
                     </span>
                 </div>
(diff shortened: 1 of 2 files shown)


b21oscpe (changes requested):
Button is not highligthed when selected. I suggest adding something along the lines with "active" along with its other class-attributes
06decddc1f84405c116ebda35c439a8a

b21oscpe:
I believe it is done with "holdplacementbuttonup()" if im not mistaken

c21adano:
> Button is not highligthed when selected. I suggest adding something along the lines with "active" along with its other class-attributes 06decddc1f84405c116ebda35c439a8a

Its not highligthed when selected due to currently being linked with er attribute like how the other sequance elements are linked to super state.

b21oscpe (approved):
I took a look and it seems like you're right, looks good

Issue #13861, opened 2023-05-08
Diagramservice.php throws a 500


b21eriho:
Currently in the network tab of the inspector any page with a diagram dugga has diagramservice.php throw a 500 error. This has been the case since the v2 branch but is not present in the v1 branch. The exception to this is the branch g1-2023-v1-#13102 which may have caused the issue though this is only might be the case. Figure out what causes this and fix it.

PR #13887, opened 2023-05-09
Diagramservice throws a 500 HTTP error #13861

a21oligu:
Fixed an issue where fetching the diagramservice.php file would cause the server to throw a HTTP 500 error. Added an if-statement for the "github" code and moved some error-prone code into the try-catch clause. Solves #13861

### DIFF
--- a/DuggaSys/diagramservice.php
+++ b/DuggaSys/diagramservice.php
@@ -27,92 +27,95 @@
         $firstname = "UNK";
     }
 
-    $opt = getOP('opt');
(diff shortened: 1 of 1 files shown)


a21oligu:
> Good job the issue if fixed, but there is a console log there, which should be removed before merging

On which line?

b21matfi:
in diagram.php line noticed that it was not you who put it there so i will do a new review.

b21matfi (approved):
Good job!

PR #13889, opened 2023-05-09
Updated svg class diagram for sectionedservice.php issue#13758

b21tilbe:


b21kurar:
This PR is for `sectionedservice.svg`, as such there shouldn't be any changes to any other files. You should remove the file `duggaedservice.svg` from the branch.

b21kurar:
There are some unnecessary parameters in the functions. Please go through and remove those variables that aren't sent as input to the service. 

Some examples for `newCode()`:
`$debug` - is created internally and outputs debug info.
`$pdo` ,  `queryz2` - just temporary variables storing an SQL query (and executes them)
`$armin` - temporary variable for exploded string

There may be more, and in the other functions as well.

One way to confirm if they should be paramters is to see if they are aquired using `getOP()` at some point in the file (usually at the top)
image

b21kurar (changes requested):
This PR should not have `duggaedservice.svg`, please delete it
There should be some minor updates regarding which parameters are relevant (see comments)

c21alest:
I can also see that the new file is in the wrong folder

Issue #13810, opened 2023-05-05
Straight SD-arrows vertical offset too large


b21eriho:
The vertical offset into SD-entities is too large for straight arrows. Fix this

Issue #13689, opened 2023-05-03
Fix rotation for triangle to align with draw-line.


b21phiro:
![bild](https://user-images.githubusercontent.com/45258330/235898757-20d289b2-074e-4e8d-8a7c-ea5def9fc82b.png)

The rotation of the triangle for lines are not aligned correctly.

d21matbo:
I've with some help from @b21eriho been able to figure out the mathmatics involved in solving this issue. However there exists a problem that is currently blocking the implementation of the solution.

The StateDiagram lines have an offset that places the end of the line inside element which is used to remove the gap between the end of the line and the element. This solution was implemented before the addition of "straight" lines, and as such didnt have any problems as the lines would always enter on a prefectly square angle (which lines up with the arrow). 
Currently the offset is only a problem for the "vertical" lines as can be seen in the image below, where the line should end on the tip of the arrow (just like how the "horizontal" line is placed).

Continuing on this implementation would be somewhat counterproductive as the offsets need to be resolved first. This is because how the lines connect to the elements greatly affects how the implement the solution to this issue.

![image](https://user-images.githubusercontent.com/102599161/236460369-b487cda7-287b-4343-8683-6c09625be826.png)

PR #13892, opened 2023-05-09
G1 2023 v5 #13810

d21matbo:
Fixed issue with  y-offset that is currently blocking the implementation of issue #13689

The lines now ends closer to the edge of the elements when the line type is straight

image


a21oligu:
This seems to only work when zoomed in, but does the opposite when zoomed out. I don't know if it is the selection rect that is scaling wrong, or the line's anchor points to the edges?

a21oligu:
> The scaling of the selection rectangle is not completely accurate, but that was never the case when you zoomed in on elements. Changed the values so that it should be closer

Appears I tested the "g1-2023-v5-#13810" branch instead of "g1-2023-v5-#13810-2". My bad. Everything looks good

a21oligu (approved):
Looks good! I approve the changes, but the merge conflict should be resolved.

a21oligu:
This pull request is ready for merge

PR #13894, opened 2023-05-09
New version of class-diagram accessedservice.svg

a20fanma:
Added a new version of class-diagram accessedservice.svg


a21oscgu (changes requested):
- The method "logUserEvent($userid, $username)" should be renamed to **"logUserEvents"**. Look at: [backend-models/accessedservice.svg](https://github.com/HGustavs/LenaSYS/blob/44f973e4d27a94e1a23932e941dd1501dc3868ce/backend-models/accessedservice.svg)

d21erisu:
great

PR #13896, opened 2023-05-09
G3 2023 v5 issue#13863

e21krida:
The issue at hand was to implement the refresh button that existed for each "code-example" onto each "section" (Not the same as each "moment"). So i simply moved it, commented out potentially irrelevant functionality and added a simple console.log on the button to test functionality. Why we only managed part of it was partly because of the git-issues earlier and no time to implement the rest asked for before the weekly merge tomorrow, hence a new issue will be opened to handle the potential back-end functionality. 

To test this, simply open a course via sectioned.php and log in as adminstrator, then test the refresh button next to each "section", for example the one in the picture below. 

Screenshot 2023-05-09 152608

If incorrect, this "refresh" arrow will appear on each code example. If correct, it should be moved to each section. On click, it should print out a simple console.log message in the form of "RefreshButton Clicked!"

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -1664,7 +1664,7 @@ function returnedSection(data) {
         }
 
         // Testing implementation
-        if (itemKind === 4 && data['writeaccess'] || data['studentteacher']) {
+        if (itemKind === "section" && data['writeaccess'] || data['studentteacher']) {
           str += `<td style='width:32px;' class='${makeTextArray(itemKind,
             ["header", "section", "code", "test", "moment", "link", "group", "message"])} ${hideState}'>`;
           str += `<img style='width:16px' alt='refresh icon' tabIndex='0' id='dorf' title='Refresh code example' src='../Shared/icons/refresh.svg'`;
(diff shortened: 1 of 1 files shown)


c21rebja (changes requested):
The console.log appears as it should, so everything seems to be working as described.

However, it feels very confusing to have the refresh button in two different positions in moment and in section (see below). Is this something that is going to be changed in the later issue for the backend part? Otherwise, we feel like this should be changed before the merge.
image

co-author @g21emmka

e21krida:
> The console.log appears as it should, so everything seems to be working as described.
> 
> However, it feels very confusing to have the refresh button in two different positions in moment and in section (see below). Is this something that is going to be changed in the later issue for the backend part? Otherwise, we feel like this should be changed before the merge. image
> 
> co-author @g21emmka

Fixed!

Screenshot 2023-05-09 155913

g21emmka (approved):

c21rebja (approved):
Looks good! :)

Talked to @a21olljo and @d21emijo and organizing the different buttons for all of the different rows (section, moment, etc) will be an issue for later anyways, since everything is bound to move around more as we progress.

PR #13898, opened 2023-05-09
G3 2023 v5 issue#13821

e21krida:
Fixed the previous issue that @a21hugni couldn't solve, what i did was essentially feed the filePath into the json object for both "trashcan" and "editor" objects as well and used the filePath to determine if the path correctly contained the word "Github", hence marking it as a file originating from Github. If it is a github file i simply emptied the column containing the pen to edit and replaced the trashcan with an octopus-emoji as requested. 

Edit: Changed octopus-emoji to github-icon, as i found it's already saved on the webserver. Was unsure about the rights regarding the icon, but if it's already in use i assume somebody smarter and wiser than me made the decision that it's okay.

Screenshot 2023-05-09 180429

Relevant files: DuggaSys/fileed.js & DuggaSys/fileedservice.php

### DIFF
--- a/DuggaSys/fileed.js
+++ b/DuggaSys/fileed.js
@@ -363,11 +363,11 @@ function renderCell(col, celldata, cellid) {
     if (col == "trashcan") {
         if (obj.showtrashcan) {
             console.log(obj.filePath);
-            if(obj.filePath == null) {
+            if(obj.filePath.includes("Github")) {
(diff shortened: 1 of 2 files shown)


a21lukpo (approved):
Looks good! :)

a21olljo:
I have now tested and verified as well, and it looks good. It was decided yesterday that @a21lukpo where to be the reviewer because of time constraints before the merge, but for the future it is still ment to go through me following that I'm also suppose to verify that the solution is good.

Issue #13906, opened 2023-05-10
Bugfix: Menu buttons such as "Sequence Life" does not work.


b21phiro:
## Description
Some of the buttons seems to notwork. Find out why and try to fix it.

![bild](https://github.com/HGustavs/LenaSYS/assets/45258330/5b405832-1b59-46a0-ab29-d1e85cac7e53)

PR #13912, opened 2023-05-10
g1 2023 v6 #13906

b21matfi:
changed what placing type it is and the issue is fixed you can now place sequence life

### DIFF
--- a/DuggaSys/diagram.php
+++ b/DuggaSys/diagram.php
@@ -571,7 +571,7 @@
                     </span>
                 </div>              
                 <!-- Sequence diagram object selection -->
-                <div id="elementPlacement13" class="diagramIcons toolbarMode" onclick='setElementPlacementType(2); setMouseMode(2);' onmouseup='holdPlacementButtonUp();'>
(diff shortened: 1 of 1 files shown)


b21eriho:
There are other buttons that do not work, these are also a part of this issue. That is why they wrote "such as", please ensure that all buttons do what they should, the exception to this is the button for the button that is the furthest down with the icon for option loop or alternative which is not yet implemented.

b21eriho:
From what I can tell selecting sequence activation will mark the sequence object button but still create a sequence activation object on click, I believe

b21matfi:
Alright i will take a look at it

b21matfi:
The issue is fixed, the bug came because we reuse alot of the code but we don't change the ID, same thing is that some of the button that dont have functionality place the attribute element because we reuse the code without removing or changing the code to apply for the element.

b21eriho:
I see

Den ons 10 maj 2023 14:23b21matfi ***@***.***> skrev:

a21aldsa (approved):
All menu buttons work except sequence object and option loop. This is how it should be, good job.

Issue #13905, opened 2023-05-10
UML Super State darkmode issue


b21phiro:
## Description
The UML Super States frame is not visible during darkmode due to the stroke being black. 

![bild](https://github.com/HGustavs/LenaSYS/assets/45258330/8e18e635-455b-43eb-81f9-635826b130e7)

## Your job
Make sure the stroke-color is inverted between dark/light mode, and are visible.

PR #13921, opened 2023-05-10
 UML Super State darkmode issue #13905 

a21aldsa:
created a variable for holding the stroke color for special cases like super state outer border and sequence actor body called nonFilledElementPartStrokeColor. Then implemented that into actor and super state as the stroke color for the parts that do not have a fill. and the actor text.


a21oligu (approved):
Looks good!

Issue #13785, opened 2023-05-05
Fix function linetest()


b21phiro:
## Description
In the file ```diagram.js``` there is a function called ```linetest()```. This function seems to have couple of TODO's in which there is a disagreement.

```
// TODO : WHY does it return EITHER a boolean OR an object??????? Either this is a TRUE/FALSE function and return booleans OR it returns objects/null/undefined!
 // TODO : Use new POINT objects to reduce amount of arguments?
function linetest(...) {...}
```

The issue seems to be that the function is unclear in what it should return. Either, it should be returning a boolean value, or objects/nulls.

## Your job
1. Try to get a grasp of what the function does.

2. Decide on if the function should return Boolean values, or objects. Not both. Remember that no lines can/will not be found, and the function must reflect on that. Do NOT change the functionality of the function, just decide on what the returns should be.

3. Remove any unused code inside the function, and the comments that are for unused code.

4. Update the block-comment, because right now it has no real description after you are done. Make sure to state what it returns.

## Tip
You can test the function by calling it, and putting some values as parameters. You could try to see if you can trigger the function by playing around with lines too.

PR #13923, opened 2023-05-10
Fix function linetest() #13785

a19clasv:
1. Made it clear what it does now. Had to brush up on my linear algebra for that.
2. It should return a boolean if it's necessary to change it. It's only used to get a boolean value, but it seems like it was also supposed to be used to get intersection points.
3. Removed.
4. Updated and stated.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -7498,10 +7498,7 @@ function linetest(x1, y1, x2, y2, x3, y3, x4, y4)
             if (!(y3 <= y && y <= y4)) return false;
         }
     }
-    return { 
-        x: x,
-        y: y 
-    };
+    return true;
 }
 
 /**
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -7498,7 +7498,10 @@ function linetest(x1, y1, x2, y2, x3, y3, x4, y4)
             if (!(y3 <= y && y <= y4)) return false;
         }
     }
-    return true;
+    return { 
+        x: x,
+        y: y 
+    };
 }
 
 /**
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -7467,11 +7467,6 @@ function sortvectors(currentElementID, compareElementID, ends, elementid, axis)
  // TODO : Use new POINT objects to reduce amount of arguments?
 function linetest(x1, y1, x2, y2, x3, y3, x4, y4)
 {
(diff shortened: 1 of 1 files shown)


a19clasv:
Forgot I wasn't supposed to change the functionality. 2. It should return a boolean if it's necessary to change it, but it returns x and y at the moment.

a21nibah (changes requested):
I am not a fan of how returns are handeled right now. Having true and false does not make sence if it does not return true and instead returns an object.
It would be better to return undefined if the lines do not intersekt and an object if it does.

a19clasv:
It's stated in the issue that I shouldn't change the functionality, just decide if the function should return either a boolean or an object. I've already done that.

a19clasv:
Basically, I'm supposed to come up with a decision and not implement it. My decision is in 2 and it's not implemented.

a21nibah:
> It's stated in the issue that I shouldn't change the functionality, just decide if the function should return either a boolean or an object. I've already done that.

the issue states that you should decide weather the function should return either boolean values(true/false) or an object and (null or undefined)
my suggestion is to return an object  and undefined
or if the intersections are not being used anywhere in the code after the test function is called return either true or false.

a19clasv:
"or if the intersections are not being used anywhere in the code after the test function is called return either true or false." I already wrote that it's only used to get a boolean value. That means that you agree with me.

a21nibah:
My bad. I did not read up on this as i should have, sorry about that. The changes that you have made looks good but i would still prefer that this issue is changed so that you can implement the return type. YOu have already changed comments  the function so  it just makes sence to go all the way.

b21phiro:
The function should only return EITHER objects or boolean values, and should NOT change the basic functionality of the function as it's being used.

As of right now, it returns either false or object. I agree with Nibar about object and/or undefined/null rather then false, if you now choose to return objects.

In my opinion the changes are acceptable, but the return of false is not correct as you return an object if found.

b21phiro:
> Right now it's only being used for boolean values, so it would be best to make it return true or false if we were to change the function. And I'm a bit confused about not changing the basic functionality because changing what it returns would do that.

True, but if the checks are for example ``if (!linetest()) `` it should not matter if it's ``null | false | undefined``

a19clasv:
It returns true or false now.

a19clasv:
[Fix function linetest() #13785 fix 3 #13992](https://github.com/HGustavs/LenaSYS/pull/13992) is the new pull request for the issue.

Issue #13911, opened 2023-05-10
Fix hardcoded values for sequence.


b21phiro:
## Description
The Sequence Activation Box have hardcoded values. This is not ideal, and should be fixed.
![bild](https://github.com/HGustavs/LenaSYS/assets/45258330/2683bf60-eab2-41c4-8f4a-33666e451a07)

```
else if (element.kind == 'sequenceActivation') {
        //div to encapsulate sequence lifeline.
        str += `<div id='${element.id}'	class='element' onmousedown='ddown(event);' onmouseenter='mouseEnter();' onmouseleave='mouseLeave()';' 
        style='left:0px; top:0px;width:${boxw}px;height:${boxh}px;`;

        if (context.includes(element)) {
            str += `z-index: 1;`;
        }
        if (ghosted) {
            str += `pointer-events: none; opacity: ${ghostLine ? 0 : 0.0};`;
        }
        str += `'>`;
        str += `<svg width='${boxw}' height='${boxh}'>`;
        //svg for the activation rect
        str += `<rect rx="12" style="height: 100%; width: 100%; fill:rgb(0,0,255);stroke-width:3;stroke:rgb(0,0,0)" />`;
        str += `</svg>`;  
        ```
        
The code is in ``diagram.js``.

PR #13928, opened 2023-05-10
Changed some hardcoded code to use variables #13911

a21oligu:
Changed some of the code for drawing a sequence activation box to use dyncamic variables. Solves issue #13911.

Note: I decided to leave `rx="12"` as is since it controls the border radius and may never be changed when using the program.

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -646,3 +646,10 @@
 .diagramZoomIcons:hover .zoomToolTipText {
     visibility: visible;
 }
+
+/* SEQUENCE */
+.sequenceActivation {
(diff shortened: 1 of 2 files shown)


b21matfi (approved):
Looks good!

b21matfi (approved):
Looks good!

d21matbo:
FYI, This had some issue that needed to be resolved, as the rx"12" does not scale with the current zoomfactor and as such the element changes apperance as you zoom in and out...

Changes requested would have required a variable for the corner radius that would scale with the current zoomfactor, just like the variable `sequenceCornerRadius` does for the actor/object element

PR #13929, opened 2023-05-10
Added pop-out for sequence and added all the buttons

b21oscpe:
Moved code from v5 to v6 which made all sequence buttons into a pop-out


d21matbo:
As a note, creating these tooltips could potentially be moved to another issue since the functionality of the pop-out is there

b21phiro:
> As a note, creating these tooltips could potentially be moved to another issue since the functionality of the pop-out is there

Yes, this seems to be a different issue. I will close and merge this issue and create a new.

PR #13940, opened 2023-05-11
Centered the cursor for ghost elements

c21adano:
Centered the cursor close to the middle of each ghost element except the current sequance element. IE entity ghost appears doesnt have an utline while its a ghost and should be taken care of in another issue.

I also removed "avgcharwidth" which was a dead variable with no current use after getting permission.


d21matbo (approved):
The "ghost" elements are now centered, Looks good!

PR #13941, opened 2023-05-11
G1 2023 v6 #13689

d21matbo:
The SD Arrow should now be rotated correctly when the line is "straight", if the line is segmented it will fallback to the previous implementation. *This applies to both the end & start icon*

image

It is however possible to place the elements in way such that the arrow is somewhat hidden behind the element, if this is deemed problematic, a new issue should be raised

image


c21adano (approved):
works great!

Issue #13919, opened 2023-05-10
Switching ER-relation to SD-type causes problems


b21phiro:
## Description
State diagrams should support logical-operations (see figure 1.). 

Figure 1.
![bild](https://github.com/HGustavs/LenaSYS/assets/45258330/6365ea0c-22df-4f87-a402-1f1b28b62120)

However, changing the type of ER-relation to SD results in problems, as it's probably not implemented as seen in figure 2.

Figure 2.
![bild](https://github.com/HGustavs/LenaSYS/assets/45258330/3635b040-0a93-4ec6-839d-cf4a40e2193f)

## Your job
Make sure logical operations are allowed for SD elements too.

PR #13944, opened 2023-05-11
G1 2023 v6 #13919

b21matfi:
You can now change the ER relation to the type SD and draw lines between SD element if the type of the relation is SD.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -2867,9 +2867,9 @@ function changeState()
         if(element.kind != 'EREntity') {
             if (newType == 'SD') {
                 element.type = newType;
-                console.log(element.type);
-                console.log(element.kind);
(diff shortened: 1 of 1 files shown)


a21aldsa (changes requested):
Switching an ER relation to UML or IE and then to SD makes the SVG invisible for the new SD relation, see screenshot.

image


Also, I believe there should be a menu button for spawning SD relations since this current method of spawning an ER relation and then switching it in the options pane is woefully non-intuitive. It also does not store switching the type in the state machine meaning that undoing/redoing this is impossible. Lastly, you can switch the relation after a connection has been made which creates nonsensical diagrams, see screenshot

image

b21eriho:
Fact of the matter is that SD-relations aren't in out notation meaning the ability to switch to them will be removed in a future issue. In the meantime keep testing so that this issue is solved

Issue #13865, opened 2023-05-08
Add functionality for all entities to elementHasLines


b21eriho:
During another issue ( #13840 )it became clear the counting lines connected to UML and SD elements does not function correctly. Correct this fault

PR #13948, opened 2023-05-11
Add functionality for all entities to elementHasLines #13865

b21oscpe:
### What is changed:

Removed condition that made it so that only ER-elements count their lines and add them to their respective array.

### For tester:

Control this by writing `getElementLines(context[0])` after selecting an element *OR* try switching between types while lines are connected.


b21matfi (approved):
Looks good!

Issue #13914, opened 2023-05-10
Add functionality for sequence loop or alternative


b21eriho:
Create an entity for the option "loop or alternative" in accordance with the notation in issue #13535 . It should have a title and the possibility for dashed dividing lines. It does not yet need to be placeable with the button nor be placeable over other entities.

PR #13949, opened 2023-05-11
 Add functionality for sequence loop or alternative #13914

a21aldsa:
Created the element kind sequenceLoopOrAlternative, then created the svg for this and made a loop that draws a number of dashed lines depending on user preference. The default is 1. This is changed via the **options pane**. If this is set to 0, the Alt will automatically change to a Loop.

Also added the ability for vertical resizing in case the user wants many alternatives. This was done via reuse of the same vertical resizing i made for actor and object (their lifeline)

It is noteworthy that naming this or actor and object does not work at the moment but it is the same in the stable branch right now so it is unrelated to this issue.

# very important: make sure functions and attributes work for UML and IE.


b21matfi (approved):
Good job! it looks amazing

Issue #13644, opened 2023-05-02
Changing the "Inheritance"-property to "overlapping" on "IE inheritance" crashes the system


d21matbo:
When the **SAVE** button is pressed after swapping from "disjoint" to "overlapping" the system becomes unresponsive

![image](https://user-images.githubusercontent.com/102599161/235627173-8017b103-e621-4f03-a527-7314dd27242d.png)

The console throws errors that could be provide insight in how to fix this issue

a21nibah:
is there a diagram i can use  to test this?

PR #13953, opened 2023-05-11
G1 2023 v6 #13644

b21matfi:
You can change the relation now so that it is overlapping or disjoint, by fixing this issue it also solved for the uml


c21adano (approved):
No error occurs when creating or changing inheritance.

PR #13957, opened 2023-05-11
G3 2023 v6 issue#13943

g21emmka:
No longer saving the commit when creating a course. It now saves the commit when pressing the refresh button and instead of just refreshing the metadata it also downloads the files. 

Test this by creating a new course with a github url. 
Check the courses folder to see if you can find a folder with the course id. There should be nothing there.
Open the SQLite database and check the gitRepos table to see that no commit has been saved for the course.
Go into the course you've created and press the refresh button.
Go back into the courses folder. Now there should be a folder with the course id as name, containing the files from the repo you used.
Also check the SQLite database. It should have the latest commit saved for the course. 

co-author @c21rebja

### DIFF
--- a/DuggaSys/gitcommitService.php
+++ b/DuggaSys/gitcommitService.php
@@ -144,8 +144,8 @@
 				$query->execute();
 
 				// Update metadata
-				bfs($url, $cid, "REFRESH");
-				print "The course has been updated!";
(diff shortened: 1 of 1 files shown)


e21krida (approved):
Works great! 👍

a21lukpo (approved):
Works correctly! : )

Issue #13904, opened 2023-05-10
Index file to control which files or directories to be downloaded


a21olljo:
Correction of issue: #13016 
After discussions with the customer; its been clarified that the index file should not control the order of code examples, as stated in the issue. 
But rather be present in the repo to specify which files or directories should be downloaded, more like a filter. Then when the code example is downloaded and generated, the newly created example should be placed in a moment to then be ordered manually through the drag and drop feature.

a21olljo:
After further discussion with the customer and with a short demo/ explanation of how this component now works. It was requested that the index file wouldn't specify what to download, but rather what shouldn't be downloaded, similar to a .gitignore. This change is small and doesn't change much, so it will be done in the already existing pull request.

PR #13962, opened 2023-05-12
G3 2023 v6 issue#13904 Index file now filters code examples

a21henhe:
When fetching a github-repo, we now search for an index file containing the names of what files to keep, located at the root of the repo with the name "index.txt". These names are to be separated with a new line, an example of this is listed below.

For Testing:
1. Fork a repo you want to download into LenaSys
2. Create an index file with the name 'index.txt' in root (/index.txt)
3. Add all the file names you want to add with file extension in the name and separate by a new line
4. Push it up to the repo
5. You should remove the previous metadata-db to start with a fresh one containing 0 gitFiles.
6. Create a course with the GitHub-URL
7. Check database if all the files you added in the index.txt are in the database (metadata-db) by entering the metadata-db from  your terminal and select * from gitFiles .
8. If they exist, try to add another file into the index.txt file or remove the index file completely
9. Push it to your repo.
10. Wait 3-5 minutes untill you refresh the course. (Github api is slow)
12. See if the files are there. 

As mentioned, github-api takes about 5 minutes to actually update which means that you will have to wait. A good way to see the changes and when they happen is by entering the link of your index-file into the browser. An example of that link is listed below, where you replace the user with your username and your reponame:

https://raw.githubusercontent.com/user/name of repo/master/index.txt

Once the update occurs in that link, then you can test on LenaSys to see the change.

When you finally remove the index-file from the repo, when you update the course the database should be filled with all of the files from the repo instead.

image

Co-authored by b21leowa

### DIFF
--- a/DuggaSys/courseed.js
+++ b/DuggaSys/courseed.js
@@ -125,11 +125,12 @@ function fetchGitHubRepo(gitHubURL)
 	regexURL = gitHubURL.replace(/.git$/, "");
 	//Used to return success(true) or error(false) to the calling function
 	var dataCheck;
+	//console.log("Inside fetchGithubRepo with Url: " + regexURL);
 	$.ajax({
 		async: false,
 		url: "../recursivetesting/FetchGithubRepo.php",
 		type: "POST",
-		data: {'githubURL':regexURL, 'action':'getNewCourseGitHub'},
+		data: {'url':regexURL, 'action':'getIndexFile'},
 		success: function() { 
 			//Returns true if the data and JSON is correct
 			dataCheck = true;
--- a/DuggaSys/courseed.js
+++ b/DuggaSys/courseed.js
@@ -125,12 +125,11 @@ function fetchGitHubRepo(gitHubURL)
 	regexURL = gitHubURL.replace(/.git$/, "");
 	//Used to return success(true) or error(false) to the calling function
 	var dataCheck;
-	//console.log("Inside fetchGithubRepo with Url: " + regexURL);
 	$.ajax({
 		async: false,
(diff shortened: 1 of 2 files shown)


c21rebja (commented):
It seems to work like it should! I created a repo and made an index file with a few of the files from the repo. Only the ones specified in the index.txt got added. When I updated the index file and refreshed, the new files got added too. 

I just have a question about removing the index file. When I removed it from my repo, all of the files from the repo got added into SQLite, so that works, but when I pressed the refresh button I got this error.

image

I can only assume it's supposed to be that way, since there's always supposed to be an index file in the repos? I just wanted to double check before approving.

a21olljo:
As mentioned in the comments of the issue https://github.com/HGustavs/LenaSYS/issues/13904#issuecomment-1547269519, the new directive is that the index file should include the files not to show. After discussion with @a21henhe this were not a big issue and required minimum modifications to the code, making it easy to change along with the error handling. So that @c21rebja knows that this has changed, and to test if specific files can be excluded with the index file.

a21henhe:
@c21rebja we have now fixed the 404-error.

Note: Concept of index file has switched, the files mentioned are now instead meant to be ignored. Fetching index file now uses curl instead of file_get_contents in order to retrieve http-headers. Now ready to be tested again 

Co-authored by b21leowa

c21rebja (approved):
I created an index file again, and it worked just like a gitignore and didn't add the files that were mentioned in the index. When I removed a filename from the index it got added to the course next time I updated it. I didn't get any errors if I removed the index file either.

So everything looks good now! :)

Issue #13961, opened 2023-05-12
State without attributes displays incorrectly


b21eriho:
![image](https://github.com/HGustavs/LenaSYS/assets/129077727/a2165c85-6287-4556-b6aa-faa8cc767c3d)

If one leaves the attributes tab os an SD state empty it still shows half of the word do: creeping up from below. Find out why the "do:" is there and report what you find so we can decide what to do from there.

b21oscpe:
I found that there was a harcoded "Do:" left in the code that was added after the "empty" textarea. I simply removed the "do:" so that it displays correctly for now

b21eriho:
Sounds good

Den fre 12 maj 2023 12:49b21oscpe ***@***.***> skrev:

PR #13968, opened 2023-05-12
State without attributes displays incorrectly #13961

b21oscpe:
### What's been done

Removed hardcoded "Do:" string which was appended to the empty textarea, this lead to it "creeping from below" as described in the issue


b21oscpe:
This PR can be closed, i have made a new one with the correct base branch

Issue #13965, opened 2023-05-12
Pressing ctrl+Z unselects the element placement tool


b21eriho:
When one presses undo while having a tool selected that places an element the tool will be unselected. Meaning you will have to select the tool again. This should not happen, resolve the issue

PR #13969, opened 2023-05-12
Ctrl+z when placing elements #13965

a21oligu:
Fixed an issue where ctrl+z when placing elements would cause the ghost to be removed, hence making it impossible to place another element without reselecting it in the toolbar. Solves issue #13965


b21oscpe (approved):
Looks good, ghost no longer disappears

PR #13970, opened 2023-05-12
Update class diagram with input data in methods#13757

b21tilbe:


a21oscgu:
Three different files have been written as class diagrams. Only the service-file is relevant, right?

a21oscgu (approved):
Looks good!

d21erisu:
great

Issue #13958, opened 2023-05-11
Not possible to rename actors and objects


b21eriho:
It is not possible to rename sequence lifeline actors nor object, figure out why and resolve the issue

PR #13971, opened 2023-05-12
Not possible to rename actors and objects #13958

b21oscpe:
### What's been done

Changed the default value `element.type` from "sequence" to "SE" for diagram.js and added the remaining implementation for SE in the option-pane


b21willu:
It works, good job!

Issue #13945, opened 2023-05-11
Further work on sequence activation


b21eriho:
Make it so that the sequence activation object can be placed on top of sequence lifeline and also make it scalable vertically.

b21matfi:
should you be able to place the object inside on top of the lifeline or is it so that you can drag it

b21eriho:
Ideally it should be placeable on top of the lifeline, does that seem feasible?

b21matfi:
i will try and make it placeable on top of it, but how it was coded it can be hard, but i will try

b21eriho:
Sounds good

PR #13973, opened 2023-05-12
G1 2023 v6 #13945

b21matfi:
could not make it so that you can place the sequence activition on top of the sequence life but you can drag it on top of it. Added so that you can expand or make it smaller vertically.

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -647,5 +647,5 @@
     visibility: visible;
 }
 .mouseData {
-    user-select: none;
+    user-select:text;
 }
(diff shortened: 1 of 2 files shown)


a19clasv (changes requested):
Sequence activation isn't even there, although the changes implemented in the branch are. The change in base branch probably removed it. Sequence activation has to be there before it can work and I can test it.

a19clasv (approved):

a19clasv:
I missed that you have to expand Sequence lifeline and choose Sequence activation. There used to be a separate button for that. Anyway, it's possible to drag it on top and change the vertical length now.

PR #13979, opened 2023-05-12
G2 2023 v6 issue#13927 Create testing database

b21kurar:
This PR adds files that creates a database meant for testing.
`DuggaSys/testingdb/testingdb_setup.php` - Sets up a new database with the username + 'testingdb'. e.g `b21kurartestingdb`. This database is then filled with the data from a fresh install of LenaSYS.

`DuggaSys/testingdb/testingdb_queries.sql` - Dump of a freshly installed LenaSYS configuration.

The script also updates `coursesyspw.php` to add a constant containing the name of the testing database. If the constant is already added, the script will not append anything to the file. 

To test:
1. Before running  `testingdb_setup.php`, ensure that the database is not existing (no database named `****testingdb` ) and that `coursesyspw.php` doesn't contain any other defines than the four from a fresh install.
```php
// coursesyspw.php

    define("DB_USER","usr");
    define("DB_PASSWORD","pwd");
    define("DB_HOST","host");
    define("DB_NAME","dbname");
```
Then check after running the file that there is a new database `*****testingdb` with 44 tables containing data from a fresh install of LenaSYS. Also check that `coursesyspw.php` has been updated

```php
// coursesyspw.php
<?php
    define("DB_USER","usr");
    define("DB_PASSWORD","pwd");
    define("DB_HOST","host");
    define("DB_NAME","dbname");
    define("DB_TESTING", "dbnametestingdb"); ?>
```

2. Make sure that the database is reset when existing by making queries to the database and then running the file. It should then return back to 44 tables. 

3. Make sure that `coursesyspw.php` is only updated once, and that the define isn't readded every time it's run.

The file may take a while to finish up, so it may look like it's frozen. Please be patient


b21kurar:
For some reason the include path gets messed up if `coursesyspw.php` is removed from `LenaSYS/coursesyspw.php`, meaning that the `database.php` can't use it's include in line 3. I don't know how this works, and quite frankly am going slightly insane, so I will leave it as is since it's working.

PR #13980, opened 2023-05-12
Add functionality for sequence loop or alternative #13914 #13949 part 2

a21aldsa:
This is a continuation of #13949 since the issue leader accidentally merged that unfinished PR into v6.

Here's what both of these PRs is about though:

I created a new element type called sequenceLoopOrAlt and made the svg for this. I also added the functionality of adding/removing alternatives via a **textbox in the context panel**. These alternatives adjust themselves accordingly to fit and also the whole element changes size to fit them. I also made sure to call state machine so undo/redo should work, although if it doesnt I believe it is a new issue.


b21matfi (approved):
Looks Good!

PR #13981, opened 2023-05-12
G2 2023 v6 issue#13638

b21ricni:
Edit title by clicking on it and then clicking away, still needs to make "press enter" to work but that is not part of my problem


b21hamwe (approved):
It is indeed working now!
Good job b21ricni.

Issue #13991, opened 2023-05-15
Implement the new test-cases


b21eriho:
See #13917 

A new design for test-cases has been developed, implement it.

PR #13993, opened 2023-05-15
G1 2023 v6 #13991

d21matbo:
If a transition has a label it will now show up like the following:
> "A" goes to "B" with lable "Test"

and if the label has square brackets like `[Test]` it will show up as the following:
> "B" goes to "C" with guard "[Test]"


c21adano (approved):
Works! adding labels or [] will show in the test case window

PR #13996, opened 2023-05-15
[NOTE: This is the PR for "v6-fix-3"] State without attributes displays incorrectly

b21oscpe:
Removed the "Do:"


a21aldsa (approved):
Looks good!

PR #13998, opened 2023-05-15
 Add functionality for sequence loop or alternative #13914 new

a21aldsa:
Since the first branch had so many issues, this is the new branch that will be merged into fix3 of v6. see #13980 for explanation on what this new element is and can do.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -9828,7 +9828,7 @@ function drawElement(element, ghosted = false)
         }
 
         let splitAlternatives = element.alternatives.map(line => {
-            return splitLengthyLine(line, maxCharactersPerLine);
+            return splitLengthyLine(line, maxLengthOfAlt);
         }).flat();
 
         //elemAlternatives = text.length;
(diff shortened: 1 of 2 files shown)


a19clasv:
It doesn't change the horizontal size if the new alternative is too long for its current size and letters get hidden if it's too narrow, but otherwise it works

a21aldsa:
> It doesn't change the horizontal size if the new alternative is too long for its current size and letters get hidden if it's too narrow, but otherwise it works

Fair, I'll implement a line break if the length of the text is suspected to be longer than the width of the element

a19clasv (approved):
I'm approving this because my suggested fixes aren't completely necessary and it works fine.

a19clasv:
Let me know when you're done pushing commits.

a21aldsa:
> Let me know when you're done pushing commits.

Done now, I ultimately decided this was not needed right now. I can be made into its own issue if need be.

a19clasv (approved):
Something hasn't broken since the last time I approved the changes. Maybe a separate issue should be made to finish off the horisontal text size issues.

b21eriho:
For tester, ensure the buttons in the pop-out work as these also had conflicts

a19clasv (approved):
Pop-out buttons work as well.

Issue #13988, opened 2023-05-15
Changing icons for IE-entities throws TypeError


b21phiro:
## Description
When setting an icon to the line for IE-entities a ``TypeError`` error is thrown, and the icon is not displayed.

Figure 1.
<img width="523" alt="image" src="https://github.com/HGustavs/LenaSYS/assets/45258330/b214380f-ff97-4a20-9c32-a63e5d24d30e">

## Your job
Find out why and try to fix it.

PR #14006, opened 2023-05-15
Changing icons for IE-entities throws TypeError #13988 

a21aldsa:
Added checks for IE where they were missing, on top of the previous checks for UML etc.


b21matfi (approved):
Works to change now!

Issue #13952, opened 2023-05-11
Border-radius is different on buttons for dark/light theme


b21phiro:
## Description
The buttons in the options panel does not have the same overall-look between light/dark theme. For example, the border-radius is different.

![bild](https://github.com/HGustavs/LenaSYS/assets/45258330/e3de76a5-3089-4650-8ef6-37fd69a0386f)

## Your job
Fix it.

PR #14008, opened 2023-05-15
Border-radius is different on buttons for dark/light theme #13952

a19clasv:
Removed it from blackTheme.css so it's the same as in style.css for light theme.


a21aldsa (approved):
Looks good

Issue #14007, opened 2023-05-15
Add popup-message for saved diagram


b21phiro:
## Description
Right now when a diagram is saved it does not give any feedback to the user. 

## Job
Make sure that a popup that displays if saving the diagram was successfull or if it failed.

PR #14015, opened 2023-05-15
G1 2023 v6 fix 3 #14007

b21matfi:
I made so that there is a message when saving, tried to make it so that you could not save as many times as possible, but there was a issue where if you change the type or placed another element instead of the one already in place you could not save.

That's why i chose to just it to check if the current index is -1 you can not save the diagram


b21oscpe (approved):
Looks good!

PR #14016, opened 2023-05-15
G2 2023 v6 issue#13617

b21hamwe:
Been discussing with G2:s "issue master" (c21alest), and we agreed on letting this issue be for the moment and leaving this solution as an work in progress file until the API and everything is complete.

Therefore, the checkboxes and searchbar are missing functionality.
Styling will later on be more defined.


a21oscgu (approved):
Looks good!

d21erisu:
Great!

Issue #13985, opened 2023-05-15
Fetch saved diagram from "CurrentlyActiveDiagram" localStorage.


b21phiro:
## Description
In the file ``diagram.js`` there's a function called ``storeDiagramInLocalStorage`` that saves the diagram as JSON in the localstorage with the key ``CurrentlyActiveDiagram``. However, this is never retrieved. 

## Job
There might allready exist functions that are commented out that do this. But, either way we need:
1. A function that retrieves the data, and translates the JSON into JS-objects.
2. A function that takes this object, and prints out on the screen.

For now, we also just loads in the saved data on load. However, the function that prints out the data must be dynamic as a "load-button" should be implemented in the future.

a21nibah:
what is the goal here? is the goal to create these 3 functions or is the goal simply to make a function that can retreave diagrams from local storage?
can i take this function adn repurpose it to solve this issue? from the comment it appears to not do anything. 
`
//code has no functionallity execpt for when the hard coded diagram was used. looks like it reloaeded the diagram and got the file. Diagram in onSetup()
/* function fetchDiagramFileContentOnLoad()
{
    let temp = getVariantParam();
    var fullParam = temp[0];
    cid = temp[1];
    cvers = temp[2];
    diagramToLoad = temp[3];
    diagramToLoadContent = temp[4];

    // Check whether there is a diagram saved in localstorage and load it.
    // Otherwise, load from VariantParam  
    if (localStorage.getItem("CurrentlyActiveDiagram")) {
        var diagramFromLocalStorage = localStorage.getItem("CurrentlyActiveDiagram");
        loadDiagramFromString(JSON.parse(diagramFromLocalStorage));
    } else if (diagramToLoadContent != "NO_FILE_FETCHED" && diagramToLoadContent != "") {
        loadDiagramFromString(JSON.parse(diagramToLoadContent));
        storeDiagramInLocalStorage();
    } else {
        // Failed to load content
        console.error("No content to load")
    }
} */
`

PR #14017, opened 2023-05-15
 renamed function in diagram.js to loadDiagramFromLocalStorage. made …

a21nibah:
repurposed function in diagram.js to now load diagrams from local storage. see issue #13985


b21matfi (approved):
Great job it works, i tested by placing some elements and then saving, in console i typed loadDiagramFromLocalStorage("CurrentyActiveDiagram") and it loaded the saved file!

Issue #13875, opened 2023-05-09
Code example locking for github files


HGustavs:
If a code example comes from the github repo there is a certain locking behavior that is important

I have updated the web programming course as a reference case that we can make use of 

https://github.com/LenaSYS/Webbprogrammering-Examples/tree/master

1. It should be possible to hide the example in the course view (if we do not want to show all examples)
2. It should be possible to reorder examples within a heading but not between headings
3. When an example is removed from github it should be hidden in the course view and not deleted
4. We should be able to rename examples in the course view at will.

If one has connected this folder to a course heading (for example called Canvas Examples or something similar)

https://github.com/LenaSYS/Webbprogrammering-Examples/tree/master/Examples%20Canvas

Canvas Animate.html
Canvas Draw Objects.html
Canvas Gradient.html
Canvas Immediate.html

We should **not** be able to move Canvas Animate.html to some other heading
We should be able to move Canvas Immediate.html to the first spot instead of Canvas Animate.html
We should be able to hide Canvas Gradient.html
We should be able to rename the examples at will. It must be possible to rename Canvas Draw Objects.html to Canvas Drawing (without the .html and other changes) but when the files are updated this connection should remain.

PR #14031, opened 2023-05-16
G3 2023 v6 issue#13897

e21krida:
What we've done is created a dropdown menu within the github buttons for each moment in sectioned.php
image
image
With this dropdown menu, you can traverse the webserver repostitory for all directories under the courses/{course id}/Github path containing the string "Examples" (This is done as the filtering for downloading files to webserver is not complete yet as far as i'm aware). When you choose one of them, this is supposed to call the automatic creation of course content via issue #13549, but this might not be testable for now. 

This also updates the database table "listentries" with a newly added column "githubDir" with the chosen directory. It will look something like this.

image

To test this, open a sample course with folders within the courses/{course id}/Github folder that match the syntax of containing the string "Examples", these should then come up within the dropdown-menu. Choose one, then check that the "listentries" table has an entry with the correct course id and a value matching the chosen directory within the "githubDir" column. 

A working example exists on my branch, under the "Webbprogrammering" course. There you can see the syntax checked for right now, if you fail to create the environment to test this, simply remove the filtering done on line 613-615 here:

`if(strstr($dirname, 'Examples')) {
	echo "<option value='$dirname'>$dirname</option>";
}		`

And simply keep the line "echo `"<option value='$dirname'>$dirname</option>";"`

Relevant files: sectioned.php

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -550,6 +550,28 @@ function sendValue(moment) {
   xhr.send("momentid=" + encodeURIComponent(value));
 }
 
+function sendValue(moment) 
+{
+  let lid = moment.getAttribute("value");
+	var dataCheck;
+	$.ajax({
+		async: false,
+		url: "../LenaSYS/sectioned.php",
+		type: "POST",
+		data: {'lid':lid},
+		success: function(data) { 
+      alert(data);
+			dataCheck = true;
+		},
+		error: function(data){
+      alert(data);
+		 	dataCheck = false;
+		}
+	});
+  console.log("ajax done" + lid);
+	return dataCheck;
+}
+
 
 // Displaying and hidding the dynamic comfirmbox for the section edit dialog
 function confirmBox(operation, item = null) {
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -553,8 +553,8 @@ function showSaveButton() {
 function sendValue(lid) {
   console.log(lid);
 	$.ajax({
-		url: "sectioned.php",
-		type: "POST",
+		url: "../DuggaSys/sectioned.php?lid=" +lid,
+		type: "GET",
 		data: {lid: lid},
 		success: function(data) {
       console.log(data);
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -474,6 +474,27 @@ function refreshGithubRepo(courseid)
 }
 
 //----------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


c21sebar:
I get an entry in the db with the name of the selected folder, I can update and change the selected folder to another name. It does what it is supposed to do. But I think filepath would be better to use since the auto-create function would rather use a filepath to a dir to start from rather than a name of a dir. 

And for a UX design it would be nice for the selected item in the list to show when clicking the button again rather than default to the first item in the list. Sort of as a confirmation if one were to forget wich dir is linked to the moment and as a confirmation that the dir is linked to the moment.

I was under the impression that the folder to select on the moment setting github button we were going to choose wich directory/folder that contains all the relevant files to be code-examples.
Now we can only select folders that match the exact string of "Examples" that lies within the /Github directory and not in any subdir "further down". But maybe this is intended?

image

We won't be able to select for instance "Demo" that contains relevant directories/files to be codeexamples for the Demo-moment 
image

I think it would be better to be able to select any directory without any filtering by name,

c21sebar:
> 



> @c21sebar
> 
> > Now we can only select folders that match the exact string of "Examples" that lies within the /Github directory and not in any subdir "further down". But maybe this is intended?
> 
> this is how it should be https://github.com/LenaSYS/Webbprogrammering-Examples/tree/master/Eamples%20Service%20Call you should not be able to see the "old demo" as it will be used as a "trashcan" according to the customer
> 
> only all theese directories in the root
> 
> https://github.com/LenaSYS/Webbprogrammering-Examples/tree/master
> 
> this is from the customer issue #13875

Im not sure I follow, @d21emijo 
This issue is to store a filepath to a directory that holds all the relevant data/files/dir for all the codeexamples connected to that moment. 
What I can tell from the repo you linked there are three dir with the name Examples, are those three different moments within a course? Moment Html, Moment Canvas and Moment Sevice call. 
So everything related to a moment should be in a dir called something with Examples and just throw every file in there?
And if so: within Examples Html there are a bunch of files, how do we know and set wich ones are connected to each other for a code example?

What we did in issue https://github.com/HGustavs/LenaSYS/issues/13549 is take all the files needed for a codeexample put them in a driectory so the filenames can be whatever to describe the file. If there are multiple codeexample within a moment those should be in different dir next to each other within that moment dir, that this issue is about and stores it in db so the auto create code example have a place to start from.

c21sebar (approved):
This works and add and updatde into the correct column the correct name of chosen directory
To fix the ideas about UXD for wich selected directory and a solution to not use hidden input is done in separate issues.

b21sebgr (approved):

PR #14034, opened 2023-05-16
getElementLines now works for UML, ER, IE and SD

c21adano:
getElementLines used to only work for ER this seems to be due to how the other types tops are defined. element.top.concat was undefined for the other types but changing top with left, right or bottom worked just fine, so i changed element.top.concat to element.bottom.concat and it should now work for all types.


a21aldsa (approved):
Works fine, tested with all sides of all different element types.

Issue #14014, opened 2023-05-15
Put ER-attribute in ER-relation toolbox.


b21phiro:
We need to have more space in the toolbox, put the ER-attribute in ER-relation toolbox.
![bild](https://github.com/HGustavs/LenaSYS/assets/45258330/62146d60-0fdf-4f47-9439-e2a6263cb053)

PR #14036, opened 2023-05-16
 Put ER-attribute in ER-relation toolbox. #14014 

a21aldsa:
It is now in the pop out together with the relations.


c21adano (approved):
attribute is now part of the pop out.

PR #13918, opened 2023-05-10
G2 2023 v6 issue#13910 Change sectionedservice-test.txt to MD and add expected output

a21oscgu:
- Converted file to Markdown .md
- Added expected output
- Deleted old file (txt)

PR #14040, opened 2023-05-16
Added expected outputs#13909

b21tilbe:
Added outputs and changed to MD

### DIFF
--- a/backend-models/service-tests/profileservice-tests.md
+++ b/backend-models/service-tests/profileservice-tests.md
@@ -1,7 +1,5 @@
 profileservice-tests.txt
 
-
-
 ------------------------------------------------------------------
 Row : 46 Check if it blocks teacher, send parameters as an teacher
 ------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


backend-models/service-tests/profileservice-tests.md
```diff
@@ -27,7 +25,12 @@ send {
 	$action = "challenge";
 }
 
+*********
+Service output-
+*********
 
+Save the values in encoded JSON string (for example '{"success": true, "status": "", "debug": "NONE!"}')
```
a21oscgu, line 32: Why is success set to true while it is not like that below?

a21oscgu, line 32: Output should look like this [example1](https://github.com/HGustavs/LenaSYS/pull/13918/files) or like this [example2](https://github.com/HGustavs/LenaSYS/pull/13924/files)

a21oscgu (changes requested):
Read the comments.

Issue #14038, opened 2023-05-16
Clipping outlines for superstate and sequence activation


b21eriho:
The outlines for these two entities are clipped, please resolve this

b21eriho:
![image](https://github.com/HGustavs/LenaSYS/assets/129077727/3d127700-ce88-4778-8d6b-4ad0bbc007a1)

PR #14043, opened 2023-05-16
 Clipping outlines for superstate and sequence activation #14038 

a21aldsa:
fixed the clipping issue.


b21eddka (commented):
The clipping issue is resolved but this change leads to the super state nodes for changing size not working as intended.

a21aldsa:
> The clipping issue is resolved but this change leads to the super state nodes for changing size not working as intended.

fixed

b21eddka:
Even though the super state still behaves abit weird when using the vertical nodes, i would approve of this pull request and create a new issue regarding this

Issue #14045, opened 2023-05-16
Allow display of saved diagram when loaded


b21phiro:
## Description
We have a load button right now that is not hooked up to anything. However it should load a save and display it from localstorage.
<img width="393" alt="image" src="https://github.com/HGustavs/LenaSYS/assets/45258330/a3a7c0bc-ca1b-4e3d-a311-14b863ab2c98">

## Job
Using fetched JSON from localstorage, display the content on the diagram when using the load-button.

PR #14046, opened 2023-05-16
 Allow display of saved diagram when loaded #14045 

a21aldsa:
made it so the button calls loadDiagramFromLocalStorage onclick


b21eddka (approved):
Works as intented, approved.

PR #14053, opened 2023-05-16
G3 2023 v6 issue#13939

a21hugni:
Changed the color of the cogwheel and trashcan so they are visible in light mode. 
To test this navigate to a course in courseseed.php and create a new heading using the fab-button. Switch dark and light mode and see if the icons are visible. You can also look through the code to see that it does not conflict/override any other styling. 

image

Worked with @b21geoyo on the some of the latest commits.


a21lukpo (approved):
Everything works correctly

Issue #13456, opened 2023-04-20
The refresh button on moment type needs a URL to function


c21sebar:
For issue:
- https://github.com/HGustavs/LenaSYS/issues/13431

From issue :
- https://github.com/HGustavs/LenaSYS/issues/13450

Save the info to db which directory is chosen in the dropdownlist, so other functions know where to look.

RaymondSWE:
This issue is to store the users selected option (for example, a specific folder) so that when the "refresh" button is clicked, only the chosen folder or file gets downloaded or updated.

PR #14056, opened 2023-05-16
G3 2023 v5 #13456

a21rammo:
# Update Selected Directory implementation

### Summary of implementations:
* Validations checks on the form inputs in Javascript. Currently only validations for the selectedDirectories is checked no other inputs are checked.
* An AJAX post request sent to sectioned.php with the 'selectedDir' and 'cid' as parameters
* PHP code to update selectedDir in the SQLite database when a form is submitted

### Needs to be fixed in the future
* **Code Cleanliness:** The current implementation includes console logs for debugging purposes. These will need to be removed once i figured out how it should integrate with other components such as code example, refresh etc.
* **Error Handling:** While there is basic error handling in place in both the AJAX error function and the PHP database, more detailed error handling should be implemented to cover a wider range of issues.
* **Directory Fetching:** I noticed that the current implementation does not have a way to fetch directories from the database. For this implementation, I had to manually insert directory data into the SQLite database. In the future, there should be a functions that automatically fetches directory data.

How to insert in to sqlite. Go into terminal the folder is
```ubuntu@webug-project-2023:/var/www/html/root/G3/students/a21rammo/githubMetadata $```

type:```sudo sqlite3 metadata2.db```

here is an example of insert: ```INSERT INTO gitFiles (cid, fileName, fileType, fileURL, downloadURL, fileSHA, filePath) VALUES (1896, 'blabla', 'dir', 'ga', 'g', 'g', 'h');```

**This does not have the automated remove metadata2.db, you need to remove it manual and redownload to fetch new columns**

issue: https://github.com/HGustavs/LenaSYS/issues/13456


a21rammo:
The issue here is that we have two duplicates of the same problem. We will proceed with the current implementation of saving selected directory from the weekly branch, but will keep some functions used to save in sqlite3 from my branch. The approach of the current weekly branch is to store the directory in lidentries, while my approach is to save it in gitFiles in metadata2.db. If the function updatedSelectedDir from my branch will never be used it can be deleted later on, but for now it will be kept.

e21krida (approved):
everything looks to be working as intended

Issue #14042, opened 2023-05-16
Allow multiple saves in localstorage


b21phiro:
## Description
Right now we save JSON in localstorage using the key ``CurrentlyActiveDiagram``. However only one save can be done each time.

## Job
Make so (for example an array) of JSON-object that we can save.

## Suggestion
Maybe also make a limit of how many saves can be done, so we don't store too much,

a21nibah:
do we only want to save diagrams to local storage that is currently being worked on or so we rather make it more flexible so that any diagram can be saved? this is important to know because there are diferent aproaches depending on when diagrams should be saved. If we want multiple diagrams to be saved, if we only want to save the diagram that a user is currently working on and allowing multiple of those saves I would suggest we use an aproach where we modify the keys so that each key is unique with a date and timestamp. But if the goal here is to allow saves at any time using any key i can make it so the function accepts parameters, that would also not break the rest of the code. But the user would then have to make sure the keys are unique.

a21nibah:
do we only want to save diagrams to local storage that is currently being worked on or so we rather make it more flexible so that any diagram can be saved? this is important to know because there are diferent aproaches depending on when diagrams should be saved. If we want multiple diagrams to be saved, if we only want to save the diagram that a user is currently working on and allowing multiple of those saves I would suggest we use an aproach where we modify the keys so that each key is unique with a date and timestamp. But if the goal here is to allow saves at any time using any key i can make it so the function accepts parameters, that would also not break the rest of the code. But the user would then have to make sure the keys are unique.

PR #14064, opened 2023-05-17
G1 2023 v7 #14042

a21nibah:
solves issue #14042 I have added a parameter to allow for other keys. it does not make sence to really make it so you can send in an array since you most likely will only be saving a diagram at once. I have asked in the issue if we should have key generation be automatic so that we generate a key according to timestamp but got no reply so that can be a separate issue if we want to have automatic key generation.
i have also fixed so that the function calls use the "CurrentlyActiveDiagram" key


d21matbo (changes requested):
The changes does not work for the "save current diagram"-button in the toolbar.

I believe that this error occurs due to the button not using any argument in its function call.

d21matbo (changes requested):
Think something went wrong with the quotation marks, as it now throws the error:

Uncaught SyntaxError: Unexpected end of input (at diagram.php:844:144)

a21nibah:
ok see if that works. works fine here.

d21matbo (approved):
The changes made have fixed the issue, looks good!

Issue #14027, opened 2023-05-16
Not all toggable-buttons have indication that they are active


b21phiro:
## Description
Some buttons in the toolbox have a highlight when active. But some buttons don't, such as ER-tables. 

Figure 1.
![bild](https://github.com/HGustavs/LenaSYS/assets/45258330/b7aa8c7f-4c6c-4cdf-9383-d34e5177cc32)

## Job
Make sure all buttons are highlighted when active. Bonus if you can make the highlight more visible, such as making the background color/border-color into "action" color (maybe the color the options panel have?).

PR #14073, opened 2023-05-17
Not all toggable-buttons have indication that they are active #14027

a19clasv:
Toggable-buttons now indicate that they're active. I didn't make the bonus effort of making the highlight more visible, but maybe that can get its own issue.


b21matfi (approved):
Looks Good!

PR #14074, opened 2023-05-17
 commented out function setPreviewValue in diagram.js

a21nibah:


a21oligu (approved):
I believe the purpose of the function is to get the previewValue from a dugga. The function works when running the "diagram dugga" in demo-courses, but not when diagram.php is run  standalone. Maybe a guard could be added to the function in order to get rid of the thrown error?

I will approve this as it does not break anything and gets rid of the error message.

Issue #14058, opened 2023-05-17
Do not allow changing to type that does not make sense


b21phiro:
## Description
Some elements, such as the SD Final State, can be changed into for example IE. This causes ``TypeError`` because most likely that does not exist in that particular context.

Figure 1.
![image](https://github.com/HGustavs/LenaSYS/assets/45258330/36dfbcda-e125-4dbe-989b-f46c5f9100aa)

## Job
Fix so the dropdown for changing type does not display options that do not make sense.

PR #14075, opened 2023-05-17
Do not allow changing to type that does not make sense #14058

a19clasv:
I made it so only the selected type shows if they have lines.


b21matfi (changes requested):
You have removed so that the relations can not change to other relations for example IE relation can not change to UML relation. It should be that they can change to eachother but they can not change to SE for example. Like in the picture. 
image

Your assignment is to make it so that relations can change to eachother but in their dropdown there should not be an option to change to SE. 

Same goes for other elements, some elements should no be able to change to other types, You can do what you did in your solution here with the array (canChangeTo) and add the different types that they can change to and remove the others from their dropdown.

b21matfi:
made a little miss click this PR is not approved

a19clasv:
SE isn't displayed in the dropdown anymore. I made it so that they can't switch when they have lines connected to them because different diagram types connecting to each other doesn't make sense. Relations not being able to change types have nothing to do with my changes, they, and the others that aren't entity types, don't even have a type dropdown anymore.

b21phiro:
## Basically
The idea is basially this:

We have for example a Initial state placed, along with a state-element. If we now try to change the initial state (using the dropbox). we can select for example ER. However, this will obviously not work and an error will be thrown.

Figure 1.
bild

## The issue
So, the ER should not even apear in the dropbox, as this will not work. Basically, only display options that makes sense.

a19clasv:
I made it so that only the type they are currently will be shown in the dropdown if they have a line connected to them, so your example would have already been solved. I also tested it and it worked for me.

a19clasv:
> Relations not being able to change types have nothing to do with my changes, they, and the others that aren't entity types, don't even have a type dropdown anymore.

a19clasv:
> Relations not being able to change types have nothing to do with my changes, they, and the others that aren't entity types, don't even have a type dropdown anymore.

Someone else did that, probably came with a merge. I see it's there in the group branch, so it might be fixed if this pull request is merged.

a21oligu:
I am responsible for the drop down and entity types being removed from it. I made it so every default entity type (and attributes and so on) has a property canChangeTo with an array of the types it can change to. The initial idea was that the types in this array would be the types showing up in the drop down. However, I was unsure what types to allow or not so I pushed my changes to the branch. It appears though that Clas started working on the branch where I left of, probably due to some miscommunication.

The issue should be solvable by assigning the right types in the canChangeTo array, but I am unsure if the solution is foolproof and actually works as intended.

a19clasv:
The grid disappeared and there's a "d" in the top left corner after the merge.
image
Don't mind the color change, it happened because of the screenshot triggering saving.

a19clasv:
Relation types still don't have the diagram type dropdown after the merge as well.

a19clasv:
The white background was fixed with ctrl + shift + r.

a21oligu:
Commit *aa34516* will not change anything as it was an empty array before. I made it so that if an empty array is assigned, the drop down will contain all entity types by default.

a21oligu:
> Commit _[aa34516](https://github.com/HGustavs/LenaSYS/commit/aa34516222abb652e2eee6afe0c201576ece481f)_ will not change anything as it was an empty array before. I made it so that if an empty array is assigned, the drop down will contain all entity types by default.

The issue right now is that some of the "defaults" has changeTypeTo assigned to `null`. Hence, the drop down will not be generated as it cannot change to any type. I assigned this to the "defaults" that I was unsure about. Giving these types to change to, if they should be able, will solve the issue.

a19clasv:
Relation types have the diagram type dropdown now.

a19clasv:
Done

b21oscpe (approved):
Looks good, can't change from SD initial or final state and inheritance only changes to relevant types. Unless im missing out on something (seeing that a lot of discussion has been had regarding this issue) I'd say it is good.

**To clarify: I have only checked inheritance and initial & final states.**

PR #14076, opened 2023-05-18
G3 2023 v7 issue#13549

c21sebar:
Redone the previous version of creating code-examples to fit the new version of how files,dir and moment is located on the webbserver and repo, see https://github.com/LenaSYS/Webbprogrammering-Examples/tree/master/Examples%20Html for moment "Examples Html".
This now creates a code example based on the file name. If multiple files have same name they are created as the same codeexample. Meaning: 
Canvas Animate.html = 1 code-example
Canvas Draw Objects.html =1 code-example
Canvas Gradient.html =1 code-example
Canvas Immediate.html =1 code-example
And 
Canvas Animate.html + Canvas Animate.css 	=1 code-example

This verision does currently not support the possibility to have two files of the same type in a code example because two files cannot have the same type and name. But maybe two or more .sql, .js, .html or any other files type is never used within the same example so this might not even be an problem? @HGustavs 

You select a dir in the github setting in the dropdown and press submit.
image

Then you press the refreshbutton on that moment.
image

Refresh the page and all code examples is created with their files.
image

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -3086,14 +3086,14 @@ function hasGracetimeExpired(deadline, dateTimeSubmitted) {
 }
 
 //Creates all examples from github that doesnt exists yet
-function createExamples(dir,momentID) {//TODO HERE
+function createExamples(momentID) {//TODO HERE
   lid= momentID;
   dirname = dir;
   console.log("* AJAX START ");
   $.ajax({
     url: "sectionedservice.php",
     type: "POST",
-    data: {'lid':lid,'dirname':dirname , 'opt':'CREGITEX'},
+    data: {'lid':lid,'opt':'CREGITEX'},
     dataType: "json",
     //TODO: reload back to coursepage
   });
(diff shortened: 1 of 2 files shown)


c21sebar:
> This verision does currently not support the possibility to have two files of the same type in a code example because two files cannot have the same type and name. But maybe two or more .sql, .js, .html or any other files type is never used within the same example so this might not even be an problem? @HGustavs

I found a few examples on the course "Databaskonstruktion" where multiple files of the same type is used within a code example. Take https://dugga.iit.his.se/DuggaSys/codeviewer.php?exampleid=7706&courseid=4&cvers=82840 as an example. Here two .cs files and two .cshtml files are used in the same code-example. This will be impossible to autogenerate with this solution since the 2 files cannot have the same name and type and exist within the same directory like https://github.com/LenaSYS/Webbprogrammering-Examples/tree/master/Examples%20Html structure.

We eather need a diffrent approch to naming files within the "moment" folder (https://github.com/LenaSYS/Webbprogrammering-Examples/tree/master/Examples%20Html) to tell what files are related to each other for the same code example, or use the previous approach by selecting all files that should be in a code-example and store them in their own folder, somthing along this: https://github.com/c21sebar/test/tree/main/Demo where "Demo" is the moment and Code-example1 is one code-example "Code-example2" is another, "Examples Test" is a third and so on while each of these folders contain their own files for the code-example. In this way the files can be named whatever and all related files for the code-example is within that specific directory.

c21rebja (approved):
It works like it should. :)
However, it is a little clunky to use. But just like Leo said this can be changed in a new issue.

PR #14083, opened 2023-05-19
New expected outputs from tests profileservice.php#13909

b21tilbe:
Updated


backend-models/service-tests/profileservice-tests.md
```diff
@@ -82,8 +99,17 @@ End pre-
 		$password = "Kong";
```
a19hakja, line 99: One thing i noticed here is that you actually have to provide the wrong password to get a response from profileservice.php, as well as the new password has to fulfill the password requirements (min 8 letters, one lowercase, one uppercase and a number) Which is why providing Kong won't do anything with profileservice.php

a19hakja: The output is correct though, i just couldn't get there the info provided here

a19hakja (changes requested):
Otherwise everything seems fine, just change Kong to something that fulfills the requirements and gets a response from profileservice.php. And maybe include a false password that gets us there

a19hakja (approved):
That did the trick :) I'll approve

d21erisu:
great

Issue #14059, opened 2023-05-17
Name of lifeline elements


b21eriho:
As of now the names of lifeline elements look like this
![image](https://github.com/HGustavs/LenaSYS/assets/129077727/abdba2a9-464f-48cc-be17-6997e0fdfbd9)
The name should have a background so that it is doesn't overlap like it does now. Further the default value should be "lifeline"

b21eriho:
I see now how I was unclear. The name of the actor should have a background so that it doesn't overlap. And since we don't and shouldn't change the name of the entity when we swap between object and actor the default name should be lifeline

a21aldsa:
> I see now how I was unclear. The name of the actor should have a background so that it doesn't overlap. And since we don't and shouldn't change the name of the entity when we swap between object and actor the default name should be lifeline

But the name is for the actor and object, no? Like, the name of the actor or object.

a21aldsa:
So the only thing I should do for this issue is add a background for the text for actor?

a21aldsa:
For documentation: Upon discovering a unexpected error related to how names that are too long to fit and consulting with the group leader about this, we came to the conclusion that there are two ways of fixing this, both of which are part of this issue.

Method one would be to use a similar linebreak that, for example, func uses in UML elements. This means that the name would extend onto a new line whenever its too long.

Method two would be to use a similar method as the name for ER entities; simply have the name cut off into nowhere if its too long.

I will first try to implement method one, but if I cannot manage this, method two was also deemed acceptable.

PR #14085, opened 2023-05-19
 Name of lifeline elements #14059 

a21aldsa:
I added a svg rect that sits behind the actors name, this caused an unexpected bug though where this rect would be immensely offset if the name of the actor was too long to fit the designated box.

As shown in the comments on issue #14059 there was a discussion between me and the group leader about how this problem should be resolved. I ultimately decided on method two: "A similar method as the name for ER entities; simply have the name cut off into nowhere if its too long."

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -9816,7 +9816,7 @@ function drawElement(element, ghosted = false)
                 }
             }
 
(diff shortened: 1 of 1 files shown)


b21oscpe (approved):
Looks good, text cuts off as described
78245ae658f4974f42f53b1038473c3d

PR #14093, opened 2023-05-19
Implement tests for fileedservice - Issue#13849 

b21kurar:
Multiple tests have been added for fileedservice, I couldn't use the documented testfile and had instead decide on values on my own.

Tests:
* Delete file - Global test: deleting a global file
* Delete file codeexample test: try to delete codeexample (shouldn't work)
* Edit file test
* Delete file - Course local file test: delete a local file 
* Delete file - Course version local file test - Delete a course version local file


b21tilbe (commented):
The syntax for all code snippets seems to be correct. But I want to emphasize that I am a bit tired today and could easily have missed something. I'm a bit confused though if it's okay to use your own values instead of using the old test-file.

b21tilbe:
If "the use of your own values" is okay from the person in charge, I approve this issue.

a21jeaha:
when i try to start the test file i get this output.
image

Is it not suppose to look like this?
image

might of course have to do with the fact that the URL are absolute. 
Im going to test around a bit on my end to see if that is the case.

b21kurar:
> when i try to start the test file i get this output. image
> 
> Is it not suppose to look like this? image
> 
> might of course have to do with the fact that the URL are absolute. Im going to test around a bit on my end to see if that is the case.

Hmm.. Interesting...
Also the tests seem to file with no return value from the service, will also investigate

b21kurar:
I have now updated the expected output to contain the entries, and now the test passes

a21jeaha:
hmm.... I'm still only getting a wall of text. And i get the same results no matter if a use 
https://cms.webug.se/root/G2/students/b21kurar/LenaSYS/DuggaSys/tests/fileedservice_test.php
or
https://cms.webug.se/root/G2/students/a21jeaha/LenaSYS/DuggaSys/tests/fileedservice_test.php
image

b21tilbe:
Do you go through Lenasys or directly into the file?

b21kurar:
I would use the developer tools to check the response (nicer JSON notation) , or temporarily changing the file to use pretty print instead

b21kurar:
Inspector tools:
image

Pretty print:
image

a21jeaha:
after changing it to write in pretty print, it seems to work. though i cannot say if it's right or wrong seeing as some of the tests fail, this problem most likely have to do with the absolute paths to the services which makes it so that i could not use my own fresh database.

b21kurar:
> after changing it to write in pretty print, it seems to work. though i cannot say if it's right or wrong seeing as some of the tests fail, this problem most likely have to do with the absolute paths to the services which makes it so that i could not use my own fresh database.

When testing these types of files you should update the service filename to set to your own database, otherwise the testing is completely unnecessary. It is a bit of an annoying workaround however...

a21jeaha:
with a fresh database the i still got quite a few test that fail. 
In the first test this value encounters problems. 

image

this site can help when comparing the JSON strings 
https://jsoncompare.org/

a21jeaha (approved):
after fixes, all test now pass. great work!

Issue #14063, opened 2023-05-17
Segmented SD lines pointing up to long


b21eriho:
A segmented SD-line with the arrow pointed upward looks like the image below at this time
![image](https://github.com/HGustavs/LenaSYS/assets/129077727/b9da3bbd-0b1a-4836-877c-dc22bc2a59a3)
As can be seen arrows pointing down (as well as to the sides) function correctly while the arrow pointing up is stretched. This happens regardless of in which direction the line was originally drawn.

b21eriho:
For tester, please ensure no other lines break as a result of solving this issue

PR #14095, opened 2023-05-19
Stretching arrow on segmented edge #14063

a21oligu:
Solved an issue where segmented arrows would strecth if pointing upwards. The underlying issue is not the arrow itself, but the "hitbox" of the SD entity being to large:

image

I would suggest that the "hitbox"-issue should be fixed in a new issue.

Solves issue #14063

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -8410,7 +8410,8 @@ function drawLine(line, targetGhost = false)
                 }
                 else {
                     if (line.ctype == 'TB') {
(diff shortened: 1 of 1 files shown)


a21aldsa (changes requested):
The arrow is still stretched 

image

a21oligu:
@a21aldsa the problem should now be fixed.

a21aldsa (changes requested):
Now arrows going downward are longer than arrows going upward

image

a21oligu:
made some changes @a21aldsa

PR #14096, opened 2023-05-19
G3 2023 v7 issue#14065

a21lukpo:
We have now added functionality to the moments dropdown were it saves the last selected option in local storage. This can be tested by choosing an item within the moments dropdown and then closing it. Open the moments dropdown once again and the last selected option should still be selected.

image


b21leowa:
The code works but change "let" to "var" for variable declaration and then merge.

e21krida:
> The code works but change "let" to "var" for variable declaration and then merge.

Var has historically had multiple problems with redeclaration, scoping and hoisting and is widely considered inferior to "let", as "let" was a replacement solving the previous issues with "var", why exactly should we replace "let" with "var" here when "var" is widely considered both inferior and legacy?

e21krida:
> Go into the code standard for this project. I agree than "let" is better than "var" but in this project the standard is "var"...

Embarrassingly bad code standard then, i didn't know that, we'll fix it.

PR #14098, opened 2023-05-19
G3 2023 v7 issue#13987

a21rammo:
# Auto-refresh for code examples when user is on course page

This pull request introduces a feature that automatically refreshes code examples when a user is viewing the course page. This PR includes a countdown timer that runs for 10 minutes while the user viewing the course page. Once the timer reaches 0, all code examples for each moment/lecture are refreshed. When the user is not viewing the course page, the timer is paused. This is used by a boolean value 'isActivelyFocued' to ensure if the course page is viewed on by user. Also withthe function from jQuery 'blur' and 'focused' it was simple to implement. 

## New issues to add
* The createExamples function was updated in another issue with different parameters, meaning my function also will need to be updated so it has same parameters as that function
* A toaster/alert needs to be fixed so the user can get a feedback when clicking refresh button
* Perform code cleanup for the createExamples, focus function, and refreshMoment. This involves removing unnecessary console logs and comments (e.g., TODO:: comments).

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -3137,7 +3137,7 @@ $(window).on('focus', function() {
   const now = Date.now();
   if (lastUpdatedCodeExampes === null || (now - lastUpdatedCodeExampes) > updateInterval) {
       lastUpdatedCodeExampes = now;
-      autoRefreshCodeExample(dir, momentID);
+      autoRefreshCodeExample();
   }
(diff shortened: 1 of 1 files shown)


a21rammo:
This will cause a merge conflict with https://github.com/HGustavs/LenaSYS/pull/14076, let me handle it, once it occur.

b21geoyo (approved):
I tested the focus and timer and it looks good

PR #14101, opened 2023-05-19
 moved some p tags in diagram.php

### LINKED ISSUE #14082: HTML semantic violation in diagram.php
## Description
ISome ``<span>`` elements have ``<p>`` tags inside them, which does not make any sense as ``<span>`` elements are inline-tags while ``<p>`` tags are block elements.

```
div id="mouseMode0" class="diagramIcons toolbarMode active" onclick='setMouseMode(0);'>
      <img src="../Shared/icons/diagram_pointer_white.svg" alt="Pointer"/>
      <span class="toolTipText" id="highestToolTip"><b>Pointer</b><br>
          <p>Allows you to select and move different elements as well as navigate the workspace</p><br>
          <p id="tooltip-POINTER" class="key_tooltip">Keybinding:</p>
      </span>
  </div>
```

## Job
Fix so for each elements of the ``toolTipText`` are semantically correct.

a21nibah:
fixes #14082 
make sure that the design is still good when reviewing.

### DIFF
--- a/DuggaSys/diagram.php
+++ b/DuggaSys/diagram.php
@@ -50,8 +50,8 @@
                 <div id="mouseMode0" class="diagramIcons toolbarMode active" onclick='setMouseMode(0);'>
                     <img src="../Shared/icons/diagram_pointer_white.svg" alt="Pointer"/>
                     <span class="toolTipText" id="highestToolTip"><b>Pointer</b><br>
-                        <p>Allows you to select and move different elements as well as navigate the workspace</p><br>
-                        <p id="tooltip-POINTER" class="key_tooltip">Keybinding:</p>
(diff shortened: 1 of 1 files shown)


b21matfi (changes requested):
the p tags are now outside of the span so they show on the whole toolbar. The toolbar can still be used but the tool tip text is always showing.

a21nibah:
any suggestions on how to fix it? I would personally think it would be best to leave it as it was previously since it does not really matter as long as it works.

b21matfi:
Yeah, because the issue is not in the code it self, you can try and just reverting the changes and then change the p tags with span tags

b21matfi (approved):
Looks good!

Issue #14086, opened 2023-05-19
Fix "diagram-toolbar"


b21phiro:
## Description
Inside ``diagram.php`` there's a div element (the toolbar) that is not correct.

```
    <div id="diagram-toolbar" onmousedown='mdown(event)' onmouseup='tup();'>...</div>
```

## Job

1. The ``event`` argument is deprecated and should be replaced according to the MDN specifications. 
https://developer.mozilla.org/en-US/docs/Web/API/Window/event

2. The the callback function for the eventlistener ``onmouseup`` is missing a parameter. Accordning to the function ``tup`` there should be an argument.

PR #14104, opened 2023-05-19
G1 2023 v7 #14086

b21matfi:
Removed parameter on mdown and tup, because event will not work in the future

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1814,6 +1814,7 @@ function mwheel(event) {
 
 function mdown(event)
 {
+    console.log("mdown");
     mouseButtonDown = true;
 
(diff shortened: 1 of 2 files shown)


b21eddka (commented):
When testing i could not get the eventlisterners to react when using the toolbar. I believe mdown and tup are declared wrongly as they are functions that uses parameters. 

After further testing i got the eventlisterners to react when having "onmousedown='mdown(event)' onmouseup='tup()'".
The tup function doesn't use the event parameter so deleting it is _probably_ best, however the mdown function only reacts when using event as a parameter in the eventlistener.

b21matfi:
image
added console log and on tup with parameter and mdown without parameters.

b21eriho:
Another PR has made changes that make this one redundant

PR #14105, opened 2023-05-19
Added functionality for "Load"-button

b21oscpe:
Added a pop-up modal that lets the user select a save from localStorage (Currently a dummy array with fake values)

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -565,20 +565,20 @@
     width: 40%;
     z-index: 999;
     overflow-y: auto;
-    background-color: #eb4;
+    background-color: white;
     color: white;
-    border: 2px solid #000;
+    border: 1px solid #000;
 }
 
 #loadTitle{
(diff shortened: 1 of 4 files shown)


a21aldsa (changes requested):
Looks great except that it does not come up after a user clicks save first.

b21oscpe:
Not sure what you mean by this but if you mean that the saved diagram does not show up if you click save before clicking load; it is because the save button saves your current canvas into localstorage and then (currently all buttons) is linked to the button you are pressing, which means it will load in nothing, as you havent changed

a21aldsa:
> Not sure what you mean by this but if you mean that the saved diagram does not show up if you click save before clicking load; it is because the save button saves your current canvas into localstorage and then (currently all buttons) is linked to the button you are pressing, which means it will load in nothing, as you havent changed

I edited my comment, I'll approve this PR

a21aldsa (approved):

Issue #14049, opened 2023-05-16
Create a "save as" popup


b21phiro:
## Description
Right now we can save and load diagrams in the localstorage. Also, we have a issue on the way that will allow to save multiple diagrams at the same time see:
https://github.com/HGustavs/LenaSYS/issues/14042

What we need is a popup of some sort, that allows the user to save the file with a custom name, so it's easier for the user later to load a particular diagram.

## Job
Create a popup on save, that allows to take in a name for the file. NOTE that you do not have to make the actual function that saves the content (as it's being done in other issue). But rather, make a popup with an input that creates a title that can later be used.

## Suggestion (Not  guide, just a suggestion, feel free to do what you want).
1. Create the popup with an input and a submit button.
2. Make a listener for the submit button that can LATER be used to save the name.
3. If the title is empty, create a default title like "Untitled 1" or something.
4. Allow cancell, to close the popup.

d21matbo:
Just a clarification; This "pop-out" or solution is supposed to be implemented on the current save button / keybinds for saving. Or should this be a completely seperate way to save the diagram?

PR #14106, opened 2023-05-19
G1 2023 v7 #14049

d21matbo:
Saving the current diagram by either 'ctrl+s' or the save button in the toolbar, now brings up a menu to where you can enter a name for the diagram.

image of the save popout, with an inputfield for the name of the save

Here you can either enter a name or leave it empty, and once the save button is pressed the diagram is added to localStorage with the given name.

**Note:** Leaving the input field empty is the only way to load that specific diagram. This is because it is currently not possible to load diagrams by name.


a21oligu (approved):
Works as described, good job!

Issue #13880, opened 2023-05-09
Github Synchronization options


HGustavs:
We want a way to change settings for github synchronization for example if we wait 10 minutes to re-check the github hash to see if repo has changed or if we want to wait 30 minutes.

Also we want a way to oversaturate github if a user re-clicks the refresh button/buttons.

If a user clicks the refresh more than (for example) once in a 10 minute interval the second and third refreshes should give some kind of feedback to the user that the refresh has been triggered... but nothing should happen until 10 minute marker.

The automatic and manual synchronization options should have individual settings for each... for exampel by changing a line at the top of one of the files. No need to have this setting in the user interface.

Are there any other relevant synchronization settings?

PR #14107, opened 2023-05-19
G3 2023 v6 issue#13999

c21rebja:
This issue was to stop a user from spamming the refresh button in the navbar. 
Every time a user presses the refresh button, it now compares with the last time the course was updated (which is saved in MySQL). For a user with "superuser" privileges, the wait-time between refreshes is 5 minutes. All other uses have a wait-time of 10 minutes. 

To test this, enter a course of your choice (that has a github link) and press the refresh button in the navbar. It should say "The course is up to date" or similar if there are no new commits. Press it again, and it should say to wait before refreshing again. 
Wait 5 minutes and try again to see if the course can be refreshed again. 

image

Right now, the updated time is saved even if the course isn't updated, since that is how we interpreted what @HGustavs said. This means the user has one refresh available every 5 minutes. Since a course counts as being updated when it's created, this also means that the user has too wait 5 minutes after creating a course before they can update it. 
The original issue: #13880 

co-author @g21emmka


a21hugni (approved):
Looks good!

Issue #14025, opened 2023-05-16
Create notation for "note"-element


b21eriho:
See issue #14024 for the details of the note element. Please create a notation that would fit the visual style of the existing system.

PR #14111, opened 2023-05-22
Added note element

c21adano:
Added a new element called note, the note exist as its own type of element and can be connected with any other element type. 
lines connected with the note will be dashed, writing more than 4 lines inside the note will make it bigger, changing the colour of the note to black will make the outline white and vice versa. 

Notation for note: #14025

Image of how the note element looks
notefinishedproduckt

The note button is currently under the sequance pop out and is keybinded to "n"

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -343,17 +343,6 @@
     background-color: #FEEFB3;
 }
 
-#zoom-message-box{
-    width: 75px;
-    left: 100px;
-    bottom: 5px;
-    color: white;
-    border-radius: 5px;
-    text-align: center;
-    z-index: 100;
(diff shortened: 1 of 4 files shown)


b21matfi (changes requested):
image
Everything looks fine but when you press save without chaging the type it gives an error

c21adano:
Cant change type from x to x seems to be a problem for all element types and not just note, therefore i would argue that its not part of this issue.

b21phiro:
Indeed, this is an issue that can be reproduced with other elements and should be looked into (if time is on our side).

b21matfi (approved):
The error message is part of this issue. Everything works well Good job!

Issue #13959, opened 2023-05-11
Redesign of refresh button


a21olljo:
After asking some other groups, if the refresh button where descriptive in its design. It was concluded that it wasn't. 
A redesign that more clearly shows what the button does, would be helpful. The new design should more clearly show what the button does from the users point of view, which is to download updates from GitHub.

c21rebja:
Made a few examples for what the new refresh-button could look like. Asked around a little and the consensus seems to be that 7 or 8 (or a combination of both) is the way to go. It is both because it's more generic (if repos on for example gitlab that will be used on lenasys in the future) and in case the github octocat is protected by copyright.

![github icon examples](https://github.com/HGustavs/LenaSYS/assets/102578746/732ec61a-1d86-40a9-88ed-a8190c60a186)

c21rebja:
Here are a few more versions of number 8.
![image](https://github.com/HGustavs/LenaSYS/assets/102578746/1162a190-864f-4222-a7ee-1b9c2650273a)

c21rebja:
Here are a few more versions of number 8.
![icon_ex_2](https://github.com/HGustavs/LenaSYS/assets/102578746/c260e3b9-e4d7-40bc-8ff4-6334c5d5fd26)

c21rebja:
Do you have any opinions on which would be best? @HGustavs 
Not just the versions of number 8, but out of all of them. :)

They are for the button in the navbar that is currently a refresh. 
![image](https://github.com/HGustavs/LenaSYS/assets/102578746/a3325758-a0c7-4581-a2a3-20ab9714fb1d)

HGustavs:
Perhaps something generic such as 8a or 8d is the best if we for example also allow support for e.g. gitlab also.

c21rebja:
Based on the feedback, here are mockups for how it would look on LenaSYS with these icons implemented. 
Icon 8:
![8](https://github.com/HGustavs/LenaSYS/assets/102578746/f5d9df2b-bc32-4058-a4ed-32426ad1a808)

Icon 8a:
![8a](https://github.com/HGustavs/LenaSYS/assets/102578746/71cf1d08-35cc-4657-b9ef-abb5da656703)

Icon 8d:
![8d](https://github.com/HGustavs/LenaSYS/assets/102578746/25ebb17c-798d-42d7-8a76-5f7e9e1c9198)

PR #14124, opened 2023-05-22
G3 2023 v7 issue#13959

c21rebja:
issue #13959

This issue was to change the refresh icon to something that made more sense. So the refresh-arrow has now been replaced with a git-download icon. Only the icons in the navheader and the hamburger-menu has been changed.

To test this, just enter any course and look in the navheader. The icon should have changed, be the same size as all the others and be the correct colour (white). Check that it works in both light and dark mode.

image
image

co-author @g21emmka


e21krida (approved):
Tested, works perfectly 👍

PR #14125, opened 2023-05-22
G3 2023 v7 issue#14021

a21rammo:
This PR handles where the manual order selection for code examples was found not functional due to our github integratioin (i think).  I decided to comment out the code for those function instead of removing them completely. In order if a  reimplementation is needed, it should take secounds. PS I wrote wrong issue number on my branch

**Please note that no code was removed in this PR.**


b21leowa (approved):
I approve these changes! Everything works as intended and it's good that you commented out the code instead of removing the code completely. Therefore it is easier to reimplement as you said.
:+1:

PR #14127, opened 2023-05-22
Added column for separate filename in lenasys for gitFiles

a21henhe:
The table gitFiles now contains a new field which is to be used for the actual name displayed in LenaSys, since the feature to be able to change the name which is displayed was to be added and the previous fileName is a primary key.

Co-authored by b21leowa


a21rammo:
works.

Issue #14089, opened 2023-05-19
Even more deprecated arguments


b21phiro:
## Description
In ``diagram.php`` there's use of the deprecated ``event`` argument for eventlisteners.
```
    <div id="container" onmousedown='mdown(event)' onmouseup='mup(event)' onmousemove='mmoving(event)' onwheel='mwheel(event)'></div> <!-- Contains all elements (items) -->

```

## Job
Change it according to the MDN specs.
https://developer.mozilla.org/en-US/docs/Web/API/Window/event

PR #14128, opened 2023-05-22
 Even more deprecated arguments #14089 

a21aldsa:
Replaced all HTML type adding of event listeners like "onmousewheel...." that used event as argument with javascript addEventListener, thereby removing use of the deprecated 'event' argument and also replacing use of the deprecated 'mousewheel' with the proper argument; 'wheel'.


d21matbo (approved):
Code looks cleaner and works aswell, good job!

Issue #14068, opened 2023-05-17
No feedback for "reset view" button


d21matbo:
Currently if you press the "reset view" button while no elements exist absolutely **nothing happens**.

![image](https://github.com/HGustavs/LenaSYS/assets/102599161/f5824601-a92d-4c67-b2da-2ec75622f3ad)

Instead of not doing anything, this should show the user an error that explains why nothing happened (i.e. nothing to centre the camera around).  *Something like this image below*.
![image](https://github.com/HGustavs/LenaSYS/assets/102599161/bfc2bf46-6c37-4d95-a370-6bd48827ad9a)

PR #14131, opened 2023-05-22
G1 2023 v7 #14068

b21matfi:
Added a display message so that it gives and error if there are no elements to be centered and then also added a success message. Removed duplicate from diagram php where it ran the function twice


c21adano (approved):
The center function works great and a error message appears when there are no elements. There are two error messages that appear but from my understanding this is cuased by how the center function actually works and it having the flaw of needing to be called twice which i think should be its own issue for the next generation.

PR #14133, opened 2023-05-22
Removed filter filtering out maps that doesn't contain the keyword "Examples"

e21krida:
Now the dropdown for choosing a directory after downloading form Github does not filter out for maps only containing the keyword "Examples", so it should now display all maps within the given directory and look something like this

image

To test this, simply open sectioned.php, click on an individual github button on a moment to open the dropdown and see if the dropdown contains multiple options not containing the keyword "Examples"


b21leowa (approved):
Works as intended. I can now add more directories to different moment.

:+1:

PR #14140, opened 2023-05-22
Box selection line is white in darkmode

b21eddka:
The box-selection lines are white when the user is in darkmode.


b21matfi (approved):
Nice work!

Issue #12921, opened 2023-03-31
Priorities between microservices


HGustavs:
We need to prioritize the different services and start with the most important services that are most crucial to the workings of the system as a whole.

The first service that is converted can/will be used as documentation of the conversion process.

c21alest:
Use documentation of available ms: https://github.com/HGustavs/LenaSYS/blob/g2-2023-v6/backend-models/microservices/duggaSys%20services/Potetial_micro_services.md

PR #14141, opened 2023-05-22
G2 2023 v7 issue#12921 priorities microservices

b21kurar:
This PR closes issue #12921 
A priority list for the microservice transition of services


a21ponkr (approved):
Looks like a good approach

Issue #14135, opened 2023-05-22
Line kinds not working


c21adano:
In the options for lines there are diffrent kinds that can be selected, this does not work. The line can't be changed to any other kind and is stuck on normal.
![fast](https://github.com/HGustavs/LenaSYS/assets/102599332/4e60d892-f5c1-409d-9c59-81447d75251e)

PR #14142, opened 2023-05-22
Line kinds not working #14135

b21oscpe:
It is now possible for ER-entities to switch between single and double-kind lines.

NOTE: Recursive does not work and im not quite sure how it is supposed to work either since a line can't be drawn to itself anyways.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -3090,19 +3090,19 @@ function changeLineProperties()
         displayMessage(messageTypes.SUCCESS, 'Successfully saved');
     }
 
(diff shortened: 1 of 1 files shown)


c21adano (changes requested):
The dashed line kind that class diagrams use does not seem to work still. Also changing IE line kind to double breaks the system.

b21oscpe:
I have now made changes so that it works for UML dashed, after investigating IE I found that it is outside the scope of this issue as it has to do with the function that handles drawing the selectionbox. I do not know how to fix that.

c21adano (approved):
UML dashed and ER double kind now work! Great work.

PR #14152, opened 2023-05-22
G2 2023 v7 issue#13955

b21ricni:
Created new microservices and created a guide


DuggaSys/microservices/shared_microservices/getUid_ms.php
```diff
@@ -0,0 +1,28 @@
...
+            $userid=$_SESSION['uid'];
+        }else{
+            $userid="1";
+        }
+
+        $log_uuid = getOP('log_uuid');
+        $log_timestamp = getOP('log_timestamp');
+
+        $log_uuid = getOP('log_uuid');
+        $info="opt: ".$opt." courseId: ".$courseId." courseVersion: ".$courseVersion." exampleName: ".$exampleName." sectionName: ".$sectionName." exampleId: ".$exampleId;
```
b21kurar, line 21: Is this tested? The variables inside the string aren't defined/declared earlier and shouldn't have data available resulting in a empty string (or maybe a crash, since the variables aren't defined? not sure how PHP handles this).

DuggaSys/microservices/shared_microservices/getUid_ms.php
```diff
@@ -0,0 +1,28 @@
...
+    include_once ("../../../../coursesyspw.php");
+    include_once ("../../../Shared/sessions.php");
+    include_once ("../../../Shared/basic.php");
+
+    function getUid(){
+        // Checks user id, if user has none a guest id is set
+        if(isset($_SESSION['uid'])){
+            $userid=$_SESSION['uid'];
+        }else{
+            $userid="1";
```
b21kurar, line 14: Just a mention, but do we **always** want to default to `uid = 1`? Not completely sure what the different values represent. If `uid = 1` represent a guest then this should be fine, as long as it is consistent in all other services.

DuggaSys/microservices/shared_microservices/getUid_ms.php
b21kurar: I have confirmed that this doesn't work. When echoing out the `$info` values, they are empty. 
image

If these are needed for the logging, then they should accessed before `$info` variable is declared/defined.

DuggaSys/microservices/shared_microservices/getUid_ms.php
```diff
@@ -0,0 +1,28 @@
...
+        }
+
+        $log_uuid = getOP('log_uuid');
+        $log_timestamp = getOP('log_timestamp');
+
+        $log_uuid = getOP('log_uuid');
+        $info="opt: ".$opt." courseId: ".$courseId." courseVersion: ".$courseVersion." exampleName: ".$exampleName." sectionName: ".$sectionName." exampleId: ".$exampleId;
+        logServiceEvent($log_uuid, EventTypes::ServiceServerStart, "getUid_ms.php",$userid,$info);
```
b21kurar, line 22: Not sure if this is working as intended, can't find any entries in `loglena6.db` with `getUid_ms.php`, I do however find my actions in `codeviewerService.php`

Shared/dugga.js
b21kurar, line 1: This works! 👍

DuggaSys/microservices/shared_microservices/getUid_ms.php
```diff
@@ -0,0 +1,28 @@
+<?php
+    date_default_timezone_set("Europe/Stockholm");
+    
+    // Include basic application services
+    include_once ("../../../../coursesyspw.php");
```
b21kurar, line 5: Not sure if this is used in this file? If removed an SQL error occurs, but it might be because it is needed in `editBoxTitle_ms.php` and therefore maybe should be included there instead?

DuggaSys/microservices/codeviewerService/editBoxTitle_ms.php
```diff
@@ -0,0 +1,48 @@
...
+
+	// Global variables
+	$exampleId=getOP('exampleid');
+	$boxId=getOP('boxid');
+	$opt=getOP('opt');
+	
+	getUid();
+
+	$exampleCount = 0;
+
```
b21kurar, line 20: Could use a comment describing what the following rows do, just a short description.

b21kurar: Such as an explanation for exampleCount, and what the queries below do

DuggaSys/microservices/codeviewerService/editBoxTitle_ms.php
b21kurar, line 1: This file works as is, but could use some clarification. Maybe an addition according to the suggestion in `getUid_ms`

DuggaSys/microservices/monolithic-to-microservices.md
```diff
@@ -0,0 +1,123 @@
+
+Disclaimer, this guide serves as a guideline and can not be followed 100% since the structure of all files differ. It's recommended to test after each step which include code changing to so its faster to spot if something goes wrong. When this guide was written no service using retrive information was used, write guide at step 5.
+***
+##Setting up file & folder:
```
b21kurar, line 4: I think there are spaces missing after the '#' characters

DuggaSys/microservices/monolithic-to-microservices.md
```diff
@@ -0,0 +1,123 @@
...
+
+###Step 4
+Add the following include in code example and call getUid(); once the code is added remove code that belongs in getUid(see code example:getUid)
+#####Codeexample: Include
+```
+include ('../shared_microservices/getUid_ms.php');
+getUid();
+```
+#####Codeexample: getUid
+```
```
b21kurar, line 46: Suggestion: writing `php` directly after the third backtick gives the code syntax highlighting.
image


```php
// code
include ('../shared_microservices/getUid_ms.php');
getUid();
````

File: DuggaSys/microservices/monolithic-to-microservices.md [hunk 8]
```diff
@@ -0,0 +1,123 @@
...
+| editBoxTitle_ms.php |
+| createBox_ms.php |
+
+***
+##Converting monolithic to microservice:
+###Step 1
+Copy the entire service file content from the monolithic php to your new made, the purpose of this is to simmer down 
+
+###Step 2
+Remove all the “opt”s that are not relevant to you. If your function is an else if, make it into an If statement.
```
b21kurar, line 31: Could be clearer, not entirely sure what this means. That means that someone further down will probably have a harder time understanding. Maybe add an example that shows what removing the "opt"s means?

Related to previous comment, this step could be different if you choose to change step 1. Either way, displaying an example of the simmer down approach would probably help clarifying.

DuggaSys/microservices/monolithic-to-microservices.md
```diff
@@ -0,0 +1,123 @@
...
+**Example:**
+| codeviewerService |
+| --- |
+| editBoxTitle_ms.php |
+| createBox_ms.php |
+
+***
+##Converting monolithic to microservice:
+###Step 1
+Copy the entire service file content from the monolithic php to your new made, the purpose of this is to simmer down
```
b21kurar, line 28: Wouldn't it be easier to just select the code that is needed rather than copying an entire file and then removing unnecessary parts?

DuggaSys/microservices/monolithic-to-microservices.md
b21kurar: This goes for all code examples

DuggaSys/microservices/monolithic-to-microservices.md
```diff
@@ -0,0 +1,123 @@
...
+        $info="opt: ".$opt." courseId: ".$courseId." courseVersion: ".$courseVersion." exampleName: ".$exampleName." sectionName: ".$sectionName." exampleId: ".$exampleId;
+        logServiceEvent($log_uuid, EventTypes::ServiceServerStart, "getUid_ms.php",$userid,$info);
+
+        $appuser=(array_key_exists('uid', $_SESSION) ? $_SESSION['uid'] : 0);
+
+        return $log_uuid;
+    }
+```
+###step 5
+If retrive information is not needed, remove it
```
b21kurar, line 68: "Retrieve Information" is what displays the response sent from the service, it might not be needed for the different services, but it is needed in some way. Either that all microservices send their individual parts of the "Retrieve Information" - functionality. Another way would be a "Retrieve Information"-service that all microservices call upon sending their `$info` and possibly microservice name

DuggaSys/microservices/monolithic-to-microservices.md
b21kurar, line 1: There are some formatting problems to be corrected and maybe some clarification on some steps.

b21kurar (changes requested):
Some changes are needed, but still good work!

DuggaSys/microservices/monolithic-to-microservices.md
b21ricni: Spaces makes no difference after ###

DuggaSys/microservices/monolithic-to-microservices.md
b21ricni: For people like you and me who understand the code yes, for people who actually will work with this, no

DuggaSys/microservices/shared_microservices/getUid_ms.php
b21ricni: It is used, when removed it causes error 500

DuggaSys/microservices/shared_microservices/getUid_ms.php
b21ricni: That is copied from old code, didn't check if it actually works

b21kurar (approved):
Make issues of the changes needed to the microservices, good job!

a21jeaha:
@b21ricni what is the ms getUid suppose to actually do?

a presumed it was geting the user id, but what it actually returns is the log_uuid. This is what i get out of that function 
image
and this is what im guessing im suppose to get 
image
the reason im asking is because it being used as a finished service by people making the micro services..

a21jeaha:
I have now made these changes to the file 
image
the function now returns the user id and gives the value "guest" if the session did not contain any valid id

b21ricni:
Seems like i forgot to send userId opsi, somehow still works in some cases. Presuming super user that dislikes not getting the id

PR #14162, opened 2023-05-22
G2 2023 v7 issue#13854

c21axepe:
Tests for duggaedservice.php.
Every test is tested and all tests passed when me and anton ran the api

### DIFF
--- a/DuggaSys/tests/duggaedservice-tests.php
+++ b/DuggaSys/tests/duggaedservice-tests.php
@@ -48,7 +48,7 @@ $testsData = array(
     
     
     'update an assignment' => array(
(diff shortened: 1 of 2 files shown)


b21antbr:
**Important notes**

- Needs a fresh database to test properly(A guide for this exists on discord).
- Last 3 tests can't have a correct expected output for the time being since the api can't handle it(this is noted with comments accordingly).
- The last 3 tests have been tested thoroughly so we know for a fact that they do work properly, again this can't be viewed with the expected output. 
- If the tester wants to see them working properly, contact either me or Axel and we can make that happen.

a19hakja (approved):
To make this work with my database, we had to make an extra commit with the path to my file instead, which worked just fine!

PR #14163, opened 2023-05-22
 changed so variables in constructElementType in diagram.js use the l…

### LINKED ISSUE #14161: Refactor function
## Description / Job
The function ``constructElementOfType`` inside ``diagram.js`` uses ``var`` declarations and either return ``undefined`` or an element (two different places). Try to refactor this function, and make it return in one place rather than two. The return types should be the same, ``undefined`` or element.

```
function constructElementOfType(type)
{
    var typeName = undefined;
    for (const name in elementTypes) {
        if (elementTypes[name] == type) {
            typeName = name;
            break;
        }
    }
    if (typeName) {
        var defaultElement = defaults[typeName];
        var newElement = {};
        for (const property in defaultElement) {
            newElement[property] = defaultElement[property];
        }
        return newElement;
    }

    return undefined;
}
```

a21nibah:
fixes #14161

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -2837,7 +2837,7 @@ function makeGhost()
  */
 function constructElementOfType(type)
 {
-    var typeName = undefined;
+    let typeName = undefined;
     for (const name in elementTypes) {
         if (elementTypes[name] == type) {
             typeName = name;
@@ -2845,14 +2845,13 @@ function constructElementOfType(type)
         }
     }
     if (typeName) {
-        var defaultElement = defaults[typeName];
-        var newElement = {};
+        let defaultElement = defaults[typeName];
+        let newElement = {};
         for (const property in defaultElement) {
             newElement[property] = defaultElement[property];
         }
         return newElement;
     }
-
     return undefined;
 }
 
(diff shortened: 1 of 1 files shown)


c21adano (changes requested):
java script usually throws an error if you try to return only in an if statement. Besides. it makes more sensce to return undefined if it can not get an object. 
"Try to refactor this function, and make it return in one place rather than two." This seems to be missing from the pull request

a21nibah:
> "Try to refactor this function, and make it return in one place rather than two." This seems to be missing from the pull request

java script usually throws an error if you try to return only in an if statement. Besides. it makes more sensce to return undefined if it can not get an object.

b21phiro:
> > "Try to refactor this function, and make it return in one place rather than two." This seems to be missing from the pull request
> 
> java script usually throws an error if you try to return only in an if statement. Besides. it makes more sensce to return undefined if it can not get an object.

True! A suggestion I have for a potential solution here, is to assign a dynamic variable that will be returned no matter what. Ex (unrelated code but for an idea)

```
function returnSomething() {
   let result = {};

  if (bla bla bla) {
      result = this
}
else {
   result = that
}

return result
}
```

b21phiro:
Other words don't return in the logical operations but rather as a result.

c21adano (approved):
calling the function returns undefined or a given element type.

PR #14164, opened 2023-05-22
#13852

a21ponkr:
Complete test, unsure if done correctly however this is the last day for creating tests.
issue #13852  (Implement test for courseedservice.php)

### DIFF
--- a/DuggaSys/tests/courseedservice-test.php
+++ b/DuggaSys/tests/courseedservice-test.php
@@ -83,8 +83,8 @@ $testsData = array(
             'coursename' => 'Testing-Course',
             'coursenamealt' => 'Course for testing codeviewer',
             'motd' => 'Code examples shows both templateid and boxid!',
-            'startdate' => '2020-05-01 00:00:00',
-            'enddate' => '2020-06-30 00:00:00',
+            'startdate' => '',
+            'enddate' => '',
 	    'blop' => '<!query-before-test-1!> <*[0][cid]*>'
         )),
         'filter-output' => serialize(array(
@@ -236,8 +236,8 @@ $testsData = array(
                 'coursename' => 'testCourseName',
                 'coursenamealt' => 'testCourseNameAlt',
                 'motd' => 'testMotd',
-                'startdate' => '2020-05-01 00:00:00',
-                'enddate' => '2020-06-30 00:00:00',
+                'startdate' => '',
+                'enddate' => '',
             'blop' => '<!query-before-test-1!> <*[0][cid]*>'
             )),
             'filter-output' => serialize(array(
@@ -281,8 +281,8 @@ $testsData = array(
                 'coursename' => 'testCourseName',
                 'coursenamealt' => 'testCourseNameAlt',
                 'motd' => 'testMotd',
-                'startdate' => '2020-05-01 00:00:00',
-                'enddate' => '2020-06-30 00:00:00',
+                'startdate' => '',
+                'enddate' => '',
             'blop' => '<!query-before-test-1!> <*[0][cid]*>'
             )),
             'filter-output' => serialize(array(
--- a/DuggaSys/tests/courseedservice-test.php
+++ b/DuggaSys/tests/courseedservice-test.php
@@ -13,7 +13,15 @@ $testsData = array(
     /* Insert into course */
 
     'Insert into course test' => array(
-        'expected-output' => '{"debug":"NONE!","motd":"UNK"}',
+        'expected-output' => '{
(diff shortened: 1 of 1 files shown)


DuggaSys/tests/courseedservice-test.php
```diff
@@ -0,0 +1,254 @@
...
+<?php
+ 
+include "../../Shared/test.php";
+ 
+$testsData = array(
+
+    /* Insert into course */
+
+    'Insert into course test' => array(
+        'expected-output' => '{"debug":"NONE!","motd":"UNK"}',
```
a21jeaha, line 16: you need to add the expected output!

it can be found here
image
for the correct output to be registered there, you must supply exactly the same values as the values used in the test.  
you will no doubt notice that there might be more values then the ones noted in the document, these are to be added as well (note, the image is not of your actual case)
image

a21jeaha: I might as well point out that there might be some data that needs to be manually removed from the JSON string.

DuggaSys/tests/courseedservice-test.php
```diff
@@ -0,0 +1,254 @@
...
+            'username' => 'usr',
+            'password' => 'pass',
+            'coursecode' => 'G1337',
+            'coursename' => 'Testing-Course',
+            'activeversion' => '1337',
+            'activeedversion' => null,
+            'registered' => false,
+	    'blop' => '<!query-before-test-1!> <*[0][cid]*>'
+        )),
+        'filter-output' => serialize(array(
```
a21jeaha, line 36: As mentioned before, some data might need to be manually removed from the JSON string. The data that is to be removed is decided here, or rather you specify what values are to be compared, these values will then be gathered from the service. But they will not automatically get filtered in the expected output, which is why you need to do it manually.

That said, you need to specify what these values are. 
When you have performed the action on LenaSYS that corresponds with the service that is to be tested, you take the output (same output as the expected output), and past it into this subservience. 

https://jsonformatter.org/json-pretty-print
if you change the display option to form, you will see what values are needed.
image

here you have an example from my own test. 
image
notice that you can specify values inside of the arrays as well if need be.

a21jeaha (changes requested):
Has this been tested at all, because if it was you would have noticed that no test are performed?
The expected output has not been changed from the example, neither has the filtered values. And i suspect that the values used as service data will cause problems. 

I have supplied some tips on what to do and where it is to be done, use them to solve this issue properly. 
If there is something you don't understand feel free to ask. Though i prefer to answer on site because of the complicated nature of the issue.

a21ponkr:
I don't see anything in network request, and I don't know if I can finish this. Would probably need some help even if it's probably too late.

DuggaSys/tests/courseedservice-test.php
```diff
@@ -0,0 +1,373 @@
...
+ 
+<?php
+ 
+include "../../Shared/test.php";
+ 
+$testsData = array(
+
+    /* Insert into course */
+
+    'Insert into course test' => array(
```
a21jeaha, line 15: the expected output is wrong.. and i have already supplied an example of how it should look..

DuggaSys/tests/courseedservice-test.php
```diff
@@ -0,0 +1,373 @@
...
+            "readonly": 0
+          }',
+        'query-before-test-1' => "SELECT cid FROM course WHERE coursecode = 'G1337' ORDER BY coursecode DESC LIMIT 1",
+        'query-before-test-2' => "INSERT INTO course (coursecode,coursename,visibility) VALUES('G1337','Testing-Course',1)",
+        'query-before-test-3' => "DELETE FROM course WHERE coursecode = 'G1337' AND cid = ?",
+        'query-after-test-1' => "DELETE FROM course WHERE coursecode = 'G1337' AND coursename = 'Testing-Course'",
+        'query-after-test-2' => "DELETE FROM course WHERE coursecode = 'G1337' AND coursename = 'Testing-Course'",
+	'query-variables' => "blop",
+        'service' => 'https://cms.webug.se/root/G2/students/a21ponkr/LenaSYS/DuggaSys/courseedservice.php',
+        'service-data' => serialize(array(
```
a21jeaha, line 32: why is the login data supplied still not changed ? 
if you test the file with pretty print enabled, you can clearly see that the test fail.
image

a21jeaha (changes requested):
this is still in an unfinished state, re read the comments i have already given.

d21erisu:
I'll close this PR since it's not ready.

Issue #14139, opened 2023-05-22
Clear gitfiles from SQLite db on refresh 


a21olljo:
Related to issue: #13934 
Refresh should clear the SQLite gitfiles table to then refill it with the data from GitHub.

Issue #13934, opened 2023-05-11
If a new file is added on github in a folder that is tracked by the synchronization module we want this file to be added automatically to the course view.


a21olljo:
When a folder is used to generate code examples, if the folder then gets a new file, that file should be automatically generated where the folder is in use. Should be done along with the refresh button. 
Related to issue: #13931

c21sebar:
We divided the update into three parts:

1. Hide an example when all files are deleted to that codeexample in the github repo.
2. Remove/unlink one or more boxes if one or more file (but not all) are removed from the github repi
3. Add boxes to the codeexample if more files are added to the githubrepo and those files should be added onto an existing code-example.

PR #14165, opened 2023-05-22
G3 2023 v7 issue#13934

c21sebar:
This is part one to solve issue  #13934 wich is the update part of existing code-examples auto create function.

We divided the update into three parts:
1. Hide an example when all files are deleted to that codeexample in the github repo.
2. Remove/unlink one or more boxes if one or more file (but not all) are removed from the github repi
3. Add boxes to the codeexample if more files are added to the githubrepo and those files should be added onto an existing code-example.

This PR will solve part 1. 
We decided to make a PR for the first part since we created this branch after we did the PR for https://github.com/HGustavs/LenaSYS/pull/14130 but before it was merged and would like this fix for part 1 (and 2 and 3).

We use the previous selection of coursevers here in this branch based on the "activeversion" so a tester must make sure to click the checkbox when creating/update a course:
image
image

We had an issue that when removing a file from the repo, clicking refresh the SQLite database did not recognize that the file was deleted. A fix for this is in place by issue https://github.com/HGustavs/LenaSYS/issues/14139 and PR https://github.com/HGustavs/LenaSYS/pull/14158 and would also like this fix before we start working on part 2 and 3.

There is a manual way around this "remove" problem.
So to test the functionality of this PR which is to hide examples when they are removed from github repo/sqlite:
sudo sqlite3 metadata2.db 
delete from gitFiles where fileName = 'FILENAME.FILETYPE' AND CID=XXXX;
This will remove the file "you deleted from the repo" in the sqlite db.
To trigger the hide part of the codeexample click the refresh button on the moment and reload the page.

This was done together with @b21sebgr .

### DIFF
--- a/DuggaSys/sectionedservice.php
+++ b/DuggaSys/sectionedservice.php
@@ -709,7 +709,7 @@ if($gradesys=="UNK") $gradesys=0;
 							$query3->execute();
 
 							$pdolite = new PDO('sqlite:../../githubMetadata/metadata2.db');
-							$query = $pdolite->prepare("SELECT * FROM gitFiles WHERE cid = :cid AND fileName LIKE ':fileName';"); 
+							$query = $pdolite->prepare("SELECT * FROM gitFiles WHERE cid = :cid AND fileName LIKE :fileName;"); 
 							$query->bindParam(':cid', $courseid);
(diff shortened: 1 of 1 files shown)


b21leowa (approved):
It works as intended, as seen below:
image

:+1:

Issue #14072, opened 2023-05-17
When opening up in darkmode, button is wrong


b21phiro:
##Description
When darkmode is active, the "darkmode" button should be filled. However, if the user opens up and allready have darkmode on, the button is not filled.

Figure 1.
![bild](https://github.com/HGustavs/LenaSYS/assets/45258330/ba73ffd7-1017-4ee4-bc6e-30f7c824236c)

## Job
When darkmode is active, the button should be filled. And then reverse.

PR #14166, opened 2023-05-22
Darkmode toggle button is now filled when user already have darkmode on #14072

c21theti:
image


a21aldsa (changes requested):
While, yes, the issue is solved I cannot approve this PR as it is right now. There are changes that do absolutely nothing in diagram.css and diagram.js, like commented out code and indentation changes that you forgot to revert before creating the PR.

I also question the reason for adding inline styling in diagram.php, when all other styling is and should be in, for example, blackTheme.css

a21aldsa (approved):
its ok

Issue #14090, opened 2023-05-19
Semicolons in HTML attributes


b21phiro:
## Decription
Semi-colons are not allowed in HTML-attributes, however some elements have that in ``diagram.js``.
```
                str += `<text class="cardinalityLabelText" dominant-baseline="middle" text-anchor="middle"; style="font-size:${Math.round(zoomfact * textheight)}px;" x="${labelPositionX + textWidth/2 + 2 * zoomfact}" y="${labelPositionY + ((textheight/2 + 2) * zoomfact) + lineLabel.labelMovedY + lineLabel.displacementY}">${line.label}</text>`;

                str += `<text class="cardinalityLabelText" dominant-baseline="middle" text-anchor="middle"; style="font-size:${Math.round(zoomfact * textheight)}px;" x="${centerX - (2 * zoomfact) + lineLabel.labelMovedX + lineLabel.displacementX}" y="${centerY - (2 * zoomfact) + lineLabel.labelMovedY + lineLabel.displacementY}">${line.label}</text>`;
```

## Job
Remove the semi-colons that are in the HTML attributes.

PR #14167, opened 2023-05-22
Deleted ; from HTML attributes #14090

c21theti:


b21eddka (changes requested):
I believe this issue refers to the use of semicolons in the actual HTML tag as that is considered wrong syntax. Only when styling with CSS should there appear semicolons. Here is a better explanation: https://stackoverflow.com/questions/22064533/why-dont-you-put-a-semicolon-after-html-and-css-tags

Issue #14119, opened 2023-05-22
Allow use of tab in toolbar


b21phiro:
## Description/Job
As of right now you can't use the TAB-key to select modes or items in the diagram. By setting the tab-index on elements (or even better, use sematic tags such as buttons) you can solve that issue. Implement a way to use the tab to navigate through the toolbox.

![bild](https://github.com/HGustavs/LenaSYS/assets/45258330/691a87df-57e9-46fa-8d55-3a34fdfd27a6)

## Tips
Tabindex: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex
Button element: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button

b21phiro:
## Yes
Indeed, it's concidered bad practice according to many, you can read more about it here:
https://adrianroselli.com/2014/11/dont-use-tabindex-greater-than-0.html

However, having no way of navigating using tab is worse. In our case, we do not have any substantial that a screen-reader could read.

I do agree, that in most cases it's a bad idea. However, in this context the flaws are not as relevent. Howver this is highly subjective.

## Replace with elements
Yes, that's a possible solution as stated, however as you said it brings problems. So, it's easier for now to use tabindex.

a21aldsa:
> ## Yes
> 
> Indeed, it's concidered bad practice according to many, you can read more about it here: https://adrianroselli.com/2014/11/dont-use-tabindex-greater-than-0.html
> 
> However, having no way of navigating using tab is worse. In our case, we do not have any substantial that a screen-reader could read.
> 
> I do agree, that in most cases it's a bad idea. However, in this context the flaws are not as relevent. Howver this is highly subjective.
> ## Replace with elements
> 
> Yes, that's a possible solution as stated, however as you said it brings problems. So, it's easier for now to use tabindex.

A screen reader can already navigate this site fairly well, the issues that arise are not because of the menu: that part works fine. The issues lie in how the options pane is always there just visually hidden. How there is no tooltips on keyboard focus. How there is no way of placing elements without clicking, etc etc etc.

b21phiro:
> > ## Yes
> > Indeed, it's concidered bad practice according to many, you can read more about it here: https://adrianroselli.com/2014/11/dont-use-tabindex-greater-than-0.html
> > However, having no way of navigating using tab is worse. In our case, we do not have any substantial that a screen-reader could read.
[...]

PR #14171, opened 2023-05-23
G1 2023 v7 #14119

b21matfi:
Made it so that you can use tab to go thru toolbar but you have to unfocus from the diagram page if you press tab in the after you have placed something or clicked in the diagram its gonna focus on to the option pane. Either click tab the first thing or first select the pointer mouse and press tab. Solved using button so you can press enter to select each tool. Popout does not work with tab yet

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -230,6 +230,7 @@
 .diagramIcons:hover, .active, .diagramIcons:focus{
     background: var(--color-primary-hover);
     border: solid 1px var(--color-primary-light);
+    outline: auto;
 }
 .diagramIcons img{
     width:34px;
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -228,6 +228,9 @@
     
    
 }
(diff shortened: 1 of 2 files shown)


a21aldsa (dismissed):
Pop out menus make the alternatives below go way down, this needs to be fixed: 

image

b21phiro:
This can be fixed by setting the container of the buttons to either a fixed height, for example the buttons width+height. The button should also be positioned relative, as the inner content (the menus) are positioned absolute.

b21matfi:
how did you get this issue i can not see get the issue 
image

a21aldsa:
> how did you get this issue i can not get the issue

I get it repeatedly and consistently on firefox, every time i open any pop out. I do not get it on chrome and edge though. Just to make sure this wasnt related to some plug in or extension i may have; I also tested it on a clean install of firefox developer edition, the issue was present here too. 

I double checked that I am testing on the right branch. (g1-2023-v7-#14119)

b21eriho:
There is not time to fix this

Issue #14110, opened 2023-05-19
Finnish the save-load system from localstorage


b21phiro:
## Description
Right now we have completed the saving of the diagram and the loading of the saves. However we need to polish the functions in such a way that they work together.

https://github.com/HGustavs/LenaSYS/pull/14106
https://github.com/HGustavs/LenaSYS/pull/14046
https://github.com/HGustavs/LenaSYS/pull/14105

## Job
Try to link/hookup the load-save system together. I.e. you can save multiple files, name them, load them.

Issue #14117, opened 2023-05-22
Override save-name issue


b21phiro:
## Job/Description
When creating a save, you can write a name for the save. However, right now you can accidentally override a save-file because you are allowed to have duplicate save-names.

If, there's a name that allready exist, the user should get notified that there's allready a save-file with that name, and if the user want to proceed, and should be warned that the previous file will be overwritten.

a21nibah:
I don't really understand the point of having diferent file names and so on for local storage. Why are we expanding functionality of local storage anyway? it is a bad way to save diagrams. Is the point not to have diagrams on the server from what i have seen from another issue? if so local storage only needs to save the currently active diagram if even that. I really do not get why we are expanding functionality of something that does not need to be there if we are going to save everything to the server. If the user switches browser or device well those diagrams will be lost anyway.
@b21phiro @b21eriho 
I believe this needs to be reassessed before making any further changes. Also the naming function is plane broken right now. It is getting the name and then it overrides that name no matter what. so the name is not even getting used. What happened during the review?
I also wrote in another issue when working with this that we needed to have unique keys but what seems to happen here now is that a name is getting treated as a key which is debatable if it should. Personally I think a name and a key should be 2 diferent things where a key is a unique timestamp so that we can allow for duplicate names if the user wants.

b21eriho:
That is a reasonable point to make. On the other hand however this functionality will hardly be harmful and the server-side saving is still under consideration and research. Conclusively this is functionality we have decided to implement in the meanwhile.

a21nibah:
I believe this issue should be reassigned to someone else to fix the front end. I have now made it so you can save and load files with diferent names from local storage and i have added the backend to check for diagram overrides. but someone eneds to complete this issue by adding a modal for the front end and then use my backend to tie in. I believe it is better if someone else makes the front end so we get it right from the beginning.

b21eriho:
Noted, someone else will be assigned as well as you when there is a vacancy so that you can collaborate on the finishing touches

PR #14176, opened 2023-05-23
G1 2023 v7 #14110

d21matbo:
Reworked the current system to store and map multiple JSON strings to their own key, while they are all in the same string (localstorage). Used this system to connect the saving and loading functions for local diagrams.

You can either save a diagram with the same name to overwrite it, or a new name to save it as a seperate diagram.
AutoSave is called the first time a diagram is saved, and then it is called by the function `saveDiagramBeforeUnload()`

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -13106,18 +13106,22 @@ function closeModal(){
     modal.classList.add('hiddenLoad');
     overlay.classList.add('hiddenLoad');
 }
-
+/**
+ * @description Check whether there is a diagram saved in localstorage and load it.
+ * @param {string} key The name/key of the diagram to load.
+ */
  function loadDiagramFromLocalStorage(key)
 {
-    // Check whether there is a diagram saved in localstorage and load it. key for current diagram is CurrentlyActiveDiagram
     if (localStorage.getItem("diagrams")) {
         var diagramFromLocalStorage = localStorage.getItem("diagrams");
         diagramFromLocalStorage = (diagramFromLocalStorage[0] == "{") ? diagramFromLocalStorage: `{${diagramFromLocalStorage}}`;
         let obj = JSON.parse(diagramFromLocalStorage);
         if (obj[key] === undefined) {
-            return;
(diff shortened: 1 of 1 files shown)


a21nibah:
would like some clerification. so is the goal that all diagrams should be saved under the key diagrams? issue #14117 uses another implementation of the following where all keys are fetched from local storage rather than this.
    let diagramKeys;
    let localDiagrams;

    let local = localStorage.getItem("diagrams");
    if (local != null) {
        local = (local[0] == "{") ? local : `{${local}}`;
        localDiagrams = JSON.parse(local);
        diagramKeys = Object.keys(localDiagrams);
    

Otherwise it looks good. saves are done as they should and loading is done correctly from what i can see. 
although i can not figure out how you go about saving multiple files as described in the issue. Is this something that has been implemented yet in the UI or what is that about?

b21eriho:
It would seem that multiple issues that realistically should have been assigned in sequence were assigned in parallell. This issue #14110 is the issue that should implement the saving and loading from local storage and make that usable to the user. Due to the aforementioned administrative fault we will have to make some choices resulting in work needing to be redone.

What will happen is that this PR will be merged when tested and then the other issue #14117 will have to be rerun to add the functionality desired there on top of this implementation. Apologies for the lost work.

Regarding the testing please elaborate on what you said about not being able to save multiple files

a21nibah:
the functionality from issue #14117  can be merged without much having to be redone. You really have to remove the functionality that displays all keys and use the eone from this pr. otherwise it should all work since the issue mostly consists of making function calls to the saveDiagram function. 
also the issue for this pr states:
Try to link/hookup the load-save system together. I.e. you can save multiple files, name them, load them.
Has that been implemented? looking at the code seems like it should work but can not find anythin where you can do this in the UI.

b21eriho:
All these things can be done in this implementation, there may be some issue with the availability though I can't say I'm sure of this.

a21nibah (approved):
from what I can see everything works as it should. loading, saving etc. The code looks good and i can not see any issues during testing. Nice work.

PR #14179, opened 2023-05-23
G1 2023 v7 #14117

a21nibah:
if a file has no name a file name will be generated containing the date and time it was saved.
Fixed the saving function not working at all. It previously would take user input and then just use the key for currently active diagrams. Files should now save properly.
Added a popout that informs the user if a duplicate name exists and gives the option to either save or cansel.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -13138,7 +13138,15 @@ function saveDiagramAs()
     if (fileName.trim() == "") {
         fileName = "Untitled";
     }
-
+    const names=getAllLocalStorageKeys();
+    for(let i=0; i<names.length; i++)
+    {
+        if(names[i]==fileName)
+        {
+            alert("about to override");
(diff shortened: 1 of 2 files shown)


a21aldsa (approved):
While I agree with the comments you made in the original issue about saving in localstorage being a generally bad way to save things, this issue is definitely still solved and works fine. Good job.

a21aldsa:
Although I do not agree with all keys from localstorage being displayed when pressing load, some of these should not be there, like themeBlack. I believe this to be a future issue though since this issue was about overwriting already existing saves.

a21nibah:
#14110 fixes this.

> Although I do not agree with all keys from localstorage being displayed when pressing load, some of these should not be there, like themeBlack. I believe this to be a future issue though since this issue was about overwriting already existing saves.

a21aldsa (approved):

Issue #14121, opened 2023-05-22
Better loadscreen


b21phiro:
## Description/Job
When loading up the diagram, there's no indication to the user if the diagram is being loaded or if it's broken. While the diagram is being loaded, there should be a spinner or _something_ that indicates that the diagram is in fact being loaded.

Figure 1.
![bild](https://github.com/HGustavs/LenaSYS/assets/45258330/a510c5c2-017c-48b6-9c62-bfa304a65ae4)

### Tips
- You can simulate slow internet using the browsers devtools on the network tab.
- Make use of eventlisteners: https://javascript.info/onload-ondomcontentloaded

HGustavs:
Spinning wheel like in for example yotube?

b21phiro:
Yeah that would be good! Just something that indicates that it is in fact loading

PR #14185, opened 2023-05-23
 Better loadscreen #14121 

a21aldsa:
Created a loading spinner that disappears after the window is loaded and also calls the startup functions.

For the tester: you can use the networking tab of developer tools to artificially throttle your connection to be slower, this allows you to see the loading spinner better if you have fast internet.

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -15,15 +15,8 @@
     align-items: center;
 }
 #spinnerTMPPlaceholder{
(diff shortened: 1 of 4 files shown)


a21oligu (approved):
Works well, good job! The zoom-container is still visible somehow, but not necessarily a problem.

Instead of just hiding the spinner with `.style="none"`, it could be removed with the element's `.remove()` function.

PR #14188, opened 2023-05-23
G3 2023 v7 issue#14050 revert

### LINKED ISSUE #14050: Remove dropdown that is used to set next and previous code example
There is a setting page located in each code example, which holds a dropdown used to go between code examples. This was thought in issue #13877 to not be synced to the order found in course list, but it was found out to be. 
The old manual way of chosing the order is now non functional and can be removed (which is this issue). 
image
What to be removed
image

a21rammo:
@a21olljo Double check if this is what u wanted

image

image

close #14050

### DIFF
--- a/Shared/navheader.php
+++ b/Shared/navheader.php
@@ -356,8 +356,8 @@
 
 			// Either generate code viewer specific nav menu or a spacer
 			if(isset($codeviewer)){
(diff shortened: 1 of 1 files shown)


a21olljo:
Yes, this is what was requested from the issue.

b21leowa (approved):
I approve these changes

Issue #14067, opened 2023-05-17
Updating anything in entity which is connected prints error when it shouldn't


b21eriho:
If an entity has lines connected and you attempt to change any aspect it will do so but print the red popout saying you can't change from one type to the same type despite not having changed anything relating to type. Further the change does not actually get applied if the change is made to a weak ER-entity and the reason stated is again that you cannot change from ER to ER

PR #14190, opened 2023-05-23
changeState() misbehaving #14067

a21oligu:
Fixed an issue where the changeState() function displayed an error message even though the changes were applied. 

Removed a function call `generateContextProperties()` as the propertySet was overridden by it, causing the changes not to be saved and applied properly. Solves issue #14067


d21matbo (changes requested):
Changes works for most elements, I was only able to find that `ERAttribute` still can't change its variant while it is connected (see image below for error message).

image

And don't forget to remove the current `console.log()` call.

PR #14193, opened 2023-05-23
It is now possible to move both lifeline and superstate without overlapp

c21adano:
The old checks never checked what kind of element you were moving with the mouse, so i added a element.kind check inside the if statment for both objects that check if you are holding the right kind of element


c21theti (approved):
Looks good! It is now possible to move both lifeline and superstate without overlapp

Issue #14113, opened 2023-05-22
Don't allow clicking on save button if there's nothing to save


b21phiro:
## Description/Job
As of right now, if you click on the save button you are still allowed to write a name, even if you can't save anything. Make so the button is disabled if there's nothing that can be saved. Make sure that the button have visual clue about this too.

Figure 1.
![bild](https://github.com/HGustavs/LenaSYS/assets/45258330/d4064c9a-540f-4e86-ae76-14b1ddab65dc)

PR #14197, opened 2023-05-23
Don't allow clicking on save button if there's nothing to save #14113

b21oscpe:
The function I made checks if data[] == 0 which means the container is empty or if the historyindex is -1. It is called whenever the mouseUp event is triggered. It should work with both ctrl+z and removing elements. The buttons is set to "disabled" (only by css) by default.

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -833,7 +833,7 @@
 }
 
 .disabledIcon{
-    background-color: white;
+    background-color: #cbb1e0;
 }
 
 #a4options {
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -832,6 +832,10 @@
     font-weight: bold;
 }
 
+.disabled{
+    background-color: #888;
+}
+
 #a4options {
     display: flex;
 }
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -833,7 +833,7 @@
 }
 
 .disabledIcon{
-    border-color: #da2727;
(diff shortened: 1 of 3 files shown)


a21oligu (changes requested):
When loading a diagram, the save buttons appears as disabled but will still allow the user to open the save-modal. There is also nothing that prevents the user from saving the same diagram twice, with zero changes made to it the second time.

b21oscpe:
How is this replicated? For me the modal doesn't show up and an error message is shown.

a21oligu:
> How is this replicated? For me the modal doesn't show up and an error message is shown.

This is also the case for me, except for when you load a diagram with the load button. After the diagram the loads, the button appears as disabled but will still open the modal on click.

If you save your changes, you can press the save button again. That is, you can save even though no changes has been done since the last save.

b21oscpe:
I have now fixed so that it does not say you shouldn't be able to save if you load in a diagram, @b21eriho or @b21phiro should decide if this is ok or not. The problem with this is that the function currently checks if data.length == 0 (no elements on the canvas) OR currentHistoryIndex == -1 which means it was just loaded in and no changes have been made. However, when you load in a diagram there are elements on the canvas and the historyIndex is saved from the last time the diagram was edited. I do not know what other condition could be used to check this without creating uneccessary variables / setting the currentHistoryIndex to 0 once you load a diagram

b21eriho:
That is fine, The implementation you have made is still a good improvement

a21oligu (approved):
Good job!

Issue #14182, opened 2023-05-23
Fix Test API relative path


c21alest:
Service in JSON uses static path at the moment but needs to be changed to relative

PR #14200, opened 2023-05-23
G2 2023 v7 issue#14182 fix test api relative path

a21jeaha:
An explanation of what has been done can be fawned when following the link below.  
https://github.com/HGustavs/LenaSYS/issues/14182#issuecomment-1559331670


c21alest:
Replace old solution? Both can't be used at same time!

a21jeaha:
i did not realize there was an old version, but seeing as that has already been merged, then this verison can be ignored and this PR can be closed.

PR #14202, opened 2023-05-23
G2 2023 v7 issue#14156

b21hamwe:


a20fanma (changes requested):
To me, it seems that the microservice does not work properly. It will not generate any outputs. I think JSON array is missing, but I'm not sure.

a21jeaha:
when i try to access the file i get this result 
image

it should look like this. 
image

a21jeaha:
i noticed that you never execute the function 
```php
pdoConnect();
```
Is this function executed somewhere ells ?

Issue #14180, opened 2023-05-23
Issues with statemachine.historylog


d21matbo:
Functionality that is based on statemachine.historylog currently has a potential risk to the system. The functionality affected include (but is not limited to) `exportWithHistory()` and `storeDiagramInLocalStorage()`.

It has been identified that placing an element on `x=200` or `y=200` that coordinate-axis is not transferred to the historylog (see below). 
*It is currently unknown if these are the only coordinates that creates the issue*. 

The issue is hard to replicate since it has to be "pixel perfect" and as such might not happen to the regular user. However it does happen and when it does the save is completely corrupted. As the save crashes the system when it tries to recreate the diagram from that saved state etc.

The following code blocks show the same diagram exported as (1) `exportWithHistory()` and (2) `exportWithoutHistory()`.
(1)
```json
{
	"historyLog": [{
		"id": "0CE105",
		"kind": "EREntity",
		"y": 43,
		"time": 1684826639157
	}],
	"initialState": {
		"elements": [],
		"lines": []
	}
}
```
(2)
```json
{
	"data": [{
		"kind": "EREntity",
		"stroke": ["#383737"],
		"x": 200,
		"y": 43,
		"id": "0CE105"
	}],
	"lines": []
}
```
As can be seen in the code, (1) does not have any value for the `x` coordinate.

d21matbo:
Upon further research I've identified that the cause for the issue occurs in `StateChangeFactory` when the filtering on the keys is applied.
```javascript
var uniqueKeysArr = Object.keys(element).filter(key => {
  return (Object.keys(defaults[element.kind]).filter(value => {
    return defaults[element.kind][value] === element[key];
  }).length === 0);
});
``` 
This code is used to only return keys that are not equal to the default key-value pairs of that element type.
The problem with this is that i.e. `ER-Element` has a default `width = 200` and `height = 50`, which means that if the elements `x` or `y` value is equal to these numbers, then they will be ignored from the list of key-value pairs that will be saved.

PR #14203, opened 2023-05-23
G1 2023 v7 #14180

d21matbo:
Fixed the issue with missing `x` and `y` coordinates in `historylog`.

Added an if-statement inside the filter function that basically ignores the filtering for the keys: `x` and `y`, where it was possible for the issue to occur. Now the coordinates should always be present in the historylog and as such the system should no longer crash.

The following code snippet is created by  the `exportWithHistory()` function. 
*Notice that the value for x is equal to 200 as this was not possible before.*
```JSON
{
	"historyLog": [{
		"id": "9DF74D",
		"kind": "EREntity",
		"x": 200,
		"y": 553,
		"time": 1684850801336
	}],
	"initialState": {
		"elements": [],
		"lines": []
	}
}
```

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -115,6 +115,7 @@ class StateChangeFactory
 
         // Get the keys of the values that is unique from default
         var uniqueKeysArr = Object.keys(element).filter(key => {
+            if (key === 'x' || key === 'y') return true;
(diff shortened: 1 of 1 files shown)


b21oscpe (approved):
Testing was complicated and i could not confirm that it works however the developer hsa tested it and the issue is very specific.

Issue #13935, opened 2023-05-11
Generating a code example should place it at the bottom of the moment 


a21olljo:
From #13879
When a new example is added, it should be placed added at the bottom of the list. To then be manually be reorder with the normal ordering operation.
Blocked by: #13934

Issue #13931, opened 2023-05-11
When an example is removed from github it should be hidden in the course view and not deleted


a21olljo:
From: #13875
Continuation of #13934

Issue #13879, opened 2023-05-09
Code example New Github File


HGustavs:
If a new file is added on github in a folder that is tracked by the synchronization module we want this file to be added automatically to the course view.

If the list currently is:

Canvas Animate
Canvas Draw Objects
Canvas Gradient
Canvas Immediate

and we add a new example "Canvas Animation Reaction.html" we want this example to be added to the bottom of the list and if we want to move this example to below Canvas Animate we manually reorder the examples with the normal ordering operation.

The naming of the code example should be Canvas Animation Reaction.html until the user renames it

New files should not be hidden by default, but be shown immediately after the first refresh (manual or automatic refresh).

PR #14204, opened 2023-05-23
G3 2023 v7 part2 issue#13934

c21sebar:
This will solve the part 2 and 3 of the update of code-examples. And creating,updating and delete/hide of codeexample should be considered done now.

1. Hide an example when all files are deleted to that codeexample in the github repo.
2. Remove/unlink one or more boxes if one or more file (but not all) are removed from the github repi
3. Add boxes to the codeexample if more files are added to the githubrepo and those files should be added onto an existing code-example.

https://github.com/HGustavs/LenaSYS/issues/13549
https://github.com/HGustavs/LenaSYS/issues/13936
https://github.com/HGustavs/LenaSYS/issues/13935
https://github.com/HGustavs/LenaSYS/issues/13879
https://github.com/HGustavs/LenaSYS/issues/13931
https://github.com/HGustavs/LenaSYS/issues/13934
In combination of this PR and previous ones related to this I think we can consider these issues completed.

This will remove a box if the files is removed from github and sqlite is updated to reflect the new githubchange. To update SQLite you need to press the github download/refresh button in the navbar. The removed box/boxes will also update the codeexample template to suit the new ammount of files in the example.

This will add boxes if new files are added to the moment folder (after updating, pressing the github dl button and sqlite has been updated) that should be in the same codeexample and update the codeexample with the correct template to suit the new ammount of files.

I have disabled the automatic refresh that runs on all lids since in this version/branch the auto refresh of lids is run on all lids for all listentries every min (meaing a lot of spam and hard to test properly).
in sectioned.js in the setInterval function
```js
 for (let i = 0; i < collectedLid.length; i++) {
                //createExamples(collectedLid[i]);
            }
```
So you need to press the refresh button manually on the moment you want to refresh/update.


b21leowa (approved):
I approve :+1:

Issue #13859, opened 2023-05-08
Implement test file for sectionedservice.php


d21erisu:
Related to: #13847

You can found a guide on how to use this API [here](https://github.com/HGustavs/LenaSYS/blob/cdb827478334ec6099763471afd3e0c16ade1332/Shared/Documentation/Test%20API/Guide%20on%20how%20to%20use%20Test%20API.md).

a21jeaha:
@d21erisu  The test document that should have been reworked to include login and expected values is not complete. There is no login values and the output is just copy pasted, with no regard to what data we want.

```
{
    "entries": [],                  
    "debug": "NONE!",
    "writeaccess": false,
    "studentteacher": false,
    "readaccess": false,
    "coursename": "UNK",
    "coursevers": "UNK",
    "coursecode": "UNK",
    "courseid": "UNK",
    "links": [],
    "duggor": [],
    "results": [],
    "versions": [
        {
            "cid": "1",
            "coursecode": "DV12G",
            "vers": "45656",
            "versname": "HT15",
            "coursename": "Webbprogrammering",
            "coursenamealt": "UNK",
            "startdate": "2014-12-29 00:00:00",     
            "enddate": "2015-03-08 00:00:00",
            "motd": "Webbprogrammering - HT15"
        },
```

a21jeaha:
I now have to waste my time doing what was already supposed to have been finished. Therefore this might take longer to complete then expected..

d21erisu:
> I now have to waste my time doing what was already supposed to have been finished. Therefore this might take longer to complete then expected..

That is unfortunate. The only upside is that now it gets done properly :)

a21jeaha:
@d21erisu im a bit lost on how this API is suppose to work, from our earlier talks i interpreted it that no output was to be written in the service test md documents, and that the expected output was to be generated in the test. but with this API, im supposed to hard code in expected output?

a21jeaha:
Documentation for how these tests are used can be found in Shared/Documentation/Test API

The service used in test removeListentries, is never used
The github related services "REFGIT" and "CREGITEX" do not have tests for the time being. they were not completed at the time. They therefore need to be added to this file.
The services changeActiveCourseVersion_sectioned and updateCourseVersion_sectioned are not in a functioning state, the tests for them are therefore not complete.

To be able to perform the tests in a passable state, you need to have a fresh database, alternatively the testdatabase if it is ready for use.

a21jeaha:
[...]

PR #14205, opened 2023-05-24
G2 2023 v7 issue#13859 v4 rebased

a21jeaha:
Documentation for how these tests are used can be found in Shared/Documentation/Test API

The service used in test removeListentries, is never used

The github related services "REFGIT" and "CREGITEX" do not have tests for the time being. they were not completed at the time this PR was made. They therefore need to be added to this file at a later date.

The services changeActiveCourseVersion_sectioned and updateCourseVersion_sectioned are not in a functioning state, the tests for them are therefore not complete.

To be able to perform the tests in a passable state, you need to have a fresh database, alternatively the testdatabase if it is ready for use.

There is one bug i would like to mention though, some of the tests will always fail the second time they are run without resetting the database. As can be seen below, all the values of "moment" change from null to 0. This bug is present in the original services code. 
image
image
This will not hinder the test, but there is a caveat. The test can only be performed once with a fresh database any further attempts before resetting the database will without a doubt result in some test failing, keep that in mind while reviewing this test!     

https://github.com/HGustavs/LenaSYS/issues/13859#issuecomment-1557185342


b21hamwe (approved):
Sufficient solution to the issue!

PR #14215, opened 2023-05-24
G2 2023 v7 14154 create microservice createnewcodeexample

b21kurar:
Moves the functionality to create new items to a microservice and changes to dugga.js to ensure that it is called when creating new items.

The call to the microservice is not commented out, as it doesn't interfere with basic functionality.

### DIFF
--- a/DuggaSys/microservices/shared_microservices/createNewCodeExample_ms.php
+++ b/DuggaSys/microservices/shared_microservices/createNewCodeExample_ms.php
@@ -1,7 +1,9 @@
 <?php
 /**
- * Handles creation of new items on a coursepage.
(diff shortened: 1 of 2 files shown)


b21ricni (changes requested):
Does work, dunno if i did it wrong tho, but it calls the correct function.
User: Toddler, course: testing, Tried to create new code item
image

b21kurar:
I have added an include to `coursesyspw.php` since this appears to fix some database related problems, other than that I don't know why the includes keep giving me troubles...

b21ricni (approved):
It now creates a new code, nice! :3

d21erisu:
great

PR #14216, opened 2023-05-24
G3 2023 v7 issue#14189

a21lukpo:
We have moved the code from "FetchGithubRepo.php" to a new file called "gitfetchService" that is now located in the "DuggaSys" folder instead of the old "recursivetesting" folder. We have also changed the links in sectioned.js and sectionedservice.php so everything still works correctly


b21leowa:
I cannot create a new course right now... See the error message below:

image

b21leowa:
You can remove the following code because we are not using the ajax service right now in old fetchGithubRepo.php
This code is in courseed.js and is not in use:
image

Because the ajax "catcher" is already commented out and is not in use...
image

b21leowa:
It works now! I approve this merge :+1:

b21leowa (approved):
:+1:

PR #14218, opened 2023-05-24
G2 2023 v7 issue#14160 created microservice createNewCourse

c21axepe:
We created a file for the microservice in coursedservice where you create a new course. The thing that bothers us is the solution in dugga.js. Check comments there for further description


Shared/dugga.js
```diff
@@ -1017,13 +1017,27 @@ function AJAXService(opt,apara,kind)
 	}
 
 	if(kind=="COURSE"){
+		// Call microservice to create new course 
+		// TODO: This is probably not a very good solution, this is done with the opt comparison
+		// For now this is usable, but should be remade when all microservices related 
+		// to courseedservice.php are implemented
+		if(opt === "NEW"){
```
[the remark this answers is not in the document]
c21axepe: The problem is that our microservice should only create a new course, not get. When you start courseed.php coursedservice.php opt=GET. This will fetch all courses. Our service as said should not do that, only opt=NEW. Thats why in dugga.js we have an if statement so that the service knows when to use our microservice.

a21jeaha (changes requested):
i left a comment where i stated that there was problems with the code. 
you need to find a solution where the old monolithic service is not required for the micro service to function.

DuggaSys/microservices/courseedService/createNewCourse_ms.php
```diff
@@ -0,0 +1,68 @@
...
+$query->execute();
+
+while ($row = $query->fetch(PDO::FETCH_ASSOC)) {
+	$username = $row['username'];
+}
+
+$isSuperUserVar = false;
+
+if (checklogin()) {
+	if (isset($_SESSION['uid'])) {
```
a21jeaha, line 39: there is an ms to check user ID, getUid_ms use that.

DuggaSys/microservices/courseedService/createNewCourse_ms.php
```diff
@@ -0,0 +1,68 @@
...
+		$userid = $_SESSION['uid'];
+	} else {
+		$userid = "UNK";
+	}
+	$isSuperUserVar = isSuperUser($userid);
+
+	$ha = $isSuperUserVar;
+
+	if ($ha) {
```
a21jeaha, line 48: dont complicate this, you only need to check if it is a super user by using 

if(isSuperUser(getUid())

DuggaSys/microservices/courseedService/createNewCourse_ms.php
```diff
@@ -0,0 +1,68 @@
...
+
+// Missing Functionality
+//		New Code Example + New Dugga
+//		Graying link accordingly
+
+date_default_timezone_set("Europe/Stockholm");
+
+include('../shared_microservices/getUid_ms.php');
+
+getUid();
```
a21jeaha, line 15: what does this even do?

DuggaSys/microservices/shared_microservices/getUid_ms.php
a21jeaha, line 1: these are changes i accidentally made while i was working on my own issue... but in your branch 😧, but they are valid and should not be changed :)

DuggaSys/microservices/courseedService/createNewCourse_ms.php
b21antbr: we followed the guide

a21jeaha: well yes, but.
image
you still need to know what your using and why it is used, that function did nothing at all useful in its original state. 
in its currents state (which i have talked over with its creator) it returns the userid of the current user. 

that said as it is, it does nothing 😥

DuggaSys/microservices/courseedService/createNewCourse_ms.php
b21antbr: doesn't work :(

a21jeaha (approved):
the code works in its current state, only supplie some comments for the next group to work on this, on the getUid() issue.

a21jeaha (approved):

c21alest:
Thanks!!!

Issue #14213, opened 2023-05-24
Weird behavior when canceling line-drawing


b21eriho:
If you begin drawing a line from one entity and then before connecting it to another press esc there will be a ghost under your pointer. resolve this.

PR #14220, opened 2023-05-24
 Weird behavior when canceling line-drawing #14213 

a21aldsa:
Added a missing if statement for drawing ER attributes, it did not check to make sure this ER attribute was actually defined before attempting to draw svg for it, which caused the erroneous and faulty ghost.


b21eddka:
Looks good, approved!

PR #14221, opened 2023-05-24
fixed a4 black box issue

c21adano:
There was a spelling mistake in the css file that didnt link the a4Rect to anything.

There appears to be another visual issue with the a4 templet too, that requires you to move the screen in some way before the a4 radius is shown, but that was not part of this issue so was left alone for now.


a21aldsa (approved):
The a4Rect is no longer a black box. This issue is solved. Good job

PR #14224, opened 2023-05-24
 fixed so note now follows the corrrect naming skeme

### LINKED ISSUE #14222: Change name for NOTE element in diagram.js
## Description
Elements are listed in ``defaults`` object in ``diagram.js``. Names for objects and types are in camelcase, while the new note element does not follow this.

```
    NOTE: { name: "Note", kind: "NOTE", fill: "#FFFFFF", stroke: "#000000", width: 200, height: 50, type: "NOTE", attributes: ['Note'],},```

## Job
Rename as camelcase, or if that does not work, find name that can follow the same name-scheme but keepts the explaining name.

a21nibah:
fixes #14222 
when testing try drawing lines and elements with type note and check for errors.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -804,7 +804,7 @@ const elementTypes = {
     sequenceLoopOrAlt: 14,
 
 
-    NOTE: 15,
+    note: 15,
     
 };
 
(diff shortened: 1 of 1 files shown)


a21aldsa (approved):
naming scheme is now correct and nothing seems to have changed site-wise. Good job.

PR #14227, opened 2023-05-24
G2 2023 v7 issue#14195

a19hakja:
Created a microservice for updating course-version! :D


b21kurar:
This is working as intended woo!

PR #14229, opened 2023-05-24
Fixed error that occurred when connecting line with note

c21adano:
there was an old if statment that is never used that got activated by notes special dashed lines, the if statment tried to change a button that does not exist in the system and there for would give an error.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -7240,7 +7240,6 @@ function generateContextProperties()
       propSet.innerHTML = str;
 
     multipleColorsTest();
-
     }
 
 /**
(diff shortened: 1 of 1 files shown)


a21aldsa (changes requested):
works fine but you forgot to remove a log message in there: "fill colours". This PR also has _a lot_ of whitespace changes where you just seem to have added spaces wherever they would not change the functionality of the code. This is technically fine but eerily similar to indentation changes which we both know are not fine. 

Essentially, remove the log message and, if you have the time, remove/restore the random spaces you have placed/removed

c21adano:
I removed the log message and most of the auto formating done.

a21aldsa (approved):
its good now

Issue #14153, opened 2023-05-22
Add abillity to remove save in localstorage Note: after 14410


b21eriho:
Make it so that it is possible to remove saves in localStorage

PR #14231, opened 2023-05-24
G1 2023 v7 #14153

d21matbo:
Added functionality to remove a specific diagram from localstorage. You can now delete a save from the load diagram menu by pressing the delete button.

Since autosave almost always updates, there is no need for the user to be able to remove it.

image


c21adano (approved):
Works great!

Issue #14026, opened 2023-05-16
Unable to group-select sequence objects


b21phiro:
## Description
Right now the user can't group selecect sequence element types. 

Figure 1.
![bild](https://github.com/HGustavs/LenaSYS/assets/45258330/3abf7b8a-870d-44a7-963a-1a3331c97736)

## Job
Try to solve this issue, so you are able to group select these elements.

b21matfi:
I have found that you can select the sequence objects, but when you change the size of the object that's when it is not possible to group select them 
![image](https://github.com/HGustavs/LenaSYS/assets/102650424/59b62e49-a742-4466-8c66-830b01ec4ce5)

b21matfi:
it is only vertically aswell

b21matfi:
![image](https://github.com/HGustavs/LenaSYS/assets/102650424/9832df34-5bbb-4cf4-b84c-9cb4fa73f2d7)

the half circle is life line

b21matfi:
This issue is much bigger then just group selecting, this is about resizing all together, i found out that you are able to group select all the elements does not matter if it is sequence or any other element. The first issue is that resizing vertically resizes both up and down for height, does not matter if you pull the lower node or upper node. Second issue is that this is not visuall so you can not see if the height changed uppwards or downwards (depending on the node pulled) you can still select the elements but you have to start from a much heigher distance because the start of the elements is much higer or lower. Third issue i came accross when using (snap to grid) that it snaps wrong which can be something with the ghost behind it. This problem is only with resizing vertically not height.

b21matfi:
This issue should also be renamed because you are able to group the sequence objects but not correctly because they are offset like all the elements

PR #14237, opened 2023-05-25
Unable to group-select sequence objects #14026

a19clasv:
The title is misleading, you can group-select sequence objects. The problem is that resizing an element vertically changes the y position detected when group-selecting and makes it so that the visual location doesn't match up to the actual location needed to include in Box Selection. 

I've fixed that by storing the height when first resizing an element and changing the compared y position based on the difference with their current height. You still need the Box Selection box to be a bit above them, but that has nothing to do with resizing and was both there before I started and was possibly intended.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -3531,7 +3531,7 @@ function getRectFromElement (element)
             console.log("resizedY "+resizedY[i].y);
             return {
                 x: element.x,
-                y: resizedY[i].y,
+                y: resizedY[i].y * 2 - element.x - resizedY[i].resizedDifference,
                 width: element.width,
                 height: element.height
             };
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -2648,14 +2648,14 @@ function mmoving(event)
                 let foundID = false;
                 if(resizedY == undefined){
                     let resizedElement = structuredClone(elementData);
-                    resizedElement.y = elementData.y - heightChange * 0.45;
+                    resizedElement.y = elementData.y - heightChange * 0.48;
                     resizedElement['elementY'] = elementData.y;
                     resizedY.push(resizedElement);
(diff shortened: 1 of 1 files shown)


b21phiro:
But does it solve the main issue objective in that you can group these elements?

a19clasv:
You can group sequence objects.

a19clasv:
The box just have to be a bit above them if Box Selection is used to group them.

a19clasv:
I'll change that then.

b21phiro:
I tried by pulling from the branch and it seems to work okay. I'd allow it, but it's up to the tester.

d21matbo:
From what I can tell, it still behaves very strange. Testing with a `sequence-lifeline` the required offset above the elements increases as the element is expanded vertically downwards.

image

The image shows that the first `sequence-lifeline` is inside the selection (since it has lost its name), while the second `sequence-lifeline` is not inside the selection

d21matbo (changes requested):
In addition to the changes made to solve this issue not working, there is also some changes made to other parts of the system that I really do not see the need of, function `drawElement` to be specific.

Newly added `var` declarations and `console.log()` is also present.

a19clasv:
preResizeHeight might need to be a var, but I'll change the rest

a19clasv:
The console log isn't mine.

d21matbo:
Some changes made now allows the selection to happen on the resized `sequence-lifeline` but not the original one.
image

b21eriho:
While changes may not have been made by you but by the person you inherited the branch from please still attempt to resolve the faults. No matter who made them we do not want to merge them into the system. 

Though as said if you will not have the time to fix them the changes will be accepted since today is the final day

d21matbo (approved):
The group selection now works well, the gap required above is small enough such that it fits with the rest of the system. Very good job!

PR #14240, opened 2023-05-25
Removed unused code

c21adano:
removed the unused code that was used to hard code a diagram into the system and debugDrawSDEntity used to predraw a sd element.


a19clasv (approved):
It's removed.

Issue #14233, opened 2023-05-24
Strange behaviour when both template and example diagrams are toggled on


b21eriho:
If you activate both the A4 template and the example diagram dropdown it looks like this:
![image](https://github.com/HGustavs/LenaSYS/assets/129077727/5d34893d-de9e-4176-9678-68a63fdd808c)
The buttons "Vertical" and "Horizontal" should be placed below the button for the A4 template instead

PR #14242, opened 2023-05-25
Strange behaviour when both template and example diagrams are toggled on #14233

a19clasv:
Placed Vertical and Horizontal buttons under A4 template.
So it looks like this:
image
Instead of this:
image


a21aldsa (approved):
Looks good

b21eriho:
@a19clasv I messed up slightly in solving the merge-conflict. If you check in the changed files you can see that in the stable branch the buttons you moved have had their styling removed, please remove the styling in the same way on your branch

a19clasv:
Looks like this now:
image

a21aldsa (approved):
Looks fine, things still seem to work after the conflict fix

Issue #14232, opened 2023-05-24
A4 template in darkmode


b21eriho:
When the A4 template is activated in darkmode it looks like this:
![image](https://github.com/HGustavs/LenaSYS/assets/129077727/c9b46155-6dc2-4ab5-8546-578d19fd11ec)
The text that says A4 should be white

PR #14247, opened 2023-05-25
A4 template in darkmode #14232

a21oligu:
Fixed color of a4 template text when in darkmode. Solves #14232


b21eddka (approved):
Looks good!

PR #14249, opened 2023-05-25
G2 2023 v7 issue#14201

a21jeaha:
an explanation of what the file does can be found inside the created micro service.
the added code to dugga.js will be removed after the code is approved.

### DIFF
--- a/DuggaSys/microservices/courseedService/createMOTD_ms.php
+++ b/DuggaSys/microservices/courseedService/createMOTD_ms.php
@@ -24,9 +24,6 @@ $opt=getOP('opt');
 $motd=getOP('motd');
 $readonly=getOP('readonly');
 
-// $isSuperUserVar=false; 
-// $isSuperUserVar=isSuperUser(getUid());
-
 // Updates the message of the day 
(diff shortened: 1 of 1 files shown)


DuggaSys/microservices/shared_microservices/getUid_ms.php
```diff
@@ -11,7 +11,7 @@ function getUid(){
         if(isset($_SESSION['uid'])){
             $userid=$_SESSION['uid'];
         }else{
-            $userid="1";
+            $userid="guests";
```
b21kurar, line 14: I am a bit unsure of this change, some services require `1`. Since this issue is not related to this microservice, should this really be altered? Maybe it is needed for the MOTD microservice?

b21kurar (changes requested):
Just a small caveat about the `getUid_ms.php`

DuggaSys/microservices/shared_microservices/getUid_ms.php
a21jeaha: the problem is that there already exist a user with userid 1, and his a super user

image

b21kurar (approved):
all good!

PR #14252, opened 2023-05-25
removed code that redirects to the created microservices

a21jeaha:
All vestiges of the test redirects to the created micro services has been removed. 
dugga.js is now reverted to its original state.


b21hamwe (approved):
Seal of approval

PR #14263, opened 2023-05-25
Master g1 context pane fix | Null checking element.type being reassigned

a21aldsa:
There was a problem of not being able to change any attributes, like name etc to anything. This completely broke the context pane. This was fixed by surrounding the reassigning of element.type with an if statement making sure its not undefined first

Please look at files changed, there is only if statements added that checks its not null

This was done together with me (@a21aldsa), the group leader (@b21eriho), the issue leader (@b21phiro) and also @a19clasv


c21adano (approved):
Works, the problems that exist on master are not there any more.

PR #14266, opened 2023-06-15
removed vertical nodes from all elements except SE & UMLsuperstate

b21eddka:
Vertical nodes are only available when using the Sequence elements or the UML superstate


b21eddka:
Everything looks good. I believe there should be a separate issue created for: using the left and top nodes makes the resizing jump a few pixels.

PR #14267, opened 2023-06-16
Changed comment typo and variable declaration

b21eddka:
The comment typo aswell as "let" for the specified variable was changed in the funciton "makeGhost()".


b21eddka:
Everything works fine after the change.

Issue #14207, opened 2023-05-24
Create microservice getGroupValues


c21alest:
Related to: https://github.com/HGustavs/LenaSYS/issues/14143

Read more about the MS [here](https://github.com/HGustavs/LenaSYS/blob/1f6fc81f85fd7cadc4652cf44f52bfab88e2e496/backend-models/microservices/duggaSys%20services/Potetial_micro_services.md#getGroupValues)

**getGroupValues_ms.php**
Folder: **sectionedservice**

a22melst:
This issue shall be split into sub-issues.

### **From potential_micro_services.md:**
**getGroupValues**
Uses service selectFromTableGroup to get information it requires from group.

a22melst:
Looking into where the DB-table "groups" is used. Creating a group in a course only seems to add a record to the table "listentries".

PR #14269, opened 2023-06-25
Created getGroupValues_ms.php #14207

b21antbr:
Created and tested file for getGroupValues_ms.php

### DIFF
--- a/DuggaSys/microservices/sectionedService/getGroupValues_ms.php
+++ b/DuggaSys/microservices/sectionedService/getGroupValues_ms.php
@@ -6,88 +6,85 @@ include_once "../Shared/sessions.php";
 include_once "../Shared/basic.php";
 include('../shared_microservices/getUid_ms.php');
 
-
 // Connect to database and start session
 pdoConnect();
 session_start();
 
+$opt = getOP('opt');
 $courseid=getOP('courseid');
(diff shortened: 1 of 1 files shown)


b21antbr:
Tested by adding
`if(opt==="GRP"){$.ajax({
      url: "../DuggaSys/microservices/sectionedService/getGroupValues_ms.php",
      type: "POST",
      data: "courseid="+querystring['courseid']+"&coursename="+querystring['courseid']+"&coursevers="+querystring['coursevers']+"&comment="+querystring['comments']+"&opt="+opt+para,
      dataType: "json",
      success: returnedGroups
    });}`
To file dugga.js

Then check network ativity and seeing if the file gets used correctly. Later reverted changes in dugga.js so everything works as normal. 

The file now exists for the microservice, however the implementation of its usage has to be sorted in another way.

f22linhe:
Researching how testing should be conducted

f22linhe:
Manual testing following the suggestions mentioned above:
The status code of the request: 200 (OK), the resource was loaded as expected. Data is requested from the server when a new group activity is created.  There is an issue with the response from the database which seems to be traceable back to database.php. However, getGroupValues.ms itself appears to be working; the problem lies elsewhere.

Issue #14235, opened 2023-05-24
Extra functionality for "reset view" button


b21eriho:
If the "reset view" button is pressed when there are no entities in the system, the camera should be taken to the origin of the diagram i.e.  the same space it starts when you reload.

PR #14271, opened 2023-07-18
Extra functionality for "reset view" button #14235

b21willu:
Fixed issue when pressing "reset view" button. Before the camera didnt move when there was no entitites in the diagram. Now the camera moves back to the origin of the diagram when pressing "reset view" although there isnt some entities in the diagram.


b21willu:
Testing done! When pressing "reset view" the camera moves back to the origin of the diagram. Tested with and without different entities in the diagram and it works with both.

Issue #14120, opened 2023-05-22
Find a better way to store save-files


b21phiro:
## Description/Job
If you don't have anything loaded, the load button is still suggesting that it can be clicked on. Implement so that the button can't be clicked on AND have a visible clue on the button itself. 

Figure 1.
![bild](https://github.com/HGustavs/LenaSYS/assets/45258330/0eea8888-55f7-4508-b296-b4597c6a9f89)

## Tip
To add better UX, also use the cursor attribute using CSS.

PR #14274, opened 2023-07-31
Better indication for loading-button (if there's nothing to load) #14120

b21willu:
Added function to check if there isnt any saved diagrams, if its not then the load button doesnt show any modal and if there is saved diagrams the modal is shown. Added a count on the "load" button as a clue to users to see how many diagrams they have saved.


b21willu:
Testing done! When no diagrams are saved the button does not do anything. And a clue is added to show the user how many diagrams that are saved on the button.

Issue #13984, opened 2023-05-15
Hook up save button with storeDiagramInLocalStorage function.


b21phiro:
## Description
Right now we have a save button that can be clicked on, however it is not hooked up to anything yet. We previously made an issue about this but found out we allready have a function for this.

## Job
Hook the button upp with the function ``storeDiagramInLocalStorage`` in diagram.js

PR #14300, opened 2024-04-02
Fixed savebutton in diagrams.php (#13984)

b22chrmo:
Changed from comma to semicolon in 'onclick' on line 1106 in diagram.php as the functions were already implemented but did not work. The semicolon allows multiple functions to be entered inside 'onclick'.


a22ricni:
Have tested this issue and don't find any problem with this solution. I change the comma to semicolon in my own environment, and don't find any problem.

PR #14310, opened 2024-04-03
Cleaned and fixed

d22hanol:
Fixed the syntax issues within the lines and tidied up the code to follow the same style as the rest of the file.


b22chrmo:
After changes, code still works as intended. Code looks cleaner and easier to read after changes. Nothing different with the functionality before or after changes.

PR #14316, opened 2024-04-03
modified the installer

a22ahmom:
Replaced the ob_end_clean() method with ob_clean() method to only clear the current output without ending the entire output buffering.


a22oscja:
solves issue but causes some weird lag on my non-docker setup (fine on docker). Either way it should reasonably be correct to not turn off the buffer, but may need further looking into.

Issue #14293, opened 2024-04-02
Reset View creates two notifications


a22solmi:
## Description
When pressing the Reset View item from the toolbar two notifications with the message "Centered the camera" shows.
![issue-v1-2](https://github.com/HGustavs/LenaSYS/assets/129150661/9cda3caf-4594-4561-b3e9-d8f08093f0a5)

## Job
Find the issue and fix the issue so that only one notification appears.

PR #14318, opened 2024-04-03
Fixed issue where two notifications would pop up when centering in diagram.php. (#14293)

b22chrmo:
Onclick function for centering in diagram.php was called twice. #14293


a22hanfa (changes requested):
# [DuggaSys/diagram.php]
**LGTM**
Was able to recreate the bug. The fix by removing the extra `centerCamera();` worked on my machine as well. 

# Remove the multiple test commits
Someone could please remove the multiple test commits that have been added to this branch, at the moment it's causing branch conflicts that must be resolved. More specifically caused by `markdown-testfiles/lenasys-markdown.txt`.

Issue #14304, opened 2024-04-02
Can't access states inside superstates


a22jacka:
## Description
When creating a superstate over some already existing states, it's somtimes impossible to access the states inside it without moving the superstate. 

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/70148d15-181b-4572-8b13-9267916884f2)

## Job
Create a way to either click past the "filled" area or to move the superstate below the states

d22hanol:
If you create the super state before you create the object inside then you can access it, but if you create the super state after the inside object then it doesnt work. 
the same is true for the loop/alt for sequence diagram

Issue #14309, opened 2024-04-03
Can't place Sequence Activation direcly over Sequence Lifeline


a22jacka:
## Description
When trying to place an Activation direcly over a Lifeline it just highlights the lifeline instead of placing the Activation.

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/234f5a60-84ff-4d08-b0b0-471c03d1e04c)

## Job
Fix so that activations can be placed directly over lifelines

a22vioja:
It is possible to place activation correctly if first activation is placed outside the sequence actor rectangle. 

function entityIsOverlapping line 3592
this functions seems to work correctly, because when trying to place an activation, it does not give error message as it gives for other elements. But it does not let place activation either - need to check why.

It could be that CSS positioning of the activation box in the HTML structure is preventing placement of the object.

Event handler “ddown” on line 1798 could be the reason that prevents from placing on the lifeline.

One more guess - overlapping is disabled on purple rectangle around the entities. If so, it needs to be enabled but only for sequenceActorAndObject , not other entities. 

Still working on it.

a22jacka:
Nice findings. Could also maybe have a check for if the mouse has a lifeline or other entity selected and change the behaviour based on that.

Small note on this however:
>  it needs to be enabled but only for sequenceActorAndObject , not other entities.
Would be nice to have a more general solution to this since other elements, like the super state, would benefit to also be able to place elements directly on it

a22vioja:
I think that the problem could be on lines 1727 -1790 , function mdown(), because it handles mouse actions and updates states considering if the mouse button is clicked. On line 1727 code suppose to handle actions if mouse is clicked on selection box. The logic is to prevent placement of objects over existing elements by checking the initial position of the object, but instead of placing or preventing from placing the object over - it takes the element which is under the cursor and starts to interact with it. 

I also think that the implementation on mouseModes is missing (starts at line 1829) but  I was not able to find a solution for this.

a22jacka:
Solved by other issue

PR #14334, opened 2024-04-04
Fixed issue where states could not be accessed inside superstates. (#14304)

b22chrmo:
States can now be accessed inside superstates. Most elements now have a fixed z-index of 1. (#14304)


b22davbe:
Test ok. States can now be accessed inside superstates. A suggestion for improvement is to make it possible for creation of a state inside a superstate.

a22jacka:
A similair placement issue is being worked in #14309, their solution might work here aswell

PR #14337, opened 2024-04-04
G2 2024 v13 issue#14208

a22marca:
The microservice is created (getCourseGroupsAndMembers.php).

Calls are now made from Dugga.js. A list of related members of groups to a specific courseid and coursevers is returned.

Check issue#14208 for more info.


b21isabe:
testing has been very unreliable, but on initial viewing it seems like the correct json list is returned, I will continue testing on monday

my first thoughts after looking at the code:
lines 19-23 could be done by getuid()

the if statement on line 41 "if(strcmp($opt,"GRP" is redundant since it already exist in dugga.js I tried removing it and the code still worked the same afterwards

b21isabe (changes requested):
From my manual testing through the browser, it seems that the original code wasn't functional and didn't actually manage to retrieve any members of groups, the array $grplist would remain empty after the "GRP" opt was called, this didn't actually matter since the code wasn't actually used in any meaningful way on the website. These issues seems to be fixed by the created microservice that now returns the $grplist array with the expected content. 

Things that could be improved:

I wrote earlier in a comment that the if-statement on line 41 should be removed, after looking into it some more I don't think it is necessary since they do inspect different things ("kind" and "opt"). I would suggest leaving it as it is right now since it's not doing any harm and removing the line now could maybe cause issues in the future if the GRP kind would get new opts. 

lines 19-23 could be done by getUid() but I have heard some say that it is currently not working, if that is the case it can be left as it is and maybe be updated at a later time.

Before merging, dugga.js must be restored to use the sectionedservice.php

a22iliru:
waiting for a22marca to implement request, then merging.

a22marca:
- The url in dugga.js has been changed back to use sectionedservice.php
- getUid_ms is currently not using a REST interface, to use this the file would have to be included and the function called. We should probably change this microservice first before adding this in getCourseGroupsAndMembers.
- Good and thorough feedback on the `GRP` selection. I will leave it as it currently is.

Issue #14322, opened 2024-04-03
Course github repo is not being saved. 


a22willi:
After adding a github repo to a course. It does not show up the next time you try to modify it.

Adding github repo.
![image](https://github.com/HGustavs/LenaSYS/assets/129264968/910220c6-dee3-4357-b95b-d8f09e312ee6)
The next time you edit, it is no longer there. 
![image](https://github.com/HGustavs/LenaSYS/assets/129264968/ec6098ac-7265-44dc-913f-ccdf1e5f90c7)

[sqlite diagram](https://github.com/HGustavs/LenaSYS/wiki/Github-Integration#sqlite-1)

a22moamy:
The bug has now been fixed. Additionally github URL should be logged as well. As a side effect, courses with no GitHub URL will display the default null value in the "GitHub URL" field from the edit course window. This null value has to be removed when updating a course. Adding a correct github URL will still make the field red, but it has the right functionality and does not affect saving the URL.

HGustavs:
Unclear about the details so I will give a short overview of my requirements regarding this functionality.

A repository is not necessary (if there is no github project for a course).
It should be possible to change or enter a github repository using the same dialog.

PR #14340, opened 2024-04-04
G3 2024 v1 #14322

a22natth:
Fixed bug where course GitHub URL was not saved nor displayed correctly when editing a course. With the current fix, the course GitHub URL will now be saved to the database and correctly fetched with a modification of a database query. OBS, if no GitHub URL was saved for a course then "null" will be displayed in its corresponding field. This has to be removed when updating a course. Additionally, when creating a course, the GitHub URL field will always be red regardless of input, although this does not affect the functionality of the whole process.

### DIFF
--- a/DuggaSys/courseed.js
+++ b/DuggaSys/courseed.js
@@ -40,25 +40,27 @@ function updateCourse()
 	// Show dialog
 	$("#editCourse").css("display", "none");
 	
+	// Updates the course (except the course GitHub repo. 
+	// Course GitHub repo is updated in the next block of code)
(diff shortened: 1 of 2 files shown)


b22marah:
Pull request will be closed since Github URL loads with a default value of null instead of being empty when a github URL has not been given which can be seen from the following image.

image

This needs to be fixed before we can merge the pull request.

b22marah:
Pull request will be closed since Github URL loads with a default value of null instead of being empty when a github URL has not been given which can be seen from the following image.

image

This needs to be fixed before we can merge the pull request.

PR #14346, opened 2024-04-04
G3 2024 v1 #14322 FIXED

a22moamy:
Fixed bug where course GitHub URL was not saved nor displayed correctly when editing a course. With the current fix, the course GitHub URL will now be saved to the database and correctly fetched with a modification of a database query. Please note that when creating a course, the GitHub URL field remains red regardless of input, although this does not affect the functionality of the whole process.

Also fixed the bug where "null" or "UNK" was displayed in the URL field for courses without a GitHub URL.


a22edvbu (commented):
Funkar galant!

PR #14350, opened 2024-04-04
G3 2024 v1 #14241

### LINKED ISSUE #14241: Timer for refresh button is not consistent
The timer to trigger the refresh button does not always work on a 10min timer(as per design), sometimes it refreshes after just 60sec. 
Pull request discovering this #14225 
Issue that implemented the feature #13987

a22edvbu:
Due to Async calls the console logs was printed in the console before the Async call was completed for all the objects. Therefore we had to wrap the whole Async call into a Promise and then use the Promise from setInterval to ensure that the Async call was completed in its entirety before logging the success.

closes #14241

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -3097,24 +3097,28 @@ function hasGracetimeExpired(deadline, dateTimeSubmitted) {
 // Function to fetch code examples for a specific lecture/moment
 function createExamples(momentID, isManual) {
   lid = momentID;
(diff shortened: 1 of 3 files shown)


a22moamy (approved):
The code works as specified. No alert is generated and only the console is updated when a moment is refreshed. OBS. It can seem as if this pull request does not work at first. To make it work, remove the web browsers cache and cookies. 

The output generated in the console is as follows:
Skärmavbild 2024-04-05 kl  12 39 57

Issue #14315, opened 2024-04-03
Show instructions when trying to refresh GitHub integration. 


a22willi:
When clicking the button `download github repo` in a course where no github repo has been set, the following message appears:
![image](https://github.com/HGustavs/LenaSYS/assets/129264968/690e2f26-b4ac-40f9-885a-c6c07fb83663)

According to the customers instructions, this should be modified to show instructions or a text box where a github repo can be added instead.

PR #14354, opened 2024-04-05
G3 2024 v1 #14315

a22oscja:
Essentially adds an html window for adding a github repo which opens when trying to download a repo from a course that does not have one.

For testing i'd recommend setting shortdeadline and longdeadline in gitcommitservice.php to 0, as otherwise there is a cooldown of 10 minutes on the button which opens the window.


a22willi:
Me and @b22marah missed that the files under courses in the installer are used to create the testing course. As such the installer fails when selecting ` Include Testing-Course.`.

a22chrfa:
Code looks good and the changes seems to work exactly as intended in the browser. All of the requirements from the issue seems to be fulfilled.

Issue #14338, opened 2024-04-04
Remove unused files


a22willi:
Remove test files, files that will not be used and so on in DuggaSYS.

PR #14359, opened 2024-04-05
G3 2024 v1 #14338 - Removal of unused files

a22natth:
The issue #14338 surrounded the removal of unused files in the "DuggaSYS" folder, to which we followed by removing files we deemed unnecessary or unused. Although the assigned scope was the "DuggaSYS" folder, we expanded the clearance on more parts of the "LenaSYS" repository. The targeted files were either not being referenced to and therefore not serving a purpose, or contained text or code which seemed to be used for creating test-commits. Other files that we discarded were old markup examples (such as list or table examples), which were in the form of code or text, rather than HiFi models for instance. Due to this, the files were deemed to be unnecessary. We made it a point to avoid messing with all diagram files as this is not our groups area.


a22willi:
Me and @b22marah missed that the files under courses in the installer are used to create the testing course. As such the installer fails when selecting  Include Testing-Course..

PR #14361, opened 2024-04-05
G3 2024 v1 #14338 FIXED

a22moamy:
Added back files in install > courses and install > md. This is to fix the bug where the installer would stop working when ckecking the Include Testing-Course.


a22willi:
There seems to be something wrong with this PR, all project files are included.

PR #14362, opened 2024-04-05
14155 create microservice createnewlistentrie

c22cargo:


a22melst:
Just noticed that the logging is not performed since $username is never retrieved from anywhere. Will look into this further next week.

PR #14363, opened 2024-04-05
G3 2024 v1 #14338, again

a22natth:
The issue https://github.com/HGustavs/LenaSYS/issues/14338 surrounded the removal of unused files in the "DuggaSYS" folder, to which we followed by removing files we deemed unnecessary or unused. Although the assigned scope was the "DuggaSYS" folder, we expanded the clearance on more parts of the "LenaSYS" repository. The targeted files were either not being referenced to and therefore not serving a purpose, or contained text or code which seemed to be used for creating test-commits. Other files that we discarded were old markup examples (such as list or table examples), which were in the form of code or text, rather than HiFi models for instance. Due to this, the files were deemed to be unnecessary. We made it a point to avoid messing with all diagram files as this is not our groups area.

Added back files in install > courses and install > md. This is to fix the bug where the installer would stop working when ckecking the Include Testing-Course. Also fixed the issue where all files were marked as modified due to installing LenaSys again while testing the changes.


a22ahmom (commented):
It seems like most unnecessary test files have been removed from DuggaSYS and files that contain unnecessary code so the overall issue seems to be solved. One or two files are left out as test files for future students to test their code but that is fine. So nicely done.

PR #14369, opened 2024-04-05
getGroupValues_ms.php #14207

a22melst:
Tested by adding the following code to the function AJAXService in dugga.js

	if(opt==="GRP"){
		$.ajax({
			url: "../DuggaSys/microservices/sectionedService/getGroupValues_ms.php",
		});
	}

Note: There is a problem with getting the user info for SQL inside of database.php causing warnings and errors. However I do not think this is due to faults inside of getGroupValues_ms.


c22cargo:
I could not get this implementation to work, but I don't know if that is just because I am testing it wrong or not. Regardless, I still think we should merge it and revisit it if needed.

Perhaps the $groups variable should be returned?

Issue #14314, opened 2024-04-03
Inconsistent keybinds


a22solmi:
## Description
Pressing 1 selects the first item in the toolbar, 2 the second and so on. However from 5 and up this pattern is not followed.
Furthermore pressing the same number several times when on an item that has a submenu should cycle between items in that submenu.

## Job
Change the keybindings to select items from the toolbar in order. Implement a way to cycle between items in a submenu when pressing the same number repeatedly.

a22solmi:
Keep track of cycling progress in #14119

a22solmi:
Also update DuggaSys/diagramkeybinds.md to reflect changes

PR #14373, opened 2024-04-05
Updated keybinds, added cycling #14314

a22solmi:
Updated keybinding to reflect order in toolbar.
Added cycling between items in all sub menus when pressing the same keybind.


a22alfna (approved):
I have tested the keybinds and made sure the cycling of multiple toolbar items with the same keybind works as it is supposed to.

Issue #14325, opened 2024-04-03
[Research] Box-select not working properly with super-states


a22jacka:
## Description
Similarly to [](https://github.com/HGustavs/LenaSYS/issues/14317), when using box select to highlight a super-state the highlight needs to be super big in order to highlight it. This time it doesn't appear to be because of margin, might be but unsure. 

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/d08c3859-95bf-4e50-b626-e4507892d43c)
The black box is the highlight

## Job
Find out why it doesn't work properly and write a comment with your findings

a22solmi:
Solved by #14377

PR #14377, opened 2024-04-05
Solved issue where box-select wouldn't properly select items. (#14325)

b22chrmo:
Box selection wouldn't properly select items inside it's area. This issue was created specifically for super-states, but more diagram-types had similar problems. Issue is now solved and you can properly select items with the box selection.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -3432,10 +3432,12 @@ function getRectFromElement (element)
                 resizedY += (element.height - preResizeHeight[i].height)/2
             }
             // Corrects returned y position due to problems with SE types
+            
             let elementY = resizedY;
             if(element.type == "SE"){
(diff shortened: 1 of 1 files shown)


d22hanol:
Tested without the branch first and found that you could select the superstate when greatly exaggerated size on the box-select, this was not a good way to do it, tried to replicate the issue with a few other entities. I found problems with the er-relation.
When switching to the new branch all of these problems disappear.
My opinion is that he solution for this problem is appropriate.

d22hanol (commented):
Tested without the branch first and found that you could select the superstate when greatly exaggerated size on the box-select, this was not a good way to do it, tried to replicate the issue with a few other entities. I found problems with the er-relation.
When switching to the new branch all of these problems disappear.
My opinion is that he solution for this problem is appropriate.

PR #14379, opened 2024-04-08
Updated event listener structure

a22solmi:
The structure of "keyup" was updated and a formatting issue was corrected.


a22cazlu:
I have tested the all funktions I can think of and everything seems to work fine.

PR #14382, opened 2024-04-08
Moved starting of session

a22moamy:
@a22willi found the solution was to start the session before including any files.


a22danwe (commented):
The issue semms to be fixed, there are no error popups when "show tests" are clicked.

Issue #14339, opened 2024-04-04
Save github repo from info popup


a22willi:
In [Issue-#14315](https://github.com/HGustavs/LenaSYS/issues/14315) a popup was created. This popup currently has no functionality, make it save the github repo URL in the proper location. 

Since the previous pull request has not yet been approved, I suggest waiting untill it has been approved and merged with the weekly branch before you try to do any development.

Also check out this [issue](https://github.com/HGustavs/LenaSYS/issues/14322) where I think they also worked on saving the github url.

PR #14387, opened 2024-04-08
G3 2024 v1 #14339

a22ahmom:
The popup now has functionality and when the user types in a github repo URL it gets saved in the database file metadata2.db in the gitRepos table. When clicking the download github repo button for the second time a minor error pops up but this only happens sometimes not when doing it for the first time for a new course.


a22edvbu:
All checks out!

a22willi:
A few things need fixing, it does not seem like the data is stored properly. And the cooldown notification appears, even when there is no gihub repo saved. It should prompt the user to enter a github repo every time, if there isnt already one when the "download github repo" button is clicked.

Issue #14345, opened 2024-04-04
Errors in diagram.css 


d22hanol:
### Description 
JETbrains IDE complains about several syntax errors in diagrams.css. Could be related to other issues such as [this one](https://github.com/HGustavs/LenaSYS/issues/14283) and by proxy [this one](https://github.com/HGustavs/LenaSYS/issues/12693)
Testing is very important for this issue as it could break a lot of stuff if it is incorrect
### JOB
Look into and potentially fix diagram.css.

Issue #14283, opened 2024-04-02
Entire toolbar not visible


a22jacka:
# Description
The toolbar uses a fixed size so that on certain screens the toolbar can be cut off with no way to access the lower icons. 
This is most apparent with different screen sizes/scaling

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/3059c2fb-335e-4a49-8001-f743ea17e026)
![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/1f3127ba-2695-413c-9e6d-2db87086dea5)

# Job
Implement a fix. Could be to either make the toolbar resize or add a scroll-feature. You could try both and check which works best

a22jacka:
This has been looked at earlier in issue [12693](https://github.com/HGustavs/LenaSYS/issues/12693) and different fixes seem to introduce new problems. The whole toolbar likely needs to be reworked a bit

d22hanol:
Adding an overflow value to the toolbar makes the popup toolbar and tooltip appears behind the diagramwindow/canvas  instead of in front of it. Did not manage to find a sufficient solution to fix this issue.

PR #14388, opened 2024-04-08
G1 2024 v1 #14345

d22hanol:
fixed the errors in the css-file but it did not fix the related toolbar issues

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -23,12 +23,12 @@
     width:100%;
     height:100%;
     user-select: none;
-    overflow: none;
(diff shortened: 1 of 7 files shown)


a22vioja (approved):
It is good that pixels are used instead of units, so the size of an element remains constant regardless of the device or screen resolution.

Since "overflow: none" has no effect, it is good that you changed it to "overflow:hidden" so that container remains intact but overflowing content is hidden. 

I do not see any new problems on the diagram behaviour which can be affected by these changes, but the main issue #12693, #14283 remains, that the whole toolbar is not visible on the smaller screen.

Issue #14385, opened 2024-04-08
Followup on #14228


a22jacka:
## Description 
Issue #14228 was completed but was never merged

## Job
Reimplement the solution so that we can merge it

Issue #14228, opened 2023-05-24
Default name for save issue


b21eriho:
The default name for issues handles numbers starting with zero incorrectly in the timestamp see image:
![image](https://github.com/HGustavs/LenaSYS/assets/129077727/cd97a95e-a234-443c-9368-8511a0df835b)
Resolve this

b21eriho:
To clarify, it reads 15:0:34 where it should 15:00:34 and it also reads 15:7:36 where it should read 15:07:36

PR #14389, opened 2024-04-08
Fix formatting for datetime. Issue #14385 (formerly #14228 )

a21liltr:
Solves issue #14385 (#14228 ).

Solution originally from another developer that had not had their work merged.


a22ricni:
Have tested it and it works

Issue #14303, opened 2024-04-02
Create example state diagram


a22jacka:
## Description
There is no example diagram for state diagrams

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/01ab03cc-5c05-4769-8b29-f80fe92d41ba)

## Job
Create a state diagram that can be used as an example

d22abdha:
![statediagram](https://github.com/HGustavs/LenaSYS/assets/58794335/084d1a65-75d8-4cac-bf9f-716c22e9e58c)

I detta diagram är vattenkokare example diagram, den har vi tre tillstånd: "Avstängd", "Uppvärmer" och "Automatisk avstängd". Vattenkokaren startar i "Avstängd". När strömbrytaren aktiveras, flyttas systemet till "Uppvärmer", och om temperaturen når 100°C, går den över till "Automatisk avstängd". Och strömbrytaren stängs av när som helst, återgår vattenkokaren till "Avstängd".

a22solmi:
This is not proper syntax for a state diagram. Please choose an example diagram from the Software Engineering course book instead. It should also be written in English to be consistent with other examples.

d22abdha:
[State diagram.drawio (1).pdf](https://github.com/HGustavs/LenaSYS/files/14878879/State.diagram.drawio.1.pdf)

"1. Waiting State: The microwave starts in this state, where it displays the time. This is the default state when the microwave is not in use.

2. Full Power and Half Power States: If the user selects either full power or half power, the microwave transitions to those states. In "Full Power," the microwave sets its power to 600, and in "Half Power," it sets the power to 300.

3. Set Time State: After choosing the power level, the user sets the time by entering a number (duration for cooking). Once the number is entered and the time is set, the microwave transitions to the "Enabled" state if the door is closed. If the door is open while in this state, it will move to "Disabled" state where it displays 'Waiting'.

3. Enabled State: In this state, the microwave displays 'Ready', indicating that it is ready to start operation. The microwave will remain in this state as long as the door remains closed.

4. Operation State: When the user starts the microwave (presumably by pressing the start button), it transitions to the "Operation" state, where the actual cooking or heating operation takes place.

5. Transitions Back to Waiting: If at any point the door is opened during the operation, or if the operation is canceled, the microwave returns to the "Waiting" state."

a22solmi:
[...]

PR #14390, opened 2024-04-08
Fixed issue #14303 to load a State diagram example in diagram.php file

a23ponek:
Its now possible to load a example state diagram in the diagram.php file.

Changed files: 

diagram.php (Line 1028): <option value="JSON/StateDiagramMockup.json>State diagrams</option>

Added file "StateDiagramMockup.json to JSON file in DuggaSys, which was created from diagram.php tool.


a21liltr (commented):
Loading a State Diagram functionality is implemented and works on Chrome, Firefox, Edge. It also works when there's already an existing diagram. Loading a State Diagram will wipe the previous diagram without saving it and then replace it with itself.

a21liltr (approved):
Approved and ready to merge and close.

a22solmi:
Full diagram is not implemented. Empty board preset should not be removed.

Issue #14341, opened 2024-04-04
Correct if-block


a22jacka:
##  Description
Several if-blocks in `diagram.js` have only 1 `=` when they should have 3 or 2. The rows might've changed is this issue is old.
The if-blocks are:
`if(document.getElementById("a4Rect").style.display = "block")` on row 5973
`if(document.getElementById("vRect").style.display = "block")` on row 5981
`if (element.kind = 'UMLRelation')` on row 6668
`if(element.kind = 'IERelation')` on row 6723

## Job
Add the extra equal signs. If things break with the new `=` signs, then just write a comment with your findings and we'll make a new issue for it

EdvinWerner:
The diagram.js contains only 2158 lines of code therefore this issue doesn't seem to exist and I recommend that it should be closed

PR #14397, opened 2024-04-08
G1 2024 v1 #14341 - Changed equal logic

a23ponek:
Changed equal logic signs from "=" to "==" on row 5973, 5981, 6668 and 6723 in diagram.js file. Things did not seem to break on change.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -5970,7 +5970,7 @@ function setA4SizeFactor(e){
 function toggleA4Horizontal()
 {
     document.getElementById("vRect").style.display = "block";
-    if(document.getElementById("a4Rect").style.display = "block"){
+    if(document.getElementById("a4Rect").style.display == "block"){
         document.getElementById("a4Rect").style.display = "none";
     }
 }
@@ -5978,7 +5978,7 @@ function toggleA4Horizontal()
 function toggleA4Vertical()
 {
     document.getElementById("a4Rect").style.display = "block";
-    if(document.getElementById("vRect").style.display = "block"){
+    if(document.getElementById("vRect").style.display == "block"){
         document.getElementById("vRect").style.display = "none";
     }
 }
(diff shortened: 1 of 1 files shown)


d22samma (commented):
# Review - Testing #14397 
Looks okay.

-  = is used for assigning the value due to the variable.
-  == , === are used for comparison purposes. 
- == compares two variables irrespective of data, a type. 
- === compares two variables in a strict check, it checks for data and returns true or false.

## Equal logic (Before changes)
In this case, _if statement_ want to asign a value instead of comparing the  equallity in meaning, This can lead to bugs and problems of the functions inside the code.
```
    if(document.getElementById("vRect").style.display = "block"){
        document.getElementById("vRect").style.display = "none";
    }
```
*Code in DuggaSYS/diagram.js  (Before changes)*
```
              else if (element.kind = 'UMLRelation') {
                //ID MUST START WITH "elementProperty_"!!!!!
                for (const property in element) {
                    switch (property.toLowerCase()) {
```
*Code in DuggaSYS/diagram.js  (Before changes)*


## Equal logic (After changes)
(==) Equal Logic seem added correctly inside the _if statements_.  
_=_ is placed correctly inside the _if statement_ functions.
_=_ sets the values when  _if statements_ equals true  (_==_) .
```
    if(document.getElementById("vRect").style.display == "block"){
        document.getElementById("vRect").style.display = "none";
    }
```
*Code in DuggaSYS/diagram.js  (After changes)*
```
              else if (element.kind == 'UMLRelation') {
                //ID MUST START WITH "elementProperty_"!!!!!
                for (const property in element) {
                    switch (property.toLowerCase()) {
```
*Code in DuggaSYS/diagram.js  (After changes)*

## Review
- File changes does not seem to harm any other code.
- Code changes shall minimize bugs and problems. 
- If statements shall get a better equal logic by those changes.
Changes looks okay.

b22chrmo:
After changes, there's a new bug where both the vertical and horizontal A4 templates are shown. This bug only happens if you click the 'horizontal' option first when you select the A4 template option. Every change after the first template-switch fixes the issue, but there is still an overlap of templates at first if you choose the 'horizontal' option.

a23ponek:
Noted, can you please provide video or images of the bug.

PR #14400, opened 2024-04-08
Fixed further issue #14303 to load a State diagram example

a23ponek:
Its now possible to load a example state diagram in the diagram.php file.

Changed files:

diagram.php (Line 1028): <option value="JSON/StateChartDiagramMockup.json>State chart diagrams </option>

Added file "StateChartDiagramMockup.json to JSON file in DuggaSys, which was created from diagram.php tool.

The State diagram example is now fully implemented and the Empty board example still exists as dropdown in the diagrams.php file.


a22robko (commented):
I have tested your code, and it looked good. What I did was first download the branch as a zip file, converted it to a folder, and placed it in htdocs. Then, I opened it in Visual Studio Code. This was because I needed to switch the branch that was created for the issue. After this, I modified the statechartdiagramMockup.json and noticed that the dropdown was still there, but when clicked on, nothing happend. Later, I removed the statechart from the mockup buy the code and saw that it was no longer there. I changed the code back to its original version and explored diagram.php to see how it looked and checked if everything worked as it should, and it looked good.
The dropdown result:
Screenshot 2024-04-09 at 11 35 24

In this picture can you see that the statechart diagram is no longer available.
Screenshot 2024-04-09 at 11 10 29
This is the statechart diagram when you press on load then will this diagram appear.
Screenshot 2024-04-09 at 11 35 41
 When I clicked on load, the state chart diagram appeared.

a22solmi:
Everything is good, but the digram needs to be cleaned up. Please resize the boxed to prevent wrapping of text and try to organize the boxed so that the arrows look better.

PR #14402, opened 2024-04-08
Added tooltip to GitHub-button

b22chran:
Added a tooltip to "dowload GitHub Repo"-button in navbar. Tooltip includes time for latest repo fetch and a cooldown until next download is available. However, in order for this to display properly the site needs to be refreshed. Tooltip is also designed for both light and dark mode.


a22willi:
Sometimes it shows up in the wrong location, make sure you test it at different window sizes, and with different content inside.
image
Sometimes it also shows a negative number like this? Maybe if the course does not have a github repo?
image
Either way, it should never show a negative number.

Issue #14367, opened 2024-04-05
Enlarge section that opens the option


a22jacka:
## Description
To open the options panel (with mouse) you need to click in the area highlighted in blue, meaning if you click too far left it wont open. A small test showed that adding `padding-bottom: {value} {unit};` to the div with id "options-pane-button" fixed it.

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/667cfdb2-0f4b-4e9b-ac52-b12b6a180541)

## Job
Add the padding to the text so that you can click to the left of the text. Find a value that fits well for all screen sizes.

PR #14405, opened 2024-04-08
Fix issue #14367

a22robko:
I fixed the button issue where it wasn't able to click on it at the left side (Issue #14367). I added some extra space at the bottom of the button by using a "padding 6px". Now it works better and you can click on the button even on the edge. I checked to make sure it looks good on the full option screen sizes.


a22solmi:
Changes to digram.js is not related to this issue.

Issue #14336, opened 2024-04-04
Fix undefined variable and other potential errors in test.php


a22iliru:
From a22marca:

I'm also having some problems when trying to run a test for the microservice getCourseGroupsAndMembers (sectionedService). For this im using: LenaSYS/DuggaSys/tests/sectionedservice_Test.php. 

All current errors are related to this file, e.g.:
 Warning: Undefined variable $QueryReturnJSONbefore in /var/www/html/LenaSYS/Shared/test.php on line 129

Seems to be some typos on some variables, and some are not declared and used correctly

test.php
Folder: /Shared

a22albjo:
Possible issues with the code:

1. There is an example on how to run the test from an other file starting from row 3 - 53. On row 20 and 37 an absolute path is used. This can cause confusion and relative paths needs to be used. The url to the services in the file is not even in use.
2. $QueryReturnJSONbefore is used on row 129 but not defined until row 146.
3. On row 163 there is an absolut path used to cms.webug.se. This should be relative.

a22albjo:
Possible solutions to the issues listed above:
1. Change absolut filepaths to relative.
2. Move the definition of $QueryReturnJSONbefore to line 126.

a22albjo:
@a22marca Do you see other possible causes?

a22marca:
@a22albjo 
I have been taking a look at the QueryReturnJSONbefore variable, seems like that is the first thing we need to address. I took a look at the outputs of `git log -L1,160:Shared/test.php` and there doesn't seem to be a earlier declaration that was removed or anyhing, as far as i can see.

a22marca:
Looks like it actually used to be declared here:
```
       if (strpos($option, 'query-before-test') === 0) {
-               $QueryReturnJSONbefore[$option] = doDBQuery($value);
+
+                // If service data !query-test! replace with actual query output
+                foreach($data as $sInput => $sValue){

```

When i add this line (function needs these parameters, not only $value), some errors occur. Looks like the parameters are not containing the right values? Running test at: localhost/LenaSYS/DuggaSys/tests/sectionedservice_Test.php
```
$QueryReturnJSONbefore[$option] = doDBQuery($value, $data, $testData, $option);
``` 
```
Warning: Undefined array key "variables-query-before-test-1" in /var/www/html/LenaSYS/Shared/test.php on line 57

Warning: Undefined variable $pdo in /var/www/html/LenaSYS/Shared/test.php on line 63

[...]

PR #14406, opened 2024-04-08
Update fileedservice.php

a22fokol:
The variable " 'filePath' => $filePath " was coming wo time in the same line code and this was causing some error . But it has been fixed now. 

But there is a problem coming  at the end line of this code, i mean at the closing  break. 

echo json_encode($array);
logServiceEvent($log_uuid, EventTypes::ServiceServerEnd, "fileedservice.php", $userid, $info);
?>

### DIFF
--- a/DuggaSys/fileedservice.php
+++ b/DuggaSys/fileedservice.php
@@ -288,7 +288,7 @@ if (checklogin() && $hasAccess) {
             'filesize' => json_encode(['size' => $row['filesize'], 'kind' => $filekindname]),
             'type' => $path,
             'uploaddate' => $row['uploaddate'],
-            'editor' => json_encode(['filePath' => $filePath, 'kind' => $filekind, 'filename' => $filename, 'extension' => $extension, 'showeditor' => $showEditor, 'filePath' => $filePath]),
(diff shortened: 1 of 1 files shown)


a22marca (changes requested):
This does not seem to address the issue that is described. From my knowledge of another issue regarding `test.php`, the mentioned code from in lines 162 to 167  in`test.php` creates a url to the files hosted on cms.webug.se if the service name does not contain a "/". So e.g. a test in `sectionedservice_Test.php` having the service entry "sectionservice.php" will be built into a `cms.webug.se/.../sectionedservice.php` url. 

We should probably use the file name as entry under "service" in a test, but retrieve the local path to it.
The code in test.php 162 to 167 should probably changed to use the function `dir(__FILE__)` to get the local path to e.g. `sectionedservice.php`. My opinion is that we should wait for the issue regarding test.php to be merged into master and then we can continue work on this issue on a new branch.

This is a rather involved issue, since it is related to the issue fixing test.php. Here is the issue regarding [test.php](https://github.com/HGustavs/LenaSYS/issues/14336).

PR #14418, opened 2024-04-09
Enlarge section that opens the option

a22robko:
I've modified line 120 as I fixed the button issue where it wasn't able to click on it at the left side (Issue #14367). I added some extra space at the bottom of the button by using a "padding 6px". Now it works better and you can click on the button even on the edge. I checked to make sure it looks good on all the full option screen sizes.

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -117,6 +117,7 @@
     transform: translate(20px,0px) rotate(90deg);
     cursor: pointer;
     width: 780px;
+    padding: 6px;
 }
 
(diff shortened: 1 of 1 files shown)


a22solmi:
You have made changes to a unrelated markdown file. Please make a new pull request with only relevant files.

b22chrmo:
Tried to recreate problem in local environment. The clickable area is smaller than size of the 'Show options panel'. After changes, you can now click and open the 'Options panel' by clicking anywhere on the 'Show options panel'. It now works as intended.

PR #14427, opened 2024-04-09
Updated the popup functionality by saving the record to the database G3 2024 v1 #14339

a22ahmom:
The popup now has functionality and when the user types in a github repo URL it gets saved in the database file metadata2.db in the gitRepos table. When clicking the download github repo button for the second time a message is shown to inform the user that it is too soon since the last update so they must wait. Also when returning to homepage and when clicking the edit course button the github repo link is already there or is saved. Sometimes a minor error pops up but this only happens sometimes not when doing it for the first time for a new course. So the overall solution seems to be working for this issue.


b22chran:
The function for saving a repo to a course is possible from the pop-up when trying to download a repo and there is no current repo saved for the course. When saving the repo it doesn't close the pop-up or giving a message that it is saved though. That can confuse the user. The repo is still saved to the database as it shows up when clicking on the edit-button (the cogwheel) in the overview for the courses (courseed.php). The problem now seems to be when the cooldown for the download-repo is refreshed and trying do perform a download, the pop-up for entering a a new repo shows up again.

DuggaSys/courseedservice.php
```diff
@@ -502,6 +502,32 @@
...
+								if(!$query->execute()) {
+									$error=$query->errorInfo();
+									$debug="Error updating entries\n".$error[2];
+								}
+							}
+						}
+		}
 
 
 	}
```
a22willi, line 533: Fix indentation.

DuggaSys/sectioned.js
```diff
@@ -3738,6 +3804,21 @@ function validateForm(formid) {
...
+      if(fetchGitHubRepo(repoLink)){
+        AJAXService("SPECIALUPDATE", {	cid : cid, courseGitURL : repoLink }, "COURSE");
+        //AJAXService("UPDATE", {	cid : cid, courseGitURL : repoLink }, "COURSE");
+        localStorage.setItem('courseGitHubRepo', repoLink);
+        $("#githubPopupWindow").css("display", "none"); 
+        updateGithubRepo(repoLink, cid);
+      }
+    }
+  }
 }
```
a22willi, line 3822: Fix indentation, tabs have been used for the rest of the file.

DuggaSys/sectioned.js
```diff
@@ -484,6 +484,72 @@ function refreshGithubRepo(courseid, user)
...
+		},
+		error: function(data){
+			//Check FetchGithubRepo for the meaning of the error code.
+			switch(data.status){
+				case 422:
+					alert(data.responseJSON.message + "\nDid not create/update course");
+					break;
+				case 503:
+					alert(data.responseJSON.message + "\nDid not create/update course");
+					break;
```
a22willi, line 543: These should probably return different error messages. [HTTP codes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status)

DuggaSys/sectioned.js
```diff
@@ -484,6 +484,72 @@ function refreshGithubRepo(courseid, user)
...
+				default:
+					alert("Something went wrong...");
```
a22willi, line 511: These should probably return different error messages. [HTTP codes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status)

a22willi (changes requested):
Some general formatting changes.

a18chrkr (commented):
For me, the saving of a github URL to a course through the pop-up window seem to work sporadically. When clicking the "save" button in the pop-up window nothing happens, and the pop-up remain open. However, when I go back to the course list and open the "setting" pop-up, the correct linked github-repo is there.

DuggaSys/sectioned.js
a22willi: case 422 and 503 could also be handled like this:
```
 switch($type) 
 {
     case 1:
     case 2:
     case 3:
         print 'success';
     break;
     case 4:
         print 'success for type 4';
     break;
 }
```

a22oscja:
files have been formatted, also some better errors have been given.

Both of us fail to reproduce the window not closing, this is assumed to be a testing error as the window closes inside the function that is called instantly upon the button being clicked (validateForm() in sectioned.js), meaning it should be impossible to have that happen without an additional error, assuming text was written at all in the input field.

Aside from that I would assume it could have seemed to work randomly due to the field having been saved from a previous usage of edit course. Make sure to check the metadata.db entry while testing as that is the more relevant part.

a22willi (approved):
Looks good now.

Issue #14352, opened 2024-04-05
Modify "download github repo" tooltip


a22willi:
![bild](https://github.com/HGustavs/LenaSYS/assets/129264968/8c4ae6af-9ec2-4e9f-996f-a7d6d87528c9)
When hovering over the button circled above, it should show a html/css styled popup instead of the default html title property. See [example](https://www.w3schools.com/css/css_tooltip.asp).

Requirements:
- The tooltip needs to be styled the same way as the rest of the page.
- It needs to work in dark mode. 
- It should show when the last fetch was made. 
- It should show how long is left on the cooldown (does not need to update dynamically).

b22chran:
Added tooltip-class to the github-button and a span-element with tooltiptext-class. This uses the created session-variables "lastFetchTime" and "fetchCooldown" from the gitcommitService.php to navheader.php to show the latest fetch and cooldown next possible fetch. This is visable when hover. Fixed the classes in style.css for the tooltip to work as wanted.

b22chran:
Pull-request:
https://github.com/HGustavs/LenaSYS/pull/14402

PR #14429, opened 2024-04-09
G3 2024 v1 #14352

b22chran:
Set a fixed width to the tooltip-box so it is centered underneath the gitRepo-button even if it has no date from last fetch. Set the cooldown to show "0" instead of a negative value if the cooldown in out of sync.


a22ahmom (commented):
It seems like the issue has been solved. When hovering over the download github repo button a popup or tooltip is shown. Based on the requirements it seems to include the same style as the rest of the page, it works in darkmode, it shows when the last fetch was made and it shows the cooldown.

a22oscja:
appears to largely work, cooldown/time of last click would however only update upon the page being refreshed on my machine.

PR #14437, opened 2024-04-09
Fixed issue where states could not be accessed inside superstates. (#14304).

b22chrmo:
Added fixed Z-index for elements created in diagram container. States can now be accessed when they are inside super states.


b22chrmo:
This is a new pull request. Old one could not be merged. 

https://github.com/HGustavs/LenaSYS/pull/14334

Issue #14160, opened 2023-05-22
Create microservice createNewCourse


c21alest:
Related to: https://github.com/HGustavs/LenaSYS/issues/14143

Read more about the MS [here](https://github.com/HGustavs/LenaSYS/blob/1f6fc81f85fd7cadc4652cf44f52bfab88e2e496/backend-models/microservices/duggaSys%20services/Potetial_micro_services.md#createNewCourse)

createNewCourse_ms.php
Folder: courseedService

c22mohel:
Upon further research this issue appears to be almost done. Just testing needs to be done

a22iliru:
please open a pull request for testers

c22mohel:
This is not ready for test

a22iliru:
seems to already be a course called createnewcourse without this issue being closed. so closing it.

PR #14438, opened 2024-04-09
G2 2024 v13 #14160

c22mohel:
fixed createNewCourse micro service and added new check in dugga.js when kind==="course"


a22marca (changes requested):
There are some issues when creating a course still, lets wait and test this more after the weekly merge.

PR #14444, opened 2024-04-10
Fixed #14358

### LINKED ISSUE #14358: Creating anything in course after deleting an existing element
When creating a new element all element that has been deleted are returned

Before deleting Elements:
before deleting old elements

after deleting Elements:
after deleting old elements

after creating new Element:
after creating new element

If you wait one minute the old elements disappear but if you leave the course page before this the elements gets saved again

a22karja:
To fix
- #14358 I added a for loop in returnedSection() because it redrew all elements that still were in the database as if they werent deleted.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -1949,6 +1949,11 @@ function returnedSection(data) {
 
 
   }
+  
+  //Force elements that are deletet to not show up unless pressing undo delete or reloading the page
+  for(var i = 0; i < delArr.length; i++){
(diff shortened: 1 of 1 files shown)


SleepyBearIV (changes requested):
Fix is not working on my system. If you delete something, it will reappear when creating something new. However, if the page is reloaded, both the function to redo something and the problem of deleted elements reappearing are not there.

SleepyBearIV (approved):
"Code works as intended, it won't bring back elements recently deleted. Tested with different elements and numbers of elements.

a22willi (approved):
Looks good.

Issue #14206, opened 2023-05-24
Create microservice deleteCourseMaterial


c21alest:
Related to: https://github.com/HGustavs/LenaSYS/issues/14143

Read more about the MS [here](https://github.com/HGustavs/LenaSYS/blob/1f6fc81f85fd7cadc4652cf44f52bfab88e2e496/backend-models/microservices/duggaSys%20services/Potetial_micro_services.md#deleteCourseMaterial)

**cdeleteCourseMaterial_ms.php**
Folder: **courseedService**

c21alest:
Depending on:
deliteFromTableCodexample
deliteFromTableListentries
deliteFromTableQuiz
deliteFromTableCourse

a22yifwa:
Need to add a File: **deleteCourseMaterial_ms.php** in the Folder **courseedService**. In deleteCourseMaterial_ms.php we need to use different service to delete _Course Material_ from different table. 
In MS : Uses service **deliteFromTableCodexample** to delete a row from the table codeexample. Uses service **deliteFromTableListentries** to delete a row from the table listentries. Uses service **deliteFromTableQuiz** to delete a row from the table quiz. Uses service **deliteFromTableVers** to delete a row from the table vers. Uses service **deliteFromTableCourse** to delete a row from the table course.

a22yifwa:
- Lines **543-548** from the courseedservice.php are use to delete the course example from the course database.
- Lines **550-555**  from the courseedservice.php are use to delete listentries from the course database.
- Lines **557-562** from the courseedservice.php are use to delete the quiz from the course database.
- Lines **564-569** from the courseedservice.php are use to delete the vers from the course database.
- Lines **572-577** from the courseedservice.php are use to delete courses that have been marked as deleted.

a22yifwa:
For make a microservice we need to use the file monolithic-to-microservices with [https://github.com/HGustavs/LenaSYS/blob/6874b5277f63c2ca2e0a60429d0c22dc1878c3fb/DuggaSys/microservices/monolithic-to-microservices.md](Link) . We need to follow the steps and enter the correct code from the courseedservice.php. Those lines of code I wrote earlier can be used in the new file.

a22yifwa:
[...]

PR #14453, opened 2024-04-10
g2-2024-v2-#14206

a22yifwa:
I created a new file called deleteCourseMaterial_ms.php. This file can only delete things about the course, not add it.

### DIFF
--- a/DuggaSys/microservices/courseedService/deleteCourseMaterial_ms.php
+++ b/DuggaSys/microservices/courseedService/deleteCourseMaterial_ms.php
@@ -1,0 +1,57 @@
+<?php
+
+
+date_default_timezone_set("Europe/Stockholm");
+
+ // Include basic application services
+ include_once ("../../../../coursesyspw.php");
+ include_once ("../../../Shared/sessions.php");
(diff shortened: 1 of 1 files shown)


a22marca:
I'm trying to find a good way to test that correct rows in the db is deleted. The functionality of marking courses as deleted doesn't seem to be properly setup at the moment, so it is not possible to check through the interface.

Will have to create queries to update rows and set them as `visiblity=3`

a22marca:
It seems the original code in the service file that was moved to the microservices was not properly written and tested. There are a few foreign key constraints that are not resolved regarding the tables `box` (related codeexamples) and `course_req` (many-to-many relation between courses). There might be more. Thus deletion of entries in `codeexample` and `course` are not able to be performed as of now.

a22marca (changes requested):
The current code is not able to perform all of the queries because of key constraints. 
I've added a few comments in the pull request on how I tested the code and found the related issues. I also added snippets of the code that fixes the issues that occured.

Before this can be merged, these problems needs to be fixed.

PR #14457, opened 2024-04-10
the used url for curl in test.php will use the service entry from eac…

a22marca:
The testing API will now use a url from the service entry of each test. Test.php will retrieve the url that is placed in the service entry of each test. 

Using for example a XAMPP/apache solution on port 80 would work without issues now. The service `sectionedservice.php` that is used in the tests in `sectionedservice_Test.php` will have the service entry `localhost/LenaSYS/DuggaSys/sectionedservice.php`. If the user uses another port for their local web server, they need to add it themselves like so:
`localhost:1337/LenaSYS/DuggaSys/sectionedservice.php`


a22fokol:
i have test it and everything is ok!

Issue #14357, opened 2024-04-05
Excessive Updates on `sectioned.php`


b22marah:
### Problem Summary
- **Frequent Updates:** Expected every 10 minutes, but occurring every 10 seconds.
- **Redundant AJAX Calls:** Up to 30 identical AJAX calls for course information, with the call's location unknown.

### Expected vs. Current Behavior
- **Expected:** Updates every 10 minutes, minimal AJAX calls.
- **Current:** Updates every second, multiple redundant AJAX calls.

### Location
- Frequent updates start from `sectioned.js` line 3138.
- AJAX call location is unknown.

### Action Needed
- Identify and fix the cause of the frequent updates and redundant AJAX calls.

PR #14459, opened 2024-04-10
G3 2024 v2 #14357 Removed redundant AJAX calls and fixed the time interval

a22moamy:
Removed unnecessary for-loop to minimize AJAX-calls. Now there will only be one AJAX-call for every moment in a course. The content of the AJAX-call remains unchanged. 

Now the AJAX-calls are only executed in a 10 minute interval, additionally also when the user refreshes a moment.

We have noticed that within the AJAX-call, there is a `duggor` attribute which contains information unrelated to the course. Another issue might be to change the structure of the AJAX-call to contain less information, thus avoiding unnecessary repetition. 

Issue: 
#14357

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -3213,11 +3213,9 @@ setInterval(function () {
       console.log("Time to update the code examples.");
 
       let returnedPromises = [];
-      for (let i = 0; i < itemKinds.length; i++) {
+      for (let i = 0; i < collectedLid.length; i++) {
         if (itemKinds[i] === 4) {
-          for (let i = 0; i < collectedLid.length; i++) {
-            returnedPromises.push(createExamples(collectedLid[i], false));
-          }
+          returnedPromises.push(createExamples(collectedLid[i], false));
         }
       }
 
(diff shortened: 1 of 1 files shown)


a22chrfa:
The timer works as intended. The superfluous AJAX requests are now eliminated and only the relevant requests are made.
The comments make the code very readable and the logic used is concise, correct and effective.

One thing: 
Since "updateMins" won't be changed during runtime it should be declared as a const instead of var, according to best practices:
https://google.github.io/styleguide/jsguide.html#features-use-const-and-let

DuggaSys/sectioned.js
```diff
@@ -31,7 +31,8 @@ var isLoggedIn = false;
 // Globals for the automatic refresh (github)
 var isActivelyFocused = false; // If the user is actively focusing on the course page
 var lastUpdatedCodeExampes = null; // Last time code examples was updated
-const UPDATE_INTERVAL = 600 * 100; // Timerintervall for code to be updated (10 minutes)
+var updateMins = 10; // Variable expressed in minutes for how often the code should be updated
```
a22chrfa, line 34: Should be declare as a const instead of var according to best practices:
https://google.github.io/styleguide/jsguide.html#features-use-const-and-let

a22chrfa (changes requested):
The timer works as intended. The superfluous AJAX requests are now eliminated and only the relevant requests are made.
The comments make the code very readable and the logic used is concise, correct and effective.

One thing:
Since "updateMins" won't be changed during runtime it should be declared as a const instead of var, according to best practices:
https://google.github.io/styleguide/jsguide.html#features-use-const-and-let

b22marah:
Looks good, and var has been changed to const, which solves the change request!

b22marah (approved):
Looks good!

PR #14461, opened 2024-04-10
Fixed #14398

### LINKED ISSUE #14398: Github download cooldown should be per-course and not global
### Problem Description

The cooldown for downloading & checking for changes in github repo is currently setting a global cooldown for all courses. This should not be the case, instead it should set a download cooldown for the specific course where the fetch-request took place while other non-related courses are un-affected by this request.

### Expected Behavior

The cooldown should be applied on a per-course basis rather than globally. This means that when a fetch request is made for a specific course, only that course should enter a cooldown period for downloads and change checks. Other courses, not related to the current fetch request, should remain unaffected and accessible without waiting for the cooldown to expire.

a22karja:
- #14398

added SESSION[updatetGitReposCooldown][$cid] to refreshCheck in gitcommitService.php
So that the update cooldown is connected to courses through the course id

### DIFF
--- a/DuggaSys/gitcommitService.php
+++ b/DuggaSys/gitcommitService.php
@@ -116,8 +116,12 @@
 			$updated = $row['updated'];
 		}
 
+
 		$currentTime = time(); // Get the current time as a Unix timestamp
 		$updateTime = strtotime($updated); // Format the update-time as Unix timestamp
+
+		$_SESSION["updatetGitReposCooldown"][$cid]=$updateTime;
+		
 
 		$_SESSION["lastFetchTime"] = date("Y-m-d H:i:s", $currentTime);
 		$fethCooldown = $longdeadline - (time() - $updateTime);
@@ -128,7 +132,8 @@
 		}
 		// Check if the user has superuser priviliges
 		if($user == 1) { // 1 = superuser
-			if(($currentTime - $updateTime) < $shortdeadline) { // If they to, use the short deadline
+			if(($currentTime - $_SESSION["updatetGitReposCooldown"][$cid]) < $shortdeadline) { // If they to, use the short deadline
+				
 				print "Too soon since last update, please wait.";
 				return false;
 			} else {
@@ -136,7 +141,7 @@
 				return true;
 			}
 		} else { 
-			if(($currentTime - $updateTime) > $longdeadline) { // Else use the long deadline
+			if(($currentTime - $_SESSION["updatetGitReposCooldown"][$cid]) > $longdeadline) { // Else use the long deadline
 				newUpdateTime($currentTime, $cid);
 				return true;
 			} else {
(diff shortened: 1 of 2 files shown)


a22moamy:
# The functionality works as intended
The added functionality works, but not visually. 

## A potential issue
After visiting a course page for the first time and entering a cooldown for the first time, the cooldown won’t be visibly expressed as a number. This can be seen in the following figure:
Skärmavbild 2024-04-10 kl  14 01 47

## Another potential issue
To see the cooldown expressed in numbers, the user has to press the button again and refresh the page through the web browser, then the cooldown will be shown. Is this per design or is it worth another issue?

Similarly, this can happen:
Skärmavbild 2024-04-10 kl  13 49 37

This, again, is since the page does not update the cooldown without having the user press on the button and then refresh the page. Perhaps the cooldown should be updated in intervals to avoid confusion? 

After downloading a Github repo and refreshing the page, the cooldown will say 0. To get the actual cooldown, you have to press the button again and then refresh the page again. 

## A third potential issue
Additionally, going into a course after having entered a cooldown in another course will display the same cooldown from that course. For example, if the user entered a cooldown in “Datorgrafik”, then headed into “Webbprogrammering”, then “Webbprogrammering” will display the same cooldown as in “Datorgrafik”. 

Nonetheless the expected functionality of this issue still works.

a22moamy (changes requested):
# The functionality works as intended
The added functionality works, but not visually. 

## A potential issue
After visiting a course page for the first time and entering a cooldown for the first time, the cooldown won’t be visibly expressed as a number. This can be seen in the following figure:
Skärmavbild 2024-04-10 kl  14 01 47

## Another potential issue
To see the cooldown expressed in numbers, the user has to press the button again and refresh the page through the web browser, then the cooldown will be shown. Is this per design or is it worth another issue?

Similarly, this can happen:
Skärmavbild 2024-04-10 kl  13 49 37

This, again, is since the page does not update the cooldown without having the user press on the button and then refresh the page. Perhaps the cooldown should be updated in intervals to avoid confusion? 

After downloading a Github repo and refreshing the page, the cooldown will say 0. To get the actual cooldown, you have to press the button again and then refresh the page again. 

## A third potential issue
Additionally, going into a course after having entered a cooldown in another course will display the same cooldown from that course. For example, if the user entered a cooldown in “Datorgrafik”, then headed into “Webbprogrammering”, then “Webbprogrammering” will display the same cooldown as in “Datorgrafik”. 

Nonetheless the expected functionality of this issue still works. The different courses still have their own cooldowns, just not always visually.

a22moamy (approved):
# Great, approved
Great! Everything works perfectly. Now the cooldown is updated live and is even expressed in minutes and seconds. Much better for human readability. The cooldown for every course is individual as well. 

Here's how it looks like:
Skärmavbild 2024-04-11 kl  13 44 38

No errors have been found through deleting history / cache either. Cooldown is also hidden when entering a course for the first time.

a22willi (changes requested):
Dont use the $_session for this, the cooldown should be for all users, not for individual users. But each course should have its own cooldown.

a22willi (approved):

PR #14462, opened 2024-04-10
fixed #14460

### LINKED ISSUE #14460: Fix code in createNewCourse_ms.php
Continue work on the microservice.

c22mohel:
simplified the code and added comments + added new check in dugga.js. Also new code uses the getUid() system to handle login.


c22mohel:
The url in dugga.js needs to be reverted to original after testing

Issue #14368, opened 2024-04-05
error on fields stay red even if correct when creating new Test


a22karja:
While editing a new test after creating it, if the field becomse red it will stay red even if the information is valid.
  
![red markings stay after being fixed](https://github.com/HGustavs/LenaSYS/assets/129260794/09cf99c9-b4f4-4933-a276-878a9f500b34)

a22chrfa:
Fixed validation for name and date
For the name:
Validation color for name was based on a variable that was defined in a function that turned out to be unused. Therefore, undefined. Unsure if the function has any other functions and if the variable might be used in other place.

For this validation I created a new variable "inputColorTheme". A function was added to initialize the variable by checking localStorage for which color theme is being used. in "validatesSectName()" the function is always run to reinitialize the variable, since the theme can be changed before or after entering the page. The function then uses the new variable insted of the old one to color the element.

For the date:
The date validation was based on the same variable, which is now changed, so the color will work (confirmed with testing). But that's only one of two issues. The remaining issue is that the date checks if the absolute deadline is set between the start and end date of the object (the test), but the when the object is created it is created with incorrect start and end dates (2020).

So the validation and coloring is fixed, but the creation of a new object is a new issue that I suggested being a separate issue on github.

a22chrfa:
Notes and testing:
Neither the checkbox nor the date-element is triggering the function that validates the date. So the function wont be found by inspecting those elements. The function being used is validateDate2() and it's being triggered by mouseover from the "editSection" ID. This is found in the code but not when inspecting the element in the browser for some reason. Looking through the callstack will reveal.

Issue #14343, opened 2024-04-04
Fix text validation when editing course


a22willi:
When editing existing courses, the text validation does not work as intended. 

When pasting a valid github url, it still complains:
![bild](https://github.com/HGustavs/LenaSYS/assets/129264968/0ec3f6bf-b11d-49b3-af67-0af7734fac2e)

When entering an invalid course number, the error message is displayed in the wrong place.
![bild](https://github.com/HGustavs/LenaSYS/assets/129264968/e2edc5e7-8693-4868-9710-ebf5fe5fd3ca)

Sometimes, even though all the information is valid, it still gives errors:
![bild](https://github.com/HGustavs/LenaSYS/assets/129264968/3d540ee1-3907-4b48-b203-1470f5a88ac7)

PR #14463, opened 2024-04-10
Fixed validation for name and date #14368

a22chrfa:
For the name:
Validation color for name was based on a variable that was defined in a function that turned out to be unused. Therefore, undefined. Unsure if the function has any other functions and if the variable might be used in other place.

For this validation I created a new variable "inputColorTheme". A function was added to initialize the variable by checking localStorage for which color theme is being used. in "validatesSectName()" the function is always run to reinitialize the variable, since the theme can be changed before or after entering the page. The function then uses the new variable insted of the old one to color the element.

For the date:
The date validation was based on the same variable, which is now changed, so the color will work (confirmed with testing). But that's only one of two issues. The remaining issue is that the date checks if the absolute deadline is set between the start and end date of the object (the test), but the when the object is created it is created with incorrect start and end dates (2020).

So the validation and coloring is fixed, but the creation of a new object is a new issue that I suggested being a separate issue on github.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -27,6 +27,16 @@ var updatedLidsection;
 var numberOfItems;
 var backgroundColorTheme;
 var isLoggedIn = false;
+var inputColorTheme;
+
+function initInputColorTheme() {
+  if(localStorage.getItem('themeBlack').includes('blackTheme')){
+    inputColorTheme = "#212121";
+  } 
+  else {
+    inputColorTheme = "#FFF";
+  }
+}
 
 // Globals for the automatic refresh (github)
 var isActivelyFocused = false; // If the user is actively focusing on the course page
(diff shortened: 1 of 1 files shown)


a22edvbu:
The validation works as intended but the validation window for the name is not on the correct line. 
image

Validation for setting a new date works as well but only for dates between 2014-12-29 and 2015-03-08 which seems to be hardcoded.

Overall the validation correctly validates the name and date for a test. The code checks out aswell.

a22willi:
> The validation works as intended but the validation window for the name is not on the correct line. image
> 
> Validation for setting a new date works as well but only for dates between 2014-12-29 and 2015-03-08 which seems to be hardcoded. image
> 
> Overall the validation correctly validates the name and date for a test. The code checks out aswell.

This issue has been fixed by issue #14343

PR #14468, opened 2024-04-10
Solved the 'keybind for note' bug

d22hanol:
Note was spelled NOTE, IDE did not give an error for this. this is now fixed.


a22cazlu:
Tested all the keybindings for the toolbar and the I get up the note tool when I press 8 så it seems to work fine now.

PR #14469, opened 2024-04-10
Fixed the collide issue

a22cazlu:
You are no longer stuck when you collide with another objekt using the arrow keys.  Implemented in some code in setPos()- if(overlapping) that make the objekt take one step back if it collide.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1569,21 +1569,65 @@ document.addEventListener('keydown', function (e) {
     }
 
     // Moving object with arrows
-    if (isKeybindValid(e, keybinds.MOVING_OBJECT_UP) && !settings.grid.snapToGrid){
-        console.log("Up");
-        setPos(context, 0, 1);
+    if (isKeybindValid(e, keybinds.MOVING_OBJECT_UP) && !settings.grid.snapToGrid) {
+        let overlapDetected = false;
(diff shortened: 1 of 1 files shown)


b22jonan:
testing failed. moved object into another one and tried moving down and it was still stuck

https://github.com/HGustavs/LenaSYS/assets/61019188/d22feeef-f229-4995-8bf6-1f821cec8482

a22jacka:
Seems like it works but only if you go in the opposite direction of the other element, could maybe be worth changing so that every direction beside the side where the other element is should work

b22jonan:
> Seems like it works but only if you go in the opposite direction of the other element, could maybe be worth changing so that every direction beside the side where the other element is should work

for clarity, i went up into the object, then pressed down arrow to try and move down and away

b22jonan:
never mind, chrome just didnt cleare cache when i told it to do so, changes did in fact work

b22jonan (approved):
changes work, can now move away with arrow keys

a22solmi (changes requested):
Looks good but please remove the console logs and comments.

b22jonan (approved):
tested moving into another entity from all 4 directions, could then move in all other directions except into the entity. also tested moving into a corner between two other entities, did not get stuck there either.

PR #14471, opened 2024-04-10
Reimplemented doSetup() function for issue#14426

a23ponek:
Removed // before onSetup() to activate function at line 1361 in diagram.js


a22jacka:
Wrong change, you need to uncomment the `setup()` function in diagram_dugga.js. Found under `DuggaSys\templates\diagram_dugga.js`

a23ponek:
Correct change is done.

Issue #14426, opened 2024-04-09
Reimplement setup()


a22jacka:
## Description
#14312 found that reimplementing the `setup()` function doesn't bring back the bug

## Job
Reimplement it

PR #14475, opened 2024-04-10
Reimplemented setup() function from issue#14426

a23ponek:
uncommented the setup() function at line 22-33 in the diagram_dugga.js in the DuggaSys\templates\diagram_dugga.js filepath.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1358,7 +1358,7 @@ function getData() {
     document.getElementById("container").addEventListener("mouseup", mup);
     document.getElementById("container").addEventListener("mousemove", mmoving);
     document.getElementById("container").addEventListener("wheel", mwheel);
-    onSetup();
+    //onSetup();
     // debugDrawSDEntity(); // <-- debugfunc to show an sd entity
     generateToolTips();
     toggleGrid();
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1358,7 +1358,6 @@ function getData() {
     document.getElementById("container").addEventListener("mouseup", mup);
     document.getElementById("container").addEventListener("mousemove", mmoving);
     document.getElementById("container").addEventListener("wheel", mwheel);
-    //onSetup();
     // debugDrawSDEntity(); // <-- debugfunc to show an sd entity
     generateToolTips();
     toggleGrid();
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1358,7 +1358,7 @@ function getData() {
     document.getElementById("container").addEventListener("mouseup", mup);
     document.getElementById("container").addEventListener("mousemove", mmoving);
     document.getElementById("container").addEventListener("wheel", mwheel);
-    // onSetup();
+    onSetup();
     // debugDrawSDEntity(); // <-- debugfunc to show an sd entity
     generateToolTips();
     toggleGrid();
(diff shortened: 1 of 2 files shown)


a22jacka:
Link to issue: #14426

DuggaSys/diagram.js
```diff
@@ -1358,7 +1358,7 @@ function getData() {
     document.getElementById("container").addEventListener("mouseup", mup);
     document.getElementById("container").addEventListener("mousemove", mmoving);
     document.getElementById("container").addEventListener("wheel", mwheel);
-    // onSetup();
+    //onSetup();
```
a22hanfa, line 1361: This should be `setup();`.

a22hanfa (changes requested):
Im getting this error when implementing the new diagram_dugga.js, which causes the diagram to not be able to load:
```
Uncaught ReferenceError: setup is not defined
    at getData (diagram.js:1361:5)
    at diagram.js:1322:5
```
And as mentioned it should be either onSetup() as it is in diagram_dugga.js or change it to setup(). You need to import the function because it's currently not recognised by the file, as the error denotes.

Good luck!

a23ponek:
If the setup() function from the diagram_dugga.js is moved into diagram.js and called with setup(), more errors occur like "returnedDugga() not defined" and if that function is also later imported, 2-3 more errors occur. @a22jacka should this be split into further issues?

a22solmi:
> If the setup() function from the diagram_dugga.js is moved into diagram.js and called with setup(), more errors occur like "returnedDugga() not defined" and if that function is also later imported, 2-3 more errors occur. @a22jacka should this be split into further issues?

Why would you move it? The function should remain in diagram_dugga.js 
Do you get errors when running the function in the correct place

a23ponek:
I have now removed the setup() call from the diagram.js file. And the setup() function in diagram_dugga.js is now active. Things seem to work without any problems in diagram.php. Awaiting new testing.

a22jacka:
Did a small test in a dugga since I had one open and the save button there appears to work aswell. Further testing is required

a22solmi:
> LGTM. Everything seems to work fine on my end.

Please read thrugh the code to see what you need to test. No meaningful changes have been made, this code is nearly identical to the weekly branch. Reviews should contain info on what was tested and the outcome.

a22solmi:
This code does not fix a issue or implement new functionallity. You have replaced a comment with another comment. The function that was uncommented is never called. I cannot approve the code as is.
Please correct me if I am wrong!

a22solmi:
> LGTM. Everything seems to work fine on my end.

What does this mean? please write what you have tested and provide screenshots.

a22hanfa:
Code added to `diagram.js`:
`console.log("Testing function of onSetup()")
    onSetup();
    console.log("Testing function of onSetup(), PASSED")
    console.log("Testing function of onSetup()")
    setup()
    console.log("Testing function of setup(), PASSED")`

a22hanfa:
Added testing code to `diagram.js`:
```
console.log("Testing function of setup()")
    setup()
    console.log("Testing function of setup(), PASSED")
```
   
    And in the function setup in `diagram_dugga.js`:
    `console.log("setup() is being used")`

Error cause by not `setup` being defined:
Screenshot 2024-04-16 at 15 14 40

These two errors will cause a permanent loading screen.

The page now loads after removing those related lines.

We can see that the setup function is being called:
Screenshot 2024-04-16 at 15 43 16

Issue #14292, opened 2024-04-02
Broken arrrows for certain elements


a22jacka:
## Description
When making "Arrow" icons between IE or UML elements sometimes the arrow is broken. 
For both the arrow is only half on the left hand side when going across.
For Class entities the top arrow is completely gone when vertical.

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/16867d54-9307-4f06-be34-a67da92979ac)
![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/397d46ed-ec9e-4149-9805-f8dc95049fde)

## Job
Fix this

PR #14477, opened 2024-04-10
arrow is not broken anymore  #14292

a22vioja:
I changed the parameter (elemsAreClose) in swich case for startIcon with same parameters as there was for other icons.


a22jacka:
Link to issue: #14292

a22wilas (approved):
### Old version
bild
bild

### This branch
bild
bild

After doing some testing on both this and the old version, I can confirm that the changes in this branch fixed the arrow bug. No new bugs with the change have been discovered. This has also been tested on the sequence activation element, where the bug now have been fixed.

PR #14478, opened 2024-04-10
Updated validation for new and old courses #14343

a22edvbu:
Updated the code so that the error messages shows on the same row as the text input. When logged in as Admin, press the "+" FAB to the bottom right and choose course. Alternatively, you could press the cog next to a course's name to edit.

(This has been updated from v1 to v2)

issue #14343

### DIFF
--- a/DuggaSys/courseed.js
+++ b/DuggaSys/courseed.js
@@ -707,15 +707,13 @@ const regex = {
 
 //Validates single element against regular expression returning true if valid and false if invalid
 function elementIsValid(element) {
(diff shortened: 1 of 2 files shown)


a22moamy (approved):
# Works great
Works perfectly. Error messages are displayed on the correct row with the correct message. Works for both editing courses and creating new ones.

For example (creating a new course): 
Skärmavbild 2024-04-10 kl  15 06 34

Another example (editing a course):
Skärmavbild 2024-04-10 kl  15 08 07

Issue #14454, opened 2024-04-10
Display cooldown in correct unit


a22willi:
![image](https://github.com/HGustavs/LenaSYS/assets/129264968/ea978813-d7c5-46cd-8043-d4c2383b18e3)
The cooldown for being able to fetch code examples from github is currently shown in seconds (I think?)
Change this to display the cooldown in minutes + seconds. Ex. `Cooldown: 1min 14s` Or `Cooldown: 42s`

This tooltip was implemented in #14352.

PR #14479, opened 2024-04-10
fixd #14454

a22karja:
By using date() i was able to get $_SESSION['fetchCooldown'] in minutes and seconds instead of just seconds

- #14454


a22willi (approved):
Looks good! And works as intended.

PR #14481, opened 2024-04-10
G2 2024 v2 issue#14435

a22marca:
I've create the microservice getDeletedListentries_ms.php. It can be called from the function `getDeletedListEntries()` in sectioned.js. Remember to uncomment the code that integrates it!

Open the file getDeletedListentries_ms.php in you'r browser to see the data retrieved by the microservice.
To test the javascript function it can be executed in the browser console when visiting a page using sectioned.php (any course page).

### DIFF
--- a/DuggaSys/microservices/sectionedService/getDeletedListentries_ms.php
+++ b/DuggaSys/microservices/sectionedService/getDeletedListentries_ms.php
@@ -1,0 +1,31 @@
+<?php
+/*
+* Retrieives ALL listentries with DELETED visibility (visible=3)
+* returns an array containing the rows
+*/
+
+date_default_timezone_set("Europe/Stockholm");
+
+// Include basic application services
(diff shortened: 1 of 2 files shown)


b21isabe (changes requested):
I have tested by setting visible=3 on a few listentries in the database and then opening getDeletedListentries_ms.php in the browser. This retrieved all of the listentries with visible set to 3.

I have also tested running the function in the browser console, this didn't work at first because of a typo in the url of the function:
`url: "../DuggaSys/microservices/sectionesdService/getDeletedListentries_ms.php",`

There seems to be an extra "s" in sectionedService, when this was corrected the function worked as it should.

I don't know if the commented out code in the function will be used in the future but to be safe I think the typo should be fixed before merging.

a22marca:
Nice catch, I have corrected the URL and pushed the commit.

PR #14482, opened 2024-04-10
replaced break; with isOverlapping = false;

a22ricni:
I replaced the break; line with isOverlapping = false. If you draw a super state or sequence actor then you should not be able to overlap with other diagrams like ER diagrams.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -3595,11 +3595,11 @@ function entityIsOverlapping(id, x, y) {
                 }
                 //if its overlapping with a super state, just break since that is allowed.
                 if (data[i].kind == "UMLSuperState" || element.kind == "UMLSuperState") {
-                    break;
+                    isOverlapping = false;
                 }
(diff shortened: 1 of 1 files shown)


a22vioja:
I tested the fix by trying to draw a sequence actor and overlap it with super state but it did not work and that is what it should be. 

Screenshot 2024-04-12 at 09 25 01

I also tried to overlap superstates but it did not work either, and that is what it should be. 

Screenshot 2024-04-12 at 09 24 41


I do not see any new problems on the diagram behaviour which can be affected by these changes.

a22vioja (approved):
I tested the fix by trying to draw a sequence actor and overlap it with super state but it did not work and that is what it should be. 

Screenshot 2024-04-12 at 09 25 01

I also tried to overlap superstates but it did not work either, and that is what it should be. 

Screenshot 2024-04-12 at 09 24 41

I do not see any new problems on the diagram behaviour which can be affected by these changes.

Issue #14446, opened 2024-04-10
Reset view doesnt center diagram


a22solmi:
## Description
When pressing the "reset view" button the view is not centered. Only after pressing it twice will the entire diagram be visible. Even so, it is not always dead center.

## Job 
Change the "reset view" functionality to put the diagram in the center of the screen. Executing the center twice is not a good solution and was removed with #14293

PR #14490, opened 2024-04-11
Updated centering #14446

a22solmi:
Centering now works. Notification is only displayed once but the rest of the function is repeated like in previous issue.


a21liltr (approved):
Tested on Chrome, Edge and Firefox.

Hard reset then started the test.

The test included adding objects to the diagram then centering after each addition of object,
adding multiple objects in a row and again recentering,
deleting objects and centering,
and adding, deleting then again adding and recentering.

Seems to be working as intended.

Issue #14408, opened 2024-04-09
Make the test for "create an assignment" in duggaedservice-tests.php work


a22iliru:
The valueoutput seem to repeat the last field making the test fail.

Example:
value expected: {"debug":"NONE!","writeaccess":true,"coursecode":"IT118G","coursename":"Webbutveckling - datorgrafik","entries":[{"qname":"Bitdugga1"},{"qname":"Bitdugga2"},{"qname":"colordugga1"},{"qname":"colordugga2"},{"qname":"linjedugga1"},{"qname":"linjedugga2"},{"qname":"dugga1"},{"qname":"dugga2"},{"qname":"Quiz"},{"qname":"Rapport"},{"qname":"HTML CSS Testdugga"},{"qname":"Clipping masking testdugga"},{"qname":"test"}]}
value output: {"debug":"NONE!","writeaccess":true,"coursecode":"IT118G","coursename":"Webbutveckling - datorgrafik","entries":[{"qname":"Bitdugga1"},{"qname":"Bitdugga2"},{"qname":"colordugga1"},{"qname":"colordugga2"},{"qname":"linjedugga1"},{"qname":"linjedugga2"},{"qname":"dugga1"},{"qname":"dugga2"},{"qname":"Quiz"},{"qname":"Rapport"},{"qname":"HTML CSS Testdugga"},{"qname":"Clipping masking testdugga"},{"qname":"TestDugga1337"},{"qname":"TestDugga1337"},{"qname":"test"},{"qname":"test"},{"qname":"test"},{"qname":"test"},{"qname":"test"},{"qname":"test"},{"qname":"test"},{"qname":"test"},{"qname":"test"},{"qname":"test"}]}

file: duggaedservice-tests.php

c22cargo:
This issue might require issue #14396 to be merged first, since it relies on the test file, which seems broken.

c22cargo:
Every time the test is re-ran there appears more instances of {"qname":"test"}. They seem to duplicate. I suspect this issue has something to do with information being saved and then not deleted. Since the amount of {"qname":"test"} do not get reset when the server is restarted, the information is likely stored in the database.

a22iliru:
This issue was fixed by linked pull request

PR #14496, opened 2024-04-11
g2-2024-v2 issue #14408

c22cargo:
The "qid" was not being fetched correctly, which resulted in assignments not being deleted. Since the old assignments remained the expected output did not match the actual output. This should now be fixed, and the tests should pass.


a22melst (changes requested):
The correct syntax for retrieving an id must be implemented for the variant tests aswell.
It is also concerning that all the variant tests passes even when the mentioned syntax is incorrect, meaning the operations are not performed but the tests passes anyway.

a22melst (approved):
It works but the variant tests should probably be changed so they cannot pass when the operation tested isn't performed.

PR #14498, opened 2024-04-11
Update navheader.php

b22chran:
It was trying to create a button in the hamburger menu with parameters from a version of a course that is not yet created. There was a if-statement to check if versname from the response from table vers is existing for the button in the navbar but not the hamburger menu. Placed the button in the hamburger menu in a new if-statement that checks if versname is existing aswell. Also removed dead code that was used for test an error from a previous issue.


a22moamy (changes requested):
# Implementation works!
The implementation works as described. No new errors are created and no alert is displayed.

## New issues found:
A side effect of the fix has made it so that some navbar buttons will jump down below the navbar at certain screen sizes. 

This can be tested by adjusting the screen size. The apperance can be seen below:
Skärmavbild 2024-04-11 kl  14 19 34

The log out button and the username will also not be contained within the navbar at some screen sizes. 

### Issues in bullet point list

- Fix the implementation so that navbar buttons will never jump down, no matter the screen size. The menu buttons should either be in the navbar or in the burger menu. 

- Also fix it so that the course version dropdown does not jump down as well. 

The current appearance when the screen gets to a certain size:
Skärmavbild 2024-04-11 kl  14 25 03

- The logout button and the username should not be outside the navbar as well.

The requested appearance (can be seen in Datorgrafik):
Skärmavbild 2024-04-11 kl  14 24 52
Note that the username and log out button is contained within the navbar in this screenshot from Datorgrafik.

Otherwise the implementation works, so when this is fixed there should be no problems.

a22willi:
PR is set to merge into the wrong branch (master).

Issue #14470, opened 2024-04-10
Arrow keys stop working


a22jacka:
## Description
When using the arrow keys to move elements, it only let's press/hold once and the de-selects the element

## Job 
Fix so that the element stays selected after multiple arrow key presses

PR #14499, opened 2024-04-11
G1 2024 v2 #14470

a22cazlu:
There was a call to clearcontext() at 1621 that was unnecessary and unwanted. Removing it fixed the issue.


a22jacka:
You might need to make another commit to this branch so that we can change the base-branch after. This pr got a bit broken becuase someone forced a massive revert on the master branch and it doesn't to change the base now since there's no new commits on it

d22abdha (commented):
I have completed a test in which I confirmed that the arrow key functions as intended. During the test, I demonstrated that when an element is selected, it can be moved around using the arrow keys on the keyboard.

a22solmi:
This branch has been created from master instead of g1-2024-v2.
I cannot merge this code as is. The branch needs to be rebased, but the automatic rebase through github does not work here.
Since we had an issue with the master branch I think the best option would be to create a new branch in the right way and redo the changes. Create a new pull request once that is done.

Issue #14450, opened 2024-04-10
Course schedule missing darkmode


a22willi:
This should change to a dark theme when dark mode is selected. 
![image](https://github.com/HGustavs/LenaSYS/assets/129264968/ef76639a-d422-4d4c-94ed-79512f5e78d6)

PR #14511, opened 2024-04-11
G3 2024 v2 #14450 Dark mode for course schedule

a22moamy:
My issue was that there was only a light theme for the course schedule (SVG). 
I made it so that the schedule SVG works in dark mode and light mode. The dark mode version now looks like this:
Skärmavbild 2024-04-11 kl  13 18 20

The theme should also update without having to refresh the page. To update the theme, you should only have to press the moon button:
Skärmavbild 2024-04-11 kl  13 19 18

Additionally, test that it also works when using the moon button from the burger menu (can be accessed with smaller devices). This button within the burger menu can be seen below: 
Skärmavbild 2024-04-11 kl  13 20 22

The SVG whould update without having to refresh the page, even when using the burger menu version of the button.

Pull request from issue #14450

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -70,9 +70,6 @@ function burgerToggleDarkmode(operation = 'click') {
     localStorage.setItem('themeBlack', themeStylesheet.href)
     backgroundColorTheme = "#fff";
   }
-  // Redraw course schedule / planner SVG to get the appropriate color scheme
-  // Course schedule = the SVG that displays the weeks and how far into a course we are
-  drawSwimlanes();
 
   //const themeToggle = document.getElementById('theme-toggle');
   //themeToggle.addEventListener('click', () => {});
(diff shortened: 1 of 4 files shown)


a22natth (approved):
The modifications of the code seems to be solving issue #14450. The course schedule now has a dark mode, which is accessible by clicking the moon button (both in the regular and the hamburger menu). When switching modes, it is done instantly and the page does not need to be refreshed (as stated in the pull request) in order to change the visuals of the course schedule. Additionally, the code follows code standards, and informative comments have been added, therefore making the comprehension of the changes easier.

DuggaSys/sectioned.js
```diff
@@ -2230,16 +2232,32 @@ function drawSwimlanes() {
...
+
+    //Changes text colors of week numbers in course SVG (course schedule with deadlines)
+    if(storedTheme.includes("blackTheme.css")){ //If theme is black then week numbers in the course schedule should be white
+      str += `<text x='${((i * weekwidth) + (weekwidth * 0.5))}' y='${(33)}'
+      font-family='Arial' font-size='12px' fill='white' text-anchor='middle'>${(i + 1)}</text>`;
+    }else{ //If theme is white then week numbers in the course schedule should be black
+      str += `<text x='${((i * weekwidth) + (weekwidth * 0.5))}' y='${(33)}'
+      font-family='Arial' font-size='12px' fill='black' text-anchor='middle'>${(i + 1)}</text>`;
+    }
+
```
a22willi, line 2260: I think this solution may be over complicating things. It should be possible to just add one class to the parent of all swimlane/schedule elements. And then use css selectors to set the style. For example, striped rows could be added in CSS via [css nthchild odd](https://www.w3schools.com/howto/howto_css_table_zebra.asp).

DuggaSys/sectioned.js
```diff
@@ -2279,8 +2297,19 @@ function drawSwimlanes() {
...
+          str += `<rect opacity='0.7' x='${(startday * daywidth)}' y='${(weeky)}' width='${(duggalength * daywidth)}' height='${weekheight}' fill='${fillcol}' />`;
+        }
+
+        //Changes text colors of assignments/duggor in course SVG (course schedule with deadlines)
+        if(storedTheme.includes("blackTheme.css")){ //If theme is black then duggor-text in the course schedule should be white
+          str += `<text x='${(12)}' y='${(weeky + 18)}' font-family='Arial' font-size='12px' fill='#ffffff' text-anchor='left'> <title> ${entry.text} </title>${entry.text}</text>`;
+        }else{ //If theme is white then duggor-text in the course schedule should be black
+          str += `<text x='${(12)}' y='${(weeky + 18)}' font-family='Arial' font-size='12px' fill='${textcol}' text-anchor='left'> <title> ${entry.text} </title>${entry.text}</text>`;
+        }
+
```
a22willi, line 2312: This could probably also be handled automatically if they have the correct parent css class.

a22willi (changes requested):
Generally I think a simpler solution could be implemented. If its not possible, please explain why in the review section, and I will approve it anyway.

a22moamy:
The pull request should now be updated. I have moved almost all of the JavaScript logic to CSS to fulfill the requirements. I did add some classes to elements to make it work, I hope this is okay. I found it particularly useful when styling the even and odd columns in the schedule. The way the schedule is structured makes it (from my knowledge) not possible to style the columns with nth:child. This is because there is no nestling of elements in the schedule and columns and rows are on the same level, weirdly enough.

a22willi (approved):
Looks good!

Issue #14419, opened 2024-04-09
Research and create microservice reorderListentries


a22iliru:
file: sectionedservice.php

Se: https://github.com/search?q=repo%3AHGustavs%2FLenaSYS%20reorderListentries&type=code

a22eddpe:
DuggaSys>sectionedservice.php

a22eddpe:
### reorderListentries
Uses the services __updateTableListentries__ to change the content of these columns:

a22eddpe:
need to create a new ms file in DuggaSys>microservices>sectionedService>

a22eddpe:
connected file createListentrie_ms.php

a22eddpe:
a suggestion on how to implement the service "updateTableListentries" 
$query = $pdo->prepare("CALL updateTableListentries(:courseid, :coursevers, :pos, :moment)");
$query->bindParam(':courseid', $courseid);
$query->bindParam(':coursevers', $coursevers);
$query->bindParam(':pos', $pos);
$query->bindParam(':moment', $moment);

a22eddpe:
include 					foreach ($orderarr as $key => $value) {
						$armin=explode("XX",$value);
						$query = $pdo->prepare("UPDATE listentries set pos=:pos,moment=:moment WHERE lid=:lid;");
						$query->bindParam(':lid', $armin[1]);
						$query->bindParam(':pos', $armin[0]);
						$query->bindParam(':moment', $armin[2]);
                     

						if(!$query->execute()) {
							$error=$query->errorInfo();
							$debug="Error updating entries".$error[2];
						}
					}

a22eddpe:
This code from sectionedservice.php compairs $opt to REORDER. If it equals 0, I need to use if for my reorder-file. 		    	
} else if(strcmp($opt,"REORDER")===0) {
					$orderarr=explode(",",$order);

a22eddpe:
updateListentrie
Uses service __selectFromTableListentries__ to _get_ information it requires from __listentries__.
Uses service __createNewCodeexample__ to makes _inserts_ into the table __codeexample__.
Uses the services __updateTableListentries__ to change the content of these columns:
- highscoremode
- gradesystem
- moment
- entryname
- kind
- link
- visible
- comments
- groupKind
- feedbackenabled
- feedbackquestion

Uses service __insertIntoTableList__ to makes _inserts_ into the table __list__.

PR #14515, opened 2024-04-11
g2-2024-v2 #14419

a22eddpe:


a22marca:
I'm having issues finding where to integrate this microservice, it with the rest of the system. @a22eddpe Do you know where this can be added in the javascript-part of the system, so that the microservice is called instead of sectionedservice.php?

a22eddpe:
@a22marca no but i can look for it

a22marca:
I was able to find where to integrate the microservice. I had to modify the current code a bit for this to work.
@a22eddpe 
The code at line 1093 in `Shared/dugga.js` has been modified to the one below to test this:
```
	}else if(kind=="SECTION"){		
		if(opt=="REORDER"){
			$.ajax({
				url: "../DuggaSys/microservices/sectionedService/reorderListentries_ms.php",
				type: "POST",
				data: "courseid="+querystring['courseid']+"&coursename="+querystring['courseid']+"&coursevers="+querystring['coursevers']+"&comment="+querystring['comments']+"&opt="+opt+para+"&hash="+hash,
				dataType: "json",
				success: returnedSection
			});
		}else{
			$.ajax({
			url: "sectionedservice.php",
			type: "POST",
			data: "courseid="+querystring['courseid']+"&coursename="+querystring['courseid']+"&coursevers="+querystring['coursevers']+"&comment="+querystring['comments']+"&opt="+opt+para+"&hash="+hash,
			dataType: "json",
			success: returnedSection
			});
		}
```

I don't think we should modify the code in the codebase yet though, since the microservice is not supposed to be integrated at this point, but it works for testing purposes.

a22marca:
One issues with the code in the previous comment is that you now have to reload the page to get the "new" ordering.

a22eddpe:
Yeah me neither. But nice work.

a22marca (changes requested):
The opt is not retrieved in the microservices, but it is used in the if-statement.
This should be added like this:
```
$opt = getOP('opt');
```

The integration seems a bit tricky, and I managed to integrate it to be able to test the microservices. And it seems that is able to reorder the list when adding the opt as well.
The integration is probably it's own issue, so i don't believe we need to fix it in this issue.

Add the opt, and we are good to go!

a22eddpe:
i added the parameter now

Issue #14485, opened 2024-04-11
Colors are hard coded


a22solmi:
## Description
All colors in diagram.js are written as strings like this: "#FFFFFF"

## Job
Create a color object/map where each string has a variable name.
Replace strings with variable names in entire file

a22solmi:
An array named colors exists with every string. Give them variable names

PR #14517, opened 2024-04-11
Created a colorMap for issue#14485

a23ponek:
Created a colormap at line 1025-1027 that maps colors to their respective color codes. Color codes can be fetched by colorMap.get('White') which returns the color code #ffffff as example for white color as a string type. This was done in the file diagram.js.
If one is to use a color that doesn't exists, add the color with its responding color code to colorMap([]) and use the .get('Color') call.

Additionally, all color codes are changed to the usage of colorMap instead of a hardcoded string ("#ffffff") in the whole file of diagram.js for all colors.

The old array called 'colors' containing color codes are removed.

Link to issue: https://github.com/HGustavs/LenaSYS/issues/14485


a22solmi:
You have replaced the entire file according to the commit history. I cannot approve this. You need to solve the issue by editing instead. Revert your branch and try again or create a new branch and pull request.

Issue #14403, opened 2024-04-08
Enable closing of load-file popup using escape


a22jacka:
## Description
Make it so that the popup for loading a file can be closed using the escape key

note: if you can't see the "load" button in the toolbar then open the console in the browser dev-tools and type `showModal()`

## Job
Implement this

PR #14519, opened 2024-04-11
Added functionality to close loaddiagram using escape key #14403

a22alfna:
I added the closeModal() function to the if statement regarding escape presses in the keyup EventListener.


b22davbe:
The sulotion works! You can now hide the pop-up using the ESC button.

b22davbe (approved):
The ESC-button now works for closing the loaddiagram pop-up.

PR #14522, opened 2024-04-12
Update navheader.php

b22chran:
It was trying to create a button in the hamburger menu with parameters from a version of a course that is not yet created. There was a if-statement to check if versname from the response from table vers is existing for the button in the navbar but not the hamburger menu. Placed the button in the hamburger menu in a new if-statement that checks if versname is existing aswell. Also removed dead code that was used for test an error from a previous issue.


d22mikwe (approved):
The functionality for creating a new course is working after the pushed commits. There are no longer any alert errors appearing when selecting the created course.

### **Noteworthy for future issues**: Wen selecting the created course the user is now prompted to create a new course version. It is not possible to select any content to copy to the course.
image

After creating a new course version the course is empty apart from course name and course id. 
image

The data is also not stored. If the user exits the course after creating a new course version all previous data is gone and the user is once again prompted to create a new course version.

PR #14524, opened 2024-04-12
Created a colorMap for issue#14485 new PR

a23ponek:
Created a colormap at line 1025-1027 that maps colors to their respective color codes. Color codes can be fetched by colorMap.get('White') which returns the color code #ffffff as example for white color as a string type. This was done in the file diagram.js.
If one is to use a color that doesn't exists, add the color with its responding color code to colorMap([]) and use the .get('Color') call.

Additionally, all color codes are changed to the usage of colorMap instead of a hardcoded string ("#ffffff") in the whole file of diagram.js for all colors.

The old array called 'colors' containing color codes are removed.

Link to issue: https://github.com/HGustavs/LenaSYS/issues/14485

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -7188,7 +7188,7 @@ function toggleColorMenu(buttonID) {
             // Create svg circles for each element in the "colors" array
             for (var i = 0; i < colorKeys.length; i++) {
                 menu.innerHTML += `<svg class="colorCircle" xmlns="http://www.w3.org/2000/svg" width="50" height="50">
-            <circle id="BGColorCircle${i}" class="colorCircle" cx="25" cy="25" r="20" fill="${colorMap[colorKeys[i]]}" onclick="setElementColors('BGColorCircle${i}')" stroke=colorMap.Black stroke-width="2"/>
+            <circle id="BGColorCircle${i}" class="colorCircle" cx="25" cy="25" r="20" fill="${colorMap[colorKeys[i]]}" onclick="setElementColors('BGColorCircle${i}')" stroke="${colorMap.Black}" stroke-width="2"/>
             </svg>`;
                 width += 50;
             }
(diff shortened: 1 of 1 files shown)


a22jacka:
You need to change this a bit since you've replaced the color inside strings, like on line 7988 it's going to become `fill=colorMap.get('White')`. In that and similair places it needs to be re-written as `fill="${colorMap.get('White')}"`

It would also be made simpler by writing the map like this:
```
const colorMap = {
    White: "#ffffff",
    Black: "#000000",
    [etc...]         
}
```
Then you can also access the values by just using `colorMap.White`

a23ponek:
Changed to the new recommendations. See last commit.

a22jacka (changes requested):
You need to change this a bit since you've replaced the color inside strings, like on line 7988 it's going to become `fill=colorMap.White`. In that and similair places it needs to be re-written as `fill="${colorMapWhite}"`

a22jacka (approved):
Looks good syntax wise, someone else will test if all the colors still work

a22ricni (changes requested):
With this changes you can't change the color of the diagrams/ elements. Get a error in function "toggleColorMenu".

It looks like this with these changes:
image

But it should look like this:
image

a23ponek:
Fixed problem with toggleColorMenu, colors now appear as intended.

a22ricni (changes requested):
Not all colors doesn't change correct when I try to change color in a diagram/ element. For example Orange-red change to color black when selected. It is other colors that change to color black.
Another thing is that in the "toggleColorMenu" menu have as stroke for the circle "stroke=colorMap.Black" and not "stroke=#000000".

a22solmi:
> Changed a variabel that seemed to cause the problem. Try again now and it should work to change between colors.

There are still variables in strings that does not use $(). You need to change this before the code is tested again.

a23ponek:
The last variables are now changed to "${colorMap.Black}" .

a22solmi:
> The last variables are now changed to "${colorMap.Black}" .

I can still see strings where you have not done this. Line 2325 for example. Please go through all the code

a23ponek:
Looked over the code, all should have been changed now. Leave a comment if there is more that I missed.

a22solmi:
Please remove the object.keys and use "for (let key in object) " instead. If you need the keys indexed, just use the object.keys directly.

a22jacka:
> Please remove the object.keys and use "for (let key in object) " instead. If you need the keys indexed, just use the object.keys directly.

Can otherwise use `for (let value of Object.values(colorMap))` or `for (let [key, value] of Object.entries(colorMap))` since you can access the values directly that way

a22ricni (changes requested):
You have missed a color in for the colormap. The hex code color is "#614875" and are in many places but is not in the colormap. A example is on the line 5864.

a23ponek:
> You have missed a color in for the colormap. The hex code color is "#614875" and are in many places but is not in the colormap. A example is on the line 5864.

Added the color to colorMap, se latest commit.

a22jacka:
The color #614875 was added to the map but it's never used in the code. 
I'd also recommend to remove the spaces in the keys since the value can't be accessed with the dot notation (`colorMap.color`) if there's spaces in the key. 
If your IDE allows for searching using regex, you can search for `"#[a-zA-Z0-9]{6}"` to see all use cases of hex-colors

a23ponek:
The color #614875 is called in the code so it needs to be in the colorMap. Changed the remaining colors that I found with regex.

a22ricni (changes requested):
I can't change the color again on the diagrams (like the first request changes)

image

a23ponek:
Colors should now be shown again.

PR #14528, opened 2024-04-12
fixed element delete function

a22vioja:
The problem was that delete button area was covering the element area because of the wrong “deleteBtnX” callcullation. 

I changed the “deleteBtnX” calculation. I added 5 pixels to the X-coordinate of the selected element's right edge (highX) to position the delete button slightly to the right of the element.

With this change the delete button is outside the element area and it is not possible to delete it without pressing on the element, when delete button is visible. 

Screenshot 2024-04-12 at 10 05 36

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -12025,8 +12025,9 @@ function drawSelectionBox(str) {
         } else if (deleteBtnSize < 15) {
             deleteBtnSize = 15;
         }
-
-        deleteBtnX = lowX - 5 + highX - lowX + 10 - (deleteBtnSize / 2);
+        
+        // Calculate delete button position relative to the selected element
+        deleteBtnX = highX + 5;
         deleteBtnY = lowY - 5 - (deleteBtnSize / 2);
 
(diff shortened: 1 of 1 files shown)


a22solmi:
Moving the cross does not solve the root issue. Please find a way to disable the delete without moving it.

a22solmi (changes requested):
Moving the cross does not solve the root issue. Please find a way to disable the delete without moving it.

a22vioja:
The problem was that delete button was always active on the right corner of an element, but not outside it. 

I changed the logic of checkDeleteBtn so that at first it would check if the mouse is over the delete button area and only after that it would check if delete button is active AND I added “!mouseOverElement” to
check if the mouse is not over an element. 

Now it is not possible to delete the element if it is not selected and the delete button is not active if element is not selected.

b22jonan (approved):
I tried to click where the "x" mark is suppose to be to delete it while its not selected, I couldnt get that to work, deleting while selected still works. Id say this works

a22solmi (approved):

Issue #14328, opened 2024-04-03
White object on white background


a22solmi:
## Description
White objects blend into background. This is not an issue in dark mode as the color black is not allowed.
![issue-v1-6](https://github.com/HGustavs/LenaSYS/assets/129150661/cb79b0bf-a80f-42af-a5e3-4edb0bbd09b8)

## Job
Create a thin black outline around affected objects or remove option for white objects on white background.

a21liltr:
Issue has been solved with weekly branch merge.

PR #14531, opened 2024-04-12
G1 2024 v2 #14328

b22samer:
We developed the existing start and end points by making the state charts' start and end points visible when they have a white color with a white background, this was possible with all colors except white. And we made their viewbox larger so that the start and end points view can be seen clearly and are not cropped.


a22solmi:
The person testing should provide screenshots of both light and darkmode

b22chrmo:
States now have a black outline making them visible on white backgrounds. No issues when using dark mode as the outline blends in. Black outline was added to all colors and not just the white states. 

Screenshot 2024-04-16 135531
Screenshot 2024-04-16 140409
Screenshot 2024-04-16 135606

a22solmi (changes requested):
Please remove the outline for color other than white as these are not affected

a23miksk (commented):
The commits work, but the changes are already implemented in the main branch. Since this commit also brings with it a lot of conflicts that need to be solved, this pull request will be closed.

Issue #14391, opened 2024-04-08
Fetch github repo


a22willi:
At the moment, when the button `Download github repo` button is clicked, the repo is never fetched, consequently, the moment selection does not work either. Make sure that whenever a new github repo is added, it is downloaded to the proper location. It should be stored in `LenaSYS/courses/$cid/Github/*`. Issue discovered in [#14351](https://github.com/HGustavs/LenaSYS/issues/14351).

There is a file called `DuggaSys/gitfetchService.php`, which should be able to perform the download. 

- If an error occurs while fetching data from github, it should preferably log the http status code, which file, and when the error occured.
- If it failed to save a file to the webserver, it should preferably log which file it failed to save, when and if possible why.
- The logging should happen to a txt log file, to prevent annoying javascript alerts. (Check if there is existing functionality for this)

The important part of this is that it should be clear why the Github download failed, was it something local, or was it a remote error.

Issue #14351, opened 2024-04-05
Github moment selection does not work


a22willi:
When opening a moment trying to add a github moment, even though there is a github repo selected for the course, the dropdown does not show any of the folders in that repo.

![bild](https://github.com/HGustavs/LenaSYS/assets/129264968/edd9b57e-71a6-4cc3-ae56-c4d1cd995c4c)

At the moment, the code for saving a github to a course does not work. This will be fixed by this [PR](https://github.com/HGustavs/LenaSYS/pull/14346). So I suggest that you try out their dev branch before writing any code. When the pull request has been merged, you can create your own branch.

a22maxra:
When adding dummy data to a folder in `"../courses/$cid/Github/*` the available repositories get displayed in the drop down menu.

![image](https://github.com/HGustavs/LenaSYS/assets/98206230/8bb82e56-fdd1-43dc-9fe6-507997c9267f)

The problem seems to be that a Github folder does not get saved to `"../courses/$cid/`  despite specifying a Github repository.
![image](https://github.com/HGustavs/LenaSYS/assets/98206230/d8b9322f-0ec5-4b7d-82e8-d0f598f84de3)

PR #14535, opened 2024-04-12
Pull request for issue #14391

a22natth:
When a user adds a GitHub repo link in the popup window (see example below), it is saved
to the course it has been assigned to when reentering the home page.

issueScreenshot1

In the example below, the GitHub repo link that was added, will now be visible when editing a course on the home page. Although, if a user has already entered a link when creating a course or when editing a course, the link will not show up when clicking the “Download GitHub repo” button. 

issueScreenshot2

Continuously, when clicking the “Download GitHub repo” button and saving a new link, it is downloaded to the correct location as requested. When a user saves the GitHub repo link, a new folder called “courses” will be created in the LenaSYS folder. The “courses” folder will then create an additional folder which has the name of the course id of the current course(the one the link was added to), which can be seen in the example below. 

issueScreenshot3

Depending on what the GitHub repo link is referencing, the Github moment dropdown will show content accordingly. Due to this functionality being fulfilled, the issue regarding GitHub moment selection seems to have more functionality than previously(see issue #14351). 

issueScreenshot4

Additionally, when clicking on the code examples, it displays further information as seen below. 

issueScreenshot5


DuggaSys/gitfetchService.php
```diff
@@ -142,38 +164,47 @@ function bfs($url, $cid, $opt)
...
+                                // If the file is not part of files to ignore, resume (Index file)
+                                if (!in_array($item['name'], $filesToIgnore) && !in_array($item['name'], $filesVisited)) {
+                                    if ($opt == "REFRESH") {
+                                        insertToMetaData($cid, $item);
+                                        array_push($filesVisited,$item['name']);
+                                    } else if ($opt == "DOWNLOAD") {
+                                        insertToFileLink($cid, $item);
+                                        insertToMetaData($cid, $item);
+                                        downloadToWebserver($cid, $item);
+                                        array_push($filesVisited,$item['name']);
```
a22willi, line 191: Somewhere here it should log wether or not the file was downloaded correctly, so that its possible to know if the code failed, or if the server was rate limited by github. 

Perhaps add some error handling for the `file_get_contents()` function used in `downloadToWebServer()` and return an error if it failed. then log that somewhere using one of the already existing logging functions in LenaSYS. Here is a relevant thread about [handling errors from `file_get_contents()`](https://stackoverflow.com/questions/272361/how-can-i-handle-the-warning-of-file-get-contents-function-in-php).

DuggaSys/gitfetchService.php
```diff
@@ -142,38 +164,47 @@ function bfs($url, $cid, $opt)
...
+                            } else {
+                                if (!in_array($item['name'], $filesVisited)) {
+                                    if ($opt == "REFRESH") {
+                                        insertToMetaData($cid, $item);
+                                        array_push($filesVisited,$item['name']);
+                                    } else if ($opt == "DOWNLOAD") {
+                                        insertToFileLink($cid, $item);
+                                        insertToMetaData($cid, $item);
+                                        downloadToWebserver($cid, $item);
+                                        array_push($filesVisited,$item['name']);
```
a22willi, line 204: Same here.

a22willi (changes requested):

d22mikwe (changes requested):
file data is stored in gitFiles as expected when fetching a github repo inside the course page or when editing the githuburl from main course page.
image
DB - gitFiles

**Note for future issues**: If a course is edited and the same githubURL is used and saved. The data is removed from gitFiles db, but not the relevant gitRepos db table. Data from other pre existing courses was also removed in gitFiles. They are still visible in gitRepo. Only courses that the user has created themselves keeps their data in gitFiles. I have not tested if this is the case if user provides a different url. 
image

Data is stored in gitRepos as expected when fetching a github repo inside the course page or when editing the githuburl from main course page.
image
DB - gitRepos

The githubURL is correctly visible when editing the course if it has been previously fetched.

However, during testing, a course folder($cid) containing a 'Github' folder with relevant files was not created for any of the courses tested apart from folder 1 in courses (Webbprogrammering). folder '1' was a pre exisiting folder and thus only the Github folder containing files was created.
Coursefolder 2 (Webbutveckling - datorgrafik) which is also a pre existing folder has not been tested due to Ajax requests having a limit of 60 /hour (it seems) to github.
**Note for future issue**Could be due to bfs() function in gitfetchService.php 


**Note for future issue**Also. A loading button could be added when fetching a repo from inside the course since screen freezes for a little bit.

d22isaka (commented):
The solution is working as intended.
Using https://github.com/LenaSYS/Webbprogrammering-Examples as example repo I can see it stored all of the files under the course id / github / when pressing on the fetchbutton.

the metadatabase got changes as well.

a22natth:
With further modifications after the pull request was denied at first, we have now added logging which stores the error message in a new textfile. If the content of a file cannot be added or fetched, the message will inform the user of what type of error occurred. Additionally, if a http error occurs (for example 403 or 404), the error code is also displayed in the textfile.

log/gitErrorLog.txt
a22willi, line 1: Log files should not be part of the PR. Add this to .gitignore and re-request review.

DuggaSys/gitcommitService.php
```diff
@@ -263,6 +262,7 @@ function clearGitFiles($cid) {
 	// getCommit: Gets the latest commit from a URL using DOM
 	//--------------------------------------------------------------------------------------------------
 
+	//this is useless, do bfs($url,$cid,"GETCOMMIT") instead where applicable, leaving for now due to it being referenced elesewhere
```
a22willi, line 265: This function is useless, yet its changed?

a22willi (changes requested):

DuggaSys/gitcommitService.php
a22oscja: This function was deemed useless as it would always return an empty string. Considering the function tries to get the latest commit by essentially loading the github website and fetching it from the element that says it, it could reasonably be assumed that this code was simply outdated. Any changes to the function are simply leftovers from testing.

It has been replaced by the new bfs() functionality with GETCOMMIT, but it is still referenced elsewhere so it was deemed reasonable to leave it there for now.

log/gitErrorLog.txt
a22oscja: fixed

d22mikwe (approved):
Function to create folder and writing github files into it is still not working for me. 
As it did however work for Isak i have talked with william and we agreed that this can be approved and be fixed in a future issue. Cause of this is unknown.

See line 102 - 108 in gitfetchService.php for reference.
During work on Issue 14525 i encountered this exact problem where it did not create a folder and write files inside it. I modified the solution to look like this:
image

This code works for me atleast, still unsure what the root problem is.

b22marah (approved):
Looks good, fixes previous requests, and besides mikwe:s comment, which we've agreed upon, it does what it's supposed to do.

Issue #14452, opened 2024-04-10
Fix bug in courseedservice.php


a22iliru:
To reproduce the bug:
Login as 
user: brom
password: password
Edit the first course "Datorns grunder" by pressing the settings icon and change visibility to deleted, save and refresh the page. All courses are now gone and it seems that we have to reinstall lenasys to get the courses back.

Error that is recived from the sevice call to courseedservice.php:
Fatal error
Uncaught PDOException: SQLSTATE[23000]: Integrity constraint violation: 1451 Cannot delete or update a parent row: a foreign key constraint fails (albin.filelink, CONSTRAINT filelink_ibfk_1 FOREIGN KEY (cid) REFERENCES course (cid)) in /Applications/XAMPP/xamppfiles/htdocs/LenaSYS/DuggaSys/courseedservice.php:574
Stack trace:
#0 /Applications/XAMPP/xamppfiles/htdocs/LenaSYS/DuggaSys/courseedservice.php(574): PDOStatement-&gt;execute()
#1 {main}
  thrown in /Applications/XAMPP/xamppfiles/htdocs/LenaSYS/DuggaSys/courseedservice.php
on line 574

courseedservice.php

c22mohel:
This bug appears to happen when you try to delete any course not just the "Datorns grunder" one

c22mohel:
When deleting newly created courses this does not happen for some reason

c22mohel:
It does not remove the courses from the database. It appears that coursedsrvice.php fails to retrieve the courses.

c22mohel:
Most likely this is a foreign key constraint fails. There is a foreign key that  uses the rows on course. this makes deleting a course impossible. 
But this should not be a problem since when deleting a course, the course should not actually be deleted but the visibility should change. 
At the moment it is unclear why newly created course get deleted from the database.

c22mohel:
All tables that use "cid" and their subtables:
all these need to be deleted for this to work 

Subparts
partresult - from subparts

codeexample
impwordlist - from codeexample
box - from codeexample
improw - from box

listentries
user_participant - from listentries

useranswer

timesheet

quiz
variant - from quiz

vers

filelink

programcourse

user_course
course_req

PR #14538, opened 2024-04-12
Fix bug in courseedservice.php #14452

c22mohel:
This should work need to be tested.


c22mohel:
This is now complete need to be tested.

a22iliru:
This does not work and causes problems for multiple files because incorrect naming of attributes. new pull request has to be made with the fixes

PR #14540, opened 2024-04-12
remove all references to the unused variable $appuser and remove unnecessary debug call from test.php.

a22albjo:


c22mohel:
After testing this code does resolve the issue and there are no more $appuser variable.

Issue #14518, opened 2024-04-11
Unwanted navbar effects at certain screen sizes


b22marah:
## New issues found:
Some navbar buttons will jump down below the navbar at certain screen sizes. 

This can be tested by adjusting the screen size. The apperance can be seen below:
![Skärmavbild 2024-04-11 kl  14 19 34](https://github.com/HGustavs/LenaSYS/assets/129263915/efa8e509-53bc-41e2-b251-4c3790f58de2)

The log out button and the username will also not be contained within the navbar at some screen sizes. 

## Issues in bullet point list

- Fix the implementation so that navbar buttons will never jump down, no matter the screen size. The menu buttons should either be in the navbar or in the burger menu. 

- Also fix it so that the course version dropdown does not jump down as well. 

The current appearance when the screen gets to a certain size:
![Skärmavbild 2024-04-11 kl  14 25 03](https://github.com/HGustavs/LenaSYS/assets/129263915/f3a87e5d-c6da-4e0d-8d9e-1655b8dec21b)

- The logout button and the username should not be outside the navbar as well.

The requested appearance (can be seen in Datorgrafik):
![Skärmavbild 2024-04-11 kl  14 24 52](https://github.com/HGustavs/LenaSYS/assets/129263915/d1b33bfd-b1b5-4ee1-acdf-246e2ab5494d)
Note that the username and log out button is contained within the navbar in this screenshot from Datorgrafik.

PR #14542, opened 2024-04-12
G3 2024 v2 #14518 Navbar bug fixed

a22moamy:
Fixed the styling of the navbar so that navbar items don't jump down at certain screen sizes anymore. 

The current appearance: 
Skärmavbild 2024-04-12 kl  15 00 36

Skärmavbild 2024-04-12 kl  15 00 47

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -1373,7 +1373,6 @@ header td {
 }
 
 .navheader {
-  table-layout: fixed;
   height: 20px;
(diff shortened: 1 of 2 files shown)


a22maxra:
Largely works well, however i was able to replicate a  similar bug with the fixed version once but it does not generally seem to be a problem.

image

a22maxra (approved):
Already left a comment

Issue #14344, opened 2024-04-04
Show spinner when saving course changes


a22willi:
Since there is a delay when updating/saving course information in the following box:
![bild](https://github.com/HGustavs/LenaSYS/assets/129264968/fd7fbbd5-a6ff-4472-b96c-017f43653ea9)
Add a spinner (maybe replace the text in the button with the spinner), and disable the button. This is to make it obvious that the page is probably working, and it has not frozen. Note: It also needs to work in dark mode, so make sure you test that. 

[Spinners/Loaders](https://www.w3schools.com/howto/howto_css_loader.asp)

PR #14543, opened 2024-04-12
Fixed a spinner when saving course changes #14344

SleepyBearIV:
added a loder alredy in the system in style.css, added functions that hide and show loader

### DIFF
--- a/DuggaSys/courseed.php
+++ b/DuggaSys/courseed.php
@@ -37,9 +37,17 @@ if(isset($_SESSION['uid'])){
 	<script src="../Shared/js/jquery-ui-1.10.4.min.js"></script>
 	<script src="../Shared/dugga.js"></script>
 	<script src="courseed.js"></script>
-</head>
-<body>
-
+	</head>
+	<body>
+	<script>
+	function showLoader() {
+		document.querySelector('.loader').style.display = 'block';
+	}
+	function hideLoader() {
+		document.querySelector('.loader').style.display = 'none';
+	}
+	</script>
+	
 	<?php
 	include '../Shared/navheader.php';
 	?>
@@ -140,12 +148,15 @@ if(isset($_SESSION['uid'])){
 				</div>
     		</div>
     		<div style='float:right; padding-top:20px; width: 464px;' >
-    			<input id='saveCourse' class='submit-button' type='button' value='Save' title='Save changes' onclick="validateForm('editCourse')" />
+    			<input id='saveCourse' class='submit-button' type='button' value='Save' title='Save changes' onclick="showLoader(); setTimeout(function() {validateForm('editCourse');hideLoader(); }, 1000); "/>
+				<div class="container">
+				<div id='loader' class='loader' style='height: 50px; width: 50px; position: absolute; top: 50%; left: 50%; transform: translate(-50%, -50%); display: none'></div>
+				</div>
     		</div>
       </div>
 	</div>
 	<!-- Edit Section Dialog END -->
-
+	
 
 	<!-- Edit Server Settings START -->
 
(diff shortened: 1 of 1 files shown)


DuggaSys/courseed.php
```diff
@@ -37,9 +37,17 @@
...
-
+	</head>
+	<body>
+	<script>
+	function showLoader() {
+		document.querySelector('.loader').style.display = 'block';
+	}
+	function hideLoader() {
+		document.querySelector('.loader').style.display = 'none';
+	}
```
a22willi, line 48: Indent this.

DuggaSys/courseed.php
```diff
@@ -140,12 +148,15 @@
 				</div>
     		</div>
     		<div style='float:right; padding-top:20px; width: 464px;' >
-    			<input id='saveCourse' class='submit-button' type='button' value='Save' title='Save changes' onclick="validateForm('editCourse')" />
+    			<input id='saveCourse' class='submit-button' type='button' value='Save' title='Save changes' onclick="showLoader(); setTimeout(function() {validateForm('editCourse');hideLoader(); }, 1000); "/>
+				<div class="container">
+				<div id='loader' class='loader' style='height: 50px; width: 50px; position: absolute; top: 50%; left: 50%; transform: translate(-50%, -50%); display: none'></div>
+				</div>
```
a22willi, line 154: Fix indentation.

DuggaSys/courseed.php
```diff
@@ -37,9 +37,17 @@
...
+	function showLoader() {
+		document.querySelector('.loader').style.display = 'block';
+	}
+	function hideLoader() {
+		document.querySelector('.loader').style.display = 'none';
+	}
+	</script>
+	
 	<?php
 	include '../Shared/navheader.php';
```
a22willi, line 52: Indent this.

DuggaSys/courseed.php
```diff
@@ -140,12 +148,15 @@
 				</div>
     		</div>
     		<div style='float:right; padding-top:20px; width: 464px;' >
-    			<input id='saveCourse' class='submit-button' type='button' value='Save' title='Save changes' onclick="validateForm('editCourse')" />
+    			<input id='saveCourse' class='submit-button' type='button' value='Save' title='Save changes' onclick="showLoader(); setTimeout(function() {validateForm('editCourse');hideLoader(); }, 1000); "/>
```
a22willi, line 151: Would it be possible to have the loader disappear after the box has closed instead of a 1 second timer?

a22willi (changes requested):

a22willi:
Also remove the background from the loader.

SleepyBearIV:
> Also remove the background from the loader.

can't, that loader is connected to a lot of functions, not just this one, is a gif with a background. You would have to change the original gif

DuggaSys/courseed.php
SleepyBearIV: It does. The timer just delays both 'validateForm('editCourse')' and 'hideLoader()'. 'validateForm('editCourse')' closes the window when it is done, and 'hideLoader()' runs directly after. Putting it before can make it disappear before it is done validating. However, I can change the time to 0.1 seconds instead, and it will work.

DuggaSys/courseed.php
```diff
@@ -37,8 +37,16 @@
...
-<body>
+	</head>
+	<body>
+	<script>
+		function showLoader() {
+			document.querySelector('.loader').style.display = 'block';
+		}
+		function hideLoader() {
+			document.querySelector('.loader').style.display = 'none';
+		}
```
d22isaka, line 48: Loader ok,
This should be in courseed.js

d22isaka (approved):

PR #14545, opened 2024-04-12
Create microservice deleteCourseMaterial

c22mohel:
deleting all related rows so no key constraints happens when deleting a course.


c22mohel:
This microservice will delete all courses and all there related rows where visibility is 3

a22marca (changes requested):
Incorrect tables written in the queries. Change all queries for these table names:
useranswer to userAnswer
filelink to fileLink

After changing these the code works.
Change the above table names, then it can be merged.

a22iliru:
merged in my pull request

Issue #14533, opened 2024-04-12
Dont show tooltip if there is no github repo


a22willi:
If there is no github repo saved for the current course, the tooltip should not be shown.
![bild](https://github.com/HGustavs/LenaSYS/assets/129264968/953f9297-6992-4cd7-96fa-12b0de9c8534)

This tooltip was added in #14352

There are also several bugs with this feature. If you find one that is an easy fix, simply make a comment here of your findings and fix it.

a22maxra:
Every course that does not have a github repo shows 9:28:31 as the last fetch time, atleast for me at this point in time.

The last fetched time is passed with .$_SESSION['lastFetchTime']

a22maxra:
Added check if 'courseGitURL' (column in course table) is set for the course.
```
// Check if there is a 'courseGitURL' set for the course
$checkIfGithubURL = $row['courseGitURL'];
if ($checkIfGithubURL) {
$updateTime = $row['updated'];
} else { $updateTime = "No Github URL set"; }
```

a22maxra:
Also fixed space between Cooldown: and the time
![image](https://github.com/HGustavs/LenaSYS/assets/98206230/32ebd873-2c26-4848-a260-78f3f82a6d95)

a22maxra:
https://github.com/HGustavs/LenaSYS/pull/14546

a22maxra:
Added some error handling

PR #14546, opened 2024-04-12
G3 2024 v2 #14533

a22maxra:
Added a check to if there is a Github URL set for the course when trying to get the latest fetch.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -3830,6 +3830,8 @@ function validateForm(formid) {
         localStorage.setItem('courseGitHubRepo', repoLink);
         $("#githubPopupWindow").css("display", "none");
         updateGithubRepo(repoLink, cid);
+        location.reload();
+        console.log("TESTAR, RELOADED DET NYSS???");
       }
     }
   }
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -3830,8 +3830,8 @@ function validateForm(formid) {
         localStorage.setItem('courseGitHubRepo', repoLink);
         $("#githubPopupWindow").css("display", "none");
         updateGithubRepo(repoLink, cid);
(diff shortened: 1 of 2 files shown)


a22moamy (changes requested):
This fix creates the following error alert on my mac: 
Skärmavbild 2024-04-16 kl  10 29 22

And then crashes the website as can be seen below:
Skärmavbild 2024-04-16 kl  10 29 28

Please fix this error and request another review.

a22maxra:
Added error some handling

a22moamy (changes requested):
# Small fix required
When you add a github repo URL to a pre-existing course by clicking the "fetch github repo" button, it does not immediately respond to the new URL. After adding the URL, it does not update visually and looks like this: 
Skärmavbild 2024-04-16 kl  14 44 10

Currently, for the tooltip to update, the user has to refresh the page. **Please try to make it so that the user does not have to refresh the page after adding a github repo to a course through the "fetch github repo" button.** If this is not possible, please state why. 

"fetch github repo" button: 
Skärmavbild 2024-04-16 kl  14 51 17

Besides this small fix, the implementation works as intended. Since this branch is behind a merge that fixes new courses, I cannot fully guarantee that it works on new courses.

DuggaSys/sectioned.js
```diff
@@ -3830,6 +3830,8 @@ function validateForm(formid) {
         localStorage.setItem('courseGitHubRepo', repoLink);
         $("#githubPopupWindow").css("display", "none");
         updateGithubRepo(repoLink, cid);
+        location.reload();
+        console.log("TESTAR, RELOADED DET NYSS???");
```
b22marah, line 3834: REMOVE

b22marah:
Previous comments has been fixed, and both me and tester with these fixes thinks it looks good.

Issue #14383, opened 2024-04-08
Resizing elements causes position changes


a22jacka:
## Description 
Elements change their position when getting resized. Expanding horizontally seems to mostly work, but with a tiny position change. Vertical expanding makes top and bottom expand when pulling on the top and pulling on the bottom causes very weird behaviour (see video below)

https://github.com/HGustavs/LenaSYS/assets/129258593/66cdfb6a-7ca2-4823-9d1c-074faf119cf9

## Job
Make the vertical expansion behave like the horizontal expansion and make it so that the element doesn't move then resized

PR #14547, opened 2024-04-12
Fixed issue where resizing elements caused the position to change. #14383

b22chrmo:
Removed 'margin-top' property from all elements. The calculations for the height of the elements when resizing was incorrect due to the 'margin-top' property. 

Resizing elements now work as intended.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -2492,7 +2492,9 @@ function mmoving(event) {
             } else if (startNodeDown && (startHeight - (deltaY / zoomfact)) > minHeight) {
                 // Fetch original height
                 var tmp = elementData.height;
+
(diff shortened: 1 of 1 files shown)


a22robko (approved):
I've tested the two codes and compared how they worked previously and how they are functioning now after the fixes and everything seems to be working well. Previously, it was possible to drag the side edges, without them moving, but not from the top or bottom.  Now, after the issue has been addressed, you can drag from the bottom and top and the "element" stayed in place. Dragging from the top of some elements isn't possible, but that will be a separate issue to be tested and worked on later. I've verified that everything else works properly and there are no bugs with other features in the diagram.php environment. 

Example how they look like:
image1
Screenshot 2024-04-15 at 17 30 10
image2
Screenshot 2024-04-15 at 17 34 00

PR #14558, opened 2024-04-15
Improve consistency in dark mode for course planning svg

a22oscja:
The issue appeared to be the styling of the progressbars being applied by simply applying styling to every odd child of swimlaneSVG.

as the demo course has 9 columns instead of 10 like webbutveckling, the styling got "desynced". Simply changing it to apply specifically to children of swimlaneSVG that also have the class progressbar appears to have fixed it well.

the progressbars themselves still span the entirety of the demo course schedule, but we believe that to be intentional.

### DIFF
--- a/Shared/css/blackTheme.css
+++ b/Shared/css/blackTheme.css
@@ -538,7 +538,7 @@ body {
 }
 
 /* Course schedule styling */
-#swimlaneSVG rect:nth-child(odd){
+#swimlaneSVG rect:nth-child(n of .progressBar){
     opacity: 0.9;
 }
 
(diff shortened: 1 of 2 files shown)


a22moamy (approved):
# Works as intended
The fix works in both light and dark mode for all courses with a planner (Datorns grunder, Demo-course, Webbprogrammering and Datorgrafik). 

The appearance of demo-course in dark mode:
Skärmavbild 2024-04-15 kl  13 11 56


The appearance of demo-course in light mode:
Skärmavbild 2024-04-15 kl  13 12 03

Very good!

a22willi (approved):
Looks good!

Issue #14525, opened 2024-04-12
Store relevant github database entries for codeexample


a22willi:
It was determined that the scope of [#14370](https://github.com/HGustavs/LenaSYS/issues/14370) was too large, therefor the filefetching and creation of db entries has been split up into 2 separate issues. Assuming that the issue has been solved, and files can be fetched from github when clicking save.

The following should be saved to the sqlite db.
![236859351-088624a4-bd36-4e12-8b5d-4635d922543e](https://github.com/HGustavs/LenaSYS/assets/129264968/89397251-8d34-4389-bad3-e21c15cbf15d)

d22mikwe:
Talked with william and a different issue is to be made to improve readability, error handling, input handling, commenting and general effectiveness. Also make sure it doesnt overwrite data etc

PR #14559, opened 2024-04-15
G3 2024 v2 #14370

d22isaka:
We have added functionality to retreive codeexamples through github.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -3929,17 +3929,19 @@ function showFeedbackquestion() {
 }
 
 //Fetch file content from github with ajax request
-async function fetchGitCode(courseid){
+async function fetchGitCodeExamples(courseid){
   
+  let fileNamesArray = [];
   var cid = courseid;
   var fileName = document.getElementById('fileName').value;
   var githubURL = document.getElementById('githubURL').value;
   var filePath = document.getElementById('filePath').value;
   
   var apiUrl = githubURL.replace('github.com', 'raw.githubusercontent.com') + '/master/' + filePath;
-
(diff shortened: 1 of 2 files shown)


a22natth (approved):
As testing was done on this issue, the requirements had slightly shifted but all parties agreed on that the issue would fetch data correlating to the inserted GitHub repo URL and the file path (on GitHub). More specifically, it was important that it was fetched from the actual GitHub repository and not locally, which the pull request accomplished. When the specified file is fetched, additional files with the same name (regardless of type, .js, .html etc.) are also received. This means that the changed requirements for the issue (can be seen in a comment of issue #14370) regarding fetching multiple files with the same name but different types has been done. An example of the fetched data can be seen below, to which the specified file put into the path is also displayed in the console (which in this case was a HTML file), and proves that the issue has been solved. Furthermore, the data for the files are put into an array which contains additional information of the exact URL which can be used to access the file on GitHub. The testing was also done in light and dark theme to which the functionality worked in both modes. In regards to the readability of the code, the comments give a short but overall description of the methods. Although, a comment on line 3933 in sectioned.js states that the array “codeExampleContents” might be changed into a global variable, but depending on what the following functionalities are, I will leave no change request about this. 

image

a22willi (changes requested):
Looks good in general, but how do you plan to store the files on the server when using ajax? Would it not be simpler to use php for this functionality?

d22mikwe:
> 

This will be adressed in 14525

d22mikwe:
> Looks good in general, but how do you plan to store the files on the server when using ajax? Would it not be simpler to use php for this functionality?

This will be adressed in #14525

a22willi (approved):
Looks good

PR #14560, opened 2024-04-15
G2 2024 v2 issue#14420

a22marca:
Created the microservice updateListentrie using code form sectionedservice.php for the `opt` "UPDATE".
Testing is done using the code described in the issue comments.

### DIFF
--- a/DuggaSys/microservices/sectionedService/updateListentrie_ms.php
+++ b/DuggaSys/microservices/sectionedService/updateListentrie_ms.php
@@ -24,17 +24,15 @@ $courseid=getOP('courseid');
 $coursevers=getOP('coursevers');
 $link=getOP('link');
 $kind=getOP('kind');
-$moment=getOP('monent');
+$moment=getOP('moment');
 $visibility=getOP('visibility');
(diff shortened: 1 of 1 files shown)


DuggaSys/microservices/sectionedService/updateListentrie_ms.php
```diff
@@ -0,0 +1,121 @@
...
+$visibility=getOP('visibility');
+$grptype=getOP('grptype');
+$tabs=getOP('tabs');
+$gradesys=getOP('gradesys');
+
+if($feedbackenabled != 1){
+	$feedbackenabled = 0;
+}
+
+if(strcmp($opt,"UPDATE")===0) {
```
a22albjo, line 37: I think this (row 37)might not be needed, if the file is executing i would imagine the opt has already been compared.

a22albjo (approved):
Updates to listentries are working as intended, the changes persists and the right data is sent to the service. The respons is not implemented as noted by the author due to retriveSectionedService.php is not implemented yet.

This can be merged, i noted one small thing that might be a redundancy on row 37, it is not important.

PR #14561, opened 2024-04-15
Fixed the broken clearcontext() call

a22cazlu:


d22abdha (commented):
I recently conducted a detailed test to assess the functionality of the arrow keys on a keyboard. The primary purpose of this test was to verify that the arrow keys operate as expected. During the test, I selected an on-screen element to observe its responsiveness to keyboard commands. I specifically checked whether I could smoothly maneuver this element using the arrow keys. The test confirmed that upon selecting an element, the arrow keys allowed for precise control and movement in all directions—up, down, left, and right. This successful outcome demonstrates the reliability of the arrow keys in facilitating accurate navigation and positioning of screen elements.

Issue #14550, opened 2024-04-15
Create dockerenv


a22willi:
Create a docker environment, that can be brought up on any system with a single command.

Requirements:
- Works on windows, mac, linux.
- After installing the system, no files should be seen as changed by git.
- A single command should bring up the environment.

PR #14562, opened 2024-04-15
G3 2024 v2 #14550

a22willi:
Implemented a docker environment config, that should let users from various systems easily setup a development environment. View the instructions in `dockerenv/docker_install_instructions.md` to test it. I have tested it on windows and linux, where it works. But I have not tested it on mac os. 

When testing, please specify your platform. If it did not work specify error messages.

### DIFF
--- a/dockerenv/.gitattributes
+++ b/dockerenv/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf
--- a/dockerenv/.gitignore
+++ b/dockerenv/.gitignore
@@ -1,0 +1,3 @@
+mysql-data
+githubMetadata
+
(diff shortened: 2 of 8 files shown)


b22chran (commented):
Changed the db: Image: mySQL from a fixed version to latest in the dockerenv/docker-compose.yml. Still got the error response:
Error response from daemon: Mounts denied: 
The path /Applications/XAMPP/xamppfiles/htdocs/LenaSYS/dockerenv/mysql-data is not shared from the host and is not known to Docker.
You can configure shared paths from Docker -> Preferences... -> Resources -> File Sharing.
See https://docs.docker.com/desktop/mac for more info.

Tested on Mac using a M1-chip.

a22danwe (commented):
Not able to set the correct permission when using MAC!
image

a22willi:
More testing and security measures need to be implemented before this can be merged with master. Currently, the docker config files are indexed by the apache server. Waiting for next merge.

PR #14567, opened 2024-04-15
14421 research and create microservice updatequizdeadline

c20juler:


a22albjo (changes requested):
Fully working service, absolute deadline and relative deadline are both working.(relative deadline seem to be relative to 29/12 -14, probably due to this beeing a development environment.) I have however made a summery of three places where there might be redundant code. 

On row 17 - 22 and 34, 35, 39-41 the code does the exact same thing. The code on row 17-22 is redundant and i think it can probably just be removed?

is: include_once "../../../DuggaSys/gitfetchService.php"; on row 11 needed? I cant find where that file is referenced.

Line 43 is probably not needed, if the service is running the opt should already be set to: UPDATEDEADLINE. The opt check/comparison should already be done if we are in the file.

a22iliru:
Will remove the dugga.js commit and merge this later.

a22albjo:
@c20juler @a22iliru This seems fine, we however don't want to merge into master but to the groupbranch i think.

PR #14571, opened 2024-04-15
G2 2024 v2 issue#14209 Create microservice deleteListentries

a22albjo:
The micro service is created and works like the one in sectioned service. However there is a problem where listentries that has a userAnswer connected to it can't be deleted.  This problem appears in both the monorepo and the microservice i will create a seperate issue to take a look at this issue. 

For example listentries in **Webbprogrammering** can be deleted but not those in **Webbutveckling - datorgrafik**

To test the microservice the trashbin can be pressed to run the service and delete a list entry:
Skärmavbild 2024-04-15 kl  14 04 25


c22emigr:
Tested the to delete entries with the micro service instead of the monolith file and this seems to work! Merging.

PR #14574, opened 2024-04-15
G2 2024 v2 issue#14504

b21isabe:
**Fix bugs in dugggaedservice-tests.php**
duggaedservice-tests.php

Issue: Variant tests passes even when the variant id is not retrieved correctly, causing the operations tested to not take place.

all "Test 3 (assertEqual)" tests should now fail when faulty values are entered.

Important for testing:
If you have made changes to the variant table before, you might need to reinstall your database or remove any new rows so that the expected outputs are correct.

If you make the "update variant" or "delete variant" tests fail on purpose, make sure to manually delete the newly created variants in the variants table before testing again (delete the last rows where variantanswer = 'Test text' or 'Test text updated').

### DIFF
--- a/DuggaSys/duggaedservice.php
+++ b/DuggaSys/duggaedservice.php
@@ -155,6 +155,7 @@ if(checklogin() && (hasAccess($userid, $cid, 'w') || isSuperUser($userid) || has
 //------------------------------------------------------------------------------------------------
 $mass=array();
 $entries=array();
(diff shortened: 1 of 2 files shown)


a22marca (approved):
Changing the expected output for each of the variant tests fails the assert tests. I tried changing different parts, like the coursecode and the variants array.

OK

Issue #14486, opened 2024-04-11
entityTypes hard coded


a22solmi:
## Description
Conditions that compare entity types in diagram.js use "ER" or 'ER' when entityTypes.ER should be used instead.

## Jobs
Replace strings in document with the corresponding element type where appropiate

PR #14576, opened 2024-04-15
I searched in diagram.js for Conditions and i added entityTypes. issue #14486

c22abdha:
I searched in diagram.js for Conditions and i added entityTypes. For the following diograms:
"ER" and 'ER'
"UML" and 'UML'
"IE" and 'IE'
"SD" and 'SD'
"SE" and 'SE'


a22solmi (changes requested):
You have put the variables in quotes which makes them into strings. This code will not work. Please remove all surrounding " and '

```
} else if (element.type == 'entityTypes.ER') {
```

a22solmi (approved):

d22hanol (changes requested):
The code has been changed in the way described earlier but the code stopped working. The line functionality is one section that was affected and is no longer working. you can also no longer select any elements, except the last one created. the properties panel has no effect on this last element. 
I do not support merging this code.

https://github.com/HGustavs/LenaSYS/assets/129263218/5b70e200-db7a-4b80-93f3-e271012a7773

PR #14582, opened 2024-04-15
Selecting with ghost entity disable

b22davbe:
You can now no longer select an element when you're in the PLACING_ELEMENT mode. Also deleted an unnessesary console.error that was thrown when trying to create an element on top of another.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1963,7 +1963,6 @@ function ddown(event) {
         switch (mouseMode) {
             case mouseModes.POINTER:
             case mouseModes.BOX_SELECTION:
-            case mouseModes.PLACING_ELEMENT:
                 startX = event.clientX;
                 startY = event.clientY;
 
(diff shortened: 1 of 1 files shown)


a21liltr (approved):
**Behaviour before change:**

When user is in a mode where an element may be placed down ("PLACING_ELEMENT" mode) and the user has previously already placed atleast one element on the diagram (e.g. ER entity, Line or Sequence lifeline, etc), it is possible to still select the placed element with an element ghost active. This can be observed by clicking an element and seeing a purple marker/border appear around the clicked element.

The image below attempts to show this but the cursor is unfortunately not captured during the print -- so please use your imagination instead.

image


**Behaviour after change:**

When the user is in the same mode as mentioned earlier with the same preconditions, a placed element cannot be selected anymore, unless using the Pointer or if the user is in Box Selection mode. Instead an error message appears and informs the user that elements cannot overlap. This happens even when the user has changed the element to place down, but is still in PLACING_ENTITY mode (e.g. trying to select an ER Entity whilst having Line active). 

Image below depicts an attempt to select an ER Entity whilst having an ER Entity ghost (meaning, user is in PLACING_ENTITY mode). Again, the cursor could not be captured so just pretend it is there.

image

Selecting the element works fine when using Pointer or Box Selection.

https://github.com/HGustavs/LenaSYS/assets/129370135/0106fe07-cca7-4460-b8c0-631054aaef3d

**Conclusion:**

The changes made seems to have achieved what has been asked of in the issue.

Issue #14523, opened 2024-04-12
Fix expanding boxes having dissapeared


a22jacka:
## Description
A bug has found it's way in that removes some of the resizing boxes (the blue squares) on some elements

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/35616d7a-6860-4bb2-a2a7-e88bb933a600)
Lifeline - missing bottom

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/35fed12c-8117-403c-97f3-87f6fd50e4a9)
UML, IE, & State entity - missing top

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/7ac022e6-ab00-4be4-805f-1e2a40b3259e)
ER-relation - missing all sides

## Job
Fix so that all elements have all 4 boxes again

PR #14585, opened 2024-04-15
Fixed issues with nodes for resizing not showing. (#14523)

b22chrmo:
Resizing nodes for some of the elements did not show. Nodes are now shown as intended.


b22davbe (commented):
The ER-relation, UML-inheritage and IE-inheritance doesn't have any nodes. Im not sure if that supposed to be that way and in that case the others work now.

PR #14586, opened 2024-04-15
deleted height and weight from CSS

a22vioja:
Weight and height was defined in two places: CSS and javascript . I deleted wight and height from CSS file.


a22cazlu (approved):
I don't have the problem when i am using the g1-2024-v2 branch so it is little hard for me to test this. But if I use the issues branch it also works fine and I can't find any problems with it.

PR #14591, opened 2024-04-15
Update sectionedservice.php issue#14573

a22melst:
Added a query under the opt "DEL" that deletes records in useranswer that references the listentry to be deleted, before deleting the listentry itself.


a22albjo:
Well done, my pull req for the deleteListentries microservice is not reviewed yet. I stole your code from here and commited it to my existing pull-request to make the microservice have youre solution as well.

a22iliru:
Merging this without testing as we can instead test a22albjos pr which includes this fix

PR #14598, opened 2024-04-16
G2 2024 v2 issue#14196

a22eriwi:
Created the ms changeActiveCourseVersion_courseed_ms.php. The features of this microservice are not being used yet because there doesnt seem to be any way to change course versions to being active. But this is probably something that is going to be implemented in the future.


a22marca (changes requested):
Before merge fix the following:
- included paths are incorrect, need to back out two times more
- Add `$opt=getOP('opt')`
- change `cid` and `vers` to courseid and coursevers (so the used variables are declared)
- Format the curly braces

I tested the microservice using the following method.
Getting test data:
select cid, activeversion from course;

test course, original data, cid: 324 vers: 12324
command will update to cid:324, vers: 13371337

The command below works to test if the microservice can update the activevers of a course.
One issue is that since it is just a curl command to the file, there is no active session (no active
user). I had to comment out the following selection: 
```
// Login is checked for function to run
if(checklogin() && isSuperUser(getUid()) == true) {
```

curl command:
```
curl -X POST -d "opt=CHGVERS&cid=324&vers=13371337" http://localhost/LenaSYS/DuggaSys/microservices/courseedService/changeActiveCourseVersion_courseed_ms.php
```
It would of course be better test it by integrating it.

PR #14601, opened 2024-04-16
G2 2024 v2 issue#14552 create updateActiveUsers_ms.php

a22albjo:
The new microservice is in: updateActiveUsers_ms.php.

I have tested it by running this in the terminal: 
curl -X POST -d "hash=AcIj-Ld6&AUtoken=9999" http://localhost/LenaSYS/DuggaSys/microservices/showDuggaservice/updateActiveUsers_ms.php
and compare the output from the instance in the groupdugga table in the database. I have not been able to test the service in the application.

This service is taken from groupDuggaservice.php monorepo.

To make the microservice run instead of the monorepo version change row 1244 in dugga.js to:
	else if(kind=="GROUPTOKEN") {
		if(opt=="UPDATEAU"){
			$.ajax({
				url: "../DuggaSys/microservices/showDuggaservice/updateActiveUsers_ms.php",
				type:"POST",
				data:"AUtoken="+groupTokenValue+"&hash="+hash+"&opt="+opt+para,
				dataType: "json"
			});
		}else{
			$.ajax({
				url: "showDuggaservice.php",
				type:"POST",
				data:"AUtoken="+groupTokenValue+"&hash="+hash+"&opt="+opt+para,
				dataType: "json"
			});
		}
	}

### DIFF
--- a/DuggaSys/microservices/showDuggaservice/updateActiveUsers_ms.php
+++ b/DuggaSys/microservices/showDuggaservice/updateActiveUsers_ms.php
@@ -1,0 +1,10 @@
+### updateActiveUsers
+Collect active users
+Uses service __selectFromTableUserAnswar__ to _get_ information it requires from __userAnswer__.
+<br>
+
+Uses service __insertIntoTableGroupdugga__ to makes _inserts_ into the table __groupdugga__.
+Uses the services __updateTableGroupdugga__ to change the content of these columns:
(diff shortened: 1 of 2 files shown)


c22cargo (approved):
I found it a bit hard to test this microservice, but the CURL command seems to return the correct value so I will assume this implementation works.

Issue #14557, opened 2024-04-15
Fix position for "Back-to-top" arrow


b22marah:
Currently there is a arrow when viewing a course-page, and when you scroll down the page, eventually a "back-to-top" arrow appears.

![image](https://github.com/HGustavs/LenaSYS/assets/129395645/ba2ede3c-5037-4320-bfac-dd1675da1fd4)

However the arrows position is very off, and needs to be adjusted. Adjust the arrow to be positioned above the "+" button, also maybe fix a transparent background for the arrow that changes to the same color as the "+" button when a user is hovered on the arrow.

a22maxra:
Switched position and styling of up ^ arrow, as well at "Back to top" text..
![image](https://github.com/HGustavs/LenaSYS/assets/98206230/d0cd357e-23c2-4087-8226-58b990c55f0d)
![image](https://github.com/HGustavs/LenaSYS/assets/98206230/7b9c2058-aaa6-4271-a4fa-6fc76aa8b13d)

PR #14603, opened 2024-04-16
G3 2024 v2 #14557

a22maxra:
The position of the arrow has been changed as well as the "Back to top" text, there is also now different styling where the arrow ^ turns yellow on hover.

image
image

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -7270,8 +7270,9 @@ only screen and (max-device-width: 568px) {
   border-radius: 6px;
   font-size: 12px;
   opacity: 0.3;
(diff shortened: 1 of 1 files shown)


b22chran (changes requested):
The arrow for "back to top" is working as intended and have a good position underneath the fab-button due to the menu that gets in the way from the fab-button. 
When analyzing the code, the set of the color when hovering is using a hex-code instead of a variable that is used for hovering (var(--color-accent-hover)). Otherwise it seems to be good.

b22chran (approved):
Looks good!

PR #14604, opened 2024-04-16
updated remove course

c22mohel:
changed filelink to fileLink


c22mohel:
This should work.

c22mohel:
fixed all queries for the table names

a22marca (approved):
Removing courses from the course list is possible.
It also removed courses after setting visiblity=3 on a course in the db.

This can be merged

Issue #14549, opened 2024-04-15
Fix the issue with the inspector messing up the grid (#14327) 


a22jacka:
## Description
The issue seems to be caused by the grid not realizing that the size of the window has been changed

## Job
Find a fix so that the grid always take up the whole screen

a21liltr:
The issue seems to appear when a user loads the diagram page with a minimized/smaller window or with inspection tool open, after which, maximizing or closing the inspection window (essentially making the inner window bigger than when initially loading the page) causes the grid to break.

1. Loading in page in Firefox.
![image](https://github.com/HGustavs/LenaSYS/assets/129370135/3e6325a5-0b10-4786-8b79-370ef9fa4011)


2. Expanding/maximizing page to a bigger window than initially (doesn't have to maximize, as long as the window after expansion is bigger than when first loading in).
![image](https://github.com/HGustavs/LenaSYS/assets/129370135/61594048-8f79-42b0-9577-4435b595855a)

PR #14605, opened 2024-04-16
Grid works properly on all webbrowsers regardless of window size and resize. #14549

a21liltr:
Change grid_rect to 100vw/vh as grid_rect is the element filled with the grid pattern and what the user actually sees. Now that it is 100vw/100vh, it covers the whole window, regardless of its parent. It will always be the size of the user's window and the rest of the diagram components are "on top" of the grid. This should work for all browsers.

Slightly changed the Event Listener for resize of window to make it more managable and easy to read. The methods for the rulers on the side are moved and grouped together in one function as it is always 2 specific methods that are needed when updating the rulers.

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -762,8 +762,8 @@
 }
 
 #grid_rect {
-    width: 100%;
-    height: 100%;
+    width: 100vw;
+    height: 100vh;
     overflow: visible;
     fill: url(#grid);
 }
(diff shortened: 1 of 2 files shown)


a22vioja (approved):
I have tested opening and closing the inspector. The grid looks as expected.
I also tried zooming in and out, with or without the inspector - it does not affect the grid.
I also tried making the screen bigger and smaller, minimize it and this also did not disrupt the grid.

I think the problem is solved.

Issue #14615, opened 2024-04-16
Delete button on select


a22solmi:
Merged pull request #14547 solving #14383 introduced a new issue where the delete button was difficult to select when not using the box select tool.

a22solmi:
introduced in  #14528

PR #14616, opened 2024-04-16
Fixed delete button on select #14615

a22solmi:
Fixed formatting issues and deleted line causing bug


a22jacka (approved):
Tried to delete selected and unselected elements and could only delete when elements were selected

PR #14630, opened 2024-04-17
Store relevant github database entries for codeexample #14525

d22mikwe:
Created functionality to store code examples in the sqlite database tables gitFiles and gitRepo in metadata2.db.
Also created functionality to store the code examples in LenaSYS/courses/$cid/

To test that this works. insert a repository and filepath
image

When save button is clicked a folder should be created in LenaSYS/courses/ named after the courseid unless the folder already exists. Inside the folder all codeexamples should appear.

Use DB Browser to check metadata2.db and see if all files have been stored in gitFiles and a gitRepo connection has been made.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -4148,7 +4148,51 @@ function fetchGitCodeExamples(courseid){
       });
     });
   }
-
+  function storeCodeExamples(cid, codeExamplesContent, githubURL){
+    var base64;
+    var decodedContent = [];
+    var shaKeys = [];
+    var fileNames = [];
+    var fileURL = [];
+    var filePath = [];
+    var downloadURL = [];
+    var fileType = [];
+    var i = 0;
+    for(i; i < codeExamplesContent.length; i++){
+        base64 = codeExamplesContent[i].content.content;
+        decodedContent[i] = atob(base64);
+        shaKeys[i] = codeExamplesContent[i].content.sha;
+        fileNames[i] = codeExamplesContent[i].filename;
+        fileURL[i] = codeExamplesContent[i].content.url;
+        downloadURL[i] = codeExamplesContent[i].content.download_url;
+        filePath[i] = codeExamplesContent[i].content.path;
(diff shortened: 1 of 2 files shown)


a22maxra (approved):
Code looks OK to me, files that get saved get stores in the SQLite database, however it is quite unintuitive to save files when you need to specify the correct repo and file path in the correct way. There is more functionality that is to be added but Mikael said this is worked on in another branch that is branched of this?

a22willi (changes requested):
Some small formatting changes requested.

a22willi:
Code looks good now.

a22willi (approved):
Looks good!

a22maxra (approved):
Files still get stored correctly after improvements, in future improvements there should be something to show a file has been retrieved as now you just click "save" and nothing happens on the client side even though the file is fetched.

a22karja (commented):
Everything on the backend seems to work. On the front end i have a few questions marks.
On this screen i have pressed the github button on the marked element
formJacascriptE1

But when i choose a different one the information is still there
HRML5E1

Another thing is that when I pressed saved at first I thought nothing happend because nothing happend on my screen, but when i checked the metadata file the files were there i would suggest either closing the window or some other way to show the user that something happend

a22willi:
> Everything on the backend seems to work. On the front end i have a few questions marks. On this screen i have pressed the github button on the marked element formJacascriptE1
> 
> But when i choose a different one the information is still there HRML5E1
> 
> Another thing is that when I pressed saved at first I thought nothing happend because nothing happend on my screen, but when i checked the metadata file the files were there i would suggest either closing the window or some other way to show the user that something happend

For this issue, nothing needs to happen regarding notifying the user that the files have been downloaded, we will create a future issue to resolve any user experience concerns. But as you pointed out, the information should not be saved when moving between different code examples. This needs to be fixed.

d22mikwe:
"But as you pointed out, the information should not be saved when moving between different code examples. This needs to be fixed." Fix in #14656

PR #14655, opened 2024-04-17
Merge issue#14650 add correct paths for logging

a22marca:
Change the paths for logging.

Before a error was shown in the response from logging.php when visiting e.g. 'show tests' as a teacher in a course


c22cargo (approved):
Works.

Issue #14516, opened 2024-04-11
Fix microservices linking to wrong files


a22iliru:
From a22eddape: 
![image](https://github.com/HGustavs/LenaSYS/assets/164884946/67b6af70-9f45-4930-b518-37567b61a831)

any files linking incorrectly

c22cargo:
It seems that at least "createListentrie_ms.php", "changeActiveCourseVersion_sectioned_ms.php", and "createCourseVersion_ms.php" have incorrect include paths.
Many also use different formats, some use parentheses around the include while others do not. Many use both formats in the same file. Preferably a single standard should be used.
There also seems to be many instances where "include" is used when "include_once" should be used instead.
Many files also have "?>" at the end of the file. That should not be there when the file only consists of php.

A branch will be created to fix this after the weekly merge.

PR #14657, opened 2024-04-17
Fixed incorrect includes, indentation, standards. Issue #14516

c22cargo:
This fixes several files with incorrect includes, indentation, and standard compliance.


a22eddpe:
i fixed some more in the code. There were som places that still needed to be fixed

PR #14659, opened 2024-04-17
Added media querys and made the icons scalable

a22cazlu:
The icons in the toolbar is now scalable and with the added media querys you can now use the program with a much smaller screen. There's still a minimum limit to where the tooltip wont fit but we find it hard to see that a user would ever use such a small screen when using the diagram editor.

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -238,21 +238,33 @@
 .diagramIcons{
     position: relative;
     margin: auto;
-    width: 34px;
-    height: 34px;
+    width: 3.7vh;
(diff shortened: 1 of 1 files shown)


a22alfna:
I have tested this since I personally had problems with not seeing the load button in the toolbar but when I switched to the branch where the issue got fixed I could see every button. I would consider this fixed.

a22alfna (approved):
I have tested this since I personally had problems with not seeing the load button in the toolbar but when I switched to the branch where the issue got fixed I could see every button. I would consider this fixed.

image

image

Issue #14651, opened 2024-04-17
Elements can be placed over the zoom-buttons


a22jacka:
## Description
It's currently possible to place elements over the zoom-buttons by placing elements nearby. Nearby placed elements appear over the buttons so you can't click them

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/bd0b3d1b-45ad-4cdd-98c3-929d1c17976e)

## Job
Place a super-high z-index on the `<div id="zoom-container">` element so that it's always in front

PR #14667, opened 2024-04-17
G1 2024 v3 #14651

a22ricni:
Added a z-index for the class "zoom-container" in css. Can test it by placing for example sequence lifeline on top of the zoom function on the bottom left and see if the element is on top of the zoom button and if you can still use the button.


c22abdha (approved):
image
i have test it and its work

PR #14670, opened 2024-04-17
created the microservice deleteDuggaVariant_ms

a22marca:
The microservice is implemented, see issue comments for help with testing.

### DIFF
--- a/DuggaSys/microservices/duggaedService/deleteDuggaVariant_ms.php
+++ b/DuggaSys/microservices/duggaedService/deleteDuggaVariant_ms.php
@@ -1,0 +1,32 @@
+<?php
+
+date_default_timezone_set("Europe/Stockholm");
+
+// Include basic application services!
+include_once "../../../Shared/sessions.php";
+include_once "../../../Shared/basic.php";
+
+pdoConnect();
+session_start();
+
+$opt=getOP('opt');
+$vid=getOP('vid');
+
+if(strcmp($opt,"DELVARI")===0){
+    $query = $pdo->prepare("DELETE FROM userAnswer WHERE variant=:vid;");
+    $query->bindParam(':vid', $vid);
+
+    if(!$query->execute()) {
+        $error=$query->errorInfo();
+        $debug="Error updating user".$error[2];
+    }
+
+    $query = $pdo->prepare("DELETE FROM variant WHERE vid=:vid;");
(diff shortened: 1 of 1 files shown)


DuggaSys/microservices/duggaedService/deleteDuggaVariant_ms.php
```diff
@@ -0,0 +1,32 @@
...
+$opt=getOP('opt');
+$vid=getOP('vid');
+
+if(strcmp($opt,"DELVARI")===0){
+    $query = $pdo->prepare("DELETE FROM userAnswer WHERE variant=:vid;");
+    $query->bindParam(':vid', $vid);
+
+    if(!$query->execute()) {
+        $error=$query->errorInfo();
+        $debug="Error updating user".$error[2];
```
a22albjo, line 21: this debug output would maybe be more clear:
```suggestion
        $debug="Error when trying to delete user answer".$error[2];
```

DuggaSys/microservices/duggaedService/deleteDuggaVariant_ms.php
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    $query = $pdo->prepare("DELETE FROM variant WHERE vid=:vid;");
+    $query->bindParam(':vid', $vid);
+
+    if(!$query->execute()) {
+        $error=$query->errorInfo();
+        $debug="Error updating user".$error[2];
```
a22albjo, line 29:
```suggestion
        $debug="Error when trying to delete variant".$error[2];
```

DuggaSys/microservices/duggaedService/deleteDuggaVariant_ms.php
```diff
@@ -0,0 +1,32 @@
...
+include_once "../../../Shared/sessions.php";
+include_once "../../../Shared/basic.php";
+
+pdoConnect();
+session_start();
+
+$opt=getOP('opt');
+$vid=getOP('vid');
+
+if(strcmp($opt,"DELVARI")===0){
```
a22albjo, line 15: I think we might want to check if the user has the right privileges to make the change.

a22albjo (changes requested):
I would suggest to take a look at making the errors a bit more clear. 

I also think we want to check if the user has the right privileges to delete the dugga variant.

Otherwise, well done!

Issue #14665, opened 2024-04-17
Fix updateCours_ms.php spelling , remove updateCourse.php, move first one to courseedService folder


a22iliru:
Delete: updateCourse.php
Fix: updateCours_ms.php to updateCourse_ms.php
Then move updateCourse_ms.php to courseedService folder.

a22marca:
I've deleted the unused file and move the misspelled one to the correct path and name.  I also took a look at the code in the file, and it seems it's not correctly done. Some parameters are not retrieived for example. I test the microservice to make see if it works before creating a pull request.

a22marca:
I've integrated the microservice, and no requests are properly handled in the microservice
Error:
```
 Fatal error: Uncaught Error: Call to undefined function getOP() in /var/www/html/LenaSYS/DuggaSys/microservices/courseedService/updateCourse_ms.php:6 Stack trace: #0 {main} thrown in /var/www/html/LenaSYS/DuggaSys/microservices/courseedService/updateCourse_ms.php on line 6
``` 
This error is likely thrown because some files are not included.

a22marca:
Will create a new separate issue to fix the implementation of the updateCourse_ms.php microservice.

PR #14677, opened 2024-04-17
G2 2024 v3 #14665

a21sahfa:


a22marca (changes requested):
This implementation has a few problems:
- the PDO is not initalized using `pdoConnect`.
- The parameters from POST is sanitized
- A echo to the page when the update is successful
- all opt's are not retrieved.

Furthermore,  I believe the query is  incorrect. It should probably be the one listed [here](https://github.com/HGustavs/LenaSYS/blob/g2-2023-v7/backend-models/microservices/duggaSys%20services/Database_related_micro_services.md#copycourseversion-5). This is based on the description of the service [here](https://github.com/HGustavs/LenaSYS/blob/6874b5277f63c2ca2e0a60429d0c22dc1878c3fb/backend-models/microservices/duggaSys%20services/Potential_micro_services.md#updatecourse).
```
UPDATE course SET coursename=:coursename, visibility=:visibility, coursecode=:coursecode, courseGitURL=:courseGitURL WHERE cid=:cid;
```

This microservice need to be looked over again

PR #14678, opened 2024-04-17
Remove gradesystem input on the kind type 3

a22albjo:
Issue: Its not possible to update the "gradesystem", "visibility" or "moment" of a listed entry in a course. 

I have not been able to recreate the problem with "visibility" or "moment", it works fine for me. however "gradesystem" was not working on listentries with kind 3(type: tests)

After doing some research it seems that only kind 4 should be able to have gradesystem changed.

I remove the gradesysteminput field from listentries with kind 3. When you go into settings for a listentrie that has type: test the inputfield for gradesystem will no longer be shown.


a22marca (approved):
Changing the gradesystem and visibility works without any issues for me using the provided code.

There is still a issue when changing the moment for a listed entry. I suggest we merge this branch and then create a new issue for "changing the moment of a listed entry in a course" specifically.

Issue #14548, opened 2024-04-15
label <<includes>> shows wrong text


a22solmi:
## Description 
When selecting a line there is a option to create a label and a button that should fill it with "<<includes>>".
Before deselecting the line the text displays text other than "<<includes>>".
![Screenshot from 2024-04-15 08-44-11](https://github.com/HGustavs/LenaSYS/assets/129150661/105677f1-622d-4fe1-9047-d9d9fca1c19f)

## Job 
Find a way to make the text display in the right way

b22samer:
We have noticed a problem in our code where we are trying to display "<<Include>>" both in a text box and on the actual diagram. We have managed to adjust the text box to correctly display the value "<<Include>>" by changing a value in the code, but when we try to display the same value on the actual diagram, there seems to be an issue. Instead of showing "<<Include>>", only the "<" part is displayed for some reason.

PR #14680, opened 2024-04-17
G1 2024 v3 #14548

a22robko:
We have now managed to make it work so that when you press the button, "\<\<include\>\>" appears instead of "&#60&#60include&#62&#62". Additionally, the white background has been adjusted to fit the text better than before when it was much longer.
So for the textfield, converted character encodings into angle brackets; Then for the label, convert angle brackets into character encodings.
Here is an image of what the diagram looks like now:
Screenshot 2024-04-17 at 14 12 00

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -3109,7 +3109,8 @@ function changeLineProperties() {
 
     if (line.label != label.value) {
         label.value = label.value.trim();
-        line.label = label.value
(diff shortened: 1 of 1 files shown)


a22alfna (approved):
The <<include>> label seems to be fixed and looking at the changes in the document it seems like the last label was hardcoded as the value and it looks like that has been fixed. I would consider this issue fixed

Before changes:
image

After changes:
image

Issue #14291, opened 2024-04-02
Toolbar selection remains open


a22solmi:
## Description
The menu from a "drop down" item in the toolbar remains open after selecting another item.
It only closes by selecting an object in the menu itself.
![Issue-v1-1](https://github.com/HGustavs/LenaSYS/assets/129150661/443cc03e-a3b7-42e8-8f58-8349296130e4)

## Job
Solve the issue so that selecting any other item closes the menu.

PR #14683, opened 2024-04-17
G1 2024 v3 #14291

a22ricni:
You can now have the pop-out menu hide regardless where you click on the website. Can test it by have the pop-out menu open and click somewhere and see if the menu hides.


a22cazlu (approved):
Looks good, the pop out disappears when I click somewhere else in the diagram and i can select all the items in the pop out.

Issue #14666, opened 2024-04-17
Fix error with `show tests`


a22willi:
When clicking `Show tests` in the navbar below:
![image](https://github.com/HGustavs/LenaSYS/assets/129264968/65c922fa-aa54-4b9d-9f7a-9e49435ab20a)
The following error appears.
![image](https://github.com/HGustavs/LenaSYS/assets/129264968/2baf2a73-bf58-496c-a962-f428732d5599)

PR #14684, opened 2024-04-17
G3 2024 v3 #14666 - Fixed "Show tests"

a22natth:
# The solution
We removed all references to timer.js and clickcounter.js since these have been removed since a previous merge. These files only consisted of commented out code and were deemed excessive (and were thus deleted). 

This issue was fixed by uncommenting the pdoConnect() in duggaed.php. 

# Potential new issue found
Pressing “Give students access to the selected version” will result in a crash in all courses. This needs to be fixed.

The “Give students access to the selected version” button can be found in the header of a course:
image
The result: 
image
This new potential issue does not have the same problem as ours and requires a different solution (since pdoConnect() is not commented out in this page).

### DIFF
--- a/DuggaSys/duggaed.php
+++ b/DuggaSys/duggaed.php
@@ -3,7 +3,7 @@
 session_start();
 include_once "../Shared/sessions.php";
 include_once "../Shared/basic.php";
-//pdoConnect();
+pdoConnect();
 $cid=getOPG('courseid');
 $vers=getOPG('coursevers');
 ?>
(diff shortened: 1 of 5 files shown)


a22maxra:
There is no longer an error and the page works however i see a warning when entering the "Show tests" page. 
image

a22maxra (changes requested):
Warning message shows see image

a22maxra (approved):
Now works, gr8.

PR #14687, opened 2024-04-17
replaced color strings #14485

a22solmi:
Replaced all colors strings with variables


a22ricni (approved):
I looked at the week branch and this issue branch and it looks good. Both the week branch and issue branch looked the same, also looked at the code and it seems good of what I could see.

Issue #14622, opened 2024-04-17
Node zoom scaling


a22solmi:
## Description 
"Down" node does not scale with zoomfact
![Screenshot from 2024-04-17 08-39-36](https://github.com/HGustavs/LenaSYS/assets/129150661/9892695c-2783-4028-b889-8c775824a692)

## Job
Make the zoom scaling consistent between all nodes

PR #14689, opened 2024-04-17
Fixed issue with some nodes not scaling with the zoomfactor. (#14622)

b22chrmo:
Some nodes (bottom node on most elements & top and bottom node on super states) would not scale when zooming in and out. Nodes now scale properly when zooming.

Also removed redundant code related to nodes.


a22solmi (approved):
All entities that can be resized now scale with zoom. Tried to resize ER, IE, UML. SD and ER attriubute.

Screenshot from 2024-04-17 08-39-36

PR #14691, opened 2024-04-17
created the microservice createDugga_ms

a22marca:
Created the microservice, check issue comments for testing was done.


a22melst (changes requested):
Never checks if it is a group assignment. Otherwise it works.

Issue #14572, opened 2024-04-15
Fix very tall states


a22jacka:
## Description
When state-entities become very tall, the top and bottom of the entity start to twist around. A guess is that this is caused by the corner-rounding using a relative unit and eventually they mess up and the top and bottom looks twisted, so it could maybe be changed by either using a constant unit or introducing limits to the bending

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/59438314-9678-49c4-a10b-2640e1063693)

## Job
Fix so that the corners are always just rounded regardless of size or proportions.

PR #14692, opened 2024-04-17
issue #14572 is now fixed

d22hanol:
Issue #14572 was a problem where the shape became wierd when adding to the height of the element. A bug regarding attributes in the SDelement was also fixed.


b22jonan (approved):
tested making a very tall object, which still looked good, i dont see any problems with the code either

Issue #14425, opened 2024-04-09
Research and create microservice setVisibleListentrie


a22iliru:
file: sectionedservice.php

Se: https://github.com/search?q=repo%3AHGustavs%2FLenaSYS+setVisibleListentrie&type=code

a22iliru:
I will reopen pull request and fix it later. it merged into master (reverted)

PR #14697, opened 2024-04-18
SubmitDugga

a22fokol:
I have done what i could on this issue to " Create test: submit dugga in showDuggaservice_test.php"  by creating a new file "SubmitDugga_Test.php".  I have aproble with the file path cause it's untracked.    

I had  also make an mistake by pulling  this issue in the issue #14425 l


a22marca (changes requested):
The added file in the commit does not have a file extension, so it's not possible to perform the code in it when navigating to it in the browser. However, the test itself already exists in the file showDuggaservice_test.php as stated in the issue description.

The issue was about gathering a new expected-output for the already implemented test "submit dugga" in showDuggaservice_test.php. The gathered expected out in the newly created file still uses the same expected-output.

a22fokol:
I'm still working on it to fix the issue but i haven't managed to resolve the "untracked file" issue that is blocking the visibility of "submit Dugga" in the repository. I tried different commands like (git add submitDugga_test.php)
   and (git commit -m "Add submitDugga_test.php") to resolve the issue.

PR #14710, opened 2024-04-18
ctrl + a sets mouse to pointer

a22cazlu:
Simulate a click at mouseMode0 when ctrl + a is pressed before the select all so the user always comes to the select mode.


a22vioja (approved):
I have tested marking all elements on the screen with "Ctrl+A" while the line creation tool is selected, and it works. After selecting all, the mouse is set back to default, which is how it should work.

Screenshot 2024-04-18 at 13 56 08

Other parts of the diagram work as before, so these changes did not affect anything else in the program.It seems that the issue is solved.

I have not noticed anything specific in the changed code part that could cause bugs or errors.

Issue #14587, opened 2024-04-15
Fix bugs in showDuggaService.php


a22iliru:
When trying to save a dugga a error as recieved as response from showDuggaService.php.

Request:
`courseid=2
did=2
coursevers=97732
moment=2003
segment=2001
hash=UNK
password=UNK
opt=SAVDU
answer=2%2097732%202003%2000000001%200%200%201408%20880%20777%20748
log_uuid=mn11hhYj2brFXjG
variant=undefined`

Response:
`Fatal error: Uncaught Error: Class "ZipArchive" not found in /var/www/html/LenaSYS/DuggaSys/showDuggaservice.php:170 Stack trace: #0 /var/www/html/LenaSYS/DuggaSys/showDuggaservice.php(378): processDuggaFiles() #1 {main} thrown in /var/www/html/LenaSYS/DuggaSys/showDuggaservice.php on line 170`

b21isabe:
I have started investigating this.
notes for showDuggaService:
line 170 - a new zipArchive object is made
line 129 - ProcessDuggaFiles() function
line 378 - processDuggaFiles() is called, and the error is thrown

b21isabe:
I have replicated the error by opening Biträkningsdugga 2 in the course Webbutveckling - datorgrafik and when trying to save it, got the error: "Unable to save dugga!"

I will look into lines 278-328 in showDuggaService to see if there is anything there that is causing the issues.

b21isabe:
I have done some more testing and discovered that the error only seems to show up when signed in as a superuser, when signed in as either a guest or student saving works fine.

b21isabe:
The initial error reported was most likely because the person who found the error was using linux and had not configured zip-archive or had some other local issues with zip-archive. This can be read about here: [install guide](https://github.com/HGustavs/LenaSYS/blob/60fa0a4f5d03b176fc48452ed0b53346defc3cbc/README.md#to-get-zip-archive-to-workused-in-download-zip-function)

I have not been able to replicate the error message in the issue description but I have found some other problems that keep users from saving/viewing a saved dugga. The first one is that when a dugga is saved the generated link links to the wrong folder for example: 
http://localhost/sh/?s=21f0f62c
 when it is supposed to be 
http://localhost/Lenasys/sh/?s=21f0f62c
[...]

PR #14712, opened 2024-04-18
G2 2024 v3 issue#14587 - Fix bugs in showDuggaService

b21isabe:
Original issue: #14587 

I could not replicate the original error message received in the issue. My guess is that the cause of that error was that the user reporting it was using linux and/or had not configured ZIP-ARCHIVE as is explained in the [install guide](https://github.com/HGustavs/LenaSYS/blob/60fa0a4f5d03b176fc48452ed0b53346defc3cbc/README.md#to-get-zip-archive-to-workused-in-download-zip-function).
I did not have any issues with ZIP-ARCHIVE since I am using windows where it is built in.

I did however find a bunch of different bugs that kept the user from being able to save a dugga.
- When a dugga is saved the generated link linked to the wrong /sh folder instead of /LenaSYS/sh.
- Loading a saved dugga in ValidateHash.php also linked to /sh instead of LenaSYS/sh.
- Uploading a pdf/zip sent the user to /sh instead of back to the dugga.
- Javascript in the dugga template files were overwriting the contents of the submission receipt, breaking the markup and undoing the fixes made to the showDugga.php file.

I have fixed these and saving/loading is working as it should on my system now.

To test:
Login to a student account, since teachers are not allowed to submit (I used the acc mestr - password) .
Enter the course: Webbutveckling - datorgrafik, this one has all types of dugga.
Try submitting and loading the different types of dugga.

### DIFF
--- a/DuggaSys/filereceive_dugga.php
+++ b/DuggaSys/filereceive_dugga.php
@@ -516,7 +516,9 @@ if(!$error){
 //echo "{$hash}|";
 //echo "{$hashpwd}|";
 //echo "{$variant}|<br>";
-header("Location: /sh/?s=$hash");
+
+//redirects back to the dugga
+header("Location: showDugga.php?did=$duggaid&courseid=$cid&coursevers=$vers&moment=$moment&cid=$cid");
 exit();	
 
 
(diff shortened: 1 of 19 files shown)


a22eriwi:
@b21isabe
Please fix the conflicts with the DuggaSys/validateHash.php file before merging.
Also there is a spelling error in the DuggaSys/showDugga.php file on line 272. It says "Besure" but it's supposed to be "Be sure"
Good job otherwise!

b21isabe:
Conflicts and typo has been fixed

PR #14713, opened 2024-04-18
G2 2024 v3 issue#14640 Create microservice deleteCodeExample

a22albjo:
See issue more info on ex how to test the service.

### DIFF
--- a/DuggaSys/codeviewerService.php
+++ b/DuggaSys/codeviewerService.php
@@ -291,10 +291,10 @@
 				return;
 			} else if (strcmp('DELEXAMPLE', $opt) === 0) {
 
-				$query1 = $pdo->prepare("DELETE FROM box WHERE exampleid=:exampleid;");
+				$query1 = $pdo->prepare("DELETE FROM improw WHERE exampleid=:exampleid;");
 				$query1->bindValue(':exampleid', $exampleId);				
-
-				$query2 = $pdo->prepare("DELETE FROM improw WHERE exampleid=:exampleid;");
+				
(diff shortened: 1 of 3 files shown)


b21isabe (changes requested):
Tested by changing the service in dugga.js and then entering the course webbprogrammering clicking on a code example, clicking the cogwheel and then clicking remove.

I first tested the original monolith service which failed to run and gave this error:
```
:  Uncaught PDOException: SQLSTATE[23000]: Integrity constraint violation: 1451 Cannot delete or update a parent row: a foreign key constraint fails (`isakdb`.`improw`, CONSTRAINT `improw_ibfk_2` FOREIGN KEY (`boxid`, `exampleid`) REFERENCES `box` (`boxid`, `exampleid`)) in C:\xampp\htdocs\LenaSYS\DuggaSys\codeviewerService.php:310
Stack trace:
#0 C:\xampp\htdocs\LenaSYS\DuggaSys\codeviewerService.php(310): PDOStatement-&gt;execute()
#1 {main}
  thrown in C:\xampp\htdocs\LenaSYS\DuggaSys\codeviewerService.php
on line 310
```

I then tested the microservice which worked as it should. This was done simply by swapping the order of query1 and query2. I didn't find any issues with the microservice.

Before merging:
It would be nice if the monolith could be fixed the same way the microservice was.
the dugga.js file should be restored to use the monolith service on line 1177.

a22albjo:
@b21isabe Thank you for the review, requested change is made

PR #14722, opened 2024-04-18
add sessions.php, remove duplicate row in database.php.

a22albjo:
Add include on sessions.php.


c22mohel (approved):
This works like it is supposed to.

Issue #14635, opened 2024-04-17
Create microservice deleteFileLink


a22iliru:
Count rows where these conditions hold. This will indicate if the file is in used.

- __FROM__ fileLink, box __WHERE__ box.filename = fileLink.filename __AND__ (fileLink.kind = 2 __OR__ fileLink.kind = 3) __AND__ fileLink.fileid=:fid ;

If not in use, then service __deliteFromTableFileLink__ is used to delete a row from the table __fileLink__.

This service contains two seperat instances of the above mentiond code, placed in an if and else statement. 

`Note that the description may be wrong or missing steps, do as u think is right.`

c22cargo:
This seems to be related to "fileedservice.php".

PR #14723, opened 2024-04-18
G2-2024-V3 Issue #14635

c22cargo:
The microservice is used when a file is deleted from a course. You can create new files and delete them for testing purposes. New files can be created from the UI and will be placed in ```LENASYS/courses```, the subfolder is determined by if the file is global or local.

The original code from the monolith can be found inside ```fileedservice.php``` on row 63-142.

The implementation was tested by changing the ```kind=="FILE"``` part in the ```AJAXService()``` function inside ```dugga.js``` to:
```
else if(kind=="FILE"){
	if (opt=="DELFILE"){
		serviceURL="../DuggaSys/microservices/fileedService/deleteFileLink_ms.php";
	}
	$.ajax({
		url: serviceURL,
		type: "POST",
		dataType: "json",
		success: returnedFile
	})
```

### DIFF
--- a/DuggaSys/microservices/fileedService/deleteFileLink_ms.php
+++ b/DuggaSys/microservices/fileedService/deleteFileLink_ms.php
@@ -3,24 +3,43 @@ date_default_timezone_set("Europe/Stockholm");
 
 include_once "../../../Shared/sessions.php";
 include_once "../../../Shared/basic.php";
+include_once "../shared_microservices/getUid_ms.php";
 
 // Connect to database and start session
 pdoConnect();
 session_start();
 
 // Global variables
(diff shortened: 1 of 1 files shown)


a22melst (changes requested):
There is no authorization (checklogin etc). I also get redirected to courseed.php after deleting a file, which doesn't happen when using the monolith. Other than that it seems to work.

a22melst (approved):
Couldn't test the permission checks because I couldn't even open the file editor without permission, but the code looks correct. The redirecting still occurs because of "echo json_encode($debug);" but that can be fixed when the microservice is integrated into the system.

Issue #14642, opened 2024-04-17
Create microservice updateDugga


a22iliru:
Seems like there is a function in duggaed.js that does this, so it should be replaced with this.
Uses the services __updateTableQuiz__ to change the content of these columns:
- coursename
- visibility
- coursecode
- courseGitURL
- qname
- autograde
- gradesystem
- quizFile (template)
- qstart (start date)
- deadline
- qrelease (release date)
- jsondeadline
- group

`Note that the description may be wrong or missing steps, do as u think is right.`

a22iliru:
created new issue

a22marca:
the microservice is now implemented and worked. I tested it by adding this code to dugga.js (line 1061):

```
	}else if(kind=="DUGGA"){
		let service_url = "duggaedservice.php";
		if (opt==="SAVDUGGA"){
			service_url = "../DuggaSys/microservices/duggaedService/updateDugga_ms.php";
		}
		$.ajax({
			url: service_url,
			type: "POST",
			data: "opt="+opt+para,
			dataType: "json",
			success: returnedDugga
		});
```

This functionality can be tested by logging in as a teacher, going into a course and then "show tests". Editing a dugga can be done by clicking the cog icon.

PR #14724, opened 2024-04-18
G2 2024 v3 #14642

a21sahfa:
created a new php file in the microservices folder in courseedservices with the name updatedugga. transfered the function updatedugga from duggaed.js to updatedugga_ms.php file


c22cargo (changes requested):
I don't know how you tested this because there is just a bunch of javascript inisde of a PHP block. The indentation is not correct either. Also, rename the file to use camelCase.

a22fokol:
When doing the testing, it says the code have a probem to connect to the database . It seems like the code need to have an "PDOException" to Management  Resources.  this is the error that comes up when testing it.

Fatal error: Uncaught PDOException: could not find driver in C:\xampp\htdocs\LenaSYS\Shared\basic.php:173 Stack trace: #0 C:\xampp\htdocs\LenaSYS\Shared\basic.php(173): PDO->__construct('sqlite:../../lo...') #1 C:\xampp\htdocs\LenaSYS\DuggaSys\microservices\duggaedService\updateDugga_ms.php(6): include_once('C:\\xampp\\htdocs...') #2 {main} thrown in C:\xampp\htdocs\LenaSYS\Shared\basic.php on line 173

Issue #12756, opened 2022-05-20
Panning the diagram to a lenght up to 1 thousand lags the process almost to a halt.


a20jonda:
This is likely due to the ruler drawing everything to the length.

b22jonan:
it starts lagging due to adding more and more objects without cleaning out the old ones, resulting in thousands of html elements which results in the lag. the fix for this would either to render the rulers another way without saving entities, or to clean up the elements as you pan further away. 
![image](https://github.com/HGustavs/LenaSYS/assets/61019188/76433f08-83ca-440d-b450-d5fe7285a426)

b22jonan:
fixing this seems to be a bit wierd since the coordinate system used for the rulerbars is somewhat screwed up, currently trying to get a range where bars are visible, but that doesnt seem to work since the coordinates for bars get moved as you pan.
however this should still be possible to fix as long as i can get the range to be correctly dynamically generated.

b22jonan:
sort of fixed now, introduced problem with 10cm bars not being drawn for some reason

PR #14726, opened 2024-04-18
fix for #12756

b22jonan:
added ranges to the drawRulerBars function so it only draws lines that would be visible 
also added a quickfix for the long bars since they dissapeared if i just added the range.


a22ricni (approved):
It seems to be working. I tested it by first be at coordinate 1000,1000 and it lagged in the week branch. Then switch branch to this issue and i went to 1000,1000 coordinate, and it didn't lag.

a22jacka (changes requested):
Tried zooming a bit and it messed up the rulers alot

bild
4x zoom
bild
1.25x zoom

Could also add some spaces on if-statements that it follows `if (...) {`. Can also be fixed afterwards so not the most important

b22jonan:
> Tried zooming a bit and it messed up the rulers alot
> 
> bild 4x zoom bild 1.25x zoom
> 
> Could also add some spaces on if-statements that it follows `if (...) {`. Can also be fixed afterwards so not the most important

fixed the problem witht the zoom, should work now

b22jonan:
fixed issue completely this time, no issues when zooming in or out

a22jacka (approved):
The rulers now take up the entire width/height like they're supposed to

a22ricni (approved):
It seems to be working. Tested it first by first be at coordinate 1000,1000 and it lagged in the week branch. Then switch branch to this issue and i went to 1000,1000 coordinate, and it didn't lag. And also tested the zoom if it still works and I didn't find any problem with it.

Issue #14632, opened 2024-04-17
Align rulers and origin


a22jacka:
## Description
The '0'-point on the rulers and the origin lines and slightly miss-aligned. 

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/1aac5781-a229-4f47-89ed-31d069a86d0c)

Because of the rulers position this can be fixed in css using left/right and top/bottom 

## Job
Align the origin and rulers

PR #14727, opened 2024-04-18
G1 2024 v3 #14632

a22ricni:
Now should the '0'-point on the rulers and the origin lines be at the right position and are aligned to the grid. Can test it by zoom in and out to see if they are aligned, the ruler to the grid.


b22jonan (approved):
zoomed in and out and panned to different places, it seems to be correctly aligned now

PR #14731, opened 2024-04-18
2 Fixed White object on white background #14328

a22robko:
We have updated our issue to 'just prevent the white color from being selected, just like how it works for the black color in dark mode'. We searched through the code and finally found that changing #FFFFFF to #ffffff made it so that the white circle would function the same with a white background as the black circle does with darkmode.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -8787,10 +8787,11 @@ function drawElement(element, ghosted = false) {
                     onmouseenter='mouseEnter();' 
                     onmouseleave='mouseLeave();'>
                     <svg width="100%" height="100%" 
-                        viewBox="0 0 24 24"
+                        viewBox="-2 -2 28 28"
                         xmlns="http://www.w3.org/2000/svg" 
                         xml:space="preserve"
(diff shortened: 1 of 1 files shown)


a23ponek:
No changes seemed to have been made on the new branch. Cant verify the created outline when the color is white.
Screenshot 2024-04-22 082529
Screenshot 2024-04-22 082852

a22solmi (changes requested):
See comment above

a23ponek (commented):
The outline is now visible when the color white is selected for the circles. All other colors behave as usual with no outline needed.
First image is from my local and the second image is from the branch g1-2024-v3-https://github.com/HGustavs/LenaSYS/issues/14328
Screenshot 2024-04-22 082529
Screenshot 2024-04-22 124336

a22solmi (changes requested):
This does not look good. Please just prevent the white color from being selected, just like how it works for the black color in darkmode

a23ponek (commented):
The initial circle now displays as a black outline and looks better than before since its visible. The final state remains the same as before and I consider it alright. See the image for how it looks now down below.
Screenshot 2024-04-24 081202

Issue #14686, opened 2024-04-17
Rename all microservices to have camelCase


a22iliru:
Also copy over all files in "showDuggaservice" to "showDuggaService" then delete showDuggaservice folder.

a22albjo:
Also merge showDuggaservice and showDuggaService to one map

PR #14733, opened 2024-04-18
G2 2024 v3 #14686

a21sahfa:


c22cargo (changes requested):
There are still two folders "showDuggaservice" and "showDuggaService". Also, you seem to have commited a microservice "Updatedugga_ms.php", which just contains a bunch of javascript inside of a PHP block, and also does not even use camelCase. The issue was not to create any microservices, it was to rename files and combine two folders.

a21sahfa:
på min dator fanns bara showDuggaService, jag kollade i terminalen, konsolen osv men showDuggaservice fanns inte.

Issue #14709, opened 2024-04-18
UML entities don't scale


a22jacka:
## Description
When the UML entity has a lot of attributes, the rest of the entity doesn't scale so the attributes appear behind the classes and eventually are hidden below the entity

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/fe1b8821-00fe-4a1f-bb1e-f18b4c133bf7)
How it's supposed to be

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/9f2a765e-093c-43a4-bd1c-dc193d0c860b)
Broken

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/562c7300-7610-4e52-9c5c-8f8f6023ad8a)
Broken with overflow visisble

## Job
Make so the entity scales correctly with the middle box

Issue #14701, opened 2024-04-18
break out UMLEntity into functions


a22solmi:
## Description
drawElement() is a 900 line long function. There is a lot of repetition and if else cases that are hard to read and understand.
Break out UMLEntity into a function.

## Job
Break up this function into smaller pieces to increase readability

PR #14737, opened 2024-04-18
fixed issue #14709

d22hanol:
UML class entities now scale with the amount of attributes and the amount of functions related to the class. 
image


a22solmi:
solved by #14701

PR #14738, opened 2024-04-18
Fixed updateListentrie test again - issue#14393

a22melst:
Only works when "Testing-Course" is installed.


c22cargo (changes requested):
The test works but you are using different indentation formats. For the service-data it looks like:
```
'service-data' => serialize(
    array(
```
And for the filter-output it looks like:
```
'filter-output' => serialize(array(
```

Also, you are currently only testing a name change. It would probably be better to test changing several attributes, since it is a test after all.

c22cargo (approved):
Looks good

PR #14739, opened 2024-04-18
 UppdateSubmittedDugga_Test.php

a22fokol:
The Create test: update submitted dugga in showDuggaservice_test.php has be created now.

### DIFF
--- a/DuggaSys/tests/UppdateSubmitDugga_Test.php
+++ b/DuggaSys/tests/UppdateSubmitDugga_Test.php
@@ -1,0 +1,48 @@
+<?php
+
+
+
+include "../../Shared/test.php";
+
+$testsData = array(
+
+    'update submitted dugga' => array(
+        
+        'expected-output' => '{"debug":"NONE!","param":"{\"tal\":\"10\"}","answer":"an answer","danswer":"{\"danswer\":\"00002 0 A\"}","score":0,"highscoremode":"","grade":"UNK","submitted":"","marked":"","deadline":"UNK","release":"UNK","files":[],"userfeedback":"UNK","feedbackquestion":"UNK","variant":"UNK","ishashindb":false,"variantsize":"UNK","variantvalue":"UNK","password":"password","hashvariant":"UNK","isFileSubmitted":"UNK","isTeacher":1,"variants":[],"duggaTitle":"Inserttobedeleted","hash":null,"hashpwd":null,"opt":"UNK","link":"https:\/\/localhost\/sh\/?s="}',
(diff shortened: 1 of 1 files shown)


a22marca:
The tests seems to pass after reinstalling lenasys (refreshing the database). Will try to manually check the test to see that it is correct.

a22marca:
After manually doing the before-test queries, and then retrieving the 'did' and 'moment', this is the response from the following request:
Request:
```
Parameters:
hash=ghj1ghj2
hashpwd=asddasdd
username=stei
password=password
courseid=9999
coursevers=1337
did=40
moment=5010

Curl:

curl -X POST -d 'hash=ghj1ghj2&hashpwd=asddasdd&username=stei&password=password&courseid=9999&coursevers=1337&did=40&moment=5010' http://localhost/LenaSYS/DuggaSys/showDuggaservice.php
```
Response:
```
{"debug":"[Superuser] Could not load dugga! no userAnswer entries with moment: 5010 \nline 338 showDuggaservice.php","param":"{}","answer":"UNK","danswer":"UNK","score":0,"highscoremode":"","grade":"UNK","submitted":"","marked":"","deadline":"UNK","release":"UNK","files":[],"userfeedback":"UNK","feedbackquestion":"UNK","variant":"UNK","ishashindb":false,"variantsize":"UNK","variantvalue":"UNK","password":"password","hashvariant":"UNK","isFileSubmitted":"UNK","isTeacher":1,"variants":[],"duggaTitle":"Inserttobedeleted","hash":null,"hashpwd":null,"opt":"UNK","link":"UNK"}
```

The debug field says that moment 5010 is not found, so i believe this part of the test is not working properly.
the `query-before-test-5` returns 5010 for me when doing each step manually. 
```
        'query-before-test-4' => "SELECT MAX(id) AS did FROM quiz",
        'query-before-test-5' => "SELECT MAX(lid) AS moment FROM listentries",
        'variables-query-before-test-6' => "did, moment",
        'query-before-test-6' => "INSERT INTO userAnswer (cid, quiz, variant, moment, useranswer, submitted, marked, vers, score, hash, password) VALUES(9999, ?, 3, ?, 'an answer', '2024-04-15 14:00:00', '2024-04-15 14:30:00', 1337, '0', 'ghj1ghj2', 'asddasdd');",
```

Note that this is not a problem with the specific implementation in this PR, but I think there is more to this issue than first described. This should probably be added to the current issue, OR, made into its own issue.

a22marca (changes requested):
- The test should not be in a separate file, but in showDuggaservice_test.php
- The expected-output field does not match the actual response when making a manual request to showDuggaService.php (see details in previous comment of the PR)

Issue #14704, opened 2024-04-18
Add '/LenaSYS/' to the path that is created at line 38 in validateHash.php


a22iliru:
Also remove any echo before header() function calls(redirects).

PR #14744, opened 2024-04-19
Deleted echoes before header() - issue#14702

a22melst:
I only deleted the echoes which came right before a header call.


a22marca (approved):
This issue is not possible to test properly before [issue#14704](https://github.com/HGustavs/LenaSYS/issues/14704) is resolved. This is because, the url in validateHash.php is not correct at the moment, so the page redirects are not done correctly. 

This code looks good however,  so I think we should merge this.

PR #14745, opened 2024-04-19
G2 2024 v3 issue#14638

a22marca:
Created the microservice, see issue comments for testing steps done.


b21isabe (approved):
I first tested by trying to change the linked service in dugga.js, this didn't work and I agree with the creator of the ms, that retrieveCodeViewerService_ms.php need to be created first to test integrating it into dugga.js.

Since I am having difficulties testing with curl commands, I instead used PowerShell Invoke-WebRequest to test the microservice. The microservice could successfully change sectionname and add and remove important words from the code example.

I couldn't find any other problems with the code.

PR #14746, opened 2024-04-19
the "copy content from" dropdown now displays existing course versions

a22ahmom:
It should be possible now to view existing course versions in the course version creator but also in choose course version dropdown. While solving the issue and trying to make it work a strange explode error occurred when you would navigate from version to version but now it's been fixed.


b22chran (changes requested):
Solves what the issue is intended to do. It puts the previous versions of the course in the dropdown of the "Copy content from". The code looks good as well.

b22chran (approved):
By accident I filled in for Request change instead of Approve. Changing to Approve.

Issue #14539, opened 2024-04-12
Improve delete functionality of courses


a22willi:
At the moment, course deletion is handled by selecting 'deleted' in the visibility section of the course edit modal. Change this to a button in the lower left corner.

**Requirements:**
- In the course edit modal, a red delete button should be visible in the lower left corner.
    - When clicked, a prompt should appear asking the user if they are sure. This prompt should include the options "Yes" and "No."
        - If "No" is clicked, the prompt should close, and the course edit modal should reappear, if possible. Alternatively, the 'Are you sure?' box could be positioned above the course edit modal.
    - If the user confirms the deletion of a course, it should not be deleted immediately. Instead, a 'toast' notification should appear for a few seconds, offering an option to undo the action. If the user presses 'Undo,' the deletion should not proceed. If the toast disappears without user action, the course should then be deleted.
- All the options should work with both dark and light mode, and fit the overall theme of the website. Try to reuse as many existing components as possible.

In this issue, you are not expected to perform the actual deletion, but you should implement the flow of the options and UI elements as described.

PR #14747, opened 2024-04-19

SleepyBearIV:


a22willi (changes requested):
There is something wrong with this PR, way too many files included, including files and commits which are not part of this issue. Also change it to merge into our weekly branch and not master.

Issue #14716, opened 2024-04-18
Make a working test for removeListentries in sectionedservice_Test.php


a22iliru:
make sure a correct expected output and filters are used
    currently this does not pass the assert test
    Be aware that some tests needs a fresh database to pass

sectionedservice_Test.php

b21isabe:
There seems to already be a test for this but it doesnt seem very good / doesn't seem to work at all. It uses way too many data points in the filtered expected output, the only data we are really interested in comparing is the listentries. The second thing is that the test first creates a list entry and then deletes it which means that the database/output will look the same before and after the test is run.

It would be better to pair this test with the createListEntry test like what was done in the duggaedservice-tests.php where one test first created something then another updated it and then the last one deleted it, this way you can see how the output is changed every step of the way.

b21isabe:
There appears to be 2 separate opts with very similar names but that do different things. There is removeListentries and deleteListentries. removeListentries sets a listentries visible attribute to 3 and the deleteListentries service deletes the listeEntry entirely. Since I have the remove one and not the delete one making a test should be pretty easy. I will first create a new listEntry before the test, then call the opt to change its visibility then get the output and compare it and then after the test delete it.

b21isabe:
There is a slight problem with making a test for this opt. At the moment no entries will show up as output, the reason for this is that there is an if statement that blocks entries from being stored. The if statement checks for many different things so I will have to investigate what it does, if its necessary and if there is another way to pass the entries to the output array.

b21isabe:
Fixed it by copying a working test and remaking it. Don’t know why it wasn’t working earlier.

PR #14748, opened 2024-04-19
G2 2024 v3 issue#14716

b21isabe:
issue: #14716 

The removeListEntries test should work now.

### DIFF
--- a/DuggaSys/tests/sectionedservice_Test.php
+++ b/DuggaSys/tests/sectionedservice_Test.php
@@ -127,6 +127,7 @@ $testsData = array(
     ),
     //-------------------------------------------------------------------------------------------------------------------
     // This test will test the micro-service removeListentries and its curresponding part in the monolith called "DELETED"
+    // THIS WILL FAIL THE FIRST TIME IT EXECUTES AFTER INSTALLING DATABASE!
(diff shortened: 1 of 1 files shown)


c22cargo (changes requested):
The test did not work for me the first time after reinstalling the database, then it worked fine. Ideally the comment should mention this. For example:

```
//-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
// This test will test the micro-service removeListentries and its curresponding part in the monolith called "DELETED". THIS WILL FAIL THE FIRST TIME IT EXECUTES AFTER INSTALLING DATABASE!
//-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

Also, the indentation is not correct, there are random blank lines. It would also be ideal to have it look something like this for improved readability:

```
'service-data' => serialize(
    // Data that service needs to execute function
    array( 
        'opt' => 'DELETED',
        'username' => 'brom',
        'password' => 'password',
        'lid' => '<!query-before-test-2!><*[0]["lid"]*>',
        'courseid' => '1',
        'coursevers' => '45656',
    )
),
```

Aside from those small nitpicks the test seems to work fine.

b21isabe:
The requested changes has been made

PR #14750, opened 2024-04-19
G2 Merge issue#14704 

a22marca:
Removed echo and change the url. It is now possible to load a dugga using a hash.
To get a example hash from the db, use:
```
select cid,hash,password from userAnswer;
```


c22cargo (approved):
Works

Issue #14707, opened 2024-04-18
fix: createListentrie_ms.php 


a22iliru:
- wrong paths to basic.php and session.php includes

PR #14760, opened 2024-04-19
G2 2024 v3 #14707

a21sahfa:


c22cargo (changes requested):
Already fixed in another issue.

PR #14762, opened 2024-04-19
fixed the pointer when dragging down

a22vioja:
I adjusted the deltaY (the change of vertical position) based on the kind of the element being manipulated. The kind attribute "elementData" determines the type of the element, such as 'UMLEntity', 'IEEntity', or 'SDEntity'. Since a UML entity consists of three parts and IE and SD entities consist of two parts, I divided the deltaY calculation by 3 or by 2 for the specific entity. 

This calculation adjusts the vertical movement speed for specific type of entity, allowing the pointer to move in parallel with the mouse.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -2585,8 +2585,14 @@ function mmoving(event) {
             deltaX = startX - event.clientX;
 
             const minHeight = 50; // Declare the minimal height of an object
-            deltaY = startY - event.clientY;
-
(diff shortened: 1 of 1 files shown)


a22hanfa (approved):
Tested elements: UML, IEE, & SD.

This works as it should now. The elements when dragged follow the mouse pointer at the same velocity. The mouse pointer is now always on top of the node being dragged, just like the ER ones.

Issue #14764, opened 2024-04-19
Make sure that reorderListentries test don't modify the data.


a22albjo:
reorderListentries test in sectionedservice_Test.php currently modifies the data making the test createListenties work the first time but not the 2nd time and onwards. 

For the tests to be consistent and able to run multiple times all data needs to be reset efter a test has been run, in this case the test reorders the listentries but don't put them back in the unordered state again afterwards.

Before the test has been running the items => pos is set from 1, 2, 3, 4, 5, 6. after the reordered test it is 0, 1, 2, 3, 4, 5.

To make the listentries get the right index from when first installing the db you can in the file LenaSYS/install/SQL/testdata.sql change the pos in the listentries with the kay vers set to: 45656 to be indexed(start counting) from 0.

Note that fixing this issue might result in problems with other tests. If that is the case you need to harvest new expected outputs or create new issues to harvest new expected outputs for those tests.

Questions? write to a22albjo in the discord-channel

Issue #14718, opened 2024-04-18
Filter output for reorderListentries test in sectionedservice_Test.php


a22iliru:
make sure a correct expected output and filters are used
    currently this does not pass the assert test
    Be aware that some tests needs a fresh database to pass

sectionedservice_Test.php

a22albjo:
The test seem to be working fine for me, im having a discussion with @a22melst on how to move forward.

The problem seem to be that its only working if testing course is included during install.

This issue is resolved, i will create a issue to make it more clear that the checkbox needs to be checked for servicetests to work.

a22albjo:
We have decided to filter the output on the tests in sectioned service to make them more readable.

PR #14770, opened 2024-04-19
G2 2024 v3 issue #14718 #14764

a22albjo:
This pullrequest solves issue #14718 #14764.

This pullreq makes createListentries test work and the output is filtered in reorderListentries test. It also makes sure thet the db entries that are tested in createListentries test is indexed correctly från 0 and not from 1.


a22marca (approved):
- The reoderListentries and createListentries test is able to pass and be run again and still pass. In fact all tests are passing!
- The updated inserts into the table listentries are correct.  I resintalled LenaSYS with a fresh database and checked the table manually after running the tests once and the values in the pos column where `vers=45656` are being indexed correctly. After the first and all succeeding times the test was run.

OK to merge

Issue #14685, opened 2024-04-17
Label misaligned when relation is completely vertical


a22jacka:
## Description
When a relation between 2 entities is completely vertical, the label on it flies away to the right. How far away it goes seems to depend on the elements position relative to the camera, bringing the elements up and right causes the label to go further and bringing the elements down and right brings it closer.

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/6990d9ae-87e3-465b-96dd-ab47ad232d6a)
![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/43e72e20-622f-4586-b1f6-6e748dff7826)

The elements also need to be closer than this for it to work
![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/f612c7c6-a1cc-4a10-9a5e-b77cae88982b)

## Job
Fix so that the label stays on the relation

PR #14772, opened 2024-04-19
G1 2024 v3 #14685

a22ricni:
Made it so the label should not be moving anymore. Can test it by have two entities close to each other and have a line between them. After set a label on the line and select on of the entity and move it by the arrow keys. It also depends on what zoom you have on the camera.


a22wilas (approved):
Old version:
bild

This branch:
bild

From testing I can confirm that the label bug happens when the entity elements are verticaly aligned and also depends on the campera position. If the snap to grind function is on, the label sometimes stays in place. 
On the new branch the label always stays in place, both if snap to grid is on or off. Tried this on multiple elements with different camera positions. I approve that the change has fixed the bug!

TLDR
bild
For some reason in this branch the icons have increased in size, this does not seem to have anything to do with the code changed in this issue. I changed the code back to it's original code from the issue but did not effect the size of the icons in the toolbar.

PR #14773, opened 2024-04-19
Removed  onchange="checkDropdownOption()" from the example diagram fi…

a22alfna:
…eldset


b22davbe (approved):
The call to the function which do not exist have been taken away and there's no longer an error when the dropdown-alternatives changes.

PR #14785, opened 2024-04-22
G1 2024 v3 #14701

a22solmi:
Broke out UMLEnity into function and updated resizing behaviour


a22cazlu (approved):
Tested the diagram and everything seems to work and the code have a good strukture

Issue #14526, opened 2024-04-12
Store relevant metadata when saving githubrepo


a22willi:
After [#14391](https://github.com/HGustavs/LenaSYS/issues/14391) it should be possible to download an entire gitrepo by selecting it for a course, and clicking fetch git repo. Make sure that the relevant data is saved when doing this. This is the data that should be saved.
![236859351-088624a4-bd36-4e12-8b5d-4635d922543e](https://github.com/HGustavs/LenaSYS/assets/129264968/9bb589a9-f7d4-4e21-88cf-abe6838da13c)
Taken from [wiki page](https://github.com/HGustavs/LenaSYS/wiki/Github-Integration) about github integration written by group 3 2023.

I noticed by reading the [pull request](https://github.com/HGustavs/LenaSYS/pull/14535) for issue #14391 that most of the db parts are probably saved already, just double check that everything regarding this is correct.

PR #14791, opened 2024-04-22
G3 2024 v3 #14526

a22danwe:
Made it so that the github repos files are saved to the githubmetadata2.db.

To be able to test this you must reinstall lenaSYS so the database is up to date. 

Add a github repo in editcourse 
image 

And also test it by adding it in fetch github.
image

we used DB Browser to look in the githubmetadata2.DB

@a22natth and @a22oscja assinged aswell.


d22mikwe (approved):
Everything works as intended when saving a github repo through the edit course option. Files are saved to gitFiles and the repo is saved to gitRepo.
Everything worked when fetching inside the course through fetch github aswell.

Code has proper indentation aswell.

LGTM!

Issue #14682, opened 2024-04-17
Create Toast functionality


a22willi:
Currently, the website is riddled with alerts, which are really annoying but sometimes useful. To fix this create a general Toast implementation. [What is a toast](https://www.w3schools.com/howto/howto_js_snackbar.asp).

IMPORTANT: If there is an existing toast implementation, simply comment that there is one, and give a short guide of where we can find it, and how we can use it. 

Requirements:
- The toast should work in both dark and light mode. 
- It should be simple to call the toast from anywhere in the project, only one file should have to be included for it to work. 
  - We recommend that you implement it somewhere within the Shared directory of LenaSYS.
- When creating the toast, you should be able to choose:
  - Duration.
  - Type: (warning, success, error)
  - Text.
- The toast should appear in the top right corner. 
- When calling multiple toasts, they should stagger downwards. 
- Every toast should have an `X` icon for immediately closing it. 
- After its timer has ran out, they should fade out.

PR #14794, opened 2024-04-22
G3 2024 v3 #14682 Toast implementation with responsive CSS this time

a22moamy:
A general toast implementation has been created. 

# How to use
### In your php file
To use the toast, simply include the `toast.php` file in your target php page. Your inclusion of toast.php should look something like this: 
`include "../Shared/toast.php";`
(OBS, the path might vary from file to file)
### In your javascript file
In your target javascript file, you can create a toast by calling the toast(type, text, duration) method where:
* **type** = the type of toast message (string)
type should be a string containing either "error", "warning" or "success". 
All other values will create a default toast without a type. 
* **text** = the text the toast method should contain (string)
* **duration** = the duration the toast message should appear for, expressed in seconds (integer)
If the duration is left null then it will become a default value of three seconds. 

For example: You can use the method like this:
`toast("success", "Hello world!", 6);`
Skärmavbild 2024-04-23 kl  10 25 30

You should also be able to stack toasts.
For example:
```
toast("success", "Hello world!", 6);
toast("warning", "You are not permitted to do that.", 12);
toast("error", "Operation failed! Something went horribly wrong...", 8);
toast("", "I'm not sure what type this toast should be.... Hmmm...", 10);
```

Result:
Skärmavbild 2024-04-23 kl  10 25 06

(These test screenshots were implemented in sectioned.php and sectioned.js)

Also, test that the toasts are responsive to different screen sizes.

The example usage will create a toast of the type "success" with the text "Hello world!". It will appear for six seconds before fading away. 

If there are multiple toasts, then they will stack downwards. Closed toasts will be deleted from the file completely.


a22danwe (approved):
works like a charm and the code looks good

Shared/toast.php
```diff
@@ -0,0 +1,104 @@
...
+
+<!-- The toast container containing all toasts -->
+<div id='toastContainer'>
+    <!-- The following commented out code shows the structure of a toast notication
+    <div class='toasterr'>
+        <span class="material-symbols-outlined" class='typeIcon'>warning</span>
+        <span class='typeText'></span>
+        <span class="material-symbols-outlined" class='closeIcon' onclick='closeToast()'>close</span>
+        <p class='toastText'></p>
+    </div>  -->
```
a22willi, line 102: Will this code ever be used, if not I think you can delete it.

Shared/toast.php
```diff
@@ -0,0 +1,104 @@
...
+                typeIcon.innerHTML = "error";
+                typeText.innerHTML = "Error";
+                break;
+            // We create a default situation in case no type is given.
+            // The default toast will not have an icon nor heading.
+            default:
+                typeIcon.innerHTML = "";
+                typeText.innerHTML = "";
+                break;
+        }
```
a22willi, line 74: Have the type based on [enums](https://www.geeksforgeeks.org/enums-in-javascript/) instead.

a22willi (changes requested):
In general, change it so that the warnings, errors and success toasts have different background colors. Success should be green, warning yellow, and error red.

a22willi:
If possible also change the shape of the toast to be more rectangular, like this:
image

DuggaSys/sectioned.php
```diff
@@ -1,7 +1,7 @@
 <?php
+	session_start();
 	include_once "../Shared/basic.php";
 	include_once "../Shared/sessions.php";
-	session_start();
```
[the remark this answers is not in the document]
a22moamy: Sadly yes, errors will be created when implementing toast otherwise.

Shared/toast.php
```diff
@@ -0,0 +1,114 @@
+<script>
+    const types = Object.freeze({ 
+        WARNING: "warning", 
+        ERROR: "error", 
+        SUCCESS: "success"
+    }); 
+    // Close the toast by clicking the X icon
+    function closeToast(toastDiv){
+        const toastContainer = document.getElementById('toastContainer');
+        if (toastDiv) {
```
a22danwe, line 10: tab this for correct indentation

Shared/css/style.css
a22danwe, line 1: be more consistent with "{" in the css. use space, ex ".toast:first-child {"

Shared/toast.php
```diff
@@ -0,0 +1,114 @@
...
+        toastRight.appendChild(closeIcon);
+         
+        // Add the toast div to the toast container (containing all toasts)
+        toastContainer.appendChild(toastDiv);
+       
+        // Changes the toast text to the input text
+        toastText.innerHTML = text;
+
+        // Decides the type of toast (changes the type icon and toast heading)
+        switch(type){
```
a22danwe, line 71: to be even more annoying the same goes for here

Shared/toast.php
```diff
@@ -0,0 +1,114 @@
...
+                typeIcon.innerHTML = "";
+                typeText.innerHTML = "";
+                break;
+        }
+        // Makes the toast visible ("show" is a class that makes the toast visible through css)
+        toastDiv.classList.add('show', 'toast');
+
+        // The duration of a toast decides how long it should be visible for
+        // If no duration is given, the default duration should be 3 seconds.
+        if(duration == "" || duration == null){
```
a22danwe, line 99: and here

a22danwe (changes requested):

PR #14795, opened 2024-04-22
retrieveUsername_ms.php issue#14765

a22eriwi:
Created the ms retrieveUsername that retrieves a username depending on the uid from the current session.
The Uid is retrieved from the ms getUid.


c22cargo (changes requested):
It works, but there are two blank lines at row 22-23 instead of just at row 22. Also, the closing ```?>``` tag should be omitted from files containing only PHP.

Issue #14756, opened 2024-04-19
accessedservice: Create microservice addClass


a22iliru:
Uses service insertIntoTableClass to makes inserts into the table class.

folder: accessedservice

PR #14796, opened 2024-04-22
Create microservice addClass issue #14756

c22cargo:
The microservice is supposed to be used when adding a new course, as can be seen from the ```accessedservice.txt``` text:
```* Add new course (opt = ADDCLASS)```. However, when creating a new course I was not able to see the ```accessedservice.php``` file being called at all. I couldn't find ```ADDCLASS``` used anywhere else either. Because of this I have not actually been able to  test the implementation.

If any tester finds where it's used you can test the implementation by changing the ```kind=="ACCESS"``` part in the ```AJAXService()``` function inside ```dugga.js``` to:

```
else if(kind=="ACCESS"){
	let serviceURL = "accessedservice.php";
	if (opt=="ADDCLASS"){
		serviceURL="../DuggaSys/microservices/accessedService/addClass_ms.php";
	}
	$.ajax({
		url: serviceURL,
		type: "POST",
		data: "opt="+opt+para,
		dataType: "json",
		success: returnedFile
	})
```

### DIFF
--- a/DuggaSys/microservices/accessedService/addClass_ms.php
+++ b/DuggaSys/microservices/accessedService/addClass_ms.php
@@ -1,0 +1,58 @@
+<?
+include_once "../../../Shared/sessions.php";
+include_once "../../../Shared/basic.php";
+include_once "../shared_microservices/getUid_ms.php";
+
+date_default_timezone_set("Europe/Stockholm");
+
+// Connect to database and start session
+pdoConnect();
+session_start();
+
+// Global variables
+$opt = getOP('opt');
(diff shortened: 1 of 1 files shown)


a22melst (changes requested):
I could not test it because the service is seemingly not used anywhere currently. According to [this](https://www.php.net/manual/en/language.basic-syntax.phptags.php), it is recommended to use a normal opening tag instead of a short tag for compatibility reasons. Other than that, the code looks good and I think it should be merged after this small change.

Issue #14664, opened 2024-04-17
Investigate why there is no error message for wrong login


a22willi:
In #14556 an error message was implemented for unsuccessful logins, this error message does not show up anymore after the merge. Investigate why, and fix it.

a20jonda:
![image](https://github.com/HGustavs/LenaSYS/assets/102598258/3ce73a87-580c-43b5-9280-319e99589c23)

Message is now showing after fetching to latest.

a20jonda:
This issue have been solved

b22marah:
Still doesn´t work for me or @a22willi, therefor we assign the issue to another group member, that this doesn´t work for to fully test and fix the issue.

PR #14799, opened 2024-04-22
G3 2024 v3 #14664 error message for wrong login is fixed

a22ahmom:
Before it wasn't possible to display an error message for wrong login because of Uncaught PDOException that caused a php error for uncaught SyntaxError for invalid JSON. But now when a user enters wrong credentials the error is displayed.

### DIFF
--- a/Shared/navheader.php
+++ b/Shared/navheader.php
@@ -691,6 +691,8 @@ const cooldownHolder = document.getElementById("cooldownHolder");
 setInterval(
 	function() 
 	{
+		//issue 14664, should be a typeerror down below: The TypeError is because of a innerHTML returns a string, not a number. We need to convert the innerHTML values to numbers, parseInt() function can be used which will convert a string innerHTML to an integer. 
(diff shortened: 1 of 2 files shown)


a22danwe (changes requested):
remove unecessary comments on line 217 in showDugga.php and on line 694 in navheader.php. 

other than that it works, good job!

a22danwe (approved):
perfecto

Shared/navheader.php
```diff
@@ -691,6 +691,7 @@ function hamburgerToggle() {
 setInterval(
 	function() 
 	{
+		
 		if(gitFetchCooldownSec.innerHTML>0 || gitFetchCooldownMin.innerHTML>0)
```
a22willi, line 695: This is not a change, remove this file from your PR.

a22willi (changes requested):

a22willi (approved):
Looks good

Issue #14775, opened 2024-04-19
Expanding doesn't check for overlap


a22jacka:
## Description
When resizing elements it doesn't check if it's overlapping so elements can be overlapped freely by resizing them.
This only needs to be done after the user finishes resizing (so in mup or similair). If that doesn't work properly then just check whenever the resize is done

## Job
Implement so that resizing also checks for overlapping.

PR #14800, opened 2024-04-22
G1-2024-v3-#14775

b22davbe:
Added global variables that saves the original properties of an entity before resizing while in "mdown()" and then lets the user resize freely when the program is in "mmoving()". and then when the resizing is done, the "mup()" reverts the entity to the original state if overlapping was detected and sends an error-message.


a22jacka (changes requested):
I'm guessing the master branch was used as a base for this branch, so a lot of extra changes came along.

This either needs to be rebased to the `g1-2024-v3` branch or a new branch with the weekly branch as a base needs to be made

Issue #14644, opened 2024-04-17
Create microservice createDuggaVariant


a22iliru:
Uses service __insertIntoTablVariant__ to makes _inserts_ into the table __Variant__.

`Note that the description may be wrong or missing steps, do as u think is right.`

a22eddpe:
### createDuggaVariant
Uses service __insertIntoTablVariant__ to makes _inserts_ into the table __Variant__.

a22eddpe:
cant find the relevant function in the dugga.js file.

a22eddpe:
this issue is not done yet because the servise insertIntoTableSettings is missing. View the pull request for the code that was created before insertIntoTableSettings and see what can be used.

PR #14801, opened 2024-04-22
g2-2024-v3-#14644

a22eddpe:
this issue is not done yet because the servise nsertIntoTableVariant  is missing. View the pull request for the code that was created before insertIntoTableSettings and see what can be used.

### DIFF
--- a/DuggaSys/microservices/showDuggaService/createDuggaVariant_ms.php
+++ b/DuggaSys/microservices/showDuggaService/createDuggaVariant_ms.php
@@ -9,5 +9,44 @@ include_once "../../../Shared/basic.php";
 pdoConnect();
 session_start();
 
+//Get the necessary parameters from the request
+$qid = getOP('qid');
+$uid = getOP('uid');
+$disabled = getOP('disabled');
+$param = getOP('param');
+$variantanswer = getOP('variananswer');
 
+
+// - quizID
+// - creator
+// - disabled
+// - param 
+// - variantanswer
+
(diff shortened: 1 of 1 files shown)


DuggaSys/microservices/showDuggaService/createDuggaVariant_ms.php
```diff
@@ -0,0 +1,58 @@
+<?php
+
+date_default_timezone_set("Europe/Stockholm");
+
+include_once "../../../Shared/sessions.php";
+include_once "../../../Shared/basic.php";
+include "../shared_microservices/getUid_ms.php";
+getUid();
```
a22marca, line 8: This function call can be removed.

DuggaSys/microservices/showDuggaService/createDuggaVariant_ms.php
```diff
@@ -0,0 +1,58 @@
...
+
+// Connect to database and start session
+pdoConnect();
+session_start();
+
+function getUid(){
```
a22marca, line 14: The function is already included no need to reimplement it.

a22marca (changes requested):
See review comments of the code.

It would also be good to add the part that checks the users privileges from duggaedservice.php (line 54 in that file).

It's not possible to completely test the microservice since this branch does not seem to include a few fixes that addresses the problem of entering the "show tests" page. You need to get the updates from the weekly group branch to be able to test this properly.

Issue #14769, opened 2024-04-19
Close toolbar selection on keyboard-selection


a22jacka:
## Description
The check to close the toolbar selection (see image) is done in a mouse button event, meaning it stays open if the keyboard is pressed instead

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/57112d0c-b9bb-487e-b08b-deccb501e1e8)

## Job
Call the function `hidePlacementType()` from keydown/keyup if another element has been chosen with keyboard

PR #14802, opened 2024-04-22
G1 2024 v3 #14769

a22ricni:
Can now press a keybinding and the sub-menu is closing when open. Can test it by have a sub-menu open and press for example "1" and the sub-menu should be closing.


d22hanol (approved):
Tested to see that the problem existed before this branch, switched to this branch  and the problem was fixed. 
I approve the merging of this branch

PR #14805, opened 2024-04-22
Program now detects overlapping while resizing objects

b22davbe:
Added global variables that saves the original properties of an entity before resizing while in "mdown()" and then lets the user resize freely when the program is in "mmoving()". and then when the resizing is done, the "mup()" reverts the entity to the original state if overlapping was detected and sends an error-message.


b22chrmo:
Solution works properly. Resizing an element so that it covers another element is not possible anymore. Overlapping an element by resizing causes it to revert back to its original size.

PR #14806, opened 2024-04-22
Fixed ER relation text position and double lines ( + added auto weak relations when creating )

a22vioja:
To fix the position of the text inside ERRelation, I changed the SVG element text positions to percentages.
To fix the ER example diagram I added  state property and set its value to "weak" in JSON file.

Screenshot 2024-04-22 at 14 34 37

I also added additional implementation for weak relations, so that if the line is created between strong and weak relation, the line will be doubled automatically and the ERRelation would change to “weak” also. 

Screenshot 2024-04-22 at 14 48 38


a22solmi:
This is branched from master

PR #14809, opened 2024-04-22
deleted some conditions

a22cazlu:
deleted some conditions in a if-state: context[0].kind != elementTypesNames.ERRelation.


a22vioja (approved):
It works with minimum size. 

Screenshot 2024-04-23 at 13 57 29

Issue #14776, opened 2024-04-22
Can't click on text in ER "Kinds" selection


a22jacka:
## Description
When selecting the kind of ER-relation you can't click on the text to select an option, instead you can only click inside the circle

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/ff31d8e7-3bb2-4c88-80a2-0685dc7ac7b0)

## Job
Make it so that an option can be selected by clicking on the text

PR #14810, opened 2024-04-22
G1 2024 V3 #14776

a22ricni:
Can now click on the text when you want to change what kind of line it should be. Test it by have a line between elements and click instead on the text when change for example to double line under the "Kinds" header.


b22jonan (approved):
switched to branch, created two EREntities, added an ERAttr between them, double clicked it, tried clicking on all the different text options, clicking the text or the circle selected that option, this works

PR #14815, opened 2024-04-22
basic test file

b21youab:
Basic test file that covers core functionalities .


a22fokol:
while doing the test, I noticed that there was an error on lines 3, 4 and 32.
Lines 3 and 4 show that (include) on both of these lines are  incorrect and the one on line 32 is a function that is undefined.

By the way for lines 3 and 4 I believe there must be an error in the location of the coursesseedservice_test.php file.
This is a test file which is located in a Microservice file, which means that the test file must instead be moved to the test file. and then we will have to resolve "function" problem of line 32.

a22solmi (changes requested):
> while doing the test, I noticed that there was an error on lines 3, 4 and 32. Lines 3 and 4 show that (include) on both of these lines are incorrect and the one on line 32 is a function that is undefined.
> 
> By the way for lines 3 and 4 I believe there must be an error in the location of the coursesseedservice_test.php file. This is a test file which is located in a Microservice file, which means that the test file must instead be moved to the test file. and then we will have to resolve "function" problem of line 32.

c22cargo (changes requested):
I just get warnings when I open the file, it does not work at all.

I can see from the first line that:
```
include "C:\Users\Youssuf\OneDrive\Documents\GitHub\LenaSYS\DuggaSys\microservices\courseedService\courseseedService_test.php";
```
is not correct, that is an absolute path, not a relative path. And it's also specific to your own computer. It seems to be including itself for some reason.

Also, this test file should be placed in DuggaSys/tests.

PR #14818, opened 2024-04-22
removed unused variable 

a22vioja:


a21liltr (approved):
Line 1856 with "var testCount = 0;" has been successfully removed.
Double checked and there is no use of testCount found anywhere else in the code either. Nothing in the code seem to be affected by the change.

PR #14820, opened 2024-04-22
Fixed #14816

### LINKED ISSUE #14816: break out SDEntity into function
## Description

drawElement() is a 900 line long function. There is a lot of repetition and if else cases that are hard to read and understand.
Break out SDEntity into a function.

## Job
Break up this function into smaller pieces to increase readability

d22hanol:
The code for SDElements has been made into a function and localized


d22abdha (commented):
The test has been verified, and the latest update confirms that the implementation now supports proportional resizing in both dimensions. This improvement allows elements to be resized with equal flexibility in width and height while maintaining the required aspect ratio.

PR #14821, opened 2024-04-22
Changed IEEntity to behave like UMLEntity

a22solmi:
Changed IEEntity to behave like UMLEntity

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1310,7 +1310,7 @@ var defaults = {
         stroke: color.BLACK,
         fill: color.WHITE,
         width: 200,
-        height: 50,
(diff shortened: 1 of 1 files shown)


a22vioja (approved):
I have tested to resize UML and IE entities and they behaves the same way (header remains fixed and other parts scales with text). Other functionalities seems to work as expected. 
Screenshot 2024-04-23 at 09 51 02

I tested also SD entity because there was code changes for this entity as well. It behaves as before. 
Screenshot 2024-04-23 at 09 50 30

Refactored version achieves the same functionality as the previous code but with clearer and more maintainable code structure.

 MAX height/width should be added for entities (could be as a new issue?)

PR #14822, opened 2024-04-23

SleepyBearIV:

### DIFF
--- a/DuggaSys/courseed.php
+++ b/DuggaSys/courseed.php
@@ -140,17 +140,13 @@ if(isset($_SESSION['uid'])){
 				</div>
     		</div>
 			<div style='float:right; padding-top:20px; width: 464px;' >
-				
-    			<input id='saveCourse' class='submit-button' type='button' value='Save' title='Save changes' onclick="validateForm('editCourse')" />
 				<button id="deleteCourseButton" onclick='openForm()'>Delete</button>
+    			<input id='saveCourse' class='submit-button' type='button' value='Save' title='Save changes' onclick="validateForm('editCourse')" />
 					<div class="form-popup" id="myForm">
 					<form action="" class="form-container DarkModeBackgrounds">
 						<h1 Class="DarkModeText" style="font-size:24px; padding-bottom:50px; padding-top:60px; border-bottom: 2px solid #ccc;">Are you sure? </h1>
-
-
-					
-						<button id="deleteCourseButtonYes" type="button" onclick="closeForm()">Yes</button>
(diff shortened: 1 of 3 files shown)


SleepyBearIV:
we added a new button the old one had not been clanged and we could not use the same, make a new issue for next weeks merge to fix a hover effect if  the feeling of one is necessary. indentation of the code: both buttons is in the container and should not have different indentation, but form should however be indented differently, that is now fixed

b22chran (changes requested):
The indentation of the buttons are fine, but the <h1> should be aligned with the buttons though it is elements inside the form. 
Also the change of the cursor to a pointer isn't implemented for the hover-effect for the delete button and the yes and no buttons in the dialog.

b22chran (approved):
Hover effects looks good and indentation looks fine as well!

Shared/css/style.css
```diff
@@ -7737,6 +7736,82 @@ textarea#mrkdwntxt {
...
+  padding: 5px;
+  position: fixed;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
+  border: 3px solid #f1f1f1;
+  z-index: 9;
+  background-color: #fff; /* white background */
+  height:260px;
+  width:460px;
```
a22willi, line 7751: Consider using dynamic sizing, this could potentially look strange on devices with different screen sizes.

Shared/css/style.css
a22willi, line 1: - Check if there is an existing css class for popup forms that you can use. 
- Improve consistency in formatting, mixed use of spaces and no spaces between property and value.
- Some properties can be combined into one. [example](https://www.w3schools.com/cssref/pr_padding.php).

a22willi (changes requested):

DuggaSys/courseed.php
```diff
@@ -139,8 +139,16 @@
...
+					<div class="form-popup" id="myForm">
+						<form action="" class="form-container DarkModeBackgrounds">
+							<h1 Class="DarkModeText" style="font-size:24px; padding-bottom:50px; padding-top:60px; border-bottom: 2px solid #ccc;">Are you sure? </h1>
+							<button id="deleteCourseButtonYes" type="button" onclick="closeForm()">Yes</button>
+							<button id="deleteCourseButtonNo" type="button" onclick="closeForm()">No</button>
+						</form>
+					</div>
     		</div>
       </div>
 	</div>
```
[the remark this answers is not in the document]
SleepyBearIV: dose not look like that in the code, the button and the input is on the same line

PR #14824, opened 2024-04-23
Fixed ER relation text position and double lines ( + added auto weak relations when creating )

a22vioja:
To fix the position of the text inside ERRelation, I changed the SVG element text positions to percentages.
To fix the ER example diagram I added  state property and set its value to "weak" in JSON file.

Screenshot 2024-04-22 at 14 34 37

I also added additional implementation for weak relations, so that if the line is created between strong and weak relation, the line will be doubled automatically and the relation would change to “weak” also. 

Screenshot 2024-04-22 at 14 48 38


a22hanfa (changes requested):
Unfortunately it looks like the text is still off center when it's a weak relation:
**Weak relation**
Screenshot 2024-04-23 at 09 47 49

**Normal relation**
Screenshot 2024-04-23 at 09 48 04

The `DuggaSys/JSON/ERDiagramMockup.json`-problem is **fixed**, "has" is now a weak relation.
Screenshot 2024-04-23 at 10 02 21

**Automatically makes an error:**
In response to:

> I also added additional implementation for weak relations, so that if the line is created between strong and weak relation, the line will be doubled automatically and the relation would change to “weak” also.

There is a weak relation and a normal entity.
Screenshot 2024-04-23 at 10 05 11

This will turn it in to a weak entity, this shouldn't happen unless the user wants it to be weak.
Screenshot 2024-04-23 at 10 05 24

a22vioja:
> This will turn it in to a weak entity, this shouldn't happen unless the user wants it to be weak.

I fixed it so now it will change to double line only if the relation is coming from "weak" state to "weak" state. Otherwise, the elements need to be changed manually.

Screenshot 2024-04-23 at 11 13 04


Screenshot 2024-04-23 at 11 14 19

PR #14829, opened 2024-04-23
UML states broken out #14823

a22solmi:
Broke out all UML states into functions, updated text placedment in superstate


a23ponek (commented):
Things seem to not have been affected and working as intended. This includes the creation of UML diagram with its different properties like inheritance, relations etc. Also the example UML diagram works aswell. No issues in console and other functions in the diagram.php seems to work as normal.

PR #14834, opened 2024-04-23
G2 2024 v3 #14686v3 - update names to camelCase

a22albjo:
Update names and paths to files to take the new names in to account.


a22fokol:
During testing I noticed that "accessedService_ms.php" and "Addservice_ms.php" both do not exist in the "microservices" file when browsing. But there is indeed "accessedService_test.php" in the "tests" file.

a22albjo:
> During testing I noticed that "accessedService_ms.php" and "Addservice_ms.php" both do not exist in the "microservices" file when browsing. But there is indeed "accessedService_test.php" in the "tests" file.

@a22fokol I don't understand, are "accessedService_ms.php" and "Addservice_ms.php" implemented microsevices that are missing?

accessedService_test.php exists because it is the name of a monorepo accessedService.php, the namingconvention we use for testfiles are to name them after monorepos.

Issue #14803, opened 2024-04-22
Identify why It's not possible to remove a listed entry in a course


a22iliru:
By a22marca: It's not possible to remove a listed entry in a course, when trying to remove it no requests are sent, so i'm not sure whats wrong in this case. I belive a listed entry should be set to visibility 3 and then removed after wards, i'm not sure?

c22cargo:
Issue can be reproduced by going into "show tests" for a course and trying to delete a test.

c22cargo:
It seems that the "useranswer" table has references to these quizes, which is why they cannot be deleted.

PR #14841, opened 2024-04-23
g2-2024-v3 issue #14803

c22cargo:
The error was occurring because the ```useranswer``` table has a foreign key constraint to ```quiz```. The solution I ended up with was to delete all instances of ```useranswer``` with a reference to the quiz to be deleted.

  This is only a good idea if useranswers are SOLELY answers in quizes, which I assume. However, I am not 100% sure this is the case. A tester should ideally come to the same conclusion as me. What the tables are supposed to do can be found in ```dbTables.md```. 

The two rows look like:
```
### quiz
The quiz table contains everything related to a dugga, and connects the dugga to a specific .js template file. It does not save answers, it only loads the dugga how it is supposed to look like.
```

```
### userAnswer
Stores all the duggas that has been saved in the system. That is, the answers that a student has entered is stored here.
```

### DIFF
--- a/DuggaSys/duggaedservice.php
+++ b/DuggaSys/duggaedservice.php
@@ -93,20 +93,19 @@ if(checklogin() && (hasAccess($userid, $cid, 'w') || isSuperUser($userid) || has
 			$debug.="Error updating dugga ".$error[2];
 		}
 	}else if(strcmp($opt,"DELDU")===0){
-		$query = $pdo->prepare("DELETE FROM quiz WHERE id=:qid");
+		$query = $pdo->prepare("DELETE FROM useranswer WHERE quiz=:qid;");
 		$query->bindParam(':qid', $qid);
-		try{
-			if(!$query->execute()) {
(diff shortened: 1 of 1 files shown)


a22eddpe:
My conclusion: The deletion of useranswer records was necessary because of a foreign key constraint linking the useranswer table to the quiz table. This constraint caused errors when attempting to delete a quiz, as the database system prevented it. To resolve this, @c22cargo chose to delete all instances of useranswer records associated with the quiz to be deleted. However, this solution is only advisable if useranswers are exclusively answers within quizzes.

DuggaSys/duggaedservice.php
a22eddpe, line 1: My conclusion: The deletion of useranswer records was necessary because of a foreign key constraint linking the useranswer table to the quiz table. This constraint caused errors when attempting to delete a quiz, as the database system prevented it. To resolve this, @c22cargo chose to delete all instances of useranswer records associated with the quiz to be deleted. However, this solution is only advisable if useranswers are exclusively answers within quizzes.

Issue #14753, opened 2024-04-19
profileservice: Create microservice changeProfileValues


a22iliru:
changeProfileValues

(writter comment: i think this service is small enough as is )

Uses service selectFromTableUser to get information it requires from user. Uses service selectFromTableUser_course to get information it requires from user_course.

Statements below are methods for changeProfileValues and not services.
updateSecurityQuestion

Uses service selectFromTableUser to get information it requires from user.
updatePassword

Uses service updateUserPassword to get information it requires from user.

folder: profileservice

f22linhe:
This function is already implemented in updateUserPassword. See issue #14757.

PR #14848, opened 2024-04-23
G2 2024 v3 #14753

a21sahfa:
needs to be tested


DuggaSys/profileServices/changeProfileValues.php
```diff
@@ -0,0 +1,150 @@
+<?php
+
+date_default_timezone_set("Europe/Stockholm");
+
+include_once "../Shared/basic.php";
+include_once "../Shared/sessions.php";
+include_once "database.php"; // Include database connection file
```
a22albjo, line 7: database.php is already included in sessions.php and can be removed.

DuggaSys/profileServices/profileservice.php
a22albjo, line 1: This file profilesevice.php needs to be put back to DuggaSys folder and the profileService folder with the microservice changeProfileValues_ms.php inside it.

DuggaSys/profileServices/changeProfileValues.php
a22albjo, line 1: rename this to changeProfileValues_ms.php to continue with the established naming convention for microservices.
The folder profileServices should be in microservices folder. DuggaSys/profileServices/changeProfileValues

DuggaSys/profileServices/changeProfileValues.php
```diff
@@ -0,0 +1,150 @@
+<?php
+
+date_default_timezone_set("Europe/Stockholm");
+
+include_once "../Shared/basic.php";
+include_once "../Shared/sessions.php";
```
a22albjo, line 6: The includes have the wrong paths to the files. This will look inside the DuggaSys after a folder called Shared.

DuggaSys/profileServices/changeProfileValues.php
```diff
@@ -0,0 +1,150 @@
...
+        }
+    } else {
+        $error = $stmt->errorInfo();
+        $isSuperUser = "Error checking super user status: " . $error[2];
+    }
+
+    return $isSuperUser;
+}
+
+function logServiceEvent($log_uuid, $event_type, $file, $userid, $info) {
```
a22albjo, line 67: There is a function in basics.php that is named logServiceEvent, that looks a bit like this, is this needed or are you suppose to use the existing function that you imported from basics.php?

DuggaSys/profileServices/changeProfileValues.php
```diff
@@ -0,0 +1,150 @@
...
+    pdoConnect();
+    $querystring = "INSERT INTO log_table (log_uuid, event_type, file, userid, info) VALUES (:log_uuid, :event_type, :file, :userid, :info)";
```
a22albjo, line 69: I cant find the table: log_table in the LenaSys database.

a22albjo (changes requested):
Well done with this separation of functionality and nice structure to the code.

I saw that you wrote:

> i have tried to test it and i believe it can be tested in the profile settings but for some reason i get this message when i click on the profile "Profile settings can only be changed on a secure HTTPS connection.

in a commit, if you are using XAMPP you can just change the url by prefix it with https:// and profile.php will work.

I think you can start by addressing the issues i have written in this review, then test the functions and then resubmitting it for a review again. I have not tested the individual functions, i think this is easier done after this review has been adressed.

PR #14852, opened 2024-04-23
G2 2024 v3 issue#14758 

b21isabe:
This service will add an existing user to a course, if the user doesn’t exist it will create a new one.

The monolith service was not working at all before, so I had to make several fixes/changes to it. Before, users were created with SSNs and className, but since these are not used in the current version of LenaSYS, I had to do some imperfect workarounds. The service still has many problems, I didn’t want to completely redesign it so I only made changes to make it functional with how the website is set up right now.

To test it, enter any course, click the padlock icon at the top of the screen. Then move your cursor to the bottom right + sign and click “create user”, fill in the fields and click “add”. If it worked a new user should have been added to the “user” table in the database and they should also be added to the current course in the “user_course” table. If the user already existed, that user should be added with the currently selected course in the user_course table.

To test the microservice you can edit from line 1093 in dugga.js to this:

```
	}else if(kind=="ACCESS"){
		if(opt === "ADDUSR"){
			$.ajax({
				url: "../DuggaSys/microservices/accessedService/addUser_ms.php",
				type: "POST",
				data: "opt="+opt+para,
				dataType: "json",
				success: returnedAccess
			});
		}
		else{
			$.ajax({
				url: "accessedservice.php",
				type: "POST",
				data: "opt="+opt+para,
				dataType: "json",
				success: returnedAccess
			});
		}
```


c22cargo (approved):
Works.

There are some things that could be changed, for example the ```?>``` tag should be omitted and:
```
include ('../shared_microservices/getUid_ms.php');
```
should be
```
include_once "../shared_microservices/getUid_ms.php";
```

but these are small changes so I will approve the PR.

Issue #14836, opened 2024-04-23
Fix microservice createNewCodeexample 


a22iliru:
USED BY

    createListentrie
    createGithubCodeexample

Uses the services insertIntoTableCodeexample to insert into the table codeexample.

    cid
    exampleid
    sectionname
    uid = 1 (static value)
    cversion

This file exists, but does not do what it is supposed to do.

PR #14853, opened 2024-04-23
g2-2024-v3 issue #14836

c22cargo:
MS was tested with the command:

```
curl -X POST -d "courseid=1&examplename=TESTEXAMPLE&sectname=TESTSECT&coursevers=123" http://localhost/LenaSYS/DuggaSys/microservices/shared_microservices/createNewCodeExample_ms.php
```

### DIFF
--- a/DuggaSys/microservices/shared_microservices/createNewCodeExample_ms.php
+++ b/DuggaSys/microservices/shared_microservices/createNewCodeExample_ms.php
@@ -1,13 +1,6 @@
 <?php
-/**
- * Handles creation of new items on a course, not just for code examples (as the name would suggest)
- * I presume that the title of the file is for consistency with the legacy system... Hey, I just work here....
- */
-
 date_default_timezone_set("Europe/Stockholm");
 
-// Include basic application services!
-include_once "../../../../coursesyspw.php";
 include_once "../../../Shared/sessions.php";
 include_once "../../../Shared/basic.php";
 include_once "getUid_ms.php";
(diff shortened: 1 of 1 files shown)


DuggaSys/microservices/shared_microservices/createNewCodeExample_ms.php
```diff
@@ -18,85 +11,24 @@
...
-			$debug = "Error updating entries" . $error[2];
-		}
-
-		$link = $pdo->lastInsertId();
-	}
-
-	$query = $pdo->prepare("INSERT INTO listentries (cid,vers, entryname, link, kind, pos, visible,creator,comments, gradesystem, highscoremode, groupKind) 
-												 VALUES(:cid,:cvs,:entryname,:link,:kind,:pos,:visible,:usrid,:comment, :gradesys, :highscoremode, :groupkind)");
+$coursevers = getOP('coursevers');
+$userid = getUid();
```
a22albjo, line 18: I think this is fine, but i notice that the global variables probably would already be harvested in the file that includes this file. This will serve as additional safety.

a22albjo (approved):
I have tried to create a new codeexample with cull as explained by @c22cargo.

I have also tried to include it in row 244 in the sectionedService monorepo and it worked fine used in practice inside a service as well. Well done! 

This can be merged.

a22albjo:
I don't know why the old code was added to the review comments and not your new. I hope my feedback in clear anyway.

Issue #14669, opened 2024-04-17
GitFetch cooldown causes errors


a22willi:
The following error appears once per second after clicking the github fetch button for a course and then navigating to another site. 
![image](https://github.com/HGustavs/LenaSYS/assets/129264968/759aa9fb-0062-41c6-9381-a6e7b5dc9b45)

a22samka:
Located error in the navheader.php file with the function setInterval:
![image](https://github.com/HGustavs/LenaSYS/assets/129301680/ea09af99-a263-48a8-8517-f1792e63e03e)

I will try to fix the error..

PR #14855, opened 2024-04-23
g3-2024-v3-#14669

a22samka:


a22willi:
There are too many files included in this PR which are not part of the issue.

a22willi (changes requested):
There are too many files included in this pull request which are not part of the parent issue.

Issue #14832, opened 2024-04-23
Remove unused code


a22solmi:
## Description
Following #14474 #14627

## Job
Remove unused and under utilized global variables

PR #14858, opened 2024-04-23
G1 2024 v3 #14832

a22solmi:
bug fixes and removed unused global variables


a23ponek (commented):
No immediate problems seem to occur after the changes. Overall functionality seems to work without any problems. Tested functionality like create example diagrams, save/load, toolbar functions and grid/dark mode, and everything works. Also no errors in console.

Issue #14654, opened 2024-04-17
Lifelines only copies lines


a22jacka:
## Description
When copying a lifeline is only copies the line and not the character or label

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/2cf429e8-f3bc-4e1c-8903-089e7954a648)

## Job
Fix so that the label and character are also copied

a22solmi:
solved by #14859

PR #14859, opened 2024-04-23
Separated the lifeline actor/object into two different elements. (#14654)

b22chrmo:
Copying a lifeline actor/object was not possible before due to the way they were drawn inside the div. 

I've separated the actor/object into two different elements, making it possible to copy the lifelines. 

I have not changed the tooltip for either of the two elements, and the new type of element has a placeholder icon. (These will have to be fixed in a new issue, due to no time before merge).


a22solmi (approved):
Could copy and paste without any problem. Toolbar and cycling also works with new entity.

Issue #14740, opened 2024-04-19
Icons scaling past the toolbar


a22jacka:
## Description
At certain sizes it's possible for the icons in the toolbar to expand past the toolbar and appear over the ruler instead

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/b2aab2ed-ef1f-4c1f-af3d-a7cecd2074c5)

## Job
Either limit the sizes with css or js, or dynimacally expand the toolbar with js. 
**NOTE:** if the choice is made to change the toolbar width, then you need to look through the code since `setRulerPosition()` and maybe other functions use the toolbars 51px width as a hardcoded value

a22solmi:
solved by #14860

PR #14860, opened 2024-04-23
G1 2024 v3 #14740

a22ricni:
Made it so the ruler on the left and the toolbar to the left looks better when the screen height it higher then 1200 pixels. Can test it by have a bigger screen and see if it looks like when it is for example 1920x1080 screen.


a22solmi (approved):
Scales well when resizing window. Issue solved

PR #14881, opened 2024-04-24
g2-2024-v4 issue#14395

b21isabe:
The setVisibleListentriePUBLIC in sectionedService_Test.php should now work.

Important:
Some tests require a fresh DB install to work.


c22cargo (changes requested):
The test works, but there are a few changes that could be made.

Firstly, where is the ```cid``` "1885" coming from? Is it a course made only for testing? If it is you can ignore this, but otherwise I would probably make a new test course first, and then delete it after the test.

There's also a few inserts into ```listentries``` which look to be unnecessary. These are ```link, kind, pos```. 

I also don't think these are necessary for the filter:
```
'debug',
'writeaccess',
'readaccess',
'coursevers',
'courseid',
````

c22cargo (changes requested):
The changes look good, but you are still using an existing course for the test. What happens if that course is removed? Instead, insert a new course into the database to use, and then delete it after the test.

b21isabe:
Now creates a new course and deleting it after test.

Issue #14886, opened 2024-04-24
microservice shared_microservices/retrieveUsername_ms.php should be a function


a22iliru:
The code inside the file should be remade into a function

PR #14890, opened 2024-04-24
g2-2024-v4 issue #14886

c22cargo:
Functionality was remade into a function.

I only updated the files that already used the old MS, but there are several more that should be updated to use the ```retrieveUsername``` MS.


c22emigr:
Implementation of function looks good.

Issue #14828, opened 2024-04-23
Make both sides of UML start and end states scale


a22jacka:
## Description
UML start and end states use the minimum of the with and the height to set it's size, this can cause the states to have a lot of unused space which is a problem since the unused space still get used in the overlap check.

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/3882b39e-23d5-4790-9743-3027b4873e06)

## Job
Make so that both sides scale at the same time. If you expand the height is should also expand the width

PR #14893, opened 2024-04-24
Synced height and width expansion for issue#14828

a23ponek:
Fixed an half-done solution for the height and width expansion so they expand together "kinda", to get a perfect circle one must fix the expansion to fit but its possible. The room that the box takes up are also decreased by the size of the circle as requested.

However, some weird problems showed up. The changes creates a small circle like a blob on top of the init state circle which doesnt expand and stays idle. Also the uncommented code at line 9672 seem to affect the circle if its removed, even if its "outcommented" which is very odd. Also the final state circle appears exactly like the init state circle for some reason. I havent been able to locate where these anomalies come from, and the diagram.css and style.css seem to have nothing affecting this. 

Link to issue: 
https://github.com/HGustavs/LenaSYS/issues/14828

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -9658,17 +9658,18 @@ function drawElementState(element, ghosted, vectorGraphic) {
     } else if (element.fill == color.WHITE && theme.href.includes('style')) {
         element.fill = color.BLACK;
     }
-    return `<div id="${element.id}" 
+    return `<div id="${element.id}"
                 class="element uml-state"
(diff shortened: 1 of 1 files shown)


a22jacka (changes requested):
You need to scale the parent div for the circle so that the whole circle is always visible

You also need to fix the issues with the white circle in the corner and final states appearing like initial states

a22jacka (approved):
Those immideate problems seem to have been fixed

Ready for proper testing

c22abdha (dismissed):
resize up, resize left and the entire object moves. please change so the object doesnt move before resize

d22abdha (approved):
the elements can now be resized and move in different directions on the screeen

PR #14895, opened 2024-04-24
included suggestions

a22eddpe:


a22melst (changes requested):
**To fix:**

- This microservice is based on the opt “ADDVARI” in duggaedservice, and should therefore be in the folder duggaedService instead of showDuggaservice.

- You need to include sessions.php and basic.php, since some functions needed such as hasAccess and checklogin is defined in those files.

- Include the microservice getUid instead of implementing it again. Then replace getOP(‘uid’) with a call to the function getUid.

- Put all the code which executes the query inside of the if statement that checks permissions.

- getOP(‘param’) should be getOP(‘parameter’).

To test the microservice, switch out the code at kind "DUGGA" in dugga.js to the following:

```
}else if(kind=="DUGGA"){
		let serviceURL = "duggaedservice.php";
		if (opt=="ADDVARI"){
			serviceURL="../DuggaSys/microservices/duggaedService/createDuggaVariant_ms.php";
		}
		$.ajax({
			url: serviceURL,
			type: "POST",
			data: "opt="+opt+para,
			dataType: "json",
			success: returnedDugga
		})
```

a22eddpe:
@c22cargo, denna issuen har blivit lite kosntig då den redan är mergead (även fast den inte skulle de) så den gamla versionen finns just nu i mastern. Marcus har redan kollat på denna (då han fick som test att testa den gamla verisonen) och gav feedback på den och bad mig ta bort includes du nu sa att jag skulle implementera. Detta skev han: 
image

Issue #14636, opened 2024-04-17
Create microservice updateFileLink


a22iliru:
Uses the services __updateTableFileLink__ to change the content of these columns:
- filesize
There exist three different versions of this update, with different _WHERE_ cases.
__WHERE__ kind __AND__ filename;
__WHERE__ cid __AND__ kind __AND__ filename;
__WHERE__ vers __AND__ cid __AND__ kind __AND__ filename;

`Note that the description may be wrong or missing steps, do as u think is right.`

a21sahfa:
issuet kan inte lösas eftersom orden den nämner inte finns.

a22melst:
The microservice is misspelled as "updataFileLink" in [potential microservices](https://github.com/HGustavs/LenaSYS/blob/d663ef4fd4f5d138017e7ce456bed6dce1350e15/backend-models/microservices/duggaSys%20services/Potential_micro_services.mdl).

a22melst:
Trying to fix the logging so this can be tested before the weekly merge.

a22melst:
Will create a new branch for this issue after the weekly merge.

a22melst:
When using getOP('fid') from the MS, the returned value is "UNK". The file ID is needed when fetching the path of a file, which in turn is used for setting the $currcwd variable for course local files.

a22melst:
Everthing works now except for logging.

a22melst:
Will make a new branch to be able to use the retrieveUsername microservice.

PR #14899, opened 2024-04-24
Created updateFileLink_ms.php

a22melst:
To be able to test the microservice, change the code at kind "FILE" in dugga.js to the following.
```
}else if(kind=="FILE"){
		if (opt=="SAVEFILE"){
			serviceURL="../DuggaSys/microservices/fileedService/updateFileLink_ms.php";
		}
		$.ajax({
			url: serviceURL,
			type: "POST",
			dataType: "json",
			success: returnedFile
		})
```
		
### Testing
Check all this for global, course-local and version-local files.

- Check that the updating of the filesize in table "fileLink" is performed.
- Check that logging to loglena6 database is performed. 
- Check that the file is successfully written to in C:\xampp\htdocs\LenaSYS\courses.

Note: I could not get the logging to work, although I am not sure why because it looks correct to me. Maybe it will work for someone else, but if not then it could be made into a separate issue. The retrieval of the username is to be done using a microservice which at the time of writing has not been merged yet.

Check the [issue](https://github.com/HGustavs/LenaSYS/issues/14636) for further notes.

### DIFF
--- a/DuggaSys/microservices/fileedService/updateFileLink_ms.php
+++ b/DuggaSys/microservices/fileedService/updateFileLink_ms.php
@@ -1,5 +1,11 @@
 <?php
 
+//---------------------------------------------------------------------------------------------------------------
+// updateFileLink_ms
+//---------------------------------------------------------------------------------------------------------------
+// Microservice that handles writing to files and updates filesize in fileLink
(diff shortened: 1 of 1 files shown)


c22cargo (approved):
Looks good

Issue #14648, opened 2024-04-17
Create reusable "loader-button"


a22willi:
Create a button, that can replace some of the existing buttons, that incorporates a loader that makes it clear the website is still working and has not crashed. 

Requirements:
- When clicked the button should become disabled, and the loader should appear inside it. (innterHtml)
- The button should preferably be class based, maybe using [builder pattern](https://www.geeksforgeeks.org/builder-design-pattern/) so that when used, appropriate settings can be made to accomodate the specific functionality that is needed.
- There should be a function to make it stop loading that can be called when something has completed.

a22maxra:
The loading button can be implemented in several places such as when saving and getting data from github. A good way could be creating a JS class that can be reused.

a22maxra:
We are trying to find a button with DOMContentLoaded and then attach an eventlistener, which we did, but the code that is supposed to run when the attached eventlistener is triggered is not running...

a22maxra:
We will try a different approach tommorow

a22danwe:
It seems like the code doesn't run on the "demo course" and courses we create ourselfs but works on the already existing courses

a22maxra:
Will do it a different way then we tried, instead of adding a class that adds functionality to buttons/inputs, we will try to replace the buttons with added functionality

a22maxra:
I have been trying to create a class that functions with the requirements but it is hard!!!! ! !!!

a22maxra:
A button container is now created instead of the old button
``` 
<div id="buttonContainerSaveRepo"></div> 
```

And the arguments are added as an action in the ´loadingButton.js´ file
```
const actions = {
  saveRepo: {
    action: validateForm, // Function
    argument: "githubPopupWindow", // Argument to pass to the function
    buttonLabel: "Save Repo", // Button label
    containerId: "buttonContainerSaveRepo" // Container ID
  },
```

Which generates a button with loading functionality 
![image](https://github.com/HGustavs/LenaSYS/assets/98206230/bcd24479-d7ce-4ba7-996b-b74b95e38a0e)

This button stops loading when an action is completed, some functions reload the page after completion as well so it won't show, but when pressing a button without an argument the loader will show for 0.25 seconds.

There is a problem with this code which i have not been able to resolve, newly created courses for some reason do not work with the code, the onclick functions seem to now activate even though the button is generated.

a22maxra:
[...]

PR #14900, opened 2024-04-24
Loader for buttons G3 2024 v4 #14648

a22maxra:
Buttons now have loading capabilities when waiting for a request to be fulfilled, replace old buttons/inputs
Old:
```
<input id='saveRepo' class='submit-button' style='float:right; margin-right:5px;' type='button' value='Save' title='Save Changes' onclick='validateForm("githubPopupWindow");' />
```
New, make a container and include `loadingButton.js`
``` js
<script src="../Shared/loadingButton.js"></script>

<div id="buttonContainerSaveRepo"></div>

```

Within `loaderButton.js` fill in the function to be called, the argument, label and containerID in the _actions_ object
```js
    const actions = {
      saveRepo: {
        action: validateForm, // Function
        argument: "githubPopupWindow", // Argument to pass to the function
        buttonLabel: "Save Repo", // Button label
        containerId: "buttonContainerSaveRepo" // Container ID
      },
```

When adding more buttons, create a new container, make sure `loaderButton.js` is included in the php file and add the correct arguments to the _actions_ object.

For new courses you need to fill in the information and give it a version for the button to work, i don't know why this is, the `onClick` should work either way from my understanding but this should not be a problem since you would fill these fields out before getting data from Github in a real world scenario.

This solution allows buttons to be changed to the loader version incrementally and if something is not working the old versions of the inputs/buttons can be used in the mean time. If the solution is satisfactory more functionality can be added and more buttons can be switched out, right now the button to edit information about a course and the get Github data are loading buttons. 

image

### DIFF
--- a/DuggaSys/courseed.php
+++ b/DuggaSys/courseed.php
@@ -141,7 +141,7 @@ if(isset($_SESSION['uid'])){
 				</div>
     		</div>
     		<div style='float:right; padding-top:20px; width: 464px;' >
-    			<input id='saveCourse' class='submit-button' type='button' value='Save' title='Save changes' onclick="validateForm('editCourse')" />
+				<div id="buttonContainerSaveCourse"></div>
     		</div>
       </div>
 	</div>
--- a/DuggaSys/courseed.php
+++ b/DuggaSys/courseed.php
@@ -33,6 +33,7 @@ if(isset($_SESSION['uid'])){
 	<link id="themeBlack" type="text/css" href="../Shared/css/blackTheme.css" rel="stylesheet">
 
 	<script src="darkmodeToggle.js"></script>
+	<script src="../Shared/loadingButton.js"></script>
 	<script src="../Shared/js/jquery-1.11.0.min.js"></script>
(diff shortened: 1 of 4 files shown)


DuggaSys/courseed.php
```diff
@@ -141,7 +142,7 @@
 				</div>
     		</div>
     		<div style='float:right; padding-top:20px; width: 464px;' >
-    			<input id='saveCourse' class='submit-button' type='button' value='Save' title='Save changes' onclick="validateForm('editCourse')" />
+				<div id="buttonContainerSaveCourse"></div>
     		</div>
```
a22willi, line 146: Indentation.

a22willi (changes requested):
Could you make the spinner slightly smaller, and have the button appear disabled when clicked loading? As detailed in the original issue.

DuggaSys/courseed.php
a22maxra: Looks good in VSCode

a22willi (approved):
Looks good.

a22willi (approved):
Looks good.

PR #14901, opened 2024-04-24
added function retrieveUsername to microservice

c22emigr:
Added function to retrieveUsername microservice. Added function so that other microservices can include our microservice and get the username by using our function.

### DIFF
--- a/DuggaSys/microservices/sharedMicroservices/retrieveUsername_ms.php
+++ b/DuggaSys/microservices/sharedMicroservices/retrieveUsername_ms.php
@@ -13,19 +13,18 @@ include_once "getUid_ms.php";
 pdoConnect();
 session_start();
 
-$userid = getUid();
+function retrieveUsername($pdo)
+{
+    $userid = getUid();
+    // Gets username based on uid
+    $query = $pdo->prepare("SELECT username FROM user WHERE uid = :uid");
+    $query->bindParam(':uid', $userid);
+    $query->execute();
 
-// Gets username based on uid
-$query = $pdo->prepare("SELECT username FROM user WHERE uid = :uid");
-$query->bindParam(':uid', $userid);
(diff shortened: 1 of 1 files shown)


DuggaSys/microservices/sharedMicroservices/retrieveUsername_ms.php
```diff
@@ -13,19 +13,18 @@
 pdoConnect();
```
a22marca, line 13: The microservice works without calling this function. This is because the PDO that is passed as a parameter. I think we can presume that the PDO is already previously initalized by the service calling this function.

DuggaSys/microservices/sharedMicroservices/retrieveUsername_ms.php
```diff
@@ -13,19 +13,18 @@
...
 session_start();
```
a22marca, line 14: I'm not sure this is needed either, like `pdoConnect()`. Most likely the calling service will have performed this function call already.

a22marca (changes requested):
The function works as intended. Tested by adding the function call to a already implemented microservices, and integrating the microservices. The username is retrieved without any issues.

 I don't think it's necessary to make the function calls, like specified in the review comments of the code. Furthermore, "getUid_ms.php" is probably the only needed include here. The reason for this is that the calling microservices who will use this function probably already included these themselves. 

But this looks good overall, everything works as intended.

a22eriwi:
There was an missunderstanding in the group and this issue had already been solver by https://github.com/HGustavs/LenaSYS/pull/14890
So i'm closing this now.

a22eriwi:
There was an missunderstanding in the group and this issue had already been solver by https://github.com/HGustavs/LenaSYS/pull/14890
So i'm closing this now.

PR #14904, opened 2024-04-24
G2 merge issue#14714

a22marca:
Added new expected-output, removed unnecessary filters and added needed parameters.


c22cargo (changes requested):
I don't think the ```INSERT INTO listentries``` and ```DELETE FROM listentries``` is needed at all, but if it is here are some points. It should use the testing-course, not course 2. I also don't think it needs to use ```link, kind, pos, visible, comments, gradesystem, highscoremode, groupKind```

The ```DELETE FROM``` should ideally not delete everything greater than a certain value. Instead you could do something like:
```
DELETE FROM listentries WHERE cid = 1885 and entryname = "<SOMETHING SPECIFIC>"
````

I would assume that instead of doing an insert into ```listentries``` it should make inserts with regards to users and groups.

For the filter I don't think ```coursename, coursevers, courseid, grpmembership``` is needed.

Also, instead of leaving the old code as comments just remove it instead to reduce cluttering.

**Comment by a22marca** [comment]
You are correct, the test can be performed without the queries. They where left there since I was trying to performing test with test data during the time I was working on the issue. The attributes where there for me to easily identify the row when manually looking in the database.

**Comment by a22marca** [comment]
I have added these queries and a new expected-output according to these values:
INSERT INTO user_course (cid, verse, uid,groups) VALUES (1885,1337,3, "No_2"), (1885,1337,101, "No_2");
INSERT INTO listentries (cid,vers, entryname, kind, visible,creator, groupKind)  VALUES(1885,1337,"test group",6,1,101, "No");
DELETE user_course WHERE uid=3;
DELETE user_course WHERE uid=101;

**Comment by a22marca** [comment]
The code is now updated, and i've left the filters that I think are relevant for this test.

**Review by c22cargo — CHANGES_REQUESTED** [review]
The test now seems to be testing what it should, but there are still a few changes that could be made.

I don't think these are necessary at all. I tried removing them and the test still works as expected.
```
'query-before-test-2' => "INSERT INTO listentries (cid,vers, entryname, kind, visible,creator, groupKind)  VALUES(1885,1337,'test group',6,1,101, 'No');",
```

```
'query-after-test-3' => "DELETE FROM listentries WHERE cid=1885 AND vers=1337 AND entryname='test group';",
```

Regarding the ```DELETE``` querys, I think they should be changed. Instead of deleting everything from ```user_course``` that a specific user is a part of, you could instead do something like:

```
'query-after-test-1' => "DELETE FROM user_course WHERE uid=3 and cid = 1885;"
```

Perhaps even more specific.

a22marca:
Yes, the tests work when not creating a new row in listentries.  Group members are retrieved from all courses specified in user_course. I guess we can remove this part as well, but this (and maybe more) will need to be added again if group members only for the specific course should be retrieved.

- The uid's used in the queries are not part of any course originally, thats why they are removed from all courses.

c22cargo (approved):
Looks good

Issue #14729, opened 2024-04-18
Make it clear that the Include Testing-Course checkbox needs to be checked for servicetests to work during installation.


a22albjo:
Something like this but shorter and better:
![image](https://github.com/HGustavs/LenaSYS/assets/53318057/99ebef5f-8181-47ca-b8fc-a0d44e51f107)

f22linhe:
Group 3 has made a new interface for the installer. This issued should be closed.

PR #14905, opened 2024-04-24
Update install.php #14729

a22fokol:
the testing checking box has been updated, and needs to be checked to work during installation


c20juler:
Reviewed and tested. Checkbox for "Include Testing-Course" is now checked as expected during install.

PR #14907, opened 2024-04-24
Create logUserEvent_ms.php

c20juler:
created microservice, implemented to be used through a function call.


b21isabe (approved):
Tested by commenting out the original logUserEvent function in basic.php and instead including the microservice.

The microservice successfully inserted new rows in the /log/lenasys6.db file with the expected data. One issue with the logs is that the timestamps are set 2 hours too early. This is not a problem with the micro/mono service, because it's the sqlite db not the service that sets the timestamps. 

Approve.

PR #14908, opened 2024-04-24
G2 2024 v4 issue#14500

b21isabe:
The Get active users test in showDuggaservice_test.php should work now.

Important:
Some tests require a fresh DB install to work.


a22marca (approved):
The change in showDuggaservice, makes it possible to retrieve the active users value from showDuggaservice.php. The test compares the hash and the value of active users without any issues, the rest is filtered.

The test is working, OK to merge.

Issue #14872, opened 2024-04-24
Remove unused options


a22jacka:
## Description
Certain elements have options that aren't valid in their notations and only cause bugs.
List is not complete and may be expanded during development
## Job
Options and elements to remove: 
* Recursive option for ER-line
![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/2cdf2323-47cb-4bbf-82c2-d31fef5daca1)

* Label for IE-line
![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/9a33292d-6e7b-496a-a51c-3fb9d615cb4d)

* Cardinalities for state entities
![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/14aac1dd-d4a1-41d4-a056-273ab7c257f6)

PR #14909, opened 2024-04-24
G1 2024 v4 #14872

a22ricni:
Now should the "Recursive" option be gone for ER diagram, label section for IE diagram and cardinalities for state diagram. Can test it by see if all the option still works and if the option that I removed is gone.


b22davbe (approved):
The buttons have been removed and the options work.

PR #14912, opened 2024-04-24
G2 2024 v3 #13852

b21youab:
Fixed issues with the first lines of include and reworked the testfile by adding new functions testing core functionalities of the courseedservice.php


a22marca:
There are changes in sectionedservice_Test.php. This file should not be changed.

a22marca (changes requested):
There are changes in sectionedservice_Test.php. This file should not be changed in this issue.
The expected-outputs don't look like the outputs of courseedservice.php. Also what parts (opts/functionalities) do these tests address?

a22albjo:
@b21youab I think you should close this pr and if needen make a new one. This pr is built on a branch from week 3.

Issue #14877, opened 2024-04-24
Key listener for ´Enter´ ignores bad data and submits a form which breaks the section in sectioned.php 


a20jonda:
![image](https://github.com/HGustavs/LenaSYS/assets/102598258/62496981-551e-4c28-b765-e9f820771cf5)
`Edit Section`: Note that the `Save` button has been disabled but pressing enter saves the form anyways

![image](https://github.com/HGustavs/LenaSYS/assets/102598258/b2ba39db-bb7d-45eb-aff7-aaff9bcfbd48)
The new section is no longer editable due to the error.

PR #14917, opened 2024-04-24
G3 2024 v4 #14877 Pressing enter with invalid information

a22moamy:
Added an if-statement that checks whether all information in the form is correct before allowing the user to update an item by pressing enter. To check if this fix works, try to update an item (for example a dugga) with an illegal character in a text-input. An example of an illegal character is **´**. For example, try pressing enter when the form is looking something like this:
Skärmavbild 2024-04-24 kl  16 23 44

It should not allow you to submit the form and update the item. Also test that no new errors have been created because of this fix.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -2566,27 +2566,28 @@ $(window).keyup(function (event) {
     document.activeElement.blur(); // To lose focus from the newItem button when pressing escape
   } else if (event.keyCode == 13) {
     // Remember that keycode 13 = enter button
-    console.log("This is from sectioned.js");
     document.activeElement.blur();
     var saveButtonDisplay = ($('#saveBtn').css('display'));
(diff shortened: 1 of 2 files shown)


a18chrkr (commented):
The if-statement seem to work very good. I've tested with several different characters including ? ] ) ¨ * ; etc and it did not let me save by clicking enter. The only character that seemed to be accepted was _ (underscore). Not sure if this is intended or not.

One thing i noticed was that when you correctly change the name of a dugga by pressing "enter" the update toast do not show.

a22moamy:
I have now hopeully fixed the alert/toast issue :) ! Underscore does not seem to be an illegal character as it does not crash anything unlike the other illegal characters. Saving a dugga with underscores in the name works good, so I think this is intended!

a18chrkr (approved):
Nice! The toast seem to be fixed! Great work! 🎉

PR #14918, opened 2024-04-24

SleepyBearIV:

note to tester: the DuggaSys/courseed.php code looks weired on the review page but shude look good in the code editor


b22chran (approved):
The hover looks good, the code has the standard indentation what i can see in VS-code and the functions works as far as intended.

Issue #14880, opened 2024-04-24
Change order of ER attribute variant


a22jacka:
## Description
When selecting what variant an ER attribute should be, the order is very arbitrary and should be changed

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/7ee056e7-a33a-4214-a7ee-a943f8a7d925)

## Job
Change the order so that more common choices are close to the top

PR #14920, opened 2024-04-24
G1 2024 V4 #14880

a22ricni:
Have change the order of the attribute states for the ER diagram. Test it by see if the order is "correct" and the most common attribute is more in the top of the drop-down menu.


a22vioja (approved):
I have tested to see the arrangement of ER attributes and it is correct, everything is working fine and the code changes do not affect anything else in the program.

Screenshot 2024-04-25 at 09 59 50

PR #14923, opened 2024-04-24
Support for legacy database in resulted.php

a22oscja:
This changes a few things regarding queries to the database to support the old version of the lenaSYS database. mostly adding a few try/catch and isset statements.

**testing instructions:**

- Get the old database version (ask William or Marcus for this) and then run that sql file towards the database.
- Go into resultedservice.php and add "true ||" to the beginning of the if-statement at row 36 (this has to be done due to the user we are logged in to no longer existing, hence this is done to make sure we have rights to see the content of the page)
- At this point, load the page to see if the purple bar at the bottom (this is the table) appears, i also suggest checking the content sent by resultedservice.php is being sent (use the network section of the inspect tools in the browser for this).
- Make sure to check that both the old and new database versions work. To get your old database back, simply run the installer.

**If you wish to further test to the point of actually seeing content in the table:**

- add a row to user, fill out the required fields.
- add a row to listentries, it must have all required fields filled out, as well as "kind" being 3 and "vers" corresponding to a course.
- lastly make a row in useranswer, link the "cid" and "moment" that was created earlier and also make sure that "vers" matches

Note that these things must use an existing course as a basis, so do not just make up vers, cid and such.

Now if that all those extra steps of making table rows sound a little too annoying (because they are), here is a picture of it working on my pc: 
Screenshot 2024-04-24 170744

### DIFF
--- a/DuggaSys/resultedservice.php
+++ b/DuggaSys/resultedservice.php
@@ -33,12 +33,21 @@ if(checklogin())
 	} 
 }
 
-if(isSuperUser($userid) || hasAccess($userid, $cid, 'w')){
+if(true || isSuperUser($userid) || hasAccess($userid, $cid, 'w')){
 	// Get data to display in table rows
-	$query = $pdo->prepare("SELECT hash, password, submitted, timesSubmitted, timesAccessed, moment,last_Time_techer_visited FROM userAnswer WHERE cid=:cid AND vers=:vers");
+	try{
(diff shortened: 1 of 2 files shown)


SleepyBearIV:
cant run/import the database on my system (running Docker), tried it on 3 different docker systems and the same error occurs 
image

thinking that xampp runs it as it ass it is imported while the docker wants to implement it and then run, causing the problem of "bytes exceeds the limit" an person using xampp to see if it works on there system

a22karja (commented):
Dosent get any errors and the page loads. But can not get results to be written out. this probobly from inserting the infromation wrong. The code works with both databases. the original issue is solved in the sense the page loads and dosent write out any errors. (exept for getFetchCooldown but that comes from another issue)

PR #14925, opened 2024-04-24
Implement test file for diagramservice.php #13853

a21sahfa:
följde efter api guiden, har dock inte testat den.


a22marca (changes requested):
These tests don't really make sense since diagramservice.php don't create tests, yet both tests are structured like so and named "create course test". 

We need to first identify what separate functionalities diagramservice.php offer, and what tests need to be created. So i'm not completely sure this issue should be about implementing the actual tests. If we identify several functionalities offered by diagramservice.php and need to create a test for each, they could be individual issues.

PR #14937, opened 2024-04-25
G2 2024 v4 issue#14926

a22marca:


c22cargo (changes requested):
I have seen several other microservices use more permissions checks. Looking at ```duggaedservice.php``` the following line is used:
```
if(checklogin() && (hasAccess($userid, $cid, 'w') || isSuperUser($userid) || hasAccess($userid, $cid, 'st')))
```

I am not completely sure if we should use these or not, but since several other MS use it I will put this here. Perhaps a group discussion is needed about these?

If not needed you can see this as approved.

a22marca:
These queries are now added:
```
        'query-before-test-1' => "INSERT INTO user_course (cid, vers, uid,groups,access) VALUES (1885,1337,3, 'No_2','W'), (1885,1337,101, 'No_2','W');",
        'query-before-test-2' => "INSERT INTO listentries (cid,vers, entryname, kind, visible,creator, groupKind)  VALUES(1885,1337,'test group',6,1,101, 'No');",
        'query-after-test-1' => "DELETE FROM user_course WHERE uid=3;",
        'query-after-test-2' => "DELETE FROM user_course WHERE uid=101;",
        'query-after-test-3' => "DELETE FROM listentries WHERE cid=1885 AND vers=1337 AND entryname='test group';",

```

a22marca:
I have added the permission check and the return value now.

c22cargo (approved):
This looks to be correct, but for the future you don't have to nest the if-statements.

Instead of
```
if(checklogin() && (hasAccess($userid, $cid, 'w') || isSuperUser($userid) || hasAccess($userid, $cid, 'st')))
```

you could do:

```
if(!(checklogin() && (hasAccess($userid, $cid, 'w') || isSuperUser($userid) || hasAccess($userid, $cid, 'st'))))
```

and then return. This improves readability a lot.

Also, the closing ```?>``` tag should be omitted from files containing only PHP.

Issue #14903, opened 2024-04-24
Break out sequence into function


a22solmi:
## Description
drawElement() is a 900 line long function. There is a lot of repetition and if else cases that are hard to read and understand.
Break out sequence object, actor and activation into functions.
Part of #14891

## Job
Break up this function into smaller pieces to increase readability

PR #14939, opened 2024-04-25
G1 2024 v4 #14903

a22solmi:
Moved all sequence drawing into own functions


a22jacka (approved):
Tested placing ever and resizing everything and it works the same as before

Issue #14906, opened 2024-04-24
Change path of multiple files


a22iliru:
![image](https://github.com/HGustavs/LenaSYS/assets/164884946/56f49e4b-a7ec-4e12-b18f-bc61bb0f7390)

PR #14948, opened 2024-04-25
g2-2024-v4 issue #14906

c22cargo:
Updated includes


b21isabe (approved):
The 5 files now links to the correct folder, I also checked if there were any other files linking to the old folder but didn't find any. 

Approve

Issue #14503, opened 2024-04-11
Create tests: get data from userAnswer (student) and get data from userAnswer (teacher) in showDuggaservice_test.php


a22albjo:
Create test: get data from userAnswer (student) and get data from userAnswer (teacher) in showDuggaservice_test.php

There are commented out version of these tests more or less finished in showDuggaservice_test.php, to resolve this issue gather expected output from the service and make this test work as intended.

PR #14950, opened 2024-04-25
getdataUserAnswer_test.php  #14503

a22fokol:
the test of get data from userAnswer (student) and get data from userAnswer (teacher) has been created


a22melst (changes requested):
The test should be inside of the file showDuggaservice_test.php. You can uncomment the test which exists in that file with CTRL + K + U and run the file in a browser to see what looks wrong, or if it passes when it shouldn't and so on. This test doesn't work because it has no expected output to compare the actual service output to.

PR #14961, opened 2024-04-25
Created microservice updateCourseversion_sectioned.php

a19gabos:
Added functionality to update MOTD, version name, start date, and end date for a course version.


a22fokol:
updateCourseversion_sectioned_ms.php has been created

c22cargo (changes requested):
There was already an almost complete version you could have built upon instead of making a new one. Also, the new one is located in the wrong folder. It doesn't even resemble the original version. How did you test this and where is it used?

I would say scrap this new version and just finish the original one.

There's currently 2 files trying to be merged instead of just 1.

PR #14969, opened 2024-04-25
G2 2024 v4 issue#14715 Filter output for test deleteListEntries

a22albjo:
I have filtered the output for the test deleteListEntries in 
/LenaSYS/DuggaSys/tests/sectionedservice_Test.php
I think it is sufficient to have the listentries together with the name of the course as output in this case.


c22cargo (changes requested):
Since the ```DEL``` OPT also deletes from ```useranswer``` I think one should also be inserted instead of just a single instance of ```listentries```.

Regarding the ```INSERT INTO listentries``` it should probably not use a fixed ```lid``` since it's auto-incremented. I also don't think it needs to use ```link, kind, pos, visible, comments, gradesystem, highscoremode, groupKind```. It should also use the test-course instead of course 1.

Even though the test should delete the entries itself I still think the ```DELETE FROM``` is necessary incase the test fails. However, the ```DELETE FROM``` should ideally not delete everything greater than a certain value. Instead you could do something like:
```
DELETE FROM listentries WHERE cid = 1885 and entryname = "<SOMETHING SPECIFIC>"
```

For the service-data I don't think ```coursename, comment, hash``` is needed.

Regarding the filter I don't think ```lid, writeaccess, studentteacher, courseid, visible, pos``` is needed. Since lid is auto-incremented I would use a very specific entryname instead.

a22albjo:
Nice feedback, to implement the test with quizAnswer will make the test better. There will be multiple inserts into atleast **listentries**, **quiz** and **userAnswer**. I will start working on this.

a22albjo:
> Since the DEL OPT also deletes from useranswer I think one should also be inserted instead of just a single instance of listentries.

_This is now implemented_


> Regarding the INSERT INTO listentries it should probably not use a fixed lid since it's auto-incremented. I also don't think it needs to use link, kind, pos, visible, comments, gradesystem, highscoremode, groupKind. It should also use the test-course instead of course 1.

_When using an autoincremented lid the result of the lid seemed unpredictable and went from 5011 to 100009 and 100010. This happened even tho i had deleted the last inserted entry. I agree that using the autoincrement woult be ideal and this should be revisited sometime in the future to look further in to how that could be implemented. A fixed lid makes it possible to test the service multiple times without a reinstall. The rest is fixed._


> Even though the test should delete the entries itself I still think the DELETE FROM is necessary incase the test fails. However, the DELETE FROM should ideally not delete everything greater than a certain value. Instead you could do something like:
> ```
> DELETE FROM listentries WHERE cid = 1885 and entryname = "<SOMETHING SPECIFIC>"
> For the service-data I don't think coursename, comment, hash is needed.
> ```

_I have specified the listentrie to delete using the lid._


> Regarding the filter I don't think lid, writeaccess, studentteacher, courseid, visible, pos is needed. Since lid is auto-incremented I would use a very specific entryname instead.

_This is done_

c22cargo (changes requested):
The reason you are getting
```
Warning: Undefined array key "did" in C:\xampp\htdocs\LenaSYS\Shared\test.php on line 82
```
is because the value is not actually retrieved. It needs to be sent in the service-data. You currently send a fixed ```lid``` instead, and no ```did```.  This means that the ```quiz``` part of an ```userAnswer``` is set to null. Make sure a fixed ```lid``` (5011) is not sent. Instead, send the one retrieved from the query.

image

I don't actually think a ```quiz``` needs to be inserted at all though. Maybe you had some issues with constraints or something? Looking at the database the ```quiz``` row in a ```userAnswer``` seems to be allowed to be NULL.

The ```WHERE``` part of these should not be necessary since the IDs should all be unique and the tests will not be performed concurrently or on a live database.
```
'query-before-test-3' => "SELECT MAX(id) AS did FROM quiz WHERE cid = 1885",
'query-before-test-4' => "SELECT MAX(lid) AS lid FROM listentries WHERE cid = 1885",
```

The deletes need to be more specific. Currently if the test succeeds, the ```userAnswer``` will be deleted (from the monolith file), then afterwards another ```userAnswer``` not related to the test is deleted again by this query.
```
'query-after-test-1' => "DELETE FROM userAnswer ORDER BY aid DESC LIMIT 1;",
'query-after-test-2' => "DELETE FROM listentries WHERE lid = 5011;",
'query-after-test-3' => "DELETE FROM quiz WHERE cid = 1885 ORDER BY id DESC LIMIT 1;",
```

Also, take a look at my previous comment regarding the filter-output and the ```listentries``` insert. It seems you didn't change anything there. The ```userAnswer``` insert can probably also be trimmed down, but I didn't look further into it.

You can probably remove all instances of ```quiz```. I would assume a ```userAnswer``` would require a ```quiz```, but apparently not.

c22cargo (changes requested):
The username needs to be changed from ```mestr``` to ```brom``` (or someone else with permission). This is because the ```DEL``` OPT doesn't get executed without certain permissions. This means that the test wasn't being executed before. It also means that the expected output is wrong. It shouldn't contain ```{"entryname":"Inserttobedeleted"}``` since that should be deleted from the test.

Also, update the ```DELETE``` querys to:
```
'query-after-test-1' => "DELETE FROM userAnswer WHERE cid = 1885 AND useranswer = 'an answer to be deleted';",
'query-after-test-2' => "DELETE FROM listentries WHERE cid = 1885 AND entryname = 'Inserttobedeleted';",
```

Other than that it looks good.

c22cargo (approved):
Looks good

PR #14971, opened 2024-04-25
G2 2024 v4 issue#14931

a22marca:
Create the microservice isSuperUser_ms, that implements a function that is similar to a function in sessions.php (same name). This function requires a pdo to be sent as a parameter. See issue comments for testing steps.


a21sahfa (approved):
den funkar

Issue #14944, opened 2024-04-25
Activations can't overlap conditions properly


a22jacka:
## Description
Sequence activations can only be placed over conditions if they're placed really far down, but they need to be able to placed all over it.

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/bb143cd8-7a9e-4198-be27-55b264132c2e)

## Job
Make it so that activations can be placed over the entire condition element

PR #14972, opened 2024-04-25
G1 2024 V4 #14944

a22ricni:
Should fix it with the sequence activation can be on top of sequence condition without it overlapping. Can test it by have a sequence condition on screen and try if sequence activation it still overlapping or not.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -3813,7 +3813,7 @@ function entityIsOverlapping(id, x, y) {
                     isOverlapping = false;
                 }
                 //if its overlapping with a sequence actor, just break since that is allowed.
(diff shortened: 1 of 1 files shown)


a21liltr (approved):
Tested by placing a 'Sequence Condition' on the diagram, followed by placing 2 'Sequence Lifeline' elements and 2 'Lifeline (object)' elements on top of the 'Sequence Condition'. 

One of each element is then resized (elongated from the bottom by dragging to expand further down). The 'Sequence Activation' element is then selected and attempted to place on top of both the resized and unchanged elements.

The 'Sequence Activation' element has also been tested by placing a numerous amount of the element all over the 'Sequence Condition' without issue.

Conclusion:
Works well but the code line is very long...

PR #14974, opened 2024-04-25
Fixed the minimum height and width for superstates

a22alfna:
I fixed so that not only superstates but also the other elements have a minWidth and minHeight variable storing their specific minimum height and width so that you can call on these variables in one switch case instead of creating multiple different if statements one for each element. And I adjusted the minimum heights and widths to be realistic for each element.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1267,7 +1267,9 @@ var defaults = {
         state: 'normal',
         attributes: ['-attribute'],
         functions: ['+function'],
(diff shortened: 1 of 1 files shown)


a22cazlu (approved):
The issue seems to be solved. But maybe we should take a look on why u can get like stuck when u resizing the element.

a22solmi:
> The issue seems to be solved. But maybe we should take a look on why u can get like stuck when u resizing the element.

The same bug is present on the current branch.

PR #14976, opened 2024-04-25
Implemented the file updateUserPassword_ms.php issue#14757

a22eriwi:
In order to test this ms you need to have an https connection and you need to implement the following url:"microservices/sharedMicroservices/updateUserPassword_ms.php"  into the ajax call on line 166 in the file profile.js

image

### DIFF
--- a/DuggaSys/microservices/sharedMicroservices/updateUserPassword_ms.php
+++ b/DuggaSys/microservices/sharedMicroservices/updateUserPassword_ms.php
@@ -1,7 +1,7 @@
 <?php
 
 //---------------------------------------------------------------------------------------------------------------
-// profileService - handles password changes and challenge question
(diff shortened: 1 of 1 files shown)


c22cargo (changes requested):
The microservices works, but some changes should be made.

You forgot to change the top comment from ```profileservice.php```.

This should be replaced with an MS call.
```
if(isset($_SESSION['uid'])){
	$userid=$_SESSION['uid'];
}else{
	$userid="1";		
} 
```

I could not get the logging to work correctly. This may just be an issue with my computer. If it works for you then it's fine.

a22eriwi:
@c22cargo Fixed the change requests

c22emigr:
Fixed requested changes and tested that the ms works. Ready for merge

PR #14977, opened 2024-04-25
G2 issue#14966 

a22marca:
Added a new file for the microservices, with the function from gitcommitService.
See issue for testing.


a22eddpe (changes requested):
@a22marca 
should have three steps back in path
E.g
include_once "../../../Shared/basic.php";
include_once "../../../Shared/sessions.php";

The created ms is not at the moment reachable at lenasys and therefor cant be tested as intended. Github repos cant currently be created and therefor they cant be updated as well...

a22marca:
Will be adding the fixes mentioned in the change request and create a new pull request, because of the weekly merge.

PR #14983, opened 2024-04-25
G2 2024 v4 issue#14929

a22marca:
Added the microservice updateListEntryGradesystem_ms.php. See issue comments for testing.


a22melst (approved):
It works. I think there should be permission checks inside this microservice eventually, but it is OK for now since it is not based on any existing service and therfore hard to test.

PR #14993, opened 2024-04-25
G3 Merge issue#14991

a22marca:
Added packages to be installed when the container is built. These are needed to upload files to a dugga.


a22willi (approved):
Looks good, and works as intended.

Issue #14979, opened 2024-04-25
Break out ER into function


a22solmi:
## Description
drawElement() is a 900 line long function. There is a lot of repetition and if else cases that are hard to read and understand.
Break out remaining ER related cases into a function.
Part of #14891

## Job
Break up this function into smaller pieces to increase readability

PR #14995, opened 2024-04-25
Broke out ER into function #14979

a22solmi:
Broke out ER into functions. Restructured drawElement()

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1275,8 +1275,8 @@ var defaults = {
         kind: "ERRelation",
         fill: color.WHITE,
         stroke: color.BLACK,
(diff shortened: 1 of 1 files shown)


a22alfna (commented):
It seems like when the ER elements where broken out the code for determining the minimum height and width of elements were affected since now the ER entity and attributes are rezisable to a much smaller size than before 

On the issue branch:
image

On the week 4 branch:
image

a22alfna (approved):
Never mind the different minWidth and minHeights are results of the new minWidth and minHeight code being merged into the week branch after this issue was formed so that code doesnt exist on this specific branch. The rest seems to work as intended.

Issue #14876, opened 2024-04-24
Replace js alerts with toasts


a22willi:
in #14682 toast functionality was implemented. Use this to replace some of the existing alerts with toasts. Replace all that you can find, where it makes sense to have a toast instead of an alert.

a20jonda:
https://github.com/HGustavs/LenaSYS/wiki/Toast

a20jonda:
Starting with courseed.php/js

a20jonda:
a fix for toast container blocking buttons will be included

PR #14996, opened 2024-04-25
Replace js alerts with toasts #14876

a20jonda:
Added toasts where they fit and could be viewed by and user and fixed toast container permanently blocking elements behind it.

Files affected
image
 
 Most Alerts are connected to debugging and are not expected to be viewed by an user, and or used to block refresh until OK response, therefore the majority is still not touched.
 
 Most toasts inserted are connected and responses to the forms listed [here](https://github.com/HGustavs/LenaSYS/wiki/Popup-form-conventions#currently-known-popup-forms-on-lenasys)

### DIFF
--- a/DuggaSys/accessed.js
+++ b/DuggaSys/accessed.js
@@ -147,25 +147,25 @@ function verifyUserInputForm(input) {
 
 	// Verify first name
 	if(verifyString = validateName(input[1])) {		// Returns null if there is no error
-		alert(verifyString);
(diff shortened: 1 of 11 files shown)


d22isaka (approved):
Looks good for me! 

I've tried the functionality in accessed, courseed, codeviewer, duggaed and sectioned.php The toasts seem to work as intended.

d22mikwe:
LGTM!

PR #14998, opened 2024-04-26
G2 2024 v4 issue#14965 Create microservicec refreshGithubRepo_ms.php

a22albjo:
Se issue 14965 for how it is tested and to what extent i have been able to test it.

### DIFF
--- a/DuggaSys/microservices/gitCommitService/refreshGithubRepo_ms.php
+++ b/DuggaSys/microservices/gitCommitService/refreshGithubRepo_ms.php
@@ -11,9 +11,9 @@ ini_set('display_startup_errors', 1);
 error_reporting(E_ALL);
 
 // Include basic application services!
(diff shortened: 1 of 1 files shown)


a22albjo:
Create microservicec refreshGithubRepo_ms.php

a22marca (approved):
As you described in the comments, it's not entirely possible to test that this functionality works. But I see what you mean, it tries to use a link to refresh the course.

In the other cases of microservices from gitcommitService, I have simply moved the functions into separate files. The way you have done this seems a lot better, and makes the file a actual microservice, since it may take POST requests. We should discuss the method of implementing the microservices from gitcommitService as a group, when we get the opportunity.

Even though this can't be fully tested yet, I believe this is ok to merge for now. We will have to look further into the issue when integrating the microservices.

PR #14999, opened 2024-04-26
G2 2024 v4 issue#14962

a22marca:
Added the function into a separate file. See issue comments for testing steps.


a22iliru (approved):
Tested through instructions and it works

PR #15001, opened 2024-04-26
Created the file updateSecretQuestion_ms.php issue#14930

c22emigr:
Created and tested a new microservice for updating or adding a security question for a user (not for teachers or superusers). Tested to confirm that it works by changing the url on row ~ 31 to use our new microservice updateSecurityQuestion_ms.php in the file profile.js

### DIFF
--- a/DuggaSys/microservices/sharedMicroservices/updateSecurityQuestion_ms.php
+++ b/DuggaSys/microservices/sharedMicroservices/updateSecurityQuestion_ms.php
@@ -14,13 +14,13 @@ include_once "../Shared/sessions.php";
 pdoConnect();
 session_start();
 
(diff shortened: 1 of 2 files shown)


c22cargo (changes requested):
It works, but a few changes should be made.

This should be an MS call:
```
if(isset($_SESSION['uid'])){
	$userid=$_SESSION['uid'];
}else{
	$userid="1";		
} 
```

There's also a lot of nested IF-statements which reduces readability. Ideally there should not be so many, a few have to be nested but some can be un-nested.

There are places with double blank rows instead of a single blank row, for example row 24 and 25.

The closing ```?>``` tag should be omitted from files containing only PHP.

Issue #14878, opened 2024-04-24
Box selection following mouse 


a22jacka:
## Description
It's possible to make the box selection follow the mouse without holding down leftclick by either:
* making a selection, press <kbd>1</kbd> while holding down leftclick, let go of leftclick, press <kbd>2</kbd>
* pulling the mouse outside the site (to broswer search/tabs, taskbar, etc) while making a box selection

## Job
Cancel the box selection if the mouse leaves the site or if another tool is selected

PR #15010, opened 2024-04-26
G1 2024 V4 #14878

a22ricni:
Added a eventlistener for when the mouse is leaving the window/website. Can test it by select the box selection tool and use the tool, but before the left mouse click is up have the mouse outside of the website/window. When your mouse is outside of the website you can have the left mouse button up, and the box selection should be gone when the mouse is returning to the website.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1754,6 +1754,10 @@ document.addEventListener('keydown', function (e) {
             displayMessage(messageTypes.ERROR, "Error: You can't place elements too close together.");
         }
     }
+
+    if (altPressed == true) {
+        mouseMode_onMouseUp();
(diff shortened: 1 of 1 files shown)


a22ricni:
Now should the change request be fixed. Added when you press alt key then you cancel the box selection. So when you press alt+tab then it should not be a problem with the box selection anymore, but don't now if this is a good solution.

a22alfna (changes requested):
The only flaw I found with these changes are that it is still possible to move the box selector without having the mouse pressed down if you leave the browser and then press down the mouse again before moving it back inside then when you release the box selector keeps following the mouse. 
image

a22alfna (approved):
Seems to be working as intended

Issue #14916, opened 2024-04-24
Fix setAsActiveCourse_ms.php


a22iliru:
Add pdo as parameter, dont create a new one with pdoconnect
Remove as many includes as possible
Remove session_start()
The reason for this: This things are already done by the calling microservice

IF IT DOESNT NEED PDO ( DOES NOT EXECUTE QUERIES) , JUST MARK ISSUE AS COMPLETE)

PR #15016, opened 2024-04-26
14916 fix setasactivecourse msphp

c20juler:
fixed issue according to description:

### Fix setAsActiveCourse_ms.php #14916

Add pdo as parameter, dont create a new one with pdoconnect
Remove as many includes as possible
Remove session_start()
The reason for this: This things are already done by the calling microservice

IF IT DOESNT NEED PDO ( DOES NOT EXECUTE QUERIES) , JUST MARK ISSUE AS COMPLETE)


c22cargo (changes requested):
You forgot indentation.

You are calling ```getUid()``` for some reason, but it's not assigned to anything and ```$userid``` is not even used.

You should also ```echo json_encode($debug)```. The variable is never used.

c22cargo (approved):
Ideally ```echo json_encode($debug);``` should not be inside the if-statement. But in this case I don't think it really matters.

Approved.

Issue #15000, opened 2024-04-26
Disable lines between ER-elements


a22jacka:
## Description
You shouldn't be able to create lines between ER entities since they need a relation between them

## Job
Make it so that ER entities can't be directly connected to eachother

PR #15018, opened 2024-04-26
G1 2024 V4 #15000

a22ricni:
Should not be able to draw a line between two ER entity anymore.


b22chrmo (approved):
Drawing a line between ER-entities is no longer possible. Code works properly. No new bugs found.

PR #15021, opened 2024-04-26
I added a margin-left to both the top and bottom squares so that they…

a22alfna:
… are centered on the dotted line going through the middle of the elements.


a22ricni (changes requested):
The square is aligned with the lifeline-lines, but the top square is not aligned with the bottom square and the whole lifeline actor is not aligned with the whole element. So I think the actor is the problem and not the little square at the bottom. The actor should be a few pixels to the left.

Issue #14882, opened 2024-04-24
Cardinalities in wrong position


a22jacka:
## Description
Placing cardinalities on ER-lines causes the 1/N to appear next to the relation instead of next to the entity like it's supposed to

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/c823907d-4587-45bd-a6bc-f37077c350c5)

## Job
Move the 1/N so that it's next to the entity again. This should happen regardless if the line was dragged from or to the entity

PR #15022, opened 2024-04-26
Changed position for cardinalities on ER-lines. (#14882)

b22chrmo:
Changed it so cardinalities are now placed next to the ER-entity instead of the relation.


a22vioja (approved):
I have tested both cardinalities(1,N)  and dragged the line from both sides (entity and relation), it works the same - cardinalities stays at the same position on entity side. 

Screenshot 2024-04-26 at 15 06 40

I also tested a longer distance between entities and it works as expected. 

Screenshot 2024-04-26 at 15 09 59

Issue #14857, opened 2024-04-23
Scale sequence actor on the height aswell


a22jacka:
## Description
If a sequence actor is scaled on the X-axis then the character expands on height aswell and then get's hidden because of the overflow. 

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/effb2d51-40ce-4d28-8761-dfc866de411d)

## Job
Make it so that sequence lifelines also scale vertically when scaled horizontally. 
**NOTE**: The lifeline is **_NOT_** supposed to scale horizontally when scaled vertically

a21liltr:
Solved by PR #15999.

PR #15026, opened 2024-04-26
fixed Scale sequence actor on the height aswell #14857

b22samer:
b22samer and a22robko have fixed the issue where adjusting the width (X-axis scaling). We are now using widthChange to manage how much the width appears to change without actually modifying it when scaled vertically. We've added heightChange to correctly adjust the height based on any width changes during horizontal scaling.
Screenshot 2024-04-26 at 15 43 53

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -2694,30 +2694,34 @@ function mmoving(event) {
             // Functionality for the four different nodes
             if (startNodeLeft && (startWidth + (deltaX / zoomfact)) > minWidth) {
                 // Fetch original width
-                var tmp = elementData.width;
-                elementData.width = (startWidth + (deltaX / zoomfact));
+                var originalWidth = elementData.width;
+                elementData.width = startWidth + (deltaX / zoomfact);
 
                 // Deduct the new width, giving us the total change
-                const widthChange = -(tmp - elementData.width);
+                var widthChange = elementData.width - originalWidth;
 
                 // Fetch original x-position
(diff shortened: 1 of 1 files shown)


a21liltr (changes requested):
Unsure if this is supposed to happen? 
1. Sequence Activation gets really big (most other placable objects also get really big when resizing on x-axis).

2. State Diagram State looks weird and the text is... well, see image below.
image

Original behaviour of State Diagram State (can be changed on the x-axis with the title kept on 1 line or "compact".
Rest of the text in the other box (attributes box) also compact between multiple lines (text unaffected by height change).
image

Definitely a bug:
When clicking the Sequence Activation, it also jumps up (seen on 0:25 in video below).

https://github.com/HGustavs/LenaSYS/assets/129370135/878c6d7a-bc9c-4633-84f2-ac44a783369b

a22robko:
We changed all the code back to how it was before, and it seemed like the error still looked the same when we reverted the code to see the state diagram break. But we reverted all the code and started from the beginning, and the state diagram was still broken. So we think the state diagram was fixed after we created our branch. And that it will be fixed when everything is merged together.

a22jacka (changes requested):
The changes appear to have been made on every object and not just the lifelines. Resizing also broke more since on some elements (sequence activation at least) you can't scale on the left side.

https://github.com/HGustavs/LenaSYS/assets/129258593/f1db41d5-b6ab-416b-8a73-1d1ed6aafba8

Issue #14734, opened 2024-04-18
IE primary keys


a22jacka:
## Description
There is currently no way to underline attributes in IE to properly mark them as primary keys.

## Job
Implement a button (and preferably a <kbd>ctrl</kbd> + <kbd>U</kbd> keyboard-shortcut) to make text underline

a22robko:
We were assigned the task of creating primary keys, and we managed to do so by adding "PK," which stands for primary key. We thought this was the simplest way, but we understand that it can be difficult to know when to write "PK" in diagram.php without that information, as it's not the most natural way to do it.
image1:
<img width="896" alt="Screenshot 2024-04-23 at 16 24 35" src="https://github.com/HGustavs/LenaSYS/assets/129367092/35c6826f-190b-49d2-9d40-43a21bf5a540">
The code checks if the text from each item is longer than two letters and ends with "PK" which usually means "Primary Key". If this is true, it makes the text underlined to show that it is a primary key property. If not, it just uses a regular 'text' style.
image2:
<img width="782" alt="Screenshot 2024-04-23 at 16 30 44" src="https://github.com/HGustavs/LenaSYS/assets/129367092/dac79ff8-dc3c-4fbc-b1e4-083e765535bd">

PR #15028, opened 2024-04-26
Primary Keys in IE Entities are now underlined #14734

a21liltr:
image

Solution includes:
+ add text area for primary key in PropertyFieldset in Option pane
+ adding a property to the IEEntity object called "primaryKey"
+ adding a switch case for "primaryKey" during the looping through properties of IEEntity when saved (in saveProperties() function).
+ adding an if condition to underline text in the textbox part inside drawElementIEEntity() function that draws the object/element on the diagram.
+ adding signs (stars * or hyphens -) based on if primary key or attribute (adds automatically if user doesnt manually type them in
+ also possible to have multiple primary keys (can easily be changed later to only have one and any additional text will be added to the first line.


d22hanol (approved):
The addition of primarykey handling works good. the code follows the structure and is clean. Everything seemed to work during testing. 

Issue #15042, opened 2024-04-27
Remove getPoint()


a22solmi:
## Description
This function creates a new Point, this can already be done with "new Point()"

## Job 
Replace all uses with "new Point()"

PR #15043, opened 2024-04-27
Replaced getPoint #15042

a22solmi:
Replaced all instances of function


b22jonan (approved):
read through the code, things look good. checked that selecting things with box selection still works. i approve

Issue #15046, opened 2024-04-27
Remove common factor drawElement


a22solmi:
## Description
All if statements in drawElement have been moved to separate functions. There is some code that is repeated across all functions.

## Job
Move the common code to drawElement

PR #15048, opened 2024-04-27
Moved common code to drawElement #15046

a22solmi:
Moved div and variable init from functions to drawElement.


d22hanol (approved):
Tested the branch by playing around with all of the elements, nothing seems to have broke. Everything seems to be behaving correctly.
image

Issue #15044, opened 2024-04-27
Replace find...FromLine()


a22solmi:
## Description
There are several functions that do the exact same thing, just for different entity types.

## Job
Create one function that can compare any entity type.

PR #15049, opened 2024-04-27
Replaced functions #15044

a22solmi:
Replaced all find line functions with one that takes a type


d22hanol (approved):
Tested the functionality related to the changes in the branch but found no immediate problems, everthing seems to work as normal.

PR #15066, opened 2024-04-29
Copy content from existing course version fixed

a22ahmom:
It seemed like the solution to the problem was to add another else if statement but for CPYVRS and fix a syntax error. So now it should be possible to copy content from existing course version when creating a new course version. The correct format for creating a new course version is

Version ID: 2001
Version Name: HT01
Start Date: 01/04/2024
End Date: 31/05/2024
MOTD: MOTD
Copy content from: ST20 - 52432

### DIFF
--- a/Shared/dugga.js
+++ b/Shared/dugga.js
@@ -1051,13 +1051,6 @@ function AJAXService(opt,apara,kind)
 				success: returnedCourse
 			});
         }
-		/*$.ajax({
-			url : "courseedservice.php",
-			type: "POST",
-			data: "opt="+opt+para,
-			dataType: "json",
-			success: returnedCourse
-		});*/
 	}else if(kind=="VARIANTPDUGGA"){
 		$.ajax({
 			url: "showDuggaservice.php",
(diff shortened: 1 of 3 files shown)


d22mikwe:
To Tester:
Check that copying of courses works.
Visual problems and copied content of duggor will be dealt with in a separate issue.

a22maxra (approved):
image
You are now able to create new versions and a course that gets saved

Shared/dugga.js
```diff
@@ -1015,17 +1015,50 @@ function AJAXService(opt,apara,kind)
...
-		}else if(kind=="VARIANTPDUGGA"){
+        }
+		/*$.ajax({
+			url : "courseedservice.php",
+			type: "POST",
+			data: "opt="+opt+para,
+			dataType: "json",
+			success: returnedCourse
+		});*/
+	}else if(kind=="VARIANTPDUGGA"){
```
d22mikwe, line 1061: Remove comment?

d22mikwe (changes requested):

d22mikwe (approved):

d22mikwe:
LGTM!

Issue #15030, opened 2024-04-26
Seperate buttons


a22jacka:
## Description
The "save" and "lock" buttons in an element's options are way to close together and need to be seperated a bit

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/ad0912ac-d28a-4f6f-bf54-ddb6f19b1bdd)

## Job
Add a margin or offset to seperate the buttons

Personal opinion but the hover effect on the buttons are too slow and should be sped up a little bit. It doesn't have to be sped up but could look better

PR #15069, opened 2024-04-29
Separated the buttons and made the hover effect faster for issue#15030

a23ponek:
Added a margin: 5px; on line 326 and decreased transition to 0.1s on line 327 in diagram.css.
Link to issue:
https://github.com/HGustavs/LenaSYS/issues/15030


b22chrmo:
'Save' and 'Lock' buttons are now separated. No new issues found. The hover effect is faster making it more clear when it's being hovered.

b22chrmo (approved):

Issue #14945, opened 2024-04-25
Add "undo" toast type 


a22willi:
Add functionality to create an "undo" toast. A toast that shows up for x seconds, and while its visible, the user should have the option to click undo, to reverse whatever action they just did.

Requirements:
- Add the undo type
- Add the undo button for the "undo" toasts
- After the deadline/timeout for a toast has passed the toast function should return true if the user clicked undo. False if the timer passed.

a20jonda:
![image](https://github.com/HGustavs/LenaSYS/assets/102598258/fcf9ad86-28c1-4383-bc72-7dd3d3c4c70f)
An already built in undo button appears in sectioned.php when you delete a section. Will add a undo toast alongside it

PR #15071, opened 2024-04-29
Undo toast type #14945

a20jonda:
Added an toast to sectioned.js where a section gets deleted.
image
Regular
image
Hover

Toast now takes 4 arguments, the last argument is a default argument and does not need to be included in every toast() call
image
The argument is used to add an method call from sectioned.js to sectioned.php.
image

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -9576,6 +9576,24 @@ margin:2px;
   padding-left: 2pt;
 }
 
+.toast .clickable{
+  border-radius: 4px;
+  height: fit-content;
+  background-color: var(--color-primary);
+  color: var(--color-text-header);
+}
+
+.toast .clickable .typeIcon{
+  cursor: pointer;
+  margin: 1rem 0 1rem 0;
+}
+
+.toast .clickable:hover{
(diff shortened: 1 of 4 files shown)


a18chrkr (commented):
Undo toast work great, both in light and dark mode. So does all the other toast. Have also tested on mobile screens. The code also looks clean.

Shared/css/style.css
```diff
@@ -9576,6 +9576,24 @@ margin:2px;
...
+.toast .clickable .typeIcon{
+  cursor: pointer;
+  margin: 1rem 0 1rem 0;
+}
+
+.toast .clickable:hover{
+  background-color: #80eb80;
+  color: var(--color-text-label);
+  cursor: pointer;
+}
```
a22moamy, line 9595: Please add a space before every "{"-symbol to follow the K&R code standard.

a22moamy (changes requested):
Look at the comment I gave, otherwise everything looks good! Will be approved once small indentation stuff is fixed.

a22moamy (approved):
Looks good!

Issue #15050, opened 2024-04-28
generateContextProperties repetition


a22solmi:
## Description
The entire generateContextProperties function is hard coded. There is a TODO for changing it to be more dynamic.

## Job
Refactor this function for readability

PR #15072, opened 2024-04-29
G1 2024 v4 #15050

a22solmi:
Moved element and line properties into own function.
Test option pane properties for combinations of elements and lines.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1,4 +1,5 @@
-// =============================================================================================
+// =
+// ============================================================================================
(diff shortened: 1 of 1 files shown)


a22jacka (approved):
I tried placing a bunch of elements and connecting them and everything seems to work the same as it did before, just with a few unnecessary settings removed.

Not sure if there's issues about these but I found:
* There can only be max 1 connecting between an ER relation and entity, 2 are needed for an ER entity to make a connection with itself
* IE and Class entities can't connect with themselves

PR #15073, opened 2024-04-29
Created mouseEnterSeq() and snapSAToLifeline()

a22cazlu:
Created mouseEnterSeq() and snapSAToLifeline(). Now the ghost element snap to a lifeline if hovering it. Also changed de default heights on seqActivation, seqObject & seqActor.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1407,7 +1407,7 @@ var defaults = {
         fill: color.WHITE,
         stroke: color.BLACK,
         width: 100,
-        height: 150,
+        height: 500,
         type: "SE",
         //actorOrObject: "actor",
         canChangeTo: null,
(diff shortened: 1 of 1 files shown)


a21liltr (changes requested):
Cannot place the Sequence lifeline after placing a Sequence Activation on the lifeline.

https://github.com/HGustavs/LenaSYS/assets/129370135/22620957-9d7c-4c55-9935-3081d6b8662e

Please fix so it can be placed.

a23aleva (commented):
The snap to lifeline functionality seems to be already implemented in the current version of lenaSYS but this particular code addition also introduces a bug where the lifeline can't be relocated if picked up after a sequence activation has been attached to it. This bug wasn't hitherto present and therefore it is deemed best to remove the pull request rather than merging it to master.

PR #15077, opened 2024-04-29
Update changeActiveCourseVersion_courseed_ms.php

a22fokol:
The file has now been renamed as requested, to  changeActiveCourseVersion_courseed_ms.php.   i have try to modify a word in the text comment to be able to create the pull resquest of this issue because it was difficult to do it just by changing the file name


a22iliru:
This should have been solved by moving the file over to desktop, renaming it and then moving back to the repo. now it has difficult conflicts and easiest way is to make a new pull request and do as i said. I will fix it

PR #15078, opened 2024-04-29
Implemented microservice getIndexFile

a22iliru:
I took the function from gitfetchService.php and put it here. What it does and how it should be tested I have no idea. Someone experienced enough to find out should test it


a22albjo (approved):
This code is a copy of the function inside the gitFetchService.php, i don't think this functionality is fully implemented yet and therefore it is hard to test. To my knowledge group 3 is working with this. However the function is internal to the backend and will be used by another service. I think this can be merged but might need to be revisited once the functionality is fully implemented.

PR #15083, opened 2024-04-29
B22samer and a22robko adjusted the calculation for determining the mi…

b22samer:
…nimum height of the note element to accommodate 2 lines of text by default instead of 4 as earlier.


a23ponek (commented):
The default minHeight now accommodates 2 lines of text and it looks better when it isn't taking up to much space. See attached image. The overall functionality of the note seem to work as intended and doesn't cause any errors. Although, the pull request have merge conflicts that I haven't been able to locate, these must be fixed before the solution is approved.
Screenshot 2024-04-30 132341

a23ponek (approved):

PR #15089, opened 2024-04-29
G2 2024 v4 issue#14959

a22marca:
Added the function into a microservice. See issue comments for testing.


a22albjo (approved):
Works as intended, can be tested and verified by removing the function including the file in the monorepo.

Issue #14698, opened 2024-04-18
Derive components from installer mockup


a22willi:
Split the [installer mockup](https://github.com/HGustavs/LenaSYS/issues/14534) into smaller units/components, that can be worked on in parallel. Examples:
- BreadcrumpPanel.
- checkBoxPanel.
- ProgressBar.
- And more. 

The fonts, buttons, and smaller units that only require one css class, and are used extensively should be implemented in [#14569](https://github.com/HGustavs/LenaSYS/issues/14569). This should be for "bigger" components, that also require the use of HTML/JS/PHP. All the components should have its text/shadows/colors defined through the `installer.css` file for a more unified design language with as little repeated code as possible.

Ideally, the components should be defined as php classes, so that if we need a progress bar we can create one like this:
```
$pb = new progressBar();
$pb.setProgress(10); // set it at 10%.
$pb.setStep("Creating new database user"); // display the current step.
```

At the end of this issue, files called something like `installer_components.php` & `installer_components.js` & `installer_components.css` should be completed. Note, there should only be one of each for all components.

This is a lot of work, so it is recommend creating a separate issue for each component, so that more people can be assigned to work on the different components in parallel.

All the files should be located in the directory `newinstaller`.

PR #15090, opened 2024-04-29
G3 2024 v4 #14698

a18chrkr:
### Added PHP file with components
** Co-commit with @b22chran **

We have updated the stylesheet for the new installer and finalized a PHP-file with functions that represent complete components. We were however unsuccessful to link the php-file to the js-file "new_components.js". All file names and functions names are at the moment test-names and may be subjects to refactoring later on.


newinstaller/new_components.php
```diff
@@ -0,0 +1,157 @@
...
+            }
+        }
+
+        function breadCrumbDecr() {
+            stepSelected -= 1;
+            if(stepSelected<=1){
+                stepSelected = 1;
+            }
+        }
+    </script>
```
a22moamy, line 156: Please separate the javascript code from the HTML-code

a22moamy (changes requested):
See the comment, then we can start testing this pull request :)

a22edvbu (approved):
All looks good. Great work!

newinstaller/style.css
```diff
@@ -244,3 +253,32 @@ input[type=checkbox] {
...
+        height: .75rem;
+    }
+
+    .progressBarIndicator{
+        font-style: italic;
+        background-color: var(--secondary-color);
+        width: 15%;
+
+    }
+}
```
a22moamy, line 284: These selectors are nestled, is this SASS or is it a mistake? Rejecting it for now.

a22moamy (commented):
See comment. If this is SASS and on purpose, please comment so and it will be approved.

newinstaller/style.css
a22moamy: Seems like CSS and not SASS, please fix.

d22mikwe:
NOT REAL MERGE. WAIT FOR OTHER MERGE

Issue #15012, opened 2024-04-26
Update accessedservice.php to make tests possible.


b21isabe:
The test-file accessedservice_tests.php is currently not able to perform tests on the services of accessedservice.php.
These are the known problems that need to be fixed:
- $userid is not set when testing, making authentication fail and no operations be run/data be returned.
- The service of adding users with a class is broken. (make an isset() to check when a class is set in the newusers json)
- In the output array: $entries, $courses, $submissions and $examiners are not properly updated and are returned empty.

b21isabe:
There are some annoying problems that are keeping me from fixing the file. The output for $entries is adding the uid to every item which makes testing impossible since uid:s are automatically incremented, the output will be different every time a new user has been added. The output can't be changed because other parts of the system is using the $entries output and it looks like changing it breaks those systems, but I will have to test that some more. It also seems like I have broken some other parts of the website, so I am going to have to undo some of the things I have done until it fixes itself and then find other ways of fixing it.

b21isabe:
Removing uid from $entries seems to work now, I think that as long as you keep the output items in json format, everything will keep working.

PR #15095, opened 2024-04-29
G2 2024 v4 #15012

b21isabe:
It is now possible to make tests in accessedservice_tests.php. I have created 3 functioning tests to check that it is working. I have commented out all the tests that haven't been updated. New issues will need to be created for these, when this has been tested/merged.


c22cargo (approved):
Looks good

Issue #14992, opened 2024-04-25
Can't delete elements inside superstate with mouse


a22jacka:
## Description
If you click on the X on an element inside a superstate then you select the superstate rather than deleting the element.

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/c270cab1-69b1-4c11-be73-29e003d79021)
After clicking the X
![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/dce6925a-3f7b-40f6-83e7-4d53463b5ea0)

## Job
Fix so that elements inside superstates are deleted if their X is clicked

PR #15099, opened 2024-04-29
G1 2024 V4 #14992

a22ricni:
Now should it be able to click on the cross/X on the element when it is in a super state or Sequence condition. Can test it by first place out a super state or sequence condition, then have a element on top of it. After that try to delete the element by clicking on the cross/X when the element is selected.


b22samer (approved):
Testning

The first Diagram.php file is before the changes, and the other one is after the changes. As the video shows, it is now possible to remove an element by clicking on the cross/X on the element when it is in a super state.

a21liltr (changes requested):
This makes it possible to delete objects/elements without first having it selected. This should not be possible. (See Video)

Find another solution. 


https://github.com/HGustavs/LenaSYS/assets/129370135/e0e0c8ca-02b5-40de-83ac-9a03ecd99d3e


https://github.com/HGustavs/LenaSYS/assets/129370135/aeed60c5-acda-4da3-9d17-a04abd294d39

Notice how the actor does not have the X and is NOT selected (Super state is selected) but was deleted anyways,

Issue #15024, opened 2024-04-26
Document the stages of the installer


b22marah:
Document all the steps that the current installer goes through in a simple to follow list. 

Ex:
- Requests db credentials.
- creates table x, y and z.
- creates user. 
- etc.

Note: this is meant to document the steps in the PHP file install.php, not what the user sees when installing.

a22willi:
The stages of the installer are as follows: 
- init.
    - Check if credentials file already exist. And import settings if it does
    - Get working directory of LenaSYS
    - Show UI.
    - Check post variable for selected settings. 
    - Verify that all credentials have been set.

- Establish pdo connection
    - Try to connect with pdo using root mysql account.
    - check if overwrite user / db.
    - delete sqlite db for github integration.
    - Check file permissions for root directory of LenaSYS. ( wrong order; should be called before deleting sqlitedb)
    - create new database. 
    - Create new user. 
    - Grant new user permissions for created db.

b22marah:
### Steps following @a22willi comment...
- Transaction?
    - Remove all comments from .sql file.
    - create transaction?

- Test data (all if checked).
    - Add testdata.
    - Create democourse.
    - create testingcourse.
    - Add codeexamples.
        - Add courses directory.
        - add course 1.
        - add course 2.
    

- Save credentials.
    - Save the current credentials to coursesyspw.php (write to file)
    - check upload size parameter.

a22moamy:
@HGustavs Is it okay to remove the "Yes I want to perform the database-initializing as an transaction" option in the installer as it does not seem to have any real function and only causes errors? @a22willi and @b22marah have analyzed the code and deemed it redundant. 

By "Yes I want to perform the database-initializing as an transaction" I mean this option in the installer:
![Skärmavbild 2024-04-29 kl  11 33 52](https://github.com/HGustavs/LenaSYS/assets/129263915/413d7172-562b-46a6-ae07-0abbe817d0aa)

a22moamy:
Regarding the issue, the documentation is great and concise. 
Well done!

Issue #14987, opened 2024-04-25
Create a new installer.php file


a22willi:
Create a new installer.php file with all of the appropriate function calls, that need to be performed to install the system. 
A list of all the potential functions needed should be listed in: #14649

NOTE: Should be created in the directory for the new installer. `newinstaller/installer.php`.

PR #15102, opened 2024-04-29
G3 2024 v4 #14987

b22marah:
## New additions to new newinstaller

We have added a folder to hold tools, which will be helper functions/classes for installer functionality. We also added installer.php & installer_ui.php which will be the main files for the installer, where installer_ui will be used explicitly for separating and containing the UI part of the installer from the Logic which will take place in the installer.php file.

#### Note
This pull request doesn't include any functionality, instead only focuses on creating a template to implement the different function from issue #15024.


a20jonda (approved):
Conforms to PHP naming convention and K&R indentation. Planning seems comprehensive.👍

d22mikwe:
LGTM!

Issue #15100, opened 2024-04-29
entityIsOverlapping repetition


a22solmi:
## Description
There is a lot of code repetition in changeLineProperties.

## Job
Change this code to be more dynamic

PR #15108, opened 2024-04-29
Removed repetition and nesting #15100

a22solmi:
Removed repetition and nesting in entityIsOverLapping


d22hanol (approved):
the functionality regarding entity overlapping still works, nothing seems to have broke, the elements that should allow overlapping still do so.
I approve the merging of this code

PR #15116, opened 2024-04-29
preventDefault() added to MOVING_OBJECT

b22davbe:
Implemented precentDefault() into the keybinds MOVING_OBJECT (UP/DOWN/LEFT/RIGHT) solves the issue that the screen scrolls when moving an object.


c22abdha (changes requested):
image
Select marker mismatch/breaks with object after moving with arrows.

Issue #15031, opened 2024-04-26
Fix the note element


a22jacka:
## Description
Scaling a note element works fine on the X-axis but breaks on the Y-axis. When expanding downwards it scales to fast and when expanding upwards it expands from the top and bottom at the same time.

## Job
Fix the note element so that it expands in the same way as every other element

PR #15122, opened 2024-04-29
fixed note resizing #15031

a22solmi:
fixed note resizing


a22jacka (approved):
I tried to resize it and the note element scales properly now.

The note element does jump slightly when scaling on the left side but that isn't caused by this PR is just how it was before. Should be fixed though

Issue #14947, opened 2024-04-25
Add diagonal resizing


a22jacka:
## Description
Right now you can only scale vertically or horizontally on the elements, but a nice QOL feature would be to scale horizontally and vertically at the same time by dragging in the corners.

Small note: this issue can easily be split into multiple by either making issues for each element, each category, etc

## Job
Make it so that elements can be resized vertically and horizontally at the same time by dragging in the corners

PR #15123, opened 2024-04-29
Corner nodes for entities #14947

d22hanol:


a22solmi (approved):
Corners can be dragged to resize x and y, works with cursor on hover as well. Code looks good!

PR #15124, opened 2024-04-29
copied duggor into the newly created course version is fixed

a22ahmom:
Before it wasn't possible to load duggor when creating a new course version and you would copy content from existing course version. Hopefully it should work now. The tester should try this with demo-course. Keep in mind that with webbutveckling - datorgrafik it won't work because it didn't work before this, so it's normal. So it concerns another issue.

### DIFF
--- a/DuggaSys/courseedservice.php
+++ b/DuggaSys/courseedservice.php
@@ -350,6 +350,7 @@ if (checklogin()) {
 				}
 
 				// Duplicate listentries
+				$momentlist = array();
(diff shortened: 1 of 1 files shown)


a20jonda (approved):
Testig without the fix in demo-course
image
^ Results for all the duggas without fix.

With the fix, it successfully copies the sections without the errors.
Checking, the database with copy and no-copy versions:
listentries appears to be correct.
image
same with codeexamples
image
useranswer:
image

The new code looks nice and follows K&R 👍

d22mikwe:
LGTM!

PR #15132, opened 2024-04-30
Added new CSS

b22chran:
Added CSS in blackTheme.css that uses both the "#Courselist" and "#fabBtn" to get a hihgher priority than previous "#Courselist" and "a". This made the color of the "+" in the fab-button to have the same color as the links in Courselist.


a22danwe (approved):
looks good, the button works and looks as intended

PR #15140, opened 2024-04-30
14986 align lifelines and expansion boxes

a22alfna:
I added 4px to the width of the sequence actor and object that scale with the zoom factor on the page, in order to center the resize squares to the lifelines.
This solution is the best one I have found that seems to scale correctly with the zoom without changing the entire SVG
image
image


b22davbe (approved):
The solution works and i couldn't find any new bugs with these changes.

Issue #14364, opened 2024-04-05
ctrl + z removes lifelines


a22jacka:
## Description
When using <kbd>ctrl</kbd> + <kbd>z</kbd> removes every lifelife along with the latest element

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/27af210a-7c3c-41d3-aca6-6623276f4395)
The most recently changed is the activtion furthest right

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/88048677-3d9f-494c-8580-652b1186071b)
After <kbd>ctrl</kbd> + <kbd>z</kbd> the latest element and every lifeline are removed

## Job
Make it so that <kbd>ctrl</kbd> + <kbd>z</kbd> don't remove lifelines

a22ricni:
What I find is that the variable "felem" is undefined on the line 7430. Also find that the actors but also the diagrams get a negative value on the width when you ctrl+z after you change the size. 
It seems like it gets the same result when you move the actor/ diagrams around first then ctrl+z.

a21liltr:
Observations:

After placing an element on the diagram then reselecting it with the pointer to adjust its size, the selection is somehow still on the element or somehow the resized elements was placed in an array or clumped together with the last placed element, which after pressing CTRL+Z, removes the it along with all the other elements which have had their sizes changed using the pointer.

a21liltr:
Additionally, it seems that after resizing an element then pressing CTRL+Z (which will cause the resized element(s) to disappear), it also breaks the diagram in the way that if you still have a placable element, you can still place it, however you cannot change tool to use, whether by clicking the tool in the toolbar with the mouse cursor or by using the keybindings. it will cause the cursor to lose the active placable element and nothing will react to the cursor at all. E.g. placed elements that were not resized will not react when clicked upon, and pressing other tools in the toolbar will not work either.

I still cannot find the solution to this issue but I believe it lies somewhere in StateMachine.save().
So,I will leave this and hope for the best for the next person that is assigned this monstrosity.
Good luck!

a21liltr:
Will continue working on this issue after weekly merge. A new issue will be created and linked to this.

PR #15141, opened 2024-04-30
ctrl + z #14364

a22jacka:
Managed to partially fixed how <kbd>ctrl</kbd> + <kbd>z</kbd> works

What works: 
* resize down
* resize right
* resize up
* resize left

What doesn't work:
* move around

I'm gonna keep working on this but keeping these changes would be nice


a22jacka:
All the resizes now work correctly. Plan is to make so it only stores once the user finishes resizing since it currently saves every step.
After that I'll look at position undo

a22cazlu (changes requested):
Most of the ctrl + z works, but with ER and UML the resizing dosen't work. se videos for better understanding.
[screen-capture (2).webm](https://github.com/HGustavs/LenaSYS/assets/131054714/e31dc344-51d3-498a-bfd5-1fa39578aab2)
[screen-capture (1).webm](https://github.com/HGustavs/LenaSYS/assets/131054714/9b1e80f6-9832-4097-83e4-177952b1a299)

a22jacka:
Mostly done. Currently known issues are it stores every step when resizing and the UML, IE, and State entities sizes getting messed up, however if this get's merged now I can focus more on them after the weekly merge

a22cazlu (approved):
what i can see nothing else get destroyed but the same issue is still there. And sometime the ctrl + z don't work until I do something else, then the block kind of disappear. And as you say it make all the jumps back to the size it was before when I have resized an object. I can approve it but it really need some more work.

Issue #14112, opened 2023-05-22
Make sequence activation snap to lifeline


b21eriho:
It should be possible for the sequence activation to be placed precisely on the center of the lifeline, naturally it shouldn't be stuck there forever but given a close-enough placement it should snap to the horizontal center of a lifeline

c21adano:
It would appear that pr #14097 has messed up the moving system for elements.

This is what i get when i move elements.
![errorors](https://github.com/HGustavs/LenaSYS/assets/102599332/fce357ef-fd04-41bc-a51a-61b71a3dc598)

b21eriho:
Indeed something is strange, does this prevent you from completing this issue?

c21adano:
a bit but not a 100%, im unable to test the ctrl z function and the site seems to slow down a bit after some time. Ill continue looking into the issue for the next day but if i feel like i havent made enough progress then ill consider giving up.

b21eriho:
Sounds very reasonable

c21adano:
I think i have to give up on this issue for now, the errors seem to break the current element position and over time the overlapping function seems to break down and not function any longer. 

This is a image from our stable branch with 5 er elements placed ontop of each other.
![errors](https://github.com/HGustavs/LenaSYS/assets/102599332/32a73880-83cc-498d-84ba-597c8b66c9c9)

a23ponek:
Tried to create a function for the snap between the sequence activation to lifeline but tools like Class for a lifeline/sequence activation was unclear and where the implementation needs to be was also not strictly obvious. 
This code is a start of but needs changes for correct variables/logic:

function snapSAToLifeline() {

    const lifeline = document.getElementById("sequenceActorAndObject");
    if (lifeline) {
        const lifelineRect = lifeline.getBoundingClientRect();
        lifelineCenterX = lifelineRect.left + lifelineRect / 2;

        for (let i = 0; i < data.length; i++ ){


            if (data.kind == "sequenceActivation") {
                const element = data[i];
                const activationCenterX = element.x + (element.width / 2);
                const distanceToCenter = Math.abs(activationCenterX - lifelineCenterX);
                const proximityThreshold = 20;

                if (distanceToCenter < proximityThreshold){
                    const newX = lifelineCenterX - (element.width / 2);
                    element.x = newX;
                }
            }
        }
    }
}

b22davbe:
[...]

PR #15156, opened 2024-04-30
G1 2024 v4 #14112 - v2

a22cazlu:
revertet a commit, should work now

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -2345,35 +2345,33 @@ function mouseEnter() {
     }
 }
 
-
-function mouseEnterSeq(event){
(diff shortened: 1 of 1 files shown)


a22solmi (changes requested):
Sequence only snapps as a ghost entity. Code is not formatted. The height of elements has been changed, unsure if that was needed. It is a good solution but is not complete. Add snapping for regular entity as well?

a22cazlu:
We have now made some changes. As we wrote in the issue https://github.com/HGustavs/LenaSYS/issues/14112 we could only fix the issue for the ghost. And we was told that it are good for now.

a22solmi (approved):
LGTM

Issue #14849, opened 2024-04-23
Investigate why courseed.php does not work with old database


a22willi:
In issue #14736 It was determined that the courseed.php page wont load when using the old database, figure out why and ensure that it works. 

Before fixing it, ensure that your fix does not break any rules regarding continuous integration. IE, don't change the tables, the best way would probably be to ensure that the LenaSYS page can read both the old and new database.

a22oscja:
Course page appears to load in data correctly, the issue with delete is seen as redundant to fix as the same feature has been worked on in another issue. Also I could not get the delete feature to work in our current database on either branch, hence it will likely be a future issue instead.

PR #15172, opened 2024-04-30
Made courseed.php backwards compatible with the old database G3 2024 v4 #14849

a22edvbu:
When using the old database that the live version of LenaSYS is using, it wasn't possible to load into courseed.php. This was due to missing attributes and tables when comparing the database to the newer one. The main problem was that the database tried to execute queries made for the new database on the old database, so we implemented try/catches that execute an alternate query that works for it. 

**!IMPORTANT!**
To test this you will need to use the old database.

### DIFF
--- a/DuggaSys/courseedservice.php
+++ b/DuggaSys/courseedservice.php
@@ -86,12 +86,21 @@ if (checklogin()) {
 		// The code for modification using sessions
 		if (strcmp($opt, "DEL") === 0) {
 		} else if (strcmp($opt, "NEW") === 0) {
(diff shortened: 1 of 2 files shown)


SleepyBearIV:
can't get the old databace file to import using dockerdesktop, tried to find a solution and found this https://stackoverflow.com/questions/72466258/phpmyadmin-on-docker-with-laravel-sail-post-content-length-exceeds-the-limit. Maybe someone else has better luck than me, or the sytem can be tested on a system using xampp.

a22karja (commented):
Courseed.php works. But I cant create new courses when logged in. I dont know if that has with this issue to do or should be its own issue. 
The page is loading and the links works. Edit the courses works aswell but cant create new ones.

PR #15198, opened 2024-04-30
link symbol is now centered vertically

a22ahmom:
It should now be possible to hover over the big and small fab buttons and see that link chain icon is now centered vertically. Although the issue did not include explicitly to center the link symbol in the small fab button on the top of the page, it is now fixed.


d22isaka (approved):
Hovering over both fab buttons, icon looks good and very centered for me =)

Skärmbild 2024-05-01 090837

Issue #15070, opened 2024-04-29
Eliminate error message in console and fix the code to implement toasts. 


a22moamy:
## The issue
When pressing the "Get Canvas Link" button, an error message is generated in the console. 

"Get Canvas Link"-button (can be found in list entries): 
![Skärmavbild 2024-04-29 kl  10 59 59](https://github.com/HGustavs/LenaSYS/assets/129263915/2ae93b09-16bb-4922-8eb3-8c47e3d80e25)

The error message on Firefox:
![Skärmavbild 2024-04-29 kl  11 00 49](https://github.com/HGustavs/LenaSYS/assets/129263915/5a17c959-77de-45e8-ae41-d296ddb6f5c6)

## The requested fix
Add code that checks whether there is a Canvas link or not before trying to get the Canvas link. If there is none, **an error toast should pop up stating so.** If there is a link, then it can be fetched. Also check if the right link is being fetched, or if the current method uses faulty logic. No error messages should be generated in the console.

d22mikwe:
The showCanvasLinkBox() function seems to be searching for a node element that contains the href link to the codeexample viewed in codeviewer. I'e the <td class="example-item".
Try adding an extra .parentNode in the querySelector.

PR #15211, opened 2024-04-30
Eliminate error message in console and fix the code to implement toasts. #15070

d22mikwe:
Test that you can click on both buttons that calls showcanvasLink and opencanvasLink


d22mikwe:
image
These buttons.
Test for both newly created items and already existing ones

a22edvbu (approved):
When clicking on the link button the code example shows in a different tab and when pressing copy canvas link button a box with he link appears. 

Great Work!

Issue #15111, opened 2024-04-29
"New Item"-button in sectioned.php does not work for smaller screen sizes


a22moamy:
When hovering over the "New Item"-button, there should be a drop-down with options, like this:
![Skärmavbild 2024-04-29 kl  15 40 20](https://github.com/HGustavs/LenaSYS/assets/129263915/76efebd9-170b-4b85-8f9f-a588d0beaed0)

But for smaller screen sizes, no drop-down is created. **Fix the issue by allowing the drop-down to be displayed even for smaller screen sizes**. 

Perhaps the solution could be similar to the "+"-button, whose drop-down works for all screen sizes.
The "+"-button in question:
![Skärmavbild 2024-04-29 kl  15 43 01](https://github.com/HGustavs/LenaSYS/assets/129263915/54fbb203-ccf8-441b-acc2-775a618d7d5f)

PR #15218, opened 2024-04-30
Fixed the top dropdown not showing on smaller screens #15111.

a22edvbu:
The issue was due to an !important set on the element that was displaying the element. By removing this it seems to work.


a22moamy (approved):
looks good

PR #15225, opened 2024-05-01
Fixed visual bug when copying a new course version 

a22ahmom:
It appeared that a red line would appear when one would create a new course version, and for some reason, I hope that it is local otherwise, it might be some weird bug with the code. But hopefully, it should work.

To tester:
Valid input format

Version ID: 2001
Version Name: HT01
Start Date: 01/05/2024
End Date: 30/06/2024
MOTD: MOTD
Copy content from: ST20 - 52432


d22mikwe:
Visual bug is fixed when copying new course.
Red line did not appear for me.

d22mikwe (approved):
Visual bug is fixed when copying new course.
Red line did not appear for me. It was only present in the pre existing demo course

Issue #15040, opened 2024-04-26
Refactor pdo implementation when inserting into database


d22mikwe:
Current implementation is in sectioned.php when fetching git codeexamples and inserting files into relevant databases.

During the rushed development for the halftime seminar a quickfix was implemented.
Restrucutre the pdo implemantation to be more effective and better structured (pdoConnect() / $global pdo?)

Relevant functions:
- updateCodeExampleDB()
- insertIntoBoxDB()
- insertIntoFileLinkDB()

PR #15227, opened 2024-05-01
Refactor pdo implementation when inserting into database #15040

d22mikwe:
Refactored pdo implementation when inserting to db


d22mikwe:
Tested.
All files are written to fileLink, codeexamples and box table in mysql db as expected.
Codeexamples can still be viewed

Issue #14819, opened 2024-04-22
Add properties to sequence lines


a22jacka:
## Description
There's currently no properties on lines that connect sequence lifelines or activations. There's used to be some but it seems like they've been removed.

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/0d0257e2-3f8d-48d8-998a-1ef6db737d52)

## Job
Re-implement the following properties for lines between sequence lifelines and activations:
* Label
* Icons
* Dashed line

PR #15248, opened 2024-05-02
Added icons, label and dashed line to SE #14819

a22solmi:
Added icons, label and dashed line to SE


a23ponek (approved):
The icons, label and dashed line functionality now works as intended and nothing seems to break or cause any issues. See images down below.

Before changes:
Screenshot 2024-05-02 145638

After changes:
Screenshot 2024-05-02 145356

Issue #15157, opened 2024-04-30
Automatic scroll bug fix


d22mikwe:
- #15143
- #14942

In above issue and pr it was noticed that sometimes when creating a new item in the entry list at the bottom the automatic scroll function doesnt scroll all they way down. 
Look into if you can recreate this problem and try and fix it.
@SleepyBearIV was tester ask him for details about recreating the problem.

Edit: Also include fix for instant scrolling in edge and chrome

a20jonda:
Also currently the scroll animation is instant both for chrome and edge when testing.

a20jonda:
![image](https://github.com/HGustavs/LenaSYS/assets/102598258/fe49f41e-11da-4580-b55b-773116990d44)
both chrome and edge uses system settings to check if smooth scrolling should be possible. Unchecking "turn of unnecessary animations" solves the problem. So it was not a bug.

a20jonda:
![image](https://github.com/HGustavs/LenaSYS/assets/102598258/00e6d7b8-dc41-415f-b5c6-209be8759b4e)
The problem is connected to a race condition. If the time is too low, both the scroll clips and the highlight might not show. Was working on a sync solution but unforseen problems occured. Adding an extra 100 ms stopped the problem on my end. 

Also
![image](https://github.com/HGustavs/LenaSYS/assets/102598258/eb6fa483-87bf-4a52-9f6d-61003399b234)
Adding a new section too fast or within 3 seconds between another would make the new toast disappear once the first toasts 3 seconds where complete.

PR #15249, opened 2024-05-02
Scroll getting halted when adding new section, G3 2024 v5 #15157

a20jonda:
Since the bug is only rare and visual, and trying to fix the race condition caused the edit section popup and no animation at all. Adding 100ms more delay is the simplest solution.


a20jonda:
Note that adding a sections faster than 3 seconds between causes jittering and the toast to not show properly

d22isaka (approved):
The solution seems to work. I've added several(~50) items on both firefox and edge and the scroll has been consistent. =)

As stated by jonda if adding items too fast the toast gets kinda wacky but that should be another matter.

Issue #15191, opened 2024-04-30
Create new test for the microservice createNewCourse_ms.php


f22linhe:
See parent issue: _Create new tests for the microservices created from courseedService.php https://github.com/HGustavs/LenaSYS/issues/15187_

Note, for this issue to be resloved the issue _Implement retrieveCourseedService_ms.php in all microservices in the courseedService folder #15304_ need to be resolved.

a22marca:
I'm continuing work in this issue. Creating a new branch to get all the recent updates.

a22marca:
The test is almost done at this point. I'm having some issues with the filters, for some reason "versions" is not filtered out from the values that i compared against the expected value when i'm using the following filters:
```
        'filter-output' => serialize(
            array(
                'entries' => array(
                    'cid',
                    'coursename',
                    'coursecode'
                ),
                'debug'
            )
        ),

```

I believe this is a issue with the actual filter. the versions array does also contain the entries cid, coursename and coursecode. So versions is still collected and compared against it seems.

PR #15272, opened 2024-05-02
G2 2024 v5 #15191 f

a22iliru:
The test does create a new course and deletes it after the test is done. However when opening the file it wont show what passes and what does not unless you delete the 'query before test 'query after test' in this file. I can change if the tester has suggestion on how to fix it, otherwise i suggest merge and look how to make it shown on the page.

I also fixed test.php include_once(sessions.php) and include_once(coursepsysw.php something)


DuggaSys/tests/microservices/createNewCourse_ms_test.php
```diff
@@ -0,0 +1,33 @@
...
+                'coursecode' => 'TEST001',
+                'courseGitURL' => 'http://localhost/LenaSYS/DuggaSys/sectioned.php?courseid=9999&coursename=Test%20Course&coursevers=8888',
+                'AUtoken' => 1
+            )
+        ),
+        'filter-output' => serialize(
+            array(
+                'none' 
+            )
+        ),
```
a22albjo, line 30: In the future if possible filter out some of the output if there are parts that are not relevant to the service.

DuggaSys/tests/microservices/createNewCourse_ms_test.php
```diff
@@ -0,0 +1,33 @@
...
+    ),
+);
+testHandler($testsData, true); // true = prettyprint (HTML), false = raw JSON
```
a22albjo, line 33: You should add a "?>" to mark the end of the file.

a22albjo (changes requested):
The structure looks good, however there are a number of things that needs to be adressed.

Use the retriveCourseedService_ms.php in the microservice you are testing to get the right output. 

I have not gone in to details testing the test it self, i think it is easier to do when the comments listed bellow have been adressed.

a22marca:
Creating a new pull request from a new branch

Issue #15195, opened 2024-04-30
Create new test for the microservice updateCourse_ms.php


f22linhe:
See parent issue: _Create new tests for the microservices created from courseedService.php https://github.com/HGustavs/LenaSYS/issues/15187_

a22marca:
There are some issues with updating a course in the frondend at this point. Group 3 are working on this issue, waiting until I know that the issue is fixed before proceeding here

a22marca:
There is a commented out function in sectioned.js that may use this microservice.

In anycase, this microservice purpose is to return listentries with the visible attribut set to 3.
retrieveSectionedService already returns all listentries, so this microservice is redundant.

a22marca:
It is now possible to update a course. I will continue in this issue.

a22marca:
I believe this is not fixed, group 3 are still working on it.

PR #15275, opened 2024-05-02
G2 2024 v5 #15195

a22iliru:
The test can now update a course. However i dont know how to make the query-before-test and query-after-test work.

### DIFF
--- a/DuggaSys/tests/microservices/updateCourse_ms_test.php
+++ b/DuggaSys/tests/microservices/updateCourse_ms_test.php
@@ -1,0 +1,31 @@
+<?php
+
+include "../../../Shared/test.php";
+ini_set('display_errors', 1);
+error_reporting(E_ALL);
+
+$testsData = array(
+    //------------------------------------------------------------------------------------------
+    // Test for updateCourse_ms.php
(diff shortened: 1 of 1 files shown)


a22marca:
The microservice being tested is not currently using a retrieve service, and it therefore not returning any values. This means that there will be no values to compare a expected-value against. We need to integrate the retrieve service before this test can be properly implemented.

a22marca:
The query-before-test and query-after-test fields are used to first create test data, and then clean up after the test has finished. In this case you could use a query-before-test like the following example to create a "testing" course:
```
'query-before-test-1' => "INSERT INTO course (cid, coursecode, visibility) VALUES (99999, 'IT123G');",
```
You can then use that cid as a parameter in the service-data array.
The query-after-test-1 could be a query that simply removes the course created before the test.
@a22iliru

a22marca (changes requested):
- To make this test work properly the microservice being tested need to use a retrieve service, so that some information is returned.
- Read my comment in the pull request on how to create test data if needed.

a22marca:
Creating a new pull request.

Issue #15230, opened 2024-05-01
Add filesize value in fileLink table db when fetching code examples


d22mikwe:
filesizes can be found for each file in var codeExamplesContent in sectioned.js.
extraction should be made and inserted into ajax request in storeCodeExamples() just like all other attributes.
insertIntoFileLinkDB() in sectioned.php is where insertion is made in db

PR #15276, opened 2024-05-02
Add filesize value in fileLink table db when fetching code examples #15230

d22mikwe:
To test:
After a codeexample has been fetched.
Check that the column filesize in table fileLink for the fetched files has recieved an insert containing their filesize.


d22mikwe:
image
Should look like this

SleepyBearIV (approved):
LGTM the database now saves file size

Issue #15055, opened 2024-04-28
Investigate why an error message is created when re-arranging list entries in sectioned.php


a22moamy:
When re-arranging list entries, an error message will be generated. You can re-arrange list entries in a course by dragging them with these buttons:
![Skärmavbild 2024-04-28 kl  19 57 47](https://github.com/HGustavs/LenaSYS/assets/129263915/d85a3e98-414b-46ac-b651-73e428f82a9d)

The generated error-message on Firefox:
![Skärmavbild 2024-04-28 kl  20 05 07](https://github.com/HGustavs/LenaSYS/assets/129263915/fbe12c8a-57aa-4090-98a7-c49c5166b017)

**Investigate why this error message is created and fix it.**

d22mikwe:
![image](https://github.com/HGustavs/LenaSYS/assets/129263393/4bc9754e-578f-4033-862d-349aa515475e)

PR #15279, opened 2024-05-02
remove unused method call hideDeleted() g3-2024-v5-#15055

a20jonda:
I assume that hideDeleted() was needed at some point to make the drag and arrange feature work properly, but have now been deprecated. There was a possibility of hideDeleted() being renamed but did not find any method matching the "hide deleted" name. The `.deleted` class used to mark the sections that are being deleted within 60 seconds, is not used for any extra hiding functionality.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -1591,12 +1591,12 @@ function returnedSection(data) {
           if (retdata['writeaccess']) {
             if (itemKind === 3) {
               if (isLoggedIn) {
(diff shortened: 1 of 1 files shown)


a22natth (approved):
The testing was done on both Chrome and Firefox, to which no error message(as mentioned in the issue) was presented. The functionalities of dragging and rearranging elements were not affected and work as intended, which is probably due to "hideDeleted()" being used at some point, as described by @a20jonda. The code itself looks good, since only minor deletions were required to fix the issue, therefore indentations remained the same.

d22mikwe:
LGTM!

Issue #15178, opened 2024-04-30
Create new test for the microservice addClass_ms.php


f22linhe:
See parent issue: _Create new tests for the microservices created from _accessedService.php_ #15177_

c22cargo:
To test the implementation, issue #15012 needs to be merged first.

A branch with the implementation will be created after the weekly merge.

c22cargo:
Errors occur because it's hardcoded in ```test.php``` for test-files to be located in a specific place.

c22cargo:
There was an issue with the user id not being set correctly when performing the test, thus making it fail. This was fixed by adding ```checklogin()``` to the ```getUid_ms```.

PR #15280, opened 2024-05-02
g2-2024-v5 issue #15178

c22cargo:
The only output from the MS is ```$debug```, so the comparison is only done to that.

Since this branch was created before the recent ```test.php``` fix, you will need to modify the file. Change row 64 to:
```
include_once __DIR__ . "/../../coursesyspw.php";
```
and row 202 to:
```
include_once __DIR__ . "/sessions.php";
```


a22iliru (approved):
Passed all tests. can be merged

Issue #14986, opened 2024-04-25
Align lifelines and expansion-boxes


a22jacka:
## Description
The blue resizing squares are not aligned with the lifeline-lines on both the actor and objects

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/4c24073b-e9a5-4852-a132-431d1ebf2a45)
![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/442f8d2e-3120-4c84-aa09-494db1457c7e)

## Job
Align them

a22alfna:
Hittade en ny lösning som kan fungera för denna issue men jag är inte säker på att den följer kodstandarder och liknande som behövs. Men den som ska fortsätta på denna issue kan testa iaf. I element.js i slutet av funktionerna för dom eleme
I found a new possible solution for this issue but I am not sure if it follows the right code standards and similar requirements. But the next person trying this issue can try to make it work if it is possible. In element.js in the "return drawSvg" I modified it to look like the following. 
![image](https://github.com/HGustavs/LenaSYS/assets/129264882/9547d1e0-b144-4757-966f-665f069cd1af)
to use the "extra" variable of the "drawSvg" function.

a23fanro:
Opening this issue again as it was not fixed and PR was closed, see PR: [#15287 ](https://github.com/HGustavs/LenaSYS/pull/15287)

PR #15287, opened 2024-05-02
I made the top and bottom resize squares centered again

a22alfna:
I added 4px to the width of the sequence actor and object that scale with the zoom factor on the page, in order to center the resize squares to the lifelines.
This solution is the best one I have found that seems to scale correctly with the zoom without changing the entire SVG
image
image

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -814,7 +814,6 @@ document.addEventListener('keyup', function (e) {
     if (isKeybindValid(e, keybinds.TOGGLE_REPLAY_MODE)) toggleReplay();
     if (isKeybindValid(e, keybinds.TOGGLE_ER_TABLE)) toggleErTable();
     if (isKeybindValid(e, keybinds.SAVE_DIAGRAM)) showSavePopout();
-    if (isKeybindValid(e, keybinds.RESET_DIAGRAM)) resetDiagramAlert(); //new keybind for reseting diagram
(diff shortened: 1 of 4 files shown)


a22ricni (approved):
I looked in first the week branch and tested the sequence actor and lifeline (object). After that i looked at the issue branch and doesn't look like to be any problems with these changes.

a21liltr (changes requested):
This solution is not good as it consist of using a magic number just to make the svg for the object 4 pixels bigger, seemingly randomly. 

This will be hard for new developers to understand when taking over the project as well as it will give the object an extra margin, (but only for this type of object). We have put in a great effort to remove all the unnecessary margins.

Please find a solution that can be implemented in DuggaSys > diagram > draw > node.js, or something better.

a23fanro:
Closing this PR as it does nothing. [Issue #14986 ](https://github.com/HGustavs/LenaSYS/issues/14986) is still open and a new PR will be made for this year (2025).

Original issue: [#14986 ](https://github.com/HGustavs/LenaSYS/issues/14986)

a23fanro (commented):
Closing this PR as it does nothing. [Issue #14986 ](https://github.com/HGustavs/LenaSYS/issues/14986) is still open and a new PR will be made when appropriate. 

Original issue: [#14986 ](https://github.com/HGustavs/LenaSYS/issues/14986)

Issue #15129, opened 2024-04-29
Fetching github repo does not work unless providing a github token


d22mikwe:
- #15125 

After issue above it seems fetching a repo only works (files are inserted into gitFiles in metadata2.db) when providing a github token in edit course option. Not when fetching from inside the course or without providing a github token in edit course. Without a github token only the repo name is inserted in gitRepos.

Job: 
Look into whats causing this issue and make sure the fetching works when user does not provide a token.

Relevant:
- #14563 
- #14391 
- #14526 

Problem with creation of github folder containing repo dirs and files is being worked on in #15091

Issue #15091, opened 2024-04-29
Fetching github repo does not create folder and write files


d22mikwe:
Due to this problem:
- #15129 
Fetching only seems to be possible when providing a github token in edit course option.
---------------------------------------------------------------------------------------------------------------------------------
There are currently problems when fetching a github repo.

A Github folder containing the github repo is not written in .../LenaSYS/courses/$cid when fetching a repo.
$cid is the course id.

Make sure the Github folder is created inside the course id folder and that it contains all the repo files and dirs.
Also make sure that the course folder is created if it doesn't already exist.
Make sure that it works both for mac and windows.

Relevant Issues:
- #14526 
- #14391

d22mikwe:
After #15288 it seems downloading repo to folder was fixed aswell. 
Could you recreate your bransch with g3-2024-v5 as source bransch and make sure it works for mac aswell?

d22mikwe:
Great work! 
#15342 has been created to solve this.

PR #15288, opened 2024-05-02
Fetching github repo now works correctly - #15129

a22oscja:
The issues I could find seemed to be related to the page reloading upon the save button having been clicked (in course) as well as the program simply not being set to "DOWNLOAD" when using one of the windows.

Tested both in course and through edit course, with both working to actually download files now, on top of inserting to metadata.

As I also noticed that the error logging was turned off if no token was present, I also took the time to fix this as i'd imagine it was causing confusion.
I also added a small function to gitcommitservice which simply lets you write into the log (as using echo/print_r no longer gives alerts, likely from toasts now existing).

The main issue currently would be that the downloading of files can be quite slow (took me 8 minutes for lenaSYS), but I could not see any good places to make it more effective with a quick look over.

### DIFF
--- a/DuggaSys/gitcommitService.php
+++ b/DuggaSys/gitcommitService.php
@@ -44,6 +44,11 @@
 	// getCourseID: Fetch the course ID from MySQL with Github URL, then fetch the latest commit from the repo
 	//--------------------------------------------------------------------------------------------------
 
+	function putInLog($message){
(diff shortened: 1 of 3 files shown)


d22mikwe (approved):
I've tested fetching from both the "edit course" option and inside the course without a github token and all files were stored in metadata.
The Github folder was downloaded to their respective folders aswell which may leave #15091 resolved too.

Code looks good aswell

PR #15290, opened 2024-05-02
G1 2024 v5 #14992

a22ricni:
Now should it be able to click on the cross/X on the element when it is in a super state or Sequence condition. Can test it by first place out a super state or sequence condition, then have a element on top of it. After that try to delete the element by clicking on the cross/X when the element is selected. And now you should not be able to press the cross/X button if not the element is selected.


a22hanfa (approved):
The task:
> Fix so that elements inside superstates are deleted if their X is clicked

The result:
This works now, I've tested by using a prior version and comparing functionality. It now works as intended by being able to delete the entities. The code also looks good.

PR #15293, opened 2024-05-02
changed the conditions in EntityIsOverlapping

a22cazlu:
The if-statement that we changed didn't work good att all. If there was a sequence object on the diagram you could place objects on eachother freely. Now you can place sequence objects on eachother as long as it's not the same kind as the overlapping object. Same with UML state objects.

While solving the issue we also fixed so that you only can insert UML-state properties into UML superstate and that you only can insert sequense-diagram properties into loop/condition.

### DIFF
--- a/DuggaSys/diagram/helper.js
+++ b/DuggaSys/diagram/helper.js
@@ -88,14 +88,22 @@ function entityIsOverlapping(id, x, y) {
     })
 
     for (let i = 0; i < data.length; i++) {
-        if (data[i].id === id) continue; 
-        if (context.includes(data[i])) break; 
-       
-        //Checks if the objects are of the same type or allowed types for overlapping.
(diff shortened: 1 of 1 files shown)


c22abdha (approved):
image

I tested the changes and we can see that there is no overlapping.

a21liltr (changes requested):
Should not work like this :

https://github.com/HGustavs/LenaSYS/assets/129370135/7eef2ac3-cd12-4548-87a2-e51f63d0b333

The code is also really hard to read.

Simplify the if-condition by putting the allowed types in a list and then check the list, or create bools with allowed combinations or something else that makes it more readable.

We do not like multiple paranthesis like this.

a21liltr (approved):
LGTM.

Issue #15035, opened 2024-04-26
Arrow keys scroll page on dugga


a22jacka:
## Description
When moving elements using the arrow keys while in a dugga it also scrolls the page at the same time.

https://github.com/HGustavs/LenaSYS/assets/129258593/fbc9e5fd-eaac-42cf-a6d8-ae6739da884a

## Job
Make so that the page doesn't scroll if you're focused on the diagram in the dugga.

a21liltr:
Solved with PR #15295 .

PR #15295, opened 2024-05-02
preventDefault() added to MOVING_OBJECT

b22davbe:
Adding e.preventDefault() to the arrows keybindings prevents the user from scrolling with the arrows while in a diagramdugga.

Regarding the change-request we (me and a22cazlu) couldn't reconstruct the problem trying different browsers and we dont see how adding the e.preventDefaults that almost all other keybindings also have would be the reason for the problem that the reviewer described. Is it possible that the reviewer didn't test the issue in the correct landscape or that any other factor might be the reason for the problem?

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -665,6 +665,7 @@ document.addEventListener('keydown', function (e) {
 
     // Moving object with arrows
     if (isKeybindValid(e, keybinds.MOVING_OBJECT_UP) && !settings.grid.snapToGrid) {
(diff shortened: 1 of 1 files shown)


a22jacka (approved):
Tried moving elements in a dugga using the arrow keys and the page didn't scroll anymore.

I did notice that the page scrolls if you enable `snap to grid` but (for now) that's fine since you can't move elements with arrow keys anyway if `snap to grid` is enabled

a21liltr:
PR for issue #15035

Issue #15266, opened 2024-05-02
Create a retrieveResultedService microservice


f22linhe:
Some of the microservices originating from the servicefile has no output, while some, such as _updateActiveUsers_ms.php_ from the _showDuggaService.php_, have a very short output. Create the microservice _retrieveResultedService_ms.php_, similar to the ones made for _sectionedservice.php (retrieveSectionedService_ms.php)_ and _courseedservice.php (retrieveCourseedService_ms.php)_. This microservice should then be called at the end of each in the resultedService folder.

PR #15297, opened 2024-05-02
g2-2024-v5 issue #15266

c22cargo:
The MS was tested by first removing the permission checks from ```getUserAnswer_ms.php```, and then running the command:

```
curl -X POST -d "cid=2&vers=97732" http://localhost/LenaSYS/DuggaSys/microservices/resultedService/getUserAnswer_ms.php
```


a22iliru (approved):
Tested by using the curl command and it works

PR #15299, opened 2024-05-02
legacy support for sectioned.php

a22oscja:
*This is changes copied over from last week, but i did also test it to make sure it still worked

old text:

This issue uses try/catch statements to work around the database queries, then uses isset() to determine if a column is present or not in the answer received. I'd also like to add that the only relevant changes are to sectionedservice.php, the other changes are due to a merge with my previous issue branch due to the changes to navheader.php being needed to load the page.

**Contact me or the current leaders for access to the old database.**

**Testing instructions:**

- Initially, check out any courses page on the current database to ensure it still works.
- Then run the database file to get the old database (if using phpmyadmin, click your database and select import)
- Following this, go into "sessions.php" and go to the function "isSuperUser" (line 426) and add "return true;" to the top of the function. This is done due to the account we use no longer existing, hence the admin buttons would not show otherwise.
- Then simply go to the following link: http://localhost/LenaSYS/DuggaSys/sectioned.php?courseid=1&coursename=Webbutveckling%20-%20Programmering%20av%20mobila%20applikationer&coursevers=928742 or any other correctly put together link and test away.

### DIFF
--- a/DuggaSys/resultedservice.php
+++ b/DuggaSys/resultedservice.php
@@ -35,11 +35,19 @@ if(checklogin())
 
 if(isSuperUser($userid) || hasAccess($userid, $cid, 'w')){
 	// Get data to display in table rows
-	$query = $pdo->prepare("SELECT hash, password, submitted, timesSubmitted, timesAccessed, moment,last_Time_techer_visited FROM userAnswer WHERE cid=:cid AND vers=:vers");
 
-	$query->bindParam(':cid', $cid);
-	$query->bindParam(':vers', $coursevers);
+	try{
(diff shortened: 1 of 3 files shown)


a22edvbu (changes requested):
It's possible to enter and see sectioned.php, however the styling have some minor faults. There is no way to edit the moments and portions of the navbar is missing. 

image
image

a22oscja:
I believe you may not have taken the steps to make sure that you have access to editing the page, this is what it looks like when you do:
image

the buttons to edit further down also appear then:
image

a22edvbu (approved):
All looks good. Great work!

Issue #15269, opened 2024-05-02
Rect class


a22solmi:
## Description 
A rect object is used in several places in the code.

## Job
Create a rect class for easier use of the object

PR #15300, opened 2024-05-02
G1 2024 v5 #15269

a22solmi:
Added rect class


DuggaSys/diagram/draw/boxSelect.js
a21liltr, line 1: SO much easier to read! VERY nice!

DuggaSys/diagram/draw/ruler.js
a21liltr, line 1: Nice, makes the code very straightforward without unnecessary additions and quotations.

a21liltr (approved):
Well done with the changes, the code is much easier to read and the class makes it so easy to use. Seems like everything was replaced properly and I like the changes from double quotes "" to backticks ``` to appropriately make use of embedded expressions instead.

Issue #15149, opened 2024-04-30
Implement missing functions for newinstaller/tools/credentials_manager.php


b22marah:
In issue #14987 a plan for the early stages of a new installer was implemented. And in #15019 we were tasked with creating new issues for implementing the missing functions from this plan. 

In this specific issue, the missing functions in `newinstaller/tools/credentials_manager.php` are to be implemented.

Requirements. 
- It should be class based. 
- It should store user credentials, hostname and db name.
- Should be able to fetch the stored data through functions that return the previous mentioned values.

PR #15302, opened 2024-05-02
g3-2024-v5-#15149

b22marah:
## Context
The file credentials_manager is a tool class thats supposed to be used when creating new databases for user as they run through the installer. This is ment for the new installer and not the old one #15149.

## What's been done
For the following pr the class credentials_manager has been implemented to fully work. The idea is to use this when users create new databases through the installer, and functions thats included are getters / setts, initial creation of database, as well as some private functions thats running to see that everything needed exists and works properly. The class also checks for user permissions, and if permissions are not sat properly, the user doesn't get to change the file. This class also tries to find coursesys.php, and if it doesn't exists, it creates one.

## For the tester
The test should include trying out a correct usage of the class (send data that should work). And should also include seding wrong data and maybe permissions. to test this, you need to navigate to the new installer.php and create an instance of the class along with a call of the function that creates the initial database.... An image will follow on how this can be achieved (See line 7-8):

image

Note that the echo is not needed, more used for testing to see that the call on the screen says that it was created.

After this the other functions can be tested separately, like the set_db_user etc....  Have fun! :)

### DIFF
--- a/newinstaller/tools/credentials_manager.php
+++ b/newinstaller/tools/credentials_manager.php
@@ -242,8 +242,8 @@
 
             // Define a pattern to find the existing DB_HOST definition
             // Assuming the host is defined like define('DB_HOST', 'current_value');
-            $pattern = "/define\('{$name}', '.*?'\);/";
-            $replacement = "define('DB_NAME', '{$value}');";
+            $pattern = "/define\('{$name}','.*?'\);/";
+            $replacement = "define('{$name}','{$value}');";
 
             // Check if the existing host is already defined in the file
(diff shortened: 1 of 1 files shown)


a20jonda (changes requested):
The code follows PHP and K&R conventions.
Testing:
image
The local variables appear correct but the file does not match. There is no other problem found.
Solution: Remove space in regex and include the tag in the replacement otherwise all the tags will be DB_NAME
image
New Test:
image

a20jonda (approved):

Issue #15032, opened 2024-04-26
Note corner should be constant


a22jacka:
## Description
The top right corner of the note element scales with the height which causes it to look weird. It should remain constant regardless of the height of the note element

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/6f50c588-a39b-4c08-8a27-7f8606c4c56f)
![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/7bdf1cac-0ea4-4232-b899-e3441d302325)

## Job
Make the size of the corner constant

PR #15313, opened 2024-05-03
Changed the design of the note and removed the folded corner for issue#15032

a23ponek:
Changed the design and look of the note since the upper right folded corner caused trouble and was hard to make it constant and not scale with the increased height or width of the note itself. The problem was connected to how the folded corner is drawn when the note is drawn which took long time to investigate and change. 

The design of the note has no folded corner now but instead it has "lines" which still gives it the "Note" looking characteristic, in my opinion.

Link to issue:
https://github.com/HGustavs/LenaSYS/issues/15032

### DIFF
--- a/DuggaSys/diagram/draw/element.js
+++ b/DuggaSys/diagram/draw/element.js
@@ -607,7 +607,7 @@ function drawElementSequenceLoopOrAlt(element, boxw, boxh, linew, texth) {
         }
     }
     //svg for the small label in top left corner
-    content += `<path 
+    content += `<path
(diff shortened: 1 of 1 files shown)


a22cazlu (changes requested):
I like the new stile and think it works fine for notes. The only issue I can find is that the resizing don't work as it should. It resize faster than the mouse moves. Se video:

[screen-capture (3).webm](https://github.com/HGustavs/LenaSYS/assets/131054714/f481fb7f-c9b9-4343-9441-af8cfa2abb69)

a23alera (approved):

PR #15316, opened 2024-05-03
G2 2024 v5 issue#14928

a22marca:
This is a quite involved microservice to test, so make sure you check some of the comments of the issue. I have been able to test it to some extent, but not completely. This is one of those microservices that are a bit unclear if the original functionality is used, even though it is possible to kind of use it on the actual website. Another thing is that this uses the github fetching functionality, which is barely working at the moment (see issue comments) and it makes you wait 5 minutes between each test. Adding a repo to a course also downloads the files into the LenaSYS/courses/<courseid> directory, so sometimes you need to add the repo to another course if you want to perform new testing.

The createGithubCodeExample microservice should use createNewCodeExample and createNewListEntry microservices, so I had to change these a little bit to work with createGithubCodeExample. I have tested createListEntry, which already used these two, to make sure that it still works after these changes.

As I mentioned in the issue: I think group 3 are working on a few issues that will at least make the testing of this microservice easier. So if this gets merged we will have to see if this properly works (all different cases that can occur in the microservice) when it is integrated.

This was very difficult to test, so there might be things I have missed. If you find a better way to test this, please do tell me :)

### DIFF
--- a/DuggaSys/microservices/sectionedService/createGithubCodeExample_ms.php
+++ b/DuggaSys/microservices/sectionedService/createGithubCodeExample_ms.php
@@ -1,6 +1,9 @@
 <?php
 
 include_once "../sharedMicroservices/getUid_ms.php";
+include_once "../sharedMicroservices/createNewCodeExample_ms.php";
+include_once "../sharedMicroservices/createNewListEntry_ms.php";
+include_once "./retrieveSectionedService_ms.php";
 include_once "../../../Shared/sessions.php";
 include_once "../../../Shared/basic.php";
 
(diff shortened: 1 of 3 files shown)


c22cargo (approved):
I couldn't completely test this since I got an error when trying to add a PAT to a course.
I think we should merge this and revisit it in the future if needed. The whole github-functionality seems quite buggy at the moment.

The error I got:
```
Fatal error: Uncaught PDOException: SQLSTATE[23000]: Integrity constraint violation: 19 UNIQUE constraint failed: gitFiles.cid, gitFiles.fileName in C:\xampp\htdocs\LenaSYS\DuggaSys\gitfetchService.php:89 Stack trace: #0 C:\xampp\htdocs\LenaSYS\DuggaSys\gitfetchService.php(89): PDOStatement->execute() #1 C:\xampp\htdocs\LenaSYS\DuggaSys\gitfetchService.php(258): insertToMetaData('4', Array) #2 C:\xampp\htdocs\LenaSYS\DuggaSys\gitcommitService.php(106): bfs('https://api.git...', '4', 'REFRESH') #3 C:\xampp\htdocs\LenaSYS\DuggaSys\gitcommitService.php(37): insertIntoSQLite('https://github....', '4', 'ghp_BW09ZDmMOEJ...') #4 {main} thrown in C:\xampp\htdocs\LenaSYS\DuggaSys\gitfetchService.php on line 89`
```

PR #15328, opened 2024-05-03
Added options for SE lines

a22cazlu:
Added the options to SE lines so now you can chose to have arrows, dashed lines and a label. We are aware that the lines don't work as they should for SE diagrams but we think we need a new issue for that with a better description of the problem.


a22ricni (approved):
It seems to be working. Tested it by first see how it was in the week branch and then changed the branch. With the new changes I am able to have arrows and use dashed lines when a line is between SE.

Issue #15257, opened 2024-05-02
Create a retrieveFileedService microservice


f22linhe:
Some of the microservices originating from the servicefile has no output, while some, such as _updateActiveUsers_ms.php_ from the _showDuggaService.php_, have a very short output. Create the microservice _retrieveFileedService_ms.php, similar to the ones made for _sectionedservice.php (retrieveSectionedService_ms.php)_ and _courseedservice.php (retrieveCourseedService_ms.php)_. This microservice should then be called at the end of each microservice in the fileedService folder.

PR #15331, opened 2024-05-03
g2-2024-v5 issue #15257

c22cargo:
I tested the implementation by creating a temporary php file in the same folder and adding the code below, then running the file from the web. The NULL values are values that would always be null when sent from the current MS files belonging to fileedService. Note that I had to be logged into ```brom``` for this to work. Otherwise the permission checks fail.

The code is mostly copied over from the monolith, but some changes were made. Some to actually make it work, and some because the old code needed to be refactored.

Code from temporary file:
```
<?php
date_default_timezone_set("Europe/Stockholm");

include_once "retrieveFileedService_ms.php";
include_once "../../../Shared/sessions.php";
include_once "../../../Shared/basic.php";

pdoConnect();
session_start();

$debug = "NONE!"; 
$studentTeacher = null; 
$hasAccess = true;
$cid = 1885; 
$coursevers = 1337; 
$userid = 101; 
$log_uuid = null;
$info = null;

echo json_encode(retrieveFileedService($debug, $studentTeacher, $hasAccess, $pdo, $cid, $coursevers, $userid, $log_uuid, $info));
```

### DIFF
--- a/DuggaSys/microservices/fileedService/retrieveFileedService_ms.php
+++ b/DuggaSys/microservices/fileedService/retrieveFileedService_ms.php
@@ -1,0 +1,142 @@
+<?php
+
+include_once "../../../Shared/basic.php";
+
+function retrieveFileedService($debug, $studentTeacher, $hasAccess, $pdo, $cid, $coursevers, $userid, $log_uuid, $info)
+{
+    $entries = array();
+    $files = array();
+    $lfiles = array();
+    $gfiles = array();
+    $access = False;
+
+    // Fetches information from the filelink table in the database, binds column result into vars and loops through each fetched result, building a new cell for each index.
+    if (checklogin() && $hasAccess) {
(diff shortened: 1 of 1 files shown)


a22marca (changes requested):
The retrieve service works well with the included test code. So no issues there.
We should add the use of the retrieve service in the microservices for fileedService.
It also makes the implementation/testing of the retrieve service better, if you do it by testing it with a real microservice.
I suggest you add those if possible.

c22cargo:
The implementation can be tested by first going to the files in a course, then update a file and check the response message from the MS. Also delete a file and check the response message from the MS. The response message will show the response from the retrieve-service.

Change row 1086 in ```dugga.js``` to:

```
}else if(kind=="FILE"){
		if (opt=="DELFILE"){
			serviceURL="../DuggaSys/microservices/fileedService/deleteFileLink_ms.php";
		}
		if (opt=="SAVEFILE"){
			serviceURL="../DuggaSys/microservices/fileedService/updateFileLink_ms.php";
		}
		$.ajax({
			url: serviceURL,
			type: "POST",
			dataType: "json",
			success: returnedFile
		})
```

a22marca (approved):
Both microservices are using retrieveFileedService correctly.

OK to merge

Issue #15183, opened 2024-04-30
Create new test for the microservice deleteCodeExample_ms.php


f22linhe:
See parent issue: _Create new tests for the microservices created from codeviewerService.php #15182_

Note, this issue cannot be resolved until the issue _Implement retrieveCodeViewerService_ms.php into all microservices in the codeviewerService folder #15387_ has been resolved.

PR #15332, opened 2024-05-03
G2 2024 v5 #15183

a22iliru:
Need some help with this one. on test 3 i get
 Test 3 (assertEqual): failed with error: no valid values to compare
value expected: null
value output: null
Fatal error: Uncaught PDOException: SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near ':exampleid)' at line 1 in C:\xampp2\htdocs\LenaSYS\Shared\test.php:87 Stack trace: #0 C:\xampp2\htdocs\LenaSYS\Shared\test.php(87): PDOStatement->execute() #1 C:\xampp2\htdocs\LenaSYS\Shared\test.php(179): doDBQuery(Object(PDOStatement), 'UNK', 'UNK', 'UNK') #2 C:\xampp2\htdocs\LenaSYS\DuggaSys\tests\microservices\codeviewerService\deleteCodeExample_ms_test.php(34): testHandler(Array, true) #3 {main} thrown in C:\xampp2\htdocs\LenaSYS\Shared\test.php on line 87

but it seems like the code is right? can anyone test it and suggest changes?

### DIFF
--- a/DuggaSys/tests/microservices/codeviewerService/deleteCodeExample_ms_test.php
+++ b/DuggaSys/tests/microservices/codeviewerService/deleteCodeExample_ms_test.php
@@ -1,0 +1,32 @@
+<?php
+
+include_once "../../../../Shared/test.php";
+
+$testsData = array(
+    'deleteCodeExample' => array(
+        'expected-output' => '"deleted": true',
+
+        'query-after-test-1' => "INSERT INTO improw (exampleid) VALUES (:exampleid);",
+        'query-after-test-2' => "INSERT INTO box (exampleid) VALUES (:exampleid);",
+        'query-after-test-3' => "INSERT INTO impwordlist (exampleid) VALUES (:exampleid);",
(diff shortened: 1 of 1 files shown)


a22albjo:
@a22iliru the syntax with a ":" for placeholders don't work here:
```
        'query-after-test-1' => "INSERT INTO improw (exampleid) VALUES (:exampleid);",
        'query-after-test-2' => "INSERT INTO box (exampleid) VALUES (:exampleid);",
        'query-after-test-3' => "INSERT INTO impwordlist (exampleid) VALUES (:exampleid);",
        'query-after-test-4' => "INSERT INTO codeexample (exampleid) VALUES (:exampleid);",
        'query-after-test-5' => "INSERT INTO listentries (lid) VALUES (:lid);",
```

You need to use select querys and "query before test" and "?" to insert values into the querys, see resultedService/getUserAnswer_ms_test.php by @a22melst, it is an example where it is used.

Issue #15311, opened 2024-05-03
Break out error handling in addLine


a22solmi:
## Description
addLine is long and has a lot of confusing checks. It is very hard to read and understand what is happening.

## Job 
Break up this function into smaller pieces to increase readability.

PR #15333, opened 2024-05-03
Broke out error handling in addLine #15311

a22solmi:
Broke out error handling in addLine into function.
Removed unused code in addLine.
Added const array with types that handle error checking in the same way.


a21liltr (approved):
Error handling still working.

It does still high light some problems that shouldn't be highlighted but that was logic error from before, not something broken from this change. Another issue will be created to address this problem.

PR #15337, opened 2024-05-03
Made courseed.php backwards compatible with the old database G3 2024 v5 #14849

a22edvbu:
When using the old database that the live version of LenaSYS is using, it wasn't possible to load into courseed.php. This was due to missing attributes and tables when comparing the database to the newer one. The main problem was that the database tried to execute queries made for the new database on the old database, so we implemented try/catches that execute an alternate query that works for it.

!IMPORTANT!
To test this you will need to use the old database.

### DIFF
--- a/DuggaSys/courseedservice.php
+++ b/DuggaSys/courseedservice.php
@@ -18,7 +18,7 @@ include_once "../Shared/sessions.php";
 // Connect to database and start session
 pdoConnect();
 session_start();
-
+ 
 $opt = getOP('opt');
 $cid = getOP('cid');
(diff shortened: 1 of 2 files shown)


a22oscja (changes requested):
Works as intended, constructs links for sectioned.php correctly aswell.

I'll admit im a bit unsure if I should be approving this or not since the changes made to navheader.php are in the weekly branch now, meaning reasonably they shouldn't be here to avoid struggles with merging. Might not matter much but i'll request changes anyway.

merge the weekly into this branch and then should be fine.

Issue #15136, opened 2024-04-30
no keybinding on elements in toolbar


a22vioja:
## Desription

When hovering on toolbar,  there is no keybinding on Sequence Lifeline, Toggle-test cases, Reset diagram, Save Current Diagram. 

<img width="321" alt="Sequence lifeline" src="https://github.com/HGustavs/LenaSYS/assets/129263559/4bb03613-0b6c-4799-8036-fbcbe1d45043">
<img width="316" alt="Toggle test-cases" src="https://github.com/HGustavs/LenaSYS/assets/129263559/634c1738-b368-477e-8df1-69e536ca8573">
<img width="341" alt="Reset diagram" src="https://github.com/HGustavs/LenaSYS/assets/129263559/fb4ca50a-4520-4fa0-a3bd-a22c7fc17234">
<img width="312" alt="Save current diagram" src="https://github.com/HGustavs/LenaSYS/assets/129263559/e1279b0a-44cf-4273-bbc5-ebdfa79726a2">

## Job

Assign keybinds to :
Sequence Lifeline: "7"
Toggle-test cases: "E"
Reset diagram: "R"
Save Current Diagram: "S".

PR #15339, opened 2024-05-03
G1 2024 v5 #15136

a22alfna:
I fixed new keybindings for the Toggle test case and reset diagram functions since they didnt have a keybind earlier but I didnt use the specific keybinds in the issue description since the R keybind was already used and using CTRL + R is the same key combination used to reload a webpage. And E was also already used by toggle er table. So the keybinds used was "i" for the reset diagram and "u" for toggle test case. And then sequence lifeline already had 7 as the keybind I just went into the php file to change the tooltip ids for these elements to be able to display the 7 as the keybind since that wouldnt pop up before and it would just say "keybinds" on these elements. And save current diagram had the same problem which was solved in the same way.

To test this you can hover these different elements to see that the right keybind pops up next to the "keybinding:" text and then try the different following keybinds:

Reset diagram: i 
Toggle test case: u
Sequence lifeline: 7 
Save current diagram: CTRL + S

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -815,6 +815,7 @@ document.addEventListener('keyup', function (e) {
     if (isKeybindValid(e, keybinds.TOGGLE_ER_TABLE)) toggleErTable();
     if (isKeybindValid(e, keybinds.SAVE_DIAGRAM)) showSavePopout();
     if (isKeybindValid(e, keybinds.RESET_DIAGRAM)) resetDiagramAlert(); //new keybind for reseting diagram
+    if (isKeybindValid(e, keybinds.TOGGLE_TEST_CASE)) toggleTestCase(); //new keybind for Toggleing test case
 
     //if(isKeybindValid(e, keybinds.TOGGLE_ERROR_CHECK)) toggleErrorCheck(); Note that this functionality has been moved to hideErrorCheck(); because special conditions apply.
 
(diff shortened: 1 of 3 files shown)


DuggaSys/diagram.js
```diff
@@ -814,6 +814,9 @@ document.addEventListener('keyup', function (e) {
     if (isKeybindValid(e, keybinds.TOGGLE_REPLAY_MODE)) toggleReplay();
     if (isKeybindValid(e, keybinds.TOGGLE_ER_TABLE)) toggleErTable();
     if (isKeybindValid(e, keybinds.SAVE_DIAGRAM)) showSavePopout();
+    if (isKeybindValid(e, keybinds.RESET_DIAGRAM)) resetDiagramAlert(); //new keybind for reseting diagram
```
a22solmi, line 817: Remove comments, they do not add any additional information

DuggaSys/diagram.js
```diff
@@ -814,6 +814,9 @@ document.addEventListener('keyup', function (e) {
...
+    if (isKeybindValid(e, keybinds.TOGGLE_TEST_CASE)) toggleTestCase(); //new keybind for Toggleing test case
```
a22solmi, line 818: Remove comment

a22solmi (changes requested):
Keybinds work but comments in code needs to be removed.

PR #15343, opened 2024-05-03
15181 create new test for the microservice updateuser msphp

c20juler:
Created test for microservice updateuser. I could not find how to trigger the updateuser function on lenasys, therefore I could not determine the output manually. Instead I looked at the first uncommented test for accessedService_test.php which seems to work. The remaining tests were created and their output determined from the first uncommented test and by looking at relevant tables in mysql. 

I am a little unsure how to test this for the microservice but the following documentation on the test API might provide some guidance: 
Shared/Documentation/Test API/Guide on how to use Test API.md

### DIFF
--- a/DuggaSys/tests/microservices/accessedService/updateUser_ms_test.php
+++ b/DuggaSys/tests/microservices/accessedService/updateUser_ms_test.php
@@ -1,5 +1,4 @@
 <?php
-
 //------------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


c20juler:
The test does not currently work, Im working on a solution.

a22iliru (approved):
Seems fine, can revisit later if needed.

Issue #15033, opened 2024-04-26
Add input validation when fetching github codeexamples


d22mikwe:
Currently the only input validation is for empty strings/inputs. An alert is sent if a user tries to save with an empty input box:

![image](https://github.com/HGustavs/LenaSYS/assets/129263393/c84f6946-2378-4a6c-ae2c-1135a6dbae92)

Your job is to provide similar user input handling like the picture below:

![image](https://github.com/HGustavs/LenaSYS/assets/129263393/b01d4ca4-84a9-49fa-a55d-02943501a98b)

The "Name" input box should only allow letters and numbers.
The "GithubURL" input box should only allow a valid Github Repo.
The "Filtepath" input box can remain the same.
Also look into if it is possible to require that the user selects a template.
The user should not be allowed to save before the input requirements are met.

Relevant Issue that implemented these requirements:
- [Fix text validation when editing course #14343](https://github.com/HGustavs/LenaSYS/issues/14343)

Standard convention:
- [Standard convention for input validation in popup forms](https://github.com/HGustavs/LenaSYS/wiki/Popup-form-conventions)

d22mikwe:
First input is handled here right now in sectioned.js
![image](https://github.com/HGustavs/LenaSYS/assets/129263393/7e98aff8-74a8-4c6c-8034-7ab052133d8e)

PR #15345, opened 2024-05-03
G3 2024 v5 #15033

b22chran:
Implemented regex to the inputs for filename and github-link so it matches the requirements. 
Added so it visualize if an invalid input is entered and that a template needs to be chosen before saving the information.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -3708,10 +3708,37 @@ function getCourseElements() {
   }
   return list;
 }
+
+const regex = {
+	fileName: /^[A-ZÅÄÖa-zåäö\d]+( ?(- ?)?[A-ZÅÄÖa-zåäö\d]+)*$/,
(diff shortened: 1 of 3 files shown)


b22marah (changes requested):
Code looks good, but indentation and inline-css could be worked on.
image
The image above shows current code indentation, and the following will show a suggested indentation, which is more consist and clear for the future reader / developer:
image
The suggested indentation change, moved variable declaration from middle of function to the top where the rest of the variables are declared. Also spaced every if-statement so it's clearer for which if-else block belongs to who.

For the inline css we see alot of inline-css which there are currently other issue active regarding the removal of inline-css to using a css file instead. There are columns in the html, which has the same class names but still css is set for every occuring instance of this class, which should be removed and placed in the css file instead, a image below shows a example of this.
image

b22marah:
Besides previous comments, the functionality of the code works well and as intended:
image

And with correct inputs:
image

Well done! 👍

d22mikwe:
LGTM!

a22moamy:
Please avoid using "var" in code and use "let" or "const" instead.

PR #15346, opened 2024-05-03
Created retrieveDuggaedService_ms.php issue#15254

a22melst:
You can test it using any duggaed microservice, but I used createDugga. First create a new dugga using the monolith and save the response. Then add this code to createDugga_ms.php.
```
include_once("retrieveDuggaedService_ms.php");
$log_uuid=getOP('log__uuid');
$array = retrieveDuggaedService($pdo, $debug, $userid, $cid, $coursevers, $log_uuid);
echo json_encode($array);
```
And change the code at the else-if for kind "DUGGA" in dugga.js.
```
}else if(kind=="DUGGA"){
	let serviceurl = "duggaedservice.php";
	
	if (opt=="SAVDUGGA")
		serviceurl = "../DuggaSys/microservices/duggaedService/createDugga_ms.php"

	$.ajax({
		url: serviceurl,
		type: "POST",
		data: "opt="+opt+para,
		dataType: "json",
		success: returnedDugga
	});
```
Then create a new dugga and compare the responses.

### DIFF
--- a/DuggaSys/microservices/duggaedService/createDuggaVariant_ms.php
+++ b/DuggaSys/microservices/duggaedService/createDuggaVariant_ms.php
@@ -40,4 +40,8 @@ if (!$stmt->execute()) {
     $debug = "Error updating entries" . $error[2];
 }
 
-echo json_encode($debug);
+include_once("retrieveDuggaedService_ms.php");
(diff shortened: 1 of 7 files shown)


a22iliru:
Now it looks good. merged

PR #15349, opened 2024-05-03
G2 2024 v5 issue#15310

b21isabe:
The retrieveProfileService_ms.php file has been created and implemented.

The profileService microservices only output basic status messages on the operations made. I have made a super simple retrieve-function that can be expanded in the future if more output is required.

To test the microservices change the urls in profile.js
line 33:
```"../DuggaSys/microservices/profileService/updateSecurityQuestion_ms.php"```

line 166:
```url: "../DuggaSys/microservices/profileService/updateUserPassword_ms.php",```

Then login and press the username in the top right corner (requires https connection). Here you can attempt to change the password and security question. This should send the output to the network tab in your browser.

### DIFF
--- a/DuggaSys/microservices/profileService/retrieveProfileService_ms.php
+++ b/DuggaSys/microservices/profileService/retrieveProfileService_ms.php
@@ -1,0 +1,10 @@
+<?php
+
+function retrieveProfileService($debug, $success, $status){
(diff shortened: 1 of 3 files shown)


a22iliru (approved):
A POST was sent in network tab and password changes. it works

PR #15381, opened 2024-05-06
getdatUserAnswer Update for showDuggaservice #14503

a22fokol:
The getdataUserAnswer has now been updated hope everything works well as it should


a23oskme (commented):
Test 1 (login) is OK
Test 2 (callService) is OK
Test 3 (assertEqual) fails with the error "no valid values to compare"

Issue #15383, opened 2024-05-06
ER entiy and relation has no cardinality option


a22solmi:
## Description
When connecting an ER entity and relation cardinality cannot be set.

## Job
Update so that the line works in the same way as between entity and attribute

PR #15384, opened 2024-05-06
Removed break for non attributes #15383

a22solmi:
Removed break for non attributes


a23ponek (approved):
The line properties between a ER entity and relation now works the same way as between entity and attribute, se images below. No problems seem to arise.
Screenshot 2024-05-06 114620
Screenshot 2024-05-06 114718

PR #15389, opened 2024-05-06
Added check for max length under 1 #14322

a22solmi:
Added check for max length under 1


a22vioja:
Closes issue #14322

b22davbe (approved):
Looks good and solves the problem.

Issue #15390, opened 2024-05-06
New errors whenever entering a course


a22moamy:
Now, when entering a course, you get these errors:
![Skärmavbild 2024-05-06 kl  10 28 21](https://github.com/HGustavs/LenaSYS/assets/129263915/2bf7f66f-f012-4bc3-be2b-6b76bd24bf54)
![Skärmavbild 2024-05-06 kl  10 28 25](https://github.com/HGustavs/LenaSYS/assets/129263915/80befdcd-8b1b-489a-b8b6-06c48d3f685b)

**Please find the problem to this and fix it.**

PR #15394, opened 2024-05-06
Moved the variable so that its declared. #15390

a22edvbu:
The variable was declared in an unreachable else state. Moved it so that its declared. The error should be gone now for Mac users when opening courseed.php.

### DIFF
--- a/Shared/navheader.php
+++ b/Shared/navheader.php
@@ -65,10 +65,9 @@
 					$checkIfGithubURL = $row['courseGitURL'];
 				} else {
 					$checkIfGithubURL = null;
+					// $row is false, which means no data was fetched
+					$updateTime = "No data found for the given course ID";
 				}
-			} else {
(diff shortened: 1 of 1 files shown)


a22natth (approved):
## Testing 
During the testing, a comparison was done on the weekly branch and the branch for this pull request. Previously (and on the weekly branch), when entering a course the user was presented with the alerts below:
Pasted Graphic
Pasted Graphic 1
On the branch of the pull request, the previous messages did not display and regular functionality seems to be achieved. 

## Problems 
Although, when creating a new course, a 500 Internal server error is presented and the new course does not appear until the page is refreshed. This does not seem to be caused by the pull request, but rather a previous merge. Following this, when entering the new course, the alerts (as seen above) will be presented. The same issue is presented on the weekly branch where a user is unable to correctly create a course.

Issue #15037, opened 2024-04-26
Enable arrow keys on snapToGrid


a22jacka:
## Description
Currently there's a check that disables the arrow keys if the option "snap to grid" is enabled.

## Job
Implement so that the arrows use the grid size if the setting is enabled otherwise just keep moving 1px.

b22samer:
b22samer and a22robko enhanced functionality to enable the movement of elements using arrow keys while snap to grid is enabled. Previously, this feature was not working as intended. One of the issues was with the selection mechanism responsible for arrow key movements, which did not account for snap to grid. Regardless of the zoom level, elements will now consistently align precisely with the grid lines, ensuring a seamless experience.

PR #15395, opened 2024-05-06
fixed enable arrow keys on snapToGrid #15037

a22robko:
b22samer and a22robko enhanced functionality to enable the movement of elements using arrow keys while snap to grid is enabled. Previously, this feature was not working as intended. One of the issues was with the selection mechanism responsible for arrow key movements, which did not account for snap to grid. Regardless of the zoom level, elements will now consistently align precisely with the grid lines.


a22jacka (changes requested):
The arrow keys work as expected with aligning everything to the grid. I did however notice that the ghost is sometimes misaligned with the grid, although it can be fixed by placing the element and then moving it. The ghost should be properly aligned tho

bild

bild

https://github.com/HGustavs/LenaSYS/assets/129258593/007efad3-c0ef-43c9-b713-4b6565273bb9

PR #15396, opened 2024-05-06
removed "Login is only possible from course page!" and check login works

SleepyBearIV:
checked if you can log do a log-in form anywhere in the system and that works for me and removed the text, i am on an docker system and would probably be good to test this on an xampp system as well.

to tester:
try and log in and out at any place you can find and report back if a place dose not work; if this massage comes up:
image
i have confirmed that this is a separate issue in the system now an id being looked at, but you should still be able to log in.


a22willi:
The text is removed everywhere. Looks good to me.

a22willi (approved):
The text is removed everywhere. Looks good to me.

PR #15397, opened 2024-05-06
Implemented first draft of installer

a22willi:
Now incorporates all the requirements of the issue:
- SSE receiver
- SSE sender
- And a coarse version of the installer for testing. (this will be replaced by a proper implementation of UI).
- Also fixed some bugs in other newinstaller files.

### How to test?
Navigate to `localhost/LenaSYS/newinstaller/installer.php`

### DIFF
--- a/newinstaller/installer.php
+++ b/newinstaller/installer.php
@@ -1,6 +1,9 @@
 <?php 
-    include_once("installer_ui.php");
-    foreach (glob("tools/*.php") as $tools) {
-        include_once $tools;
+    if (isset($_GET['stream'])) {
+        // Script that runs the installer
+        // Sends installation progress messages at: installer.php?stream=true
+        include_once("tools/install_engine.php");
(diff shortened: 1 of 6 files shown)


a22danwe (approved):
bild
It seems to be working. i cant figure out how to use phpmyadmin so cant check if everything actually works as intended but the install works and it gives the correct output.

a22natth (approved):
When testing the installer, the installation was successful when inserting the right credentials in the install_engine.php file. The installer creates a database with the name “LenaSYStest”, which when successfully created can be seen in phpmyadmin. 
Pasted Graphic 4

The database can be seen in the example below which shows some of the tables that were created. 
Pasted Graphic 2

For instance, the table “course” contains the right columns as seen below. 
Pasted Graphic 3

The table itself also has the correct inserts. 
Pasted Graphic 5

PR #15398, opened 2024-05-06
Ghosts now follow mouse centered.

b22davbe:
Due to the fact that some elements (like UML-entity, IE-entity e) is created in a different way than the "normal" way we needed to make different calculations for those cases. The IE-relation didn't center because the program treated it like a square. We changed that, solving the issue with it not being in the center of the mouse while also improving the element with better properties resulting in no more false errors where the program before thought it were overlapping when it wasn't.


a22solmi (changes requested):
This is the closest I can place anything under UML, IE and SD.
Screenshot from 2024-05-06 16-33-03

Issue #15263, opened 2024-05-02
ER example digram dot


a22solmi:
## Description 
The ER example diagram has a dot in the top right corner that causes the window resize to center the camera wrong.
<img width="962" alt="Screenshot 2024-05-02 at 11 34 33" src="https://github.com/HGustavs/LenaSYS/assets/129263559/3c63c0ec-72ca-414c-8388-9776dabcd21e">

## Job 
Remove the dot.

PR #15401, opened 2024-05-06
Remove object with ID BA4BCD  #15263

c22abdha:
Remove object with ID BA4BCD in JSON file.

{
            "kind": "ERAttr",
            "name": "Ghost",
            "stroke": [
                "#383737"
            ],
            "width": 5,
            "height": 5,
            "x": 1805.5,
            "y": -37.5,
            "id": "BA4BCD"
        },

Closes issue #15263


a21liltr (approved):
When first loading in the example ER diagram, there's a dot at the far right (see image)

image

After the change, it is removed properly as seen in the image below:
image

I like the reformatted JSON, It is easier to read and I have also validated it with JSONlint, so everything is OK.

PR #15402, opened 2024-05-06
Pull request for G3-2024-v5-#15320

### LINKED ISSUE #15320: Implement missing functions for newinstaller/tools/testdata_setup
In issue https://github.com/HGustavs/LenaSYS/issues/14987 a plan for the early stages of a new installer was implemented. And in https://github.com/HGustavs/LenaSYS/issues/15019 we were tasked with creating new issues for implementing the missing functions from this plan.

Requirements:
- function to add language support.
- function to add test course.
- function to add demo course.
- function to add testdata.

b22marah:
# NOTE
See previous comments for this issue before testing. The requirements has come to change....

Whats has been implemented is a class that takes sourcePath, destinationPath and callback function.
There are two public functions that are used, these are copy_test_course, which effectively copies all available test-courses to the correct folder when choosen to, through the installer. The second function is copy_course, which copies specific courses.

## For the tester
When testing this you can use the following image as a guide to setup you're testing.
image

Try switch between existing courses when trying the copy_course function (available courses are '1', '2', 'global'). Also play around with no / true / false for the verbose boolean (True gives better feedback, false / empty provides little to none feedback).

closes #15320

Have fun! :)

### DIFF
--- a/newinstaller/tools/testdata_setup.php
+++ b/newinstaller/tools/testdata_setup.php
@@ -1,44 +1,77 @@
 <?php
 class TestdataSetup {
+	private $sourceDirectory;
+	private $destinationDirectory;
 	private $callback;
 	
-	public function __construct($callback = null) {
+	public function __construct(string $sourceDirectory, string $destinationDirectory, $callback = null) {
+		$this->sourceDirectory = $sourceDirectory;
+		$this->destinationDirectory = $destinationDirectory;
 		$this->callback = $callback;
 	}
 
     /**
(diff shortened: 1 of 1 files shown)


a22maxra (changes requested):
image

image

Copying courses works but there are now errors when trying to edit courses for some reason

a22maxra:
Since errors and unrelated to the coded added in the new installer they are probably because of some previous merge so request i accepted

a22maxra (approved):

d22mikwe:
LGTM

PR #15403, opened 2024-05-06
Small fix to getUserAnswer_ms.php issue#14500

a22melst:
Tested by looking at the resulteditor of a course before and after implementing the changes. Before nothing could be seen, but after the changes, it works.


c22cargo (approved):
Works

Issue #15097, opened 2024-04-29
Removed list entries come back after refreshing the page


a22moamy:
After removing a list entry and refreshing the page, the list entry comes back. 

How to find this issue:
1. Navigate to a course with content
2. Press the trash icon on a list entry, like so:
![Skärmavbild 2024-04-29 kl  14 48 44](https://github.com/HGustavs/LenaSYS/assets/129263915/6242ab95-63fc-409f-ae06-609b3fc8cf7f)
3. Refresh the page through the web browser and the removed list entry will be back.

Fix this issue and add a toast instead, asking if the user wants to undo their actions. A similar implementation of the toast can be seen in #14945.

d22mikwe:
Undo toast has been added and merged in #14945 and works in current v4 bransch.
Also i tested on Windows 11 Microsoft Edge and Chrome. Works for me

d22isaka:
It seems the deltimers value was initially set for 60 seconds blocking the delete, which was way longer than the toast.
If refreshing the page before the sleep was done, the delete action did not complete.

Changing the value for delTimer to match the undo toast makes it more natural. Now it is set to 5 seconds. Refreshing the page before the toast is done will bring back the list entry otherwise it will be deleted.

Created a variable to set both the toast and sleep to have the same timevalue.
sectioned.js row 1011

![Skärmbild 2024-05-06 132708](https://github.com/HGustavs/LenaSYS/assets/129263107/67a53f00-7ade-4330-9021-ffd406ab7388)

PR #15409, opened 2024-05-06
G3 2024 v5 #15097

d22isaka:
Entries will now be deleted after the undo-toast is done appearing.


a22maxra (changes requested):
image

Despite choosing _JS_TEST template 1_ in this picture it does not get deleted

PR #15410, opened 2024-05-06
G2 2024 v5 issue#15358

b21isabe:
Made the test-file updateUserPassword_ms_test.php.

The first test tries to change the password of a user which should work.
The second test tries to change a teachers password, this should not work and return: ```{"success":false,"status":"teacher","debug":"NONE!"}``` since teachers are not allowed to change their own passwords.


a21sahfa (approved):
den funkar

Issue #15159, opened 2024-04-30
Improve highlight function when creating new list item.


d22mikwe:
Improve the highlight to make it more intuitive and noticable

SleepyBearIV:
after reconsideration it is better perhaps to make the whole row highlighted, continuation after weekly merge

SleepyBearIV:
when you ad a new course element (code) the whole row should be highlighted  
![image](https://github.com/HGustavs/LenaSYS/assets/96497530/7daa3e35-cdd4-40a0-8a29-e352c6fdc5fc)


current animation function can be found here 
![image](https://github.com/HGustavs/LenaSYS/assets/96497530/2f0dff87-1b3e-404e-af1a-562898c27169)

PR #15412, opened 2024-05-06
New branch for week 4 issue #15159

a18chrkr:
Has fixed the highlight to light up the whole table row instead of just the text. also created a function in order to not use repeating code. This could need a double check/test in order to confirm functionality.


d22isaka (approved):
Highlight looks good, fully highlighting the row and fading away. Functionality of the new function seems to be ok, not sure how to test it aditionally. 

I've tried on Firefox and Edge and it works on both.

PR #15413, opened 2024-05-06
legacy support for showDugga.php

a22oscja:
showdugga seemed to work correctly for the most part, I have tested by inputting data needed into the database. As I noticed that all showdugga pages use showdoc.php instead if you do not have access to editing the page. This did not work but had support implemented already, so it was an easy fix.

From what I could tell from looking around everything should work now. However, it is not possible to actually get the current system to display anything (as the files/data needed is not included in the data we were given). Because of this, the testing becomes quite rudimentary. All in all i'd say out of all the pages, this one is the most likely to break with proper files/data, but I believe there is no way to get around this from our side.

**Testing:**

- get access to old database from me/leaders/anyone who has it.
- go to a course page, then to any dugga.
- mess around all you'd like, there really is not much to see.

**Switching between access and no access:**

- go to sessions.php
- on line 426 lies the function isSuperUser(), make this return true upon being called (add return true; to the top of it)

### DIFF
--- a/DuggaSys/showdoc.php
+++ b/DuggaSys/showdoc.php
@@ -1,5 +1,5 @@
 <?php
-	ini_set("auto_detect_line_endings", true);
+	//ini_set("auto_detect_line_endings", true);
 	include_once "../Shared/basic.php";
 	include_once "../Shared/sessions.php";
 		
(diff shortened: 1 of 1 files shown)


a22edvbu (approved):
Looks good to me!

PR #15414, opened 2024-05-06
Changed .toastText overflow-y from scroll to auto.

a22edvbu:
Having overflow-y set to auto enables it to switch to a scroll-wheel whenever the text overflows in the y-direction. If the text does not overflow the toast, it will no longer show a scroll-wheel. Whenever the text overflows the toast, that's when the scroll-wheel appears.

The easiest way to test this is to go into toast.php and extend the timers for the toasts and then edit the content inside.


d22isaka (approved):
YupYup, the scrollbar only appears for me when the toast if overflowed. 
Tested this by editing a particular toast with loads of teeext.

PR #15415, opened 2024-05-06
Create microservice updateCodeExampleTemplate issue#14637

a22eriwi:
Created the ms file that updates the template for a certain code example.

### DIFF
--- a/DuggaSys/microservices/codeviewerService/updateCodeExampleTemplate_ms.php
+++ b/DuggaSys/microservices/codeviewerService/updateCodeExampleTemplate_ms.php
@@ -12,6 +12,8 @@ $userid = getUid();
 $opt=getOP('opt');
 $templateNumber=getOP('templateno');
 
+if($exampleCount>0){
+
 if(checklogin() && ($writeAccess=="w" || isSuperUser($_SESSION['uid']))) {
     $writeAccess="w"; // TODO: Redundant? Is set a couple of rows above
 
@@ -259,4 +261,5 @@ if(checklogin() && ($writeAccess=="w" || isSuperUser($_SESSION['uid']))) {
         echo (json_encode(array('deleted' => true, 'debug' => $debug)));
         return;
     }
+}
 }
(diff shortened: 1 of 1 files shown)


c22cargo (changes requested):
The microservice works, but some changes should be made.

Add a comment at the top to indicate what the microservice is used for.

You are calling ```getUid();``` but not assigning the return value to anything. ```isSuperUser($_SESSION['uid'])``` should use the uid returned from ```getUid();```.

The variable ```$writeAccess``` is used in an if-check, but is never assigned a value.

The code could use some blank lines to make it easier to read. Currently everything is clumped together. For example, line 26 could be a blank line. The original code from the monolith has _okay_ code structure that you could just copy.

The if-statements below should perhaps be a switch-statement instead, but it might make it less concise in this case. I'll let you decide on what to use. If you decide to keep it as if-statements you should use ```else if``` instead of different if-statements.

```
if($templateNumber==10) $boxCount=1;
if($templateNumber==1||$templateNumber==2) $boxCount=2;
if($templateNumber==3||$templateNumber==4 ||$templateNumber==8) $boxCount=3;
if($templateNumber==5||$templateNumber==6 ||$templateNumber==7) $boxCount=4;
if($templateNumber==9) $boxCount=5;
```

c22cargo (changes requested):
There should be an ```if``` first, currently there are only ```else if``` statements.

Change it to:
```
if($templateNumber==10) $boxCount=1;
else if($templateNumber==1||$templateNumber==2) $boxCount=2;
else if($templateNumber==3||$templateNumber==4 ||$templateNumber==8) $boxCount=3;
else if($templateNumber==5||$templateNumber==6 ||$templateNumber==7) $boxCount=4;
else if($templateNumber==9) $boxCount=5;
```

You forgot to address the ```getUid()``` issue from my previous comment.

Simply do:
```
$userid = getUid();
```

This should use the already retrieved value instead of calling the function again:
```
isSuperUser(getuid())
```

There are simpler and cleaner ways of doing the permission checks, but since it works it's fine.

Issue #15174, opened 2024-04-30
IE and UML Class Entities can't connect with themselves but should be able to.


a21liltr:
IE and UML Class Entities can't connect with themselves but should be able to.

Example: (illustrated by video below)
There can only be max 1 connecting between an ER Relation and an ER Entity however 2 are needed for an ER Entity to make a connection with itself.

https://github.com/HGustavs/LenaSYS/assets/129370135/c6b538cf-c04f-4c1f-985f-b2043f771ab2

Make an implementation that will solve this problem.

PR #15421, opened 2024-05-06
Added recursion to UML IE #15174

a22solmi:
Added recursion to UML IE


a22ricni (changes requested):
Can't see two lines when ER entity with ER relation only on because they are at the same place and looks to be one line. Also can't find what is change with the UML entities or if it was something that was going to change.

a22ricni (approved):
Tested the new commit and is work as indented. I also don't see any issue when resize the entity. 
image

PR #15424, opened 2024-05-06
getdataUseAnswer lasted updated version  

a22fokol:
I have updated the Url of the by changing showDuggaservice.php wiht getdataUserAnswer.php.  Changed also     //     'query-after-test-1' => "DELETE FROM userAnswer ORDER BY aid DESC LIMIT 1;",      with              'query-after-test-1' => "DELETE FROM userAnswer WHERE hash = 'ghj1jfg2';",

### DIFF
--- a/DuggaSys/tests/getdataUserAnswer_test.php
+++ b/DuggaSys/tests/getdataUserAnswer_test.php
@@ -1,0 +1,36 @@
+<?php
+
+
+    include "../../Shared/test.php";
+
+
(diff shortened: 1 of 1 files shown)


a22iliru:
@a22fokol What is the issue related to this?

a22iliru (approved):
The last test fails, however the structure of the code looks good.
probably only the output needs to be fixed.
Can merge

Issue #14623, opened 2024-04-17
Move and Resize jump


a22solmi:
## Description
Resizing an entity by dragging the "left" or "up" node will cause it to jump a few pixels. This only happens once when you move the mouse when dragging the node. The moving and resizing itself works as expected.

## Job
Find out why the entity is moved and implement a solution.
Most likely caused by "StateChangeFactory.ElementMovedAndResized" or "screenToDiagramCoordinates"

a22hanfa:
Okay, I'm pretty sure I've figured out what is causing the jump.

This line in the`screenToDiagramCoordinates`-function. But I'm kind of stuck on how to redo this function.
```
return new Point(
        Math.round((mouseX / zoomfact - scrollx) + zoomX * scrollx + 2 + zoomOrigo.x), // the 2 makes mouse hover over container
        Math.round((mouseY / zoomfact - scrolly) + zoomX * scrolly + zoomOrigo.y)
    );
```

a22hanfa:
# Overview of the issue
I think I have found the where the issue stems from. The function ``screenToDiagramsCoordinates`` is intended to convert screen coordinates to diagram coordinates. This function uses a lot of different variables like zoom, scroll, and an origin adjustment. However, I believe the function's implementation has led to significant and unexpected movements of elements, as evidenced by various tests and adjustments tried.

***All test** where done on the Y-coordinate but it should apply the same for the X-coordinates, since it's outputting the same behavior.*

# Where does the errors occur?
I'm pretty sure that the error occurs in the ``case pointerStates.CLICKED_NODE`` in `mmoving`-function. It's hard to know what causes what because there are so many variables in this.

My research has concluded that the error happens here:

    tmp = elementData.y;
    elementData.y = screenToDiagramCoordinates(0, (startY - deltaY + 14)).y;

    // Deduct the new position, giving us the total change
    const yChange = -(tmp - elementData.y);

I also believe that the code here is the reason for all the errors:
This return statement in thescreenToDiagramCoordinates-function. 
```
return new Point(
    Math.round((mouseX / zoomfact - scrollx) + zoomX * scrollx + 2 + zoomOrigo.x), // the 2 makes mouse hover over container
    Math.round((mouseY / zoomfact - scrolly) + zoomX * scrolly + zoomOrigo.y)
    );
```
I've found that the elementData.y is different before vs after the `screenToDiagramCoordinates`-function. Here is an example of what happened once:
[...]

PR #15429, opened 2024-05-06
Fixed move and resize jump. (#14623)

b22chrmo:
Fixed issue where an element would make a big jump when moving or resizing it. 

There is still a very small jump when resizing using the left node and the upper node (also affects the upper left and lower left nodes), but this seems related to the 'startX' and 'startY' values.

Also removed the 'preResizeHeight' code from the upper node (This code is not used anymore from what I can see).

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -972,7 +972,6 @@ function mmoving(event) {
             calculateDeltaExceeded();
             break;
         case pointerState.CLICKED_LINE:
-
             if (mouseMode == mouseModes.BOX_SELECTION) {
                 calculateDeltaExceeded();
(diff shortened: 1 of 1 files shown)


a22vioja:
Closed issue: #14623

a22hanfa (changes requested):
This works well on the x-axis but when it comes to the y-axis I still experience crazy jumps.

**See:**

https://github.com/HGustavs/LenaSYS/assets/129273186/f2c2833b-ea14-4ce8-bc25-eefbc16e7a63

a22hanfa (approved):
This works as intended on Windows-machines. 
However, the issue still exists on macOS and Linux.

Issue #15423, opened 2024-05-06
When toast has "0" duration, it should not have a timer


a22moamy:
Implement toasts without a duration, which should be created by entering "0" as the duration. 
When a toast has "0" duration, it should not have a timer at all. This should be useful for toasts with longer messages where you do not want them to fade away. 

So, if I call a toast like this, then it should not fade away and should only dissappear when the X-button is pressed:
`toast("success", "Hello world!", 0);`

See [the Toast wiki](https://github.com/HGustavs/LenaSYS/wiki/Toast) on how to implement and use toasts.

PR #15431, opened 2024-05-06
Duration set to "0" will now not be given a timeout G3 2024 v5 #15423

a22edvbu:
When calling the toast() function and assigning duration to 0 or null, it will now not be given a timeout. In other words, the user has to manually close the toast.

To test this i recommend changing the toast on row 1011 in sectioned.js. This toast will pop up when you delete an item.


b22marah (approved):
LGTM!

d22mikwe:
LGTM!

Issue #14940, opened 2024-04-25
Fix alignment of course settings modals


a22willi:
When editing group activities/code examples/tests etc, the input boxes are not properly aligned. Fix this.
![image](https://github.com/HGustavs/LenaSYS/assets/129264968/b5af7a70-45b0-43af-89a5-4d9babe68935)

Create item:
1. Navigate to a course.
2. Hover over the floating plus button in the bottom right.
3. Select an item
4. Click the settings icon for the newly created item.

a22chrfa:
In this case. The name has a set pixel width of 440, unlike the other elements. This is probably a bandaid for some previous fix. Most likely not a good solution as the consistency with other elements might be compromised, as in this case.

For the last element. The select element has a margin of 10 px, which is inherited by all select elements in except for the last one. Where there is a higher priority (ID) styling attached, which sets the margin to 0. This is likely also a bandaid fix.

Fixing these issues will not take much time, however. Removing these might lead to new issues with alignments as these bandaid fixes (although not a good solution) might have fixed something somewhere else.

a22chrfa:
Removed the specific pixel width for the inputwrapper-name ID. Removed the specific higher priority style for the last element.

This cause new issues to arise, as I feared. There were two new input fields that was now shifted. After some digging I found that these elements now inherited the correct styles, but as they probably had issues before this. It seems the span element covering those elements had a wrapping margin for those elements. Which has probably been a quick fix to solve the alignment for those elements, instead of fixing the source of the problem.

PR #15432, opened 2024-05-06
Fixed aligments on input and select elements (issue: #14940)

a22chrfa:
There were specific styles being prioritized on certain elements. Which has probably been quick fixes. When these were solved, new issues with alignment appeared in other places. Most likely due to other quick fixes. There was a span element covering two elements which had a margin covering them both. Causing them to visually have double margin values when these elements themselves inherited the correct margins.

Hopefully, whenever a new element is added in the future, they should now inherit the correct styles. However, there might potentially be some places on the website where new issues arise. I can't find any for now, but they might pop up in the future unless a thorough search is conducted.

For testing, just check the alignments on all input elements you can find. For the specific elements in the issue, it's been solved.

image

### DIFF
--- a/DuggaSys/sectioned.php
+++ b/DuggaSys/sectioned.php
@@ -299,7 +299,7 @@
 					<div id='inputwrapper-deadline' class='inputwrapper'>
 							<legend><h3>Deadline</h3></legend>
 							<span>Absolute</span>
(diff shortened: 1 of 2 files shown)


a22edvbu (approved):
The text fields are all aligned now. Great work!

Issue #14946, opened 2024-04-25
Change order of code example settings icons


a22willi:
![image](https://github.com/HGustavs/LenaSYS/assets/129264968/0f33b532-d661-4c33-bf4e-cf07960a7230)
As seen in the image above, code examples have the buttons in a different order than everything else, make sure you change the code example rows to match the others.

PR #15435, opened 2024-05-06
Changed order of buttons to make them align (from issue #14946)

a22chrfa:
Moved the github buttons to the front of the row. Since this buttons seems to be one of the few that doesn't always appear on a row, they would shift the entire row by one step, therefore making it asymmetrical with other items.

Now it looks like this:
image

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -1844,6 +1844,25 @@ function returnedSection(data) {
           }
         }
 
+        // github icon for moments (itemKind 4 is moments)
+        if (itemKind === 4 && data['writeaccess'] || data['studentteacher']) {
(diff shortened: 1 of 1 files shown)


a22ahmom (changes requested):
When entering/leaving courses I get these errors and have to click 3 times to remove them. This happens with all courses and should be fixed before merging.

Screenshot (408)
Screenshot (409)
Screenshot (410)

a22ahmom (approved):
The issue seems to be solved. LGTM!

a22chrfa:
To clarify and document. The issue is related to navheader.php (which hasn't been changed here). This issue seems to originate in the new weekly branch. While investigating this, I noticed via git blame that it had been changed prior to the weekly merge, and that an issue already is dealing with this. It is therefore expected to be solved elsewhere, and this pull request should be able to ignore the issue.

Issue #15084, opened 2024-04-29
Size difference


a22solmi:
## Description
There is a large size difference between IE Inheritance, ER Attribute and everything else
<img width="418" alt="Screenshot 2024-05-07 at 08 27 16" src="https://github.com/HGustavs/LenaSYS/assets/129263559/ab492774-ac25-4c4c-be06-c992e11c2ed5">

## Job
Change the dimensions to look similar to other elements

PR #15443, opened 2024-05-07
Balanced the size of the different element for issue#15084

a23ponek:
Changed the size of the boxw and boxh arguments for the different drawElement function calls in element.js file. The changed calls were for drawElementERAttr, drawElementIERelation and drawElementUMLRelation on lines 48-62.

Link to issue:
https://github.com/HGustavs/LenaSYS/issues/15084


a22solmi (changes requested):
You have +600 and -400 lines but the issue is only to change default values. You have some problem with your git versioning.

PR #15444, opened 2024-05-07
Added async false

d22mikwe:
Added async false to avoid xhr error in courseedervice

In previous version when copying new course version or editing version there is an xhr error in the network traffic caused by asynchronous calls being made.

image

With these changes a response can be retrieved


b22marah (approved):
LGTM!

PR #15447, opened 2024-05-07
Created the test - issue#15350

a22melst:
To test this you have to change retrieveResultedService.php and getUserAnswer_ms.php to their newest versions which are found in the group branch. You should also use a freshly installed database. Then you can integrate the microservice in dugga.js and run the test. I am aware the output is large but that is needed in order to show all the changes made by the queries before the test. The opt is sent to the service even though it is not checked for there, since I think that is something that should be implemented in the future.


a22marca (approved):
Everything works as it should when following the steps in the PR.

OK to merge

PR #15449, opened 2024-05-07
G2 2024 v5 issue#15304

a22marca:
I have added the use of the retrieveCourseedService to all the microservices in microservices/courseedService. 
Some of the microservices could not be tested properly:

changeActiveCourseVersion - done/untested - original opt "CHGVRS" not used, no place to change active version specifically on the website
deleteCourseMaterial - done/untested - retrieveCourseedService already performs this functionality, might be a unnecessary microservice.
createCourseVersion - done/TESTED
updateCourseVersion - done/untested - currently does the same thing as updateCourse, needs to be changed
createNewCourse - done/TESTED
updateCourse - done/TESTED
createMOTD - done/TESTED

For testing and integration see the issue comments.


c22cargo (changes requested):
Regarding ```retrieveCourseedService_ms.php``` 

* This should be able to be an MS call:

```
if(isset($_SESSION['uid'])){
        $userid=$_SESSION['uid'];
    } else{
        $userid="UNK";
    }
```

* Is there a reason for one include to be at the top before the actual function, while another include is at the start of the function? If there is no reason then they should probably be in the same place.

* The comment ```//echo json_encode($array);``` can be removed, it's at the bottom of the file.

* In the logging the variables ```$log_uuid``` and ```$info``` are used, but they are never assigned a value anywhere.

```
logServiceEvent($log_uuid, EventTypes::ServiceServerEnd, "retrieveCourseedService_ms.php",$userid,$info);
```

Regarding the other microservices

* They seem to have a few issues. However, fixing all of them I feel is outside the scope of this issue. But as an example, they forget to use the ```getUid``` function and instead use:

```
if (checklogin()) {
	if (isset($_SESSION['uid'])) {
		$userid = $_SESSION['uid'];
	} else {
		$userid = "UNK";
	}
```

* The integration of the microservices in ```dugga.js``` should use ```switch``` instead of ```if```.

* I couldn't get ```updateCourse_ms.php``` to work. I assume it's used when editing a course from the home page, but maybe I am mistaken. It just kept loading for me.

The microservices that I was able to test seem to be using the retrieve MS correctly.

a22marca:
I addressed the points in the change request. 

updateCourse_ms.php works on my end. The only thing is that i need to refresh the page to see the actual update of the course.

c22cargo (changes requested):
The logging at the top uses ```courseedservice.php```, which is incorrect:
```
logServiceEvent($log_uuid, EventTypes::ServiceServerStart, "courseedservice.php", $userid, $info);
```

Using ```getOP``` is a good idea if it works correctly. I don't know if it can bug out in certain scenarios. I have seen that some retrieve MS instead receive these as parameters, but this approach is simpler. I checked the values of ```$opt``` and ```log_uuid```, and they seem to be correct. This probably needs more testing, but I don't have time today and the merge is tomorrow. In other words, this looks good.

```
$opt = getOP('opt');
$cid = getOP('cid');
$coursename = getOP('coursename');
$visibility = getOP('visib');
$versid = getOP('versid');
$courseGitURL = getOP('courseGitURL');
$log_uuid = getOP('log_uuid');
```

The logging doesn't seem to work for me. This might just be on my end though. If it works for you then it's fine.

a22marca:
Yes there are some issues with the logging, but this is not only happening in this microservices. Lets fix that in a separate issue.

I've changed the filename

PR #15450, opened 2024-05-07
Internal server error when making new course

a22oscja:
It was attempting to insert data into the database twice, leading to the error. I also added support for the old data. As a result of this fix, the course should also now show up without reloading the page, as it was getting stopped from making that happen before.

**testing:**
if you do not have access to old data, get it from me/leaders/anyone really
make a new course, check for errors.
repeat on current/old data as much as you'd like.

### DIFF
--- a/DuggaSys/courseedservice.php
+++ b/DuggaSys/courseedservice.php
@@ -95,7 +95,6 @@ if (checklogin()) {
 				$query->execute();
 			}
 			catch(Exception $e){
-				$userid = 1102;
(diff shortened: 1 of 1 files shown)


a22danwe (commented):
On the current database im not getting any errors when creating a course. and on the old database i cant seem to create one at all, when i try this appears 
bild. the toast says a course is created but it doesnt show up in the database or on the page. ill look more into it tomorrow

PR #15452, opened 2024-05-07
Scale sequence actor on the height aswell #14857

a22robko:
The issue is fixed with being able to zoom out on the sequence lifeline, but there appears to be a bit of bugging when you drag the element up. We have been trying to fix it so that it doesn't jump, but what we've achieved is the best we could do. We went back to the weekly branch and noticed the elements were jumping too, so we're not sure if it was like this before we started coding or if it's something we caused.


a22vioja:
Closes issue #14857

Issue #15386, opened 2024-05-06
ER attribute element can be used as ER relation


a22vioja:
## Description

It is possible to connect  ER Attribute in between of entities and it is wrong.

<img width="677" alt="Attribute" src="https://github.com/HGustavs/LenaSYS/assets/129263559/8a78bca9-0dfb-4340-8e92-9647430b776d">
<img width="651" alt="Screenshot 2024-05-06 at 09 10 18" src="https://github.com/HGustavs/LenaSYS/assets/129263559/58ee1b47-6a4c-4acd-8d1d-42449e534b77">

 ER Attribute CAN be connected only two ways:
1. to ER Entity
3. as composite attributes 
<img width="1060" alt="Screenshot 2024-05-06 at 09 11 14" src="https://github.com/HGustavs/LenaSYS/assets/129263559/cf82ad87-f054-4397-84f4-e1faea5830df">

#Job
Fix this so that it would be possible to connect attributes to ER entity and to each other.  (As in the third  picture above)

PR #15453, opened 2024-05-07
ER attrbiutes can only be connected to 1 ER entity (#15386)

a22jacka:
Added another check so that 1 ER attribute can only be connected to 1 ER entity


a23ponek (approved):
Each distinct ER attribute can now only be connected to one ER entity. An error message notifies the user when more than one ER attribute is trying to be connected to a ER entity, which works smoothly. No errors or anomalies seem to occur after changes. See images down below for before and after changes.

Before changes:
Screenshot 2024-05-07 151131

After changes: 
Screenshot 2024-05-07 151055

PR #15454, opened 2024-05-07
remove unecessary console logs

a22oscja:
It appears to have been the same lines that were removed in the old issue, likely someone accidentally put them back in.

**testing:**
check if the "User is focusing on course page, isActivelyFocused is now true" and "User lost focus on course page, isActivelyFocused is now false" console logs persist anywhere (although they were only present in sectioned.php)


a22moamy (approved):
LGTM!

PR #15455, opened 2024-05-07
Add retriveSectionedService.php to all microservices.

a22albjo:
I have tested the services, they work. They do have some smaller annoying side effects still that needs to be resolved. Due to the scope of this issue is starting to get very big and i want all of the changes added before the merge i will create new issues to resolve those problems. Read more about the annoying issues further down in this comments. 

Those issues should be resolved before migrating sectionedservice to microservices.

I have tested the following services and i have noted some things about them here, i have fiddled around with the page in general as well to se if i can find any issus, all issues noted are bellow:

**createListEntry_ms.php** 
- After a listEntrie is created all listentries with visibility 3(hidden) is not visible on the screen even if you are a superuser. You need to refresh the page to get the entries to show. Unclear if this is a problem with the monorepo as well.
- After creating a group activity the page will no longer render the listentries, cant get them to show without reinstalling db. All other kinds of listentries works fine and are tested.

**deleteListentrie_ms.php** 
- After a listEntrie is deleted all listentries with visibility 3(hidden) is not visible on the screen even if you are a superuser. You need to refresh the page to get the entries to show. Unclear if this is a problem with the monorepo as well.
- Get a alert that is saying null for no apperent reason.

**removeLestEntrie_ms.php**
- After a listEntrie is hidden all listentries with visibility 3(hidden) is not visible on the screen even if you are a superuser. You need to refresh the page to get the entries to show. Unclear if this is a problem with the monorepo as well.
- When visibility is changed with this service it is not crossed out as it is when visibility is changed by the settings. This might not be in the scoope of the microservice but should be done by section.js. I have changes the trash-bin icon to run this service and that could explain why the services dont get crossed out. But this is good to have in mind.
- Get a alert that is saying null for no apperent reason.

**updateListEntriesTabs_ms.php**
- After a listEntrie is hidden all listentries with visibility 3(hidden) is not visible on the screen even if you are a superuser. You need to refresh the page to get the entries to show. Unclear if this is a problem with the monorepo as well.
- Get a alert that is saying null for no apperent reason.

**setVisibleListentries_ms.php**
- After a listEntrie is hidden all listentries with visibility 3(hidden) is not visible on the screen even if you are a superuser. You need to refresh the page to get the entries to show. Unclear if this is a problem with the monorepo as well.
- When visibility is changed with this service it is not crossed out as it is when visibility is changed by the settings. This might not be in the scoope of the microservice but should be done by section.js. I have changes the trash-bin icon to run this service and that could explain why the services dont get crossed out. But this is good to have in mind.
- The change gets applied after refresh.

**reorderListentries_ms.php**
- After a listEntrie is hidden all listentries with visibility 3(hidden) is not visible on the screen even if you are a superuser. You need to refresh the page to get the entries to show. Unclear if this is a problem with the monorepo as well.
- Get a alert that is saying null for no reason.

**getListEntries_ms.php**
- After a listEntrie is hidden all listentries with visibility 3(hidden) is not visible on the screen even if you are a superuser. You need to refresh the page to get the entries to show. Unclear if this is a problem with the monorepo as well.
- Get a alert that is saying null and then its ok after pressing ok.


DuggaSys/microservices/sectionedService/getDeletedListEntries_ms.php
```diff
@@ -20,6 +21,8 @@
 $courseid = getOP('cid');
 $coursevers=getOP('coursevers');
 $versid = getOP('vers');
+$log_uuid=getOP('log_uuid');
+$userid=getUid();
```
a22marca, line 25: `$uid` is used in the function call.

DuggaSys/microservices/sectionedService/getListEntries_ms.php
```diff
@@ -20,5 +20,6 @@
 $opt=getOP('opt');
 $courseid = getOP('courseid');
 $coursevers=getOP('coursevers');
+$log_uuid=getOP('log_uuid');
 
 echo json_encode(retrieveSectionedService($debug, $opt, $pdo, $uid, $courseid, $coursevers, $log_uuid));
```
a22marca, line 25: `$debug` is not declared

DuggaSys/microservices/sectionedService/getUserDuggaFeedback_ms.php
```diff
@@ -24,6 +24,7 @@
 $courseid=getOP('courseid');
 $moment=getOP('moment');
 $versid = getOP('vers');
```
a22marca, line 26: $courseverse is used in the function call (undeclared), but versid is declared up here. I'm assuimg this should be used

DuggaSys/microservices/sectionedService/setVisibleListentries_ms.php
```diff
@@ -21,6 +21,8 @@
 $courseid = getOP('cid');
 $versid = getOP('vers');
```
a22marca, line 22: $courseverse is used in the function call. I'm assuming verseid should be used. $debug needs to be declared as well

DuggaSys/microservices/sectionedService/updateQuizDeadline_ms.php
```diff
@@ -18,6 +18,7 @@
 $courseid=getOP('courseid');
 $link=getOP('link');
 $versid = getOP('vers');
+$log_uuid=getOP('log_uuid');
```
a22marca, line 21: $courseverse and $debug is undeclared

a22marca (approved):
I've added some comments about variables that are still undeclared. This looks good enough to merge, so we get all the retrieve service integration in after the weekly merge.
Create a new issue to fix the remaining small problems left here.

PR #15456, opened 2024-05-07
Created test for createDugga_ms.php issue#15199

a22melst:


c22cargo (changes requested):
* Course 2 should not be used, instead the testing course 1885 should be used.

* The DELETE statement should probably be more specific. You could do something like this:

```
'query-after-test-1' => "DELETE FROM quiz WHERE qname = 'createDuggaTest' AND cid = 1885",
```

Other than that I think the test looks good.

It seems that a lot of values remain NULL even though they are sent.

```
{"qname":"createDuggaTest","autograde":0,"gradesystem":1,"qstart":null,"deadline":null,"qrelease":null,"jsondeadline":"{\"TestJSON\":\"2024-04-04 04:04:04\"}"}]}
```

This doesn't seem to be the test's fault though, it appears that the microservice is not fetching all the values it needs. An issue should probably be created for this if one doesn't already exist, and afterwards the expected output for the test should be updated.

c22cargo:
The microservice was fixed in issue #15461. It would be nice to have ```group``` in the filter, but currently the retrieve MS doesn't retrieve that value for some reason. You could still add it to the filter if it doesn't cause any issues since this will be fixed in the future.

Issue #14375, opened 2024-04-05
Icon does not change when cycling with keybinds


a22solmi:
## Description
When choosing a toolbar item in a sub menu the toolbar icon changes to the selected item.
When pressing and cycling through the sub menu by pressing the key bind multiple times the icon is not updated.
![issue-v1-7](https://github.com/HGustavs/LenaSYS/assets/129150661/937c7c94-5f79-43f9-8aa8-91feee62e9ee)

## Job 
Implement a way to change the icon in the same way as using the mouse.

PR #15460, opened 2024-05-07
icon change on cycling #14375

a22solmi:
icon in toolbar now changes when pressing  keybinds


a22jacka (approved):
Cycling through the menus with the keyboard now also updates the toolbar icons to show what is selected

image
Icons changed with arrow keys

Issue #15054, opened 2024-04-28
Add a HTML title-attribute to create tooltips for certain buttons in sectioned.php


a22moamy:
Adding a title-attribute to an element makes them display a tooltip when hovering above them. It gives the user a clue as to what a button does. Like so:
![Skärmavbild 2024-04-28 kl  19 59 13](https://github.com/HGustavs/LenaSYS/assets/129263915/a7be2fa3-f589-4dd5-9907-397955455ccf)

The following buttons are missing a title-attribute: the theme-toggler, the "fetch github repo button" and the draggable parts of list entries. These buttons are marked in red:
![Skärmavbild 2024-04-28 kl  19 53 35](https://github.com/HGustavs/LenaSYS/assets/129263915/361a3b2d-7359-4a1e-8276-3d5584246f6d)
![Skärmavbild 2024-04-28 kl  19 57 47](https://github.com/HGustavs/LenaSYS/assets/129263915/48f89601-c584-4b09-a364-26bddaf7de54)

a22diako:
Cant fix this issue because of revert, therefor this one is set on pause.

PR #15465, opened 2024-05-07
G3 2024 v5 #15054

a22diako:
Added title to icons: 
Themetoggle.svg
refreshIMG
select.png

Now user should have a title that shows up when the mouse is over the icon. 

Screens shows the icons that have been fixed.

Skärmbild 2024-05-07 141818

Skärmbild 2024-05-07 141923

image

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -1592,12 +1592,12 @@ function returnedSection(data) {
             if (itemKind === 3) {
               if (isLoggedIn) {
                 str += "<td class='LightBox" + hideState + "'>";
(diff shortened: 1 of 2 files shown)


a20jonda (changes requested):
Titles successfully shows up on Chrome and Edge and are not blocked. 

Changes are needed.

image
Should be more clear that it is a toggle.

image
I think "Press and drag to arrange" would be clearer.

image
I think `list_docfiles.svg` should be titled "Moment" instead.

image
I think `PenT.svg` should be titled something like "Quiz"  instead

Issue #15267, opened 2024-05-02
Lock causes arrows to be displaced


a22solmi:
## Description
When locking an element it is added to the end of the div which causes all arrows/line to point at nothing.
![Screenshot from 2024-05-02 11-07-51](https://github.com/HGustavs/LenaSYS/assets/129150661/e717c115-64ea-432a-a41d-77f93cac2e30)

## Job
Move the lock to be placed on top of the current div.

a22ricni:
What I have found is that the "lock" is a image and the whole element is in a div tag and when you want to lock a element then the height of the div expands and that's why the arrow moves down a bit from the element. Have tried to use position absolute in the img tag, but doesn't work the way I wanted. The lock image is on the code DuggaSys/diagram/draw/element.js on row 124 to 144.

PR #15474, opened 2024-05-07
fixed padlock expanding elements #15267

a22jacka:
Moved the padlock with `position: absolute`, removed `overflow: hidden` if a padlock is present, moved the padlock with zoom


a22solmi (changes requested):
Darkmode toggle moves padlock. Background color also does not change on darkmode.
Screenshot from 2024-05-07 14-59-12
Screenshot from 2024-05-07 14-59-21

PR #15476, opened 2024-05-07
Deleted an condition

a22cazlu:
The condition did so that we couldn't select the lines when it was in an object. When we did this it created a bug that makes the app to think that we do an double click when we just do one, this just hapends when the line is in an object. Maybe we can create a new issue about that so we can merge in this?


b22davbe:
[screen-capture.webm](https://github.com/HGustavs/LenaSYS/assets/129262690/5a6192fd-efa1-4b2d-9999-ee414dacef15)

Issue #15461, opened 2024-05-07
Fix createDugga_ms


c22cargo:
Currently ```createDugga_ms.php``` does not retrieve all the variables it needs to function. An example is ```$qstart```. This results in inserts having NULL values even though the values were not sent as NULL.

PR #15479, opened 2024-05-07
g2-2024-v5 issue #15461

c22cargo:
Necessary code was added, unnecessary code was removed, refactoring was done.

You can test the microservice by going into a course, clicking "show tests" at the top, and then creating a new test. The frontend doesn't seem completed so you will have to check the database to see that the correct values are entered.

Change row 1062 in ```dugga.js``` to:
```
}else if(kind=="DUGGA"){
		let service_url = "duggaedservice.php"
		
		if (opt == "SAVDUGGA")
			service_url = "microservices/duggaedService/createDugga_ms.php"

		$.ajax({
			url: service_url,
			type: "POST",
			data: "opt="+opt+para,
			dataType: "json",
			success: returnedDugga
		});
```

### DIFF
--- a/DuggaSys/microservices/duggaedService/createDugga_ms.php
+++ b/DuggaSys/microservices/duggaedService/createDugga_ms.php
@@ -1,63 +1,79 @@
 <?php
+include_once "../../../Shared/basic.php";
+include_once "../../../Shared/sessions.php";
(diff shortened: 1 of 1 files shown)


a22melst (approved):
It works.

PR #15480, opened 2024-05-07
Sequence Lifeline object is black instead of white

a22robko:
I added a CSS code to make the image inside the class "placementTypeBoxIcons" white.

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -702,6 +702,7 @@
 .placementTypeBoxIcons img{
     width:34px;
     height:34px;
+    filter: brightness(0) invert(1);
 }
 .placementTypeToolTipText{
     top:0%;


a22vioja:
Closes issue #15481

a23ponek (changes requested):
The sequence Lifeline object is now white in the "slide" of the toolbar when the different options are viewed. But when its chosen the color of the "eye" still remains the color black. See video capture before and after changes below.

Before changes:

https://github.com/HGustavs/LenaSYS/assets/164863298/ef444e22-77f8-4702-a37c-c1f71ce50423

After changes:

https://github.com/HGustavs/LenaSYS/assets/164863298/5e191a5b-67c8-4a2b-aff7-2db38de3d702

a22vioja:
Changes where made on a new branch:  #15522

Issue #15081, opened 2024-04-29
Min height not checked during type change


a22solmi:
## Description
When changing the type of an entity in the options>property sidebar the minimum height is not checked.
UML, IE and SD entities have been reworked to resize with text and start with the height of 0.
ER entities entire height is determined by element.height which means it will not be visible if set to 0.

## Job
Change drawElementEREntity to draw the element with a minimum height simmular to drawElementIE / UML / SD  -entity

PR #15482, opened 2024-05-07
#15081 fix, when changing an UML/SD/IE element to an ER entity the height doesnt default to zero anymore. 

b22jonan:


a22solmi (approved):
This does not solve the core issue but it works as a quick fix

PR #15483, opened 2024-05-07
15352 create microservice highscoreservice msphp

c20juler:
The microservice seems a bit redundant as the original mono file is already reasonably small. Im not sure if the file should be simmered down further so will just create the file and add it to microservices/highScoreService/highscoreservice_ms.php. 

Only file to review is highscoreservice_ms.php located in microservices folder


a22iliru (approved):
Tested the ms and it works.

PR #15485, opened 2024-05-07
G1 2024 v5 #14235

a22ricni:
Now it is possible to center the camera when there is no element on the screen. Can test it by have element on the screen and press the reset button to see if the camera is on the default position or not (it should only remove all the element). And if there is no element on screen then you should be able to center the camera to default position.


a22vioja:
Closes issue #14235

PR #15500, opened 2024-05-08
Errors when entering new courses

a22oscja:
It was never declared if there was data in the database, specifically it would only declare if there was no github repo set.

Now it sets depending on if its column has a value and also gets set if there is no data at all.

**Testing:**

- open basically any page on the website(sectioned, courseed, etc), check for an alert that complains about updateTime when the page loads.

### DIFF
--- a/Shared/navheader.php
+++ b/Shared/navheader.php
@@ -66,14 +66,14 @@
 				} else {
 					$checkIfGithubURL = null;
 				}
-				if (isset($row['updated'])){
+				if (isset($row['updated'])) {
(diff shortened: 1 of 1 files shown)


b22marah (commented):
LGTM! But one quick fix before i approve. The code doesn't follow the K&R indentation, the '{}' should be spaced after other parameters. At the moment your code looks like this "if (isset($row['updated'])){
}"

But should look like this to follow K&R:
"if (isset($row['updated'])) {
}"

b22marah (approved):
LGTM!!!

PR #15503, opened 2024-05-08
Internal server error when making new course

a22oscja:
*copied over changes from last week, did a quick test to make sure it worked and fixed an edge case.

old text:

It was attempting to insert data into the database twice, leading to the error. I also added support for the old data. As a result of this fix, the course should also now show up without reloading the page, as it was getting stopped from making that happen before.

**testing:**

- if you do not have access to old data, get it from me/leaders/anyone really
- make a new course, check for errors.
- repeat on current/old data as much as you'd like.


b22marah (approved):
LGTM!

PR #15505, opened 2024-05-08
G2 204 v6 issue#15499

a22marca:
Change the parameters of createNewListentry from a array to variables instead, as well as the same name as the microservice file.


a22iliru (approved):
Works, POST request on the ms is sent in network tab.

Issue #15478, opened 2024-05-07
Code cleanup


a22solmi:
## Descriptions
There are functions that take parameters it does not use and other small formatting issues all across the codebase.

## Job
Fix various issues before weekly merge

d22hanol:
2 errors in line.js in drawLineLabel and drawLineCardinality

PR #15506, opened 2024-05-08
G1 2204 v5 #15478

a22solmi:
Updated unused paramters, inconsistent whitespace, const/let instead of var and missing semicolons


a21liltr (approved):
LGTM

PR #15511, opened 2024-05-08
Created "Add User" button - G3 2024 v6 #15491

a22natth:
In the example below, when hovering the “+” button, an additional icon has been added named “add user”. 
Pasted Graphic 8

When pressing on the icon, the following modal will appear with the label “Choose user” combined with a searchable dropdown menu. A save button has also been added. 
Pasted Graphic 9

In the example below, the search ability is displayed with example “data”. 
Pasted Graphic 10

Furthermore, the example below portrays the ability to just press the dropdown menu which then will display the different options. 
Pasted Graphic 11

No further functionality has been implemented since it was not requested. Additionally, no specification regarding the “add user” icon has been made, therefore, the icon used is the same one as for “create user”. 

Co-Author: @a18chrkr


d22mikwe:
LGTM

a22ahmom (changes requested):
Everything looks good except the modal seems to have some CSS issues. 
Screenshot 2024-05-08 144407

PR #15512, opened 2024-05-08
Added ability to restart the installer from specified step

a22willi:
The installer now creates an associative array containing all the operations the installer needs to perform. This array is constructed based on the settings provided via json from the frontend. Specifying the parameter: `starting_step: "add_test_data"` will for example restart the installer from the add_test_data step. 

# How to test?
1. First change the file `LenaSYS/newinstaller/tools/install_engine.php` to contain the proper pdo authentication settings. 
image

2. Since everything should be working correctly, (assuming it does, thats why you are testing haha) you first have to get it to fail somewhere. This can be achieved by changing the installer options in the file `LenaSYS/newinstaller/installer_ui.php`. For example, change the language support list to a language that does not exist. This should result in a fail, where it tells you where it failed. 
image
Running the code by navigating to [localhost/LenaSYS/newinstaller/installer.php](http://localhost/LenaSYS/newinstaller/installer.php) should now result in the following exception.
image

3. Now remove the language that you added that does not exist. And set the starting step to `add_language_support_html`. Should look like this:
image
And give the following output.
image
Notice how the first step now was to install the html language keywords.


a22karja (commented):
The install works when run without error.
The install works when running with error and then restarting it. and it start from where i tell it start.
It might be good if the text sais which step it is in so the user can easily see what starting step they should pick
image
from this issue the starting step should be "add_language_support_html" but user dosent know this.

I suggest keeping the old installer until the UI is done and can be tested more. Even if this backend works

a22karja (approved):

Issue #15388, opened 2024-05-06
Replace all alerts with toasts (except for in Diagram files)


a22moamy:
There are still a lot of alerts in LenaSYS. You can find every alert in the system by pressing Ctrl + Shift + F (Cmd + Shift + F on mac) and writing something like _alert(_ in the searchbar. 

Here you can see some files that use alerts:
![Skärmavbild 2024-05-06 kl  09 50 31](https://github.com/HGustavs/LenaSYS/assets/129263915/d10ecb38-ea63-496f-b820-43808f37cc97)

Aim to replace all alerts with toasts, **except for in the diagram files which group 1 is working on.** The toasts should have a default duration of 10 seconds and have the correct type. See [the Toast wiki](https://github.com/HGustavs/LenaSYS/wiki/Toast) to learn how to implement toast. 

If you deem a longer/shorter duration is needed for a specific toast, then you may decide the duration.

a20jonda:
Since I was the last to add toasts. The debug alerts and the alerts which halts a refresh are exempt.
```javascript
alert(data['debug']);
```

PR #15513, opened 2024-05-08
Changed important alerts to toasts. G3 2024 v6 #15388

a22edvbu:
I changed the most prominent alerts to toasts when browsing LenaSYS. Some alerts are left untouched due to them being needed for debugging or having important functions on the backend. By doing changes in courseed or codeviewer etc the alerts should now be replaced by toasts. 

One issue that this has generated is that toast does not pause the execution of code like alerts did. This means that some alerts, for example in codeviewer, shows for a short moment before the site is refreshed.

### DIFF
--- a/DuggaSys/codeviewer.js
+++ b/DuggaSys/codeviewer.js
@@ -77,7 +77,7 @@ function setup() {
 			cvers: cvers
 		}, "CODEVIEW");
 	} catch (e) {
-		alert("Error while setting up: " + e.message)
+		toast("error","Error while setting up: " + e.message, 10);
 	}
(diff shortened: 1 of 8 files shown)


d22isaka (commented):
Not sure how to trigger all of these toasts.
But the few I managed is working(sectioned.js and codeviewer). I'll ask tomorrow about how to trigger duggaed and contribution. Right now duggaed redirects me back to sectioned without any notif

d22isaka (approved):
Looks good! =)

d22mikwe:
LGTM!

Issue #15514, opened 2024-05-08
Make retrieveDuggaedService_ms.php retrieve all needed values from the table "quiz"


c22cargo:
Currently at least the ```group``` value is not being retrieved. 

The table looks like:

```
CREATE TABLE quiz (
	id						INT(11) NOT NULL AUTO_INCREMENT,
	cid 					INTEGER UNSIGNED NOT NULL,
	autograde 				TINYINT(1) NOT NULL DEFAULT 0, /* bool */
	gradesystem 			TINYINT(1) NOT NULL DEFAULT 2, /* 1:U-G-VG & 2:U-G & 3:U-3-5 */
	qname 					VARCHAR(255) NOT NULL DEFAULT '',
	quizFile 				VARCHAR(255) NOT NULL DEFAULT 'default',
	qrelease 				DATETIME,
	deadline 				DATETIME,
	relativedeadline		varchar(10),
	modified 				TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
	creator 				INTEGER,
	vers					VARCHAR(8),
    qstart					DATE,
	jsondeadline			VARCHAR(2048),
	`group` 				TINYINT(1) DEFAULT 0,
	PRIMARY KEY (id),
	FOREIGN KEY (cid) REFERENCES course(cid) ON DELETE CASCADE ON UPDATE CASCADE
)
```

PR #15518, opened 2024-05-08
g2-2024-v6 issue #15514

c22cargo:
The file should now be fixed.

I tested it by doing:
```
INSERT INTO quiz (cid, autograde, gradesystem, qname, quizFile, qrelease, deadline, relativedeadline, creator, vers, qstart, jsondeadline, `group`)
VALUES (1885, 0, 1, 'Test321', 'group-assignment', '2024-02-02 02:02:02', '2024-01-01 01:01:01', 'Snart', 101, '1337', '2024-03-03', '{"TestJSON":"2024-04-04 04:04:04"}', 1);
```

And then executing ```createDugga_ms_test.php``` and looking at the response (no filter).

Now all relevant values should be retrieved.

Since ```createDugga_ms_test.php``` is the only test for duggaedService implemented at this time, the expected output only needed to be updated there.

### DIFF
--- a/DuggaSys/microservices/duggaedService/retrieveDuggaedService_ms.php
+++ b/DuggaSys/microservices/duggaedService/retrieveDuggaedService_ms.php
@@ -1,19 +1,28 @@
 <?php
-
-date_default_timezone_set("Europe/Stockholm");
-
-// Include basic application services!
-include_once "../../../Shared/basic.php";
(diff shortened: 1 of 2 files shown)


c20juler:
Reviewed, retrieveDuggaedService_ms.php works as expected.

PR #15519, opened 2024-05-08
G1 2024 v6 #14235

a22ricni:
Now it is possible to center the camera when there is no element on the screen. Can test it by have element on the screen and press the reset button to see if the camera is on the default position or not (it should only remove all the element). And if there is no element on screen then you should be able to center the camera to default position.

### DIFF
--- a/DuggaSys/diagram/camera.js
+++ b/DuggaSys/diagram/camera.js
@@ -4,7 +4,7 @@
 function centerCamera() {
     // Stops execution if there are no elements to center the camera around.
     if (data.length == 0) {
-        displayMessage(messageTypes.ERROR, "Error: There are no elements to center to!");
+        centerToOrigo();
         return;
     }
     // Centering needs to happen twice for it to work, temp solution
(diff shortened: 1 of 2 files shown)


a22cazlu (changes requested):
As i understand it the camera should focus at origo when there is no elements in the diagram and I press "Reset view". And when I test it that dosen't happened.

see video:

[screen-capture (4).webm](https://github.com/HGustavs/LenaSYS/assets/131054714/d08dc915-fba4-4c9e-ac9a-34f2ce1a7999)

a21liltr (changes requested):
## Please read the issue again. Camera should reset when diagram is empty.

Your solution does not include a reset to the camera when pressing **reset view**.

https://github.com/HGustavs/LenaSYS/assets/129370135/aaaae487-ff27-4e1e-b8d9-a5f9e255b614

Also code should be in DuggaSys > diagram > camera.js

a22ricni:
Should now work when "reset view" button is clicked if no element is on screen.

DuggaSys/diagram/camera.js
a21liltr, line 1: NOT OK: Duplicated code.

centerToOrigo is 90% duplicated code from centerCamera.

Remove the duplicated code in centerCamera and add centerToOrigo to centerCamera to use instead.
Make use of arguments to take in parameter to center to either origo or to the middle of the diagram with objects.
(Whether you use variables or bools to determine the recentering is up to you.)

a21liltr (changes requested):
See comment on file below.

a22cazlu (changes requested):
a21liltr mentions the issue in the code where there is duplicated code. But the diagram seems to work and the reset view now works when there is no elements in the diagram.

a22ricni:
I have removed the duplicated code and hopefully made it more clean.

a22cazlu (approved):
What I can see it works. No meaning for me to request changes when a21liltr already done it.

a21liltr (approved):
Lgtm

PR #15520, opened 2024-05-08
deleted some math to fix centering and fixed EIrelation

a22cazlu:
We deleted "- (settings.grid.gridSize / 2)" from updateElementDivCSS() so now the elements are centered except IEEntity, SDstate, UML class and note. It is because those elements are multiple objects in one and their real heights are stored in arrays. This causes the program to use the height of zero for calculating the mouse position.

The arrays:
var UMLHeight = [];
var IEHeight = []; 
var SDHeight = [];
var NOTEHeight = [];

### DIFF
--- a/DuggaSys/diagram/defaults.js
+++ b/DuggaSys/diagram/defaults.js
@@ -99,11 +99,11 @@ const defaults = {
         fill: color.WHITE,
         stroke: color.BLACK,
         width: 50,
-        height: 50,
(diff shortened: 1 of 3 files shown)


a22cazlu:
And the IErelation is now fixed so it is centered.

a22solmi (changes requested):
This only works for some elements

PR #15522, opened 2024-05-08
Sequence Lifeline object is black instead of white

a22robko:
I added CSS code to make the image inside the class "placementTypeBoxIcons" white. And a class with name class="white-icon".


b22davbe (approved):
Looks good!

PR #15523, opened 2024-05-08
Updated margin-right and style for forms in 3 files g3-2024-v6-15495

a20jonda:
There was a conflict with the `Add user` form in accessed and all other forms in courseed- and sectioned-, duggaed.php. The solution was to accept and implement the style without the extra 10px margin. 

Extra: Added missing style class to submit button and missing body element in `Github Moment` form in sectioned.php which was more apparent without the 10px margin. Changed flexwrapper to inputwrapper in `Edit Dugga` to ensure it implements the style of the other forms.

image
Note that `Add User` stays the same and will no longer conflict

image
image
image
image
image


a22natth (approved):
The testing was done on on the popup windows for editing a dugga, editing new group, creating a user, editing a GitHub moment, creating a new course and editing a course. In comparison to the weekly branch (where the same type of testing was done), the implemented solution seems to have solved the issue. 

The code was also checked in VS code, to which it **not** present any errors or problems in regards to indentation or *added* inline styling for instance.

PR #15528, opened 2024-05-08
Created test for createDuggaVariant_ms.php issue#15197

a22melst:
I wanted to have more output by filtering like this, but it did not work for some reason. I guess it has to do with test.php. 
```
'filter-output' => serialize(array(
        'entries' => array(
            'variants' => array(
                'param',
                'variantanswer',
                'disabled',
            ),
        )
    )     
),
```
Just having it like this works, but then 'modified' is included which depends on the time the test was executed.
```
'filter-output' => serialize(array(
        'entries' => array(
            'variants'
        )
    )     
),
```


a22iliru (approved):
All tests pass and output match

Issue #15508, opened 2024-05-08
Overlapping broken


a22solmi:
## Description 
Issue #15395 broke the overlapping on drag for all objects.

## Job
Find a change that fixed overlapping without breaking the issue snap to grid features

PR #15533, opened 2024-05-08
Fixed overlapping check #15508

a22solmi:
Overlapping is checked for objects when not in snap to grid.
newX and newY was undefined for all non snap checks


a22vioja:
closes issue #15508

a21liltr (approved):
Check works properly.  There is now a check for all placable objects to see if they overlap when placed, moved to overlap or resized to overlap.

Some objects (that consist of multiple boxes) can overlap partly, but that's another issue.

PR #15535, opened 2024-05-08
Migrated the microservices that are done for the sectioned page.

a22albjo:
I have migrated the services that are created as microservices on the sectioned page.

sectionedservice.php are used as a fallback.

The opts not implemented are: REFGIT and CREAGITEX.


c22cargo (changes requested):
In ```dugga.js``` there is first the if check:
```
if(kind=="SECTION" || kind=="GRP")
```

and then afterwards:

```
else if(kind=="GRP")
```

This seems like a mistake. The ```else if``` will never be reached. Also, it has code under it that isn't indented correctly (that code could perhaps be removed, not sure). 

The AJAX function call should _probably_ send ALL the different OPT values, even if some are NULL (I assume you are currently doing that). This way there only needs to be 1 function call, like you have it now.
There is another part of ```dugga.js``` that has implemented microservices, but instead of only changing the URL they change the whole function call for every microservice.

Your approach seems simpler, and I assume you are already sending all necessary OPTs. We should probably discuss within the group what the best approach for this is. I personally think that sending all OPTs for every MS is likely better, but I am not completely sure if it will work how I think it will, and it preferably needs to be tested.

I am not familiar with javascript, but I think the ```serviceURL``` should look different. It should probably be:

```
let serviceURL = "sectionedservice.php"
```

And then you use it like

```
serviceURL= "../DuggaSys/microservices/sectionedService/removeListEntries_ms.php";
```

I will test the integration of all the microservices after the changes have been made and the code looks okay. This is because I don't want to have to re-test everything unnecessarily.

a22albjo:
I have implemented the feedback.
I don't understand entirely what you mean here, would you be able to help me with code example that shows the different scenarios:
> The AJAX function call should probably send ALL the different OPT values, even if some are NULL (I assume you are currently doing that). This way there only needs to be 1 function call, like you have it now.
> There is another part of dugga.js that has implemented microservices, but instead of only changing the URL they change the whole function call for every microservice.
> 
> Your approach seems simpler, and I assume you are already sending all necessary OPTs. We should probably discuss within the group what the best approach for this is. I personally think that sending all OPTs for every MS is likely better, but I am not completely sure if it will work how I think it will, and it preferably needs to be tested.

Shared/dugga.js
```diff
@@ -1127,32 +1127,69 @@ function AJAXService(opt,apara,kind)
...
+				serviceURL= "sectionedservice.php";//change when MS i; created
+				break;
+			case "SETVISIBILITY":
+				serviceURL= "../DuggaSys/microservices/sectionedService/setVisibleListentries_ms.php";
+				break;
+			case "REFGIT":
+				serviceURL= "sectionedservice.php";//change when MS i; created
+				break;
+			case "CREGITEX":
+				serviceURL= "sectionedservice.php";//change when MS i; created
```
a22albjo, line 1164:
```suggestion
				serviceURL= "../DuggaSys/microservices/sectionedService/changeActiveCourseVersion_sectioned_ms.php";
```

c22cargo (changes requested):
Looking at the monolith, I think ```case "DELETE"``` should be ```case "DELETED"```. If it's not a mistake you can ignore this.

I could not find where the following microservices are used:
* removeListEntries
* updateListEntriesTabs
* setVisibleListentries
* changeActiveCourseVersion_sectioned
* getCourseGroupsAndMembers (You need to be able to create a new group-activity to test this, but I can't create any new listentries.)

I could not get the following microservices to work:

* deleteListEntries
* createListEntry

It seems like the following microservices work:

* reorderListEntries
* updateListEntries
* updateQuizDeadline

For those microservices that I could not find where they are used, you can see if you can find where they are used and if they work. If you can't find them you can probably just leave them implemented as is.

For the microservices that I could not get to work, if you also cannot get them to work you should change their URL back to the monolith and place a comment.
I think for those 2 that didn't work it might have to do with the ajax function call, because I tried changing the URL back to the monolith for the "NEW" opt, but I still couldn't create any new listentries.

For the microservices that I got to work, you can leave them implemented as is.

a22albjo:
> Looking at the monolith, I think `case "DELETE"` should be `case "DELETED"`. If it's not a mistake you can ignore this.
> 
> I could not find where the following microservices are used:
> 
> * removeListEntries
> * updateListEntriesTabs
> * setVisibleListentries
> * changeActiveCourseVersion_sectioned
> * getCourseGroupsAndMembers (You need to be able to create a new group-activity to test this, but I can't create any new listentries.)
> 
> I could not get the following microservices to work:
> 
> * deleteListEntries
> * createListEntry
> 
> It seems like the following microservices work:
> 
> * reorderListEntries
> * updateListEntries
> * updateQuizDeadline
> 
> For those microservices that I could not find where they are used, you can see if you can find where they are used and if they work. If you can't find them you can probably just leave them implemented as is.
> 
> For the microservices that I could not get to work, if you also cannot get them to work you should change their URL back to the monolith and place a comment. I think for those 2 that didn't work it might have to do with the ajax function call, because I tried changing the URL back to the monolith for the "NEW" opt, but I still couldn't create any new listentries.
> 
> For the microservices that I got to work, you can leave them implemented as is.

**createListEntries** works fine for me, i hover the plusicon and pick a entrie type and it gets added to the lisentries.
**deleteListEntries** was broken by another group. it can still be tested by pressing the checkbox before delete.
**removeListEntries** have been tested by me by changing opts to the trash can icon.
**updateListEntryTabs are used by pressing this:** Skärmavbild 2024-05-13 kl  16 28 20
**setVisibleListentries** are used by marking entries and pressing one of theese: 
Skärmavbild 2024-05-13 kl  16 29 15
**changeActiveCourseVersion_sectioned** not sure, @a22marca can you help us with this, i think you were having an issue involving this?
**getCourseGroupsAndMembers** cant be tested.

a22marca:
> changeActiveCourseVersion_sectioned not sure, @a22marca can you help us with this, i think you were having an issue involving this? getCourseGroupsAndMembers cant be tested.
@a22albjo There is no part of the javascript side that makes a request using "CHGVERS". This basically means there is no part of the frontend that makes a request to perform this specific functionality.

c22cargo (approved):
I removed those microservices that don't seem to currently work. They will need to be fixed and then integrated afterwards.

The microservices are:
* createListEntry
* deleteListEntries (listentries appear to get deleted, but after refreshing they are back)
* updateListEntries (some parts work, some don't)
* updateListEntriesTabs (MS seems to update wrong thing)
* getListEntries

Issue #15338, opened 2024-05-03
Fix functionality to delete a whole course


a22moamy:
Currently, you cannot delete a whole course. You can press the cogwheel for a course (`in courseed.php`) and press the red delete button, but nothing will happen. Investigate why the button does not work and fix it.

b22chran:
Researched old issues to see where the previous delete-function was removed. Found that the edit course section is not functional because it needs a gitkey to save changes and test the delete function.

Issue #14688, opened 2024-04-17
Investigate why loading of old SQL data failed


a22willi:
In issue #14653 it was determined that loading data from the current LenaSYS system (the one the school is currently running live) did not work, investigate the first error, and fix it, so that the script can create the data successfully. 

> **cannot drop tables:** _“table” - “has issue with” - “table with foreign key”:_
> 
>     - box - improw_ibfk_2 - improw
> 
>     - class - programcourse_ibfk_2 - programcourse
> 
>     - codeexample - box_ibfk_1 - box
> 
>     - course - course_req_ibfk_1 - course_req
> 
>     - groups - user_group_ibfk_1 - user_group
> 
>     - listentries - useranswer_ibfk_3 - useranswer
> 
>     - quiz - variant_ibfk_1 - variant
> 
>     - template - codeexample_ibfk_3 - codeexample
> 
>     - user - course_ibfk_1 - course
> 
>     - wordlist - fk_word_joins_wordlist - word

a22oscja:
The issue, as mentioned, appears to lie in tables from the current database being unable to be dropped due to foreign keys in different tables referencing their values. To this end a working order was determined, it is the following:

1. improw
2. box
3. programcourse
4. class
5. impwordlist
6. codeexample
7. course_req
8. user_course
9. useranswer
10. variant
11. timesheet
12. quiz
13. vers
14. filelink
15. partresult
16. subparts
17. announcement
18. userduggafeedback
19. shregister
20. user_participant
21. opponents
22. listentries
23. course
24. user_group
25. groups
26. template
27. word
28. wordlist
29. user_push_registration
30. user
 
This order is not completely set in stone, to make it I simply went through the errors given when trying to drop tables in the order the original file attempts to. Then i would simply drop whatever it complained about first and repeat until everything dropped correctly.

At this point the simplest solution seemed to be to drop all of these in the beginning of the script, which is exactly what has been done.

The next part of the issue was the fact that the order the tables are created had to be changed to accommodate for the tables being dropped so early. This is obviously fine to do anyway, since a problem with referencing to data that hadnt been written yet would occur later on anyway.
To do this a similar process to before was used, simply go from top to bottom in the document and check for tables with foreign keys, then make sure that the table it references lies above it in the creation order, then repeat until it fully works. Here is the order used:

1. user
2. course
3. template
4. codeexample
5. box
6. class
7. course_req
8. eventlog
9. filelink
10. groups
11. improw
12. impwordlist
[...]

PR #15540, opened 2024-05-08
Moved code from v5-branch

a22danwe:
Moved code from branch "g3-2024-v5-#15338" to "g3-2024-v6-#15338"


a22danwe:
we managed to make the delete button work after reorganising the delete queries in courseedservice.php. what we noticed is that courses with code examples in them can not be deleted due to nesstled tables in init_db.sql. we spoke to the groupleader and agreed that this issue is too large and should be split into smaller issues.

when testing:
avoid deleting these courses: Demo-Course, Webbprogrammering, Testing-Course, Webbutveckling-datorgrafik.

when deleting a course you press the cogwheel to the right of the course name
bild

then you press the red deletebutton
bild
also create courses and delete them, see if any internal error message appears.

a22oscja (commented):
While there is nothing implicitly incorrect about the solution and it working as intended. It does cause issues regarding backwards compatibility in its current form.

The problem itself lies in the old data having different foreign keys as well as a few tables that exist/dont exist when compared to our version of the database. While I have not looked into it in detail, I did confirm that it breaks.

This issue becomes annoying to fix with the current way deleting of courses is done (setting the visibility to 3 and then deleting all courses with that visibility when loading courseed.php). I'm not saying it would be impossible to fix, but worst case it doubles the amount of code used for deleting (because of order of operations difference)

To this end it may be easier to implement the DEL command in AjaxService (it currently exists, but does nothing). This would if nothing else, make it easier to work around the problem (and not further clutter courseedservice.php).

Either way, I have spoken to the leaders about this and will leave it up to them on what to do. Also, to whoever ends up working more on deleting - more work was done on what order things can be deleted and how they were built up in #14688

a22edvbu (approved):
Looks good to me!

Issue #15433, opened 2024-05-06
updateCourseVersion_ms.php (courseedservice) is performing the same query as updateCourse_ms.php


a22marca:
It seems both are providing the same functionality. The updateCourseVersion should provide the same functionality as the code for the opt UPDATEVRS in courseedservice.php. Since that microservice is used to edit a course version, not a course itself.

Furthermore, if we are following the potential microservice documentation, we might also want to rename the microservice to
updateCourseVersion_courseed_ms.php.

a22marca:
Waiting for [15304](https://github.com/HGustavs/LenaSYS/issues/15304)

a22marca:
Looking further into this issue I can see that updateCourseVersion_sectioned_ms.php is actually performing almost the exact same code as the code that would be implemented in updateCourseVersion_courseed_ms.php.

courseedservice.php and sectionedservice.php both handles the opt "UPDATEVRS" and perform a similar functionality. Furthermore, I can't find any part of the code base that sends the opt "UPDATEVRS" to the courseedservice.php file, so I'm not sure this is used. As of now the only way of changing a course version is by navigating into a course (sectioned.php). So maybe this microservice is not needed at all?

a22marca:
I have added the microservice now. It does the same thing as the updateCourseVersion_sectioned_ms.php currently does.
It seems some functionality like copying a course version was previously sent to courseedservice.php, but updating a course version was sent to sectionedservice.php. So the previous years group added multiple microservices regarding course versions to both courseedservice and sectionedservice microservices.

Since all these course version functionality is performed when the user is in sectioned.php, i believe the microservices regarding course versions should be added in the sectionedservice microservices directory instead. The opt "CPYVRS", "UPDATEVRS", "NEWVRS", "CHGVRS" should be performed like so:
```
      AJAXService("CPYVRS", param, "SECTION");
```
But it currently looks like:
```
      AJAXService("CPYVRS", param, "COURSE");
```

a22marca:
I will create a new issue to move the *CourseVersion microservices to `microservices/sectionedService`.

a22marca:
To test the microservice add this code:
```
	}else if(kind=="SECTION"){
		let serviceURL="sectionedservice.php";
		if (opt=="UPDATEVRS"){
			serviceURL="../DuggaSys/microservices/courseedService/updateCourseVersion_ms.php";
		}
				$.ajax({
					url: serviceURL,
					async: false,
					type: "POST",
[...]

PR #15541, opened 2024-05-08
G2 Merge #15433

a22marca:
Changed the microservice to actually update a course version. See issue comments for testing.


a22iliru (approved):
A post request is sent to the microservice. so it works

Issue #15543, opened 2024-05-08
Allow the filter-output to include an endless amount of nested arrays


c22cargo:
Currently ```test.php``` doesn't allow too many nested arrays. For example, when creating a test you cannot do:

```
'filter-output' => serialize(
    array(
        'entries' => array(
            'variants' => array( // This array cannot be added
                'vid',
                'param',
                'variantanswer',
                'disabled'
            ),
        ),
    )
)
```

You can try this out with ```updateDuggaVariant_ms_test.php``` (if it has been created at the moment you are reading).

c22cargo:
It kind of works at the moment (not finished), but you need to change the filter-output to:

```
'filter-output' => serialize(
            array(
                'entries' => array(
                    'variants' => array(
                        'vid'=>"",
                        'param'=>"",
                        'variantanswer'=>"",
                        'disabled'=>""
                    )
                )
            )
        )
```

This is not ideal since we preferably don't want to have to change the filter-output for every single test.

We will look into the possibility of fixing this.

PR #15545, opened 2024-05-08
g2-2024-v6 issue #15202

c22cargo:
Unfortunately there seems to be an issue with ```test.php```, which is why the filter-output looks like it does. Until #15543 has been fixed it will have to look like this, unless a tester finds another solution.


a22iliru (approved):
All tests pass. seems fine

PR #15546, opened 2024-05-08
If a user is logged out in profile.php, they should be taken to the home page and no errors should appear

a22oscja:
If you log out while in profile.php you will now be directed to courseed.php.

Also the error regarding push notifications will no longer appear if not working in a setup that supports it. Essentially it was trying to interact with an element that would only be created if support for push notifications was installed, hence it was never created.

I also included some improvements to pushnotifications.js (changing var to let, and making parts of it use the standard for pushnotifications). I did not touch the other parts as I was unable to test it, so did not want to touch.

**Testing instructions:**

- To reach profile.php, be logged in and click your name in the top right from any page.
     - if at this point it tells you that you need a secure https connection (which it will, unless you use one), go into profile.js
     - comment out lines 212-214 (they do nothing except this)
- at this point, check for errors and such
- if you wish to try to load the page with the push notification support and you do not already have support for it installed
     - add ! to the if-statements at lines 27 and 95 in profile.php
     - note that errors will appear if you click the button that now exists since the support is not installed
- at this point, log out from within profile.php and any other page to verify that it does not always move you.

### DIFF
--- a/DuggaSys/profile.php
+++ b/DuggaSys/profile.php
@@ -23,8 +23,13 @@ pdoConnect();
 	<script src="../Shared/js/jquery-ui-1.10.4.min.js"></script>
 	<script src="../Shared/dugga.js"></script>
 	<script src="profile.js"></script>
-
-	<script src="pushnotifications.js"></script>
+	<?php
+	if (defined('PUSH_NOTIFICATIONS_VAPID_PUBLIC_KEY')) {
+	?>
+		<script src="pushnotifications.js"></script>
+	<?php
+	}
+	?>
 </head>
 <body>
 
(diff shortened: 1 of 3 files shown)


a22edvbu (commented):
I've checked the login functionally and it seems to be working fine. When logging our from profile.php i get redirected back to courseed.php as mentioned. This does not happen when the HTTPS connection is not established however, but this is probably a minor issue. 

Sadly I'm not able to test the new push notifications due to me not having that set up. Might need a second opinion on this test to be sure that it works.

Otherwise great work!

a22moamy (approved):
Looks good! Works as intended.

Issue #15539, opened 2024-05-08
UML IE SD NOTE total height wrong


a22solmi:
## Description
The height for all elements that resize with text check overlapping wrong after weekly merge.

## Job
Fix height so that overlapping works

PR #15548, opened 2024-05-08
Corrected IE height #15539

a22solmi:
IE height now includes primary keys


b22chrmo (approved):
Adding text to an element will now update the real height of the element, so it doesn't overlap with other elements. 

It is still possible to put an element directly above another element and then add text, resulting in overlapping elements, but this is not a problem that this issue needs to solve.

PR #15549, opened 2024-05-08
G2 2024 v6 issue#15200

b21isabe:
Created the test: deleteDuggaVariant_ms_test.php

The test creates 2 variants and deletes one of them, in the output only the non-deleted variant can be seen.


a22iliru (approved):
All tests pass with correct output

Issue #15341, opened 2024-05-03
Pressing the "Show files"-button from the navbar of a course generates error message


a22moamy:
When pressing the "Show files"-button from the navbar of a course, this error message is generated in the console:
![Skärmavbild 2024-05-03 kl  14 56 43](https://github.com/HGustavs/LenaSYS/assets/129263915/7118df86-263d-45d6-bd62-919ca114478d)

**Please fix this.**

The "Show files"-button (from `sectioned.php`):
![Skärmavbild 2024-05-03 kl  14 57 05](https://github.com/HGustavs/LenaSYS/assets/129263915/9cd1fef1-b282-4cfc-9c5a-2b7a15ab161a)

a22samka:
![image](https://github.com/HGustavs/LenaSYS/assets/129301680/97bcd89b-71b6-43ed-8f82-763318dfabb3)
Fixed another error aswell

PR #15552, opened 2024-05-08
G3 2024 v6 #15341

a22samka:
Fixed error by including ace.js in the fileed.php file. Also fixed error in navheader.


a22karja (approved):
Cant find anything breaking

PR #15554, opened 2024-05-08
G1 2024 v6 #15032

a22ricni:
Can now change the size of the note element and the top right corner should not be able to resize weird.


a22solmi (approved):
Looks great. Corner stays the same size regardless of resizing now.

PR #15556, opened 2024-05-08
background overlap fix #15555

### LINKED ISSUE #15555: overlap on entity creation
## Description
Background objects such as the UML SuperState can only overlap with other SD elements. Background object should only check for overlap with other background objects.

## Job
Change overlapping to only check background objects with each other.

a22solmi:
UML super state and loopOrAlt can no longer overlap. They can overlap with all other elements.


a22vioja (approved):
I have tested all the elements to place on the Superstates and they all work. 
Overlapping with other elements or overlapping between Superstates is still not possible, which is good.

Screenshot 2024-05-10 at 09 15 00

PR #15559, opened 2024-05-09
Added ESC to go back in accessed.php

d22mikwe:
Fixed current iteration of keyup for enter and esc. 
As the popups already has a close function attached to them we only go back if none are open.
Only current popups/modules in accessed are "createUser" and "addUser".
"removeUser" and "createClass" will be added later. These should be inserted into allPopups array when implemented.

For testing:

### DIFF
--- a/DuggaSys/accessed.js
+++ b/DuggaSys/accessed.js
@@ -930,20 +930,26 @@ document.addEventListener('click', function(e){
 //----------------------------------------------------------------------------------
 // Keyboard shortcuts - Edit functionality in the accessed table
 //----------------------------------------------------------------------------------
-document.addEventListener("keyup", function(event)
+document.addEventListener('keydown', function(event)
 {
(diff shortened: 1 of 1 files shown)


a22natth (approved):
## Testing 
The testing was done on accessed.php to which the functionality works as intended. When opening a modal and pressing Esc, the modal is closed but the user remains on the accessed.php page. Furthermore, if a popup window is not open, the user is able to go back to sectioned.php. The code was also viewed in VS Code and looked good. 

## Notes
 When  pressing Esc to close a modal in sectioned.php, the user is transported to courseed.php instead of just closing the popup window. The same issue occurred in the weekly branch, therefore, it is out of the scope of this specific pull request. This might be a future issue or something worth checking out.

d22mikwe:
Thanks for heads up, will create separate issue for sectioned.php

PR #15560, opened 2024-05-09
Added ESC to go back in fileed#15439

d22mikwe:
Fixed current iteration of keyup for esc and control.
Integrated current filePreviewClose() call when pressing esc and popup is open.
Current popups are when adding a file, previewWindow (Edit file icon) and fileViewPreview (clicking on the file name).
Added intended integration with "x" key. this was not implemented but previous devs had intended for it.
Also fixed previous indentation.

For testing:

Also try closing fileView with x key. It should not produce duplicates in view if opened again.


b22marah (approved):
LGTM! Functionality also works as intended, good job!

PR #15561, opened 2024-05-09
Added ESC to go back in codeviewer.php #15438

d22mikwe:
Implemented keyup event for esc.
Integrated a check if popups are open in codeviewer.
Should work on all popups.

For testing:

--------------------------------------------------------------------------
Problem when pressing esc while previewWindow loginBox is open. 
Its container "previewWindowContainer loginBoxContainer" is set to none while "previewWindow loginBox" is still not changed from block to none. Circumvented this in this pullrequest but this will be looked into in another issue.

### DIFF
--- a/DuggaSys/codeviewer.js
+++ b/DuggaSys/codeviewer.js
@@ -5099,10 +5099,10 @@ function toggleTitleDescription(toCheck){
 	}
 }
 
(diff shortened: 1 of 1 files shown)


a22maxra (approved):
Works as intended

Issue #15392, opened 2024-05-06
Cannot set default version when creating new course version


d22mikwe:
Users are experiencing problems when creating a new course version.
When checking the box to use the new copied course version as default the course version is not created.

Look into whats causing this and solve it.

a22maxra:
I'll start doing commits tomorrow after the weekly merge

a22maxra:
![image](https://github.com/HGustavs/LenaSYS/assets/98206230/a523988c-31ef-4dd0-9c0b-5cd40e5b09e1)

![image](https://github.com/HGustavs/LenaSYS/assets/98206230/fbf0c178-d850-42c0-94eb-49e324d10ca6)

Despite a success message when checking the "Change this to default version" the is no new course version

a22maxra:
Even the regular new version of a course does not seem to work even without checking the box

a22maxra:
I'm trying to see why new versions no longer get created in either case, when using a switch with cases the case is not triggered for some reason, but when switching back to a if a new version is still not created even though the request is sent

a22maxra:
The problem is in when the "CPYVRS" argument is passed to `courseedservice.php`, when "NEWVRS" is passed a new version is created, so looking at logic in in `courseedservice.php` when "CPYVRS" is passed.

a22maxra:
There was a missing $ sign in front of the _allOperationsSucceeded_ variable in `courseedservice.php`

a22maxra:
There is some problem when copying certain groups where the progressBar class styling doesn't show

a22maxra:
There is an error message when copying a course and entering a dugga

![image](https://github.com/HGustavs/LenaSYS/assets/98206230/7a23cac8-99af-482b-a303-960749d6755b)

PR #15566, opened 2024-05-10
Copying and setting default versions #15392

a22maxra:
You can now copy and set default courses

image

There was a missing $ in front of a variable

### DIFF
--- a/DuggaSys/courseedservice.php
+++ b/DuggaSys/courseedservice.php
@@ -459,7 +459,7 @@ if (checklogin()) {
 					}
 				}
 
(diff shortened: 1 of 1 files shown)


a22maxra:
I just created this version from version - 1337 in the testing course and it works
image

a22maxra:
However styling is missing when copying some things will check it out

image
image

a22maxra:
The weeks do show when choosing a bigger span for the course like 2024-01-01 to 2024-03-01 and the progress bar depends on the _duggalength_ variable but the copying of courses is working

Issue #15537, opened 2024-05-08
Bad dropdown design in accessed.php


a22moamy:
How to reach `accessed.php` and find the issue

1. Go into a course
2. Press the "Give students access to the selected version"-button:
3. Hover over the search bar

A dropdown is displayed with scrollable text, but it is impossible to scroll:
![Skärmavbild 2024-05-08 kl  14 28 57](https://github.com/HGustavs/LenaSYS/assets/129263915/8e2ef125-37e2-4977-a703-4c3146ed558d)

**Please implement a new design of this dropdown which does not require scrolling. The new design should have a little padding and rounded corners. The light mode version of the dropdown should have the same light gray seen in the following screenshot:**
![Skärmavbild 2024-05-08 kl  14 25 01](https://github.com/HGustavs/LenaSYS/assets/129263915/082ea3f2-c4b0-4e46-824e-43293839f6a5)

The dark theme color for the dropdown is fine as is and should **not** be changed.

PR #15568, opened 2024-05-10
G3 2024 v6 #15537

a22karja:
made the pop up on accessed.php search bar to change size depending on what is put into it. No more scroll. Changed the lower corners to be rounded and it now matches the color of the table shown

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -4528,7 +4528,6 @@ div.reset-pw:hover{
 
 .dropdown-list-container {
   padding: 8px;
-  width: 200px;
   color: var(--color-text-list);
   font-size: 12px;
   overflow-y: auto;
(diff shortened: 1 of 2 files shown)


a22karja:
and changed the instruction to actually use an example that would work

a22willi (changes requested):
2024-05-1011-03-44-ezgif com-video-to-gif-converter
It still does not work quite right, see the attached gif.

a22willi (commented):
Works better now, but the info popup is not aligned properly with the search bar. @a22moamy What do you think?
image

a22moamy:
> Works better now, but the info popup is not aligned properly with the search bar. @a22moamy What do you think? image

Should be fixed.

a22willi (approved):
Looks good.

Issue #15086, opened 2024-04-29
Line appears over the cursor


a22vioja:
## Description

When connecting two objects, the line currently appears over the cursor. 
Instead, it should run alongside the cursor.

<img width="887" alt="Screenshot 2024-04-29 at 14 06 20" src="https://github.com/HGustavs/LenaSYS/assets/129263559/1e505e7f-f6a4-4afc-a5c6-0b5b032e0915">

## Job

Make that cursor would be beside the line.

a22hanfa:
This problem is actually a problem for all entities, so all entities are not under the mouse pointer.

a22vioja:
> This problem is actually a problem for all entities, so all entities are not under the mouse pointer.

Yes, it applies to all entities. There's an ER example of an issue in the image, but the problem described in the issue is that it occurs when connecting objects.

a22hanfa:
> Yes, it applies to all entities. There's an ER example of an issue in the image, but the problem described in the issue is that it occurs when connecting objects.
 
 Yes! Sorry, was just talking to myself haha! Making some notes on the page

PR #15571, opened 2024-05-10
G1 2024 v6 #15086

a22hanfa:
**Original task:**
>Make that cursor would be beside the line.

**The finished result:**

https://github.com/HGustavs/LenaSYS/assets/129273186/2e2d3417-1f5d-4553-a43e-3f964593fb47

The result has been achieved by having a boolean value that says if it's drawing or not. While it's drawing, the coordinates for x & y should follow the cursor.

**To test:**
Make sure that the cursor follows the draw being drawn, that the sit besides each other.

### DIFF
--- a/DuggaSys/diagram/draw/line.js
+++ b/DuggaSys/diagram/draw/line.js
@@ -1,13 +1,15 @@
 /**
- * @description Constructs an string containing the svg line-elements of the inputted line object in parameter.
+ * @description Constructs a string containing the svg line-elements of the inputted line object in parameter.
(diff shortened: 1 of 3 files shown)


a22jacka (approved):
The line now follows the cursor much better than it did before. It could still be a bit better when above the element but it's still good

https://github.com/HGustavs/LenaSYS/assets/129258593/ff405fe6-a216-4f29-bf16-5f256bcaf69c

a21liltr:
Closes issue #15086

PR #15572, opened 2024-05-10
When editing a course it will require you to input a github key

a22oscja:
It was not actually requiring you to insert a github key, it stemmed from the bfs function reaching the data limit. 

Since there reasonably is not much I can do about the data limit as well as the bfs function being worked on in another issue I just changed the error message to a more descriptive and made it close the window upon it being sent.

Dont think this really needs testing, but simply reach the data limit and insert only a link for a course that you do not already have files downloaded for.

### DIFF
--- a/DuggaSys/courseed.js
+++ b/DuggaSys/courseed.js
@@ -61,7 +61,8 @@ function updateCourse() {
 			//Check FetchGithubRepo for the meaning of the error code.
 			switch (data.status) {
 				case 403:
-					toast("error", data.status + " Error \nplease insert valid git key", 7);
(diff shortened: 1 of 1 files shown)


b22chran (changes requested):
The new error-message is more descriptive than before, and closes the window instead of leaving it open with errors. Not sure how much further the issue can be solved from here, as described in the pull-request, the data limit is the problem and looks like it's not solved here. 
The code looks like it follows the standard and is nothing I can see that needs to be change.
I will approve this and then the group leader can make a decision if it the solution meets the requirement for the problem that was issued from the beginning.

b22chran (approved):
Chose the wrong radio-button. Was meant to approve, not request a change...

Issue #14813, opened 2024-04-22
Expand toast to offer confirmation


a22willi:
The current toast implementation implemented in #14682, currently only supports informational messages, expand upon this so that a yes/no choice can be made. 

Example:
```javascript
// The current toast is called like this:
toast(type, informational message, duration);

// Allow for a confirmation instead, that is using the same code like this:
confirmation(type, informational message); 

// Or if it would be possible, using polymorphism 
answer = toast(confirmation, informational message)
```

PR #15574, opened 2024-05-10
Added a confirm toast to the toast roster G3 2024 v6 #14813

a22edvbu:
Now there is a confirm toast available to be used among the rest of the toasts. I did this through adding additional elements to toastContainer that shows the two buttons (thanks Moa). To pass the arguments for the buttons, the last two arguments in the toast functioned are used for this (arguments and arguments2). 

image

To call this function, you just add the following code to a php-file that implements toast.php (for example courseed.php):

``` javascript
toast("confirm", "Are you sure you want to do that?", 0, "", "");
```

### DIFF
--- a/DuggaSys/courseed.js
+++ b/DuggaSys/courseed.js
@@ -23,7 +23,6 @@ $(document).ready(function () {
 	});
 });
 
-toast("confirm", "Test", 0);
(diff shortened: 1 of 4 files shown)


a22danwe (approved):
looks good and scales with different screen sizes

PR #15577, opened 2024-05-10
fixed padlocks #15267

a22jacka:
The padlocks now don't expand the div, have the same position in light and dark mode, and is visible in light and dark mode


a22vioja:
Closes issue #15267

a22solmi (approved):
Padlock no longer displaces when switching to dark mode. Arrows stay the same even when locking an object. Looks great!

PR #15579, opened 2024-05-10
G2 2024 v6 issue#15194

b21isabe:
The test updateCourseVersion_ms_test.php has been created.

The microservice updates a couple of things but the only change that can be seen in the output is the versname. I used the service to change the versname of the Testing-Course which is seen in the output.  After the test is run the versname is restored to the old one.


a22marca (approved):
The test works without any issues.

Merging this.

PR #15550, opened 2024-05-08
G3 2024 v6 #15041 is partly fixed as files are fetched and downloaded asynchronously

a22ahmom:
The main problem of this issue is to fetch and download all files and directories from the course URL that is saved in download github repo window asynchronously. However, some parts of this issue have been fixed as files are fetched and downloaded locally asynchronously. The tables gitRepos and fileLink are also populated with relevant data but are incomplete as not all files within directories are fetched properly and downloaded. So the main thing to try to fix is that directories should be fetched and downloaded associated with the URL provided. Right now this is not working. However, it all comes down to the bfs function. Initially, this function used a synchronous approach as requests for files and directories were made synchronously, meaning one at a time. This resulted in longer waiting times. 

What made this function synchronous was the file_get_contents( ) function. However, to make the function asynchronous cURL was used. Specifically curl_multi_init( ) and some other functions related to it. View the source code for a detailed description. Essentially some modifications are needed in the bfs function to also fetch and download directories. View the images below to understand properly the functionality of the bfs at the moment.

The path where the content is downloaded. C:\xampp\htdocs\LenaSYS\courses\$cid\Github
Screenshot 2024-05-08 154732

DB Browser (SQLite) the gitRepos table
Screenshot 2024-05-08 154848

PhpMyAdmin the fileLink table
Screenshot 2024-05-08 154912

Issue #15041, opened 2024-04-26
Research possibility for asynchronous fetching of github repo


d22mikwe:
The current iteration of fetching a github repo to a course seems to be synchronous, thus making the webpage freeze until it has been completed. Seems to be the bfs() function that is synchronous. Correct me if im wrong.

Look into what parts of the whole fetch process is synchronous.
Research if it is possible to refactor the current fetching to be asynchronous in its current php file.
If not, is it possible to integrate the function partly or all of it into a js file where asynchronous requests are natively supported?

Relevant issue:
- #14391

a22ahmom:
this one got reverted

PR #15580, opened 2024-05-10
G3 2024 v6 #15041 (second attempt) bfs is partly fixed

a22ahmom:
The issues raised in the previous PR should be fixed now.

Previous PR https://github.com/HGustavs/LenaSYS/pull/15550


a22ahmom:
Seems like creating a new PR does not solve the issue with 95 files.

PR #15583, opened 2024-05-10
migrate showDuggaService to microservices by changing the service redirections in dugga.js

a22albjo:


c20juler:
Reviewed and tested. showDuggaService is successfully migrated to microservices.

PR #15586, opened 2024-05-10
Updated thrashcan and deletion functionality

a18chrkr:
When the user clicks the trashcan item, the "confirmationBox" opens up. I've implemented so the clicked dugga get pushed to the "selectedItemList" when the user clicks "Yes" in the confirmation window. For safety reasons have I included an if-statement that don't push the dugga-id if it's already inside the "selectedItemList" array.

When yes is clicked, the function "deleteItem" deletes all the items in the array "selectedItemList".

I have left some commented code that could be helpful for testers. If everything work properly the commented code should be removed.

Commented code is on line: 650, 651, 653, 671 and 672

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -646,7 +646,11 @@ function showSaveButton() {
 
 // Displaying and hidding the dynamic comfirmbox for the section edit dialog
 function confirmBox(operation, item = null) {
+
+  // console.log("Clicked id" + active_lid)
+  // console.log("Selected IDs:" + selectedItemList)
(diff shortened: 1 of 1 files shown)


a22chrfa (changes requested):
Doesn't appear to work. I can see the item disappearing for a second before returning to the row after deletion.

I checked the code that was commited and added a log to it. I can see that the item is indeed added to the selectedItemList array, but the changes to the database is never made. So it seems to be some other function at fault?

a22moamy:
Issue is currently blocked by [#15632](https://github.com/HGustavs/LenaSYS/issues/15632)

Issue #15504, opened 2024-05-08
validateSectName() from sectioned.js is called excessively


a22moamy:
`validateSectName()` from `sectioned.js` is called everytime you move your mouse in a list entry (dugga/code example etc) form. You can test this by adding this simple debug code to the function:

```
let i = 1;
function validateSectName(name) {
     console.log("validateSectName() has run "+i+" times.");
     i++;
     . . .
}
```

Then, go into a course (`sectioned.php`) and press the cog wheel for a list entry (dugga/code example etc), like so: 
![Skärmavbild 2024-05-08 kl  10 18 53](https://github.com/HGustavs/LenaSYS/assets/129263915/093fffa7-790d-4a55-903f-cc3ccb7b6c4b)

You will find that `validateSectName()` is run every single time you move your mouse. **Find a way to eliminate the excessive calling of the function. In your fix, the function should only be called when the "Name"-input is altered.**

a22moamy:
This one is still fixed

PR #15590, opened 2024-05-10
G3 2024 v6 #15504

a22danwe:
the issue was that the validation function was called multiple times in both editSection and sectionname. i removed the calls to the function in editSection. 

testing:
add this code into to the validateSectName function in sectioned.js.
let i = 1;
function validateSectName(name) {
     console.log("validateSectName() has run "+i+" times.");
     i++;
     . . .
}

then press the cog wheel on and go at it.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -3624,6 +3624,12 @@ function validateSectName(name) {
     return false;
   }
 }
+let i = 1;
+function validateSectName(name) {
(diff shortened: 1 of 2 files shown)


b22marah (changes requested):
Code looks good, but there's an event checking the validation when hovering over the "save button", this should probebly be removed! 👍

b22marah (approved):
LGTM, functionality works as expected, well done!

PR #15592, opened 2024-05-10
G2 Merge issue#15378

a22marca:
Created the test for copyCourseVersion_ms.php. The test do not work for courseedservice.php.


a22iliru (approved):
Works :)

PR #15595, opened 2024-05-10
Created new test for the microservice changeActiveCourseVersion_cours…

a21sahfa:
…eed_ms.php

i have tested and all 3 tests passed. can be tested here http://localhost/LenaSYS/DuggaSys/tests/sectionedService_test.php


a22marca (changes requested):
This test should be for changeActiveCourseVersion_courseseed. Furthermore, the test should be added into its own file in LenaSYS/Dugga/tests/microservices/courseedService/changeActiveCourseVersion_courseseed_ms_test.php

Issue #15553, opened 2024-05-08
No keybinds on Test case, Load and Save


a22vioja:
## Description
There is no keybinding on Test case, Load and Save.

<img width="356" alt="Screenshot 2024-05-08 at 15 57 40" src="https://github.com/HGustavs/LenaSYS/assets/129263559/2e96061e-14da-41a0-8969-342ac6c1f4c4">

## Job
Make keybinding:
Test case: ctrl + t;
Load :  ctrl + l;
Save: ctrl + s;

PR #15596, opened 2024-05-10
G1 2024 v6 #15553

a22ricni:
Added the new keybinding for save diagram, load diagram and test case. The keybinding is CTRL + S for save diagram, CTRL + L and CTRL + ALT + T for test case. Needed to have CTRL + ALT for test case because if I use CTRL + T a new tab pop out.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -794,6 +794,21 @@ document.addEventListener('keydown', function (e) {
         }
     }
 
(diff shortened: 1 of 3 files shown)


a23ponek (approved):
The keybinding for save diagram works after changes. The image below shows how the keybinding for creating a testcase with ctrl + alt + T works for a state chart example.
Screenshot 2024-05-13 103941

PR #15598, opened 2024-05-10
G2 2024 v6 issue#15191

a22marca:
Created the test for createNewCourse_ms.php
versions is included in the filter because it was not filtered out because that array also contains, e.g. "coursename".
Cid is not included in expected-output or the filtered values because it is a autoincrement column in the db, which means a user needs to reset the db before running the test. Coursecode and name should be sufficient.


a22iliru (approved):
Works

PR #15602, opened 2024-05-11
Scale sequence actor on the height aswell

a22robko:
We have committed the task again because it was not merged on Wednesday, and the code/file was changed, so we had to redo the task with diffrent type of codes.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1103,15 +1103,51 @@ function mmoving(event) {
             deltaY = startY - event.clientY;
 
             // Functionality for the four different nodes
-            if (startNodeLeft && (startWidth + (deltaX / zoomfact)) > minWidth) {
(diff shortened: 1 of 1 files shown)


b22samer:
b22samer and a22robko redid the issue as it was not merged for some reason, we do not know why. We adjust new x and y values as there was new type of codes. In this commit, we have only changed startNodeLeft and startNodeRight as the others did not need to be changed according to the instructions. We tried to change the other four nodes, but got new information yesterday, that we did not need to change their functionality. So the four up nodes have not been changed.

a22vioja:
Closes issue #14857

a22solmi (changes requested):
This only scales sequence actor, the sequence object has the exact same problem. Initial and final states might be outside the scope of this issue but they also have the same problem.
Resizing works well on actor.

Issue #15062, opened 2024-04-29
Remove as much inline css as possible from courseed.php


a22moamy:
Currently, there is a lot of unnecessary inline css in `courseed.php`. Having inline css and external css for the same website makes development harder, and thus it should be separated as much as possible into the external css files. Inline css is especially unneccessary for elements with ids and often unnecessary for those without them as well. **Please move as much inline css as possible from `courseed.php` to the external css files.**

PR #15603, opened 2024-05-11
fixed all style in courseed.php to go in to style.css #15062

SleepyBearIV:
moved all "style='X' " from courseed.php to style.css, check that everything visually works as normal. 
if you fined spelling mistakes or better suggestions for naming thing add that to.


d22isaka (approved):
Looks good. The courseed appears the same (atleast for me) after the changes. =)

Issue #15562, opened 2024-05-09
Add functionality to "Add User" && "Remove User" button


d22mikwe:
After #15491 functionality should be added.

Intended functionality is to be able to search for all users in db table user in the dropdown menu.
Then when a user i chosen and save button is pressed this user should be added to that course in db table user_course.
Also intended functionality is to remove users using the same dropdown list.
May split this into 2 issues

Add functionality for this.

PR #15606, opened 2024-05-12
Add functionality to "Add User" && "Remove User" button #15562

d22mikwe:
Whats done:
- Dropdown list in both Remove user and add user is filled with all users.
- Add User now adds a user to the course without having to create one again. 
- Remove User now removes selected user from that course.
- Changed so readaccess users is seen is the list and not just teachers.
- Dynamic tooltipTerm for input validation.

For some reason the Term element has to be in Remove User popup. I dont know why but when its removed the dropdown options are not shown. Will look into this

Todo in future issues:

- Add Classes?
- Add Groups?
- Maybe update Term value if user already exists in course while trying to Add User.

Testing:
All users should show in both dropdown lists (Add User and Remove User)

Try adding any user into any course. Then reload the page and it should show up as a list entry.

Try removing a user from a course. It should disappear from the list entries

### DIFF
--- a/DuggaSys/accessed.js
+++ b/DuggaSys/accessed.js
@@ -134,7 +134,58 @@ function hideRemoveUserPopup() {
 //----------------------------------------------------------------------------
 //-------------==========########## Commands ##########==========-------------
 //----------------------------------------------------------------------------
-
+function addUserToCourse() {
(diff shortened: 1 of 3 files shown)


a22edvbu (approved):
Looks good to me. Great work!

PR #15609, opened 2024-05-13
Implemented course directory creation for new installer

a22willi:
# How to test?
1. First change the file `LenaSYS/newinstaller/tools/install_engine.php` to contain the proper pdo authentication settings. 
image

2. Run the code by navigating to [localhost/LenaSYS/newinstaller/installer.php](http://localhost/LenaSYS/newinstaller/installer.php). The output should look something like this: 
image

3. You can ensure that everything works properly by deleting the course directory `LenaSYS/courses`. Then run the installer again. If the course directory is created again, it works as intended.

### DIFF
--- a/newinstaller/installer_ui.php
+++ b/newinstaller/installer_ui.php
@@ -34,6 +34,7 @@
 				add_test_data: 'true',
 				add_demo_course: 'true',
 				add_test_course_data: 'true',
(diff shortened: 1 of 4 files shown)


a22oscja (approved):
works as described

a22maxra (approved):
Works as intended

Issue #15507, opened 2024-05-08
Toast icons not working for new code examples after saving template


a22moamy:
1. Create a new code example
2. Enter the code example (click on it)
3. Select a template and save
The toast icons will stop working and look like this:
![Skärmavbild 2024-05-08 kl  11 01 41](https://github.com/HGustavs/LenaSYS/assets/129263915/565d9fef-28a8-4a92-8523-dd7030954582)

**Please fix this.** A cheap solution is that it can likely be fixed by adding another link to Google icons in `codeviewer.php`. 

The link to Google icons:
```
<link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Material+Symbols+Outlined:opsz,wght,FILL,GRAD@24,400,0,0" />
```

a20jonda:
This got reverted

PR #15612, opened 2024-05-13
Added reference to icons. g3-2024-v6-#15507

a20jonda:
image


a22moamy (approved):
Looks good and simple

PR #15615, opened 2024-05-13
G2 2024 v6 issue#15611

a22marca:
I've made deleteCourseMaterial into a function that retrieveCourseedService is using. This is because retrieveCourseedService was already courses marked for deletion, making the deleteCourseMaterial microservice redundant before.


a22iliru (approved):
Good

Issue #15581, opened 2024-05-10
You can no longer successfully edit and update a course.


a22moamy:
You can no longer edit and update a course. Trying to edit a course will just result in errors and nothing will be updated. 

When editing a course in `courseed.php`, errors are generated in the console and in a toast.

The console error:
![Skärmavbild 2024-05-10 kl  12 57 32](https://github.com/HGustavs/LenaSYS/assets/129263915/0881c0e5-c51d-4570-bebd-828c82f94550)

The toast:
![Skärmavbild 2024-05-15 kl  14 59 20](https://github.com/HGustavs/LenaSYS/assets/129263915/77972f15-c959-481d-8797-a1dceafb9083)

A video displaying the toast error:
https://github.com/HGustavs/LenaSYS/assets/129263915/69ef2449-2535-484e-9ba4-d265dd8c9ddf

LenaSYS will complain about "Something went wrong with updating git token and git URL..." even though nothing has been tinkered with the token and URL.

**Fix the broken functionality of editing a course to actually work again. Also make sure no new errors are generated.**

PR #15618, opened 2024-05-13
G3 V6 #15581 successfully edit and update a course

a22ahmom:
You should not see any errors in the console anymore since it has been handled and it should work to update/edit any course.


DuggaSys/courseed.js
a22danwe, line 1: put spaces before curly brackets to be consistent

a22danwe (changes requested):
no errors when editing or creating a course. just some minor consistency flaws in the code to be fixed.

a22danwe (approved):
LGTM. Nice work

PR #15619, opened 2024-05-13
Pressing ESC button does not turn element to "none" in markdown editor

a22oscja:
It was likely made this way to avoid the event firing for the window underneath, which would move you back to the sectioned page.
Now the one underneath has been given priority on checking for events and the eventlistener on the popup window itself is using the same function as the other buttons.

**testing:**

- go to any course with a codeexample (webbprogrammering for example)
- go to a codeexample
- click the edit button:
image
- press escape and check if it sets previewWindow loginBox to display:none


a22willi (approved):
It looks good on the website, but for me, only `previewWindowContainer loginBoxContainer` gets set to display none, `previewWindow loginBox` remains on display: block. But this does not seem to be a problem.

`previewWindow loginBox` does get set to display: none, if I use the close button, but not using ESC.

a22oscja:
I'm unable to get it to not work correctly, this together with it calling the same function as the close button (which from what i could see has no steps inbetween) leaves me unsure where to go from here.

d22mikwe:
Issue is scrapped

PR #15627, opened 2024-05-13
IE Inheritance ghost element appears under the object

a22robko:
Implemented a CSS styling to fix the issue where the Inheritance ghost element was appearing underneath other objects. By adding a z-index of 2 to the "Inheritance" element, it now correctly appears above other elements.


b22davbe (approved):
Looks good, the IE-inheritance ghost element appears over other placed elements. 
image

Issue #15565, opened 2024-05-10
Create element class


a22solmi:
## Description
A element object is used in several places but has no class. There are several global variables, such as UMLHeight, that could be replaced by class methods.

## Job
Implement a class for elements and make sure it is used where an object was previously used. Create class methods that could replace some global variables.

PR #15628, opened 2024-05-13
G1 2024 v6 #15565

a22solmi:
Added class for creating elements. Moved default values to class. 
Removed global height variables. Changed entityIsOverlapping to use DOMRect.

For testing: Try copy/pase, overlapping and general element use.


a22robko (changes requested):
I have tested this issue and everything seems to be working as it should, but when creating a line between class and inheritance, the text does not display the name "inheritance"? I checked the weekly branch as well, and it was not like this? Otherwise, it looks good.
image1: 
Screenshot 2024-05-13 at 11 59 58 
image2:
Screenshot 2024-05-13 at 11 53 32

a22solmi:
I have updated names for all elements. Messages should display correctly now

a22robko (changes requested):
I try to connect Class Relation and UML elements with lines. It looks like it can't be done, as shown by the error message in the screenshot: "Not possible to draw lines between: Class Relation- and UML-elements".

image1:
Screenshot 2024-05-13 at 13 43 36
image2:
Screenshot 2024-05-13 at 13 43 10

a22robko (approved):
I tested it again and it looks good, and checked the functionality it works as usual.
Screenshot 2024-05-13 at 15 41 41

a21liltr (changes requested):
Problems:
* Sequence Activation should be able to overlap on Sequence Actor and Sequence Object
* UML Super State can't be placed on top of other entities (but other entities can be placed on top of UML Super State if Super State is placed first)
* Same problem for Sequence Condition as UML Super State
*  Note element is not working, after which changing to other entities will not work either
* Reset view also not working

DuggaSys/diagram/classes/element.js
a21liltr, line 1: very nice

DuggaSys/diagram/defaults.js
a21liltr, line 1: good

a21liltr (approved):
Tested functionality of each placeable element as well as drawing lines between them. Works as before, nothing immediately standing out or broken, unless it already was before. Code also looks clean and well written.

PR #15631, opened 2024-05-13
Created test for updateFileLink_ms.php issue#15207

a22melst:
The microservice does not just handle database operations, but it also writes content to files. It then retrieves the new size of the file, which is then updated in the database. Since I also wanted one test for each kind (global, course local and version local) and no course local or version local files is created by the installer, I opted for creating test files before the test.


a22iliru (approved):
Works

PR #15634, opened 2024-05-13
Changed css for the loginbox

b22chran:
Changed the max width to fit content in the loginbox and set the logintable to have some margin to look a bit nicer. The only thing that is now is that on some places where the loginbox-class is used as css when the content inside is not filled wider than 200px (min-width) it get pretty small. Attached some pictures where it is changed with the new css.

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -1897,6 +1897,7 @@ div .navButt:hover {
   color: var(--color-text-list);
   min-height: 100px;
   min-width: 200px;
+  max-width: fit-content;
   max-height: 80%;
   overflow-y: auto;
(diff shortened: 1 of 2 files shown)


a22moamy (changes requested):
The login form looks great! Plus, as a bonus, you also solved another issue (logout form CSS was wonky too)!! However, like you said, the other forms using the same class now look a bit too skinny. **Please use a more specific selector with the fit-content to only affect the login and logout form.**  The login and logout form both have ids, so i think it should be possible.

a22moamy (approved):
Looks great! Good job 😄

Issue #15614, opened 2024-05-13
Dialogs fade in/out for every input change in "Edit course" and "New course" forms


a22moamy:
When changing an input in the "Edit course" and "New course" forms, the dialogs/tooltips will fade in and out at inappropriate times. Tooltips should stay visible for as long as the input is faulty, it should only fade in when the user first inputs a faulty value and only fade out when the user finally corrects it. If the user keeps changing the input without fixing it, it should just stay visible with no CSS animations. 

A video has been attached on the issue:
https://github.com/HGustavs/LenaSYS/assets/129263915/d3e21330-bd3b-438f-a869-676d7b06d55e

a22moamy:
Because of merge difficulties, this issue needs to be done again. This has nothing to do with the developer or the quality of the code they produced, it was strictly because of merge difficulties. I apologize for the circumstances.

PR #15635, opened 2024-05-13
Elements now correctly fade in and out.

a20jonda:
`$(messageElement).stop(true, true);` does not exist in other form validations and seems to be very useful in stopping spam causing repeat fade-ins and -outs

Affected forms: Edit Course and New Course in courseed.php (quickValidateForm in courseed.js)

Extra:
* Added an neutral class for when optional fields are empty.

Note that the save button does not get disabled due to it currently being dynamic.
image

### DIFF
--- a/DuggaSys/courseed.js
+++ b/DuggaSys/courseed.js
@@ -756,6 +756,7 @@ function elementIsValid(element) {
 			$(messageElement).fadeOut();
 			element.classList.add("color-change-neutral");
     		element.classList.remove("color-change-invalid");
+			element.classList.remove("color-change-valid");
 			return true; // Optional fields
 		}
 	}
--- a/DuggaSys/courseed.js
+++ b/DuggaSys/courseed.js
@@ -746,23 +746,25 @@ function elementIsValid(element) {
 	// Stop any ongoing animations
 	$(messageElement).stop(true, true);
 
-	// Reset initial validation styling and messages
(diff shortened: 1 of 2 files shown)


a22natth (changes requested):
## What works
When editing a course, the tooltip only appears when an invalid value has been typed into the field. Furthermore, if the value continues to be invalid as the user is typing, the tooltip remains and does not fade in/fade out weirdly. This means that as long as the value is incorrect, the tooltip will remain and not disappear/reappear during typing. 

Previously, when encountering a faulty value, every active tooltip would fade in/fade out at the same time (as seen in the video linked in the issue #15614). This was also tested, to which this problem seemed to be fixed. This means that animation will only be applied to the relevant tooltip, instead of every single one at the same time. 


## Problems
The functionality of the tooltips when creating a new course, seem to be acting strangely. It is as if the tooltips are not reacting to the specified fields but both. Here is a video referring to what I mean so you can fix it:

https://github.com/HGustavs/LenaSYS/assets/129295853/a728c9e3-172a-47b1-98d0-b49fa9132773

a22natth (approved):
After speaking with the group leader, it is determined that the issue has been solved. This is due to the animation being correct, although the validation as seen in the video in the previous comment is still a problem. But this problem is not part of this issue(which I thought initially),so great job!

Issue #15589, opened 2024-05-10
Change the icon for Sequence Object


a22vioja:
## Description 

There is an "eye" icon for the Sequence Object in the toolbar. It should be changed to Sequence Object element.
They are in a folder 2 ../Shared/icons/diagram_sequence_object.svg

## Job

Change the icon in the toolbar.

c22abdha:
The icon has been replaced in the diagram.php file.
By replacing
shared/icons/eye_icon.svg
to
diagram_sequence_object.svg

![image](https://github.com/HGustavs/LenaSYS/assets/129261791/fa27e85b-7629-4a16-9d6b-7daccb7a3536)

PR #15639, opened 2024-05-13
g1-2024-v6-#15589

c22abdha:


a21liltr (changes requested):
The icon changes back to the eye after it has been selected. The icon is only changed when inactive. This should not happen.

Make sure both the active icon and the one in the submenu is changed too.

https://github.com/HGustavs/LenaSYS/assets/129370135/cf73ecd7-f21d-403b-a98b-5e2a80b30c5d

a21liltr (approved):
Icon has properly been replaced everywhere necessary.

The icon can be seen both when active, and inactive in submenu as expected.

PR #15640, opened 2024-05-13
DeleteButton scales with size and zoomfact

b22davbe:
The deleteButton now scales with the size of the element and the zoomfact but with a max-size of 50 and a min-size of 15 (when zoomed out it becomes to small otherwise). i also placed the deletebutton just outside the edge of the bounding box instead of over it so it doesn't overlapp with the upper-right node that would sometimes result in deleting an element instead of resizing it.
Also made the code for calculating the bounding box shorter and more effective and hopefully made the code (just a bit) easier to read.


a22hanfa (approved):
**Task:**
> Make that X would scale with the size of an object.

**It works!** The X-button scales along with the object. Tested for different elements.

**Proof (Using the same scale):**
Screenshot 2024-05-13 at 16 11 11

Screenshot 2024-05-13 at 16 11 23

a22vioja:
Solved bu another pull request :  #15883

Issue #15193, opened 2024-04-30
Create new test for the microservice retrieveCourseedService_ms.php


f22linhe:
See parent issue: _Create new tests for the microservices created from courseedService.php https://github.com/HGustavs/LenaSYS/issues/15187_

a22iliru:
This is a big microservice that will require a lot of testing in the test file, will take time

a22iliru:
The querys return foreign key constraint error, not sure how to fix it yet

a22iliru:
The constraints are fixed, remaining is fixing expected output.

PR #15643, opened 2024-05-13
G2 2024 v6 #15193

a22marca:
Created the test for retrieveCourseedService_ms.php. The it uses the service getCourseed_ms.php, which retrieves the information from `retrieveCourseedService()`and returns it.


a22iliru (commented):
Works

Issue #15570, opened 2024-05-10
Min size prevents resize


a22solmi:
## Description
The new changes to min width/height causes the resizing to stop in both directions if one direction is at its minimum. While resizing diagonally down right, place the cursor above the element and then drag to the left. The element will not get wider until the cursor is once again below the min height.

## job
Allow resizing in one direction even if the other is at its minimum.

PR #15649, opened 2024-05-13
G1 2024 v6 #15570

a22solmi:
Separated left/right and up/down resize


DuggaSys/diagram.js
a21liltr, line 1: nice

PR #15659, opened 2024-05-13
Fix tooltip for "Edit section" form

a22danwe:
removed the inline styling and put it in style.css. and because im not allowed to change the css using class selectors i added 100px on the id dialog10 instead.

Testing:

go into a course and press the cog wheel on a moment:
bild

then add an "illegal" character in the Name inputfield:
bild


a20jonda (approved):
Looks good 👍

Issue #15475, opened 2024-05-07
A4 horizontal template is not size A4


a22vioja:
## Description

A4 horizontal template has not same width and length as the vertical one.

<img width="1228" alt="Screenshot 2024-05-07 at 14 49 11" src="https://github.com/HGustavs/LenaSYS/assets/129263559/38dd8b9a-748a-469f-9ab8-fd698c61a219">

## Job

Fix this so that the horizontal template would have same dimensions as vertical.

PR #15660, opened 2024-05-13
G1 2024 v6 #15475

a22hanfa:
**Original Task:**
> Fix this so that the horizontal template would have same dimensions as vertical.

**Result:**

https://github.com/HGustavs/LenaSYS/assets/129273186/5c8c54b3-d732-497f-9442-612091f87a91

Fixed the A4 error. Then fixed a bug, where when you first toggled A4, both vertical and horizontal a4 paper would be active

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -823,6 +823,11 @@
     z-index: -11;
 }
 
+
+#a4Rect, #vRect {
+    display: none; 
+}
+
 #a4Rect {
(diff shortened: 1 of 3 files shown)


a22ricni (approved):
Tested first at the week branch and find the problem when click on horizontal for the A4 template. After that I changed the branch and the horizontal A4 template is fixed and is the same size as the vertical A4 template. I also don't find any issue or error with these changes.

PR #15665, opened 2024-05-13
Fixed deletion of duggor

a18chrkr:
**Co-commit with @d22isaka** 

We have now fixed the deletion of duggor. The selected duggas get added to the "selectedItemList" array, from which all items are deleted and hidden when a user click "yes" in the confirmation box.  This happens in "deleteItem()" and in "deleteAll()" functions. "Undo" still works as usual.

The only bug that still persist is that duggas come back if the page is refreshed while the "undo" toast is still active. At the moment we have not found a way to combat this issue.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -647,10 +647,7 @@ function showSaveButton() {
 // Displaying and hidding the dynamic comfirmbox for the section edit dialog
 function confirmBox(operation, item = null) {
 
-  // console.log("Clicked id" + active_lid)
-  // console.log("Selected IDs:" + selectedItemList)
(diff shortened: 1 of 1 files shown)


b22chran (changes requested):
The function to delete entries in a course works and the undo-option still works fine. Like they say in their pull request, if the page is reloaded or if the user goes back a page at the time undo-toaster is active, the deletion will not be completed. In my opinion this looks good for now.

When watching the code, it's still some "console.log" entries left commented out that can be deleted. Change this and it is a good to go from me!

b22chran (approved):
Looks good to me now :)

PR #15666, opened 2024-05-13
Double arrow when creating Sequence diagram

a22robko:
The new code makes rotating arrow tips better by using 45-degree turns instead of the old 90-degree ones. This change makes it easier to adjust how arrow tips look, making them both work better and look nicer.


b22davbe (approved):
Looks good, there's no longer double-arrows in any directions when youre making sequence diagrams. 
image

PR #15673, opened 2024-05-13
G2 2024 v6 issue#15188

a22marca:


a22iliru (changes requested):
Test 3 (assertEqual): failed 

The test fails because it expects the numbers to be without strings. ex 2 instead of "2".

PR #15676, opened 2024-05-13
Sequence Actor name disappearing when selected

b22samer:
The text element for the Sequence Actor name was given a class of 'text' and to ensure that the name remains visible even when several objects is selected, I changed the class of the text element to 'nameLabel'. Then added a CSS rule to ensure that the text color is set to black (#000) and won't change.


a22robko (approved):
I have tested your issue, everything looks good. When you select, the 'name' does not disappear and the remaining functions work as they should.

Before:
Screenshot 2024-05-14 at 10 57 08


After:
Screenshot 2024-05-14 at 10 57 31

f23linan (approved):
Works well, name does not disappear when selected in both light and dark mode. 

image

image

PR #15708, opened 2024-05-14
Fixed positioning of "New Item"-dropdown

a22natth:
After fixing this issue, the dropdown now stays centered in relation to the “+” button for new items. The example below shows the menu at full screen.
Pasted Graphic 2
The image below presents a more narrow screen size.
Pasted Graphic 3

Lastly, the example underneath illustrates the menu with the smallest screen size. Additionally, z-index was added so that the menu does not fall behind the big “+” button at the bottom of the screen. Kindly, made sure that this works on both chrome and firefox.
Pasted Graphic 4


a18chrkr (approved):
The problem has been fixed gracefully. Good work!

PR #15709, opened 2024-05-14
create service getAccessedService_ms.php

a22albjo:
This service should run when the GET opt is called. The point of this service is to run the retrieve service without running an other microservice before, this has been implemented in sectioned and showDugga as well.

The service only gathers the opts, makes the necessary includes and runs the retrieve service.


a22iliru (approved):
Works

PR #15710, opened 2024-05-14
Sequence Actor name disappearing when selected

b22samer:
The text element for the Sequence Actor name was given a class of 'text' and to ensure that the name remains visible even when several objects is selected, I changed the class of the text element to 'nameLabel'. Then added a CSS rule to ensure that the text color is set to black (#000) and won't change.


b22jonan (approved):
made sure that this was still a problem in current weekly branch, after that i switched over to the branch for this issue. placed down 3 actors, and selected them using box selection, thier names were still there so id say this solution works

a23fanro (changes requested):
tested again, and as b22jonan said earlier, it looks like it's working.
Can be merged, but has conflicts so waiting for further instructions.
Screenshot 2025-04-08 130837

a23osckr:
Conflict fixed and base branch changed to G3-2025-V1

Issue #15462, opened 2024-05-07
Toggling theme in all pages except sectioned.php will cause errors


a22moamy:
Toggling the theme (dark/white) in all pages except `sectioned.php` will cause errors. This is because it calls `drawSwimlanes()` which is not defined in all other pages. Please remove the calling of `drawSwimlanes()` in all pages except for `sectioned.php`.

PR #15714, opened 2024-05-14
G3 2024 v6 #15462

a22diako:
An error appared in the console before where it said that 'drawswimlane is not defined' when toggling from light to dark mode. 

Therefor the function has been erased from all pages except sectioned and an if or else has been made where the function is only called if available.


a22moamy:
Too many files affected that have no real purpose, PR will be closed

PR #15717, opened 2024-05-14
The drawswimlane function should not be called 

a22diako:
An error appared in the console before where it said that 'drawswimlane is not defined' when toggling from light to dark mode.

Therefor the function has been erased from all pages except sectioned and an if or else has been made where the function is only called if available.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -2284,6 +2284,13 @@ function drawSwimlanes() {
 
   var momentno = 0;
 
+   // If the drawswimlane exist it will be called here
+   if (typeof drawSwimlanes === 'function') {
+  } else {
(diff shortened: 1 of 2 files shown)


DuggaSys/sectioned.js
```diff
@@ -2284,6 +2284,13 @@ function drawSwimlanes() {
 
   var momentno = 0;
 
+   // If the drawswimlane exist it will be called here
+   if (typeof drawSwimlanes === 'function') {
+  } else {
+    console.log("drawSwimlanes function is not available");
+  }
+
+
```
a22moamy, line 2293: This serves no purpose, remove

a22moamy (changes requested):
See comment

a22moamy (approved):
LGTM!

Issue #15428, opened 2024-05-06
Add new column with cogwheel icon for each list entry in accessed.php


d22mikwe:
After 
- #15425
- #15418 

Currently there are 5 columns for the listed user entries.

![image](https://github.com/HGustavs/LenaSYS/assets/129263393/357830e1-29e1-4e78-b24e-fd6ed2f3b180)

Your job is to add a 6th column named "Edit" to the right of "Password" column.
You should also add a new colum in each user list entry to the right of "Reset PW" button. It should align with the "Edit" Column.
For each user list entry there should be a clickable cogwheel icon in this new column.

Make sure the cogwheel icons are created for newly created users aswell.

The icon should be clickable and open a form. Content of the form will be in another issue.

d22mikwe:
Reverted after weekly merge

PR #15721, opened 2024-05-14
Added an edit column and option to users in accessed.php G3 2024 v6 #15428

a22edvbu:
Added an edit column and option to users in accessed.php. The cogwheel will show a box where users can be edited (this is yet to be implemented, but a placeholder will show).

image

To test this, you can use this link: http://localhost/LenaSYS/DuggaSys/accessed.php?courseid=2&coursevers=97732
Click on the cogwheel and see if the popup will show. This should also be tested for newly created users, which can be made from the fab button.


a22danwe (changes requested):
Part of the red border around the inputfield is hidden behind the next inputfield except for the last one.

bild

all should look like this:
bild

a22danwe (approved):
LGMT! great job you're the best <3

PR #15722, opened 2024-05-14
Updated options for IE-lines

b22davbe:
Changed the options available for IE-lines because the problem that the issue described shouldn't be possible in IE because its just the symbols that should be on the lines using this notation.
image


a22solmi (approved):
This removes labels from IE. Works as intended, all other options are still functional

PR #15723, opened 2024-05-14
Created ms retrieveAllCourseVersions_ms.php issue#15652

a22eriwi:
Created the ms file retrieveAllCourseVersions_ms.php because the code hadn't been included in any other ms yet. The code is taken from the sectionedservice.php file from row 912 - 921.
Have not found where it is used yet but still implemented it because now everything from sectionedservice.php has been or will be implemented shortly.


a22iliru (approved):
Looks good

PR #15726, opened 2024-05-14
Add handling of http status codes in the bfs()-method from gitfetchService

a22oscja:
Essentially the error logging that was put in when bfs() was made asynchronous was outdated. This meant that an issue with it that i had previously fixed was there again (it would simply not log if there was no key).

Other than that I also made the bfs function download more files, as it was very annoying trying to trigger the 403 error when it only downloaded roughly 8 files every time the button was pressed. For that I have simply made it run the download part until it has exhausted its queue by adding a while loop. Unsure if there is a better solution for this, but it is only tangentially related to this issue anyhow.

Testing:
download a github repo until you hit the rate limit (make sure it is actually downloading files and you do not have a key set, as otherwise you will not get anywhere)
see if there is a 403 error and also check the error log in LenaSYS/log/errorlog.txt (it will appear if you do not already have it there)

### DIFF
--- a/DuggaSys/gitfetchService.php
+++ b/DuggaSys/gitfetchService.php
@@ -143,7 +143,6 @@ function getIndexFile($url) {
 
 function bfs($url, $cid, $opt) 
 {
-    $counter = 0;
     // Different URL depending on operation
     date_default_timezone_set("Europe/Stockholm");
     if($opt == "GETCOMMIT"){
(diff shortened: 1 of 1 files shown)


a22oscja:
Also i'll add that the download takes quite a bit of time again (roughly 5 min, better than before atleast), so do not be alarmed and start spamming the button and such.
I also had a weird interaction on the button press where it triggered the 403 error but was unable to test that again, so look out for something like that too.

a20jonda (commented):
Error 403 Shows up after trying to downloading a github repo twice
image
Shows up as  "forbidden" in console.
image
Shows up as a token error in the gitErrorLog.txt
image

Folder and files successfully downloaded in first test course. Took some time before the second test course could download due to the 403.
image

Everything seems to work as intended however I'm unsure about the gitErrorLog.txt and the toast having different descriptions of the error value.

a22edvbu (approved):
Looks good to me. Great work!

Issue #15259, opened 2024-05-02
Sequence Activation overlaps with Actor element


a22vioja:
## Description

It is possible to place Sequence activation over the Actor stickman. It should not be allowed.
<img width="177" alt="Screenshot 2024-05-02 at 10 47 43" src="https://github.com/HGustavs/LenaSYS/assets/129263559/5ccc4f24-44f1-4ad2-8237-0597a283bfe0">

## Job

Find a way to prevent placing Sequence Activation over the stickman.

For clarification, it should not be able to be placed on the stickman,  but should be able to be placed on the dashed line underneath the stickman.

PR #15728, opened 2024-05-14
G1 2024 v6 #15259

a22cazlu:
Created function getTopHeight() that calculate the height of the top object of sequence actor and sequence object. Then there is now a minimum value for y in snapSAToLifeline().


b22jonan (approved):
tried placing a sequence activation on the stickman using the snapping. i was not able to place an activation on the stickman, aswell as the ghost snapping to below the stickman which is a nice feature.

b22jonan:
though it is still possible to get the stickman behind those objects by moving it behind them or by shortening it so that it ends up behind them, but thats another issue for another day

PR #15729, opened 2024-05-14
added functionality to open toolbar by rightclicking on the buttons instead of having to hold click

b22jonan:
added event listener to the buttons that call a function which opens the toolbar, also defered the calling of diagram.js so that its loaded at the end, otherwise this fix wont work.


a22hanfa (approved):
**This works as intended** 

Result:

https://github.com/HGustavs/LenaSYS/assets/129273186/1e9b2de5-785f-4f14-961e-4a217cb1b5c4

PR #15732, opened 2024-05-14
inital/final state resizing #14828

a22solmi:
Removed nodes that only resize in one direction. Modified delta to always resize both directions to the largest size.


a22robko (approved):
I have tested this issue and everything looks good. When you expand the height, the width also expands. I have checked the other functions as well and there seem to be no problems.
In the screenshot, you can see the before and after images of how they look when expanded.
Screenshot 2024-05-14 at 15 39 05Screenshot 2024-05-14 at 15 39 07

Result:
Screenshot 2024-05-14 at 15 43 18

PR #15733, opened 2024-05-14
Refactored two CSS class names

a22natth:
## How to test:

When testing, if possible, try both dark and light theme since the refactoring affects both css-files. Additionally, you might want to test it with different screen sizes to make sure the css has been correctly applied after refactoring. For each popup window, make sure functionalities such as tooltips, validation, date-selection and buttons still work as intended. Although, it is important to think about that some of the functionality is kind of broken to begin with so if possible, compare the pull request solution to the weekly branch. Below, the different pages are listed with their respective popup windows.  

courseed.php
- Edit server settings
- Create new course
- Edit course

sectioned.php
- Edit course version (cogwheel in navbar)
- New course version (plus sign in navbar)
- Github repo(download button to the right of lock)

duggaed.php
- Edit dugga (the window is actually for creating a dugga, press big “+” button

fileed.php
- Try all of the options in the picture below
Pasted Graphic 5

accessed.php
- Try all of the options in the picture below:

Pasted Graphic 6

Further testing surrounds manually searching for references of “loginBox” and “loginBoxheader” to see if I have missed any. Note that I have ONLY changed the id and class name of “loginBox” and “loginBoxheader”, and not “loginBoxtr” and similar values that end in different letters. To do this, use Ctrl + Shift + F, or Cmd + Shift + F on Mac.

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -1893,7 +1893,7 @@ div .navButt:hover {
   max-width: fit-content;
 }
 
-.loginBox,
+.formBox,
 .logoutBox{
   padding: 4px;
   background-color: var(--color-background-2);
@@ -2049,7 +2049,7 @@ div .navButt:hover {
   border-radius: 2px;
 }
 
-.loginBox textarea,
+.formBox textarea,
 #import {
   width: 100%;
   height: 200px;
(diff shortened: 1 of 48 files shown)


Shared/css/style.css
a22willi, line 1: Same here.

a22willi (changes requested):

a23elism (changes requested):
Whilst the refactored class names work and dont cause any issues in the respective branch, it causes conflicts in 4 files (resolving required)

Issue #15693, opened 2024-05-14
Create new test for the microservice removeListEntries_ms.php


a22marca:
See parent issue: [Create new tests for the microservices created from sectionedService.php](https://github.com/HGustavs/LenaSYS/issues/15681)

a22melst:
Leaving this issue for the group of next year.

PR #15734, opened 2024-05-14
G2 2024 v6 #15693

a22iliru:


b21isabe (changes requested):
Assert test 3 is currently failing. The test also has some problems that I have highlighted that needs to be fixed before merging.

d22marwi (changes requested):
No description, to vague. Remove?

PR #15735, opened 2024-05-14
G2 2024 v6 issue#15684

a22marca:


b21isabe (approved):
I have tested by running the tests and investigating the output. All tests work and pass, as expected. The output could be filtered a lot more but I understand that you might want to test that every aspect of the service is working, so we can leave it as it is.

Well done, merging.

PR #15743, opened 2024-05-14
Removed a call to function

b22chran:
Removed the call to the function "initInputColorTheme" from the validateSectName that is supposed to check the local storage if it is blacktheme or not. This seemed to get called wrong or something. After this was removed I wasn't able to get the fault of color for the input.

Check original issue for the sequence to manipulate the fault of input-background color.


a22edvbu (approved):
Works splendid! Great work! :)

Issue #15584, opened 2024-05-10
Highlight when editing list entries is faulty


a22moamy:
When creating a new list entry (dugga/code example etc) the highlight works great. However, it does not work as expected when editing and updating a list entry, as only a section of the row gets highlighted. Furthermore, when editing certain types of list entries, for example `tests` or `messages`, only the name is highlighted, not even the entire section for the name.

![Skärmavbild 2024-05-10 kl  13 16 06](https://github.com/HGustavs/LenaSYS/assets/129263915/0884d571-3ba5-4311-be08-5bfa1e717f05)

![Skärmavbild 2024-05-10 kl  13 15 55](https://github.com/HGustavs/LenaSYS/assets/129263915/659b5656-53dc-415c-927c-f4e6d0ea26de)

![Skärmavbild 2024-05-10 kl  13 15 32](https://github.com/HGustavs/LenaSYS/assets/129263915/fe2dccd2-5316-433d-a345-3a6ba761f032)

**Fix it so that the entire row is highlighted when updating a list entry, no matter the type of list entry.** Another issue will fix the color of the highlight (will be changed from purple to green), so leave the color as is.

PR #15744, opened 2024-05-14
G3 2024 V6 #15584 list entry is highlighted when updating/editing a list entry

a22ahmom:
The function setCreatedDuggaAnimation() seems to do the work. By observing how it worked when a new list entry was being created it appeared that the same functionality should be used.


a22edvbu (changes requested):
The animation works as intended. However, when editing a course and saving without changing anything, an error occurs. This problem is not present in the current weekly branch.

image

a22moamy:
Issue encountered is unrelated and has since been fixed in weekly branch.

a22edvbu (approved):
As of last merge of g3-2024-v6, the error no longer persists. Great work! :)

Issue #15651, opened 2024-05-13
 Fix deletecodeexample  test


a22iliru:

a22iliru:
I cant fix it because of:
Fatal error: Uncaught PDOException: SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near ':exampleid)' at line 1 in C:\xampp3\htdocs\LenaSYS\Shared\test.php:88 Stack trace: #0 C:\xampp3\htdocs\LenaSYS\Shared\test.php(88): PDOStatement->execute() #1 C:\xampp3\htdocs\LenaSYS\Shared\test.php(179): doDBQuery(Object(PDOStatement), 'UNK', 'UNK', 'UNK') #2 C:\xampp3\htdocs\LenaSYS\DuggaSys\tests\microservices\codeviewerService\deleteCodeExample_ms_test.php(34): testHandler(Array, true) #3 {main} thrown in C:\xampp3\htdocs\LenaSYS\Shared\test.php on line 88

Will be moving on to another issue

PR #15752, opened 2024-05-14
Fixed test for deleteCodeExample_ms.php #15651

a22melst:
The test cannot check whether the deletion of the listentry works because the retrieve service does not return any listentries.


a22iliru (approved):
Works

PR #15754, opened 2024-05-14
G2 2024 v6 issue#15753 fix getCourseVersions_ms.php

a22albjo:
Fix getCourseVersions_ms.php. This makes the retrieveSectionedService run as intended.


a22marca (approved):
This seems to fix the issues with `checklogin()` when using sectionedService microservices.

Issue #15662, opened 2024-05-13
Make input from "Edit dugga" form be adjusted to dark mode


a22moamy:
This issue is a little tricky to find, so here is a video showing how to cause it:
https://github.com/HGustavs/LenaSYS/assets/129263915/9a054213-ebfa-4b14-afb5-dbc6b38da417

1. Go into a course
2. Go into dark mode
3. Enter form to edit a dugga
4. Enter an illegal character in the first input, like # for example. 
5. Close/exit the form without saving it
6. Toggle to light mode
7. Enter form to edit a dugga
8. remove illegal characters in the first input, and make the value legal, "dugga 1" for example 
9. Close/exit the form without saving it
10. Toggle into dark mode
11. Enter form to edit a dugga and the input will be white instead of gray.

Fix the color of the input to match the others. The current (unwanted) appearance: 
![Skärmavbild 2024-05-13 kl  14 58 56](https://github.com/HGustavs/LenaSYS/assets/129263915/696fdfc7-a123-4c8a-805e-f543b1e71dc5)

a22moamy:
Because of merge difficulties, this issue needs to be done again. This has nothing to do with the developer or the quality of the code they produced, it was strictly because of merge difficulties. I apologize for the circumstances.

PR #15772, opened 2024-05-15
Removed call to function

b22chran:
Removed the call to the "initInputColorTheme". This function sets the black theme properties but it seems to be unnecessary though the black theme works as it should without this call.

Look in the issue (https://github.com/HGustavs/LenaSYS/issues/15662) to see how to test the fault of the input.


a22edvbu (approved):
The textfield is no longer getting the anomaly shown in the issue. Great Work!

Issue #15700, opened 2024-05-14
Create new test for the microservice updateListEntriesTabs_ms.php


a22marca:
See parent issue: [Create new tests for the microservices created from sectionedService.php](https://github.com/HGustavs/LenaSYS/issues/15681)

c22cargo:
It seems that "tabs" is always retrieved as NULL, even when the value is not set to NULL. This likely has to do with the retrieve MS, and it will have to be fixed before the expected-output can be updated.

c22cargo:
The problem above was fixed in #15748.

PR #15775, opened 2024-05-15
G2 2024 v7 issue #15700

c22cargo:
The test has been created. The MS needed to be modified since it previously updated ```gradesystem``` instead of ```tabs```.

The branch has been rebased to v7.


a22melst (approved):
Works

PR #15780, opened 2024-05-15
G2 2024 v7 #15697

a22iliru:


a22albjo:
Has this pr been reviewed before merge? The test doesn't seem to work and i remember reviewing an identical pr you made waiting for a change request, i think you should address that change request, you can find it here: https://github.com/HGustavs/LenaSYS/pull/15731

PR #15803, opened 2024-05-15
More space between numbers and icons

b22davbe:
I've set the offsetOnLine variable to 35 * zoomfact and added 10 to x if the line is vertical. This makes the numbers and symbols to not overlap eachother as long as you stick to normal UML-lingo (example: 1..*). If you would write an essay on the line it will still overlap.


a22hanfa (changes requested):
**It works!**

Screenshot 2024-05-15 at 14 31 04

However, the problem is that someone else has been has worked on the wrong branch that needs to be removed.

a22hanfa (approved):
Fixed. LGTM

PR #15807, opened 2024-05-15
Reimplemented fix 'Added reference to icons. g3-2024-v6-#15507'

a20jonda:
image


a20jonda:
Looks like this pull got included in #15809. Since this one is only 1 line, It will be better to just close this one

PR #15809, opened 2024-05-15
half Reimplementation of G3 2024 v6 15495 Updated margin-right and style for forms in 2* files

a20jonda:
Copy from #15612.
image

Copy from #15523. (Images are the same on this branch)
There was a conflict with the `Add user` form in accessed and all other forms in courseed- and sectioned-, duggaed.php. The solution was to accept and implement the style without the extra 10px margin. 

Extra: Added missing style class to submit button and missing body element in `Github Moment` form in sectioned.php which was more apparent without the 10px margin. ~~Changed flexwrapper to inputwrapper in `Edit Dugga` to ensure it implements the style of the other forms.~~

image
Note that `Add User` stays the same and will no longer conflict

image
image
image
image


a20jonda:
Looks like change for #15507 got accidently included in this pull

DuggaSys/sectioned.php
b22chran, line 1: A better way to get the styling done here might be a css-class instead of inline-styling. By the look of the code both deadlineminutes and deadlinehours use the same values and could be a class instead.

a20jonda:
image
image
image

b22chran (changes requested):
Good job with creating classes for the styling. The only thing is that the use of "!important" it's not good practice. Should be able to solve it without the use of it.

a20jonda:
I don't really know how to solve !important. Should I just place the classes above the others?

a20jonda:
Hmm, it automatically resolved the merge conflict, that interesting.

b22chran (approved):
Looks great now!

a22moamy:
Will be closed since one of these issues have been solved in another PR and I'm too scared to merge another fix in. Please make another pull request for just issue #15507. Not devs fault, accidentally made duplicate issue.

Issue #15473, opened 2024-05-07
At some screen sizes, "New item"-dropdown does not work due to faulty position


a22moamy:
At some (smaller) screen sizes, the "New item"-dropdown in `sectioned.php` will be slightly pushed to the right, disabling the user from actually selecting an option. 
It will look like this:
![Skärmavbild 2024-05-07 kl  14 28 18](https://github.com/HGustavs/LenaSYS/assets/129263915/037ae69b-dd1f-43d8-a117-b6e322a12128)

It should look more like this:
![Skärmavbild 2024-05-07 kl  14 36 47](https://github.com/HGustavs/LenaSYS/assets/129263915/1e1c2971-c979-439f-8546-47f321c7b728)

To solve the issue, maybe look at the "+"-button, which actually works well no matter the screen size.
The "+"-button in question:
![Skärmavbild 2024-05-07 kl  14 38 47](https://github.com/HGustavs/LenaSYS/assets/129263915/6596a1c7-a84e-40a5-bbf6-851bb4edc8fe)

PR #15811, opened 2024-05-15
Added changes from old branch to new branch

a22natth:
Added the changes from the old branch to the new branch due to problems with the weekly merge, which had resulted in reverting a few commits back (including the one for the issue #15473). Therefore a new branch had to be created which included the adjustments to which fixed the problem of the weird positioning of the dropdown for the new item "+"- button.

### DIFF
--- a/DuggaSys/sectioned.php
+++ b/DuggaSys/sectioned.php
@@ -186,8 +186,8 @@
 				<img src="../Shared/icons/desc_complement.svg" alt='Hide List Content' id="sectionList_arrowStatisticsClosed">
 			</div>
 			<div class='fixed-action-button2 sectioned2'  id="FABStatic2" style="display:none">
-				<input id='addElement'  type='button' value='+' style="margin-right:20px; top:-493px" class='submit-button-newitem' title='New Item' >
-				<ol class='fab-btn-list2' style='display: none;'  reversed id='fabBtnList2'>
+				<input id='addElement'  type='button' value='+' style="margin-right:20px; top:-493px; position:fixed;" class='submit-button-newitem' title='New Item' >
+				<ol class='fab-btn-list2' id='olFabBtnList2' reversed id='fabBtnList2'>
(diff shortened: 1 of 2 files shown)


DuggaSys/sectioned.php
```diff
@@ -186,8 +186,8 @@ function updateGithubDir($pdo, $githubDir, $lid) {
 				<img src="../Shared/icons/desc_complement.svg" alt='Hide List Content' id="sectionList_arrowStatisticsClosed">
 			</div>
 			<div class='fixed-action-button2 sectioned2'  id="FABStatic2" style="display:none">
-				<input id='addElement'  type='button' value='+' style="margin-right:20px; top:-493px" class='submit-button-newitem' title='New Item' >
-				<ol class='fab-btn-list2' style='display: none;'  reversed id='fabBtnList2'>
+				<input id='addElement'  type='button' value='+' style="margin-right:20px; top:-493px; position:fixed;" class='submit-button-newitem' title='New Item' >
```
b22marah, line 189: Fix spacing between attributes, id - type has 2 spaces for instance... Maybe not within the scope of your issue, but would prefer inline-css to be removed since input has an id, and could easily be moved to the css-file

Shared/css/style.css
```diff
@@ -7069,13 +7069,27 @@ only screen and (max-device-width: 568px) {
 
 .fab-btn-list2 {
   position: absolute;
-  padding: 18px 0px 0px;
-  margin: 0px 0px 0px -3px;
-  left: auto;
+  padding: 28px 25px 0px 0px;
+  margin: 0px;
+  left: -55px;
```
b22marah, line 7074: Up for discussion, but would like the use of REM instead of pixels for unit positionings > 4px.. Makes it more responsive

Shared/css/style.css
```diff
@@ -7069,13 +7069,27 @@ only screen and (max-device-width: 568px) {
...
+  height:fit-content;
+}
+
+#FABStatic2{
+  z-index:2;
+}
+
+@media only screen and (max-width: 430px) {
+  .fab-btn-list2 {
+    left: 55px;
```
b22marah, line 7091: Same here

b22marah (changes requested):
Functionality works as expected, and looks good. Besides small adjustments to the code everything LGTM!

b22marah (approved):
LGTM, well done! :)

Issue #15636, opened 2024-05-13
Implement credentialsmanager into newinstaller


a22willi:
Implement the credentials manager into new installer. Make the installer accept json data that specifies db credentials, and writes them to the coursesyspw.php file.

PR #15813, opened 2024-05-15
Merge g3-2024-v7-#15636 into g3-2024-v7.

a22willi:
This has already been merged and tested, but disappeared after fixing several issues related to the weekly merge.


a22moamy (approved):
Looks good!

Issue #15805, opened 2024-05-15
Fix getListEntries_ms_test.php


c22cargo:
The test is failing because no filter is used, which means that the time of creation is included, which will be different depending on when the database was installed.

Use the relevant filter-output.

PR #15822, opened 2024-05-15
G2 2024 v7 issue #15805

c22cargo:
The test should pass. If it doesn't, try reinstalling the database before trying again.


a22iliru (approved):
Works

Issue #15703, opened 2024-05-14
A4 invisible until zoom


a22solmi:
## Description 
When reloading the site the a4 does not show until vertical/horizontal button is pressed and zoom is used.
https://github.com/HGustavs/LenaSYS/assets/129150661/460feeea-4e19-4376-8e85-466648e22a46

## Job
Add an update to show the a4 directly when created. Also show the vertical a4 by default  when pressing the a4 template button in the options pane.

PR #15824, opened 2024-05-15
G1-2024-v7-#15703

a22cazlu:
updateA4Size() when click on vertical and horizontal. Also made so when the user click on A4 Template it will show a vertical A4 in the diagram.


a22solmi (approved):
A4 now shows when first selecting a4 template. Looks great

DuggaSys/diagram/toggle.js
a21liltr, line 1: Remove blank rows.

Issue #15736, opened 2024-05-14
StateMachine constructor types


a22solmi:
## Description 
The StateMachine constructor has the wrong types and variable names in its JSDoc. 

## Job
Change the types on the constructor to reflect the parameters

PR #15841, opened 2024-05-15
JSDoc update #15736

a22solmi:
Changed JSDocs and minor code styling issues


b22jonan (approved):
looks good, not much to test this since its mostly just comments and refactoring code to look good. since i can understand the changes and they specify things more clearly i think this is good 👍

PR #15843, opened 2024-05-15
Updated sectionedService.php

### LINKED ISSUE #15330: Update query on line 756 in sectionedservice.php
_sectionedservice.php_ line 756, query should look like this:

     SELECT boxid AS bid FROM box WHERE exampleid = :eid AND filename=:boxName;

f22linhe:
Updated the query on line 755. Closes #15330.

### DIFF
--- a/DuggaSys/sectionedservice.php
+++ b/DuggaSys/sectionedservice.php
@@ -752,7 +752,7 @@ if(checklogin()){
 					
 					//Check if remove box
 					}else if ($boxCount > $exampleCount){		
-						$query = $pdo->prepare("SELECT filename FROM box WHERE exampleid = :eid;"); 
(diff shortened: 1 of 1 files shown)


a22marca (changes requested):
The row that should be changed is incorrect in the issue description, probably because of changes in the file in earlier issues.

So the row that should be changed is in the query at line 770:
```
if($exist==false){										
    $query = $pdo->prepare("SELECT boxid AS bid WHERE exampleid = :eid AND filename=:boxName;");
    $query->bindParam(':eid', $eid); 
    $query->bindParam(':boxName', $boxName);
    $query->execute();
    $result = $query->fetch(PDO::FETCH_OBJ);
    $bid = $result->bid;

```
The query is missing the `FROM box` part.

a22marca (approved):
The functionality itself of adding a github repo url to a course and fetching the data isn't working so I can't really test if it is working. But that specific query is valid when performing in the mysql DB, so this is good to merge.

Issue #15838, opened 2024-05-15
Fix createListEntry_ms_test.php


c22cargo:
The tests are failing. Find out why and fix it.

Issue #15748, opened 2024-05-14
Fix "tabs" being retrieved as NULL


c22cargo:
The value of ```tabs``` from the table ```listentries``` is always retrieved as NULL, even when the value is not set to NULL. This most likely has to do with ```retrieveSectionedService_ms.php```.

PR #15845, opened 2024-05-15
G2 2024 v7 issue #15838

c22cargo:
The test was failing because the retrieve MS was updated in #15748.

Previously, "tabs" was always retrieved as NULL even if it had a set value. Since the test sets "tabs" to 0 the expected-output should be 0, not NULL.


a22marca (approved):
The tests are passing and working as expected, merging.

PR #15846, opened 2024-05-15
G3 V7 #15581 successfully edit and update a course. (Changed the code back to work again)

a22ahmom:
the code is now fixed since it was missing because of v6 merge problems


a22moamy (approved):
Looks good!

PR #15852, opened 2024-05-15
BFS function is now functioning properly when fetching and downloading files (the old code is implemented back)

a22ahmom:
this happened because of v6 weekly merge problems


a22ahmom:
For reference https://github.com/HGustavs/LenaSYS/pull/15582

a22maxra (changes requested):
For me fetching files does not work in the main branch g3-2024-v7
image

And in this branch g3-2024-v7-#15041 the same problem occurs. Someone else should test and see if the fetching is working?
image

Issue #15837, opened 2024-05-15
Page won't reload automatically when creating a new course


a22moamy:
See attached video:
https://github.com/HGustavs/LenaSYS/assets/129263915/8d87df34-d696-4324-8f91-cc89e64aa3f6

When you create a new course, LenaSYS is supposed to be automatically refreshed, enabling you to see the new course. This does not happen however, and can such be very confusing for the end-user. Please fix this.

PR #15859, opened 2024-05-15
Added an location.reload() to createNewCourse() to #15837

a22edvbu:
Courseed.php has not been reloading after creating a new course for some time now. To fix this, i added location.reload() to createNewCourse(). This location.reload() is inside of a setTimeout() to give the site enough time to save the new course. 

To test this, just create a new course and see if the site reloads and that the new course appear in the list.


a22moamy (approved):
Works!

Issue #15633, opened 2024-05-13
Login form uses ` <table>`  instead of ` <div>` 


a22moamy:
The login form uses `<table>` to build its structure instead of a `<div>` element with nestled divs like all other forms, please fix this.  To get started, inspect other forms in LenaSYS to see how they are built and try to implement a similar solution for the login form. 

OBS! only change the form in loginbox.php

PR #15862, opened 2024-05-15
G3 2024 v7 #15633

a22samka:
Changed all tables in loginbox to a div structure instead

### DIFF
--- a/Shared/loginbox.php
+++ b/Shared/loginbox.php
@@ -137,19 +137,15 @@
 				<div class='cursorPointer' onclick="closeWindows()" title="Close window">x</div>
 			</div>
 			<div style='padding: 20px;'>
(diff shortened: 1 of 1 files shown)


d22isaka (commented):
With the new loginbox there appears to be a new ''Sign in'' text above the username-field.
Not sure if intended as the issue states to only change tables to divs. Only 3 lines of code needs to be deleted if that's not the case.

Other than that, looks good and it works fine =)
Skärmbild 2024-05-16 094407

a22moamy (changes requested):
Fix the comment left by @d22isaka

d22isaka (approved):
Looks good to me! =)

a22moamy (approved):
looks good now

PR #15864, opened 2024-05-15
documenting zoom.js #15804

b22jonan:
added parameters where needed and added - optional to params that have default values


a23ponek (approved):
Changes look good and the new documentation for the changed functions etc seems correct to what it does.

PR #15865, opened 2024-05-15
Revert "Now saving db credentials when running installer"

a22moamy:
Reverts HGustavs/LenaSYS#15861


a22moamy (commented):
Changes requested. I don't even know what the problem is but somehow something goes wrong since it does not translate "localhost" to "%" on my mac. William and Marcus will fix :)

PR #15866, opened 2024-05-15
G2 2024 v6 issue#15694

a22marca:
**Note: ** This test fails if any new listentries has been added to Testing-Course. This is because the json used has to compare all listentries (it's not possible to just pick the ones we are modifying in the microservice)


f22linhe:
All tests passed. OK to merge!

Issue #15817, opened 2024-05-15
Test microservices from sectioned service


a22albjo:
To test:
opt "DELETE": URL= "../DuggaSys/microservices/sectionedService/removeListEntries_ms.php";
opt "DEL": URL= "../DuggaSys/microservices/sectionedService/deleteListEntries_ms.php";
opt "NEW": URL= "../DuggaSys/microservices/sectionedService/createListEntry_ms.php";
opt "REORDER": URL= "../DuggaSys/microservices/sectionedService/reorderListEntries_ms.php";
opt "UPDATE": URL= "../DuggaSys/microservices/sectionedService/updateListEntries_ms.php";
opt "UPDATETABS": URL= "../DuggaSys/microservices/sectionedService/updateListEntriesTabs_ms.php";
opt "UPDATEDEADLINE": URL= "../DuggaSys/microservices/sectionedService/updateQuizDeadline_ms.php";
opt "SETVISIBILITY": URL= "../DuggaSys/microservices/sectionedService/setVisibleListentries_ms.php";
opt "GRP": URL= "../DuggaSys/microservices/sectionedService/getCourseGroupsAndMembers_ms.php";
opt "CHGVERS": URL= "../DuggaSys/microservices/sectionedService/changeActiveCourseVersion_sectioned_ms.php";
opt "get": URL= "../DuggaSys/microservices/sectionedService/getListEntries_ms.php";

c22cargo:
Working:
* createListEntry (Works fine unless there is a "group activity" in the course. If there is one the MS still works, but you need to refresh the page to see new listentries. This may not be the fault of this MS.)
* removeListEntries (Only seems to work a few times, after that it stops working. Changing back to the monolith doesn't fix this though so the MS will stay implemented.)
* deleteListEntries
* reorderListEntries
* updateListEntries
* updateListEntriesTabs (Works but there is no frontend)

Not working:
* getListEntries (works until there is a "group activity" in a course.)
* updateQuizDeadline

Can't test:
* setVisibleListentries
* changeActiveCourseVersion_sectioned
* getCourseGroupsAndMembers

--------------------------------------------------------------------------------------------------------------------------------------------

The microservices that seem to be working and those that I was unable to test will be implemented. Those that weren't working will have to be fixed before implementation.

Issue #15319, opened 2024-05-03
Codeexamples are not created under chosen moment header


d22mikwe:
- #14935 

The current functionality that seems to work for Github moments is that the fetched github repo folders show up in the dropdown list. (You need to fetch/link a github repo to the course)
![image](https://github.com/HGustavs/LenaSYS/assets/129263393/9651486c-fb22-4eeb-8e2f-dedc9badbc51)

However after pressing the submit button for one of the folders, all files and dirs inside that folder is individually created as a codeexample at the bottom (Not at the intended Github moment).
The whole github moments development will be separated in multiple issues.

Your job in this issue is to make sure that when a user selects a folder and submits it, the codeexample list entries are created under the Github Moment header that was used and not at the bottom.
![image](https://github.com/HGustavs/LenaSYS/assets/129263393/01d07bab-50db-4961-b98c-f072bfdcdb63)
These examples at the bottom should have been created under Geometri 2HP Header if that github icon was used. (As an example)

This is a good repository to link to a course for testing.
https://github.com/LenaSYS/Webbprogrammering-Examples

PR #15867, opened 2024-05-15
G2 2024 v7 issue #15817

c22cargo:
See issue comment for information.


a22marca:
@c22cargo 
It seems like createListEntry stops working for me after creating one item (Doesn't matter which one). 
The response from the microservice:
```
Fatal error: Uncaught PDOException: SQLSTATE[HY000]: General error: 1366 Incorrect integer value: 'UNK' for column 'pos' at row 1 in /var/www/html/LenaSYS/DuggaSys/microservices/sharedMicroservices/createNewListEntry_ms.php:32 Stack trace: #0 /var/www/html/LenaSYS/DuggaSys/microservices/sharedMicroservices/createNewListEntry_ms.php(32): PDOStatement->execute() #1 /var/www/html/LenaSYS/DuggaSys/microservices/sectionedService/createListEntry_ms.php(48): createNewListentry(Object(PDO), '2', '97732', '101', 'New Heading', '9011', '0', 'undefined', '0', '0', 'UNK', 'UNK', '0', 'UNK') #2 {main} thrown in /var/www/html/LenaSYS/DuggaSys/microservices/sharedMicroservices/createNewListEntry_ms.php on line 32
```
Looks like the position is causing the issue here. I found the function that prepares the items before the ajax call: `prepareItem()` in sectioned.js.
There was a old solution that computed a position for the item that should be created, using that worked:
```
  // Places new items at appropriate places by measuring the space between FABStatic2 and the top of the scrren (Old solution)
  //var elementBtnTop = document.getElementById("FABStatic2").getBoundingClientRect();
  //screenPos = Math.round((-1 * elementBtnTop.top) / 350);
  //if (screenPos < 1) {
  //  screenPos = 5;
  //} else {
  //  screenPos = 4 * screenPos;
  //}

  //Place element at bottom if the user has scrolled all the way down, otherwise at the top. (Stopgap solution)
  let screenPos = 0
  if(Math.floor(window.scrollY) === (document.documentElement.scrollHeight - document.documentElement.offsetHeight) 
  && document.documentElement.scrollHeight > document.documentElement.clientHeight) {
   screenPos = document.getElementById("Sectionlistc").childElementCount;
  }
  param.pos = screenPos;
  return param;
```

The commented out code is the one that is working for me. After changing to that code I can create many listentries, including group activities.
 
The new solution was added in [issue#15319](https://github.com/HGustavs/LenaSYS/issues/15319). This seems to solve their issue, so I don't think we should change anything before talking to them (group 3) first. I will speak to them about how we should handle this.

a22marca:
removeListEntries only works in the frontend, and only when marking the item using the checkbox. 
When deleting a listentry without using the mark, no value for `lid` is sent as a parameter.
When deleting using the mark, the `lid` value is included in the request. 
Another issue has to do with the microservice itself.
It performs this query:
```
        $query = $pdo->prepare("UPDATE listentries SET visible = '3' WHERE lid=:lid");
        $query->bindParam(':lid', $sectid);

```
However, the `$sectid` variable is never declared before (using the `getOP()` function). This needs to be added.
So the result is basically that marking and removing a item, removes it visually, but it still exists in the database. This is because its not marked for deletion by the microservice and retrieveSectionedService perform the actual deletion when the function is called.

a22marca:
getListEntries stops working after creating a group activity like you said. This error is thrown:
```
ncaught TypeError: data.grpmembershp is null
    returnedSection http://localhost/LenaSYS/DuggaSys/sectioned.js:1674
    jQuery 6
    AJAXService http://localhost/LenaSYS/Shared/dugga.js:1196
    setup http://localhost/LenaSYS/DuggaSys/sectioned.js:106
    onload http://localhost/LenaSYS/DuggaSys/sectioned.php?courseid=2&coursename=Webbutveckling - datorgrafik&coursevers=97732:1
```
I believe grpmembershp is originally retrieved by the code in sectionedservice.php that is in getCourseGroupsAndMembers_ms.php now.

a22marca (commented):
I just added some further troubleshooting information regarding the microservice where I could find information above. The other microservices you tested are in the state like how you described them for me as well.

Should we merge these changes even though some of the microservice that are integrated are not really working fully (like createListEntry)?

c22cargo:
setVisibleListentries seems to work after the merge of #15868.

c22cargo:
createNewListentry works fine for me, I suggest we merge it and if it breaks we fix it.

I think the reason removeListEntries is so janky is because another group messed up the frontend for it, so I still think it should be merged. It has some other issues with it though. I don't think the "undo" feature will work with the current implementation, but that can be fixed afterwards.

I think we should merge this and then fix the microservices that needs fixing.

a22marca:
> createNewListentry works fine for me, I suggest we merge it and if it breaks we fix it.
> 
> I think the reason removeListEntries is so janky is because another group messed up the frontend for it, so I still think it should be merged. It has some other issues with it though. I don't think the "undo" feature will work with the current implementation, but that can be fixed afterwards.
> 
> I think we should merge this and then fix the microservices that needs fixing.

Yes, that sounds like a plan!

PR #15880, opened 2024-05-15
Created test for highscoreservice_ms.php issue#15354 

a22melst:
The feature branch has been rebased on g2-2024-v7.

I do not know the intended purpose of highscoreservice or retrieveHighscoreService, but the test works now.

### DIFF
--- a/DuggaSys/tests/microservices/higscoreService/highscoreservice_ms_test.php
+++ b/DuggaSys/tests/microservices/higscoreService/highscoreservice_ms_test.php
@@ -3,25 +3,31 @@
 include "../../../../Shared/test.php";
 
 $testsData = array(
-    'update password (user)' => array(
-        'expected-output' => '{}',
-        'query-before-test-1' => "INSERT INTO listentries(lid, entryname,cid,vers,creator) VALUES (9789,'TestEntry',1885,1337,101);",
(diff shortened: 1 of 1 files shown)


a22marca (changes requested):
The url used for the service is wrong. Also, I'm not sure why, but I don't get the "user" entry in the json, even in the unfiltered output.
```
{"debug":"NONE!","highscores":[{"username":"a99erijo","score":"99"},{"username":"k90kakli","score":"99"},{"username":"a87antal","score":"99"},{"username":"c13timan","score":"99"},{"username":"a13jacan","score":"99"},{"username":"b13syeth","score":"99"},{"username":"a99marka","score":"99"},{"username":"teacher2","score":"99"},{"username":"a99karka","score":"99"},{"username":"mofre","score":"99"}],"user":{"username":"a13andka","score":"99"}}
```

a22melst:
The output of highscoreService_ms.php varies from time to time and depends upon things like whether you are logged in or not. I do not think retrieveHighScoreService_ms.php is working as it should, since the select query which fetches the files is not joining tables user and userAnswer , which results in seemingly random usernames being attached to the scores. Since no foreign-key relationship exists between user and userAnswer, such a join cannot be done at the moment. Considering I do not know the purpose of highscoreService_ms.php, and that it does not really do anything other than forwarding data to the retrieve-service, I think this test should be scrapped.

PR #15881, opened 2024-05-15
Remerge after revert

d22mikwe:
[a22edvbu](https://github.com/a22edvbu) commented [yesterday](https://github.com/HGustavs/LenaSYS/pull/15721#issue-2295179803) • 
Added an edit column and option to users in accessed.php. The cogwheel will show a box where users can be edited (this is yet to be implemented, but a placeholder will show).

image

To test this, you can use this link: http://localhost/LenaSYS/DuggaSys/accessed.php?courseid=2&coursevers=97732
Click on the cogwheel and see if the popup will show. This should also be tested for newly created users, which can be made from the fab button. @a22edvbu

### DIFF
--- a/DuggaSys/accessed.js
+++ b/DuggaSys/accessed.js
@@ -90,6 +90,10 @@ function leaves() {
 	$('#dropdowns').css('display', 'none');
 }
 
+function showEditUserPopup(id) {
+	$("#editUser").css("display", "flex");
+}
+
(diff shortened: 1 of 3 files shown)


d22mikwe:
New users cannot be shown yet due to reverts of previous issues

a22edvbu (approved):
Really good code, my compliments to the programmer! Great work!

PR #15883, opened 2024-05-15
(Doing this again cause some error) DeleteButton scales with size and zoomfact

b22davbe:
The deleteButton now scales with the size of the element and the zoomfact but with a max-size of 40 and a min-size of 15 (when zoomed out it becomes to small otherwise). i also placed the deletebutton just outside the edge of the bounding box instead of over it so it doesn't overlapp with the upper-right node that would sometimes result in deleting an element instead of resizing it.

Also made the code for calculating the bounding box shorter and more effective


a21liltr (approved):
lgtm.

PR #15884, opened 2024-05-15
(Doing this again cause some error) Updated options for IE-lines

b22davbe:
Changed the options available for IE-lines because the problem that the issue described shouldn't be possible in IE because its just the symbols that should be on the lines using this notation.


a22solmi (changes requested):
The issue is to move labels so that they don not intersect but you have removed labels all together. I think this is correct as IE uses cardinality on arrows but it is not what the issue intended. I need some input from group leader before I can aprove

a21liltr (approved):
> The issue is to move labels so that they don not intersect but you have removed labels all together. I think this is correct as IE uses cardinality on arrows but it is not what the issue intended. I need some input from group leader before I can aprove

IE should not have cardinality so removing is fine. The labels have been moved on the weekly branch and does not overlap, so as a solution this is accepted.

PR #15888, opened 2024-05-16
Balanced the size for the different elements for issue #15084, new branch/PR

a23ponek:
Changed the default sizes for the UMLRelation, ERAttr and IERelation in the defaults.js file for the new weekly branch.

Link to issue:
https://github.com/HGustavs/LenaSYS/issues/15084


a22vioja:
Change requests from: #15501 and #15443 . Closes issue #15084

a23ponek:
Changed the width/height for the elements in elements.js instead, reason was that a logical change had been made to elements and it is now its own class. See my latest commit: 1e7f5ee

a22solmi (approved):
The element changed are quite large but they are consistent with eachother

PR #15893, opened 2024-05-16
If a user is logged out in profile.php, they should be taken to the home page and no errors should appear

a22oscja:
Copied over previously merged changes and made sure they still worked.

See #15546 for more detailed information.

**Testing:**
- Go to profile, you may be stopped by requiring a secure http connection - to solve this, comment out lines 212-214 in profile.js
- log out
- see if it takes you to homepage or not
- the pushnotification changes are harder to test, but the box at the bottom of profile saying "Notifications subsystems not installed, notifications unavailable." means that it has fully ran without crashing, although it can be more in-depth tested through console logging in InitialiseState in pushnotifications.js, again see the other pull request for more info.

### DIFF
--- a/DuggaSys/profile.php
+++ b/DuggaSys/profile.php
@@ -24,7 +24,13 @@ pdoConnect();
 	<script src="../Shared/dugga.js"></script>
 	<script src="profile.js"></script>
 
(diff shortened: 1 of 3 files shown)


a22moamy (approved):
Works great!

PR #15894, opened 2024-05-16
[re-merge] Added a confirm toast to the toast roster G3 2024 v7 #14813

a22edvbu:
@a22edvbu commented [May 10, 2024] (https://github.com/HGustavs/LenaSYS/pull/15574#issue-2289339746) •

Now there is a confirm toast available to be used among the rest of the toasts. I did this through adding additional elements to toastContainer that shows the two buttons (thanks Moa). To pass the arguments for the buttons, the last two arguments in the toast functioned are used for this (arguments and arguments2).

329528532-6be947c8-93f3-4006-b6f7-8e971af787d2

To call this function, you just add the following code to a php-file that implements toast.php (for example courseed.php):
```javascript
toast("confirm", "Are you sure you want to do that?", 0, "", "");
```

### DIFF
--- a/Shared/css/blackTheme.css
+++ b/Shared/css/blackTheme.css
@@ -1372,4 +1372,8 @@ border-left: 14px solid var(--color-sectioned-table-hi);
 .error {
   background-color: rgb(243, 129, 129);
 }
+
+.confirm { 
+    background-color: #ecf7ff;
+}
 /*/ END /*/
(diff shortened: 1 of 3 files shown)


a22moamy (changes requested):
Toast does not close when pressing yes or no

a22moamy (approved):
Looks good now!

PR #15899, opened 2024-05-16
Fix the automatic scroll function for new items

a22danwe:
managed to make the scroll function work but i couldnt make the scroll go any slower. also put up the time on the highlight so you cant miss which element is added

Test:
add new items in a course and make sure that it scrolls to the newly added element.


b22marah (approved):
LGTM, functionality works nicely, well done! :)

Issue #15626, opened 2024-05-13
Clear console when creating new list entry


d22mikwe:
When creating a new item in a course (Codeexample, moment etc) lots of text is written to the console.
![image](https://github.com/HGustavs/LenaSYS/assets/129263393/70ac2b85-7bae-4811-b1fc-1b45832548c0)
![image](https://github.com/HGustavs/LenaSYS/assets/129263393/100dab9f-d5cf-4cfc-97cf-80413138af52)

Remove the logs that are not necessary / clutters the console.

PR #15901, opened 2024-05-16
G3 2024 v7 #15626

b22chran:
Removed the console.log that cluttered the console. See the issue https://github.com/HGustavs/LenaSYS/issues/15626 to see how it looked before. Also check with the responsible of issue (d22mikwe) or group leader (a22moamy) if it is correct.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -211,8 +211,8 @@ function toggleHamburger() {
 //----------------------------------------------------------------------------------
 
 function selectItem(lid, entryname, kind, evisible, elink, moment, gradesys, highscoremode, comments, grptype, deadline, relativeDeadline, tabs, feedbackenabled, feedbackquestion) {
-  console.log("myConsole lid: " + lid);
-  console.log("myConsole typeof: " + typeof lid);
+  // console.log("myConsole lid: " + lid);
+  // console.log("myConsole typeof: " + typeof lid);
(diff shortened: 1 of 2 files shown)


a22ahmom (changes requested):
When adding new items in a course these messages popup in the console. 
Screenshot 2024-05-17 110126

However when creating a new item the item does not seem to be created at all.
Screenshot 2024-05-17 110435

By talking to group leader (a22moamy) the only relevant messages that should popup in the console are the below.
Skarmavbild_2024-05-16_kl

b22chran:
I tried to put the code back that i removed and it still not creating new items. I don't know if that is another issue that have appeared from something else? I get this error when trying to create a code-item in a course:
Skärmavbild 2024-05-17 kl  12 08 46
I think this is supposed to be the call that gets posted in the console if the creating of items is successful.

b22chran:
Have been trying some more of the creating of items, and it seems like it works fine in "Demo Course" but not in some of the other courses. When creating an item in Demo Course this will be the outcome: 
Skärmavbild 2024-05-17 kl  14 29 01

The console-text in the green square is an Ajax-call that is called by an intervall and not by creating an item.

a22ahmom (changes requested):
The clutter still exists in the console, remove it.

Screenshot 2024-05-17 162549

a22ahmom (approved):
Looks good!

Screenshot 2024-05-20 095901

PR #15905, opened 2024-05-16
G2 2024 v7 issue#15856

b21isabe:
Removed the $variants output from duggaedservice.php and retrieveDuggaedService_ms.php and updated all tests using it to instead use the $entries array. 

Deleting the $variants variable should not break any systems since it was not used for anything other than testing, I added it a few weeks ago to make tests work before double nested arrays were possible in the testing API.


a22iliru (approved):
good

PR #15909, opened 2024-05-16
G3 2024 v7 #15446

### LINKED ISSUE #15446: Implement UI for new installer
In #14698 components were created for use in the new installer. And in #15317 most of the backend of the installer was implemented. In this issue, create a UI for the installer. All the settings do not need to actually toggle the setting, this will be a future issue. 

Requirements:
- Implement all the pages of the installer. 
- Make the installer run with default settings when you click start.

b22marah:
Finnished working on the UI for the installer. From a UI perspective the installer is completed, small fine tunes might come in future, but for now works & looks as intended.

### Notes

The start_installer function does not currectly work due to it not been properly sat. This is to be fixed in the future since the function works, but it's not being initialized correctly at the moment, this is because the function tries to run before being called in previous attempts to test it. Instead this should be fixed in another branch / issue which holds correct data to test it with. So regarding this, ignore it :).

Also the breadCrum errors in the console can be ignored, this will also be fixed in future issues, where we fix all small tweeks, these errors however doesn't affect code since it work's, but as i said, will be fixed to get rid of annoying console errors.

Closes #15446


a22moamy (approved):
Looks and works good to me!

I don't know if you want some tips, but maybe this could be good for the lesser technical user :)

Skärmavbild 2024-05-16 kl  13 41 01

Skärmavbild 2024-05-16 kl  13 42 23

a22moamy (approved):
Looks great!

d22mikwe (approved):
LGTM!

PR #15914, opened 2024-05-16
Create new test for the microservice removeListEntries_ms.php #15693

c22mohel:


a22marca (changes requested):
The assertion test is not passing for me. I tried reinstalling and clearing the database as well.
There are several diffs in the output from the microservice and the expected-output, that seems to start in the fileid's in the "links" entry.

a22marca (changes requested):
We can filter out codexamples and links. But I don't think it's a good idea to filter out "entries" since that array actually contains all listentries. The test doesnt actually compare the listentries anymore.
Add "entries" back as a filter and to the expected-output and it should be good after that.

a22marca (changes requested):
image
Test is not passing, seems the expected-output does not have double quotes around the values, but the value output does

b23johgu (commented):
Verkar vara ett test som inte fungerar. Har inte fortsatts arbetas på och det är svårt att avgöra vad koden egentligen ska göra eller hur man ska fortsätta arbeta på den.

Issue #15760, opened 2024-05-15
Page `duggaed.php` does not show logged in state


a22willi:
### Background
When navigating to the tests page the user is not shown as logged in. See image below.
![image](https://github.com/HGustavs/LenaSYS/assets/129264968/77262082-a9db-4676-923e-33840cc00ced)
Fix this problem.

### Recreate
1. Login to LenaSYS.
2. Navigate to LenaSYS -> `any course` -> show tests 
3. The user is no longer shown as logged in.

PR #15915, opened 2024-05-16
Added missing session_start() #15760

a20jonda:
It turned out to be more simple than I thought. session.start() was missing in duggaed.php.
Login now stays when going into duggaed and through a refresh.
image


a22danwe (approved):
LGTM!

PR #15919, opened 2024-05-16
Renamed getUserDuggaFeedback_ms.php -> readUserDuggaFeedback_ms.php

f22linhe:


a22marca (changes requested):
A comment at the top of readUserDuggaFeedback still has getUserDuggaFeedback. Doesn't really matter, but you could changse the name there as well.

See my previous note about a duplicate switch. If you'd rather create a new issue for that, thats ok.

a22marca (approved):
The test is passing Looks good now. Everything is renamed correctly. Merging

PR #15921, opened 2024-05-16
g3-2024-v7-#15341 

a22samka:
Added the code (that was used to solve this issue last week) again as it seems it dissapeared


a20jonda (approved):
Since this is a revert revert, I'll leave the choice on the conventions up to the merger. Convention enforcement could probably be fixed in a larger issue instead.

PR #15922, opened 2024-05-16
Added old changes to new branch (added remove user button)

a22natth:
Since the solution to this issue was lost after the weekly merge, the old code was added and slightly modified.

## OLD PR TEXT
In the example below, an additional icon for removing a user has been added. 
Pasted Graphic 12

When clicking the icon, the popup below will be presented.
Pasted Graphic 13

The field to the right is a dropdown which also acts as a search bar for different users. 
Pasted Graphic 14


a22moamy (approved):
Looks good!

Issue #15877, opened 2024-05-15
It's possible to select existing lines while dragging a new one


a22vioja:
## Description

It's possible to select existing lines while dragging a new one:

https://github.com/HGustavs/LenaSYS/assets/129263559/a8d763d8-efcb-4b78-ba52-f5bd8b999864

## Job
Fix this so it would not be able to select the line while dragging another line (or element).

PR #15925, opened 2024-05-16
fix for issue #15877 "It's possible to select existing lines while dragging a new one"

b22jonan:
The fix is really just not updating any lines you click if youre currently creating a line.


a22solmi (approved):
This works as intended, lines can no longer be selected in during edge creation

PR #15928, opened 2024-05-16
Renamed getCourseVersions_ms.php -> readCourseVersions_ms.php

f22linhe:
Renamed the microservice and everywhere it is called in the system.


a22marca (changes requested):
The old name is still used as the function name in readCourseVersions_ms.php.
When changing the function name to the new name, don't forget to change the name where the function is called at line 199 in retrieveSectionedService_ms.php

a22marca (changes requested):
I think I missed one file that is not actually renamed here in the last review.
There is a test for getCourseVersions here:
```
/DuggaSys/tests/microservices/courseedService/getCourseVersions_ms_test.php
```
It's in the wrong microservice folder for the test, so you should move it to sectionedService as well.

a22marca (approved):
All names are changed and test is passing. Merging

PR #15932, opened 2024-05-16
correct parameters are sent in the ajax request for the kind SECTION

a22marca:
Some parameters like `pos` where not being sent in the ajax request being made in the kind "SECTION". It should now work.


c22cargo (approved):
Looks good, everything seems to work.

Issue #15908, opened 2024-05-16
Fix removeListEntries


c22cargo:
Currently removeListEntries does not completely work. Listentries get set to visibility 3, but they don't seem to get deleted (only sometimes).

The idea is that removeListEntries sets the visibility to 3, and then everything with that visibility gets removed by the retrieve MS.
This idea seems flawed though, because of the "undo" button. I assume the "undo" button is supposed to set the visibility back, but if the listentry is instantly deleted by the following retrieve call, that cannot be done.

Another solution for this problem will have to be thought up and implemented.

c22cargo:
The toast for undoing seems to not be working, but I found some code which suggests that the way it's supposed to work is that the delete function should sleep for 60 seconds, to allow "undo". This doesn't seem to be working correctly at the moment but I will probably change it to use the deleteListEntries MS instead. This won't fix the "undo" issue, but it should make deletion work again.

```
function deleteItem(item_lid = []) {
  for (var i = 0; i < item_lid.length; i++) {
    lid = item_lid ? item_lid : $("#lid").val();
    item = document.getElementById("lid" + lid[i]);
    item.style.display = "none";
    item.classList.add("deleted");
  
    document.querySelector("#undoButton").style.display = "block";
  }

  toast("undo", "Undo deletion?", 15, "cancelDelete();");
  // Makes deletefunction sleep for 60 sec so it is possible to undo an accidental deletion
  delArr.push(lid);
  clearTimeout(delTimer);
  delTimer = setTimeout(() => {
    deleteAll();
  }, 60);
}

// Permanently delete elements.
function deleteAll() {
  for (var i = delArr.length - 1; i >= 0; --i) {
    AJAXService("DELETE", {
      lid: delArr.pop()
    }, "SECTION");
  }
  $("#editSection").css("display", "none");
  document.querySelector("#undoButton").style.display = "none";
}
```

c22cargo:
I found some other code that suggests that listentries marked for deletion should be deleted when the page is unloaded, but something isn't working here.

```
// Delete items marked as deleted when page is unloaded
window.addEventListener('beforeunload', function (event) {
  var deletedElements = document.querySelectorAll(".deleted")
  for (i = 0; i < deletedElements.length; i++) {
    var lid = deletedElements[i].id.match(/\d+/)[0];
    AJAXService("DEL", {
      lid: lid
    }, "SECTION");
  }
});
```

PR #15934, opened 2024-05-16
G2 2024 v7 issue #15908

c22cargo:
I don't think the "undo" button not working is our group's problem. There seems to be some code in sectioned.js which is supposed to handle undoing by using classes, arrays, and timers.

I changed sectioned.js to use deleteListEntries instead of removeListEntries, which fixes the deleting issue. Unless I am misunderstanding something I think deleteListEntries is actually the correct microservices to use. But then I am not sure what removeListEntries should be used for.

It's almost like 2 different groups had different ideas on how to implement the "undo" function and implemented both at the same time.

See issue comments for more information.


a22albjo (approved):
The function is supposed to delete items, this is done with the DEL opt. This can be merged.

PR #15935, opened 2024-05-16
documentation for element.js

b22jonan:
just a bunch of documentation


a22alfna (approved):
This documentation looks good to me right descriptions and "returns" on all of the new descriptions. And the default "param"s look good aswell

PR #15937, opened 2024-05-16
Renamed changeactiveCourseVersion_sectioned_ms.php -> updateActiveCourseVersion_sectioned_ms.php

f22linhe:


Shared/dugga.js
```diff
@@ -1164,13 +1164,13 @@ function AJAXService(opt,apara,kind)
 				serviceURL= "sectionedservice.php"; //change when MS is created
 				break;
 			case "CREGITEX":
-				serviceURL= "../DuggaSys/microservices/sectionedService/changeActiveCourseVersion_sectioned_ms.php";
+				serviceURL= "../DuggaSys/microservices/sectionedService/updateActiveCourseVersion_sectioned_ms.php";
```
a22marca, line 1167: Seems like the original microservice url was wrong here. It should be using createGithubCodeExample_ms.php

a22marca (changes requested):
The test is passing so I'd say these changes are working. The microservice cant be tested on the website itself since there is no frontend for it. 

I know it's not in the scope of this issue, but maybe you can change the case for  "CREGITEX" to the correct url in dugga.js (see previous comment).

a22marca (approved):
Looks good. Merging

PR #15943, opened 2024-05-17
Fixed allignment issue on edit dugga and edit course

a22karja:
Added classes to the inputs of. And removed offset on select in the css


a22moamy (approved):
Looks great! Well done :D

PR #15946, opened 2024-05-17
G2 Merge issue#15941

a22marca:
The test is using the service getListEntries_ms.php to be tested. This is because readCourseVersions can't be called by the testing service since it's just a function.  getListEntries_ms uses retrieveSectionedService()`, which in turn calls readCourseVersions() to retrieve all course versions.


a22iliru (changes requested):
image
Merge anyways?

a22marca:
> image Merge anyways? (Test 3 fails)

I'll take a look and see what the problem is

Issue #15907, opened 2024-05-16
Make relations elements smaller


a21liltr:
![image](https://github.com/HGustavs/LenaSYS/assets/129370135/9eb5ebbf-8183-41f5-9e5b-646778a5dc1f)

## Description
Relations elements were made bigger to have a constistent sizing, however they now overlap on example diagrams and it looks a bit funny with them being so big.

## Job
Change them to be smaller but still consistent in sizing.

## Tip
You have to change both defaults.js and /class/element.js

PR #15949, opened 2024-05-17
Made elements smaller for issue #15907

a23ponek:
Changed the default size of the ERRelation, ER-Attribute, UML-relation and IE-relation with a factor of ~70%. This should make the elements smaller but still enough size for text/form etc. The same changes was made in defaults.js and element.js for the respective element.

Link to issue:
https://github.com/HGustavs/LenaSYS/issues/15907


a21liltr (changes requested):
Make your changes in defaults.js and class/elements.js instead.

a21liltr (approved):
Relation elements are now smaller. Nothing else was affected and they look good both as lonely elements and in the example templates.

PR #15951, opened 2024-05-17
fixed ESC button when closing modals in sectioned.php

a22ahmom:
just before when you would press ESC you would be transported form sectioned.php to courseed.php when closing these popups with ESC, get canvas link, settings (JS-TEST template 3) and settings (JavaScript-Code) In the issue description it wasn't mentioned which specific popups this happened to but these three popups were however found and fixed.

Screenshot 2024-05-17 103934

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -617,7 +617,9 @@ function closeOpenPopupForm(){
     "#loginBox",
     "#loadDuggaBox",
     "#sectionHideConfirmBox",
-    "#sectionShowConfirmBox"
+    "#sectionShowConfirmBox",
+    "#canvasLinkBox",
+    "#editSection"
   ];
   for (let popup of allPopups){
     if ($(popup).css("display") !== "none"){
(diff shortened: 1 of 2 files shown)


b22chran (changes requested):
The escape-function works fine when in the popup-windows that are described. However, when reading the issue it seems like it also should be closing the modal that appears when deleting items in sectioned.php instead of returning to coursed.php.
When this modal appears and pressing ESC it still returns to the courseed.php. I believe this was supposed to be fixed.
Skärmavbild 2024-05-17 kl  11 57 34

DuggaSys/sectioned.js
```diff
@@ -617,7 +617,9 @@ function closeOpenPopupForm(){
     "#loginBox",
     "#loadDuggaBox",
     "#sectionHideConfirmBox",
-    "#sectionShowConfirmBox"
+    "#sectionShowConfirmBox",
+    "#canvasLinkBox",
+    "#editSection"
   ];
```
d22mikwe, line 623: Add "#toastContainer" here

Shared/toast.php
```diff
@@ -124,6 +124,14 @@ function toast(type, text, duration, arguments = null) {
                 }
             }
         });
+
+        // Close the toast by pressing the ESC button
+        document.addEventListener('keydown', function (event) {
```
d22mikwe, line 129: change to keyup

d22mikwe (changes requested):
See comment. This should fix the issue

d22mikwe (approved):
Works now! LGTM!

PR #15953, opened 2024-05-17
Reverted remerge "Add functionality to "Add User" && "Remove User" button #15562"

d22mikwe:
Whats done:

Dropdown list in both Remove user and add user is filled with all users.
Add User now adds a user to the course without having to create one again.
Remove User now removes selected user from that course.
Changed so readaccess users is seen is the list and not just teachers.
Dynamic tooltipTerm for input validation.

Todo in future issues:

Add Classes?
Add Groups?
Maybe update Term value if user already exists in course while trying to Add User.
Testing:
All users should show in both dropdown lists (Add User and Remove User)

Try adding any user into any course. Then reload the page and it should show up as a list entry.

Try removing a user from a course. It should disappear from the list entries

### DIFF
--- a/DuggaSys/accessed.js
+++ b/DuggaSys/accessed.js
@@ -98,12 +98,14 @@ function showCreateUserPopup() {
 	$("#createUser").css("display", "flex");
 }
 
-function showAddUserPopup() {
+function showAddUserPopup(id) {
 	$("#addUser").css("display", "flex");
+	loadUsersToDropdown(id);
 }
 
-function showRemoveUserPopup() {
+function showRemoveUserPopup(id) {
 	$("#removeUser").css("display", "flex");
+	loadUsersToDropdown(id);
 }
 
 function showCreateClassPopup() {
@@ -128,6 +130,14 @@ function hideCreateClassPopup() {
 
 function hideEditUserPopup(id) {
 	$("#editUser").css("display", "none");
+}
+
+function hideAddUserPopup() {
+	$("#addUser").css("display", "none");
+}
+
+function hideRemoveUserPopup() {
+	$("#removeUser").css("display", "none");
 }
 
 //----------------------------------------------------------------------------
--- a/DuggaSys/accessed.js
+++ b/DuggaSys/accessed.js
@@ -144,8 +144,8 @@ function hideRemoveUserPopup() {
 //-------------==========########## Commands ##########==========-------------
 //----------------------------------------------------------------------------
 function addUserToCourse() {
-	let input = document.getElementById('addUsernameAdd').value;
-	let term = $("#addTermAdd").val();
+	let input = document.getElementById('addUsername2').value;
+	let term = $("#addTerm2").val();
 	$.ajax({
 		type: 'POST',
 		url: 'accessedservice.php',
@@ -173,7 +173,7 @@ function addUserToCourse() {
 	hideAddUserPopup();
 }
 function removeUserFromCourse() {
-	let input = document.getElementById('addUsernameRemove').value;
+	let input = document.getElementById('addUsername3').value;
 	$.ajax({
 		type: 'POST',
 		url: 'accessedservice.php',
--- a/DuggaSys/accessed.js
+++ b/DuggaSys/accessed.js
@@ -143,6 +143,61 @@ function hideRemoveUserPopup() {
 //----------------------------------------------------------------------------
 //-------------==========########## Commands ##########==========-------------
 //----------------------------------------------------------------------------
+function addUserToCourse() {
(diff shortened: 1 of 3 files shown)


DuggaSys/accessed.js
```diff
@@ -130,9 +132,72 @@ function hideEditUserPopup(id) {
...
+			}, "ACCESS");
+		},
+		error: function(xhr, status, error) {
+			console.error("Error", error);
+		}
+	});
+	hideAddUserPopup();
+}
+function removeUserFromCourse() {
+	let input = document.getElementById('addUsernameRemove').value;
```
b22marah, line 176: Same here, could / should be changed to removeUsername, not addUsernameRemove, since were not adding anything this id seems a little odd.

DuggaSys/accessedservice.php
```diff
@@ -285,10 +287,71 @@
 				}
         	}
 		} // End of foreach user
-	} // End ADD_USER
+		//End of ADDUSR
+	} else if (strcmp($opt,"RETRIEVE")==0) {
```
b22marah, line 291: Please fix spacing of parameters

DuggaSys/accessed.php
```diff
@@ -135,22 +135,22 @@
 					<input type='hidden' id='uid' value='Toddler' />
 					<div class='flexwrapper'>
 						<div class="tooltipDugga"><span id="tooltipFirst" style="display: none;" class="tooltipDuggatext">  </span></div>
-						<input placeholder="Greger" class='textinput' type='text' id='addFirstname' onchange="tooltipFirst()" onkeyup="tooltipFirst()"/>
+						<input placeholder="Greger" class='textinput' type='text' id='placeholderID4' onchange="tooltipFirst()" onkeyup="tooltipFirst()"/>
```
b22marah, line 138: Bad ID, could be changed to something like 'addFirstname1' or something along these lines, to keep relevant id but not get duplicate ids.

DuggaSys/accessed.php
```diff
@@ -135,22 +135,22 @@
...
 					</div>
 					<div class='flexwrapper'>
 						<span>Last Name:</span>
 						<div class="tooltipDugga"><span id="tooltipLast" style="display: none;" class="tooltipDuggatext">  </span></div>
-						<input placeholder="Gregersson" class='textinput' type='text' id='addLastname' onchange="tooltipLast()" onkeyup="tooltipLast()"/>
+						<input placeholder="Gregersson" class='textinput' type='text' id='placeholderID3' onchange="tooltipLast()" onkeyup="tooltipLast()"/>
```
b22marah, line 143: Same here regarding ID names.

DuggaSys/accessed.php
```diff
@@ -135,22 +135,22 @@
...
 					</div>
 					<div class='flexwrapper'>
 						<span>Term:</span>
 						<div class="tooltipDugga"><span id="tooltipTerm" style="display: none;" class="tooltipDuggatext">  </span></div>
-						<input placeholder="HT-11" class='textinput' id='addTerm' onchange="tooltipTerm(this)" onkeyup="tooltipTerm(this)">
+						<input placeholder="HT-11" class='textinput' id='placeholderID2' onchange="tooltipTerm(this)" onkeyup="tooltipTerm(this)">
```
b22marah, line 148: Same here regarding ID names.

DuggaSys/accessed.php
```diff
@@ -135,22 +135,22 @@
...
 					</div>
 					<div class='flexwrapper'>
 						<span>Email:</span>
 						<div class="tooltipDugga"><span id="tooltipEmail" style="display: none;" class="tooltipDuggatext">  </span></div>
-						<input placeholder="b17mahgo@student.his.se" class='textinput' id='addEmail' onchange="tooltipEmail()" onkeyup="tooltipEmail()"/>
+						<input placeholder="b17mahgo@student.his.se" class='textinput' id='placeholderID1' onchange="tooltipEmail()" onkeyup="tooltipEmail()"/>
```
b22marah, line 153: Same here regarding ID names.

b22marah (changes requested):
Some quick-fixes to id:s needs to be changed, also left comment regarding other IDS thats being used don't know if this is relevant to this issue or not though. Besides this the code LGTM, but i can't see the list of users, so can't test functionality.... Don't really know the problem as to why i can't see the list.

DuggaSys/accessed.php
d22mikwe: The whole form is a placeholder

b22marah (approved):
LGTM!

Issue #15131, opened 2024-04-30
Ghost element for Sequence Lifeline is hard to see on "dark mode"


a22vioja:
## Description 

When the dark mode is on, it is hard to see the ghost element for Sequence Lifeline because the color of the figure is black. 

<img width="358" alt="Screenshot 2024-04-29 at 22 15 50" src="https://github.com/HGustavs/LenaSYS/assets/129263559/683db5d6-74db-4699-8259-7e9b96f6f583">

## Job

Change a color for the ghost element for Sequence Lifeline figure.

PR #15955, opened 2024-05-17
change color of ghost #15131

c22abdha:
in draw element.js i have changed ${element.stroke} to gray color

image


a23ponek (approved):
The changes to the Sequence lifeline as shown in the earlier images is visible in dark mode and works fine without dark mode toggeled on. No problems seem to arise overall in the diagram.

Issue #15426, opened 2024-05-06
Currently, sorting users by user or password will cause faulty css in accessed.php


a22moamy:
To see this issue:
1. Go into a course, like "Datorgrafik"
2. Press the "Give students access to the selected version"-button:
![Skärmavbild 2024-05-06 kl  16 00 52](https://github.com/HGustavs/LenaSYS/assets/129263915/fed6c814-cc21-4557-ad0c-f6e9cff001d2)
3. Sort rows by user or password by pressing "User" or "Password" in the table.

The css for odd/even rows will be broken. 

The current appearance (sorting by user or password). **Notice the background color of every row in the table**:
![Skärmavbild 2024-05-06 kl  16 02 51](https://github.com/HGustavs/LenaSYS/assets/129263915/d14262d5-20fa-4156-9624-799768b6a297)

Fix it so that every odd row is darker and every even row is lighter in dark mode and the opposite in white mode. Just like how the colors are when sorting by "First name" or "Last name". Like this:
![Skärmavbild 2024-05-06 kl  16 05 12](https://github.com/HGustavs/LenaSYS/assets/129263915/128e98e2-ac9e-42db-aaf7-cbeb55216b2c)

PR #15956, opened 2024-05-17
 Applied css and function for changing color

b22chran:
Created a function that looks at the rows that are shown every time it re-renders and apply the even and odd classes to them. Removed the class "changeColorInDarkModeTable" that is no longer needed because of the use of correct variable in the style.css and applies it in darkmode as well.

Se issue https://github.com/HGustavs/LenaSYS/issues/15426 to find out where to test it.


a22moamy (approved):
Looks good! Good job :D

PR #15964, opened 2024-05-17
reorderListEntries_ms.php -> updateOrder_ms.php

f22linhe:

Note, I´m not updating the names for the microservice i the database_related_micro_services since it will be obsolet when the documentationen is done. I´m also not updating the sectionedService_test.php the function there is not the microservice per se.


f22linhe:
updateListEntryOrder_ms.php is a better name than updateOrder_ms.php. Scrapping this pr and will make a new one!

Issue #15664, opened 2024-05-13
Implement Flexbox solution to tooltips


a22moamy:
## Background
We have had many issues with tooltips being placed weirdly. This is because they're all placed using pixel measurements, which results in them breaking whenever the form is updated somewhat. 

## The issue
To finally end this evil spiral and make a more long-lived solution, we have decided to switch to using Flexbox for all tooltips in DuggaSYS. With your fix, tooltips should be automatically aligned next to an input without having to cook up the perfect amount of pixels to add. 

Your solution should look good on wider screens (PC and laptops). Adjustments for mobile devices will be handled in another future issue.

a22chrfa:
I ended up not using flexbox because of the extra layer of a wrapper div.

In it's original form the html structure looked like this:

`<div class="inputwrapper">
    <span> course code: </span>
    <div class="formdialog">
        <span> text </span>
    </div>
    <p> text </p>
    <input> inputfield </input>
</div>`

Instead, this is the solution:

`<div class="inputwrapper">
    <p> text </p>
    <div class="dialogwrapper">
        <div class="formdialog">
            <span> text </span>
        </div>
        <input> inputfield </input>
    </div>
</div>`

**Why another wrapper?**
Mostly semantics. 
I could've moved the input into the formdialog div and achieved the same thing with some editing. But it seems semantically incorrect to move an input field into a div called "formdialog". Secondly, the formdialog class has functions tied to it so I didn't think renaming and/or editing it too much would be a good idea.

With some reference to the previous comment. An absolutely positioned element will be anchored to it's closest ancestor that isn't static. By making the dialogwrapper have "position: relative" with the formdialog absolute, it will now be anchored to the dialogwrapper. Also, since formdialog is taken out of normal flow due to it's absolute position, the only element remaining inside the dialogwrapper is the input, making the dialogwrapper-div always be the same exact size and position as the input. When we position the formdialog now, it will effectively always be anchored to the input element.

With some adjusments to the css, the result looks like this:
![image](https://github.com/HGustavs/LenaSYS/assets/128893264/d7a227d5-f9f9-4694-a806-36e84c73b01a)

What needs to be done now is removing all the specific css/inline rules that are applied to all the other validation texts. Theoretically, once this is done. Whenever a valdiation text is added with the same html structure, the validation text should fit automatically. Hopefully nothing unexpected shows up.

a20jonda:
The wiki should probably be updated after this is done.
[...]

PR #15966, opened 2024-05-17
Generalized validation text positioning on (almost) all forms (#15664)

a22chrfa:
For inputs we have created a wrapper (class "dialogwrapper") that can be easily added to inputs and will automatically position the validation texts to the left of the input field. The classes .formDialog, .formDialogText have been altered to utilize the new dynamic positioning and loads of specific styling for validation texts (4px left here and 5 px right there) have been removed.

**For testing:**
This is an issue that ONLY touches in the actual validation texts position relative to their input field. The actual input field (which is sometimes a bit misaligned) is not within the scope of this issue.

1. The Edit dugga form will not be touched in this issue. It won't look good here. It's a future issue, as planned.
2. Responsiveness (smaller screens) are not handled here.
3. The dialogs that are positioned below their input field (formDialogWide) are considered special cases. They will not adhere to the new standard and will remain unique. This is specified in the css rules.

Just check every form you can find in courseed.php and sectioned.php, don't forget darkmode.


a22edvbu (approved):
Everything looks good to me! Great work!

Issue #15961, opened 2024-05-17
add opt for specialUpdate_ms ("SPECIALUPDATE") in dugga.js under kind "COURSE"


a22marca:
There is no opt for calling a specialUpdate_ms.php

PR #15981, opened 2024-05-17
G2 2024 v7 issue #15961

c22cargo:
I wasn't able to find where the OPT was used, so the microservice hasn't been implemented. Instead, only the relevant case has been added.


a22melst (approved):
Looks good

PR #15983, opened 2024-05-17
Added keypress functionality to creation of new course

a22chrfa:
Added eventlistener and also added tabindex to make the surrounding div focusable. If a user clicks somewhere inside the form and isn't focused on an actual input field, the eventlistener wouldn't trigger. Adding tabindex makes the entire div focusable, which makes the eventlistener active.

### DIFF
--- a/DuggaSys/courseed.js
+++ b/DuggaSys/courseed.js
@@ -836,6 +836,16 @@ function quickValidateForm(formid, submitButton) {
 	return false;
 }
 
+//adds keypress enter to the form when creating a new course. Calls the same function as clicking the submit-button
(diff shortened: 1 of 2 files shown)


a22oscja (approved):
Works the way it is supposed to, code also looks good.

Technically this allows you to trigger the function for creating a course without filling in all the fields (as normally the button would be greyed out), one could argue this hotkey should work similarly, but there are checks in the function as well so this should not be a problem.

Very minor detail would be that you cannot click enter right after clicking the button (which is as described), but since the value is stored if you write something and exit the window there could already be valid text in there, it probably shouldn't store the text in the fields.

Issue #15617, opened 2024-05-13
"Product number" font is bold in ER Example diagram


a22vioja:
## Description
"Product number" is bold in ER Example diagram. It should look same as other attributes and entities.

<img width="449" alt="Screenshot 2024-05-13 at 10 39 48" src="https://github.com/HGustavs/LenaSYS/assets/129263559/8b247b0e-da01-48a4-88fc-737adfaf75bd">

## Job
Fix the font to regular.

a23ponek:
I have tried different approaches to make the bold text normal, but it seems like its heavily connected to the weakKey variant that can be chosen in the options panel in diagram.php, when a certain element like ER attribute is selected. All other variants like multiple, candidate or primary doesn't cause the text to be bold font. It seems like the code on rows 416-424 causes the bold text but its not obvious, the function drawElementERAttr() in elements.js. further investigation of the matter is required to solve this issue. Also I couldn't find any problematic CSS code that would interfere with this issue in style.css and diagram.css.

b22jonan:
from what i can see this is added to the text: 
![image](https://github.com/HGustavs/LenaSYS/assets/61019188/df80a807-bec7-4b8b-a917-977a410aa3c0)
which is what makes it look bolded, i just gotta find where its added and then its fixed :)

PR #15985, opened 2024-05-17
fix for #15617

b22jonan:
the line removed is the cause for adding style="stroke: rgb(0, 0, 0)" which caused the text to appear bolded. from what i saw while messing around with the diagram i dont think that this messes anything up but im not 100% sure so it will need to be tested a bit better than i did


a21liltr (approved):
Dashed underline slightly hard to see in darkmode, but issue is fixed.

PR #15986, opened 2024-05-17
G1 2024 v7 #15777

a22cazlu:
Updated the documentation in stateChange.js


a23ponek (approved):
The documentation updates seemes feasible and more explaining what the various functions and attributes in stateChange.js actually do. Looks good!

PR #15988, opened 2024-05-17
G1 2024 v7 #15782

a22ricni:
In DuggaSys/diagram/draw/element.js are some documentation for the function.


a22cazlu (approved):
looks good

b23danhe (changes requested):
The documentation is a good feature to keep, but there are still some minor grammatical errors that need to be corrected. Read all the documentation and make sure it's grammaticaly correct.

PR #15990, opened 2024-05-17
Renamed getCourseGroupsAndMembers_ms.php -> readCourseGroupsAndMembers_ms.php

f22linhe:

Note, I´m not updating the names for the microservice in the database_related_micro_services since it will be obsolet when the documentationen is done. I´m also not updating the sectionedService_test.php, the function there is not the microservice per se.


a22marca (approved):
Everything seems renamed. Test is passing and the microservice is working on the website.
Merging

PR #15994, opened 2024-05-17
Implemented input validating for course feedback

a18chrkr:
I have implemented the same input validation used for the name of the course. This correctly turn the input field red and display a tooltip when the field has been filled incorrectly. The save button also gets disabled. I've also changed the declarations of some variables to 'let' instead of 'var'.

Worth mentioning is that the tooltip is displayed at the wrong location. However, this problem existed before this implementation and also affect the "Name" input field.

courseFeedback

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -3606,7 +3606,7 @@ function validateDate2(ddate, dialogid) {
   return false;
 }
 
-function validateSectName(name) {
+function  validateSectName(name) {
   var element = document.getElementById(name);
   var errorMsg = document.getElementById("dialog10");
   if (element.value.match(/^[A-Za-zÅÄÖåäö\s\d():_-]+$/)) {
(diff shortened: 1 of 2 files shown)


DuggaSys/sectioned.js
```diff
@@ -3736,12 +3736,14 @@ function quickValidateForm(formid, submitButton) {
...
-    var deadlinepart = document.getElementById('inputwrapper-deadline');
-    var deadlinedisplayattribute = deadlinepart.style.display;
+    let sName = document.getElementById("sectionname").value;
+    let item = document.getElementById("editSectionDialogTitle").innerHTML;
+    let feedback = document.getElementById("courseFeedback").value;
+    let deadlinepart = document.getElementById('inputwrapper-deadline');
+    let deadlinedisplayattribute = deadlinepart.style.display;
     valid = true;
     valid &= validateSectName('sectionname');
+    valid &= validateSectName('courseFeedback');
```
a20jonda, line 3746: The 'validate section name' (`validateSectName`) function cannot be reused as it is connected to a specific error dialog. You have to create-new/copy the function and a create a new dialog element to reference

DuggaSys/sectioned.php
```diff
@@ -323,7 +323,7 @@ function updateGithubDir($pdo, $githubDir, $lid) {
 					<div id='inputwrapper-visibility' class='inputwrapper'><span>Visibility:</span><select style='align:right;' id='visib'></select></div>
 					<div id='inputwrapper-group' class='inputwrapper'><span>Group type:</span><select style='align:right;' id='grptype'></select></div>
 					<div id='inputwrapper-Feedback' class='inputwrapper'><span>Enable Student Feedback:</span><input type="checkbox"  style='align:center;' id='fdbck' title='Student feedback checkbox' onchange='showFeedbackquestion()'></input></div>
-					<div id='inputwrapper-FeedbackQuestion' class='inputwrapper' style='display:none;'><span>Student Feedback Question:</span><input type="input"  class='textinput' id='fdbckque' value='How would you grade the dugga?'></input></div>
+					<div id='inputwrapper-FeedbackQuestion' class='inputwrapper' style='display:none;'><span>Student Feedback Question:</span><input type="text" class='textinput' onkeyup="quickValidateForm('editSection', 'saveBtn')" onchange="validateSectName('courseFeedback')" id='courseFeedback' placeholder='Enter feedback...'/></div>
```
a20jonda, line 326: A bit of redundancy here. You only need to run quickValidateForm. I'm pretty sure I removed the redundant method calls some weeks ago. You could optionally remove the redundant validateSectName calls including the one further above.

a20jonda (changes requested):
A new function is required. Look how the other dialogs is fitted to the forms and try to match when creating a new. Examples can be found in courseed.php and sectioned.php with the class tag formDialog.

Everything else works.

a20jonda:
image
The error dialogs are currently visually incorrect and are being corrected in a pending pull request. You only need to add the element.

a20jonda:
You are allowed to rename and change valideSectName so that it can be reusable like this:
```javascript
function  validateSectionText(name, dialog) {
  var element = document.getElementById(name);
  var errorMsg = document.getElementById(dialog);
```

PR #15996, opened 2024-05-17
G1 2024 v7 #15784

a22alfna:
I believe the documentation for node.js file should be done now. Havent done documentation issue before so feel free to let me know if i missed something.


a22vioja:
Closes issues #15784

a22solmi (changes requested):
This is not valis JSDoc syntax. You cannot write a doc for an if statement. You could write one for a variable but not with @param. The task is to write docs for every function. This cannot be added as is

Issue #15377, opened 2024-05-04
Go over test for the microservice addClass_ms.php again


f22linhe:
See parent issue: _Create new tests for the microservices created from accessedService.php #15177_

The test for this microservice was created before the retrieveAccessedService_ms.php was created. Therefore, go over the test again and make sure it works once the retrieveAccessedService_ms.php has been implemented. If it does not work as aspected, make sure it does.

Note, this issue cannot be resolved until these issues has been resolved:

Create a retrieveAccessedService microservice #15250

Create new test for the retrieveAcessedService_ms.php #15306

c22cargo:
The retrieve MS always returns NULL from addClass_ms. This needs to be fixed before the test can be fixed.

PR #16001, opened 2024-05-17
G2 2024 v7 issue #15377

c22cargo:
The test was updated with a few changes.

The microservice was updated to use the retrieve MS.

The retrieve MS was updated to retrieve more information from the table "class".


a22iliru (approved):
Works

PR #16002, opened 2024-05-17
documentation for event/mouse.js as specified in #15790

b22jonan:


a22solmi (changes requested):
Missing return type on checkDeleteBtn. Please remove the text after "@see pointerStates" in function tup. JSDocs wont recognize it otherwise

b22jonan:
went ahead and removed the text after "@see pointerStates" on mup aswell

a22solmi (approved):
Looks great!

Issue #15878, opened 2024-05-15
Undo button doesn't work as expected while line is selected


a22vioja:
## Description
Pressing undo while a line is selected causes everything to become invisible. Everything can be brought back by clicking somewhere else to remove the selection and then pressing undo again.

https://github.com/HGustavs/LenaSYS/assets/129263559/87818ce1-c65f-460d-ab56-c0d476daf279

TIPS from Jakob:

This can be fixed by adding a null check in drawSelectionBox()

if (!tempLines[0]) return; // <- add this
let hasPoints = tempLines[0].getAttribute('points'); // Polyline

## Job
Fix "undo" button so that when a line is selected and the 'undo' button is pressed, everything does not disappear.

PR #16007, opened 2024-05-17
g1-2024-v7-#15878

a22cazlu:
Put in two if-statements, one to check if bounding box coordinates have changed and one to check if the element is a line.


a22wilas (approved):
### Old version
Before undo:
bild
After undo:
bild

### This branch
Before undo:
bild
After undo:
bild

Seems to have fixed the problem with undo when a line is marked. Tried this on different elements, both if they were just created or if they were moved. In all cases, the fix seems to have solved the issue.

PR #16009, opened 2024-05-17
G2 2024 v7 issue#15987

b21isabe:
Changed the microservices in the courseedService folder to all assign $userid with the getUid function, look in issue for more info.

This shouldn’t break any systems since $userid is still being retrieved in the same way, just using getUid instead of retrieving it manually. I tested most of the microservices after I updated them and didn't find any issues but a second look would be good.

I tested by using the courseedService test-files and also tried out all the microservices I could find on the website, after switching them out in dugga.js.


a22marca (approved):
Tests are passing and microservices are still working on the website itself as far as I can test. Merging.

PR #16010, opened 2024-05-17
Added classes formBody and formFooter

a20jonda:
This pull is mostly due to unnamed body and footer divs which could cause confusion. The divs are now properly named and will be reported in the [wiki](https://github.com/HGustavs/LenaSYS/wiki/Popup-form-conventions#currently-known-popup-forms-on-lenasys)

Extra:
* Added more conventional headline for the Github repo form
* Did not touch 'Choose Template' because it appears to be dynamic and uses tables.
* Fixed missing body elements and moved some elements
* Fixed misaligned cancel button which became visible with these changes.
* Slight adjustments to some sizes and paddings.

Note: These changes are needed to be in forms in other files, but it is currently not possible yet due to merge conflicts. So only sectioned.php is affected in this pull request.

image
image
image
image
image
image


a22danwe (approved):
LGTM!

PR #16015, opened 2024-05-17
G2 2024 v7 issue#15980

a22melst:
Added fetching and checking of "path" to updateFileLink_ms.php, based on how it is done in the monolith. It does not have a use currently, since path is not set for any file when created using filereceive.php. However, I think it best to include the path-code in the microservice since it is exists in the monolith and will probably be relevant in the future.

### DIFF
--- a/DuggaSys/microservices/fileedService/updateFileLink_ms.php
+++ b/DuggaSys/microservices/fileedService/updateFileLink_ms.php
@@ -30,6 +30,15 @@ $username = retrieveUsername($pdo);
 $debug = "NONE!";
 $log_uuid = getOP('log_uuid');
 
+// Get the path, should be done by using json in the future 
+$query = $pdo->prepare("SELECT path from fileLink WHERE fileid = :fid");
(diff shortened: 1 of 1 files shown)


c22cargo:
It seems that ```$fid = getOP('fid');``` cannot be used because $fid is not sent. Because of this
```
SELECT path from fileLink WHERE kind=:kindid AND filename=:filename
```
was used instead of
```
SELECT path from fileLink WHERE fileid = :fid
```

I will see if I can fix this before approving the review.

c22cargo (changes requested):
You can get the path by using:

```
$encoded_path = getOP('filepath');
$decoded_path = html_entity_decode($encoded_path);
```

I assume this is what you want to retrieve.

This logic will probably need to be updated:

```
if (is_null($path)) {
            $currcwd .= "/courses/" . $cid . "/" . $filename;            
        }
        else {
            $currcwd .= "/courses/" . $cid . "/Github/" . $path;        
        }
```

c22cargo (approved):
Looks good

Issue #15993, opened 2024-05-17
"X" resizes with element height


a22solmi:
## Description
#15883 Introduced a bug where the "X" to delete elements scale with element height. 

## Job 
Change scaling to only work with zoom

PR #16040, opened 2024-05-20
G1 2024 v7 #15993

a22solmi:
Fixed scaling and color of "X" as well as  box select for lines


a22cazlu (approved):
looks good and works fine

PR #16042, opened 2024-05-20
fixed the id on the lock button

a22jacka:
the ID being passed to the button was `id='lockbtn'`, so the ID the button got was `id="id="`, which caused it to not be found by the toggle function

### DIFF
--- a/DuggaSys/diagram/draw/options.js
+++ b/DuggaSys/diagram/draw/options.js
@@ -45,7 +45,7 @@ function generateContextProperties() {
     if (context.length > 0) {
         showProperties(true, propSet, menuSet);
         let locked = context.some(e => e.isLocked);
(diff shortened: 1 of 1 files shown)


a22robko (approved):
I have tested the issue and the button shows "unlocked" and a lock icon appears on the element when pressed. In the weekly branch, the button's behavior did not change, and no lock icon did appear on the element. Additionally, I have tested the functionality in diagram.php and everything appears to be functioning good.

Before:

https://github.com/HGustavs/LenaSYS/assets/129367092/069d77a3-2fef-4687-95b1-41e9b0d4cc66

After:

https://github.com/HGustavs/LenaSYS/assets/129367092/bb95ae10-87aa-476a-a8ee-af367751cb97

PR #16043, opened 2024-05-20
G2 2024 v7 issue#15959

a22marca:
This test only uses the debug property in the json for the assertion test. This is because there is no github url property in the data returned from the microservice. This means it's not possible to compare the actual url that was added. If the microservice fails to update a course with the new github url, the debug value will change which means the test will also fail.


a22iliru (approved):
works

Issue #15440, opened 2024-05-07
ESC to go back in accessed


d22mikwe:
When navigating the website there is usually a back button, as seen in the image:
Implement a keybind ESC that when pressed brings you back to wherever this back button takes you. Ensure that the implementation does not override the exising functionlity connected to the ESC key, such as closing modals/popups.

d22mikwe:
TODO after merged. Look into changing it to keyup?

PR #16049, opened 2024-05-20
ESC to go back in accessed #15440

d22mikwe:
d22mikwe
Collaborator
[d22mikwe](https://github.com/d22mikwe) commented [2 weeks ago](https://github.com/HGustavs/LenaSYS/pull/15559#issue-2287532008)
Fixed current iteration of keyup for enter and esc.
As the popups already has a close function attached to them we only go back if none are open.
Only current popups/modules in accessed are "createUser" and "addUser".
"removeUser" and "createClass" will be added later. These should be inserted into allPopups array when implemented.

For testing:


a22diako:
-When the pop up is open the esc button closes it down. 
-When a pop up is not open the esc button takes the user back to sectioned.php 

Everything that should work and is implemented works well. 
Looks good. Good job!

Issue #16003, opened 2024-05-17
Fix addUser_ms_test.php


c22cargo:
The test is failing. Find out why and fix it.

PR #16051, opened 2024-05-20
G2 2024 v7 issue #16003

c22cargo:
A few changes were made to both the test and the MS.

The expected output for this test does not include everything that ideally should be included. But since users are not retrieved from the retrieve MS, this is what it has to look like.

### DIFF
--- a/DuggaSys/microservices/accessedService/addUser_ms.php
+++ b/DuggaSys/microservices/accessedService/addUser_ms.php
@@ -5,6 +5,7 @@ date_default_timezone_set("Europe/Stockholm");
 include_once "../sharedMicroservices/getUid_ms.php";
 include_once "../../../Shared/sessions.php";
 include_once "../../../Shared/basic.php";
+include_once "retrieveAccessedService_ms.php";
 
 // Connect to database and start session
 pdoConnect();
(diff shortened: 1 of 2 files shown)


DuggaSys/microservices/accessedService/addUser_ms.php
```diff
@@ -154,5 +155,7 @@
         	}
 		} // End of foreach user
 	} // End ADD_USER
+
+	$array = retrieveAccessedService($pdo, $debug, $userid, null, $log_uuid, $opt, null);
```
b21isabe, line 159: The $entries array that shows users in the output is currently showing as empty, this is because the retrieve-service is being called with $courseid = null. If you instead assign it to $cid, users will show up in the output and can then be filtered. Like this:
```
	$array = retrieveAccessedService($pdo, $debug, $userid, $cid, $log_uuid, $opt, null);
```

b21isabe (changes requested):
The test passes now but the output is lacking. The microservice creates a new user and adds it to a course so the most important thing that should be shown in the output is that a user has been created.

I have highlighted some of the problems and suggested some fixes. When these fixes have been made you should update the filter and expected output to show that the user has been created.

b21isabe (approved):
Looks good! Merging.

Issue #16048, opened 2024-05-20
Element class and defaults


a22solmi:
## Description
The element class was supposed to replace the defaults file but couldnt because implementation details in statemachine.
This means that there is duplicate declaration of defaults now

## Job
Make the element class use default values from defaults.js

PR #16052, opened 2024-05-20
G1 2024 #16048

a22solmi:
Made constructor use defaults


b22jonan (changes requested):
this solution breaks the site, just clicking the note causes the entire program to crash:
image

b22jonan (approved):
this seems to have broken example diagrams but i think thats a separate issue since the there have been changes to elements with these changes, but making diagrams as per usual works fine and i dont see any difference in normal use cases. great job 👍

b22jonan:
for reference on example diagrams breaking, loading ER diagram results in this:
image
and the other ones are also messed up so they need to be redone

Issue #15063, opened 2024-04-29
 Remove as much inline css as possible from sectioned.php


a22moamy:
Currently, there is a lot of unnecessary inline css in `sectioned.php`. Having inline css and external css for the same website makes development harder, and thus it should be separated as much as possible into the external css files. Inline css is especially unneccessary for elements with ids and often unnecessary for those without them as well. **Please move as much inline css as possible from `sectioned.php` to the external css files.**

SleepyBearIV:
Remove undo button dose not work remove the code if no longer in use

SleepyBearIV:
can no longer delete stuff

SleepyBearIV:
![image](https://github.com/HGustavs/LenaSYS/assets/96497530/35dc3dbf-07c0-498d-ba49-5d8265a99698)
dose not line up, Probably needs to be dynamic

SleepyBearIV:
![image](https://github.com/HGustavs/LenaSYS/assets/96497530/4eceb217-e3ae-4d26-8c8e-35dffb2d83b8)
![image](https://github.com/HGustavs/LenaSYS/assets/96497530/46c145cb-4c42-451f-a322-8c93b08615f0)
sometimes the cancel button jumps around, should be changed to dynamic

PR #16053, opened 2024-05-20
G3 2024 v7 #15063 v2

SleepyBearIV:
start work on: Remove as much inline css as possible from sectioned.php #15063
everything shude look the same as before on the site compare the two and se if you find any differences


Shared/css/style.css
```diff
@@ -2379,15 +2420,64 @@ Fonts
...
+}
+
+.align_left {
+  align:left
+}
+
+.align_right {
+  align:center
+}
+
```
b22marah, line 2446: Should be align: right;

Shared/css/style.css
b22marah, line 1: Looks over the classes you've made and format them correctly, some are good, but alot are not following the syntax of formating we're using:

Faulty formatting: 
.class{
 margin:20px;
}

Correct formatting:
.class {
  margin: 20px; 
}

DuggaSys/sectioned.php
```diff
@@ -162,30 +162,30 @@ function updateGithubDir($pdo, $githubDir, $lid) {
...
 		<div id='Sectionlist'>
 
-		<div class='course' style='display:flex; align-items:center; justify-content:flex-end; '>
+		<div id='courseHeader' class='course' >
 			
-			<!-- Undo button -->
+			<!-- Undo button 
 			<input id="undoButton" value="&#9851;" type="button" class='submit-button-newitem' title="Undo deleted example" style="position: absolute; padding-right:5px; margin-right:165px; display: none;" onclick="cancelDelete();">
-			<!-- Undo button END -->
+			 Undo button END -->
```
b22marah, line 169: You need to revert this code... When you open and close a comment with code within it, the code will also be commented, therefor this change makes runnable code not runnable anymore, since it's now a comment.

b22marah: Also fix the styling for the input here, remove inline

b22marah (changes requested):
After fixing out-commented code, fix the inline css for this part of the code, and formatting the css correctly everything should be fine, the site looks like it should, so besides these small changes everything LGTM!

DuggaSys/sectioned.php
SleepyBearIV: the button is not used or ever apers anymore so i commented it out because i did not know if the person working on it would need it again, it is replaced by a pop-up toast

b22marah (approved):
LGTM!

Issue #16044, opened 2024-05-20
Add missing sorting functions for positions for section add/delete functions


a20jonda:
* Currently, duplicate positions are created when adding a new section element. 
* Deleting a section also leaves empty positions in the database.
* There needs to be an simple function implemented which shifts elements

a20jonda:
Deleting is currently not possible and will be on hold for now

PR #16061, opened 2024-05-20
Added insertion sort function to createNewListentry in createNewListEntry_ms.php #16044

a20jonda:
Since the delete function is not currently testable, only sorting when adding to `listentries`  will be implemented in this pull.

To test 
1. Create a new course
2. Add several elements
3. Ensure that there is no duplicate positions in the database. (positions start at 1)
image

4.Optionally, add a github fetch to course and choose code examples for moment section (don't forget to press refresh button on the section) and  then check if the added code examples appear under correct moment section.


a22edvbu (approved):
Everything seems good to me. No duplicate pos for moments in listentries. Great Work!

PR #16065, opened 2024-05-20
G3 2024 v7 #16038

### LINKED ISSUE #16038: Add generic modal implementation for new installer
Add a generic modal implementation for the new installer. 
This modal will be used to display errors in the future and allow the user to retry or change settings if it fails. 

Requirements:
- Implement modal.
- Responsive design.
- Should be simple to use (one function call should be enough to use it. 
- Should be able to add various different buttons in the bottom left, right etc.
- Recommended to make it object oriented.

b22marah:
## What has been done

Implemented a modal following the OOP structure. The modal is generic and works for multiple tasks, the functions that has been included is for creating a unique modal, one used for sql queries, one for database overriding, one for forcing creation of database, and one for supplying user with commands when right are not sat properly.

Included here is:
* Javascript for the modal funcionality.
* PHP class for the modal.
* CSS for the styling.

## For the tester

Not much to try right now, since it's not usable in real situations at the moment. Therefore just calling the different functions to see that the correct header, text, and buttons are being served is enough for testing.

The test should include:
* Creating a button with the id: openModal. (onclick listener for this id is alredy implemented for you.)
* Try different functions to see that they work and provide correct data.
* See if it's responsive, should work with mobile devices, but doesn't need to be perfect here since LenaSYS is not used on mobile devices usually.

The code is approved when:
* The modal is somewhat responsive.
* Code looks well-planned and formatted.
* Correct data for each function call is provided.

Closes #16038

### DIFF
--- a/newinstaller/installer_ui.php
+++ b/newinstaller/installer_ui.php
@@ -7,9 +7,11 @@
 	<link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Material+Symbols+Outlined:opsz,wght,FILL,GRAD@20..48,100..700,0..1,-50..200" />
 	<link rel="stylesheet" type="text/css" href="style.css">
 	<?php include 'tools/components.php'; ?>
+	<?php include 'tools/modal.php'; ?>
 	<script defer src="installer_ui.js"></script>
 	<script defer src="tools/components.js"></script>
 	<script defer src="tools/sse_receiver.js"></script>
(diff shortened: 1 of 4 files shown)


Sinmick (approved):
LGTM!

d22mikwe:
Works for me. LGTM!

d22mikwe (approved):
LGTM!

Issue #16056, opened 2024-05-20
Fix updateUser_ms_test.php


c22cargo:
The test is failing. Find out why and fix it.

PR #16067, opened 2024-05-20
G2 2024 v7 issue #16056

c22cargo:
Both the test and the microservice was updated.

I had to change ```$username = getOP('username');``` to ```$username = getOP('user_name');``` in the MS because of duplicate entries in the service-data. In the monolith they get around this by using ```$val``` instead of separate variables like ```$firstname```.
I wasn't able to find any frontend use for the microservice, but it's possible that further changes need to be made because of the name change.

Perhaps the MS should use ```$val``` instead? I assumed the way it was implemented is how it's supposed to look, so I didn't change it to use ```$val```.

I also added an OPT, but I don't think anything more needs to be changed regarding that since the microservice hasn't been integrated.

### DIFF
--- a/DuggaSys/microservices/accessedService/updateUser_ms.php
+++ b/DuggaSys/microservices/accessedService/updateUser_ms.php
@@ -1,54 +1,77 @@
 <?php
-date_default_timezone_set("Europe/Stockholm");
-// Include basic application services!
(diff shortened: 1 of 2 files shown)


a22marca (approved):
All tests are passing. Merging this.

Issue #15887, opened 2024-05-16
Fix arrows on straight lines


a21liltr:
https://github.com/HGustavs/LenaSYS/assets/129370135/6ede40c1-2031-4641-8d44-a4974eb5cef3

## Description
For SD entities and UML Initial State as well as UML End State, some error causes the arrows to not show on straight lines.

## Job
Fix the error. Make sure that:
* SD entities can have arrows going to to and from both sides,
* UML Initial State only has arrows going out,
* UML End State only has arrows going in.

For testing later: Make sure no other arrows have disappeared or that other elements doesn't have multiple arrows overlapping on one or both ends. And that the lines between the elements still work.

PR #16069, opened 2024-05-20
G1 2024 v7 #15887

b22davbe:
You can now create arrows on straight SD-lines. Those arrows are also calculated to look good on a straight lines with diffrerent angles (see printscreen). 
image
 
Also implemented a new function that checks if an SD-line is connected to a initial or final-state and if that is the case you can no longer change the arrows on that line in the option-panel. This results in the standard-arrows always being in use where if you connect a line from an initial-state an arrow will be drawn FROM it and if the line connects to a final-state the arrow is drawn TO it.
image


a22vioja:
Closes issue #15887

a22wilas (approved):
Old version:
bild
bild

This branch:
bild
bild
bild

Checked all the changes and in the old version no arrows were possible even if they were selected in the properties' menu. In this branch, you can select arrows on both sides of the line if it's connected from SD to SD, and it automatically creates an arrow for the initial and end states when they are connected to a SD entity.
The change has fixed the issue and no new error has occurred.

PR #16072, opened 2024-05-20
Updated JSDocs #15798

a22solmi:
Wrote new JSDocs and modified previously written docs


b22davbe (commented):
Theres two console logs in diagram.js that i don't think should be there. The updated JSDocs looks good to me so if the console logs should be in it's approved and if else just delete the logs.

PR #16077, opened 2024-05-20
[Documentation] update.js

a22robko:
It's possible that there may be small errors since it's difficult to know exactly what is expected in a function, but I tried to make it as good as possible.


a23ponek (approved):
Documentation changes looks good and nothing seem to break or cause any problems in the diagram.

PR #16080, opened 2024-05-20
whitespace tag for the break to work

a22diako:
A white space has been implemented for the paragraph to break. 
Other functions have been implemented, but white space is the only one to have an effect. 

Click on the icon for the hover to apear, inside the hover- check that the paragraph breaks and DO NOT go outside the dropdowns whitelane. 

Screen of how the hoover looked before p break:

Skärmbild 2024-05-20 141630

Should look now:

Skärmbild 2024-05-20 163043


a22edvbu (approved):
Looks good to me. Great work!

PR #16081, opened 2024-05-20
G1 2024 v7 #15786

a22ricni:
In DuggaSys/diagram/draw/options.js are some documentation for the function.


a22robko (approved):
Looks good, the comments are clear and explaining what each part does. And nothing seems to break or cause any problems in the diagram as well.

PR #16086, opened 2024-05-20
modified the naming scheme for all constants, improving code maintain…

d22abdha:
…ability, while keeping constant objects unchanged.  During testing, I encountered a challenging TypeError related to the toLowerCase() function within the isKeybindValid function, triggered by undefined properties during key event handling. This issue highlighted the need for thorough debugging and system resilience.

modified the naming scheme for all constants, improving code maintainability, while keeping constant objects unchanged. During testing, I encountered a challenging TypeError related to the toLowerCase() function within the isKeybindValid function, triggered by undefined properties during key event handling. This issue highlighted the need for thorough debugging and system resilience.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -480,7 +480,7 @@ function getData() {
     document.getElementById("container").addEventListener("mousemove", mmoving);
     document.getElementById("container").addEventListener("wheel", mwheel);
     document.getElementById("options-pane").addEventListener("mousedown", mdown);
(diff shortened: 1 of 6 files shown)


a22solmi (changes requested):
The object value is and should be unchanged but when comparing a with element.kind you have capitalized the string. This will never match with anything.
```element.kind != 'SD_ENTITY' ```
```SDEntity: "SDEntity" ```

a20artje (commented):
Conflict, new file has been changed since this commit. Cannot merge.

Issue #15923, opened 2024-05-16
Add ID functionality to loadingButton.js and switch out buttons


a22maxra:
Add ID functionality to `loadingButton.js` so that when creating buttons you can give them an unique ID as one of the parameters here: 
![image](https://github.com/HGustavs/LenaSYS/assets/98206230/d832148a-eb22-48bd-ad9e-eb3b84e716e8)

And switch out buttons so it uses the loadingButton.

a22maxra:
![image](https://github.com/HGustavs/LenaSYS/assets/98206230/3bcfa1bf-49da-49d1-8f38-5255e0e50830)
For some reason this weekly branch makes the homepage courseed.php not load for me i have been troubleshooting for a long time i don't know why but i can still enter a sectioned.php page so i'll just do the issue and let someone test the buttons on the courseed.php later.

a22maxra:
There is now a loading button for editing a selected course version
![image](https://github.com/HGustavs/LenaSYS/assets/98206230/64e1ac62-546a-45aa-9eb9-c359fc40fde0)

And you can give a button an ID
![image](https://github.com/HGustavs/LenaSYS/assets/98206230/78294b73-5232-4468-9afb-a016aae1f406)

a22maxra:
Buttons now fade out when they are disabled as such
![image](https://github.com/HGustavs/LenaSYS/assets/98206230/5a0cc086-5efd-4c2f-866a-726189edbd02)

a22maxra:
Updated the WIKI to include buttonId

PR #16091, opened 2024-05-21
loadingButton.js functionality #15923

a22maxra:
Buttons can now be given IDs, such as this one has:
image

They also fade out when disabled:
image

This applies to buttons which uses buttonLoader.js


a22edvbu (approved):
Looks good to me. Great work!

d22mikwe:
LGTM

PR #16100, opened 2024-05-21
Moved the "First Name" label to the correct place

a22natth:
Moved the "First Name" label to the correct place since it was in the header element previously. To access this, go into a course and press the keylock button. This is what it looked like before:
image

And this is what it looks like now:
image


a18chrkr (approved):
Nice! Looks good in both light and dark mode. Also seem to be consistent through different screen sizes. It gets my 'Seal of Approval'! 🐋

PR #16104, opened 2024-05-21
Update errorHandling.js

a22wilas:


a22wilas:
Removed unnecessary code that checked if two different relations had the same name. This was over 500 lines of code that probably was used in an older version of the application.

One thing that now is allowed is having two relations with the same name that is connected to the same entities. This should not be possible and a new issue for this should be created.

a22wilas:
bild
The error check does not allow the same name on entities, how did you review this?
Some screenshots and information how the review went would been good!

a22solmi (approved):
LGTM

PR #16106, opened 2024-05-21
G2 2024 v7 issue#16092

a22marca:
I have removed the include of "coursesyspw.php" at the top of files that already includes "sessions.php". 
I checked if this broke anything by navigating to the part of the sites that is still used in LenaSYS. Tests and files like e.g. codeview.php still works without any issues.


a22iliru (approved):
Looks good

Issue #16105, opened 2024-05-21
Integrate the fileedService microservices


c22cargo:
If it's possible the microservices should be used instead of the monolith.

PR #16108, opened 2024-05-21
G2 2024 v7 issue #16105

c22cargo:
The microservices have been integrated. See dugga.js at ```kind=="FILE"```.

I tried creating getFileedService_ms.php, however I couldn't get it to work. I didn't want to broaden the scope of this issue too much so it will be looked at in another issue. Because of this the "GET" opt still uses the monolith.

You can test the integration by going into a course, clicking "show files" in the toolbar (the plug icon), and then updating and deleting files. All different kinds of files/links should work correctly (global, etc).


c20juler:
Reviewed and tested. Tested according to description, can delete and update different kinds of files for courses. 
Microservices are successfully integrated.

Issue #16076, opened 2024-05-20
Fix dark mode alignment for forms


a20jonda:
![image](https://github.com/HGustavs/LenaSYS/assets/102598258/3f75abc7-cc4b-49cf-8318-7bba08a406e2)
The 'visibility' select element is shifted 10px in dark mode and does not use the same dark color as the other input elements.

a20jonda:
![image](https://github.com/HGustavs/LenaSYS/assets/102598258/8f04fe96-07af-4d72-af64-d9ac87af1ecf)
The same problem appears in sectioned.php. The alignment should be the same as the "name" input field.

a20jonda:
![image](https://github.com/HGustavs/LenaSYS/assets/102598258/cadb2dd7-b13c-479b-9185-5463c809e21b)
Just for reference, the problem comes from darkmode.css I believe

PR #16110, opened 2024-05-21
Updated blacktheme.css (#16076)

a22chrfa:
There was a margin on the select element, that has been removed. There was also a background color not using the predefined color variables. I guess it's used for something, but it was applied to the formBox select as well, so I removed that. It seems to be working fine.

I suspected the loginbox might have had consequences, but it didn't. It looks like it dide before the changes.

When editing a course. Det dropdown should now be aligned with the rest of the inputs.
image


b22chran (approved):
The alignment looks good in darkmode now! The code looks correct as well.

Issue #15231, opened 2024-05-01
Make Name input change the codeexample name


d22mikwe:
Currently Name input is not used when fetching codeexamples.
Implement code so that the name input is set as the codeexample name

PR #16111, opened 2024-05-21
Make Name input change the codeexample name #15231

d22mikwe:
The Name input is now updated to the list entry name and the codeexample name.

Testing:
Create a new code example.
Fetch a file/files from the github icon.
Click Save.
image

The Name input should now be displayed as the list entry name and code example name.
image

Title is also updated (codeexample name)
image

This can be doublechecked in the db table "listentries" and "codeexample"


a22chrfa (approved):
Seems to work perfectly. The "storeCodeExamples()" doesn't seem to be used anywhere else so any unexpected errors elsewhere shouldn't be a problem. Solid work!

PR #16114, opened 2024-05-21
G2 2024 v7 issue#16097

a22marca:
Both createNewCodeExample and createNewListEntry are now returning a debug value. This value is used in createListEntry.


a22melst (approved):
Good solution

Issue #16096, opened 2024-05-21
Find a better solution than setting users to superusers with: $userid="1"; by default


a22albjo:
In 6 service files code like this can be found:
```
	if(isset($_SESSION['uid'])){
		$userid=$_SESSION['uid'];
	}else{
		$userid="1";
	}
```
It makes the user a superuser if not logged in.

Here are the files:
<img width="240" alt="Skärmavbild 2024-05-21 kl  10 22 11" src="https://github.com/HGustavs/LenaSYS/assets/53318057/2742a418-b404-4165-986a-07e6481e6d6b">

This makes the user have superuser privilages by default in the service file. It seems that it has been done to make tests work, however this is not a good way to make a test work.

PR #16115, opened 2024-05-21
G2 2024 v7 issue #16096

c22cargo:
```$userid``` was changed from "1" to "guest" in several php files. To make this work, ```checklogin()``` had to be added before setting ```$userid```.

Markdown files were not changed.


a22marca (approved):
Test by going into a codeexample as a guest and as a superuser. Everything works. Merging

PR #16117, opened 2024-05-21
Added a check for data entry

b22chran:
Added an if-statement that checks if the data in the entry "grpmembershp" is not null to prevent the code from brake from the split() function of null-value.


a22maxra (approved):
Adding a group works and there are no errors in the console

PR #16120, opened 2024-05-21
g2-2024-v7-issue#16070

b21isabe:
The updateFileLink tests should work now.

There were some error messages from sending arrays in $data with curl. So I deleted the arrays and just sent the variables and it seemed to work.

For the failing tests, the formatting of the output had changed but the contents of the output was the same. So I just updated the expected output with the new formatting.


c22cargo (changes requested):
Test already works on group branch.

PR #16122, opened 2024-05-21
Fixing buggs with github repo button 

a22karja:
The timer no longer starts without a Github repo link

If the string dosent contain certain elements that all github links have you cant save them

WARNING TO TESTER:
If you get a red toast warning when pressing the repo button you have to enter a github token on that course. You do this here.
image

It has to be a valid github token not just random text

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -3685,6 +3685,37 @@ const regex = {
 	githubURL: /^(https?:\/\/)?(github)(\.com\/)([\w-]*\/)([\w-]+)$/
 };
 
+function checkGithubLink(link) {
+  var element = document.getElementById(link);
+  var savebtn = document.getElementById('buttonContainerSaveRepo').children[0];
+  console.log(element.value);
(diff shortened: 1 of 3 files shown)


a18chrkr (changes requested):
Looks good and the implementation works. The 'github.com' part of the link is case sensetive though, which will be no problem if the user copy the link (which will probably be most cases) but this _could_ be considered, but is by no means necessary to change.

Other than that there is a few inconsistencies to spaces before curly brackets and also some parameters are declared with space before and after '=' while some are not.

Below is a screenshot of some examples:

whiteSpaceForCurlies

Double check the code and make sure everything is consistent - then this is an ez approve!

a18chrkr (changes requested):
Still a couple variable declarations that e.g. look like this "status=true". Please include space between '='. Also change 'var' to 'let' for declarations.

a18chrkr (approved):
Changes look good! Seal of Approval! 🐋

d22mikwe:
LGTM!

PR #16130, opened 2024-05-21
G2 2024 v7 issue#16125

a22marca:
Added checks for login, permissions and opt before the queries in the microservice are made. If any of them fails a return is made.


a22iliru (approved):
Good

PR #16131, opened 2024-05-21
made md file and wrote instructions

a22eddpe:
go through the instructions. Maybe any inputs?

### DIFF
--- a/includeInstructions.md
+++ b/includeInstructions.md
@@ -1,0 +1,32 @@
+## How to think with includes: 
+
+## Microservices:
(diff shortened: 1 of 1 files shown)


c22emigr (changes requested):
Documentation looks good and makes sense but you should fix the spelling for include_once (not include_one).

a22eriwi (commented):
Fix the spelling error that c22emigr wrote about, everything else looks good!
After the changes are made we can merge this file

PR #16132, opened 2024-05-21
G2 merge issue#16129 

a22marca:
Added a check for login and change some incorrect parameters in retrieveCourseedService calls.


b21isabe (approved):
Tested by first running the test-file with brom as the user and it passed, then tested it with a student account instead and it failed as it should. I then tested the microservice on the website. I could update the version name, start-date and end-date while logged in as brom so the authentication seems to be working as expected.

I could not update the MOTD but this seems to be a separate issue so I am approving this and merging.

PR #16133, opened 2024-05-21
Changed animatiom for new list items in sectioned.php

a22chrfa:
The animation was done by the glowAnimation keyframes.

It has been made 20% faster (8 seconds instead of 10) and the color has been changed to the same green that is used as the background for the logout button, as requested.

For testing:
Create a new code item ( or whichever item you want) and make sure that the animation of the row is a fading green over 8 secs.


a22moamy (approved):
Looks good!

PR #16134, opened 2024-05-21
begining on Remove as much inline css as possible from fileed.php #15065 will continue after merge

SleepyBearIV:
check the header on the different sites


d22mikwe (approved):
LGTM! Everything looks the same compared to before

Issue #16098, opened 2024-05-21
Remove unused variables in processDuggaFile_ms.php


a22albjo:
There are unused variables in row:
14,15,19,38,44,47.

Check to see if the services works if they are removed, it should.

The ununsed variabled can be found using intelephense

PR #16135, opened 2024-05-21
Removed unused variables in processDuggaFile_ms.php #16098

a21sahfa:
Row 14: Removed $tmphashpwd
Row 15: Removed $tmpvariant
Row 19: Removed $result
Row 38: Removed $currcvd
Row 44: Removed $isFileSubmitted (it was set but not used)
Row 47: Removed $isFileSubmitted (it was set but not used)


a23gabli (approved):
The changes look good, i have checked that the duggor load and are correctly categorised under courses.

Issue #16151, opened 2024-05-22
Copy course files from old installer to new installer


a22willi:
The new installer is currently using the SQL files and course files located in the directory for the old installer. Copy these over to the new installer directory.

PR #16158, opened 2024-05-24
Merge g3-2024-v8-#16151 into g3-2024-v8

a22willi:
The new installer is now using installation files located in the same directory as the new installer, meaning they are separate from the files in the old installer. 

> **NOTE:** Do not be alarmed by the high number of files, this commit contains all the SQL installation files, and demo course files.


a22moamy (approved):
Looks good :)

Issue #16167, opened 2024-05-31
Deleting multiple items in courseversion doesn't work


a22chrfa:
There seems to be multiple issues surrounding this. My intention is to do them all, but they will be split into smaller more manageable issues in an arbitrary fashion.

![image](https://github.com/HGustavs/LenaSYS/assets/128893264/095988d6-b875-4ad6-ad51-2eab38d0de9f)

In the image, I have marked "HTML5 Example 2", "HTML5 Example 3" and "HTML5 Example 4". When clicking the trashcan and clicking "yes" in the confirmation box. Only the first item in the list is removed. Furthermore, if clicking "yes" again, nothing happens.

a22chrfa:
Log:
Back from lunch. I just popped in to the database to confirm what happens there. I have heard from previous conversations that the items are in certain cases only hidden at first due to the undo selection and so on. So I just wanted to confirm what happens in the database.

The first item is correctly removed from the database. The second item (even if I click "yes" again) is not removed from the website, and is not removed from the database. So the web seems to reflect the database in this regard.

I will now look at the code and document how this process actually works in the background.

a22chrfa:
Log:
added some potential issues to the original issue-text.

a22chrfa:
When marking items, the items are stored in an array called "selectedItemList". I've added an eventlistener in order to track this list during interactions.

Action called when clicking trashcan:
confirmBox("openConfirmBox", this);

When marking multiple items and clicking one of the trashcans, the confirmation windows pops up. When clicking yes, this is what the log looks like:
![image](https://github.com/HGustavs/LenaSYS/assets/128893264/36b51c06-2121-470b-a242-95da763477f5)

Even when closing the popup window, the array looks the same. So one issue is that the item isn't cleared from the array.
Another issue is that it doesn't clear the entire array. It seems from both the UI and the function that this was meant to only delete one item.

a22chrfa:
![image](https://github.com/HGustavs/LenaSYS/assets/128893264/ee327466-dc57-44fd-bbcd-c89493051434)
![image](https://github.com/HGustavs/LenaSYS/assets/128893264/a6b16b0e-1e5a-4a69-ba6b-d51514e1aeb2)

The attached images visualize the callstack up to deleteAll() which then makes an ajax call.

[...]

PR #16168, opened 2024-06-03
G3-2024-a22chrfa-#16167

a22chrfa:
It is now possible to select multiple items and deleting them with the trashcan. Unlike before, the array containing the selected items is also cleared after deletion, which means you can select n items, delete them, and then do the same to a new set of items. This required reloading the page before since the array wasn't cleared.

**NOTE:**
1. The undo button wasn't showing before, but is thanks to the code not stopping because of errors. It is not working however.
2. The styling (which was the cause of log errors) are not working properly. Hence the log errors, which is why they've been commented out.
3. It's a bit weird that you have to select a row when each row has a trashcan. It's also non-intuitive that after selecting multiple rows, you can click any trashcan belonging to an item row (rather than a separated trashcan that would be more in line with UXD principles). But this is the design laid out.

Those issues are considered outside the scope here, but had to be commented out due to them not functioning. Be aware when testing that UNDO is not working and is NOT inside the scope here.

**For testing:**
Select a few items in the and click a trashcan, they should now disappear. Now do it one more time without reloading the page.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -30,15 +30,6 @@ var backgroundColorTheme;
 var isLoggedIn = false;
 var inputColorTheme;
 
-document.addEventListener('keydown', function(event) {
-  if (event.key === 'm') {
-      console.log("eventLOG: ", selectedItemList);
-      // Add any action you want to perform here
-  }
-});
-
-
-
 function initInputColorTheme() {
   if(localStorage.getItem('themeBlack').includes('blackTheme')){
     inputColorTheme = "#212121";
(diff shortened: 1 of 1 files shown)


a22chrfa (commented):
Review:
Event handler used for logging should be removed first.

a22chrfa:
Now that the eventhandler is removed the code looks fine. I once again tested the functionality and it works as intended within the scope of the issue.

**NOTE:**
It seems I can't approve, changerequest, nor deny a pull request where I'm not the reviewer. It also seems it isn't allowed on github to assign yourself as the reviewer. So this is what I can do at the moment.

a23dilel (changes requested):
I have tested the functionality and it's works well, but there are some code comments are unnessary and should be removed or uncomment if codes are important and testet it again. If this is resolved then issue manager or group leader can merge this branch.

Also a22chrfa mentioned here (https://github.com/HGustavs/LenaSYS/pull/16168#issue-2331827347) that we need to create three seprate issues which are add funtionality undo button, fix styling error which are commented out and fix UXD principles trashcan.

Issue #15637, opened 2024-05-13
Impossible to draw two arrows separately in Sequence diagram


a22vioja:
## Description
It is not possible to draw two separate arrows from one sequence object because it appears as a branch of one arrow. 

<img width="518" alt="Screenshot 2024-05-13 at 10 52 35" src="https://github.com/HGustavs/LenaSYS/assets/129263559/08ce9971-0188-49c6-a851-e07ba197f3bd">

# Job
Make it possible to draw two separate arrows from one Sequence Object.

PR #16170, opened 2024-06-05
Komplettering b22jonan #15637

b22jonan:
solved #15637 by adding offsets to lines depending on how many elements they share with other lines, it takes into account which direction the other lines are going into aswell. remaking the image shown in the issue results in this:
image

### DIFF
--- a/DuggaSys/diagram/draw/line.js
+++ b/DuggaSys/diagram/draw/line.js
@@ -3,7 +3,7 @@
  * @param {Object} line The line object that is drawn.
  * @param {boolean} targetGhost Is the targeted line a ghost line
  */
- function drawLine(line, targetGhost = false) {
+function drawLine(line, targetGhost = false) {
     let str = "";
     // Element line is drawn from/to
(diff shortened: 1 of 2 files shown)


b22jonan:
since i cant assign myself to review this.
the solution works pretty well from what ive tested, the offsets can be a bit wonky depending on the order which you click in since they offset on the x or y axis depending on the direction, not looking at where the other lines are, so it can result in cases like this:
image
but its still pretty clear that they are separate lines so it still works

b22jonan:
reviewed this while directly after i worked on it, checked again now 2 months after and it still works, looking at the code it still looks good to me so i approve these changes

a23aleva (changes requested):
Tested b22jonan's solution in the lenaSYS diagram creator and the changes made doesn't seem to make a difference. When drawing two arrows to the same event container they still snap together instead of standing separately on their own. Since the problem remains perhaps this pull request could be issued once more to get fixed.
testdiagram

Issue #16169, opened 2024-06-04
The new toast system is missing in fileed.php


a22chrfa:
There was an issue created (#16121) which has not yet been acted upon. No commits has been done as far as I can tell.

The issue reads "Integrate toasts to replace all alerts in fileed.php. I'e when deleting a file an alert is shown."

I've taken a quick look. The alert is not only an alert, it is a chance for the user to select yes or no in case a trashcan was missclicked. I don't think this should be removed, but I still think implementing the toast as a confirmation of deletion is good idea.

a22chrfa:
Read the documentation here: https://github.com/HGustavs/LenaSYS/wiki/Toast

Have made a first implementation and test. Seems to work. As of now the toast always shows after clicking the button. Will have to add conditionals.

a22chrfa:
I've tried wrapping the call in AJAXService in a promise-object´:
![image](https://github.com/HGustavs/LenaSYS/assets/128893264/401404ce-44a0-443f-bc5c-4872717c4a0c)

And then using this to show the correct toast and log:
![image](https://github.com/HGustavs/LenaSYS/assets/128893264/35815f65-efc8-4653-b249-2954e57ad2a7)

But it isn't working. It seems the promise isn't working. I'm not sure why, and I would have to research promise objects more in order to solve this.

Since it's taking up too much time and research would only result in me spending time. I will likely make a pull request with the toast for deletion, as the toast is implemented, but the handling of errors would probably be better suited in a separate issue since the first few attempts suggest that it might need a more thorough looking into.

a22chrfa:
The problem is that the confirmation box uses the alert. And the alert that is part of another library is making a follow up check.

Essentially, the toast is now showing after clicking yes. But when clicking yes you might get information that it can't be removed. Still, the toast is there.

This would require alot to fix. And there is no time in this issue. My problem now is that I've produced 2 lines of code (not sufficient) and trying things that in the end led nowhere, or that a new issue should be created.

a22carsv:
Issue #16121 and #16169 are essentially the same and will hence be treated as one in this issue. 

The problem is essentially that we want to use the toast system to deliver alerts/notifications in fileed.php instead of the current where a pop-up appears:

![Image](https://github.com/user-attachments/assets/c8941c17-a6b5-4237-870a-4ebec9421821)

The plan is to implement the toast system in 3 different scenarios for fileed.php:
- When user deletes file
- When user edits file 
- When user adds file.
[...]

PR #16171, opened 2024-06-05
Toast implemented to fileed.php for #16169

a22chrfa:
Toast is now included in fileed.php. The toast function has also been added to fileed.js under the deleteFile-function. As of now, it always toasts when clicking "yes" due to a more fundamental aspect in how confirmation is handled on this particular page. I tried implementing conditionals by wrapping the the ajax-call in a promise but was ultimately unsuccessful. As this would require more research on promise objects in order to increase my own understanding - this issue would be lengthier and would therefore be better handled in a follow-up issues.

**Testing:**
Delete an item fileed.js (those that can be deleted, some are tied to code examples) and see the toast appear.

### DIFF
--- a/DuggaSys/fileed.js
+++ b/DuggaSys/fileed.js
@@ -789,6 +789,7 @@ function deleteFile(fileid, filename, filekind) {
     if (confirm("Do you really want to delete the file/link: " + filename)) {
         AJAXService("DELFILE", tempData, "FILE");
     }
+    toast("success", "Item deleted", 6);
 }
 
 
--- a/DuggaSys/fileed.php
+++ b/DuggaSys/fileed.php
@@ -4,6 +4,7 @@ session_start();
 
 include_once "../Shared/basic.php";
 include_once "../Shared/sessions.php";
+include "../Shared/toast.php";
 
 // Connect to database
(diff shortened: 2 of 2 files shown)


a22chrfa (commented):
Code seems to adhere to the wiki. I'm not sure if include is correct over include_once in this case. But I think it's a good idea to adhere to the written wiki.

a22chrfa:
This is ok. If this was in the project - it would be OK with the condition that a follow up issue was created that either altered the confirmation box, or that another developer looked into the promise-object so that different cases could be handled, since the current developer has tried but failed.

a23kevre:
I think include_once should be used here, since its only used once. If the toasts are used in more than one place with other condition it might aswell stay like this. To later be updated with more toasts by another developer in a follow-up issue regarding toasts. Its good to get a popup when a delete is successful so we should keep the code, definitly. :)

a23kevre (commented):
I think include_once should be used here, since its only used once. If the toasts are used in more than one place with other condition it might aswell stay like this. To later be updated with more toasts by another developer in a follow-up issue regarding toasts. Its good to get a popup when a delete is successful so we should keep the code. :)

PR #16174, opened 2024-06-06
komplettering b22jonan 15751

b22jonan:
removed the ability to add lables to lines between ER elements, labels are still available to other lines. 
the solution was to remove 
"str += includeLabel(line);"
from the ER case in drawlineproperties in options.js


b22jonan:
cannot add myself as a reviewer in github, but coming back to this 2 months after making this change i can say that this still works fine as you cant add a lable to an ERline:
image

Issue #16172, opened 2024-06-05
Create a canvas background in courseed.php


a22chrfa:
courseed.php is not only made for teachers, but also for students. In order to make it more appealing, draw attention, and maybe (since it's a school) show what can be done with animations, we want to create a canvas using animations and/or some sort of interactivity.

It could be something simple or advanced. The important thing is that it should look good and draw attention. The developer is largely free to design something subjectively appealing and give his suggestion in the form of a pull request.

a22chrfa:
Info:
The plan was to use mathematical expressions instead of dots, which I thought was a neat idea considering it's clear relation to education. I tried the Mathjax library but it seems to be dependent on dom manipulation to generate the expressions which doesn't work well with canvas animations. Dom manipulation means this:
![image](https://github.com/HGustavs/LenaSYS/assets/128893264/9e4807aa-9251-4264-afd6-17f94db470b4)
I tried converting the equations to images instead but this means creating and removing elements (as in the picture) in a loop. This didn't work well and started to feel like bad design.

This idea is probably a dead end. I'll have to make some other trail effect instead.

a23elism:
Awaiting review for "solution" to issue #16172 in PR #16374

a23erigu:
This issue appears to not really be a issue to begin with so it will be closed

PR #16175, opened 2024-06-06
Added trail effect to pointer in courseed.php #16172

a22chrfa:
In order to show something a bit cool I created a trail effect for the pointer. With this as a starting point, it could be further developed by anyone interested.

image

The pointer isn't shown in the screenshot, but it is to the bottom left of the trail.

**For testing:**
Play around with it. More important here is to check for coding standards

### DIFF
--- a/DuggaSys/canvasanimation.js
+++ b/DuggaSys/canvasanimation.js
@@ -12,8 +12,14 @@ resizeCanvas();
 window.addEventListener('resize', resizeCanvas);
 
 const trailArray = [];
-const size = 10;
-const dotLife = 20; //the number means the amount of frames
+const size = 4;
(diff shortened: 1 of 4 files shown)


a22chrfa (commented):
The code seems OK, apart from the if-statements in the update-method where the wrong variable (red) is used for all statements (green, blue).

a22chrfa (commented):
Apparently I missed that one declaration "green" is missing it's semicolon and that the variables are declared without any keywords. Should be fixed.

a22chrfa:
Apart from missing a few things during review, the could should finally be ok.

c22isako:
Cannot click on anything, I believe that the "blobs" are blocking the hyperlinks and therefore breaks the functionality. 
It looks cool though!

Issue #16176, opened 2024-06-07
Change highlights in fileed.php to comply with Gestalt laws


a22chrfa:
![image](https://github.com/HGustavs/LenaSYS/assets/128893264/f44acb8e-c9fb-4bc9-b92d-6d6adf856c67)

This seems to break some gestalt laws and therefore makes it a bit unintuitive. While understanding possible actions in this case might be relatively easy, it would probably be better to adhere to the gestalt laws any way.

According to the gestalt laws, when elements (in this case in a row) belong together, they should be visualized as such. When hovering over cells the highlight emphasizes specific cells within the row, which might mistakenly be interpreted as if the cell could be interacted with independently. Instead, we want to highlight the entire row to inform the user that the items in the row belong to the same group or item. The editing icon (pencil) suffices to show the user that when clicked, the entire row will be edited.

a23simry:
Issue is no longer relevant, fix has been made by unknown pull request

PR #16177, opened 2024-06-10
Edited styling for files table in fileed.php (#16176)

a22chrfa:
1. fileed.js instantiates a new SortableTable object. This also had its property "hasRowHighLight" set to true which I instead set to false.
2. I added a new css rule for all even table rows which uses the same color as the color-primary variable, but it uses an 8 digit hex code to include the alpha channel.
3. I changed the styling of the td-elements to the tr-elements, and set the border on top an bottom instead of left and right.

It works for both black and white theme. This is the result:
image

**For testing:**
Go to fileed.php and see the list. Hover your pointer over the list to check the highlight. Check for both themes.

### DIFF
--- a/DuggaSys/fileed.js
+++ b/DuggaSys/fileed.js
@@ -112,7 +112,7 @@ function returnedFile(data) {
         renderSortOptionsCallback: renderSortOptions,
         rowFilterCallback: rowFilter,
         columnOrder: colOrder,
-        hasRowHighlight: true,
+        hasRowHighlight: false,
         hasMagicHeadings: false,
         hasCounterColumn: true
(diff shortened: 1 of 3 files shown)


a22chrfa (commented):
Looks good. Since the instantiation in fileed.js uses the hasRowHighlight no functionality has to be edited, we can therefore be very certain that this change won't lead to any surprises elsewhere. Also checked that the highlight and the borders didn't change elsewhere, and it doesn't look lite it.

This pull is therefore OK.

a23yutta (approved):
The code works and fullfills it's purpose

Issue #16161, opened 2024-05-24
Elements top middle node missing


a21liltr:
![image](https://github.com/HGustavs/LenaSYS/assets/129370135/d881f213-10f2-4f89-8b17-3e398733b75e)

## Description
All elements except ER entities have their top node missing.

## Job
Implement a solution so that all elements have a top middle node.

PR #16178, opened 2024-06-11
Komplettering-b22samer-#16161

b22samer:
For most of the time, I thought the problem was that the node wasn't being drawn for all elements except the "er entity." However, after testing various changes to how the nodes are drawn, I realized that this code has no errors. Finally, I was able to see that the node was being drawn but was hidden behind the bottom node in the middle, and only a position adjustment was needed to solve the problem.

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -315,6 +315,7 @@
 .node.mu {
     right: calc(50% - 4px);
     cursor: ns-resize;
+    top: 0;
 }
 .node.tr{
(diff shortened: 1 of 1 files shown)


b22samer:
Before:
testning1

After:
testning1 1

After comparing how the nodes looked before and after, it appears that the task is correctly solved since the node that was missing before is present in the proposed solution.

a23wilgr (changes requested):
Remove cursor completely on node.mu

Issue #16179, opened 2024-06-11
Separate groups of items in sectioned.php


a22chrfa:
![image](https://github.com/HGustavs/LenaSYS/assets/128893264/61d68b51-d27b-4141-ad3e-066ceff75985)

It would be easier on the eyes to separate the groups of items instead of having them lumped together. A moment or a section for example could use some spacing. This would also make it adhere more strictly to the gestalt laws. Look into the prod version of canvas to see how moments, sections and headers are used and come up with a style that separates groupings a bit clearer.

a22chrfa:
**Canvas current structure and usage:**

- A header seems to sometime be followed by a moment, and sometimes not.
- A section seems to be a subsection of a moment. A subsection can also be very long.
- They seem to use tabs for further/alternate sectioning. This doesn't work in our version of the application though.
- Can't 100% determine which type of grouping would be the best. But since it should be relatively easy to change in the code at a later stage, it shouldn't be a huge issue if it turns out that we want to group by moment instead of section, or something like that.

**Other notes:**
Each row is actually a table. And it all seems to be created in javascript.

a22chrfa:
Just documenting an example that includes every type of item. To be used later for comparison.
![image](https://github.com/HGustavs/LenaSYS/assets/128893264/0bbeb8a7-5a9f-4bfe-8f22-f0714a4c4385)

PR #16180, opened 2024-06-11
New styling and separation of groups in Sectionlistc (#16179)

a22chrfa:
Added logic to the styling and grouping, along with necessary class and CSS. This is based on the assumptions made from the prod version of canvas. Essentially, a header seems optional and a section seems to be used within a moment.

The added logic does the following:
1. Every header div except the first one gets a top margin
2. Every row div gets a left and a right border
3. Every moment div that isn't preceded by a header gets a small top margin
4. Every row div that precedes a moment div but isn't a header gets a bottom border
5. Every row div that precedes a header row also gets a bottom border
6. The very last row div in the array gets a bottom border
7. Every row div except for the header gets a class for shorter width

This will give us a box-like interface where a header is considered a group, and a moment (that includes sections) is considered a subgroup. If however it turns out that a section should be a parent of a moment, then this could easily be flipped. Also keep in mind that this is not a beauty contest. The purpose is to view a header as a group, and moments as subgroups.

image

**For testing**
Create different items and move them around. Make sure the borders change as expected (a moment is a box that includes all rows until a new moment or a header). Keep in mind the assumptions made, structuring the list in other ways will give the result that this is coded for.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -2006,6 +2006,27 @@ function returnedSection(data) {
     var slist = document.getElementById('Sectionlisti');
     slist.innerHTML = str;
 
+
+    //Code testing
+    const sectionListDivs = document.querySelectorAll('#sectionlistc > div');
+    const divFullIdArr = [];
+    const divShortIdArr = [];
+    const regexId = /\d.*$/; //matches the first digit and everything following
+
+    for (let i = 0; i < sectionListDivs.length; i++) {
+      divFullIdArr.push(sectionListDivs[i].id);
+      let currentId = sectionListDivs[i].id;
+      let newId = currentId.replace(regexId, '');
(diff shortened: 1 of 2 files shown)


a22chrfa (commented):
The code appears to look good. I've also played around some more with the varius row types and it seems to work as intended and it doesn't appear to break in any unexpected way.

It also does look nice to separate the moments. This pull request is OK

a23elism (approved):
Styling + changes don't seem to break anything, effective use of changes would require re-styling (through admin view) to effectively use the new styling

Approved 👍

a23elism:
Changed base branch into group branch "G3-2025-V1"

a23elism (approved):
Re-reviewed code/changes for merge with group branch, no conflicts seen or problems diagnosed

Approved 👍

Issue #16181, opened 2024-06-12
The button to show/hide hidden items in sectioned.php doesn't work


a22chrfa:
![image](https://github.com/HGustavs/LenaSYS/assets/128893264/164fa0e7-c015-4b5e-b7e7-7a5d0eac182d)

It seems this button has the intention to be a toggle-function for hidden items. As of now, the hidden items are always displayed and the button doesn't actually do anything.

Explore if this button can be activated in order to toggle the hidden items

a22chrfa:
In reference to the commit that was just made, here's some useful information to anyone looking to improve or alter the code in the future. (some of this is still beyond my understanding).

The tables are generated with javascript, which uses the draggable function of jquery. This seems to make the element dynamic and therefore does not exist when the document is fully loaded. Even though we know it exists by looking in the browser.

![image](https://github.com/HGustavs/LenaSYS/assets/128893264/85624609-4432-408a-a9ef-7ccd815a48d5)
![image](https://github.com/HGustavs/LenaSYS/assets/128893264/0acd9928-6f71-498b-90ee-81c348364e12)

They do exist within the function that creates them though, so we can add logic there, which is what I did in the commit.

I've also noted that the ID's in the list are based on type and position, so an item doesn't really have a static ID. It gets a generated ID depending on its position. This might require some extra functionality later on...

PR #16183, opened 2024-06-13
Toggling hidden items now work (#16181)

a22chrfa:
The previous buttons didnt seem to work, and judging by their behaviour it was a bit hard to determine what they were actually meant to do. The existing "eye"-button did nothing, and the ghost button seems to have been intended to remove the hidden status of the items, but it only managed to remove the strikethrough in the text, not the status. I hid these buttons by setting their display to None. If their functionality is to be looked at, it'll have to be in another issue. This issue is about toggling hidden items in the list.

I created a new button for toggling, it borrows the eye-icon.
image

When clicking this button, the items in the list will now have their visibility toggled.

Toggled on:
image

Toggled off:
image

**For testing**
Try toggling. Without reloading the page, edit, add and remove items and use the toggler. It should work regardless.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -1371,6 +1371,17 @@ function duggaRowClick(rowElement) {
     }
   }
 }
+
+//
+function toggleButtonClickHandler() {
+  if (showHidden === true) {
+    showHidden = false;
+  } else {
+    showHidden = true;
+  }
+  toggleHidden();
+}
+
 var itemKinds = [];
 function returnedSection(data) {
   retdata = data;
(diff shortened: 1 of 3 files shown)


a22chrfa (commented):
The code seems fine, comments are inserted where they might be needed. It seems to function as expected, I've tried breaking it but so far been unsuccessful.

This pull request is OK

a23carpa (approved):
After testing it appears to do what it goes out to do. In Main the ghost button and eye as stated don't seem to do anything and with these changes toggeling hidden with the eye button works well.

PR #16184, opened 2024-06-14
Making a4 appear as it should

b22samer:
I used the code that draws a4 vertically and implemented it in the function that is called when the "a4 template" button is activated. By doing this, a4 is drawn correctly when the button is activated.


b22samer:
In the first image, the A4 pattern wasn't displayed correctly as only the text "A4" was visible and not the actual pattern. In the second image taken after the pull request, the A4 pattern is displayed correctly as it should be, and no new issues arose with the new solution proposal.

Before changes:
nyissue

After changes:
A4Testning

a23emisn (commented):
Verified that the A4 does show up correctly with the revised toggle.js file. The implementated solution works.

Issue #16185, opened 2024-06-14
Make large swimlanes dynamic in some form


a22chrfa:
When a swimlane has enough items, it covers more than the viewport height. Make it so that it is either dynamic or a choice for the user.
It would be nice if the entire thing could be seen at once.

![image](https://github.com/HGustavs/LenaSYS/assets/128893264/8269bae3-acf7-4dc1-ba66-5f16389991ca)
![image](https://github.com/HGustavs/LenaSYS/assets/128893264/974b86d5-4e8c-4694-ae6a-429cb6b08cd4)

a22chrfa:
After checking what is possible and how it looks, the plan will be to create a button for toggling how the swimlane should be displayed, with two or three options.

PR #16186, opened 2024-06-14
Added toggling option for large swimlanes (#16185)

a22chrfa:
If a swimlane is large enough (> 70vh) a button will appear and give the user a choice to toggle between different views of the swimlane. One is the normal view, one is a cut height with a scrollbar for the overflow, and the last one is a smaller version of the swimlane that will fit the viewport.

Apart from the eventhandler, there is a switch at the end of returnedSection. This has to be there since returnedSection will run if any draggable item gets moved. Which originally caused the view to reset for the user. I had thought of making this section and the eventhandler share the code since they are very alike, but I didn't get it to work. Maybe I'm too tired now.

**For testing**
Enter a sectioned.php where the swimlane has enough height, the demo course seems good for this. Try toggling around and move some items in the list.

Now enter a course where it doesn't have enough height. Datorgrafik has one of these. The button shouldn't be visible here.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -1377,48 +1377,46 @@ function SwimlaneToggleFunction() {
   const swimlaneDiv = document.getElementById('statisticsSwimlanes');
   const swimlaneSvgElement = document.getElementById('swimlaneSVG');
   const currentChoiceIndex = swimLaneViewOptions.indexOf(swimlaneViewChoice);
+  const toggleButton = document.getElementById('swimlaneToggleButton');
 
   let nextChoiceIndex = (currentChoiceIndex + 1);
   if (nextChoiceIndex > 2) {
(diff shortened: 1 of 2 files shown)


a22chrfa (commented):
clean up the logging, maybe add clarity to the logging in the default case. Doesn't say what function went wrong.

a22chrfa (commented):
Code looks ok. Perhaps it should have been more efficient by sharing some styling in a shared function. Maybe that could be optimized later. It seems to function well, I've tried breaking it without succeding. There is one const that have the same name in both functions, but that shouldnt be a problem since const are block scoped.

I'll give it a pass, as mentioned. It might be optimized in the future.

Pull is OK

a22maxda (changes requested):
Button is visible but does not appear to change anything, regardless of course.

PR #16187, opened 2024-06-18
Fixing snap on Sequence Actor and Sequence Object

b22samer:
I split a selection that handled the snapping for the sequence actor and the sequence object. This function calculated the position of the sequence activation incorrectly, so it worked the opposite way for the sequence actor and the sequence object. By splitting the selection into two, I was able to adjust the calculation individually for the sequence actor and the sequence object.

### DIFF
--- a/DuggaSys/diagram/helpers/mouse.js
+++ b/DuggaSys/diagram/helpers/mouse.js
@@ -203,9 +203,23 @@ function snapSAToLifeline(targetId) {
     if (lifeline) {
         for (let i = 0; i < data.length; i++) {
             const element = data[i];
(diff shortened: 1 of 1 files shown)


b22samer:
Before the pull request, the calculation was incorrect and caused the sequence activation to be placed incorrectly. After the pull request, the calculation looks correct, ensuring that the sequence activation can only be placed correctly.

Before:
https://github.com/HGustavs/LenaSYS/assets/129371343/840c98b3-e73e-4808-93a8-9d1ac2ab6a72

After:
https://github.com/HGustavs/LenaSYS/assets/129371343/fcc57414-51ed-49a6-b4ec-4ae974b26db2

PR #16189, opened 2024-06-25
Fixed Resizing with upper nodes on elements weird

a22robko:
The issue was that when dragging the nodes, they would jump while being dragged. Initially, the element would change its position based on the start and end point of a mouse click-and-drag event, causing the element to jump to where the mouse was at the start. This behavior has now been modified to depend on the delta movement of the mouse, preventing the element from jumping.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1148,13 +1148,10 @@ function mmoving(event) {
 function movementPosChange(element, start, delta, isX) {
     // mouse position is used causing the line to "jump" to the mous pos.
     // The magic numebers are used to center the node middle with the mouse pointer
-    let property = (isX) ? 'x' : 'y';
-    let x = (isX) ? start - delta - 6 * zoomfact : 0;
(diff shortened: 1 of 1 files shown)


a22robko (commented):
I have tested this issues entities, and they no longer jump when dragged, as they did initially. I have tested all four and checked to ensure that nothing else has been broken in diagram.php.

Before:
You can see how it jumps when you drag it.

https://github.com/HGustavs/LenaSYS/assets/129367092/19548349-25cd-4ffe-b4f6-9dc1bc3e51da

After:
Here you can see that when dragged, it no longer jumps.

https://github.com/HGustavs/LenaSYS/assets/129367092/29e2469c-41c2-4d6b-87b9-c01ddea86ff6

a23elism (approved):
Resize/dragging issue fixed for specific node, node no longer jumps/glitches when resizing.

Approved 👍

a23elism:
Changed base branch to group branch "G3-2025-V1"

Issue #16191, opened 2024-06-25
The example diagrams doesn't load correctly


a22robko:
Description: The elements and lines in the UML and State Chart Diagrams are not loading correctly. 

Job: Fix the loading of example diagrams.

<img width="993" alt="Screenshot 2024-06-25 at 21 11 24" src="https://github.com/HGustavs/LenaSYS/assets/129367092/6ecbc8e2-42b1-4d61-9c0e-336b856a7b9f">

PR #16192, opened 2024-06-27
Fixed The example diagrams doesn't load correctly #16191

a22robko:
The problem is when the element 'height' and 'fill' are not set, the value will be 'undefined' when building the element. The solution is to set 'height' and 'fill' to a default value ('0' and '#ffffff') when they are not provided, so the element can render correctly.

### DIFF
--- a/DuggaSys/diagram/draw/element.js
+++ b/DuggaSys/diagram/draw/element.js
@@ -9,7 +9,8 @@ function drawElement(element, ghosted = false) {
     let texth = Math.round(zoomfact * textheight);
     let linew = Math.round(strokewidth * zoomfact);
     let boxw = Math.round(element.width * zoomfact);
-    let boxh = Math.round(element.height * zoomfact); // Only used for extra whitespace from resize
+    let boxh = element.height ? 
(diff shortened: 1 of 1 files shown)


a22robko (commented):
I have tested this issue and it appears that both the UML and state chart diagrams are working and clear. Previously, the diagrams were not fully visible, but now they can be seen completely.

**UML**:

Before:

https://github.com/HGustavs/LenaSYS/assets/129367092/6a0a69b4-e1f8-4c94-b39a-712de0b4be50

After:

https://github.com/HGustavs/LenaSYS/assets/129367092/4fdaf7e0-062d-4a50-852f-23614aa34798

**State chart:**

Before:

https://github.com/HGustavs/LenaSYS/assets/129367092/4c25228a-602b-4b40-8bc8-6e2d1756675d

 After:

https://github.com/HGustavs/LenaSYS/assets/129367092/01efcbed-9dd9-440c-bf0a-47967e1d5596

Now everything works as it should, and both the UML and state chart diagrams are clear and fully visible. And I have checked to ensure that nothing else has been broken in diagram.php.

PR #16193, opened 2024-06-30
Adjusting the positioning between the IE entity and IE inheritance

b22samer:
I found an issue with IE inheritance where the line from the IE entity was not connected to the actual IE inheritance, as it was positioned a few pixels below the IE inheritance. The solution involved adjusting the padding and margin for the line so that it was correctly positioned on the IE inheritance and afterward it looked much better.


b22samer:
After testing and comparing the placement of the line before and after the pull request, it appears correct. The line was not connected with IE inheritance before the pull request, and after the pull request the line is connected with IE inheritance.

Before:
ny

After:
ny1

c22isako (changes requested):
It works. But, could be done in a better way. As it looks, it does affect the lines on top of the IE inheritance. So by doing the fix, it might break other functionality.

PR #16202, opened 2024-07-09
15974 add documentation for microservices in highscoreservice

c20juler:
Documentation is created for both service files that constitutes the highscoreservice. It is located in the same folder as the microservice highscoreservice. Test to see if markdown syntax behaves as expected and check if documentation looks fine.


c20juler:
Documentation already exists for these microservices in "Microservices.md". Will therefore close this PR.

PR #16204, opened 2024-07-10
Update Microservices.md

c20juler:
createNewVersionOfCourse_ms.php in the "shared microservices" section has been deleted from the microservice documentation. I assume line 452-471 should still be included regarding createNewVersionOfCourse as it is not specified in the issue description that this section should be deleted.


c20juler:
Changes are done according to description.

a23edvto (approved):
I think it should be removed as it doesn't seem like the file exists

PR #16205, opened 2024-07-11
Fixed courseed and sectioned motd button

d22isaka:
Buttons fixed, they now appear the same


a23lisjo (approved):
The created MOTD button is working, a variable that changes the h3 title for MOTD banner does not work but it's not causing any problems.

PR #16218, opened 2024-07-18
Fixed common typo in Microservices.md

c20juler:
fixed common spelling error in 71 places


c20juler:
fixed, md syntax is preserved after change.

a23petny:
I think this is group 2 concern.

a23bensh (commented):
I'd say these 71 changes seem pretty unnecessary, as they don’t contribute to functionality and offer minimal improvement to readability. While "queries" is the correct spelling, updating it across so many instances feels excessive and would likely just clutter the commit history. I'd recommend probably closing it.

PR #16220, opened 2024-07-18
new test refreshGithubRepo

b21youab:
Had problems with headers but test should work as intended.


a20artje (changes requested):
What is this supposed to fix? Very unclear description.

Issue #15534, opened 2024-05-08
Fix alignment of navbar items in accessed.php


a22moamy:
Navbar items in accessed.php need vertical alignment. Currently, some items are aligned to the top of the navbar, and others are aligned to the bottom. Please make their vertical alignment more centered. 

How to reach accessed.php
1. Go into a course
2. Press the "Give students access to the selected version"-button (the lock-button):
![Skärmavbild 2024-05-08 kl  14 05 44](https://github.com/HGustavs/LenaSYS/assets/129263915/27369fe2-2124-44f9-87fa-e7879a01aa30)

The current alignment (faulty icons are highlighted):
![Skärmavbild 2024-05-08 kl  14 05 55](https://github.com/HGustavs/LenaSYS/assets/129263915/bc56190f-a05c-4cce-95aa-2f66ab9bc39f)

The current alignment followed by a red curve:
![Skärmavbild 2024-05-08 kl  14 05 59](https://github.com/HGustavs/LenaSYS/assets/129263915/897b04d9-741b-4ca9-b587-dbe115367948)

a22samka:
![navbutt](https://github.com/user-attachments/assets/6afe1a6e-8f00-4907-9b49-057899e1ddb4)

It seems like the filter icon has a class called "navButt", the other icons dont. The navButt class has a margin top that makes it so the icon is moved down slightly.

a22samka:
I cannot remove this since it is connected to alot of other buttons on other pages

a22samka:
Searchbar seems to also be affected by the navButt class and therefor gets moved out of place. This class has now been removed from the td testSearchContainer which seem to have fixed the issue

PR #16225, opened 2024-07-23
Fix alignment of navbar items in accessed.php #15534

a22samka:
Check if this solution fixed the alignment issue in the accessed.php file


a22samka:
After testing it seems that this fixed the issue without any problems, the icons do now have the correct alignment.
image

a22maxda (approved):
Looks good, doesn't appear to break anything. Approved

PR #16227, opened 2024-07-23
Changing the color of the objects that were filled with gray color instead of black

b22samer:
By locating where the objects 'sequence lifeline', 'sequence lifeline (object)' and 'sequence condition' are drawn, I was able to determine where the coloring was added. Then I changed the color of these three from gray to black.


b22samer:
The top image was taken before the pull request was made and at that time some elements were filled with gray color, but after the pull request was executed with the proposed solution the elements are filled with black color as they should be.

Before changes:
ny

After changes:
twa

f23linan:
Looks good! Ready to merge :)

Issue #16226, opened 2024-07-23
Error for "cannot read properties of null" appears when entering accessed.php 


a22samka:
This error appears when entering accessed.php, find what is causing it and fix it so that the error is removed.
![image](https://github.com/user-attachments/assets/81bbe628-f4d3-4e89-a7e0-e1a8cb4c74de)

a22samka:
I think the error is caused by the JavaScript running before it checks if the elements exist in the DOM, the dom might not be fully loaded before the js is running.

PR #16231, opened 2024-07-24
Fixed error that appears when entering accessed.php - #16226

a22samka:
Fixed error by making an eventlistener and a if( ) to check if the elements exist before the js is run.


a23alera (approved):

EdvinWerner (approved):
No errors appear when entering the accessed.php file anymore therefore the problem originaly described no longer persists

PR #16232, opened 2024-07-24
fixed SD Entity Stroke Not Visible in Dark Mode

a22robko:
The box for the SD Entity wasn't showing its top border correctly because the top part didn’t have the class="text" attribute, which helps control how the box looks. The bottom part had this attribute, so it looked fine. To fix this, I added class="text" to the top part of the box in the code too. This made the stroke show up correctly around the entire box.

### DIFF
--- a/DuggaSys/diagram/draw/element.js
+++ b/DuggaSys/diagram/draw/element.js
@@ -313,6 +313,7 @@ function drawElementSDEntity(element, boxw, boxh, linew, texth) {
     let height = texth * 2;
     let headPath = `
         <path 
+        class="text"
             d="M ${linew + cornerRadius},${linew}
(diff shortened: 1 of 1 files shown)


a22robko (commented):
I have tested the issue and I have checked the different colors as well as the other entities and nothing has broken. I have recorded two videos where you can see the before and after of how they look.

Before:

https://github.com/user-attachments/assets/60ff81ef-c871-4cc5-9eac-2b6eefcc8d27

After:

https://github.com/user-attachments/assets/6fd5e2e8-412b-4975-95e7-3c2bdd34e232

Before, you couldn't see the color on the upper part, but now it is possible and  improves ease of use.

a23kavis:
While this works, the problem can be fixed by making sure toggleBorderOfElements() is handling borders of this entity properly.

PR #16233, opened 2024-07-24
updated specialUpdate_ms_test.php

c22mohel:
now it compare the correct values.


c22mohel:
this will need a fresh database in order to work correctly, but now it not only compares the debug values but all of entries.

c22mohel:
To test this issues i opened the specialUpdate_ms_test.php file that is located in 
http://localhost/LenaSYS/DuggaSys/tests/microservices/courseedService/specialUpdate_ms_test.php.
It now works as it is supposed to. Before only the debug was being compared but now all of entries is being compared. 
The test will need a fresh database in order to work properly.

a23eliax (commented):
Checked the code and I verify that it compares the right values

PR #16236, opened 2024-07-25
fixed UML Inheritance Diagram Disappearance

a22robko:
Changed stroke colors in drawElementUMLRelation. I added logic for the stroke color in the new code. Now, if the fill is black (fill === 'black'), the stroke is white ('white'), and if the fill is white, the stroke is black ('black'). This adjustment makes the overlapping element clearer and easier to see with black background in the diagram.php.

### DIFF
--- a/DuggaSys/diagram/draw/element.js
+++ b/DuggaSys/diagram/draw/element.js
@@ -434,10 +434,11 @@ function drawElementERAttr(element, textWidth, boxw, boxh, linew, texth) {
 
 function drawElementUMLRelation(element, boxw, boxh, linew) {
     let fill = (element.state == 'overlapping') ? 'black' : 'white';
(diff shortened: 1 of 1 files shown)


a22robko (commented):
I have tested the issue and everything looks and works as it should, and nothing has broken. Now when you press overlap, the triangle's strokes are outlined with white lines and it is now visible in dark mode, something it wasn't before, as you can see in the before video.

Before:

https://github.com/user-attachments/assets/939b98af-f959-4fbb-b6e1-59cdef28d42e

After:

https://github.com/user-attachments/assets/1347ab91-12ee-4efe-9a80-fb3bf47eb6bc

a23carpa (approved):
Seams to be working as intended and do not conflict with the Main branch
Approved!

PR #16237, opened 2024-07-25
Making it possible to resize IERelation

b22samer:
To make it possible to resize IE Relation the same way as other objects, I tried using the existing nodes. I quickly noticed that there were a lot of problems when I tried to reuse them. The nodes that worked normally were the edge nodes but the bottom edge nodes were positioned incorrectly and I found no way to only change the position for these on IE Relation. Therefore I created two new bottom nodes and positioned them correct. There are small errors that I can't fix, for example, what is shown in the image with the selection area. Apart from the small errors, it works as it should.

image

### DIFF
--- a/DuggaSys/ace.js
+++ b/DuggaSys/ace.js
@@ -16374,8 +16374,8 @@
             return (row - layerConfig.firstRowScreen) * layerConfig.lineHeight;
         };
     
-        function getBorderClass(tl, tr, br, bl) {
-            return (tl ? 1 : 0) | (tr ? 2 : 0) | (br ? 4 : 0) | (bl ? 8 : 0);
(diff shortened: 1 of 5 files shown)


b22samer:
Apart from the issues mentioned in the pull request, the solution looks okay. Before the pull request it was not possible to resize IERelation but after the pull request resizing is possible.

Before:
image

After:
image

EdvinWerner (changes requested):
The ER relation can now be resized so that part works, there are however a number of follow up issues, one can now make the borders of the relation way to big compared to it's actual size. It is also possible to make white outliner of the relation too small compared to X in the middle. Functionality for not extending the relation downwards and better border control must be added.

These issues are shown in the attached screenshot.
Skärmbild 2025-04-07 114344

PR #16238, opened 2024-07-25
fixing deleteFileLink_ms_test.php

c22mohel:
Removing the debug filter to be able to compare the full output


c22mohel:
This does now fix deleteFileLink_ms_test.php, you will need to have a fresh database for this test to work but now it not only compares the debug values but all of entries just like it is suppose to do.

c22mohel:
Uploaddate is now removed from the test due to it changing every time the database is installed. This made it fail the test. But now uploaddate have been removed and it should not matter when the database was installed.

Issue #15334, opened 2024-05-03
broken arrow with incorrect start and end points


a22vioja:
# CAN BE DONE AFTER ISSUE #15328 MERGE (VECKA 5)

## Description
It is possible to make a line from Sequence Lifeline to Sequence object, but the arrow is broken and does not go from line to line but hangs in the middle.

<img width="448" alt="Screenshot 2024-05-03 at 14 07 49" src="https://github.com/HGustavs/LenaSYS/assets/129263559/5f55d1fc-5ae0-4853-b1b7-b71a0db37ff7">

## Job
Make the arrow straight;
Adjust the start and end points for the arrow.

PR #16239, opened 2024-07-28
G1 2024 komplettering #15334

c22abdha:
The arrow is now straight, with the start and end points adjusted.


c22abdha:
I have tested the implementation, and it works correctly. The arrow is now straight, with the start and end points adjusted as required. The attached image confirms the successful adjustment.

PR #16241, opened 2024-07-31
Found out, why "Toggle Error Check" gives error on ER elements and fi…

a21sahfa:
..xed it.

the code has an issue which is causing the whole FOR statement to not work.
 
On line 18 and 19 the word (lines) is indicating error. instead of writing (lines) . i wrote (line) on line 18 and 19, this activated the FOR statement and things worked right.

### DIFF
--- a/DuggaSys/diagram/errorHandling.js
+++ b/DuggaSys/diagram/errorHandling.js
@@ -13,21 +13,19 @@ function checkEREntityErrors(element) {
             errorData.push(element);
         }
     }
-
-    // Checks for entity connected to another entity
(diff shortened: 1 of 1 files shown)


a21sahfa:
https://github.com/user-attachments/assets/72b3d230-d544-41e2-b2dd-36d42287d0bc

this is how to test it

a21sahfa:
for (let i = 0; i < lines.length; i++) {
        line = lines[i];
on these two lines i removed the s from lines
    for (let i = 0; i < line.length; i++) {
        line = line[i];

Issue #16199, opened 2024-07-08
Dropdown is not responding, hover disappears


a22diako:
When the mouse is over the plus icon the hover diseppears on the way to the fab-buttons.

GOAL: Make it possible for the mouse to have a smooth transition from icon to fab buttons while hover is activated.

Also add function (a spin) for the small plus icon (addElement) for it to show that it can be interacted with through the mouse!

![Skärmbild 2024-07-08 123211](https://github.com/HGustavs/LenaSYS/assets/129268487/07f8c365-e830-4797-81e6-33b3f948a281)

a22diako:
EDIT: When the mouse is over the plus icon the hover diseppears on the way to the fab-buttons.

GOAL: Make it possible for the mouse to have a smooth transition from icon to fab buttons while hover is activated.

Also add function (a spin) for the small plus icon (addElement) for it to show that it can be interacted with through the mouse!

a23fanro:
This issue has been fixed as of [PR #16242](https://github.com/HGustavs/LenaSYS/pull/16242).
Still works, +-button spins, and there is a smooth transition from icon to fab when hoovered over.

![Image](https://github.com/user-attachments/assets/96fad969-87a1-42c0-8bf6-fc7046a98073)

PR #16242, opened 2024-08-02
Komplettering a22diako #16199

a22diako:
The issue was to fix so the addElement and fabBtns hovers would work fine and be clickable which was done with css for the hover to be moved so the mouse still can interact with the new buttons without them diseppearing because of it being too far away. 

The functional issue within the same elements was to show that the addElement button (small plus) is clickable through it doing a spin when the mouse goes past it. As it is the 'head-element' on the page that shows the other buttons the user can interact with. This was made through changes in the if or else.

### DIFF
--- a/DuggaSys/diagram/classes/stateChangeFactory.js
+++ b/DuggaSys/diagram/classes/stateChangeFactory.js
@@ -1,0 +1,201 @@
+/**
+ * @description Constructs state changes with appropriate values set for each situation.
+ * This factory will also map passed argument into correct properties in the valuesPassed object.
+ */
+class StateChangeFactory {
+    /**
+     * @param {Object} element The new element that has been created.
+     * @returns {StateChange} A new instance of the StateChange class.
+     */
+    static ElementCreated(element) {
+        const values = {kind: element.kind};
+
+        // Get the keys of the values that is unique from default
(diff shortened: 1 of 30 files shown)


a22diako:
Testing works fine. Short and clean code. 

The hover and button is well placed and its possible to click the fab buttons inside of it, which was the issue before. 
Screenshots of the relevent icon and fab button placements:

Skärmbild 2024-07-08 123211


Skärmbild 2024-08-02 214020


The addElement button also does a full spin when in contact with the mouse, obs. the spin continues even if the mouse is not over the button which is positive as the button stands out and show that its important and clickable.

a23fanro:
This PR has fixed [issue #16199](https://github.com/HGustavs/LenaSYS/issues/16199).
Still works, +-button spins, and there is a smooth transition from icon to fab when hoovered over.
431808641-96fad969-87a1-42c0-8bf6-fc7046a98073

Because of that I am deleting this branch.

a23fanro (commented):
This PR has fixed https://github.com/HGustavs/LenaSYS/issues/16199.
Still works, +-button spins, and there is a smooth transition from icon to fab when hoovered over.
431808641-96fad969-87a1-42c0-8bf6-fc7046a98073

Because of that I am deleting this branch.

Issue #16240, opened 2024-07-31
The code includes variables directly in the HTML output, which could lead to cross-site scripting (XSS) vulnerabilities if not properly sanitized.


c22abdha:
Security Concerns with Unescaped Output:

$cid=getOPG('courseid');
$vers=getOPG('coursevers');
$quizid=getOPG('did');

PR #16244, opened 2024-08-03
G1 2024 komplettering #16240

c22abdha:
The code includes variables directly in the HTML output, which could lead to cross-site scripting (XSS) vulnerabilities if not properly sanitized.


a23matsh (commented):
Try and catch block prevents error handling but does not adress the XSS concern in the issue, needs further implementation since it does not solve the issue

PR #16246, opened 2024-08-03
Adding try-catch blocks to critical sections of code to gracefully ha…

c22abdha:
…ndle any unexpected errors.


a22nevan (approved):
Checked so that added try/catch blocks work, which they do. If an error occurs, it is displayed. 

Some sections still give errors (_like saving a diagram, which gave me errors when trying both the issue- and main branch_), but the added code works and doesn't seem to affect anything else.

Changes can be merged, but I wouldn't consider the issue to be fully resolved.

PR #16249, opened 2024-08-04
Making the elements stop jumping when SnapToGrid is activated

b22samer:
While solving this problem there were two issues. The first issue was that the calculation for vertical movements when snap to grid was enabled was incorrect. I fixed this by removing the part of the calculation that caused the element to jump up a few pixels. The second issue was that only ER entities were considered in the selection, which made only the ER entity function correctly. To address this I created a list of all the different elements that the selection then went through. This way all elements are taken into the calculation.

### DIFF
--- a/DuggaSys/diagram/theme.js
+++ b/DuggaSys/diagram/theme.js
@@ -11,10 +11,9 @@ function updateCSSForAllElements() {
             left -= deltaX;
             top -= deltaY;
         }
-        const includedElementTypes = [elementTypesNames.EREntity, elementTypesNames.IEEntity, elementTypesNames.UMLEntity, elementTypesNames.SDEntity, elementTypesNames.UMLRelation, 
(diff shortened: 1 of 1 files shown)


b22samer:
I came up with a better solution where all the elements don't need to be added to the list manually. Instead, I'm using 'Object.values' to fetch all the different types of elements that are available. I made a new commit that includes the new solution.

b22samer:
After the pull request has been completed, the elements no longer jump when snap to grid is enabled. Snap to grid works as it should and the same goes for the elements.

Before:
[2024-07-21 13_46_13 - Screen - NYISSUE1.webm](https://github.com/user-attachments/assets/3f3dd0b1-4e3b-4cb8-a569-1e451c6d3c25)

After: 
https://github.com/user-attachments/assets/16f0700e-16f0-4d93-bc92-b78d2c218e27

a23miksk (changes requested):
The code changes did not cause any noticeable changes in the diagram program, the grid snapping behaved in the same way after the changes as before.

PR #16250, opened 2024-08-05
Update selection.js

a21sahfa:
removed the code from line 36-39. duggaSys>diagram>draw>selection.js

### DIFF
--- a/DuggaSys/diagram/draw/selection.js
+++ b/DuggaSys/diagram/draw/selection.js
@@ -33,10 +33,7 @@ function drawSelectionBox() {
                     }
                 }
             }
-            let lineSelectionBox = Rect.FromPoints(
-                new Point(minProperty(tempLines, 'left') - margin, minProperty(tempLines, 'top') - margin),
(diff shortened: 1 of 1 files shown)


a21sahfa:
https://github.com/user-attachments/assets/0e5372dc-e484-40f4-aee5-cee7b70d3238

this is how you test it

a21sahfa:
In selection.js the codes are for the square /rectangular shape on elements and line after they are selected. The shape looks perfect for all the elements but for the line it looks weird. On top of the code you can see margin =5. If we increase the margin then the square /rectangular shaped box becomes bigger (for all shapes and line) if we think we realize that any shape around the line will look weird so i deleted the code that was responsible for the shape and now it only shows the line being highlighted

a21sahfa:
Even if we add a shape around the line even then it wont look the same for all lines. Like a vertical line will have a different box shaped and a horizotal line will have another weird shape and a line that is tilted will have different so best option was to just remove the box around the line as it looks better when line is just highlighted

PR #16252, opened 2024-08-10
fixed this issue

a21sahfa:
In the code below the value of X has been changed to 1100 that is slightly changing the position of the diagram in the row


a21sahfa:
this will move the diagram with the heading Subject to few pixels right. then there wont be any flicking or movement in the line after zooming in or out or by moving the diagram
The reason it was having issues before was because of the location of the diagram. By changing the direction to any other place simply fixes the problem.

a21sahfa:
https://github.com/user-attachments/assets/8cfc2d0a-ec3f-4776-b039-8856ca25306f

this is how to test it

a20artje (commented):
Already fixed on master, new code doesn't seem to affect any other things. Could be merged but kind of pointless.

PR #16254, opened 2024-08-13
Fixed highlight in fileed.php

d22isaka:
Fixed the higlighting so the full row is highlighted instead of a single cell.

Before 
Skärmbild 2024-08-13 143319

After
Skärmbild 2024-08-13 143243


d22isaka:
Looks ok. QoL would be fixing the borders. Do it if there's time.

a22carsv (approved):
Now highlights the entire row of a cell when pressing a cell instead of just highlighting the cell.

f23linan (approved):
Works! Approved and changed to correct branch

PR #16258, opened 2024-08-16
fixed this issue

a21sahfa:
1) The code has Switch statement with a case of line DIrection. UP
2) The value 1 is added to make sure the line isnt too big or small when released and connected
to an element
3) The value Px added is basically equal to 3 and it is used to make sure the starting point of line
is not too big or small.
4) In these line i replaced y2 to y1 and y1 with y2 in the result , this causes the line to move forward instead of backward when dragged and released on an element,

### DIFF
--- a/DuggaSys/diagram/draw/line.js
+++ b/DuggaSys/diagram/draw/line.js
@@ -279,10 +279,11 @@ function getLineAttrubutes(f, t, ctype) {
     };
     switch (ctype) {
         case lineDirection.UP:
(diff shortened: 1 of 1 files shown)


a21sahfa:
Uploading Skärminspelning 2024-08-16 kl. 18.06.10.mov…

this is how to test it

a21sahfa:
https://github.com/user-attachments/assets/cf2753d7-289d-4bd5-bd2a-4ea55bb04445

Issue #16256, opened 2024-08-15
fix the issue where the inheritance line doesnt touch the inheritance, make it so it touches the inheritance


a21sahfa:
![Skärmavbild 2024-08-15 kl  21 48 45](https://github.com/user-attachments/assets/76e8bd0f-d6a2-41cb-85f1-528c9df7c2cc)

PR #16262, opened 2024-08-19
fixed this issue

a21sahfa:
in issue #16256 i worked on the line direction up which means changes will be in line drawn upward. So similarly when i spotted the issue where drawing line from inheritance to entity was having issue since that line was drawn downward so it had to be fixed in Line Direction. Downward. Therefore in the i changed fy2 to fy1 which solved some of the issue and i changed -px *2 to 1 which basically means value was -6 before and now its 1 . So the line that was way to bigger at a certain angle was fixed when 1 was written.

### DIFF
--- a/DuggaSys/diagram/draw/line.js
+++ b/DuggaSys/diagram/draw/line.js
@@ -279,14 +279,14 @@ function getLineAttrubutes(f, t, ctype) {
     };
     switch (ctype) {
         case lineDirection.UP:
-            offset.y1 = px;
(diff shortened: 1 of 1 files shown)


a21sahfa:
Offset.y2= px is for the part at the end of the downward line

a21sahfa:
https://github.com/user-attachments/assets/65cc69c0-342b-4253-95af-e0355c85f0ae

this is how to test the issue

PR #16263, opened 2024-08-19
Removed line of code preventing property changes for IE-relations.

b22chrmo:
In the stateChange.js file there was a line of code preventing IE-elements from saving the newly chosen property. This must be old code from when the IE-element had a dropdown that could break the site. This line of code is no longer necessary and only prevents the IE-relation from changing inheritance type.

IE-relations can now properly switch between disjoint and overlapping types.


b22chrmo:
Solution works as intended. Checked if other elements were affected or if anything was broken, but no new bugs can be found. 


https://github.com/user-attachments/assets/56b5cd09-949f-430b-a182-06a2f3ad58e0

PR #16264, opened 2024-08-19
Implementing a solution for the issues related to the "note" and "UML super state" elements

b22samer:
I am modifying the maximum number of characters that can be written per line in the "note" element and doing the same for the "UML super state" element. Additionally, I am adjusting the spacing between the rectangle and the text because the text was overlapping the right side of the rectangle in the "UML super state" element.

### DIFF
--- a/DuggaSys/diagram/draw/element.js
+++ b/DuggaSys/diagram/draw/element.js
@@ -477,13 +477,21 @@ function drawElementState(element, vectorGraphic) {
 }
 
 function drawElementSuperState(element, textWidth, boxw, boxh, linew) {
+    const MAX_TEXT_LENGTH = 30;
     element.stroke = (isDarkTheme()) ? color.WHITE : color.BLACK;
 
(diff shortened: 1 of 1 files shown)


b22samer:
After the pull request it looks significantly better than it did before, the "note" element looks good. However, the "UML super state" element doesn't look exactly as expected because three dots ("...") are added at the end when the maximum number of characters is reached. otherwise, this solution still looks okay.

https://github.com/user-attachments/assets/dfb2e0b1-04a7-4009-b414-149e7e3318f0

b23ebbst (commented):
The note element mostly looks as expected, and the size adjusts to the amount of text. The UML Super State does not show the three dots "..." as mentioned in the pull request. See image. Both elements hve the remaining issue of the text not fitting (or being abbreviated) to fit within their respective containers. 
Proposed fix: Adjust so the text fits within the elements no matter the length of text, or abbreviate the text with elipses ("...") to fit without getting cut off. 
Skärmbild 2025-04-07 113400

PR #16266, opened 2024-08-19
fixing setVisibleListentries_ms_test.php

c22mohel:
changing the expected-output to the correct values in order for the test to pass. Also removing the filters to compare all the values.


c22mohel:
This will require a fresh database install to work. But this does now compare the correct values in order to pass the last test.

a20artje (commented):
Seems to be already done by updateVisibleListentries_ms.
Marked as finished in documentation and that setVisibleListentries is the old file.

[](https://github.com/HGustavs/LenaSYS/blob/master/DuggaSys/microservices/Microservices.md)

image

Issue #16257, opened 2024-08-16
New installer gets stuck


a22karja:
When using the new installer it get stuck on one of the first things it does
![Skärmklipp 6](https://github.com/user-attachments/assets/21b0c523-df0e-43c6-a0b5-14e748528b5b)

Im unsure why this happens but think it can have something to do with the order in which the SQL commands are processed

a22karja:
Update:
Installation gets to 18% when i run flush privileges before trying to create user

a22karja:
2 Fault identified.
Wrong ' symbol was used in db_setup.php set_permissions() (` was used instead of ')
New error message appear. 
![image](https://github.com/user-attachments/assets/0d4d3040-6c33-46f6-949b-77eeb59b594e)

For this reason i will start checking the create user function in db_setup.php

Issue #16153, opened 2024-05-22
Add download installation logs feature


a22willi:
The installer_ui.js file receives an SSE stream from the installer backend (install_engine.php). Save all the logs in a variable. Convert this to a file when the SSE stream is closed, and allow the user to download the logs.

a22karja:
Current problem.
When i try to use the new installer it fails on the first step
![Skärmklipp 1](https://github.com/user-attachments/assets/00f12985-1521-425e-9f3b-92a110e99e7e)
![Skärmklipp 2](https://github.com/user-attachments/assets/05e26ee2-3234-476c-b861-2d1198cb18ca)
![Skärmklipp 3](https://github.com/user-attachments/assets/f4b9335b-b088-4fc7-bdff-aa1e2b50fcf5)
![Skärmklipp 4](https://github.com/user-attachments/assets/220a01ac-19ee-40e3-be61-0c86af691470)
![Skärmklipp 5](https://github.com/user-attachments/assets/43ea3c85-2851-459c-9cf6-5f6a3cef4af6)
![Skärmklipp 6](https://github.com/user-attachments/assets/40986e34-b426-4797-b4fa-460f9d51229b)

a22willi:
@a22karja Try enabling the option: "Overwrite existing user".
And if that does not work, try with a different database and username. I suspect it cannot create it because it already exists.

a22karja:
Tried that, This came up
![image](https://github.com/user-attachments/assets/e39d2f0c-eba7-4cb7-856a-1af6b8369daf)

PR #16267, opened 2024-08-21
A22karja 2024 extra work #16257

a22karja:
This pull request will fix 2 issues at once because i had to do a install log (#16153) to figure out why the install wouldnt finish no matter what i put in (#16257)


a22karja:
Have tried breaking the installer and it seems to work. Log also notes down what the installer does

a22karja:
Thanks for approving my work. if you guys have any questions feel free to ask me

Issue #16251, opened 2024-08-09
New quote button


a22diako:
EDIT: When you click on the 'message' button a new row pops up with the text 'New quote', 
Fix the icon so it matches with the other message icon and the same thing appears when clicking ´. 

ALSO. Make a button that will showcasee quotes through a alert.

a23erigu:
This issue is not a feature that seems worth to add

PR #16268, opened 2024-08-21
Komplettering a22diako #16251 version2

a22diako:
The 'New message' button had a missleading printing as it was 'New Quote' in a new row. 

The outprint has been changed to match the other 'New Message' icon. 
Skärmbild 2024-08-21 130505

The issue below inspired a new issue, to make a new 'Show quote' button that will randomized display a motivational quote everytime the button is being activated. 
Skärmbild 2024-08-20 150124

Skärmbild 2024-08-21 130747


a23danme (commented):
The quotes works as intended (but only when logged in as admin). Not a useful functionality for the project. The code can be removed.

PR #16275, opened 2024-08-21
added the corner node code to the start/end if statement

d22hanol:
When creating the nodes to all elements there was a if statement that made sure the start/end elements did not get nodes. when the new corner nodes were added, the creation of the corner nodes was placed outside of this if statement. this is now fixed by adding the corner creation to the inside of this if statement.


f23linan:
Changed the branch its commiting into from master to the group branch

PR #16276, opened 2024-08-21
Added class to upper part of SD entity

b22chrmo:
The upper part of the SD entity did not have the class that gives each entity some of their properties, so the color of the line did not change properly when switching to dark mode.

The upper line of the SD entity now works the same as the other elements.

https://github.com/user-attachments/assets/fd23e515-db98-4af6-85bb-d04e8f3722e8

### DIFF
--- a/DuggaSys/diagram/draw/element.js
+++ b/DuggaSys/diagram/draw/element.js
@@ -313,6 +313,7 @@ function drawElementSDEntity(element, boxw, boxh, linew, texth) {
     let height = texth * 2;
     let headPath = `
         <path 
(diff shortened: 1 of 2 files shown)


b22chrmo:
Solution works. SD entities that are black are now more clear in darkmode. Upper part of SD entity now belongs to the class "text", similar to the lower part. No new bugs can be found.

b23sebth (approved):
The State Diagram entity lines now changes color correctly when switching to darkmode.

a22maxda (approved):
Changed base to group branch

PR #16277, opened 2024-08-21
fixed the copyCourseVersion_ms_test.php

c22mohel:
removes created rows from the database


c22mohel:
This removes unnecessary redundancy from the database when doing this test. This then allows for other test to be made without reinstalling the database.

PR #16279, opened 2024-08-23
fixed the issue

a21sahfa:
added a speed factor variable to control the speed of the elements in all four directions meaning right, left, up and down

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -664,85 +664,89 @@ document.addEventListener('keydown', function (e) {
     }
 
     // Moving object with arrows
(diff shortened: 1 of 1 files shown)


a21sahfa:
https://github.com/user-attachments/assets/182e56f9-a081-4a62-b072-63573e589a21

this is how you test it, you can change the speed factor variable to make it as fast as you want on line 668 
  const speedFactor = 10; // Change this to any value greater than 1 to increase speed

a21sahfa:
if you use a negative speed factor lets say -10 the element will move at the same speed as 10 but in the opposite direction.

PR #16281, opened 2024-09-19
found the issue and added a fix

d22hanol:
The problem was that the normal line that usually connects two object would be drawn behind the elements. This was also not just a problem for the SD entities but for all the entities. By adding a clause that checks if the line type is not a recursive this was then fixed


d22hanol:
this did not fix issue for the other entities, fixing the issue again and requesting a new pull

PR #16282, opened 2024-09-19
recursive arrow issue fix

d22hanol:
The problem was that the normal line that usually connects two object would be drawn behind the elements. This was also not just a problem for the SD entities but for all the entities. By adding a clause that checks if the line type is not a recursive this was then fixed


d22hanol:
The previous issue connected to the recursive arrows has now been fixed, nothing else has been destroyed in the process.

PR #16284, opened 2024-09-19
kompletering-d22hanol-15229

d22hanol:
When converting the wordlistid and boxcontent columns in box table to be added dynamically  I realized that the kind column in filelink also needed to be dynamically inserted. This is why i also fixed the filelink insertion in this issue.


d22hanol:
I can't find anything that broke when this implementation was used to fix the issue.

PR #16285, opened 2024-09-20
Kompletering-d22hanol-16283

d22hanol:
arrows were not visible as there was another layer put on top of them. Additionally they were too small and pointed in the wrong direction, this is now fixed.


d22hanol:
Arrows are now visible and I cannot find anything that became bugged because of this

PR #16322, opened 2025-04-08
Moving code from Pullrequest #16181 

a23carpa:
Copying code from pull request  #16183 for issue #16181 to new branch to be merged later.

There issue was to fix the Eye button so you could hide the viability of crossed over tests.
And from what I've tested it does its job.
image

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -29,6 +29,7 @@ var numberOfItems;
 var backgroundColorTheme;
 var isLoggedIn = false;
 var inputColorTheme;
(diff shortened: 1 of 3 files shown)


a23aleva (commented):
The eye symbol seems to be working as intended as it hides crossed over tests when clicked and makes them reappear when clicked again. This is something that doesn't work in the master branch. However, there is a "ghost" button (with unknown purpose) in master that seems to disappear when this new code is applied. This might not be the intention.

_Original master_
image

_New version_
image

_New version after hide button click_
image

PR #16324, opened 2025-04-08
Implement debouncing for event listeners to improve performance.

a23gabst1:
I took an older PR (#16248) and fit it with the updated version of 2025. This update  impoves performance by minimzing unnecessary and redundant executions.


a23kevre (approved):
This made performance better even if it just was a little bit.

PR #16325, opened 2025-04-08
Solution for text not fitting in the "note" and "UML super state" elements 

b23ebbst:
Text in the UML super state and note elements are fixed to fit within their containers

Taken from the old pull request #16264, and fit with the updated version for 2025. The text in "note" and "UML super state" are fixed to be able to handle longer text . The note element adjusts to fit the content without overlapping the borders, and the UML super state ends with an elipsis ("...").

image


a22abdjo (approved):
Works as accepted and can be merged.

PR #16333, opened 2025-04-09
2025 Grupp 1 Integrating pull request from #16236

a23carpa:
Tested and implemented the solution from pull request #16236, so it can me merged with Master.


b23danhe (approved):
Tested and works as promised! Tested in Firefox and Safari.

PR #16349, opened 2025-04-09
Update line.js

a23matsh:
There was no function for Straight type so I added another if and else if statement containing the functionality for Straight type, the same as Segmented


a23aleva (changes requested):
While the arrows do appear on straight lines after the fix they do not line up with the inclination of the actual line. So the arrows aren't invisible anymore but the solution is somewhat ugly. Also, the initial state of the straight line starts with an arrow already set while it should perhaps start clean with no set arrows on either end.

_Old version_
image

_Fixed version_
image

a23matsh:
Changed the functionality in line.js of how the arrow rotates for a more adaptive look when moving around the entities and also set z-index on svgbacklayer so arrows appear above boxes in update.js. It now works better than before but still not perfect.

image

Issue #16224, opened 2024-07-21
Almost all elements jump around when clicked and are in 'snap to grid' mode.


b22samer:
Description: Almost all elements jump up a few pixels when clicked while the 'snap to grid' mode is enabled.

Job: Make sure that the elements do not jump up when clicked on while the 'snap to grid' mode is enabled.

https://github.com/user-attachments/assets/77cd0f6e-5146-4924-ae2b-4791f1993fb9

b22samer:
I found out that as soon as you try to move the element with the mouse while snap to grid is enabled, the element jumps up. It doesnt only happen when you click on the element but also while moving it.

PR #16361, opened 2025-04-09
Updated theme.js

d22abdha:
issue #16224.
Elements no longer jump when 'snap to grid' is active. Replaced hardcoded condition for EREntity with all different element type check. Changed also the top position calculation to stop elements from jumping when clicked in 'snap to grid' mode.


a23miksk (approved):
The snap function now works better where it wont snap weirdly when starting to drag the element.

PR #16363, opened 2025-04-09
Fixed failed initialization of database because of query (in init_db.…

### LINKED ISSUE #16338: Failed initialization of database because of query (in init_db.sql) while using docker
I went to direcory LenaSYS/dockerenv and run this command docker-compose up --build. When container was build and running then I typed this url http://localhost/LenaSYS/install/install.php for install lenasys. In first page I left default username, password and more, and then press to next page. Second page I typed username: root and password: password. In third page I left default also and press to next page. In fourth I also left default (all was uncheck unless if user and/or database that already exists then check all boxes to overwrite). In fifth page I pressed install green-button and show installation and then progress bar show at 31% and show an error "Failed initialization of database because of query (in init_db.sql) ..." 

Image

I think init_db.sql is missing or couldn't find this file or something. So far as I know that there is two init_db.sql files which are located at LenaSys/Shared/SQL/init_db.sql and LenaSys/newinstaller/SQL/init_db.sql.

In newinstaller folder has init_db.sql, but not in install folder, so maybe it cause an error or myabe oudated docker file?

I'm using windows, but I think macos also has a problem.

a23dilel:
I fixed #16338. To fix this I need to add one year old mysql package in docker-compose.yml, otherwise mysql will get latest version in docker and our mysql code will be outdated.

### DIFF
--- a/dockerenv/docker-compose.yml
+++ b/dockerenv/docker-compose.yml
@@ -12,7 +12,7 @@ services:
       - "./githubMetadata:/var/www/html/githubMetadata:delegated"
     restart: no
   db:
-    image: mysql:latest
+    image: mysql:8.0.36-debian
     container_name: mysql-server
     environment:
       MYSQL_ROOT_PASSWORD: password


a23dilel:
If you still get an error then need to run this commend "docker-compose down" and remove mysql-data which located at LenaSYS/dockerenv/mysql-data. After you did then run this commend "docker-compose up --build" and try to install lenasys. Or go to this http://localhost/LenaSYS/install/install.php and go to Write over section and there check both "Yes I want to write over an existing database" and "Yes I want to write over an existing user." 
image

a22linre (commented):
I tested the installer with a new user and database, everything worked fine and went to 100% with no errors.
I also ran it using Docker, and it worked there too.  Looks good to me!

PR #16369, opened 2025-04-09
Fixed too large selection box for a recursive arrow.

a23emisn:
The selection box for a recursive arrow had too large of a margin to the left of the arrow before, now sits flush with the arrow.


a22abdjo (approved):
The solution works and can be merged, the selection box has an appropiate size for a recursive arrow.

PR #16371, opened 2025-04-09
Added a view mode toggle (Normal / Scroll / Overview)#16185

a23alera:
A small feature was added that allows the user to change how the section list is displayed. The user can switch between:
- Normal (default view)
- Scroll (limits height and adds internal scroll)
- Overview (zooms out to give a bird's-eye view of all items)

Skärmavbild 2025-04-09 kl  16 11 44
Skärmavbild 2025-04-09 kl  16 11 57

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -4443,3 +4443,18 @@ function getLocalStorage() {
     dropdown.value = selectedValue;
   }
 }
+
+// Toggle betwwen different view modes for the section list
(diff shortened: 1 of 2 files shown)


a23simry:
This solves issue #16185 in a larger capacity, though normal and scroll view buttons are essentially the same. The three buttons do as described and don't seem to mess with anything else. 

Screenshot 2025-04-10 120417
mobile overview is almost impossible to read but not that important at the moment.

PR #16374, opened 2025-04-09
Create canvas background in Courseed.php#16172

a23elism:
Implemented canvas-skeleton + necessary transparent styling

### DIFF
--- a/DuggaSys/canvasanimation.js
+++ b/DuggaSys/canvasanimation.js
@@ -1,0 +1,118 @@
+const courseedCanvas = document.getElementById('canvasAnimation');
+const ctx = courseedCanvas.getContext('2d');
+
+//if you resize after the page has loaded, the dot will be offset from the mouse. Hence this function
+function resizeCanvas() {
+    courseedCanvas.width = window.innerWidth;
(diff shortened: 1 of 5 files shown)


a23elism:
Everything should work as intended "light skimming" on other pages seem to work fine and dont seem to be affected by the canvas and its tiny css changes

The only page that should be affected is courseed.php that now should show a "trail" following the mouse (as seen in attached image, mouse not shown cause windows is windows when it comes to screenshots)
image

a23alera (approved):

EdvinWerner (commented):
The PR clearly fulfills the intent described at the end of the issue: _"The important thing is that it should look good and draw attention. The developer is largely free to design something subjectively appealing and give his suggestion in the form of a pull request."_ Some users might find the animation distracting since it draws a lot of attention, but given that this was the stated goal of the issue, I believe the PR addresses it properly.

However, there are some issues with clicking the links that are caused by this PR. Combined with the fact that courseed.php looking boring isn't an issue to begin with makes me come to the conclusion that this PR—along with its issue—should be scrapped.

PR #16376, opened 2025-04-09
Band-aid solution for fixing issue confirmation popup not closing

EdvinWerner:
This is a quite bad solution that fixes the issue with not being able to close the conformation pop-up-window when deleting a dugga. It's bad since it doesn't solve the underlying issue that caused an error in the function deleteitem. see screenshot for log of this error 
image


a23erigu:
This issue was fix in another issue

Issue #16196, opened 2024-07-03
Some elements disappear when they are held over other elements while they are in ghost mode.


b22samer:
Description: When using certain elements in ghost mode, they are placed behind other elements as soon as they are held over them, which makes it difficult for the user to see where they should be placed.

Job: Make these elements visible when they are in ghost mode while being held over other elements.

https://github.com/HGustavs/LenaSYS/assets/129371343/9aa9f250-c178-4f6e-8607-d93162af97a1

PR #16378, opened 2025-04-10
G1 2025 v2 issue #16196 elements ghost hover behind

a23carpa:
The problem was that the objects Z value where set to low to be able to appear above when hovering as ghost.
Setting these values to 2 seems to have fixed it.


a23aleva (approved):
Ghost elements that were moving behind static elements before now seems to hover above them. The change to z-index value depth seems to work as intentioned.

image

image

Issue #16283, opened 2024-09-19
SD-line problem


d22hanol:
The arrows doesnt show up if they are in straight mode, only if they are in segmented mode do they show up. 
this is a bug, fix this.
![Screenshot 2024-09-19 172242](https://github.com/user-attachments/assets/e0075dda-8d95-418c-b3e8-943225f2732c)
![image](https://github.com/user-attachments/assets/e2f5939f-2292-4612-a488-d6e929f6ab96)

PR #16389, opened 2025-04-10
G1 2025 v2 #16283 sd line problem

a23matsh:
image

### DIFF
--- a/DuggaSys/diagram/draw/line.js
+++ b/DuggaSys/diagram/draw/line.js
@@ -537,6 +537,13 @@ function drawLineIcon(icon, ctype, x, y, lineColor, line) {
             } else if (line.type == entityType.SE) {
                 str += iconPoly(SD_ARROW[ctype], x, y, lineColor, color.BLACK);
             }
+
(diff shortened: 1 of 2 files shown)


a23matsh:
This was supposed to go into #16349

a23aleva (changes requested):
The arrows are now visible in a straight line and the also gyrate when moving the objects they're attached to. There are still some problems though. The segment line now has some strange offshoot to the side of its target arrow which wasn't there before. 

When using the straight line the target arrow is applied above the 'to' object and not to its frame which should be the case. Only when the objects are lined up vertically above and below each other are both arrows correctly displayed. The 'from' arrow seems seems to behave fine though. 

Also found some inconsistencies within the code where some documentation names doesn't line up with the actual variable names used in the implementation such as the documented function.

_Master Segment Line_
image

_Master Straight Line_
image

_Static Line Artifact when attaching_
image

_New Segment Line_
image

_New Straight Line inclined_
image

_New Straight Line vertical_
image

_Inconsistencies with code documentation, [tip] should be [point]_
image

Issue #16360, opened 2025-04-09
Text is outside of button/box


a23danme:
The red circle shows the issue, located on Dugga 5 under Demo. Box/button is too short. Text "XYZ" and white boxes, "Lägg till" is not centered aswell.

![Image](https://github.com/user-attachments/assets/21b4f387-3f10-4c8f-badd-6ac69fcf1545)

PR #16398, opened 2025-04-10
Fixed text inside buttons

EdvinWerner:
Made the width of the buttons larger so now the text fits inside the buttons.


a23elism (approved):
Fixes reported Issue (#16360), text now text is now inside the buttons

Approved for merge 👍

PR #16399, opened 2025-04-10
G3 2025 V2 #16240 fix variables in html output

a22maxda:
Stripped & escaped exposed variable outputs to mitigate XSS vulnerabilities.


EdvinWerner (approved):
Looks good to me, no problems that I can find at least

PR #16402, opened 2025-04-10
Added inverse dependencies for createNewListEntry_ms.php #16381

a22linre:
Added inverse dependencies for createNewListEntry_ms.php to the documentation file for #16381.
Found 3 files where this microservice is included.


a22linre:
Done, added folder names

PR #16406, opened 2025-04-10
Updated microservices_inverse_dependencies

d22marwi:
No include logUserEvent_ms.php exist in the microservices files. Fixes issue #16384


a22linre (commented):
I couldn't find anything the include "logUserEvent_ms.php" either, so its correct. But the question is should it be as a header with "###" or just a normal text? "### logUserEvent (No dependencies)"

PR #16413, opened 2025-04-10
Imported React and Babel

f23linan:
Imported React and Babel to the files that have JQuery in order to be able to replace them in the furture.


f23linan:
For testing, add a div with ID "root" and add the script 
"<script type="text/babel">
    const App = () => <h1>Hello from React 👋</h1>;
    ReactDOM.createRoot(document.getElementById('root')).render(<App />);
  </script>" 

For testing purposes only! But this should make it possible to add future react components.

a23elism (approved):
The implementation of React + Babel works without any problems (functionality stayed the same from testing/usage of each page [were possible])

Testing method (i.e making a "Hello from React 👋") appeared on every page where react and babel was imported.

This is approved and ready for merge for further implementation of react 👍

PR #16422, opened 2025-04-10
Documented inverse dependencies for isSuperUser_ms.php

a23bensh:
No inverse dependencies found. Fixes issue #16383.


a23oskme (approved):
I did not find any missed inverse dependencies.

Issue #16332, opened 2025-04-09
Installer freeze


b20isato:
The installer freezes when using "Yes I want to perform the database-initializing as an transaction."

Issue #16373, opened 2025-04-09
Freeze occurs when the transaction is to be comitted.


a23gabli:
I've used logs to pinpoint the freeze to the `$connection->commit();` line in install.php. Error logging has shown that there are no faults with the queries in the init_db.sql file, as well as no apparent problems when running the SQL code by itself. There are no issues with permissions or preexisting transactions. So i cannot find what the problem is.

PR #16423, opened 2025-04-10
Removed transaction checkbox

a23gabli:
This fixes issue #16332 and #16373 .

The html code for the Transaction checkbox has been commented out, since it is broken and there is no real reason to have that option in this instance. The code remains in case anyone wishes to fix it.

### DIFF
--- a/install/install.php
+++ b/install/install.php
@@ -248,8 +248,10 @@
           Yes I want to write over an existing database.<br>
           <input title="Write over existing user." id="writeOver2" type="checkbox" name="writeOverUSR" value="Yes" />
           Yes I want to write over an existing user.<br>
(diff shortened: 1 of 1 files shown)


a23bensh (approved):
This change doesn't fully address the underlying issue but serves as a temporary solution **(can be merged for now)**. To properly resolve this, I'd suggest either:

Removing/commenting out both the UI checkbox and functionality if deemed unnecessary, not just commenting out the UI. 
**OR**
You fix the transaction functionality by investigating the transaction handling, and then restore the UI checkbox (if the feature is still needed of course).

a23louma:
A new issue #16429 was created for the remaining problems.

PR #16426, opened 2025-04-10
Remove possible inline css #15064

a23lisjo:
Fixes issue #15064. All inline css in codeviewer.php has been moved to external CSS files, style.css or codeviewer.css


a23kavis:
No inline css remains in the codeviewer.php, and fitting element IDs have been assigned for elements that required one for the move. This can be merged successfully

PR #16435, opened 2025-04-10
Fixed A4 text to align with the corner of the A4 outline

a23kevre:
Update update.js
Changed the x coordinate value of the A4 text to only move a little bit but always stay close to 0,0 like the corner of the A4 outline.

image


a23gabst1 (approved):
It works as it should, both for when the A4 is vertical and horizontal

PR #16441, opened 2025-04-11
Updated UML-entity header-text

b23ebbst:
Fixed the issue with the text in the header not fitting the element width. Adjusted the "Header"-portion of the updateElementUMLEntity-function to create line-breaks if the text is wider than the entity. Text no longer goes out of bounds.

image


b23danhe (approved):
This solution works as promised and fixed the problem. Tested in Safari and Firefox.

PR #16443, opened 2025-04-11
Fixed ER relations disappear

b23danhe:
Changed the value from 0 to 1, since there is no element matching "0" of "type 1" in the togglePlacementType() function. As a result, the menu disappeared. But when "1", there is a match and the ER relation element stays and the menu remains.
Skärmavbild 2025-04-11 kl  09 47 20

**Before:**

https://github.com/user-attachments/assets/a6ce4d11-76bd-4d62-a893-c633c0d86d81

**After:**

https://github.com/user-attachments/assets/8586f5da-7412-4f0f-88b2-d301853d6020


a22abdjo (approved):
ER relations försvinner ej från menyn när man väljer det första elementet. Därmed så funkar denna lösning som den ska.

PR #16462, opened 2025-04-11
G3 2025 v2 fix alignment of navbar items in accessed.php #15534

EdvinWerner:
Fixed the allignmentissue for the navbar items in accessed.php


c22isako (approved):
Alignment looks good. Doesn't break anything from what I can see.

PR #16463, opened 2025-04-11
Updated daywidth

EdvinWerner:
Updated daywidth so now the collums fill the entire svg. Due to the hardcoded nature of the weekwidth calculation there is a possibility that tihis solution only works when theres 9 weeks in a course. I havn't found another course with the same layout where my theory can be tested unfortunately


EdvinWerner:
I looked in to the possibility of having some sort of dynamic values for the collums but I at least can't think of a solution that based on the current code standard is responsive. I therefore recommend that this PR is closed.

PR #16465, opened 2025-04-11
Added a text about inverse dependencies

### LINKED ISSUE #16449: Document inverse dependencies: copyCourseVersion_ms.php
)

a23dilel:
Fixed #16449 for adding a text about inverse dependencies.


a23axehe (approved):
Looks OK, could not find any inverse dependencies for copyCourseVersion_ms.php either.

PR #16471, opened 2025-04-11
Documenting dependencies for 'editContentOfExample_ms.php'

a23oskme:
Fixes issue #16445


a23dilel (approved):
I also check if there are inverse dependencies which I couldn't find and also you also added new line each section which is unnecessary, but looks good!

a23louma:
There is a conflict and maybe the best is to only add the inverse dependencies without the new lines and then a new review can be done.

PR #16485, opened 2025-04-11
Documented inverse dependencies for updateCourseVersion_ms.php

### LINKED ISSUE #16460: Document inverse dependencies: updateCourseVersion_ms.php
)

a23bensh:
Fixes #16460, documented no inverse dependencies for updateCourseVersion_ms.php in Microservices_inverse_dependencies.md


a22linre (commented):
Couldn't find any inverse dependencies for updateCourseVersion_ms.php either, so it looks good.

Issue #16370, opened 2025-04-09
Self Calling Line assigned arrows error


a23aleva:
When using the self calling line in a class diagram the line automatically comes with a pre-set arrow. However the option to assign arrows (or icons) to the ends of the line still remains available. As shown in the pictures the attached new icons selfinsert at strange positions. At least the icon connected to the right icon drop down which appears on the opposite side of the actual class box and arrow it is assigned to. The best solution would make the original line appear clean without arrows and the attached icons would then connect to the actual ends of the line.

_Self calling line_
![Image](https://github.com/user-attachments/assets/69876d98-f347-415b-8075-4cfaea92a0ca)

_Left icon attached_
![Image](https://github.com/user-attachments/assets/55399590-c5e7-4f3f-b231-638c4296ee43)

_Right icon attached_
![Image](https://github.com/user-attachments/assets/7f1ec82e-a0bc-4a29-86de-0fb9ab579ada)

PR #16488, opened 2025-04-11
Fixed recursive arrows #16370

a23miksk:
The recursive line is now able to be modified with arrows. The cardinalities, label and changing to dashed is still broken.
State diagram states are a bit broken because it wants to do a straight line when the recursive relation line is a segmented one (changing the line to segmented fixes this). A fix might be to make recursive lines a segmented line directly when created on a state.

[Screencast_20250411_113443.webm](https://github.com/user-attachments/assets/e1a3298d-801c-4f54-8949-b81342d871c9)
_The recursive line now supports different arrows_

### DIFF
--- a/DuggaSys/diagram/draw/line.js
+++ b/DuggaSys/diagram/draw/line.js
@@ -23,6 +23,8 @@ function drawLine(line, targetGhost = false) {
     if (isSelected) lineColor = color.SELECTED;
     let fx, fy, tx, ty, offset;
     
+    // Sets the to-coordinates to the same as the from-coordinates after getting line attributes
+    // if the line is recursive
     if (line.kind === lineKind.RECURSIVE){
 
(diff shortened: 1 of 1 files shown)


a22nevan (changes requested):
As stated, the recursive line is modifiable with arrows. All types of arrows are available and are aligned properly at 1x zoom level. At higher or lower zoom levels however, the bottom arrow gets misaligned.

This should probably be fixed before merging with the weekly branch.

Classes with recursive lines at 1x zoom. Arrows work as intended:
image

Same classes at 0.5x zoom. Bottom arrows are positioned below intended location:
image

2x zoom. Bottom arrows are positioned above intended location:
image

a22nevan (approved):
Arrows now align properly at all zoom levels. 

0.5x Zoom:
image

1x Zoom (Already functioned as intended):
image


2x Zoom:
image

PR #16490, opened 2025-04-11
Updated IE-entity header

b23ebbst:
The IE-entity is now updated in the "drawElementIEEntity"-function to be able to print longer names in the header without the svg clipping the text.

image


a23emisn (changes requested):
The implemented solution works and the header is adjusted dynamically by longer texts. However the margin below the the text in the header is larger than necessary and should be corrected before merging. See picture. 
image


This margin can be reduced by correcting the value added to "headlerLines.length" at line 370 in element.js to: "0.5", or whatever value < 1 that produces the most accurate margin .

370 `let height = texth * (headerLines.length + 0.5) * lineHeight;`

a23emisn (commented):
Header text margin is fixed. I approve this change, and the PR may be merged.

Issue #16334, opened 2025-04-09
Bug: Adding an element or reloading the page restore deleted elements


a23erigu:
In a course if you delete a element and then reload the site or add a new element then the deleted elements get returned

HGustavs:
There should still be a difference between deleted and hidden elemnents.

Som elements that are connected to other elements may not be allowed to be deleted, and may therefore be hidden instead. 

A newly created element should not be connected to anything and it should be deleted without isssues.

PR #16491, opened 2025-04-11
G3 2025 #16334 bug restoration of deleted objects

c22isako:
Fixed delete function, wrong handling was made within the function deleteAll and deleteItem which resulted in error.


EdvinWerner (approved):
Looks good, nothing wrong with it, approved for merge

Issue #16430, opened 2025-04-10
The diagram save button isn't working


a23aleva:
When trying to save a diagram you can enter a name for it but when you push the 'save' button nothing actually happens. The window should close and the saved diagram should appear as a loadable option. Instead pressing load only states that no saves could be found. 

![Image](https://github.com/user-attachments/assets/ea6d380d-ecb4-4db9-bdc8-1170a9fe90cb)

![Image](https://github.com/user-attachments/assets/470dba51-dce2-45f5-a3d9-4be2d3c918df)

_Somewhere in an old version of lenaSYS the save/load option was functioning and looked like this:_
![Image](https://github.com/user-attachments/assets/fac41ac2-787a-48aa-aadc-adf2590e0b3f)

PR #16498, opened 2025-04-11
G1 2025 v2 16430 the diagram save button isnt working

a23aleva:
This is the solution to issue #16430. 

image

image


a23eliax (approved):
Working as expected. I don't know why it says that it's 1 save found when it's zero but the button works, it saves and deletes correctly.

https://github.com/user-attachments/assets/0988e166-3150-4be7-aeea-0a464c4b24a9

PR #16499, opened 2025-04-11
G1 2025 v2 #16191 the example diagrams doesnt load correctly

d22abdha:
Example state diagram is now fixed, lines and elements now load correctly without any problem


a23miksk (approved):
The examples seems to be fixed now with complete elements and arrows
_state diagram before_
state broken
_state diagram after_
state fixed
_uml diagram before_
uml broken
_uml diagram after_
uml fixed

PR #16500, opened 2025-04-11
Documented getCourseed in Microservices_inverse_dependencies

### LINKED ISSUE #16454: Document inverse dependencies: getCourseed_ms.php
)

a23wilgr:
getCourseed is never used as an include. In the Microservices.md it say getCourseed has been replaced with retrieveAllCourseedServiceData

fixes #16454


a23dilel (changes requested):
I also couldn't find inverse dependencies. So remove both and add someting "No inverse dependencies"

a23petny (approved):
getCourseed has no inverse dependencies

a23petny:
I approved these changes, as getCourseed has no inverse dependencies

Issue #16427, opened 2025-04-10
Find out where the top bar components are located


a23erigu:
Try to find where the top bar and its commponents are created and find out if anything there uses jquary

This is the top bar
![Image](https://github.com/user-attachments/assets/bb160b51-f974-4a7f-83f5-a09d602297d8)

a23elism:
### Correction to all instances shown above

Instance 5 and 6 dont seem to be used at all so BOTH can be removed in the future
never called anywhere in the project

PR #16524, opened 2025-04-11
Replacing/removing Jquery from topbar with React/Babel

a23elism:
Removed Instance 5 and 6 of JQuery from #16427 
Simply removed dead JQuery code

### DIFF
--- a/Shared/navheader.php
+++ b/Shared/navheader.php
@@ -682,25 +682,6 @@ function mouseOverSearchInput() {
    }
 }
 
-
-/*Toggle the hamburger list containing the icons to appear*/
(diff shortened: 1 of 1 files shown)


a23elism:
### Temporary Fix for remaining instances of JQuery
These temporary fixes are only here to replace JQuery until further information about
going from JQuery to React/Babel have been clarified

a22maxda (approved):
Cannot find any issues with the changes, and the code appears to be correctly translated from JQuery to React. The dead code shouldn't pose any issues either as I cannot find any other references to the removed parts.

PR #16526, opened 2025-04-11
Implemented overlap-detection for pasting elements.

a23emisn:
Elements could previously be pasted on top of each other, as shown in the picture below.
image

The entityIsOverlapping() method, which detects entity overlaps, has been added to the pasteClipboard() method. This allows the parent function to flag the overlap, deny pasting and notify the user. Trying to paste objects on top of eachother will result in an error as shown in the picture below.
image


a23miksk (approved):
This issue is now largely fixed. Trying to copy/paste elements over each other is now prohibited. It is still possible to cause some overlap by mass copy/pasting a lot of elements, but this will not realistically come up with a normal usage of the system.
I approve this change.

Issue #16508, opened 2025-04-11
Logout: Change jquery to react


a23erigu:
Switch the jquery used in the logout to react (babel)

Picture of the logout
![Image](https://github.com/user-attachments/assets/5beb9b9d-a162-477a-bc91-c0d97333796a)

There is some helpful info for this issue in [#16427](https://github.com/HGustavs/LenaSYS/issues/16427) and [#16428](https://github.com/HGustavs/LenaSYS/issues/16428)

PR #16527, opened 2025-04-11
Logout: Change JQuery to React/Babel

a23elism:
Switched Jquery to React/Babel in the logout popup, requiring a rebuild (only solution i could think of) of popup and buttons
as per #16508

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -1898,7 +1898,6 @@ div .navButt:hover {
   background-color: rgba(0, 0, 0, 0.6);
   z-index: 8000;
   margin:0 auto;
-  display:none;
 }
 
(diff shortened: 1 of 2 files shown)


a22maxda (approved):
Converted popup seems to work and code looks to function about the same. There is a bug in firefox which sometimes causes the react/babel code to break causing the popup to not appear, only fix seems to be to reload site. IT IS A FIREFOX ISSUE, NOT WEBSITE / CODE. Attached are screenshots of the error and Firefox's source documentation regarding it.

Screenshot 2025-04-14 at 11 37 36

Screenshot 2025-04-14 at 11 59 58

Screenshot 2025-04-14 at 12 02 39

Issue #16344, opened 2025-04-09
Bug: Undo doesn't work


a23erigu:
The undo for delete doesn't work as intended, it only appears when pressing the trash can without selecting an element (which does not delete) and does not appear when actually deleting

PR #16531, opened 2025-04-11
G3 2025 v2 #16344 fix undo

c22isako:
Changed the function cancelDelete() to affect the right "class"-items. Also made some changes to deleteAll() function so that it removes from the internal class as well, not just the database.


a23elism (approved):
New changes to the sectioned.js file appear to have fixed the "buggy" undo, all functionality seems to be retained (couldnt find anything that wasn't working atleast).

Bug removed/Fixed and Approved for merge 👍

Issue #15647, opened 2024-05-13
Snap to Lifeline or object after placement


a21liltr:
https://github.com/HGustavs/LenaSYS/assets/129370135/419d5813-ed41-4c4b-b22a-39cfc0a66e05

## Description
Sequence Activation works really well as a ghost. It snaps to Sequence Lifelines and Sequence Lifeline Objects as it should. But after placing the activation, it will no longer snap to Lifelines or Objects anymore. 

## Job
Make Sequence Activation snap to the dashed line on Sequence Lifelines and Sequence Lifeline Objects after the user has already placed but want to move the Activation.

a22jacka:
After trying around for a bit I found that the issue seems to be that the code uses the `mouseenter` event on the lifeline to check when to snap, which works fine for ghosts. However when dragging an existing element, the `mouseenter` event doesn't trigger for the lifeline since the activation is front of it. There are events that can see through elements though, I've tried using `document.elementsFromPoint(x, y)` which works to grab all the elements that the mouse is over.
![image](https://github.com/HGustavs/LenaSYS/assets/129258593/da4509ae-e6c1-40d3-a44a-59872e79c807)

My current attempt at using this however causes the site to lag a decent bit when dragging activations and the dragging stops working if the mouse leaves the activation for any reason.

a22jacka:
After testing different stuff for a couple of days I can't get this to fully work. I managed to detect the lifeline by using the method in the previous comment but something in the calculation doesn't want to work. Actual elements use the use the same X-calculation as the ghost element but for some reason it causes the jump seen in the video below. After the jump it keeps following the mouse but with an offset.

https://github.com/HGustavs/LenaSYS/assets/129258593/738a2d7d-95d4-41a5-b231-9c4c6a6cf15b

To properly freeze the element you need to somehow change value of `left` in `updateElementDivCSS` in theme.js, setting that to a constant value when an activation is over a lifeline, then you can probably freeze it.

The code can be view in #16113

a23eliax:
This was an hard issue with a lot of code in different files. It will probably take some days before I get this to work.

PR #16532, opened 2025-04-11
G1-2025-v2-#15647 snap to lifeline or object after placement

a23eliax:
**helper/mouse.js:**
Extracted ghost-snapping logic into a reusable function: snapElementToLifeline(element, targetId) to allow both ghost elements and placed elements to snap to the center of the nearest lifeline.
     snapSAToLifeline() still handles ghost snapping on hover, but its logic now matches the real element snapping logic more closely.
     When placing a new element (placeElementAt(x, y)), the code creates a new element, finds the nearest lifeline using findNearestLifeline(x, y) and snaps the element to that lifeline before adding it to the diagram.

**events/mouse.js:**
Modified the mouse release (mouseup) behavior. If elements were moved (context.length > 0), each is snapped to the nearest lifeline after release using the same snapElementToLifeline() logic.

Uploading Inspelning 2025-04-11 170037.mp4…

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1001,11 +1001,6 @@ function mouseMode_onMouseUp(event) {
  * @description Event function triggered when the mouse has moved on top of the container.
  * @param {MouseEvent} event Triggered mouse event.
  */
-
-function snapToLifeline(container, lifeline) {
-    const snapThreshold = 10; // Max distance for snaping
-}
-
 function mmoving(event) {
     lastMousePos = new Point(event.clientX, event.clientY);
     switch (pointerState) {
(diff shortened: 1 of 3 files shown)


b23danhe (changes requested):
The snapping function works, but the recent changes have introduced some new issues. If an object is placed anywhere other than on the lifeline, it cannot be moved to a new position, instead it snaps back to its original placement this applies on all other objects. 
The actor-object moves down when trying to move the actor-object.

**Video 1:**

https://github.com/user-attachments/assets/1b4773bd-515c-42d1-ac2d-e02379d6d487

**Video 2:**

https://github.com/user-attachments/assets/d6204c10-621c-4a17-b9f1-f12ad7a8a7f1

b23danhe (approved):
Works as expected—tested in Firefox!
The only minor issue is that when replacing an existing object on the lifeline, there's no visual indicator (ghost) showing that the object will snap to the lifeline.
However, this isn't a requirement for the current issue and can be handled in a separate issue.

https://github.com/user-attachments/assets/2460263c-d1b9-4368-b698-4e826001788d

Issue #16368, opened 2025-04-09
If statement that is 400 lines long: createGithubCodeExample_ms.php


a23wilgr:
Perhaps make it shorter, and/or split it up

PR #16533, opened 2025-04-11
G3 2025 v2 #16368 reduce if statement size

a22maxda:
Removed old IF-statement - instead broken up into several functions. Logic and functionality should remain the same as it was before. Necessary to ensure logic and variables (parameters) are the same as it was previously


a23lisjo (approved):
The If-statement is now divided into functions instead and the logic, variables and parameters are the same as the old versions. The code looks good!

Issue #16412, opened 2025-04-10
Item ordering using Jquery


HGustavs:
I believe that one of the last pieces of code that still makes use of jquery, is the ordering of elements.

Screenshot from old version of system (feel free to generate screenshot from most recent version) When we drag and drop course elements, we can reorder those elements dynamically. 

![Image](https://github.com/user-attachments/assets/e220a5a3-e2ed-45bd-b237-2d84ad6455c1)

One key part of the system that needs to be addressed early is this reordering, without which we cannot remove jquery completely.

Is this still using jquery, and if so, how can we implement this using react without downloading react components made by other people?

PR #16534, opened 2025-04-12
Replace JQuery drag-and-drop with custom React-style #16412

a23alera:
Disabled and commented out the old jQuery function and replaced it with a custom React-style drag-and-drop sorting. Added a login check to allow sorting only for logged-in users.


EdvinWerner (approved):
I can't find any faults in this so in my opinion it's ready for merge

PR #16546, opened 2025-04-14
Documented inverse dependencies for deleteDugga_ms.php

### LINKED ISSUE #16474: Document inverse dependencies: deleteDugga_ms.php
)

a23bensh:
Fixes #16474, documented that no inverse dependencies were found.


a22linre (commented):
I couldn't find any inverse dependencies for deleteDugga_ms.php either, so it looks correct

PR #16547, opened 2025-04-14
Remove as much inline css as possible from fileed.php #15065

a23osckr:
Resolves issue #15065. I have moved as much inline css from fileed.php to style.css as reasonably possible as well as some light cleanup. Nothing appears broken.


a22maxda (approved):
Buttons appear to retain functionality and there is no change in style.
Screenshot 2025-04-14 at 14 58 49
Pictured are the affected buttons.

Issue #15036, opened 2024-04-26
UML cardinalities overlapping


a22jacka:
## Description
Since UML lines always connect at the same point it causes their cardinalities to be displayed in front of each other

![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/f7b1d29e-29e8-4df0-90a0-286cf093d8db)
![bild](https://github.com/HGustavs/LenaSYS/assets/129258593/00d7983e-88cd-48da-9a0b-e5ef8ac6d2aa)

## Job
Seperate the lines of 2 more lines are connected to the same side

PR #16562, opened 2025-04-14
Fix for issuie #15036, overlapping cardinalities on UML

a23carpa:
Added the line offset value to the from x, to x, from y and to y values before the cardinality is added. 
Screenshot from 2025-04-14 12-39-46


a22abdjo (approved):
This fix has been tested and approved, cardinalities no longer overlapp on UML.

PR #16563, opened 2025-04-14
Text in header no longer gets clipped

b23ebbst:
Fixed issue with the text in the header getting clipped if it is longer than the entity is wide - the text now fits within the header even if it is longer than the entity is wide. 

image


a23gabst1 (approved):
It works as intended, the text no longer gets clipped.

image

Issue #15262, opened 2024-05-02
Example IE/UML line error


a22solmi:
## Description
When using box select in IE and UML example digrams there is an error in the console. Removing the lines and replacing them removes the error. 

## Job
Find what is causing the error and fix it so all examples work with box select

PR #16567, opened 2025-04-14
Fixed the issue "Example IE/UML line error" #15262

a23gabst1:
I solved the issue regarding the errors generated in the console by removing a misplaced comma in the iconPoly function, and also the errors don't appear in the console does not appear when using the box selection. PS, also indented the code.


a23emisn (commented):
The implemented patch has solved the problem of diagram-line errors appearing in the console when loading example diagrams and using the box select for the diagrams. This PR can be merged.

Issue #16519, opened 2025-04-11
Endpoint directory name query


HGustavs:
We want the directory to handle queries that provide part of the name of a micro service, such as 

directoryendpoint?name=course

Related to issue "Microservice Directory Database #16538 ".

HGustavs:
sub-issue of #16518 16518

PR #16568, opened 2025-04-14
Add communication between EndpointDirectory and nameQuery - issue 16519

a20artje:
Still needs to add database connection, right now there is only a link between the directory and the nameQuery file and the display of the information.

To test go to the php page of endpointDirectory:  http://localhost/LenaSYS/DuggaSys/microservices/endpointDirectory/endpointDirectory.php

Add ?name=whatever.

Some data should show up (same no matter the parameter).

Partly fixes issue #16519

### DIFF
--- a/DuggaSys/microservices/endpointDirectory/endpointDirectory.php
+++ b/DuggaSys/microservices/endpointDirectory/endpointDirectory.php
@@ -1,16 +1,19 @@
 <?php
 
 include_once "./nameQuery.php";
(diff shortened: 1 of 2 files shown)


a23bensh (approved):
endpointDirectory.php executes and works. nameQuery.php and its mockup is functional aswell. Looks good (as a placeholder of course).

(noticed that the commented out "directoryRendering" is missing .php)

PR #16575, opened 2025-04-14
Update Microservices_inverse_dependencies.md

### LINKED ISSUE #16472: Document inverse dependencies: createDugga_ms.php
)

a22wilas:
Found no inverse dependencies for createDugga_ms.php
fixes #16472


a23petny:
Found no inverse dependencies for createDugga_ms.php

a23petny (approved):
The file createDugga_ms.php indeed has no inverse dependencies.

PR #16576, opened 2025-04-14
Create Documentation folder and template

### LINKED ISSUE #16574: Add new folder and template for microservices documentation
> We could structure the documentation in a folder called "Microservices Documentation" where we create separate subfolders or documents for each microservice, instead of having all documentation in the same document. This would make it more structured and documentation for specific services would be easier to find.
> 
> One map called ex. "Shared microservices documentation" that contains one md per service - so one md for ex. "getUid", another for "logUserEvent" etc. 

 _Originally posted by @a23wilgr in [#16537](https://github.com/HGustavs/LenaSYS/issues/16537#issuecomment-2801400047)_

b23johgu:
Added folder for documentation and template
fixes #16574

### DIFF
--- a/DuggaSys/microservices/Microservices Documentation/template.md
+++ b/DuggaSys/microservices/Microservices Documentation/template.md
@@ -1,0 +1,32 @@
+Example of template for the documentation:
+
+# Name of file/service
+
+## Description
+*Description of what the service do and its function in the system.*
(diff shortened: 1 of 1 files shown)


d22marwi (approved):
Good tamplete for documentation, approved!

PR #16578, opened 2025-04-14
Update line.js

a23matsh:
The problem where arrows were not visible in straight mode is now fixed.

image

Known issues that need to be solved in **NEW** issues:
- When the "to-entity" is on the left side of the "from-entity" the selection box of the line becomes too big so the arrow is there but its under the entity. But when for example the arrow starts from the entity on the left side and goes to an entity on the right side its perfect.
image
image

- In segmented mode in some angles the arrows dont adapt to the rotation of the segmented mode where they should be straight, they rotate the same way as in straight mode.


a23miksk (changes requested):
The fix works like mentioned in the comments, but also the problems. A fix probably shouldn't create two new problems, and these should be fixed before merging. 
Also, the added lines 705 through 723 in line.js are completely incomprehensible compared to the code that was there before. So either add comments describing what happens here, or make the code simpler.
Nitpicks: avoid leaving un-commented, and there are now a redundant break line in the switch statement (line 595). There are also now some branch conflicts so make sure you use the latest version of the weekly branch.

a23matsh:
The code is now changed to fulfill the demands of the review, the following has been done

- The arrow rotation implementation had also been applied to the Segmented mode which caused the arrows in segmented mode to also rotate in different angles when they should be straight, this is fixed.

- There was no documentation for the newly implemented code row 705-723 which now has been made for extra clarity.

The regarding code section:
_function rotateArrowPoint(base, point, clockwise) {
    const angle = Math.PI / 4;
    const direction = clockwise ? 1 : -1;
    const angle = Math.PI / 4; // 45 degrees in radians
    const direction = clockwise ? 1 : -1; // Decides rotation direction

    // Calculate how far the point is from the base
    const dx = point.x - base.x;
    const dy = point.y - base.y;
        return {

    // Rotate the point around the base
    return {
            x: base.x + (dx * Math.cos(direction * angle) - dy * Math.sin(direction * angle)),
            y: base.y + (dx * Math.sin(direction * angle) + dy * Math.cos(direction * angle))
        };_

- There was a redundant break which is fixed and also "out-commented code" which could just be removed which now is done.

- There was a branch conflict which now is resolved by merging G1-2025-v2 into Rebranched and resolved conflicts in line.js to make sure the latest version is used.

a23vinge (approved):
As demanded, comments has been added between the section **705** through **723** which has added context to the code, given a better understanding of the situation and the purpose. The adding of @param and @description is great too.


comments as: "Rotates a point around another point (the base) by 45 degrees. This mainly used to create angled corners of an arrowhead" Comments like this give more context and a better understanding of the codes purpose and recommend that comments as this will be given in the future.

PR #16581, opened 2025-04-14
Fab button 16345

f23linan:
Replaced 2 buttons with react buttons instead, witht the exact same functionality as the previous buttons. 

In both the courseed page and the sectioned page


a23jesja (changes requested):
Hi, for some reason when i check out this branch, nothing shows up. I tried checking out the group branch and that works fine, but this specific branch doesn't work at the moment.. 

This current branch:
image

Group branch:
image

Issue #16551, opened 2025-04-14
Basic.php investigation


HGustavs:
No part of basic.php shoudl be doing other things than scafffolding e.g. processing parameters.

It sohuld not do things such as providing functiuonality for reading data that is specific for any sub-service.

PR #16584, opened 2025-04-14
Documented Basic.php investigation #16551

a22linre:
I reviewed basic.php for issue #16551, which is about making sure the file only contains scaffolding (general purpose & reusable code). Most of the file looks fine, functions like getOP(), logging helpers, and error handling are generic and reusable.

However, I flagged a section that I don't think is scaffolding, database setup and table creation (logEntries, userLogEntries, etc.). This sections feel more like app-specific logic and might be better placed in separate files.

### DIFF
--- a/Shared/basic.php
+++ b/Shared/basic.php
@@ -169,6 +169,8 @@ $dbVersion = 6;
 $metadataDbVersion = 2;
 //---------------------------------------------------------------------------------------------------------------
 
+
+// This section might be doing more than scaffolding.
 try {
 	$log_db = new PDO('sqlite:../../log/loglena'.$dbVersion.'.db');
 } catch (PDOException $e) {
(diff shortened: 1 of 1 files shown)


a23arvme:
I was thinking maybe "gettheOP" could be removed and replaced by getop?

a23danme (commented):
I think this should stay in the file: 

function customErrorHandler
function getOP
function getOPG
function gettheOP
function makeRandomString
function endsWith
function getOS()
function getIP() 
function getBrowser()
function formatted_debug
function debug_to_console.
(and obviously the things around them that are needed such as "set_error_handler("customErrorHandler");")

The rest, other parts as mentioned and commented in the file (logEngtries etc.). This is what as I understand it should be put in another file and the result will be basic.php being clean, small and more generall without app-specific-logs etc.

I think gettheOP could as mentioned be removed and replaced by getOP. In my opinion the names getOP and getOPG is not the best, is it not simpler to just give them some names like postX and getX  due to the first one using POST and second one using GET? X is just something i wrote to represent something short. Could be for example Param.

PR #16587, opened 2025-04-14
Made the A4 text bigger

a23kevre:
I have come to the conclusion that it is not actually the A4 text scaling upwards and downwards. The A4 text always stays the same size. It is the whole program that is zooming in and that is why the text looks smaller because of the perspective the zoom gives to the ruler and background etc. 

Therefore I updated the diagram.css file and changed the font-size of #a4Text to be a bit bigger to somehow help with the issue. No other file except some enter on update.js has been changed so it shouldn't do any harm to other stuff.

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -830,7 +830,6 @@
     z-index: -11;
 }
 
-
 #a4Rect, #vRect {
     display: none; 
 }
@@ -852,7 +851,7 @@
 }
 
 #a4Text {
- /* no styling yet */
+    font-size: 32px;
 }
 
 /****************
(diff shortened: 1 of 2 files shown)


a23eliax (commented):
The text is bigger. I don't really see the issue as a problem although the issue was to fix it from scaling. In both videos below, the text stays the same size relative to the screen the whole time, but if you look at it relative to the grid, it does scale in both. So I didn't quite know what to look for.

_Before:_

https://github.com/user-attachments/assets/ef238035-b605-4aab-8aa3-7efcf61c4cee

_After:_
https://github.com/user-attachments/assets/b6d2f3ef-841d-4666-90dd-ed8c3675ea67

Issue #16538, opened 2025-04-14
Microservice Directory Database


HGustavs:
We want the microservice directory to be stored in a sqlite databas.

So we need to produce SQL code and installation script additions for creating this database.

We want a separate database so that power users can download the database and query that database directly.

HGustavs:
I created a separate issue for documenting the create table statements as I did not find any discussion on the details of this in this issue? Or do we discuss those details in some other issue?

PR #16600, opened 2025-04-15
G2 2025 v2 #16538 a23oskme

a23oskme:
Works as is. However, it will have to be changed whenever the structure of the microservice md files has been decided.
Also, I left the test md file in there (testMarkdownFile) along with the temporary php file that displays the database (microserviceList.php). These are for testing and debugging and can be removed when the database is completed. 

We should consider creating some documentation or guide for this when it has been changed to follow the chosen markdown file structure. 

Fixes issue #16538. But I don't know if it fulfills this "We want a separate database so that power users can download the database and query that database directly." specifically the power user part.


a20artje (approved):
SQLite database is created with an installation script, content is taken from a md file and added to the db.
Works as described in PR.
Can't find any bugs, doesn't break the rest of the system,
Needs to be tested with the actual microservices documentation file.
Looks good to me.

PR #16601, opened 2025-04-15
Lines are correctly adjusted to correct anchor points between entities

a23emisn:
Anchor points are now correct, before the fix lines anchored at the wrong edges when placed next to each other. This led to lines getting stuck underneath entities which was faulty behavior. The problem were faulty offsets in the getLineAttributes() method, they've been corrected.

The lines are now correctly adjusting to the placement of entities and anchors at the correct edges, the closest possible.
image

Element moved to the other side: 
image


a23carpa (approved):
This solves the target issue. The lines drawn from Left to Right no longer get a increased length.

Note.
All the white spaces added and removed kinda hides where the solution is done.

PR #16604, opened 2025-04-15
Inverse dependencies for createListEntry and deleteListEntries

### LINKED ISSUE #16594: Document inverse dependencies: createListEntry_ms.php AND deleteListEntries_ms.php
)

b23johgu:
No inverse dependencies found for either createListEntry or deleteListEntries.
fixes #16594


a22linre (commented):
I couldn't find any inverse independencies for these either, so it looks correct.

PR #16607, opened 2025-04-15
Documented getCourseGroupsAndMembers_ms.php AND getCourseVersions_ms.php

### LINKED ISSUE #16595: Document inverse dependencies: getCourseGroupsAndMembers_ms.php AND getCourseVersions_ms.php
)

a23wilgr:
Are not used anywhere. getCourseGroupsAndMembers has been replaced with readCourseGroupsAndMembers, and getCourseVersions has been replaces with readCourseVersions.

fixes #16595


b23johgu (approved):
Looks good, I found no inverse dependencies for either getCourseGroupsAndMembers or getCourseVersions.

PR #16610, opened 2025-04-15
Documented getListEntries_ms.php AND getUserDuggaFeedback_ms.php

### LINKED ISSUE #16598: Document inverse dependencies: getListEntries_ms.php AND getUserDuggaFeedback_ms.php
)

a23wilgr:
getListEntries is not used anywhere, but is supposed to be replaced with retrieveAllSectionedServiceData, but that file does not exist. getuserDuggaFeedback is not used, is replaces with readUserDuggaFeedback

fixes #16598


a23gabli (approved):
Documentation is correct.

Issue #16455, opened 2025-04-11
The "hitbox" for some object does not match the object it self


b23danhe:
The "hitbox" for some object does not match the object it self. It seems to apply on all shapes that are tilted or not a square. This makes it unable to place elements within a certain distance from other elements.

https://github.com/user-attachments/assets/c5d8b0bc-4b67-4960-971e-76bcc975f07b

PR #16611, opened 2025-04-15
G1 2025 v2 #16455 the hitbox for some object does not match the object it self

a23aleva:
Solved the hitbox issue for the ER Relation element. Its angled side can now be placed closer to other elements without the collision detection kicking in to reject its position.

https://github.com/user-attachments/assets/5e5002b9-21fb-453b-82aa-14c6e09843fb


a22abdjo (approved):
Hitbox issue has been tested and can now be confirmed that the ER relation element can now be placed closer to other elements without the collision detection kicking in.

PR #16615, opened 2025-04-15
Connect name query to database - Issue #16570

### LINKED ISSUE #16570: Endpoint directory name query database connection
The name query file in microservices/endpointDirectory is not completed, it has communication with the endpointDirectory file, but right now it only sends the same array no matter what the parameter in GET is.

Fix the connection to the database and use the query to correctly return the information from the database,

a20artje:
Added a database connection to name query, now correctly checks database for data.
To test open /endpointDirectory/EndpointDirectory.php and add ?name=example_ms at the end of the url.
Need to install db and fill with data first.

This error is normal and will be fixed by another issue.
Fatal error: Uncaught Error: Call to undefined function render() in C:\xampp\htdocs\LenaSYS\DuggaSys\microservices\endpointDirectory\endpointDirectory.php:32 Stack trace: #0 {main} thrown in C:\xampp\htdocs\LenaSYS\DuggaSys\microservices\endpointDirectory\endpointDirectory.php on line 32

fixes #16570

### DIFF
--- a/DuggaSys/microservices/endpointDirectory/nameQuery.php
+++ b/DuggaSys/microservices/endpointDirectory/nameQuery.php
@@ -2,14 +2,19 @@
 
 function getServicesByName($name){
     //get names from database
-/*
+    $dbFile = __DIR__ . '/endpointDirectory_db.sqlite';
+    $pdo = new PDO('sqlite:' . $dbFile);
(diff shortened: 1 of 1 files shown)


a23axehe (approved):
Code works, solves issue by enabling searching for microservices in endpointdirectory_db.sqlite by name. OK

Issue #16589, opened 2025-04-15
Add a button go to course page when new installer finsihed install the lenasys


a23dilel:
Here you can see that new installer shows 100%, but there is no button which can lead to somewhere. 

![Image](https://github.com/user-attachments/assets/cb6430d6-bb7a-4420-87e1-e92814fe9de5)

I was thinking, when new installer is finish install the lenasys then should be showing a green button which lead to course or home page.

PR #16623, opened 2025-04-15
Added a button for go to course page when installer is finish to inst…

a23dilel:
Now new LenaSYS installer has a button. When installer is finish to install the lenasys then the button will appear which lead to course page when the the button is pressed. This fixed issue #16589.

### DIFF
--- a/newinstaller/installer_ui.js
+++ b/newinstaller/installer_ui.js
@@ -212,6 +212,7 @@ function start_installer() {
 			if (targetValue === 100) {
 				setProgressBarWidth(100);
 				//progressPercentage.innerHTML = "100%";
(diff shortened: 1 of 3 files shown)


b23johgu (commented):
Skärmavbild 2025-04-15 kl  14 16 31

Looks and works great! Idk if its our responsibility to style it, but maybe add some margin around the button to separate it from the progress bar if its possible? 

As for the chosen text "Go to course" I'd suggest we go with "Finish" as that is what most installers have as "final" button. Some other suggestions are "Done" or "Complete installation" :)

a23dilel:
Ok, I have added a margin and renamed text!

PR #16632, opened 2025-04-15
Added microservice documentation viewer

### LINKED ISSUE #16522: Endpoint directory rendering
We want to reuse the markdown renderer from lenasys to render .md documentation of the micro services into html.

If a service such as courseed has documentation and description in .md format we can get this returned in md format or html format.

for example

microservicedirectory?name=course&render=html

or

microservicedirectory?name=course&render=md

md should be default if no render parameter is given

microservicedirectory?name=course

Related to issue "Microservice Directory Database #16538 ".

a23bensh:
This fixes #16522 and serves as a simple documentation viewer in MD (with the possibility of viewing in HTML). For expansion more files will need to be added not just 'testMarkDownFile.md' and for the HTML rendering support to work you would need to uncomment the segment mentioned in the code.

### DIFF
--- a/DuggaSys/microservices/endpointDirectory/directoryRendering.php
+++ b/DuggaSys/microservices/endpointDirectory/directoryRendering.php
@@ -1,0 +1,26 @@
+<?php
+header('Content-Type: text/plain');
+
+// Get parameters
+$serviceName = $_GET['name'] ?? '';
+$renderMode = $_GET['render'] ?? 'md';
+
+// Check if service name was entered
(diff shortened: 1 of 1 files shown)


a20artje (changes requested):
Code works as expected, but doesn't use the search queries from nameQuery, description search etc..

Is it possible to either:
Use those search results in this file with the render function already defined in endPointDirectory.php.

or

Add the functionality under everything in endPointDirectory.php.

So we can use the search queries already defined in those files to find the correct microservices.

PR #16633, opened 2025-04-15
G2 2025 v2 #16566

### LINKED ISSUE #16566: Find connection between Javascript and PHP
We need to find the file which calls the microservices(PHP) from Javascript. It is important to document the connection between php and javascript since it is not completely clear. We need this information so that can choose an approach on how to re-engineer the microservices. Either by commenting, but maybe easier in a .md document.

d22marwi:
I have updated the document further with information about the connections. There are many more, but I will continue with the documentation after the merge.

partly fixes #16566


a23gabli (approved):
Documentation looks good and information provided is correct.

Issue #16597, opened 2025-04-15
UML Inheritance triangle hitbox


a23aleva:
The standard collision detection for the UML Inheritance is creating a hitbox with the _width x height_ of the element which creates borders outside of the object itself - like a square or rectangle. The problem is that this makes the element overlap other objects even though the collision happens outside the visible boundaries of the element (the edges of the hitbox). The collision detection is calculated within the _helper.js_ file in the _entityIsOverlapping_ function.

https://github.com/user-attachments/assets/ab0e68bc-5c16-461f-aa06-0954c48db750

PR #16640, opened 2025-04-15
Better collision for UMLRelation #16597

a23miksk:
A small change where UMLRelation is added to a if statement that calculates collision better than before. The UMLRelation collision box is till not a perfect triangle but will now be a bit more precise.
image


a23eliax (approved):
a small change but made a good difference, works as expected!

Issue #16152, opened 2024-05-22
Enable retry on fail [newinstaller]


a22willi:
When the installer fails, it sends a json string containing the error message and what step it failed on. 

The installer can, using this be restarted from the step it failed on by sending parameter "starting_step" into the installer.

a23dilel:
I was thinking if installer failed to install the lenasys then a button should appear retry-button which does reset all steps and go to first step in the begining.

PR #16641, opened 2025-04-15
Added a button when installer failed to install lenasys

a23dilel:
I added a button when installer failed to install lenasys. This fixed issue #16152.


d22marwi (approved):
The button works as it should.

Issue #16536, opened 2025-04-14
UML final/initial state lines faulty attachment


a23aleva:
When adding lines to initial and final state elements the line only looks correct when the elements are vertically located (and even then it looks a tiny bit scuffed). Attaching lines should target the border of the circle states from all angles. This problem is mostly visible when using an UML final state since you can see the line crossing the empty inner ring. Since the initial state is just a dark spot the line's anchorage is never fully viewable.

![Image](https://github.com/user-attachments/assets/0994dfd2-48fc-4ed6-8078-07fa0603320c)

https://github.com/user-attachments/assets/b51222de-e866-4a38-9f7a-4c7af0115af3

PR #16645, opened 2025-04-15
[Issue #16536] Fixed line anchoring to UML FinalState elements

a22nevan:
The main issue here was that getLineAttributes returned the incorrect value for t.x when the lineDirection was left, returning x1 instead of x2, which didn't align with the value pattern the other directions were returning. **This should and probably will affect all types of lines**, although it will most likely expose faulty handling of line alignment.

Added special offsets for the UMLFinalstate, making sure that the line anchors to the outer ring at all zoom levels. 

image

SDElementLines demo

Also took the opportunity to tweak alignments for the other SD elements, making the line anchor a bit into the element, instead of right on the edge, which in my opinion looked and felt nicer during usage (very easily removed if unwanted).

image

**KNOWN ISSUES:**

- Line spacing permanently changes anchor position, never reverting back to the center of the object, even if additional line is removed. Had to make sure that FinalState elements never have their anchor point changed when multiple lines were attached to them. This seems to be an issue with the line spacing in general but noticed it the most on normal lines (not just for SD elements). **Should probably be made into separate issue.**

- Non-recursive segmented lines have not been tweaked and are somewhat misaligned with finalState objects, alongside the rest of the SD elements (_this is most likely a direct result of having changed the value in getLineAttributes_). These affect more than SD elements, and seem to have a unified way of drawing the line for all elements. Seeing as other elements would have been affected by an attempt to tweak these lines, it felt out of scope for this particular issue. **Will make into separate issue if/when pull request gets merged**.

### DIFF
--- a/DuggaSys/diagram/draw/line.js
+++ b/DuggaSys/diagram/draw/line.js
@@ -44,7 +44,7 @@ function drawLine(line, targetGhost = false) {
 //Looks if the lines have gotta an index value from the function getLineAttrubutes
 //If so then there are multiple lines on the same row and the offset is changed
 if (typeof line.multiLineOffset=== 'number' && typeof line.numberOfLines === 'number') {
-    const lineSpacing = 30; //Can be changed to change the spacing between lines
+    const lineSpacing = 15; //Can be changed to change the spacing between lines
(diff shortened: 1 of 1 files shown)


b23ebbst (approved):
It works like it is proposed in the issue. The UML final state anchors as expected and the anchor moves as the element is moved around (as seen in the video/gif in the issue).

Issue #16612, opened 2025-04-15
Remove view buttons from mobile screens.


b23sebth:
The different view buttons are unnecessary for mobile platforms and should therefore be removed. Remove them from view, but only on mobile versions of the system. 
![Image](https://github.com/user-attachments/assets/dbfb61a1-d3da-40e4-98a3-f005b318dd55)

HGustavs:
Is implementing these buttons using react a separate issue?

PR #16648, opened 2025-04-15
G3 2025 v2 issue #16612 remove view buttons from mobile screens

b23sebth:
I have removed the three different view buttons from the DOM for screens smaller than or equal to 570px. I have also moved the related styling to the external stylesheet and improved the semantics of the elements.
# Before:
image
# After: 
image


b20isato:
I made fast "test" before week merge and it looks fine

PR #16667, opened 2025-04-16
G2 2025 v3 #16520 a23oskme

### LINKED ISSUE #16520: Endpoint directory parameter query
We want to be able to query the endpoint directory based on a part of the name of any parameter that is used by that service

a23oskme:
Made a query for searching for parameters in the microservice documentation database. Fixes #16520.
Although, when the database is finished the query might need to be updated (small change)


a23bensh (approved):
Tested with available parameters (userID) and code seems to work as intended, looks good!

PR #16674, opened 2025-04-16
Updated template for Microservices Documentation

a23wilgr:


a23oskme (changes requested):
Works well to take data from. 
I would however like to add one small thing. It would be nice to clarify the parameter part, since there could be more than one parameter for each microservice. 
If you put parameter, type and description before the text (like a small header) it will be easier to extract data and will be less prone to errors. (the " - Parameter: " part etc)

This example shows how it could look if there were two parameters in one microservice: 

## Input Parameters
- Parameter: userID
   - Type: int
   - Description: This parameter is used for this

- Parameter: otherParam
   - Type: String
   - Description: This parameter is used for something else
   
   
   Otherwise its a good looking template!

a23louma:
I will merge this and create a new issue for the things that have to be fixed.

PR #16677, opened 2025-04-16
Document inverse dependencies for updateVisibleListEntries

b23johgu:
No inverse dependencies found


a23axehe (approved):
Looks good, I could not find any missed inverse dependencies.

Issue #16658, opened 2025-04-16
Copy pasting is no longer working in the diagram site.


a23miksk:
After this [merge](https://github.com/HGustavs/LenaSYS/commit/0068ebb4b3046108647d6fea6509ffa81041a475#diff-a76e6303d7d6036644f76c4f18917b5df830747f146abfe0f86a6a06fa8fa9c2) it no longer seems possible to copy paste elements using CTRL+C and CTRL+V.

PR #16679, opened 2025-04-16
Fixed: Copy pasting is no longer working in the diagram site.

a23gabst1:
The problem was fixed by explicitly ensuring that the COPY and PASTE events do not have meta as true but false.
Fixed this issue #16658


a23gabst1:
Message for the reviewers: Try out the other keybinds that exist other than COPY and PASTE to ensure that other keybinds works. You can find the keybinds in constants.js and how they are used in diagram.js

a23kevre (approved):
Copy with CTRL+C and Paste with CTRL+V is working after this change, and it didn't affect other keybinds.

b23danhe (approved):
Works as a charm! Tested on a Mac in firefox and safari.

PR #16697, opened 2025-04-16
issue#16659

a23petny:
I found no inverse dependencies


a23edvto (approved):
Did not find any includes for readGroupValues or readUserDuggaFeedback, so it is approved

PR #16705, opened 2025-04-16
Documented inverse dependencies.

a23gabli:
Fixes issue #16663


a20artje (commented):
Looks good, found the same

Issue #16647, opened 2025-04-15
Line spacing for multiple lines permanently offsets line anchor points


a22nevan:
In the drawLine function there's an if-statement checking if there are multiple lines connected to the same side of an element. If there are, they will all be offset to accommodate each other. This offset seems to be permanent however, and doesn't revert if the connected elements are moved to another side. On top of this, it alters the offset for **both** the anchor point to element that has multiple lines anchored to it, as well as the one for element that's on the other end of the line. This doesn't seem to be revertible **without** either undoing everything up to the point that the offset got added or redrawing the line completely. This affects most, if not all lines and elements.

**Example 1 - Segmented line to UML Class:**

1. Classes before offset. Note that the lines are anchored to the center of the side for each respective object.
![Image](https://github.com/user-attachments/assets/8d0668a8-b838-4e0b-9807-7499357a815c)

2. Classes after having moved to the same side. Note the anchor points being offset from the center of Class0 (and Class1/Class2, not as noticeable though).
![Image](https://github.com/user-attachments/assets/a889c41b-3b34-4b22-b998-0c994e5e9cfe)

3. Classes after having been moved back to their original side. Lines are still offset for both anchor points.
![Image](https://github.com/user-attachments/assets/260d8836-f8b5-4cd0-a112-22f0d2db2ac8)

**Example 2 - Straight line between ER elements:**

1. Before
![Image](https://github.com/user-attachments/assets/ad87a27a-17f5-4a6c-9a47-0b26fac66544)

2. During
![Image](https://github.com/user-attachments/assets/95cdce7d-21fa-434d-a758-7dbecaff7b30)

3. After
![Image](https://github.com/user-attachments/assets/016e402e-5981-4309-9ae6-b2ada04bd64a)

PR #16706, opened 2025-04-16
Lines should now move towards center when neighbor lines are removed

a23carpa:
Solution to issue: [#16647](https://github.com/HGustavs/LenaSYS/issues/16647)
To have a line being able to have different offset on each side is done here by adding new properties to the line object inside the checkAdjacentLines and then inside drawLine give different offset defendant on these properties. checkAdjacentLines checks if i object have lines going out or coming in, and dependent on that set values to these new properties so the right x and y gets increased.

[Screencast from 2025-04-16 14-07-19.webm](https://github.com/user-attachments/assets/5957393a-212d-4554-92b2-7768b37281a3)

### DIFF
--- a/DuggaSys/diagram/draw/line.js
+++ b/DuggaSys/diagram/draw/line.js
@@ -41,19 +41,31 @@ function drawLine(line, targetGhost = false) {
     }
 
 
-//Looks if the lines have gotta an index value from the function getLineAttrubutes
-//If so then there are multiple lines on the same row and the offset is changed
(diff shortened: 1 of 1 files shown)


a23miksk (approved):
This change seems to fix the problem in the issue where multiple relations didn't adjust graphically when elements were moved around on the screen. The lines now jump back to where they are supposed to be. Approved.

_Before_

https://github.com/user-attachments/assets/87ca22cb-001b-42e3-933e-03e5c8442fde

_After_

https://github.com/user-attachments/assets/f7007cd5-ed3f-48d3-ba6b-8c1ae2368d3b

Issue #16704, opened 2025-04-16
The keybindings for zoom-in, zoom-out and zoom-reset does not work


a23gabst1:
The problem is that you cannot use the key combinations ctrl + "+", ctrl + "-" and ctrl + 0, these key combinations are for zoom-in, zoom-out and zoom-reset.

PR #16707, opened 2025-04-16
Fixed: The keybindings for zoom-in, zoom-out and zoom-reset does not work

a23gabst1:
Fixed the issue #16704 by setting the meta-prpperty in the objects to false instead of true.


a23carpa (approved):
This fixes the issue. You can now use the function of zooming with the stated keybinds.
https://github.com/user-attachments/assets/080480a6-cb9d-4e81-bf84-b52332a869ad

PR #16708, opened 2025-04-16
Fixed microservice endpoint output format

### LINKED ISSUE #16522: Endpoint directory rendering

a23bensh:
Fixes #16522, now you get a markdown/html (html conversion can be added later) output of the relevant microservice documentation (from the database). Also removed debug outputs for less clutter and easier testing. The searchable parameters include: name, description and parameter, these are the ones that should be used when testing. More parameters can be added later if requested.

### DIFF
--- a/DuggaSys/microservices/endpointDirectory/directoryRendering.php
+++ b/DuggaSys/microservices/endpointDirectory/directoryRendering.php
@@ -1,0 +1,36 @@
+<?php
+header('Content-Type: text/plain');
+
+$search = $_GET['name'] ?? $_GET['description'] ?? $_GET['parameter'] ?? '';
+if (empty($search)) {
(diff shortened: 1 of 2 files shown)


a23axehe (approved):
Tested all parameters available. Working as intended and solves the issue. Looks Good.
We might want to create a new issue to remove the old "endpointDirectory.php" if it is not being used anymore?

PR #16711, opened 2025-04-16
16617 fix the tooltip for scaling panel

a23vinge:
he task bestowed upon me is now complete.
I have removed right: 0; from the tooltip, as it caused the text to hide behind the toolbar. In its place, I added left: 0;, which ensures the tooltip appears correctly—spaced out and visually distinct from surrounding elements. If change is demanded for better UI then i shall make it so.

before:
Skärmbild 2025-04-16 142852

after:
image

The code awaits your judgment.

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -784,7 +784,7 @@
     padding: 5px 0;
     pointer-events: none;
     position: absolute;
-    z-index:1;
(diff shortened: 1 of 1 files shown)


a23gabst1 (approved):
The tooltips meant for the scalings are fixed and works as intended! They are well positioned and visibly for the user.

PR #16713, opened 2025-04-16
Fixed the tooltip for options panel

a23kevre:
Update diagram.css
Changed the position of options tooltip.

Before:
image

After:
image


a23kevre:
Might override this PR #16711 since i changed the same file before this PR was merged.

a23gabst1 (approved):
Works as intended! The tooltip for the options panel is fixed by making the whole tooltip visible by altering the position of it.

Issue #16694, opened 2025-04-16
Add a title section about how to setup XAMPP in README.md


a23louma:
Add a title section in LenaSYS/README.md about how to setup XAMPP, it should be easy to follow the steps for this before installing LenaSYS. Another issue was created to add docker instructions in README.md: #16690 .

PR #16714, opened 2025-04-16
Added instructions on how to install XAMPP in README.md

b23johgu:
README.md updated with step by step instructions on how to install XAMPP on the computer. Added neccesary information from the [Wiki](https://github.com/HGustavs/LenaSYS/wiki/Configuring-web-server-for-LenaSYS-installation) to keep all instructions in the same place.

Fixes issue #16694


a23danme (commented):
Overall looks good, clean instructions, some thoughts:

.2 Developer Files, I'm not sure what is meant by that. To select which file it should be installed  in?
.5 Should be Extract instead of Expand?

b23johgu:
Added an image for clarification on developer files and edited the typo. It should be extract!

Issue #16348, opened 2025-04-09
White space on mobile devices


HGustavs:
There seems to be a lot of white space around the on-screen elements on mobile platforms.. 

Especially on the course editor and code viewer we want the white space to be more or less removed. We want no padding around the screen edges or between the on-screen course elements.

We also want to reduce the padding around the course elements on mobile platforms, especially on smaller screens we ant the padding arond the elements to be smaller so that more items fit on the screen at the same time (on mobile devices)

b23sebth:
https://github.com/HGustavs/LenaSYS/pull/16586

PR #16719, opened 2025-04-16
G3 2025 v2 issue #16348 white space on mobile devices

b23sebth:
This is a retread of a previous PR that did not get implemented before the weekly merge. 

## What's changed:
Unnecessary white space have now been removed. I have left the white space between the header, course title and load dugga heading. I have only changed styling for mobile screens (width <= 570px). A separate issue for white screen on desktop will need to be created. I have also changed some HTML and moved inline styling to the external stylesheet. More courses is also visible on mobile devices now per the request, although this feels a tad bit excessive for mobile platforms. 

## Future issues:
Future issues include reviewing the different SectionList ID's, small course names and egregious use of inline styling as well as creating tables correctly.

# Before:
image

# After: 
image


b20isato:
Will merge since it was reviewed previous weekly merge (and if it breaks it will only be some wrong leftshift on "mobile" view)

Issue #16669, opened 2025-04-16
Multiple lines does not work properly when dragging


a23miksk:
When multiple relations exists between two elements, and one is dragged to the left of the other element, the lines get cut off and the mouse gets stuck.

_Bug in action_
https://github.com/user-attachments/assets/b2f6d88f-b443-4c6f-b964-e5d559225004

PR #16723, opened 2025-04-16
[#16669] Fixed issues with lines not updating when moving to left side

a22nevan:
This issue seems to have stemmed from duplicate code being added during the merge to the weekly branch. The function getLineAttribute had been updated in between the opening of the merged branch and the merging of the branch, which caused duplicate code to get added that didn't align with how the updated function worked, breaking before it could return the correct values for the left side, causing all lines to stop updating if lineDirection.LEFT was true. 

That issue may mostly have been my fault, since it was my code that got added that messed up the function and was outdated - most likely due to not fetching origin or something of the sort, so apologies for that.

The lines should now work as intended again.

fixedlines demo

### DIFF
--- a/DuggaSys/diagram/draw/line.js
+++ b/DuggaSys/diagram/draw/line.js
@@ -399,16 +399,9 @@ function getLineAttrubutes(f, t, ctype) {
             return [f.cx, f.y2, t.cx, t.y1, offset];
 
         case lineDirection.LEFT:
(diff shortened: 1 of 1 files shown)


b23danhe (approved):
This solution works as promised. Tested with all object and the lines now updates when moving an attached object to the left side of the other object.

PR #16725, opened 2025-04-16
Changed the fab button to a react component

a23jesja:
Jag har omvandlat FAB i courseed.js-filen till en React-komponent. Här är vad jag gjort:

Skapade en React-komponent för FAB-knappen som ersätter den traditionella HTML-knappen:

- Komponenten behåller samma funktionalitet och stil som originalknappen
- Lade till korrekt tillgänglighet (role, aria-label, tangentbordsnavigering)
- Följde React-kodstandarden från den bifogade dokumentationen med PascalCase för komponentnamn och camelCase för 
funktioner

Modifierade returnedCourse-funktionen för att:

- Lägga till en dedikerad container-div för React-komponenten
- Dynamiskt ladda React och ReactDOM från CDN
- Rendera React FAB-komponentet när sidan laddats
- Inkludera en fallback till traditionell HTML om React-laddningen misslyckas

Gjorde newCourse-funktionen globalt tillgänglig genom att lägga till den i window-objektet så att React-komponenten kan anropa den: "window.newCourse = newCourse;"

### DIFF
--- a/DuggaSys/courseed.js
+++ b/DuggaSys/courseed.js
@@ -552,169 +552,156 @@ function createVersion() {
 // Renderer
 //----------------------------------------
 
-function returnedCourse(data) {
-	versions = data['versions'];
-	entries = data['entries'];
-	if (data['LastCourseCreated'][0] != undefined) {
-		LastCourseCreated = data['LastCourseCreated'][0]['LastCourseCreatedId'];
-		localStorage.setItem('lastCourseCreatedId', LastCourseCreated);
+	function returnedCourse(data) {
+		versions = data['versions'];
+		entries = data['entries'];
+		if (data['LastCourseCreated'][0] != undefined) {
+			LastCourseCreated = data['LastCourseCreated'][0]['LastCourseCreatedId'];
+			localStorage.setItem('lastCourseCreatedId', LastCourseCreated);
 
-	}
+		}
 
-	var uname = document.getElementById('userName').innerHTML;
+		var uname = document.getElementById('userName').innerHTML;
 
-	// Fill section list with information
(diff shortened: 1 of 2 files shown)


EdvinWerner (changes requested):
My interpretation of the coding standard for react is that this button should be created as it´s own component inside of the shared component folder

a23jesja:
I've now moved the FAB button to a separate React component as requested. 

The changes include:

- Created a new FabButton.js component in the Shared/components folder
- Implemented it as a class component that follows the React coding standard
- Updated the returnedCourse function in courseed.js to dynamically load and use this external component
- Made sure the component has the same functionality and accessibility features as the original

b20isato (changes requested):
Even tho that the feature seem to work so is there so much unnecessary indentation changes that make it hard to make an actual review whether this complies or not with the standards. For example: is the inline css new? or was it already there.

And also its harder to see the logic flow of how the html/css would be constructed.

a23jesja:
I'd like to clarify that I didn't add any new inline CSS - all styling in the component was already present in the original inline implementation of the FAB button in the returnedCourse function. I only moved the existing functionality to a separate component.
Regarding the indentation changes, these may have occurred when moving the code. I understand this makes it harder to review the actual logical changes.
The real changes I made were:

- Creating a separate FabButton.js component in Shared/components
- Modifying the returnedCourse function to load and use this component instead of the inline-defined version

If you prefer, I can create a new, cleaner PR where im more careful about preserving the existing indentation to make it easier to see the actual changes.

b20isato:
sure, why not?

PR #16737, opened 2025-04-17
fixed so weeks are 'taller' in sectioned.js

a23fanro:
And changed so the text in css for #svimlaneSVG is 1em, which will make the font change size dynamicly based on parent element. 
(Also removed an unnecessary <style>-tag in style.css since it gave me an ulser)

- Need to make a new issue for fixing on mobile! 

Old swimlanes
image

New swimlanes
image


EdvinWerner (approved):
Looks fine to me, it does however just like the author mentioned need to be made mobile friendly but I belive that this is a separate issue

PR #16739, opened 2025-04-17
Corrected the documentation of reorderListEntries/updateListEntryOrder

a23wilgr:
Previous documentation said updateOrder should replace reorderListEntries, but it is supposed to be updateListEntryOrder. Changed the documentation from updateOrder to updateListEntryOrder, since updateOrder was incorrect.

Also, reorderListEntries and updateListEntryOrder contain the exact same code, reorderListEntries should be completely replaced with updateListEntryOrder and then removed (?).


a23axehe (approved):
Changes to documentation looks ok.
I will create an issue to research whether we can remove one of the microservices since they are identical. However i believe this is something we will have the answer to when research for inverse dependencies are done

Issue #16726, opened 2025-04-17
Delete the old file endpointDirectory.php


a23louma:
This file is not used anymore and can be removed.

PR #16740, opened 2025-04-17
Delete old file endpointDirectory

a23axehe:
The file endpointDirectory was not needed anymore because of the new file directoryRendering.php. The purpose of the files are them same only the new one formats the output. Fixes Issue #16726


a23bensh (approved):
Tested locally and everything seems to work as intended without endpointDirectory, directoryRendering works, looks good!

PR #16741, opened 2025-04-17
Documented inverse dependencies for clearGitFiles and fetchOldToken.

### LINKED ISSUE #16729: Document inverse dependencies: clearGitFiles_ms.php AND fetchOldToken_ms.php
We need to go through all the microservices and list their inverse dependencies. This is so we later can remove the includes. After finding all the inverse dependencies for **these two files** in /gitCommitService, add it to the documentation-file **with file path**: 
)

a23bensh:
Fixes #16729, no found for fetchOldToken, one for clearGitFiles.


b23johgu (approved):
Looks good, found the same!

Issue #16742, opened 2025-04-17
Fix so you do not get a lot of errors in the when loading up the example diagrams


a23gabst1:
The problem is that when you load up and use the box selection on of the example diagrams, it generates a lot of error messages in console. 

![Image](https://github.com/user-attachments/assets/e8bf272f-ca75-4af0-be9d-7bc33b1ff927)

HGustavs:
This can be important for the acceptance testing

PR #16743, opened 2025-04-17
Fixed: Fix so you do not get a lot of errors in the when loading up the example diagrams

a23gabst1:
The solution was to remove a misplaced comma in the iconPoly function so the points for polyline dont get any unexpected attributes and it follows the correct syntax for declaring the points.
Fixed this issue: #16742

image


a23eliax (approved):
This solution works, the error is gone!

image

PR #16762, opened 2025-04-17
Documentation inverse dependencies for insertToFileLink_ms.php and insertToMetadata_ms.php

a23axehe:
Documented inverse dependencies for insertToFileLink_ms.php and insertToMetadata_ms.php. None found. Fixes issue #16753.


a23bensh (approved):
Found the same, looks good!

Issue #16631, opened 2025-04-15
The UML example diagram uses IE cardinalities instead of UML cardinalities


a23thefo:
When loading the UML exemple diagram it gives an IE diagram. We already have a IE example diagram there for make the UML example diagram actually be a UML diagram
This is wrong and should be change to the correct cardinalities type (arrows and text ex. 1..*) 

## How the cardinalities should look
![Image](https://github.com/user-attachments/assets/561e6f14-4f98-4f0f-b799-5ba5a6721142)

## How they look now
![Image](https://github.com/user-attachments/assets/1095b071-f8f5-4fc1-bfd8-3f8ed17cf56c)

HGustavs:
We should perhaps split this issue into two issues one for example ie diagram and one for example class diagram. 

Either / Both need some changes.

a23emijo:
> This may be an ie diagram and not a uml diagram

This is actually a UML diagram because it has the box under the variables with the functions for the class

PR #16776, opened 2025-04-17
[#16631] Fixed UML example diagram using icons instead of cardinalities

a22nevan:
The solution to this problem was to change the JSON file for the UML Mockup diagram, removing the text from start- and endIcon and adding the corresponding cardinality to start- and endLabel.

The cardinalities should now be in the UML format and values should be correct as well.

**Known Issue:** I personally had some issues getting the JSON file to reload with the new updated data, which seems to be a browser cache issue or something of the sort, not properly reloading the JSON file since it already existed in the cache. I managed to get the new data to show by opening up the website in incognito mode, and also outside of incognito after restarting my pc with some additional hard browser refreshes after that.

**UML Mockup diagram with updated cardinalities:**
image

### DIFF
--- a/DuggaSys/JSON/UMLDiagramMockup.json
+++ b/DuggaSys/JSON/UMLDiagramMockup.json
@@ -205,10 +205,10 @@
             "time": 1681378942204,
             "type": "UML",
             "label": "",
-            "startLabel": "",
-            "endLabel": "",
-            "startIcon": "1-M",
-            "endIcon": "1"
+            "startLabel": "1..*",
+            "endLabel": "1",
+            "startIcon": "",
+            "endIcon": ""
         },
         {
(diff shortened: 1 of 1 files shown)


a22nevan:
I think the way the website handles browser caching for the JSON files should be looked at in a separate issue. It's a very niche issue that will rarely be noticed, but could cause some problems for people trying to update the already existing ones.

a23aleva (approved):
The changes to the JSON file apply like intended and cardinalities are now matching the UML class diagram type instead of IE. Personally I had no problem swapping between the different example diagrams when I tested. Seems like a clear improvement to how it was before!

https://github.com/user-attachments/assets/7277856a-9c62-4a19-8af2-4c8f99d3bd34

PR #16777, opened 2025-04-17
Issue#16731

### LINKED ISSUE #16731: Document inverse dependencies: getCourseID_ms.php AND insertIntoSQLite_ms.php
We need to go through all the microservices and list their inverse dependencies. This is so we later can remove the includes. After finding all the inverse dependencies for **these two files** in /gitCommitService, add it to the documentation-file **with file path**: 
)

a23petny:
found 1 inverse dependency for insertIntoSQLite and 0 for getCourseID

fixes #16731


b23johgu (approved):
Looks good, found the same results.

Issue #16754, opened 2025-04-17
Document inverse dependencies: highscoreservice_ms.php AND retrieveHighscoreService_ms.php


a23louma:
We need to go through all the microservices and list their inverse dependencies. This is so we later can remove the includes. After finding all the inverse dependencies for **these two files** in /highscoreService/, add it to the documentation-file **with file path**: 
[This file!](https://github.com/HGustavs/LenaSYS/blob/G2-2025-v3/DuggaSys/microservices/Microservices_inverse_dependencies.md 
)

HGustavs:
I think it may be useful to check if this file is still in use or deprecated.

It may have very limited use.

PR #16779, opened 2025-04-17
Documented inverse dependencies: highscoreservice_ms.php AND retrieveHighscoreService_ms.php #16754

a22linre:
Document inverse dependencies: highscoreservice_ms.php and retrieveHighscoreService_ms.php, found one inverse dependency for retrieveHighscoreService_ms.php. 

Also discovered that highscoreservice_ms.php is not in use anymore. It's commented out in the code in dugga.js.(highscoreservice.php is used instead).

Fixes issue #16754


d22marwi (approved):
Correct

PR #16784, opened 2025-04-17
Fixed activation element in sequence diagram always is above actor

a23kevre:
The actor element had a zLevel = 2; and the activation element didn't have any zLevel.
I added zLevel = 3; to the activation element to always be above the actor.


a23vinge (approved):
The issue has been fixed and there seems to be no problems with the code. The adding of zlevel has made it possible to add the sequence activation over the sequence lifeline. 

There could be a possible add for a comment just describing the purpose of the zlevel because it does change the relationship between the sequence activation and the sequence lifeline. 

but beside that the issue has been fixed as you can see on the image the sequence activation is over the sequence lifeline.
image

PR #16786, opened 2025-04-17
Updated diagram.css

d22abdha:
Increased toolbar width and icon size for better visibility on large screens

Before
before
After
after

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -774,7 +774,7 @@
 }
 .zoomToolTipText {
     bottom: 40px;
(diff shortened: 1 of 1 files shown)


a23thefo:
Yo before i give this PR out for review. Could you add a margin similar to the one in issue #16711 . This would allow us some consistency of the margins on the tooltips :)

### Meaning the margin the tool tip has from the zoom panel here
image

a23thefo (approved):
Looks great the margin is consistent with other tool tips in the diagram view and is much easier to read

Issue #16542, opened 2025-04-14
Fix overlapping elements with titles using copy/paste


a23thefo:
Copy pasting elements mostly check if it is overlapping with an element except for if you hover slightly above an element with a title. Then you are allowed to paste an element on top of that element.

How it looks now:

https://github.com/user-attachments/assets/723c6b9e-b821-434c-afa4-37e58d0cd0a4

PR #16788, opened 2025-04-17
Fix for copypasting elements with labels #16542

a23miksk:
The program now correctly prohibits the user from copy pasting elements on top of each other if the element has a label.

_before_

https://github.com/user-attachments/assets/da688d1e-5b23-426f-8cdc-66fb3a023b19

_after_

https://github.com/user-attachments/assets/0519878a-6109-4a59-a0df-caabf65a9a22


a22abdjo (approved):
This solution now prohibits the user from copy and pasting elements on top of eachother (if the element has a label). Therefore the solution provided is approved after testing.

PR #16789, opened 2025-04-17
Example diagrams now show lines again

b23ebbst:
Removed part of the line.js code that was seemingly used as debugging previously, but was causing the issue with not fully showing the example diagrams. Now all example diagrams show lines as they are supposed to!

image


a23thefo:
Looks good! The lines are visible when you load in the example diagrams and are easily readable! 10/10 gj!

PR #16796, opened 2025-04-22
Inverse dependencies: getShowDugga_ms.php & loadDugga_ms.php

b23johgu:
Updated document with found inverse dependencies.

loadDugga had an inverse dependency inside of an if-statement in retrieveShowDuggaService_ms.php.  

Fixes issue #16793


a22wilas (approved):
I can approve this pull request!
No inverse dependencies could be find for the getShowDugga_ms.php.

Could find the one inverse dependence in retrieveShowDuggaService.

bild
Picture shows the code for the inverse dependence.

PR #16797, opened 2025-04-22
Documented inverse dependencies for saveDugga and updateActiveUsers

### LINKED ISSUE #16795: Document inverse dependencies: saveDugga_ms.php AND updateActiveUsers_ms.php
We need to go through all the microservices and list their inverse dependencies. This is so we later can remove the includes. After finding all the inverse dependencies for **these two files** in /showDuggaService, add it to the documentation-file **with file path**: 
)

a23bensh:
Fixes #16795, no inverse dependencies found for either.


a23petny:
I could not either find any inverse dependencies. I will approve this change.

Issue #16588, opened 2025-04-15
Recursive line overlapp when multiple lines exist


a23carpa:
With the added functionality of having multiple lines on one side, a new problem of the recursive line was created.
Change it so the recursive line no longer overlaps with the additional lines. Possibly move it to the side as the paint example, but that's not a requirement. 

As of now:
![Image](https://github.com/user-attachments/assets/dffa96b4-dfac-4e3c-af2b-1139714c720c)
Example:
![Image](https://github.com/user-attachments/assets/9875dab3-dfcf-4430-b762-48cc6c910d2f)

PR #16798, opened 2025-04-22
Fix for the multiple lines overlapping with the recursive line 

a23carpa:
Solution for issue [#16588](https://github.com/HGustavs/LenaSYS/issues/16588) 
Moved the recursive line to now instead be on top of the element and in the right corner.
The lines should now also always stick to the corner when the element is re-sized to keep out of the way.

[Screencast from 2025-04-22 10-13-21.webm](https://github.com/user-attachments/assets/537eff61-bef2-4ab7-9645-b4b12945789d)

### DIFF
--- a/DuggaSys/diagram/draw/line.js
+++ b/DuggaSys/diagram/draw/line.js
@@ -169,14 +169,14 @@ function drawLine(line, targetGhost = false) {
         //Arrow/icon location dependant on element length, so its always in the top right corner of the element.
         const length = 40 * zoomfact;
         const elementLength = felem.x2 - felem.x1;
-        let startX = felem.x1 + elementLength - length * zoomfact;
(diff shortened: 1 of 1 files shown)


a22nevan (approved):
Works as shown in the given video. Recursive lines now sit nicely at the top right corner of an element and different zoom levels, as well as resizing of the elements have all been accounted for. Icons align properly as well. 

Screenshot from own test. Lines are working as intended:
image

a23aleva (approved):
Functionality looks enhanced and recursive lines are now convenient to use. Added cardinalities and symbol elements works as expected. The line has a minor sprout artifact where it is attached to the middle of the diagram element but this seems a minor defect.

https://github.com/user-attachments/assets/757a938a-60b7-41b4-94ba-f05bd6e862e9

PR #16799, opened 2025-04-22
Document inverse dependencies: the two files in /resultedService #16792

### LINKED ISSUE #16792: Document inverse dependencies: the two files in /resultedService
We need to go through all the microservices and list their inverse dependencies. This is so we later can remove the includes. After finding all the inverse dependencies for **these two files**(getUserAnswer_ms.php AND retrieveResultedService_ms.php) in /resultedService , add it to the documentation-file **with file path**: 
)

a22linre:
Documented inverse dependencies for getUserAnswer_ms.php and retrieveResultedService_ms.php. Found one inverse denpendencies for retrieveResultedService_ms.php.

Fixes #16792


a23bensh (approved):
Found the same, looks good!

Issue #16657, opened 2025-04-16
Change new microservice documentation template for compatibility with database


a23louma:
The new template in /microservices/Microservices Documentation/ appeared to be difficult to get data from, especially the tables. Make some research which structure can make this easier and change the template. The tables in text format looks like the image below and it's difficult to ignore all characters with regex.

Related to issue #16538 and #16537

![Image](https://github.com/user-attachments/assets/795a298a-d678-469f-94c2-edb09354c791)

a23wilgr:
Example of template for the documentation:

# Name of file/service

## Description
*Description of what the service do and its function in the system.*

## Input Parameters
*Parameters will be described in tables for easier readability*

- Parameter
   - Type
   - Description

- Parameter
   - Type
   - Description

## Calling Methods

- GET
- POST
- etc.

## Output Data and Format
*Output Data will be described in tables for easier readability*

- Output
   - Type
   - Description

- Output
   - Type
   - Description

## Examples of Use
`CODE`

### Microservices Used
*Includes and microservices used*

b23johgu:
Looks good! In my research I found a format that's called YAML, a data serialization language, that seems to be in this list-format as well. It should make it easier to get the data but still have high readability. So I think this would work well.

b23johgu:
It could look something like this in YAML.

# Service: <name-of-service>

## Description
Describes what the service does and its role in the system.

## Input Parameters
```yaml
input_parameters:
  - name: exampleid
    type: string
    description: Example ID Description
```
## Calling methods
```yaml
methods:
  - GET
  - POST
  - etc
 ```
## Output Data and Format
```yaml
output_data:
  - name: exampleid
    type: string
    description: Example ID Description
```
## Examples of Use
```yaml
GET /example?exampleid=123
```
## Microservices Used
```yaml
dependencies:
  - <name-of-microservice>
```

PR #16804, opened 2025-04-22
G2 2025 v3 #16635

d22marwi:
Finished the documentation in issue #16566 and #16635

### DIFF
--- a/DuggaSys/microservices/microservice_connections.md
+++ b/DuggaSys/microservices/microservice_connections.md
@@ -986,34 +986,77 @@ retrievevers.php
 ----------------------------------------------------------------------------------------------
 # Name of file/service
 
+sectioned.js
+Function getStudents
+
 ## Description
 *Description of what the service do and its function in the system.*
+
+This function retrieves student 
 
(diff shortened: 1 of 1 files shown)


b23johgu (commented):
Documentation looks ok but the template used is the old one, didn't we decide to use the one in issue "Change new microservice documentation template for compatibility with database #16657" ?

The headline "**Name of file/service**" maybe should be the actual name of the service to make it easier to find specific services.

The descriptions from the template should be removed to make it more clean and easier to read e.g.:
*Description of what the service do and its function in the system.* 
*Parameters will be described in tables for easier readability*

a23louma:
I think it's fine for this documentation to not follow the template as the compatibility with database is not necessary for this, it's only to see how the communication is between javascript and PHP.

a22linre (commented):
Looks good! Just a question, are we supposed to double check that the connections between JavaScript and PHP really exist, or mainly check that the documentation looks right?

a23louma:
For now only the documentation, it would be too much to check that the connections really exist I think.

PR #16815, opened 2025-04-22
G2 2025 v3 #16602 a23oskme

### LINKED ISSUE #16602: Update Microservice Directory Database installation script
The installation script created in "Microservice Directory Database #16538." needs to be updated to scan files for a different md-file structure. 
At the moment it scans for a structure that was used in examples.

a23oskme:
Changed the script that takes data from microservice documentation and stores it in a SQLite database. 
The previous version only took data from an example markdown file and stored in the database. This new version follows the established structure of how microservices should be documented. Fixes #16602 

This new version also includes some changes in the database, the previous version only had one table for microservices. The new version still has the table for microservices, but also has a table for parameters. Since there can be more than one parameter for each microservice. There is a one-to-many relationship between the microservice and its parameters, this means that the parameter takes the id (auto incremented integer) from the microservice of which it belongs to. 

I also changed the microserviceList.php which is a simple site that displays the content of the database. (This can be removed later on, its main function for now is to provide a fast way to check so that the data has been successfully inserted in the database)


a23gabli (approved):
The script functions as intended.  installEndpointDb and fillEndpointDb were executed and the microserviceList.php showed that all data has been correctly read and stored in the sqlite file.

PR #16819, opened 2025-04-22
A4 template title text size is now static

a23emisn:
The A4 template title text scaled with zoom before. This added patch makes the title text static, which effectively solves the target issue. The aim for the issue was to make the title text static, since the A4 template size was static, the fix involved copying the code that scales the A4 accordingly with the current zoom-factor, and applying it in a similar manner to the text element.

0.5x zoom:
image

2x zoom: 
image

### DIFF
--- a/DuggaSys/diagram.php
+++ b/DuggaSys/diagram.php
@@ -1090,7 +1090,7 @@
     <!-- A4 template -->
     <div id="a4Template">
         <svg id="svgA4Template">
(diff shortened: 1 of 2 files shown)


a23aleva (approved):
As you zoom in or out the A4 template text now appears to remain a static easily viewable size instead of scaling accordingly. Great fix and approved!

https://github.com/user-attachments/assets/aaca039e-adf5-4f34-8d4d-eced29a592c9

Issue #16818, opened 2025-04-22
You can't toggle the error check using its keybind


a23gabst1:
The issue here is that you cannot use the keybindings for toggling the error check which is pressing "h".

PR #16822, opened 2025-04-22
Fixed: You can't toggle the error check using its keybind

a23gabst1:
The solution was to place back the if-statement that sends the keybinds of toggle_error_check and if it is true then it executes the toggleErrorCheck() function. 

Fixed the issue: #16818


a23kevre (approved):
Does exactly what it says it should do. You can toggle the error check with key "h" and it shows.

Issue #16336, opened 2025-04-09
Bug: Selecting to delete


a23erigu:
When trying to delete you have to select the element then press the trach can not just press the trach can on the element like intended

a23erigu:
wait with this untill the button is made in react

c22isako:
Delete button only works when selecting the elements. Deletion only happens for selected items, independent on which delete-button is pressed. Issue might stem from lids only being passed based upon selections.

PR #16840, opened 2025-04-23
G3-2025-v3-#16336-delete-without-selection

c22isako:
Fixed trashcan by calling the markedItems function to select listItems before deletion. Also had to fix selection since it had some bugs when solo-trash was introduced as well as beforehand. You can now select several section-headers & only affect the right section - and solo-trash will not affect the selection. Section trashcan (marked with red) will now select everything in the section when pressed. Solo trashcan will only select the list item which it was pressed on.
bild

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -654,9 +654,6 @@ function showSaveButton() {
 
 // Displaying and hiding the dynamic confirmbox for the section edit dialog
 function confirmBox(operation, item = null) {
-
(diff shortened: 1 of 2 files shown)


c22isako:
How to test:
select course of your choice (which has duggor) -> log in -> test selection & trashcans
section-header trash and selection-box (marked in red) should select everything, or deselect everything. 
- Note that if an item is selected before pressing section-header trash, it should remove the item and not select everything for deletion.
bild

Issue #16767, opened 2025-04-17
Locate and change the JQuery used in codeviewer.js with normal JS - line 0 to line 1029


a23erigu:

a23elism:
### NOTE: Ignore all AJAX, this will be made its own issue

a23erigu:
@a23alera you have this to fix

PR #16849, opened 2025-04-23
change the JQuery used in codeviewer.js with normal JS - line 0 to li…

a23alera:
…ne 1029 #16767

Converted all remaining jQuery code to vanilla JavaScript from line 0 - line 1054, Certain parts related to AJAX were intentionally left unchanged as they are handled in a separate issue.

### DIFF
--- a/DuggaSys/codeviewer.js
+++ b/DuggaSys/codeviewer.js
@@ -112,7 +112,7 @@ function returned(data)
 	if ((retData['templateid'] == 0)) {
 		if (retData['writeaccess'] == "w") {
 			toast("warning","A template has not been chosen for this example. Please choose one.",10);
-			$("#chooseTemplateContainer").css("display", "flex");
+			document.getElementById("chooseTemplateContainer").style.display = "flex";
 			return;
 		} else {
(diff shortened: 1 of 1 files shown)


a23lisjo (changes requested):
All jquery have been changed to vanilla Javascript. But the pull request has conflicts to merge problably because more code has been added after merge with the master branch and the changes made in this pull request is not in the same place as it was. 

Something in the code is causing the programs functions to stop working. I will list them below:

**Cogwheel - change content in the code viewer**
When clicking on the cogwheel for a view, it should appear a window where an admin can change the content inside the viewer. Now nothing appears. 
image

How it should be:
image

**Zoom in/ zoom out buttons** 
When loading in for the first time and clicking on the zoom in or out buttons, nothing happens and the buttons are greyed out after clicking it once as if it's as zoomed in/out as possible. The only way to use them is to use the refrech button next to the zoom out button and then users can use zoom in/out. 
image

EdvinWerner (commented):
Nothing has changed since last review so nothing new to look at

PR #16851, opened 2025-04-23
Removed broken comment

EdvinWerner:
The issue was that the jquerycode that was responsible for the drag and drop feature had been commented out, removed the comments so now it works again, it does however need to be rewritten in javascript but that is a separate issue


a22carsv (approved):
Testing displayed no issues. Works fine, great job!

Issue #16848, opened 2025-04-23
The keybinding for toggling the keybind list is not working


a23gabst1:
The issue is that the user cannot toggle the keybind list by pressing f1 on the keyboard.

PR #16853, opened 2025-04-23
Fixed: The keybinding for toggling the keybind list is not working

a23gabst1:
The solution was to alter the keybind for toggling the keybind list by changing ctrl from false to true. So, when you press f1 + ctrl, then the keybind list should toggle. Also changed the instruction on keybind list document so the keybind matches with the list. I have also tested the remainder of the keybind. 

Fixed the issue: #16848


a23eliax (approved):
Looks good and I confirm that these changes doesn't disturb any other function

PR #16872, opened 2025-04-23
Locate and change the jQuery used in sectioned.js with normal js commands line 902 - 1388 (Revamp)

a23elism:
### This is a Re-opened PR from before Weekly Merge (org ver. G3-2025-V3)
This PR has been updated to ver. G3-2025-V4


a23elism:
### Testing of this PR will simply require a rundown of any course on LenaSYS
i.e any site that uses sectioned.php (since this uses the sectioned.js) and testing to create sections, headers etc


image


most important is creation of new sections, since these were mostly touched within lines 902 and 1388

PR #16873, opened 2025-04-23
refactor(gitCommitService): replace legacy includes with cURL-based

### LINKED ISSUE #16861: Re-engineer microservice clearGitFiles
Follow the "coding standard for removing includes" for this file in folder /gitCommitService. The coding standard can be found in /microservices/Microservices Documentation
/microserviceCodingStandard.md.

a23petny:
Removed every include_once statement from the seven *_ms.php files. Re-implemented data exchange with cURL:
  - GET requests now use CURLOPT_RETURNTRANSFER true and json_decode the response.
  - Fire-and-forget POST requests set CURLOPT_POST true and CURLOPT_RETURNTRANSFER false. Added `header('Content-Type: application/json'); echo json_encode(...);` to all receiving endpoints. Normalized `$baseURL = "https://" . $_SERVER['HTTP_HOST'];` path handling to keep dev/prod parity. Consolidated error reporting into consistent `{ status, message }` JSON objects. Deleted dead wrapper functions and ensured every script exits immediately after echoing JSON to prevent accidental output.

fixes #16861

### DIFF
--- a/DuggaSys/microservices/gitCommitService/clearGitFiles_ms.php
+++ b/DuggaSys/microservices/gitCommitService/clearGitFiles_ms.php
@@ -1,40 +1,31 @@
 <?php
-
-// Remove includes
-// Include basic application services!
-// include_once "../../../Shared/basic.php";
-// include_once "../../../Shared/sessions.php";
-
-//--------------------------------------------------------------------------------------------------
-// clearGitFiles: Clear the gitFiles table in SQLite db when a course has been updated with a new github repo. This function is used by other github microservices.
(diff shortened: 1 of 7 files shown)


a22linre (commented):
The code looks good to me! One small thing, you could set header('Content-Type: application/json') just once at the start instead of repeating it. This is just a small readability improvement and doesn't affect the functionality.

About testing, I'm not sure how to properly test it myself, so maybe someone else can double-check that part.

a23louma:
@a23petny please see the small change request above and comment here when it's fixed, then someone can review it again.

Issue #16512, opened 2025-04-11
Acceptance test script for Course creation


HGustavs:
What steps are there for creating new course and connecting it to the repo at

https://github.com/LenaSYS/Webbprogrammering-Examples

PR #16875, opened 2025-04-23
Create courseCreationAcceptanceTest.php #16512

a20artje:
Test to check if course creation with giturl works.

Location of file is not good, don't know where it should be.

Should fix issue #16512 if I understood it correctly.

To test you can comment the delete part of the test and the course should appear in the database.

### DIFF
--- a/DuggaSys/microservices/courseCreationAcceptanceTest.php
+++ b/DuggaSys/microservices/courseCreationAcceptanceTest.php
@@ -14,7 +14,7 @@ $opt = 'NEW';
 $username = 'brom';
 $password = 'password';
 $coursename = 'Test Course';
-$coursecode = 'TE001S';
+$coursecode = 'TE002S';
 $courseGitURL = 'https://github.com/LenaSYS/Webbprogrammering-Examples';
 $AUtoken = 1;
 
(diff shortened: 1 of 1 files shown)


a22wilas:
Hi. I would want some clarification on what this issue/pull request are supposed to achieve. 
Is it supposed to verify if a course exists in the microservice: http://localhost/LenaSYS/DuggaSys/microservices/courseedService/createNewCourse_ms.php and then add it to the database.

If that's the case the code does not achieve the right outcome but would want verification on this before I submit a review!

a23louma:
@a20artje could you make a comment for this?

a20artje:
What it should do is test the createNewCourse microservice. 

It should call the service with data regarding a course to create. 
Then check that it was correctly added to the database,. 
Finally it should remove the added test course from the db.

a20artje:
Changed comments to more accurately reflect what happens

a22wilas (approved):
After some clarification I can now Approve this issue!

The code creates the course depending on if the circumstances are correct. 

To test the Issue I removed the "Delete" query.
Running the script shows that the course is created in the database with the GitHub link as well as showing a message telling the user that the test was a success. 

bild
Picture shows the message if the acceptance test were successful.

bild
We can now see the correct row with the GitHub in the database.

If the test is ran a second time with the same data an "error" message now shows that the test was not successful and given an explanation for why.

bild

Issue #16639, opened 2025-04-15
Remove the jquery in course editor


a23erigu:
Chance the jquery used in course editor to javascript

a23erigu:
more info found in parent

PR #16876, opened 2025-04-23
Remove jQuery in section #16639

a23elism:
### Removal of JQuery from Section
This part has only removed JQuery (AJAX if present is untouched), and is an re-opened PR from before last weekly merge (G3-2025-V3 is this ones original verison but has been updated to G3-2025-V4)


a23alera (approved):
Everything works as expected after the jQuery-to-JS changes. Can't find anything wrong. 

However, I did notice that the mobile view does not behave as intended (e.g. missing back button and view options for dugga entries). After testing, I can confirm that this issue exists in the base branch as well and is not caused by this PR.

image
image1
image2
image3

We'll create a separate issue for the mobile view problems.

Issue #16718, opened 2025-04-16
Remove the jquery UI in code viewer


a23erigu:
Swap the jquery UI that is still used in code viewer to javascript

PR #16877, opened 2025-04-23
Remove the jquery UI in code viewer #16718

f23linan:
Removed all the Jquery UI in the codeviewer javascript file.


a23simry (approved):
The use of jQuery UI .resizable for method _resizeBoxes_ in codeviewer.js has been removed and changed to JavaScript, tested and works as before. Can't find any other missed uses of jQuery UI, so issue has been fixed.

Image

Some other lines have been edited that aren't strictly necessary but they also don't seem to brake anything.

PR #16837, opened 2025-04-22
Refactoring of the cogwheel button to use a React button component

a23yutta:
Cogwheel button has been refactored to use a reusable React component `Button.js` located in `/Shared/components/Button.js`.

jQuery logic for the related to the cogwheel button has been replaced with JavaScript in `sectioned.js` located in `/DuggaSys/sectioned.js`.

HTML for save/cancel button of  "edit section dialog" has been converted to React using the `Button.js` component

PR #16882, opened 2025-04-23
FAB button bug fixed: refactoring of the cogwheel button to use a React button component

a23yutta:
Cogwheel button has been refactored to use a reusable React component `Button.js` located in `/Shared/components/Button.js`.

jQuery logic for the related to the cogwheel button has been replaced with JavaScript in `sectioned.js` located in `/DuggaSys/sectioned.js`.

HTML for save/cancel button of  "edit section dialog" has been converted to React using the `Button.js` component.

This is the continuation of the pull request[#16837](https://github.com/HGustavs/LenaSYS/pull/16837#event-17364560509) fixing the bug: broken/non-functional FAB button. 
PR #16837 merged into branch G3-2025-V3 and got unmerged with PR [#16844](https://github.com/HGustavs/LenaSYS/pull/16844).

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -241,9 +241,10 @@ function selectItem(lid, entryname, kind, evisible, elink, moment, gradesys, hig
     item.style.boxShadow = "none";
   });
   const element = document.getElementById("I" + lid);
-  element.style.border = "2px dashed #FC5";
-  element.style.boxShadow = "1px 1px 3px #000 inset";
-
+  if (element) {
+    element.style.border = "2px dashed #FC5";
+    element.style.boxShadow = "1px 1px 3px #000 inset";
+  }
   // Default showing of gradesystem. Will show if has type "Test" or "Moment"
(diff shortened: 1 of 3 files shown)


a23fanro (approved):
Looks good!! Proper use of react as far as I can see. 
I think it would be better if each component had it's own css-file, so for instance on line 200 it could say 
`element.classList.add("selected-item");`
and then there would be a react class component css related to 'button.js' named 'button.css' that has the lines:
`.selected-item {
  border: 2px dashed #FC5; //(new line here, could not add in comments for some reason...)
  box-shadow: 1px 1px 3px #000 inset;
}`
 there could then also be another function for removing the styling (as is done on line 295) with something like:
 `items.forEach(item => {
  item.classList.remove("selected-item");
});
`

But as far as this PR goes, with just 'translating' the jquery to react, it does what it is supposed to do. 

I suggest making another issue with sub-issues for each react component made thus far, implementing this approach. This will prevent inline styling.

TO BE CLEAR: this PR is approved for merge.

PR #16885, opened 2025-04-23
Fixed: autosave counts as manual save

a23kevre:
The autosave no longer counts as a manual save. 
image
image


a23eliax (approved):
I confirm that this works very well!

image

PR #16886, opened 2025-04-23
Fixed the lines between objects when zooming in andor out

a22abdjo:
By changing the line spacing between the lines, you are now able to zoom out and in without the it looking too messy. Especially when zooming out, the lines no longer deattaach from the source and target object.

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -118,7 +118,7 @@
     width: 320px;
     bottom: 0px;
     z-index: 5000;
-    background-color: #eb4;
+    background-color: rgb(34, 33, 32);
     color: #FFF;
     border-top-left-radius: 15px;
     border-bottom-left-radius: 15px;
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -118,8 +118,7 @@
     width: 320px;
     bottom: 0px;
     z-index: 5000;
-    background-color: rgb(243, 141, 39);
-    color: #FFF;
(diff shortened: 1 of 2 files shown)


a23miksk (changes requested):
This pull request should probably be scrapped and remade, it solves the problem in some ways but introduces a lot of other problems. _A lot_ of code is changed or remade so it's hard to say what is working or broken. It also changes the option panes colour? 

We could get a working solution to the issue by editing the line 49 in line.js from 
` const lineSpacing = 30 ; //Controlls spacing between lines` to 
` const lineSpacing = 30 * zoomfact; //Controlls spacing between lines`. 
This is a single line instead of the +188 -371 changed in the branch.

This pull request is a few merges behind G1-2025-v4 as well so make sure that the branch is synchronized with the weekly branch before asking to merge!

a22abdjo:
The last commit shows the neccessary modicifations on line 49 regarding the line spacing. This change was the only one made to fix the issue.

a23aleva (changes requested):
Lines between objects now work as they should while zooming in and out thanks to the zoomfact change. I don't know about the change of color to the options-pane, which was initially set to **#eb4** but is now **rgb(243, 141, 39)**. Although a trivial change I'm not sure it is a desired inquiry for this particular issue. Nevertheless it's a small modification to make so please just change it back and the issue will get approved!

https://github.com/user-attachments/assets/77486eb3-0e4f-4e37-882b-6f42c77fb169

Issue #16769, opened 2025-04-17
Locate and change the JQuery used in codeviewer.js with normal JS - line 1499 to line 2269


a23erigu:

a23elism:
### NOTE: Ignore all AJAX, this will be made its own issue

PR #16887, opened 2025-04-23
Swapped the jquery to Javascript

EdvinWerner:
Swapped the Jquerycode to javascript as mentioned in issue #16769
named my branch to V3 instead of V4 by misstake.

### DIFF
--- a/DuggaSys/codeviewer.js
+++ b/DuggaSys/codeviewer.js
@@ -2257,8 +2257,8 @@ function rendercode(codestring, boxid, wordlistid, boxfilename) {
 	str += "</div>";
 
 	// Print out rendered code and border with numbers
-	printout.css(createCodeborder(lineno, improws) + str);
- 	var borders = [...document.querySelectorAll('.codeborder')];
+	printout.innerHTML = createCodeborder(lineno, improws) + str;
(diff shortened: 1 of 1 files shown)


c22isako (changes requested):
### Both replaced lines breaks functionality in different ways.
Tested scenarios:
- Replacement of only line 1855
- Replacement of only line 2126
- Replacement of both line 1855 and 2126

### Issues:

**Replacement of only line 1855:** Testboxes are not visualized properly, "show box" (marked in red via image1) -dropdown isn't displayed but for a small gray box, and code is not visualized.

After replacement of line 1855: (image1)
bild
Before replacement of line 1855:
bild

**Replacement of only line 2126:** Code and boarders are not visualized properly.

After replacement of line 2126:
bild

Before replacement of line 2126:
bild

**Replacement of both line 1855 and 2126:** With both replacements applied, the code is shown but testboxes still doesnt work, and "show box"-dropdown is not shown.

After both replaced:
bild
Before both replaced:
bild

f23linan (approved):
After the changes the issues seems to have been resolved, textboxes shows in multiple different templates, example shown below.

image
image

Approved for merge!

PR #16888, opened 2025-04-23
Fix for overlapping elements

a23matsh:
The issue where ER and UML elements could overlap is now fixed by updating helper.js file. The issue was that ER relations and UML inheritance was using a custom center-point overlap check inside the function entityIsOverlapping(), which wasn't reliable when detecting collisions with other elements. This caused them to be placeable on top of UML classes, IE entities, and SD states unlike other elements which correctly triggered an overlap error.

The issue was fixed by removing the custom center-based logic and applying the same collision detection method used by all other elements. So the elements stated in the issue that could overlap, can't anymore either. .

### DIFF
--- a/DuggaSys/diagram/helper.js
+++ b/DuggaSys/diagram/helper.js
@@ -146,32 +146,18 @@ function entityIsOverlapping(id, x, y) {
             });
         });
 
-        // Collision detection for the ER and UML Relation elements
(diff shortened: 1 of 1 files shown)


a23eliax (approved):
Works as expected! The screenshot shows that an UML final state cannot overlap an ER entity.

image

Issue #16839, opened 2025-04-23
Problem in the options text fields


a23carpa:
Something have happend and made it so the text inside the textboxes in options now are vertical.
Addint text still works and is shown corretly in the class elements.
But this text is also then changed to be vertical.

https://github.com/user-attachments/assets/c34bd1a7-788b-42e5-a311-b946a4d85d29

PR #16890, opened 2025-04-24
Solution to issue #16839, Text box in options no longer display text vertically

a23carpa:
The text inside the diagram entities should not be horizontal as previous.
Don't know what caused the problem, but the solution is small so it should not interfere if the original cause is also fixed.

Before and after:
<p align="center">
  Before
  After
</p>

### DIFF
--- a/DuggaSys/diagram/draw/options.js
+++ b/DuggaSys/diagram/draw/options.js
@@ -448,9 +448,19 @@ function toggleOptionsPane() {
  */
 function textboxFormatString(arr) {
     let content = '';
+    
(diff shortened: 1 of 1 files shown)


a23aleva (approved):
The text inside diagram entities is now editable without every new attribute and function stacking up on the same line, instead they get their own row. Text is no longer placed in a vertical manner which was practically unreadable. Seems to work for both ER, UML, IE and state. Approved!

https://github.com/user-attachments/assets/d6676c9e-c2af-40d2-bc56-5e7de4c54e60

PR #16891, opened 2025-04-24
G3 2025 v4 issue 16738

a23fanro:
The mobile styling where overwriting my previous styling that fixed the visibility of the text in swimlanes, so I removed it.


a23lisjo (approved):
Looks good! I have not encountered any errors during testing.

PR #16893, opened 2025-04-24
Updated template to make it even more clear

a23wilgr:


a23oskme (changes requested):
Works well, I used the template to create an example file that got scanned as planned. 
Below is the entry in the database from taking the data using the template.

image

The script however, uses the same structure for extracting output data that it uses for extracting the parameter data. So I would suggest that the output part uses the same structure (its just a small clarification to avoid missunderstanding). Example below: 
image

Or, we can do changes in the script that reads the data so that it follows the current structure in the output part.

a23oskme (approved):
Looks very good, the template and the script work well together.

PR #16894, opened 2025-04-24
Changed jquery between lines 2184-2573 in sectioned.js

### LINKED ISSUE #16748: Locate and change the JQuery used in sectioned.js with normal JS - Commands - Line 2184 - 2573
Locate and change the JQuery used in sectioned.js (if present) from line 2184 going til 2573 with normal JavaScript

a23lisjo:
Fixes #16748 I have changed the found jquery between lines 2184-2573. I haven't found any error when using the program.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -2200,9 +2200,8 @@ function returnedSection(data) {
       drawSwimlanes(); // Create the swimlane used in the statistics section.
     }
 
-
     // Change the scroll position to where the user was last time.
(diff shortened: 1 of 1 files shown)


a23elism (approved):
All functionality remains from testing on sectioned.php (php file that uses the js version), as only a few JQuery instances appeared and have been converted not much can really break. From what i've seen/tested it all still has 1:1 functionality with the main group branch (G3-2025-V4)

example of 1:1 behaviour through the highscore box since it's JQuery styling was switched with normal JS:

image

looks identical to the group branch (only used 1 image since they look exactly the same)

Approved for merge 👍

PR #16895, opened 2025-04-24
Changed the JQuery to JS #16671

a23alera:
Changed the JQuery to JS between line 2502 - 2981


a23yutta (approved):
The code works as intended, it has the same functionality as before and does not break any other features.

All jQuery has been replaced between L2502-2981.

Approved.

PR #16898, opened 2025-04-24
Corrected IE relation line anchoring points

a23emisn:
The line anchoring points for the IE-relations were previously calculated by the height and width of the element. Due to the curvature and smaller size of the element inside, it appeared as if the lines didn't connect to the actual element, but instead floated mid-air. This has been corrected so that the lines now connect to a set of adjusted coordinates within the IE relation shape, which makes the lines visually connect to the IE-relation. 

Before. Line endings floating mid-air:
image

After. Corrected anchoring points and lines connect to the element:
image


a22abdjo (approved):
The issue has been solved appropriately and is approved.

Issue #16772, opened 2025-04-17
Locate and change the JQuery used in codeviewer.js with normal JS - line 2981 to line 3478


a23erigu:

a23elism:
### NOTE: Ignore all AJAX, this will be made its own issue

c22isako:
Can not find any usecase of maximizeBoxes() function within codebase. There might be a button missing, since the documentation mentions how, by pressing a button, you can maximize the code-window. The only button which has similar name is zoom-in, which doesn't utilize maximizeBoxes(). Since maximizeBoxes() contains the requested-change JQueries, unsure if necessary to change.

PR #16903, opened 2025-04-24
G3-2025-v4-#16772-change-jquery-in-codeviewer-2981-to-3478

c22isako:
**The requested JQuery changes are within the function maximizeBoxes() which isn't used, therefore removed the function and its documentation.**
Factors which brought the removal:
- Function hasn't been updated in the last 3 years.
- There seems to be no uses of the function within the codebase.

_For additional explanation, see connected issue, https://github.com/HGustavs/LenaSYS/issues/16772#issuecomment-2826783686_


f23linan (changes requested):
So unfortunately i did find the use for the maximizeBoxes function.
From the actual lenasys page:
image
image

So now with the function removed you cant minimize or maximize the left side of the code example.  Video below shows the changes in this branch vs the "normal" lenasys.


https://github.com/user-attachments/assets/eb44a60f-c9ab-4af4-b727-0185420f31f2

f23linan:
If ive misunderstood it please let me know!

Issue #16830, opened 2025-04-22
Re-engineer microservice retrieveCodeviewerService


a23louma:
Follow the "coding standard for removing includes" for this file in folder /CodeViewerService. The coding standard can be found in /microservices/Microservices Documentation
/microserviceCodingStandard.md.

PR #16906, opened 2025-04-24
Re-engineered microservice retrieveCodeviewerService #16830

a22linre:
Re-engineered retrieveCodeviewerService and its inverse dependencies. Also formatted the code using the built-in formatter, so some changes might look more than they are beacuse of the formatting. I hope I did the re-engineer part correct, still not sure. 

This fixes issue #16830.

### DIFF
--- a/DuggaSys/microservices/codeviewerService/deleteCodeExample_ms.php
+++ b/DuggaSys/microservices/codeviewerService/deleteCodeExample_ms.php
@@ -3,7 +3,7 @@ date_default_timezone_set("Europe/Stockholm");
 
 // Include basic application services
 include_once "../sharedMicroservices/getUid_ms.php";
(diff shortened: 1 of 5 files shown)


a23louma:
Don't review this until changes have been made.

a23louma:
I will just tag @a23axehe so you know that changes have been made.

a23axehe (changes requested):
I believe the code works, i could not find any breaks between this branch and the original branch when looking at the codeviewer and testing the solution. However it is not according to the standard to use GET in files which are responsible for either create, update or delete operations. I think the Get operations needs to be changed to Post operations in order to preserve the standard

d22marwi (approved):
The Get operations has been changed to Post operations in order to preserve the standard. Correct.

Issue #16892, opened 2025-04-24
clarify microserviceCodingStandard.md


a23louma:
It would be good to make this documentation more clear, for example what has to be added to the "main file" and what has to be added to the "inverse dependency file". Also if possible with some examples with what has to be replaced with the new code in the original file.

PR #16908, opened 2025-04-24
Create microserviceRefactorPOSTnoAnswer.md

a20artje:
Tutorial for POST without return answer.

Partly fixes issue #16892


a23danme (commented):
I  think the tutorial is solid and understandable, but I guess I think that there are people that dont understand it due to different aspects  such as not working with it earlier. It probably is posssible to exaggerate showing some steps with even more details. But too much info could also reverse it and make the tutorial more like an essay and harder to understand so I am not 100% sure.

"POSt" on line 75 hehe.

header("Content-Type: application/json"); isnt always needed unless its outputting JSON right? if so maybe put a comment before it saying like "When JSON something". I am not good in that area, maybe all POST calls need it. Just thought that people could think that it is always needed and a part of an normal implementation of the POST call.

a20artje:
I think i have resolved the issues brought up in the comments now.

PR #16909, opened 2025-04-24
Documentation for createNewCodeExample and createNewListEntry

### LINKED ISSUE #16678: Research document microservices in microservices/sharedService
Document services(starting from the beginning of the folder) that can be tested against new template so that it conforms to that template. When one or several files have been documented (depending on how much this takes). Create a new sub-issue to continue with the following files.

The template can be found in issue #16537 (OBS! has been changed).
see also #16824 for the new template

a23wilgr:
Documented the two first files in the sharedMicroservices file. Unsure of calling methods in createNewCodeExample, and if examples of use is necessary in both?

fixes #16678

### DIFF
--- a/DuggaSys/microservices/Microservices Documentation/sharedMicroservices.md
+++ b/DuggaSys/microservices/Microservices Documentation/sharedMicroservices.md
@@ -55,8 +55,8 @@ It retrieves the user ID and username, inserts a new entry into the table, gener
 -
 
 ### Microservices Used
-- getUID_ms.php
-- retrieveUsername_ms.php
+getUID_ms.php
+retrieveUsername_ms.php
 
 
 # createNewListEntry_ms.php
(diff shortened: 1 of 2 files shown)


a23louma:
Don't review until some small changes have been made.

a22wilas (changes requested):
Minor changes needs to be done. Looking at the "createNewListEntry_ms.php" there seems to be no post happening. At the moment a function is being called using inverse dependence and not HTTP commands. Also in the same file the only include I could find was "retrieveusername_ms.php" and not "basic.php".
bild

This issue also changed the document template.md. Not sure why, this issue should probably only change sharedMicroservices.md file. If the template need to be changed a separate issue should be created.

a23louma:
A review can be done again now, I will fix the changes of the template file so it's fine.

a22wilas (approved):
Looks good. Follows the set documentation standard and after some adjustments, now also provides correct information about the two microservices.

I approve this Issue!

PR #16910, opened 2025-04-24
16836 implement scrolling the canvas in the mobile version

b23danhe:
Implemented the ability to scroll on the grid/canvas with touch. Added events in _**diagram.js**_ that listens to events caused by a touch input. _**tmoving**_ function is handling the movement of the finger touch and the _**tstart**_ function is storing the start position of the finger.

_There is still more functions that has to be adjust to touch but this will be handled in other sub-issues related to the mobile implementation of the diagram tool._

**Before:**
_(No ability to use touch to scroll on canvas)_

https://github.com/user-attachments/assets/2c52eb53-576c-4d20-8990-ad2cf9417446

**After:**
_(**New feature!** Ability to scroll on canvas)_

https://github.com/user-attachments/assets/f0ae3339-5009-441b-984a-387b410cc5ba

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -914,33 +914,48 @@ document.addEventListener("mouseout", function (event) {
 
 // --------------------------------------- Touch Events    --------------------------------
 
+/**
+ * @description Event function triggered when touch is registered on top of the container.
+ * @param {TouchEvent} event Triggered touch event.
+ */
+
+function tstart(event) {
+    // Only responds to single-finger touch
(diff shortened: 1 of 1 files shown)


a23miksk (approved):
I confirm that the added code makes it possible to now scroll up and down in mobile-mode using touch. I tried with my own mobile and it worked as you expect, with other parts of the diagram(not scrolling) still being broken using touch. I tried recording on my mobile to show it working but it could not record the taps which makes the recording useless.
The code doesn't seem to break anything in the desktop version either.

Issue #16745, opened 2025-04-17
Locate and change the JQuery used in sectioned.js with normal JS - Commands - returnedSection(data)


a23elism:
Change the JQuery used in Sectioned under Commands, specifically the returnedSection(data) function (Line 1390-2180) with normal JS

a23elism:
### NOTE: Ignore all AJAX, this will be made its own issue

PR #16912, opened 2025-04-24
Locate and change the JQuery used in sectioned.js with normal JS -= commands =- returnedSection(data)#16745

a23elism:
This PR is focused on the returnedSection(data) function (alot of code)


a23simry (approved):
I've tested adding heading and sections, changing properties, changing dates, etc.
I wouldn't say it worked well previously but these changes don't seem to cause any issues.

PR #16916, opened 2025-04-24
G3 2025 v4 locate and change the j query used in sectioned.js with normal js   setup and event listeners   line 3530   4082

EdvinWerner:
Swapped the Jquery for javascript as mentioned in the issue

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -2740,7 +2740,7 @@ $(window).keyup(function (event) {
         }, 200);
         //Duration time for the alert before remove
         setTimeout(function () {
-          $("#updateAlert").removeClass("createAlertToggle");
+          document.getElementById("updateAlert").classList.remove("createAlertToggle");
           document.getElementById("updateAlert").innerHTML = "";
         }, 3000);
       }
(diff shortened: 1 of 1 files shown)


a22maxda (changes requested):
### Overall the code seems to work, however there appears to be an issue with error message variables.

Screenshot 2025-04-28 at 18 11 54
This error pops up when trying to edit a current course or create a new course.

Below is a comparison of how the windows differ between branches:
Screenshot 2025-04-28 at 18 12 58
This is on the current branch.

Screenshot 2025-04-28 at 18 14 22

This is on the group branch.

The biggest problem is the errors in the console and that the Create button does not function on the current branch, I assume it is connected to the issue but could be something else as well.

The issue appears to stem from this line in sectioned.js:
Screenshot 2025-04-28 at 18 18 13
It is worth noting that neither errorMsg1 or errorMsg2 are in use on the current branch whereas both are in use on the group branch.

I have not found any other issues so I believe this is the only change that needs to be made.

c22isako (approved):
**Code works, can not find any earlier noticed issues (or issues linked to the PR at all.)**
- Create works.
- Edit works.
- Invalid end date (both for course and deadline) displays the correct error message & relative and absolute deadline works as well as it did before.
- None of the earlier error messages are displayed within console.

(left = correct input, middle = invalid date (course), right = invalid date (deadline)
bild

**Errors which isn't connected to the PR:** (for issuecoordinator /group leader)
- Relative deadline - works if checked/de-checked absolute deadline first, but doesn't update as well as it should. The wrong date is displayed.
- Errors within dates - When selecting invalid date, ex. start 4/4 & end 4/3, and then changing to 4/4 & 4/4, error persists. Error isn't present if selecting 4/4 & 4/4 without invalid first, though you can not apply the change.
- Edit - works given that the version name is filled out, which can be troublesome with the test data 1337 for testing course which doesn't have an version name. - not linked to PR though. If anything, should be an separate issue to look into (if worth)

Issue #16800, opened 2025-04-22
fix message of the day on mobile


a23erigu:
The "message of the day" has a close button that covers things on mobile view

pls move or change the button on mobile

picture of "message of the day" on mobile
![Image](https://github.com/user-attachments/assets/5cafbf27-8742-4482-b012-cc4108577cdf)

PR #16918, opened 2025-04-24
Fix message of the day on mobile #16800

a23kavis:
Rewrote the logic for motd banner
Previous
image

Now with changes
image

The content has been placed in flexboxes and aligned correctly relative to the size of their parent containers, and smaller devices with extra support using media queries.


EdvinWerner (approved):
Looks good to me, works as described in the commits and I at least can't find any more issues caused by these changes, therefore it is approved for merge.

PR #16921, opened 2025-04-24
Fixed the keybinds to be consistent

b23ebbst:
All element/entities keybinds are now updated to be consistent. 

_**How it looks as of this fix:**_
image
image
image

The previous solution where every entity had their own tooltip-name (such as "_tooltip-PLACE_ATTRIBUTE_") was not working as these tooltips seem to not exist since they generate no keybind. 

To fix this all elements on the same row use the same tooltip to make sure that the correct keybind is generated. For example, switching them from "_tooltip-PLACE_ATTRIBUTE_"  to "_tooltip-PLACE_RELATION_" might make it less clear in the code, but the "_ATTRIBUTE_" generates no keybind while "_RELATION_" generates the correct one for the specific elements on that row - this achieves the consitency we were looking for and fixes the issue.

### DIFF
--- a/DuggaSys/diagram.php
+++ b/DuggaSys/diagram.php
@@ -139,6 +139,8 @@
                                 <span class="placementTypeToolTipText"><b>ER entity</b><br>
                                     <p>Each entity represents an object which is a representation of concepts or data.</p>
                                     <p>The entity only holds the name of the object and if it depends on another object.</p>
+                                    <br>
+                                    <p id="tooltip-PLACE_ENTITY" class="key_tooltip">Keybinding:</p>
                                 </span>
                             </div>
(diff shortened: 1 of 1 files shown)


a23aleva (approved):
Keybinds are now consistent and show up for all elements. Code is altered to give better clarity in regards to the different tooltip labels which appears more unisonally coherent than before. All good and approved!

When testing this I found that some of the information inside the tooltips is misspelt and sometimes using the same description for different objects. Also the tooltip for **Box Selection** is overlapped by the diagram frame. However, this isn't anything related to this issue and I will make a separate issue for that.


https://github.com/user-attachments/assets/239efc5e-be1b-48c7-a1d8-106cf998a167

https://github.com/user-attachments/assets/93213ff5-6bd2-499d-a193-0ee188c94eda

PR #16924, opened 2025-04-24
Fixed margin for the subicon tooltips

a23kevre:
Before:
image
After:
image


a22abdjo (approved):
The solution provided solves the issue with reasonable changes and is approved.

Issue #16869, opened 2025-04-23
The align for labels becomes messy when adding cardinality to a line


b23danhe:
The issue is that the labels shift out of alignment when cardinality is added to the line. This only occur when there’s more than one line between classes.

https://github.com/user-attachments/assets/f33677cd-dcfa-48ae-9f70-bb3b7c7d192f

PR #16930, opened 2025-04-25
Line labels now dont move when cardinality is added

a23carpa:
Solution to #16869
Fixed so that when cardinality values are added to a line, they no longer effect the positionering of that same lines label.
Previously cardinality added offset the fils global variables fx, fy, tx and ty. Now instead its a self contained variable just for cardinality. 

https://github.com/user-attachments/assets/9d897d94-8e0e-4c0e-9d7b-05e83f1a9e02


a23eliax (approved):
I confirm that this works as expected!

image

PR #16932, opened 2025-04-25
G1 2025 v4 16838 sequence activity box cant connect to the top a sequence object

d22abdha:
The activity box can be placed on the top of object version.


a23eliax (approved):
The issue is solved. The only thing is that the object-box went smaller, but since this was not a part of the issue and you can make it bigger by draging the corners out, I approve!

_Before:_
image

_After:_
image

a23emijo (changes requested):
Why did the object-box go smaller? This should not be the case because it can be fixed without the re-size. Fix this and then we can review again!

a23matsh (approved):
The object box is now restored back to its original size that it had from the start, and the solution still works as presented in the screenshot down below. This is now approved.

image


image

PR #16935, opened 2025-04-25
Fixed: Box selection hover-box is cut off 

a23gabst1:
The solution was to add a specific css by targeting only the tooltip text that is a child of the element with the id "mouseMode1", only meant for the tooltip text for the box selection, so the remainder of the tooltip texts dont get affected by this change of position.

PS: Auto indented the css

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -1,31 +1,34 @@
 @keyframes loadingSpin {
-    0%{
+    0% {
         rotate: 0deg;
     }
-    100%{
+
+    100% {
         rotate: 360deg;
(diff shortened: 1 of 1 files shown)


a23eliax (changes requested):
The hover-box still cuts of a little at the top, the pointer hover-box doesn't and that is the one that's even higher up so I assume the box selection hover-box shouldn't be cut of at all

_Pointer:_
image

_Box selection:_
image

a23eliax (approved):
My screen was zoomed in, when I fixed that, this solution works! So this issue is solved.

Issue #16925, opened 2025-04-24
Fix bug: delete without selection of items of the type header


a23yutta:
Items of the type header problems:
- Cannot directly delete an item when clicking the trashcan button without selecting the selection-box
- Unable to select/deselect selection-boxes of all items within the header by selecting the header selection-box
- All items within the header is not being selected for deletion when clicking the trashcan button 

This is related to the PR #16879 of the issue #16336

PR #16940, opened 2025-04-25
G3-2025-V4-#16925-delete-without-selection-header

c22isako:
**Since sections already had the functionality, added it to header-type items as well.**
Unsure if this will be used, since I haven't found any headers with trashcans from browsing existing ones. Although, if trashcans were added to them, the problem is solved. for additonal information surrounding delete via trashcan, view: https://github.com/HGustavs/LenaSYS/pull/16879#event-17385722221

select course of your choice (which has duggor, ex. Testing Course) -> log in -> from here, if you press the cogwheel on a dugga you can change its type. By doing this, you can "create" applicable headers.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -746,7 +746,7 @@ function markedItems(item = null, typeInput) {
   var subItems = [];
 
   //if the checkbox belongs to one of these kinds then all elements below it should also be selected.
-  if (kind == "section" || kind == "moment") {
+  if (kind == "section" || kind == "moment" || kind == "header") {
     var itemInSection = true;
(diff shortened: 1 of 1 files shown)


a23osckr (approved):
Fixes issues mentioned in issue #16925. However, instead of empty boxes remaining for deleted items for ~15 seconds as mentioned in #16879, the items now completely restore themselves for ~15 seconds when selecting another header/section/moment after deleting one or more items. This is similar to the empty boxes and not a huge issue. But it is worth looking into as it causes some confusion at first. However, I don't see how these changes could cause this and overall it appears good. These changes are approved, but the bug should be fixed.

Issue #16870, opened 2025-04-23
Re-engineer microservice retrieveHighscoreService


a23louma:
Follow the "coding standard for removing includes" for this file in folder /highscoreService. The coding standard can be found in /microservices/Microservices Documentation
/microserviceCodingStandard.md.

PR #16941, opened 2025-04-25
Re-engineer microservice retrieveHighscoreService #16870

a22linre:
Re-engineered retrieveHighscoreService and its inverse dependencies. Also formatted the code using the built-in formatter, so some changes might look more than they are beacuse of the formatting. 

This fixes issue #16870


a23danme (approved):
Looks good!

PR #16943, opened 2025-04-25
Implemented stereotypes for UML-elements

a23miksk:
Implemented the stereotype header for uml-elements. Is also added in local storage so it can be saved and loaded. 

image

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1386,6 +1386,7 @@ function saveProperties() {
         }
         addToLine("primaryKey", "*");
         addToLine("attributes", "-");
+        addToLine("stereotype", "");
         addToLine("functions", "+");
     }
     stateMachine.save(element.id, StateChange.ChangeTypes.ELEMENT_ATTRIBUTE_CHANGED);
--- a/DuggaSys/diagram/classes/element.js
+++ b/DuggaSys/diagram/classes/element.js
@@ -24,6 +24,7 @@ class Element {
             functions = null,
             altOrLoop = null,
             alternatives = null,
+            stereotype = null,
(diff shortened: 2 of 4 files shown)


a23vinge (approved):
The developer did as they mentioned: they implemented the stereotype header for UML elements. They also made sure it is saved and loaded through local storage, as they described.

I do not see any problems with the code, nor any bugs introduced by these changes. I tested it using both Google Chrome and Microsoft Edge and did not find any issues.

One small observation: when changing the class name, it's possible to simply press "Enter" to confirm the new name, but when editing the stereotype, the user must click the save button — pressing "Enter" just adds a space. This could be a small usability tweak for the future but does not block this change.

a23emijo (changes requested):
The one small obersvation from a23vinge should be fixed before merging 😄

a23vinge (approved):
The developer addressed the previously mentioned issue: pressing "Enter" now correctly confirms the stereotype input, so that’s no longer a problem.

I did notice one new inconsistency that might (a very small detail)  be worth looking into. When entering a long string in the Name field, the text wraps and flows downward to stay visible within the box. However, in the Stereotype field, the text continues past the visible bounds of the box instead of wrapping. You can see this difference in the image I attached, both fields contain the same text. 
"123456789101112131415161718" = 27 characters

This isn't a critical issue since it's unlikely someone will input long strings for stereotypes or names during typical use. Still, if consistency in UI behavior is a priority, it might be worth adjusting the Stereotype field to behave similarly to the Name field.

Aside from that, everything looks good — no bugs or broken behavior found in testing.


_The inconsistency I mentioned_
image

image
image

Issue #16642, opened 2025-04-15
Remove the jquery in course editor


a23erigu:
Chance the jquery used in course editor to javascript

PR #16946, opened 2025-04-25
G3-2025-V4-#16642 JQuery to JS in courseed.js

a23osckr:
All JQuery I could find has been converted to JS in courseed.js. I have done testing and all appears fine. I made JS functions for the fade animations that JQuery implemented. The JQuery .ajax() requests have been replaced with fetch and in some places XMLHTTPRequest since fetch was not able to return a boolean without changing every implementation where the function is called. The refactor would be too large for this issue, but will have to be done at some point. Added flatpickr library to replace the JQuery datepicker.

### DIFF
--- a/DuggaSys/courseed.js
+++ b/DuggaSys/courseed.js
@@ -14,7 +14,6 @@ var LastCourseCreated;
 var lastCC = false;
 var updateCourseName = false;
 
-
 $(document).ready(function () {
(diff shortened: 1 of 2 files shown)


a23alera (approved):
Couldn't find any problems with the changes. Everything looks fine, and the changes to the JavaScript file look correct

Issue #16923, opened 2025-04-24
Cardinalities for lines between ER Entities and ER Relations aren't shown


a22nevan:
The lines between ER Entities and ER Relations are able to have cardinalities added to them. Usually, when the cardinality is added to a line, it shows up as text next to the line. This is due to the lines having their anchor points in the center of the ER Entity, causing the cardinalities to be positioned behind the element.

Visible cardinalities for line between UML Classes:

![Image](https://github.com/user-attachments/assets/099a0dcf-e5ab-4e6d-9180-aa1aa51b373d)

Cardinalities for line between ER Entities and ER Relation not visible. Can see parts of the cardinality sticking out from behind the Entity itself:

![Image](https://github.com/user-attachments/assets/78d8b2b7-0bb1-4826-947c-776f81c59075)

PR #16947, opened 2025-04-25
Fixed cardinality issue with ER lines #16923

a23matsh:
Cardinalities were being hidden in ER relation lines due to inconsistent line calculations.

- Removed the separate logic recursiveERRelation() function and replaced it with getLineAttributes() to unify line positioning logic across all diagram types.
- getLineAttributes() was refactored to return clear coordinates and offsets.
- To eliminate a visual gap between ER relation diamonds and their connecting lines introduced by the adjustment, a small ER-specific offset tweak was also made.

**Master version**
image

**Fixed version**

image


a22abdjo (changes requested):
The issue with the cardinalities regarding the ER lines are fixed and approved. But while zooming out, the lines disconnect from the target and source element. This is not an issue in the G1-2025-v4 branch and therefore needs to be looked over before merging.

PR #16949, opened 2025-04-25
HelperFunction that is supposed to replace includes

a23wilgr:
Created a function that should be able to be reused in all files, instead of using includes. Also added a short explanation in the end.

### DIFF
--- a/DuggaSys/microservices/HelperFunction_ms.php
+++ b/DuggaSys/microservices/HelperFunction_ms.php
@@ -1,0 +1,40 @@
+<?php
+
+    //Supposed to be used instead of includes
+
+    function callMicroserviceGET($microservicePath){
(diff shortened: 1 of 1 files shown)


a23dilel:
I try to use HelperFunction, but I can't test it, because tried to use a retrieveSectionedService_ms microservice and replace include from inverse denpendency to HelperFunction and use callMicroserviceGET(). I try to test and try to find retrieveSectionedService_ms in lenasys website, but I couldn't find it. I leave someone else to test it.

a22linre (commented):
Looks good and it worked when I tested it. 

For testing I made a small local test file to try out the three functions. Had to switch to http instead of https because it's local, and also had to add header('Content-Type: application/json'); and echo json_encode([...]); in the services that was needed for the testing to get a proper JSON response for testing purposes.

Issue #16950, opened 2025-04-25
Update microserviceList.php so that it is relevant to the UI mock-up


a23bensh:
As of now the microservice documentation/microservice list is outdated and needs to resemble the UI mock-up in a way where additional functionality can be implemented: Searchability, filtering, adding of microservices, editing microservice details and deleting microservices. These functionalites should only be created as placeholders in this issue and later implemented individually in the remaining sub issues.

**Image of outdated microservice list:**

![Image](https://github.com/user-attachments/assets/cf300f09-bc83-4149-a685-19e32df811a4)

PR #16952, opened 2025-04-25
Updated microservice directory UI to match mock-up

a23bensh:
Fixes issue #16950. This updates the microservice directory to better resemble the mock-up and includes placeholder functionality for future implementation. The directory can be viewed at microserviceUI.php (for testing) and currently only supports viewing different microservices (no interactive features yet). More resemblance to the mock-up can be added later if wanted. As of now for example, only "parameters " show if there are no values inserted, but a more refined detailed view can be done at a later time. Other issues need to be created for the implementation of the search function, the filter function, adding a microservice via a form, editing a microservice and deleting one. An issue for a more appealing design could also be argued is needed.


a20artje (approved):
Looks good, microservices contain relevant information. Will make searching for documentation much easier.

As mentioned in comment, search, filter, add, delete, edit functionality needs to be added.

PR #16953, opened 2025-04-25
Documentation of getUid_ms.php and isSuperUser_ms.php

### LINKED ISSUE #16913: Document Microservices in /sharedService: getUid_ms.php AND isSuperUser_ms.php
Document services that can be tested against new template so that it conforms to that template. When these files have been documented (depending on how much time this takes). Create a new sub-issue to continue with the following files.

The template can be found in issue https://github.com/HGustavs/LenaSYS/issues/16537 (OBS! has been changed).
see also https://github.com/HGustavs/LenaSYS/issues/16824 for the new template

a23wilgr:
Documented getUid and isSuperUser from the sharedMicroservices folder. Unsure if examples of use is necessary.

fixes #16913

### DIFF
--- a/DuggaSys/microservices/Microservices Documentation/sharedMicroservices.md
+++ b/DuggaSys/microservices/Microservices Documentation/sharedMicroservices.md
@@ -1,1 +1,87 @@
 # sharedMicroservices Documentation
+
+# getUid_ms.php
+
+## Description
+Retrieves the user's UID (user ID) from the session. If a user is not logged in, guest ID is returned instead.
+Also logs events into serviceLogEntries-table.
+
+## Input Parameters
+
+- Parameter: $opt
+   - Type: String
+   - Description: Operation type
+
+- Parameter: $courseId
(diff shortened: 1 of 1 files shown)


b23johgu (commented):
The descriptions for getUid seems fine, but I'm unsure about the types as I can't find anything clear about it in either the database nor the files. 

The description for isSuperUser seems to be a duplicate from getUid. I think that this ms only checks if the user and returns a true or false. It also doesn't log anything. Otherwise everything seems good!

b23johgu (commented):
The types seems to be correct now as of what I can find as well. But the description for isSuperUser_ms on row 62-63 still seems to be a duplicate if i'm not misinterpreting something

a23louma:
@a23wilgr please see the comment above.

b23johgu (approved):
Looks good now! I don't find a parameter type for opt either so I think its ok to leave it like this (for now).

PR #16955, opened 2025-04-25
16700 better lines from sequence lifeline to sequence object2.0

a23vinge:
So I used to work on this issue before under the name "16700 better lines from sequence lifeline to sequence object" but i took my time and needed to make a new branch "16700 better lines from sequence lifeline to sequence object2.0" under G1-2025-v4, so there is no confusion.

"In the functions related to SE (Sequence Element/diagram), the coordinates were automatically generated by a function, but they didn't include all the information needed to provide clear coordinates for the arrow to be drawn between the relations without getting stuck in the middle. The solution was to remove the function (getLineAttrubute) and then add a new one that works more with getBoundingClientRect(). This is a predefined method provided by the browser that returns a DOMRect object containing more accurate and detailed positioning data for elements.  By using getBoundingClientRect(), we now have clearer coordinates for all objects involved in the diagram, which ensures the arrows are drawn without any unnecessary gaps. (yippy)

The function "getSequenceElementConnectionPoint" is to define getBoundingClientRect(). 

Then the adjustment to 24-39 is to adept to the new information that is available through getBoundingClientRect()

Before: " If we take a good look at the image, we can see that nothing is really connected, there is  a small space between everything."
before SE

After: That small space has been removed with the help of getBoundingClientRect() that has given better result, essentially we are asking DOM (Document-object-model); "Hey browser, where exactly is this element on the screen right now? Tell me its coordinates." With this a person can manipulate better. 
After SE

_Something to know:_ The problem with this solution is that if "something" does not have DOM then this will not work, but does every website have DOM, Generally speaking yes so should not be a problem. 

There is also alot of changes in this code and this might have affected other things, as for example on the image "After" it can be seen that the UML relationship no longer have a gap.


a23miksk (changes requested):
This code solves the problem of "hovering" lines that are only attached to bounding boxes and not the graphical element. However, since the solution is to set every lines "from" and "to" point as the centre of every object, the lines from and to icons (arrows, relationship types, etc...) are completely hidden. This also causes the lines to look a bit skewed and sometimes be in the way when attempting to click elements. 
I also think that in a sequence diagram a line should never go from the inner lines of an object, but from the "sequence object" box.

a23miksk (approved):
The changes now prohibit the user from creating lines between sequence diagram objects and actors, which was the core issue from the start.

PR #16958, opened 2025-04-25
Update updateCourseVersion_ms.php

a22wilas:
Fixes Issue https://github.com/HGustavs/LenaSYS/issues/16927

Now the call to make a course version active is being done with HTTP request POST.

To test this use the cogwheel on a course page and mark "change this to default version" and then check the table course in the database.


a23gabli (approved):
The calls have been changed to use POST and everything works as intended.

Issue #16868, opened 2025-04-23
AutoSave is only a copy of the first save file


a23aleva:
AutoSave needs to be changed since its purpose right now is questionable. When you create your first saved file AutoSave also gets added with the same save state. Even if you change your first save file, or makes any changes at all, AutoSave will never save anything again. The only way to alter AutoSave is to choose the save name AutoSave and overwrite it as if it's just another save file. Instead it would be better if AutoSave actually saved every change the user did as a failsafe to revert to in case an actual save was never made.

_AutoSave is created with a new save file_

https://github.com/user-attachments/assets/3fc5dd2a-feb0-4b97-8912-a7f97cde3fc7

_Changes never apply to AutoSave itself_

https://github.com/user-attachments/assets/901fbc59-517a-4a79-8a92-4641b2a36e4d

PR #16969, opened 2025-04-28
AutoSave is now latestChange (name Subject to change)

a23carpa:
A sollution for [#16868](https://github.com/HGustavs/LenaSYS/issues/16868) 
Changed so the autosave file is now called latestChange and keeps track of the latest user action on the canvas and saves them. So instead if it having only the first save, it now keeps the latest thing a person did. As its not a save per say any more, it just keeps track of what he last thing the user did was. 

If the program crash before a user saves, it should keep track of how the canvas looked. As long as no new input is given. 
It does not distinguish between save files, there is only one latest change.

[Screencast from 2025-04-28 09-31-47.webm](https://github.com/user-attachments/assets/95498ef8-966f-4b60-9d7f-8615c4e07163)

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -135,7 +135,7 @@ class StateMachine {
                 console.error(`Missing implementation for soft state change: ${stateChange}!`);
                 break;
         }
-
+        updateAutoSave()
     }
 
     /**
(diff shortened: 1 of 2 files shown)


a23aleva (changes requested):
latestChange now saves newly added or deleted elements actively as the user is working instead of only being a copy of the first save file. Looks good and is approved!

https://github.com/user-attachments/assets/5eebef51-0abc-4e54-85f7-ff7a88ebb99a

a23aleva (approved):
Sorry I clicked request changes instead of approve in my last review.

PR #16971, opened 2025-04-28
Visual snap indicator for mmoving element

a23aleva:
When the sequenceActivation object has already been placed on the diagram grid and is moved close to a lifeline there is now a visualization which indicates snappable locations just like for the ghost version.

There are some issues with this interaction that were already there before. The hitboxes when placing something on a lifeline seem wonky and for some reason the actual snap applies everywhere along the lifeline's vertical line, even outside its boundaries. To fix this changes will need to be done to the actual snap logic, which could be the target for potential new sub-issues.

https://github.com/user-attachments/assets/9c6a215e-da10-49d1-ad84-a10181bef3a0

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1051,6 +1051,21 @@ function mmoving(event) {
                 deltaX = startX - event.clientX;
                 deltaY = startY - event.clientY;
                 // We update position of connected objects
+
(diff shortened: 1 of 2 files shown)


a23carpa (approved):
From my testing i can see that it works as displayed in the video and fixed the issue in question. If you pick up a "sequence activation" object after placing it the snap effect still works.

And as stated the lifeline object it self creates a line the whole x value its on, not just the dotted line. For future issues to fix.

PR #16986, opened 2025-04-28
Get tutorial for refactoring - Issue #16933

a20artje:
A tutorial for the replacing of functions with a return and no parameters sent.

Fixes Issue #16933


a22linre (commented):
Easy to follow and understand! Nice work!

Just curious, this only mentioned GET, what about POST?

PR #16990, opened 2025-04-28
Updated the datesize

EdvinWerner:
Now it doesn't display the date after a breakpoint of max-width of 570 pixels

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -884,6 +884,7 @@ p {
   .dateSize {
     width: 25%;
     font-size: 10pt;
+    display: none;
   }
 	.internal-link {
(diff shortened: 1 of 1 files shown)


a23lisjo (changes requested):
I don't really see any difference. The date is still visible even when the screen is 570px or smaller. 

Now:
image

EdvinWerner:
I think that the reviewer forgot to branch correctly. Since the date is hidden once the screen width is smaller than 570 pixels, I've doublechecked this on my own computer and the computer of a friend to make sure that it works. 

See attached screenshot to know how it should look once the correct branch is selected. 
image

PR #16991, opened 2025-04-28
G2 2025 v4 #16966 a23oskme

### LINKED ISSUE #16966: Make microservice endpoint db easier to use
Fix so that the microservice endpoint database wont need to be manually removed, installed and then filled (with two scripts) before it can be used.

a23oskme:
Created a file that removes the database (if there is any) and creates a new one . Then it runs both fill scripts to fill the database. 
This means that if you wish to create the database you will only need to run 'setupEndpointDirectory.php' instead of having to run multiple files. 

This can be tested by simply running 'setupEndpointDirectory.php' and then open both 'microserviceList.php' and 'dependenciesList.php' and check if the database has been filled. Fixes #16966


a23danme (approved):
Looks good and seems to work well. Well-structured and good comments on the code.

PR #16997, opened 2025-04-28
Documentation for microservices createListEntry and deleteListEntries.

### LINKED ISSUE #16985: document microservices in microservices/sectionedService: createListEntry_ms.php AND deleteListEntries_ms.php
Document services that can be tested against new template so that it conforms to that template. When one or several files have been documented (depending on how much this takes). Create a new sub-issue to continue with the following files.

The template can be found in issue https://github.com/HGustavs/LenaSYS/issues/16537 (OBS! the template has been changed)
see also https://github.com/HGustavs/LenaSYS/issues/16824 for the updated template

b23johgu:
Documentation for microservices createListEntry and deleteListEntries in sectionedService.

Could not find the parameter type for $opt and I am a little unsure about $log_uuid as well.

fixes #16985

### DIFF
--- a/DuggaSys/microservices/Microservices Documentation/sectionedService.md
+++ b/DuggaSys/microservices/Microservices Documentation/sectionedService.md
@@ -1,1 +1,141 @@
 # sectionService Documentation
+
+# createListEntry_ms
+
+## Description
+Adds a new list entry to a course, handling the creation of a new code example if necessary, and then retrieves the updated sectioned data for the course. List entries are duggas, headers, tests etc.
+
+## Input Parameters
+
+- Parameter: $opt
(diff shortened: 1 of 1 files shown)


d22marwi (approved):
Everything is correct, I approve even though we are missing $opt because I could not find the parameter type for $opt either. Is just documentation so maybe we should merge anyway?

I think char is correct for $log_uuid, UUIDs are usually a bit number, Its 36 characters, 32 digits + 4 dashes. so maybe:

Parameter: $log_uuid
   - Type: char(36)
   - Description: For logging purposes.

(I am not 100% sure if this is correct in this context, or necessary to implement)

Issue #16335, opened 2025-04-09
Deleting an element in a course


a23erigu:
There are multiple problems with deleting an element

Most of the problems should be in
LenaSYS/DuggaSys/sectioned.php
and
LenaSYS/DuggaSys/sectioned.js

PR #16998, opened 2025-04-28
G3-2025-V4-issue-16720-tables-second-try

a23fanro:
> **~~~~DO NOT MERGE THIS, IT IS A VERY BAD PR~~~~**

made tables for Sectionlistc into a real table with each item being its own row instead of a new table. 

But there are alot of dependencies... there is now no styling for the tables, and the functionalities do not work anymore. The drag-and-drop does not work, since it seems to be made for dragging entire tables, not rows. The same is applicable to the delete-function, which shows the dialog but does not delete anything anymore. 
I will list the functions that still work below, since that will take less time.

The functionality that still works is:

- Click on Dugga-name to see dugga
image

- Open link in new tab
image

- Get Canvas Link
image

- Settings
image

Oh and there is also a bunch of divs being created over the table. If removed all other tables vill stop existing.
image

This issue needs to be divided into sub-issues.

**TO BE CLEAR: DO NOT MERGE THIS, I just need it for the future when creating sub-issues**

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -1659,12 +1659,6 @@ function returnedSection(data) {
         menuState.idCounter++;
         // All are visible according to database
         
-        // Create a row in the main table
-        str += `<tr id='lid${item['lid']}' value='${item['lid']}' value='${makeTextArray(item['kind'], valarr)}'`;
-        str += " >";
-        
-
-
         var hideState = "";
         if (parseInt(item['visible']) === 0) hideState = " hidden"
         else if (parseInt(item['visible']) === 3) hideState = " deleted"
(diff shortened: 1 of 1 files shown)


a23fanro:
made it a little more viable with 2nd commit [WIP](https://github.com/HGustavs/LenaSYS/pull/16998/commits/1f04d8cbcc7e11ace44a1a4ada81a77a94bcb626)

- The drag-and-drop now works

https://github.com/user-attachments/assets/217ea3cb-d260-4ba2-a122-a63ecde76b35

a23fanro:
Delete does still not work

a23fanro:
Can now see that the delete dows not work because it never worked [#16335](https://github.com/HGustavs/LenaSYS/issues/16335).

a23fanro:
last [commit](https://github.com/HGustavs/LenaSYS/pull/16998/commits/777afb11c7c5a4cfee385c6b07b239ebf3784113) made the rows into their own tables again, fml

Issue #16951, opened 2025-04-25
Re-engineer microservice retrieveResultedService


a23louma:
Follow the "coding standard for removing includes" for this file in folder /resultedService. The coding standard can be found in /microservices/Microservices Documentation
/microserviceCodingStandard.md.

PR #17001, opened 2025-04-28
Re-engineer microservice retrieveResultedService #16951

a23danme:
Fixing issue "Re-engineer microservice retrieveResultedService #16951 "


a20artje (commented):
I cannot find where this is used in the website, so I am unable to test this. I was able to tes retrieveResulted service with cmd commands, that part works. Could not test the getUserAnswer.

Unsure if the whole thing works until I can find where it is used.

a20artje:
I have been able to test it now and it works as intended.

PR #17003, opened 2025-04-28
Improved tooltips information

a23kevre:
All tooltips now have correct grammar (the ones in modes menu). One thing to note is that dots are used in some places and not in others. Could be made into a sub-issue to remove dots or add dots.


a23aleva (changes requested):
ToolTip looks good with some minor nitpicks that are easily altered. Might as well fix these straight away. Do this and the PR subsequently will get approved!

Change to a UML
_Insert an UML inheritance to the diagram_ - line 428

Remove last comma between **hardware** and **or** for the lines below:
Example: "This version is used for human users, external hardware or other subjects"

_This version is used for human users, external hardware, or other subjects_ - line 753

_This version is used for human users, external hardware, or other subjects_ - line 766

_This version is used for human users, external hardware, or other subjects_ - line 817

_This version is used for human users, external hardware, or other subjects_ - line 869

a23aleva (approved):
Changes requested are now applied so this issue should be closed and approved for merge!

Issue #16758, opened 2025-04-17
Locate and change the JQuery used in sectioned.js with normal JS - Setup and Event listeners - Line 4084 - 4476


a23elism:
Locate and change the JQuery used in sectioned.js with normal JS under Setup and Event listeners - going from line 4084 til 4476 (end of file)

a23elism:
### NOTE: Ignore all AJAX, this will be made its own issue

PR #17004, opened 2025-04-28
Replaced jQuery with JS in sectioned.js between L4084-4757(end of file)

a23yutta:
jQuery code has NOW been replaced with JavaScript in the file `sectioned.js` between  L4084-4757(end of file). 

Make sure you...
- log in as admin 
- enable student feedback: Settings for a dugga --> Enable student feedback

...before testing.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -4386,10 +4386,10 @@ function refreshMoment(momentID) {
 //------------------------------------------------------------------------------
 function showUserFeedBack(lid, feedbackquestion) {
   AJAXService("GETUF", { courseid: querystring['courseid'], moment: lid }, "USERFB");
-  $("#userFeedbackDialog").css("display", "flex");
-  $("#feedbacktablecontainer").html("");
-  $("#statscontainer").css("display", "none");
-  $("#duggaFeedbackQuestion").html(feedbackquestion);
+  document.getElementById("userFeedbackDialog").style.display = "flex";
+  document.getElementById("feedbacktablecontainer").innerHTML = "";
+  document.getElementById("statscontainer").style.display = "none";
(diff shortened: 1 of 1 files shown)


b23sebth (dismissed):
**Approved: Fixes #16758**
## What's done:
The code successfully replaces the JQuery code with a JS counterpart. Everything worked as expected while testing. An icon is also correctly displayed when student feedback is enabled. The changed code follows the established coding standard in regards to indentation and variable naming.

I did not manage to test the updateTemplate function, but the code is correct and nothing appears to break.

## Considerations
There appears to be no actual way to give feedback even though the option is enabled. This does not relate to the JS implementation seeing as that option does not appear with the JQuery implementation either.

Further, while the code does follow the standard, the standard in of itself is outdated in certain parts. An example is the use of var instead of let or const. The coding standard may need to be updated. That does not affect this PR however.

### Feedback activated
feedback_icon
### Dugga example
bit_dugga

b23sebth (commented):
**Fixes #16758**
## What's done:
The code successfully replaces the JQuery code with a JS counterpart. Everything worked as expected while testing. An icon is also correctly displayed when student feedback is enabled. The changed code follows the established coding standard in regards to indentation and variable naming.

I did not manage to test the updateTemplate function, but the code is correct and nothing appears to break the website. The JQuery is replaced with Javascript correctly and should work exactly as before.

## Considerations
There appears to be no actual way to give feedback even though the option is enabled. This does not relate to the JS implementation seeing as that option does not appear with the JQuery implementation either.

Further, while the code does follow the standard, the standard in of itself is outdated in certain parts. An example is the use of var instead of let or const. The coding standard may need to be updated. That does not affect this PR however.

### Feedback activated
feedback_icon
### Dugga example
bit_dugga

Issue #16722, opened 2025-04-16
View buttons positions


b23sebth:
The buttons are currently placed on the left side next to the tables which creates a lot of unneccesary white space directly underneath them. One possible solution to this would be to move the buttons to a dropdown located on the header. These settings could become visible by a button on the left side of the header with the text "View". I have added a rough mockup of the placement.  

## Example from Demo-Course:
![Image](https://github.com/user-attachments/assets/24164a31-f302-4124-a867-cf0b1d022bd7)

## Rough mockup: 
![Image](https://github.com/user-attachments/assets/31b91e42-745e-41a7-93ef-5f3a327cab16)

PR #17006, opened 2025-04-28
G3 2025 v3 #16722 view buttons position

b23sebth:
## What's changed:
I have added a reusable React Dropdown component that's placed inside of Shared/components. The actual implementation uses the Button component found in the same directory. An open source icon from Google fonts has also been added. When the dropdown is pressed, each row quickly slides down with a small delay added to each one for a nicer effect. The dropdown will animate up to 9 buttons placed inside it, if more is needed, CSS keyframes will need to be mofified to accomodate it. A function to close the dropdown by clicking anywhere on the page has also been implemented. 

This dropdown is then used to move the view buttons onto the header, which removes the unneccesary space underneath them previously. The placement is similar to that of the mockup in the issue and can be seen in the video below.

### Dropdown example
image

### Dropdown example when clicked
image

### Dropdown animation example
https://github.com/user-attachments/assets/677cd7ca-3551-4d4c-bd59-f0201f490212

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -4477,7 +4477,8 @@ function setViewMode(mode){
 
 //elementID is the ID that gets passed to the dropdown component.
 function showDropdown(elementID) {
-  let element = getElementById(elementID);
+  let element = document.getElementById(elementID);
+  element.classList.toggle("dropdown-content-show");
 }
 
 // Adds the ability to close dropdown by clicking anywhere on the page.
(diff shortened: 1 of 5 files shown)


c22isako (approved):
**The solution generally works well. It relieves space from the side and looks better than original solution, with the same functionality + it now works on mobile as well!**

Though, it has some small issues which isn't really "breaking":
- Header placement when using small dimensions (such as for the phone) is affected. Though, I believe this should be a separate issue to solve since it seems like the original solution had some issues with placement as well. 
(right side = mobile, left side = "normal")
bild


- When pressing the icon-image on the dropdown, it's not activated which may pose some irritation. This might also benefit from being a separate issue // Talked with issueCordinator.
bild

PR #17007, opened 2025-04-28
Added search functionality (safe from sql injections aswell)

### LINKED ISSUE #16963: Implement the search function for Microservice directory UI

a23bensh:
Fixes #16963 and adds search functionality to the microservice UI. Searchable terms include microservice name and description as of now. More terms can be added later if desired. "Search name/description" placeholder will also need to be changed when filter functionality is added, to something like "Search...". To test the search function, go to microserviceUI.php.


a23gabli (approved):
The search works and the code follows correct procedure to protect from injections.

PR #17015, opened 2025-04-29
Added recursive relation for SE

a23miksk:
Added the ability to set a recursive relation for a sequence activation box. This implementation causes the recursive relation to always be at the top of the box, which means that users will have to make additional boxes in sequence if they want to show the recursive sequence between two sequence changes. Labels should probably also be able to be added to the recursive sequence, but that does not work yet.
image


a22abdjo (approved):
The solution provided has resolved the issue and therefore is approved.

PR #17021, opened 2025-04-29
Create script to be able to download Microservice directory database #17011

### LINKED ISSUE #17011: Create script to be able to download Microservice directory database

a23oskme:
Created a script that starts a download for the database. Fixes #17011 

Simply clicking 'downloadDb.php' in LenaSys should start the download.


b23johgu (commented):
I only get "Database not found" as a result. So the file works but dont seem to find the database file.
I can't seem to find the database file "endpointDirectory_db.sqlite" in the system at all. Not in the database either. Unsure of where else to look for it.

I found this on row 22 in .gitignore but I don't know if it has anything to do with anything.. 
`DuggaSys/microservices/endpointDirectory/endpointDirectory_db.sqlite`

a23oskme:
> I only get "Database not found" as a result. So the file works but dont seem to find the database file. I can't seem to find the database file "endpointDirectory_db.sqlite" in the system at all. Not in the database either. Unsure of where else to look for it.
> 
> I found this on row 22 in .gitignore but I don't know if it has anything to do with anything.. `DuggaSys/microservices/endpointDirectory/endpointDirectory_db.sqlite`

Sorry, should have mentioned that you need to install the database first. So try to run the file "installEndpointDb.php" before, and then try to download it again.

If its not installed this else statement will execute instead of the download. Maybe I should add some more information to the echo telling the user that it needs to be installed? We could also make the download script run the installation, but I don't think thats a good idea, since people can press download by misstake etc. 
image

b23johgu:
I can't seem to install the database, the file is in endpointDirectory, right? 

Screenshot 2025-04-29 at 11 50 52

I think it's a really good idea to add more information to the "Database not found" to check that the one trying to install it has done all the neccessary steps before.

a23oskme:
> I can't seem to install the database, the file is in endpointDirectory, right?
> 
> Screenshot 2025-04-29 at 11 50 52
> I think it's a really good idea to add more information to the "Database not found" to check that the one trying to install it has done all the neccessary steps before.

I think that might be a local problem with your LenaSys installation and the permissions. Since the installation script has been tested by multiple people. 
I can send you the database file over discord and you can plant it in the folder manually, since this pull request is wether the download script works or not :)

b23johgu (approved):
Works well, look at comment above, I approve 👍

Issue #16827, opened 2025-04-22
Re-engineer microservice createNewListEntry


a23louma:
Follow the "coding standard for removing includes" for this file in folder /SharedMicroservices. The coding standard can be found in /microservices/Microservices Documentation
/microserviceCodingStandard.md. **Now the functions in DuggaSys/microservices/curlService.php can be used.**

PR #17022, opened 2025-04-29
Re-engineer microservice createNewListEntry #16827

d22marwi:
Added POST implementation to createNewListEntry inverse dependencis and removed include_once createNewListEntry from coursedService/copyCourseVersion_ms.php because its not used.


a20artje:
When i try to add a new listEntry on the website nothing shows up sadly.

Issue #16880, opened 2025-04-23
Deleting sections with items still in them


a23erigu:
If you press the mark icone on a section it selects all the items inside it as well.
but if you then manualy de select all the items inside then you can delete the section without geting rid of the things inside

example picture
![Image](https://github.com/user-attachments/assets/682da851-e827-4702-b7eb-42f183617565)

PR #17027, opened 2025-04-29
G3-2025-V4-#16880-deletion-of-section-without-children

c22isako:
**Prevented deletion of section whiles items within the section still exists.** 
- Streamlined selection of Section/Header/Moment - items by making sure that it stops when it found the next section (meaning current sections items end.)
- Made sure that if a code-dugga has been deselected, the connected section is also deselected. This means that the section can not be deleted without everything within it also being deleted.
- Created function for gathering the code-duggas parent (section) that might also be reused in the future.

select course of your choice (which has duggor, ex. Testing Course) -> log in -> test selection box!
- box marked in red should be able to select all items within the section (and deselect if everything is selected). 
It should also be deselected when a deselection of items within the section happens - which is this fix.

bild

_Note to be discarded if deemed unnecessary: The solution could possibly be made more effective. If so, should probably be another issue since it **_might_** involve revamping the selection system - and may not currently be worth it. (Just a headsup)_

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -854,23 +854,49 @@ function markedItems(item = null, typeInput) {
   if (kind == "section" || kind == "moment") {
     var itemInSection = true;
     var sectionStart = false;
+    const elements = document.querySelectorAll('#Sectionlist .item');
 
-    $("#Sectionlist").find(".item").each(function (i) {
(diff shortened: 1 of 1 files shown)


a22carsv (approved):
Testing revealed no way of deleting a section without its children, hence resolving the issue. Great work.

PR #17032, opened 2025-04-29
Improve error handling for microserviceUI.php when database is missing. #17024

### LINKED ISSUE #17024: Improve error handling for microserviceUI.php when database is missing.
When opening microserviceUI.php without creating a database using installEndpointDb.php, a simple message is displayed stating that "http://localhost/LenaSYS/DuggaSys/microservices/endpointDirectory/microserviceUI.php might have a temporary problem or it could have moved.". This makes it unclear what the issue is and there should be a message stating that a database has not been created yet for the purpose of correct error handling.

a23oskme:
Added a try/catch block to the pdo connection. If its unsuccessful there will be an error message, else the page will render. Fixes #17024 
(Remember to remove 'endpointDirectory_db.sqlite' if you have it to test this)

Suggestion: when the setupEndpointDb.php file has been approved we can add a button to the error message, letting the user setup the database directly in the UI. If the user is supposed to be able to setup the database of course, otherwise ignore this suggestion. 

image

### DIFF
--- a/DuggaSys/microservices/endpointDirectory/microserviceUI.php
+++ b/DuggaSys/microservices/endpointDirectory/microserviceUI.php
@@ -86,8 +86,84 @@ try {
     <div class="line">
         <h1>Microservice Directory</h1>
     </div>
-    <?php if (!isset($dbError)) {
-        if (!isset($microservice)) { ?>
+    <?php
+    if (isset($dbError)) {
+        echo "<p style= 'color:red';>" . $dbError;
+    } else {
+        if (!isset($microservice)) {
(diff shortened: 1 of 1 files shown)


a23danme (commented):
Works well, as shown in the picture "install" is spelled wrong which should be changed. Also I think it would be good to maybe not only have "Install it first!". Maybe add some type of information to simplify it like "Run file "setupEndpointDirectory.php"" or something? Otherwise, if you aren't aware of how it works, it could take some uneccassary minutes of trying to "install it".

PR #17034, opened 2025-04-29
Documented addCladd and addUser

### LINKED ISSUE #17012: document microservices in microservices/accessedService
Document services: **addClass_ms.php AND addUser_ms.php** that can be tested against new template so that it conforms to that template. When one or several files have been documented (depending on how much this takes). Create a new sub-issue to continue with the following files. 

The template can be found in issue https://github.com/HGustavs/LenaSYS/issues/16537 (OBS! the template has been changed)
see also https://github.com/HGustavs/LenaSYS/issues/16824 for the updated template

a23wilgr:
Documented addClass and addUser from the accessedService map. Unsure if the $newusers input parameter in addUser should be the same as $newclass in addClass, since newusers can affect 2 tables depending on the situation?

fixes #17012

### DIFF
--- a/DuggaSys/microservices/Microservices Documentation/accessedService.md
+++ b/DuggaSys/microservices/Microservices Documentation/accessedService.md
@@ -1,0 +1,97 @@
+# accessedService documentation
+
+# addClass_ms.php
+
+## Description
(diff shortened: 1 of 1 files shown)


b23johgu (approved):
Looks great! I can't figure out what $responsible would be as well so I think it is fine to leave it like that.

Issue #17008, opened 2025-04-28
Note text won't flow across entirety of spread


a23aleva:
When writing things in the note text field it only seems to use half the space of the spread even though the note's sole purpose is to contain text. The text also has some weird formatting/cut to next line instead of prioritizing moving the full word. Enlarging the note makes no difference as the text will never fill its entirety anyway. This should be fixed.

https://github.com/user-attachments/assets/23c4f1af-8f98-4f5d-bed7-7c292099f77d

PR #17035, opened 2025-04-29
Fixed issue #17008

a23eliax:
Improve note rendering by calculating character width more accurately and allowing text to fill full width.

https://github.com/user-attachments/assets/a01226b5-adba-4710-8bab-805505993494


a23gabst1 (changes requested):
This almost fixes the issue #17008, however one thing that needs to be improved before receiving approval is the breakpoint for when the words begins at a new row, see video below for illustration.

https://github.com/user-attachments/assets/9a724cef-9272-4d64-9261-9b6598a10056

PR #17039, opened 2025-04-29
Fixed sequenceActiation to not snap to the lifeline when outside the range

b23ebbst:
Updated mouse.js so the sequenceActivation doesn't snap to the lifeline on the y-axis, when outside the range. Checks have been implemented to see it the lifeline has ended, and if that is the case the sequenceActivation does not snap to the y-axis. 
If the sequenceActivation is moved into the lifelines range it will still snap to it!

https://github.com/user-attachments/assets/fdf65ecc-8b9d-4f97-8d5f-09c23ccb66a8


b23danhe (commented):
This solution works as intended, objects now snap only to the lifeline. The implementation looks solid and does not affect any other existing logic. 
Good job!

b23danhe (approved):
Approved!

Issue #16972, opened 2025-04-28
The toggle error check does not show errors on the ER-diagram


a23gabst1:
The problem to fix is to make sure that the toggle error check shows the errors on the ER-diagram when it is on by marking the elements in red borders.

This is how it should look like based on what I found in wiki
![Image](https://github.com/user-attachments/assets/0c574d06-38b5-47dd-8013-b85e8606078d)

how it look now
![Image](https://github.com/user-attachments/assets/ead1cb32-37be-4132-aab6-b3b84fdbae21)

PR #17040, opened 2025-04-29
Fix for error checking & added one new error check issue #16972

a23matsh:
Reviewed the code and identified that there was logic for validating ER entities and relations, so this wasn't the issue. The main issue was a broken loop, where the iteration over the lines array was not handled correctly, causing connection checks to fail. This was corrected by fixing the loop to iterate through all lines. Also implemented a new error check that seemed reasonable that ensures that ER entities without any connections (lonely entities) are flagged as errors.

In the screenshot down below you can see that after creating your model and then clicking the error detection button it now flags errors that are implemented in the logic. For example entities with no connections are marked as errors, entities with no primary key, entities with a relation that are missing cardinality, models with the same name and so on.

**Note that all the models in the screenshot are from the fixed version, just different examples**

image


a22abdjo (approved):
The issue has been solved and works as intended and therefore is approved.

Issue #16843, opened 2025-04-23
QuickSave should be implemented as a feature


a23aleva:
Currently, pressing ctrl+s brings up the "save as" menu that allows you to either save the file to a new name or overwrite the current file if the name is the same. If you've already saved a file and picked a name for it pressing ctrl+s should implement QuickSave which just auto-saves the changes to that already existing file without having to bring up the "save as" option. For users unfamiliar with the ctrl+s command a FAB (Floating Action Button) could be implemented as an alternative to use QuickSave. These changes might force a name change of the existing save button to "save as" instead of just "save" to avoid confusion.

_Example of quick save FAB_
<img width="703" alt="Image" src="https://github.com/user-attachments/assets/98e5d6ca-8293-4bbe-81ad-202073428759" />

_Pressing ctrl+s in the current version_

https://github.com/user-attachments/assets/4cdb6740-4079-4f23-a585-e16522c9588f

_Can't quick save to existing file without going through these steps_

https://github.com/user-attachments/assets/0cfa14af-b1ee-434f-a1a9-827abd454e47

PR #17046, opened 2025-04-29
[#16843] Reworked the save function to add Quicksaving functionality

a22nevan:
Quick save functionality has now been added to the regular save button. On top of that, a new button has been added to the toolbar called "Save As". This button essentially has the same functionality as the old Save button had. The Save As button has a new (mostly placeholder) icon. Feel free to change the icon if it doesn't meet the quality standards for the sidebar icons. The new button also has a new keybind (ctrl + shift + s (Had to update the code for the tooltip generation to include "shift" as a print out. Also made it include "alt" as well)).

**New toolbar button:**

image

The save button now functions as a regular save (or quicksave, as the issue called it), saving the diagram directly to the most recently saved to file. If there isn't any recent file it acts as the new "Save As" button, prompting the user to input a file name for the diagram to be saved to. This functionality is the same for both the button and the keybind (ctrl + s).

The tooltip for the save button has been updated with a new type of tooltip text, making extra information in the tooltip less jarring to look at at a first glance. Haven't applied this to any other tooltip, but it is there to be used if needed.

**Tooltips for "Save" and "Save As" buttons:**

image

image

**Videos showcasing the new saving functionality:**

https://github.com/user-attachments/assets/043c808a-ea09-42e7-a5e0-1a52ffad929f

https://github.com/user-attachments/assets/395879c6-0b34-4d9d-adfd-788de006a7c5

**Known issue:**

- The event listener for keypresses in the diagram doesn't have any functionality to prevent registering multiple button presses from holding down a button. This usually isn't noticeable since not many keybinds are bound to anything that could happen multiple times. However with the new quicksave function the save happens instantly, causing it to happen multiple times if the button is held down for too long. This doesn't seem to have any impact on functionality but is jarring to see when trying to save a file. Wasn't sure how to fix this without risking impacting all keybinds so left it as is.

(I also want to point out that I added periods to the end of some tooltip sentences for other buttons in the diagram.php file. Not relevant to testing the save changes, but could potentially mess with other code when merging so thought I'd mention it just in case.)


b23ebbst (approved):
This works just as expected, really well done! The diagram saves to the current file (if there is one), otherwise it promts the user to create a new file. I approve!

https://github.com/user-attachments/assets/aafd104f-fb73-46d8-9e66-4e740bf72103

PR #17050, opened 2025-04-29
G1 2025 v4 16838 sequence activity box cant connect to the top a sequence object

d22abdha:
The sequence activity box would not be placed on top of the actor/object

https://github.com/user-attachments/assets/5584140c-1ea2-4d8e-8a2a-cb7be51e898b

### DIFF
--- a/DuggaSys/diagram/helper.js
+++ b/DuggaSys/diagram/helper.js
@@ -119,8 +119,12 @@ function entityIsOverlapping(id, x, y) {
 
         // No element can be placed over another of the same kind
         if (data[i].kind !== element.kind) {
-            // Sequence life-lines can be placed on activations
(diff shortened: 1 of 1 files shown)


a23eliax (changes requested):
This solution almost works. The issue was to fix so the activation cannot connect under or on top of the lifeline but it still can connect under. When this is fixed I can approve!

https://github.com/user-attachments/assets/ba9513a9-a71c-4dfa-96e9-66a936c2c77c

a23emijo:
> This solution almost works. The issue was to fix so the activation cannot connect under or on top of the lifeline but it still can connect under. When this is fixed I can approve!
> 
>  review.snapping.mp4

This is already another issue so this PR doens't need to fix this. I'll approve and merge this PR 😊!

a23thefo:
Was fixed with #17039

PR #17057, opened 2025-04-30
Updated str to lineStr

a23miksk:
Small update to hopefully fix the problem that occurred after merge where no recursive relations worked.


b23ebbst (approved):
Seems to be working really well! 
The self-referencing is working for both actors and objects. What I did notice though is that it only works if there are activations, if the lifelines are missing activations the errors seem to persist, but all lifelines that are active should have activations on them, so for that reason: I approve! 
Well done! 

https://github.com/user-attachments/assets/8bb46ccf-3728-4b70-a693-c88a050c8361

https://github.com/user-attachments/assets/08504f26-01f2-4d03-9afd-c4623b52be29

Issue #17064, opened 2025-04-30
Fix the layout of the title


f23linan:
Fix the layout of the course name in the header in mobile view

PR #17066, opened 2025-04-30
Fixed the course name in mobile view for issue #17064

f23linan:
Fixed so that the course name is as centred as it can be with the space in the header. Its still not perfect but the fab buttons needs to be sized down before it can be perfect but it is as good as it can get right now. 

image


EdvinWerner (approved):
Looks fine to me, good job!

Approved for merge

PR #17071, opened 2025-04-30
G3 2025 v5 j query to js in sectioned.js line 2575 3013

a22maxda:
Updated branch to be in line with Group branch and changed name to V5 from V4, related to PR #17000


a23alera (changes requested):
Several parts of the converted functionality are currently not working as expected based on my manual testing. The floating action button does not respond to mouse hover, and when clicked, it immediately triggers a message submission without letting the user select an option first. The announcement form also appears to be broken, on mobile, the button is visible in the hamburger menu but does nothing when clicked, on desktop, I was not able to locate the button at all, which makes the feature inaccessible.

The scroll position is, as I understand it, not being saved correctly when scrolling the page. Additionally, the yellow arrow next to “View” is visible but does not respond to clicks.

Note that these are the issues I was able to observe during my manual testing. It is possible that I may have missed some elements or misunderstood how certain features were intended to be accessed or triggered, such as the announcement functionality.

a22maxda:
The announcements button still does not exist in desktop view, however I am unsure if that should be it's own issue as the button does not appear on the group branch either. Otherwise code should work.

Screenshot 2025-05-06 at 13 23 15
Current

Screenshot 2025-05-06 at 13 23 40
Group

Issue #17025, opened 2025-04-29
Change the JQuery used in profile.js with normal JS


a23erigu:
I don't know how much this is actually used but there is jquery in it

a23erigu:
@a22maxda you goning to need to fix this

a22maxda:
I will update the PR as soon as I get the ajax function converted

PR #17073, opened 2025-04-30
G3 2025 v5 #17025 jquery to js in profile.js

a22maxda:
Changed JQuery in profile.js to JS, very small file but I can't find any issues with it


f23linan (changes requested):
The changes that have been made are still working however there are som jquery left that have not been removed. Will put everything here so it is a easy to find and fix!

    "var password = currentField.val();
    var newPassword = newField.val();"
    http://api.jquery.com/val/ 

"message.html("Challenge has been updated!!");
message.html("Teachers are not allowed to change challenge question!");
message.html("Incorrect password!");
message.html("Unknown error.");
message.html("Error: Could not communicate with server");"
https://api.jquery.com/html/

"field.css("background-color", "rgba(255, 0, 6, 0.2)");
 field.css("background-color", "white");"
https://api.jquery.com/css/

also the part with the $ajax is jquery that needs to be switched out!
https://api.jquery.com/jQuery.ajax/

a22maxda:
The PR is not ready for review as the ajax parts still need to be converted.

a22maxda:
All JQuery has now been changed to vanilla JS, I am unsure if the ajax conversion is correct as I could not find much about it but as far as I can tell nothing breaks. Should be ready for re-review

a23osckr (changes requested):
Profile.php has several issues and requires some changes.

* When changing challenge question the boxes do not become red if there is an error. 

* Trying to change challenge question produces an error in the console if both inputs have been typed into.

* Trying to change the password when entering valid inputs produces an unknown error

Below are the errors I found in the code that may cause these issues.

Typo in code below (#challengeAnswer) should not have a #
image

I believe error = function is incorrect. I believe a catch would be more appropriate. Will link an example of a catch directly below this image. This is true for both fetches.
image
image

response.json should be response.json();
image

This is invalid js syntax since backgroundColor is not a display property.
image

Aside from these issues everything appeared fine. The JQuery has been removed and when the issues are fixed I would assume no more problems.

Issue #17058, opened 2025-04-30
Error handler for the new installer


b23johgu:
Continuation of "Added specific error handling for events modal in new installer. #17028  ".

The error handler is supposed to show a pop up where you can change the wrong information that caused the error (names for root, database name etc.). I don't get this frame, only the one that is supposed to show after according to the dev that created it. Needs to be fixed. 

<img width="647" alt="Image" src="https://github.com/user-attachments/assets/f0931b6d-c248-4ab7-8433-7f490538c6a2" />

a23dilel:
Can someone test it with macOS, when you type wrong root username, root password or hostname to appear modal, there you can change these values? 

This pull request is old #17028, but this pull request #17083 is merged, so someone can test G2-2025-v6 branch.

a23oskme:
You can see my comment on this matter in the link below:

https://github.com/HGustavs/LenaSYS/issues/17081#issuecomment-2866211959

PR #17028, opened 2025-04-29
Added specific error handling for events modal in new installer.

a23dilel:
Now have implemented modal this issue #16039. There are four specific error handling which you can check this issue #16039. If you want to know how to get these error let me know.

Issue #16039, opened 2024-05-20
Add specific error handling for events in new installer


a22willi:
Currently the backend sends json data containing both an error message and what step it failed on. 
Implement a handler in the frontend that tries to give the user the option to fix the various errors. 

Some of the things that should be implemented. 
- if db connection fails. Show modal with ability to change root username, password and hostname.
- if it fails to copy course files. Show modal with chmod command to fix it.
- If creation of database or db user fails, show modal with force option.
- if execution of sql files fails, show option to restart installer. 

Note that the modal implementation (#16038) should be done before working on this issue. 

All steps should also have a retry button, which simply starts the installer again from the given step using the data that the installer sends when failing.

a23dilel:
I soon finish implement the modal, but how to get error "if execution of sql files fails, show option to restart installer." in lenasys? So that I can implement and show modal error message. Also what restart installer option does? I was thinking when user press the restart installer option, then reset all values (username, password, hostname and more) and go back step 1. Do I think correct? @HGustavs

PR #17083, opened 2025-04-30
Added specific error handling for events modal in new installer. Use Docker and Windows to test it.

a23dilel:
I have now implemented the modal for issue #16039. This pull request continues work from pull request #17028, but it doesn't work on macOS with XAMPP and merge deadline has passed, it was closed. I'm having problems with XAMPP, so I can't test it there, instead, I'm using Docker on Windows. This issue #17058 was tested on macOS with XAMPP and got some problems with the new installer, so there is sub-issue for that, but I know it works on Windows with Docker. 

There is Docker install guide here: https://github.com/HGustavs/LenaSYS/blob/master/dockerenv/document/docker_install_instructions.md

To get a database connection error, you need to enter an incorrect root username, password, or hostname. Im using docker, so hostname is db and if you use xampp then read this https://github.com/HGustavs/LenaSYS/blob/master/Shared/Documentation/newinstaller/documentation.md

Here are steps to get modal:
In step 1 select Windows

In step 2 check both:
oa

Follow these values or create own values:
1a

Here I typed wrong root username and password:
2a

And then continue all steps in default. 

When lenasys begin to install and get an error, then it will show a button "View problem" under progress bar:
image

When you press this button then it will show "Database Connection Failed":
3a

To get an error when creating a database or database user, you must first create a database and user with names that already exist, and then install Lenasys. When you did, these are already created, run the installation again using the same database and user names to trigger the error.

To get an error when executing the SQL files, need to type a typo in one of the SQL files located in Lenasys/newinstaller/SQL/. Add the typo there to get this error.

To get an error when copying course files, uncheck both “Create new MySQL database” and “Create new MySQL user” in step 3. Then type names for the database and user, but do not use existing ones instead, add new values.

### DIFF
--- a/newinstaller/installer_ui.js
+++ b/newinstaller/installer_ui.js
@@ -121,7 +121,7 @@ function navigateToPreviousPage() {
     }
 }
 
-function start_installer() {
+function start_installer(newData) {
 	let error_occured = false;
 	let languageSupportChecked = false;
 
@@ -131,7 +131,7 @@ function start_installer() {
                ['text', 'password', 'radio', 'checkbox'].includes(element.type);
     });
 
-	const data = {
+	let data = {
         verbose: false,
         overwrite_db: false,
         overwrite_user: false,
@@ -189,6 +189,10 @@ function start_installer() {
         }
     });
 
+	if (newData != null) {
+		data = newData;
+	}
+
 	console.log("Form Data:", data);
 
 	fetch('installer.php', {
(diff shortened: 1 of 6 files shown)


a23axehe (changes requested):
Reviewed using docker and windows. 
First tried to download with wrong credentials, did not work and error message appeared as it should. 

Then i tried to overwrite existing users without checking that box which made the correct error message appear. 

By adding a typo in one of the sql files the installer showed that there was a issue but the view error button was unresponsive. The typo added were: INERT INTO course (cid,coursecode,coursename,created,creator,visibility,activeversion,hp) VALUES (1894,'G420','Demo-Course',NOW(),1,1,'52432','1');
on line 1 of demoCourseData.sql. The download progress stopped and everything looked good except the button "view error" were unresponsive and did not appear to have an action connected.

Error while copying coursefiles on a non existing user without creating one did not work and error message appeared as it should.

Something i noted were that the new installer does not turn on db using docker(see attatched image) as the old installer did. Unsure of the consequenses of this.
Skärmbild 2025-05-02 140612

The error message not appearing when a typo existed in the sql file are worth checking out along with the db_using_docker option being turned off when docker is in fact being used(not really related to this issue so maybe a new one can be created regarding that).

a23axehe (approved):
SQL error message now appears as it should, great work. Good plan to fix db_using_docker in another issue if it is related to it.

Issue #17082, opened 2025-04-30
Change to "curlService" instead of "HelperFunction" in the instructions in the "curlService-php" file


a23louma:
The name of the file has been changed and the instructions(comments) in that files have to be updated as well.

PR #17084, opened 2025-04-30
Update curlService.php

a22wilas:
Fixes Issue https://github.com/HGustavs/LenaSYS/issues/17082

Updates the guide documentation in the bottom of the file.
Now shows the correct file-name that should be used for the curlService.


a23dilel (approved):
I double checked for typos and compared the HelperFunction_ms.php file, and everything looks the same and good!

Issue #16987, opened 2025-04-28
Creating a table in the database for all pages using the Diagram-drawing functionality


b23ebbst:
A table for all pages using the diagram-drawing functionality is needed in the database. 
This is needed to be able to dynamically affect the diagram-pages. For example, to add a filename in the nav-bar which should not be visible on any other page than the diagram-pages (this can be done by calling the table in the database and adding the functionality to only the pages in the table). Without this table the only solution is to hard-code which pages to affect, which in turn causes problems if pages are renamed, moved, or deleted as this would break the code. 

_**Example of the desired functionality**_
![Image](https://github.com/user-attachments/assets/010a9ca0-85e1-4cb3-b9c4-83de876d1c0c)

b23johgu:
@b23ebbst I need a little bit more info about what the files are about, do you create them yourself or are they already created? Which file in the system can I find the workspace pictured in issue #16807 ?

b23johgu:
I found the files yey, but the files seem to be saved in local storage?

I found this code in diagram.js (rows 2237-2242), maybe that could be used to display the file name in the header?

```
//Get the current file name that the user wants to use for saving to local storage. Also stores that current file to the activefile variable.
function getCurrentFileName() {
    let fileName = document.getElementById("saveDiagramAs");
    activeFile = fileName.value;
    return fileName.value;
}
```

PR #17087, opened 2025-04-30
Update createTables.sql

b23johgu:
Created a table to be used for filenames for issue "Creating a table in the database for all pages using the Diagram-drawing functionality #16987"

To test this you need to re-install LenaSYS and see if the new table is in the database.

### DIFF
--- a/Shared/SQL/createTables.sql
+++ b/Shared/SQL/createTables.sql
@@ -8,7 +8,7 @@ CREATE TABLE groupdugga (
 /* Table for filenames in courses and duggor associated with issue #16987 */
 CREATE TABLE duggaFiles(
 	filename				VARCHAR(100) NOT NULL UNIQUE,
-	quizID                  VARCHAR(100) NOT NULL UNIQUE,
-	course_id				VARCHAR(45) NULL UNIQUE
+	quizID                  INT(11) NOT NULL,
+	course_id				INT(10) NOT NULL
 	PRIMARY KEY (filename)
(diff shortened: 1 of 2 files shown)


a23oskme (changes requested):
It didn't work when I first tried to reinstall LenaSys. I did some researching and it appears that 'createTables.sql' actually doesn't appear to execute or install the tables inside it.
 
I tried to remove 'groupdugga' from 'createTables.sql' as well, but it still appears in the database when i reinstall LenaSys. This means that it gets installed somewhere else. 
I found that 'groupdugga' gets installed in 'init_db.sql', so I tried to remove 'groupdugga' from 'init_db.sql' and keep it in 'createTables.sql'. When I did this, 'groupdugga' didn't get installed. This led me to the conclusion that 'createTables.sql' doesn't work. 


I copy/pasted the 'duggaFiles' table into 'init_db.sql' as the picture below. Which seems to successfully install the table into the database when reinstalling LenaSys. (There was a small syntax error in 'duggaFiles' though, a missing "," between 'course_id' and the primary key)

I would recommend either looking into why 'createTables.sql' doesn't seem to work, or use a solution that is similar to the one I tried when reviewing this. 
I attached some images of the solution.

image

image

b23johgu:
I have moved duggaFIles to 'init_db.sql' and will create an issue that looks in to why 'createTables.sql' doesn't work (or if it's even used).

a23oskme:
You are missing a "," between cid and the primary key, which is causing the installation to skip it at the moment. Add it and it works perfect :)

image

b23johgu:
Fixed!

Issue #16858, opened 2025-04-23
Re-engineer microservice deleteCourseMaterial


a23louma:
Follow the "coding standard for removing includes" for this file in folder /courseedService. The coding standard can be found in /microservices/Microservices Documentation
/microserviceCodingStandard.md.

PR #17090, opened 2025-04-30
DeleteCourseMaterial reengineered - Issue #16858

a20artje:
Fixes deleteCourseMaterial dependency to retrieveCourseedService.

It says all rows changed in retrieveCourseedService but i only fixed indentation, only thing I added is pdo_connect() and session_start(), I also removed the function wrapper in that file, no other changes.

To test, open up the database and go to codeexamples. There should be a long list. Find the course which has all those codeexamples (I think one of the courses is test something, open it up and you should see on the bottom the codeexamples). When that course gets deleted the codeexamples should all be deleted too.

Fixes Issue #16858


a23danme (approved):
Works! Looks good.

Issue #17048, opened 2025-04-29
View button not fully clickable


a23erigu:
The view button in section does not work if you press the part with the hamburger menu on it.

picture of what part you can't click on
![Image](https://github.com/user-attachments/assets/aac7938d-bb3e-4813-9520-a5a0fcd0733f)

a23erigu:
It also appears that the picture for the button can't be found so fix that to

PR #17093, opened 2025-04-30
G3-2025-V5-#17048-view-button-not-fully-clickable

c22isako:
### The dropdown is now fully interactable, and the icon is visible for everyone.
**Fixed icon blocking button**
- Since the problem stemmed from the icon receiving clicks instead of the button, made the icon non-interactable so that the button instead receives the clicks. Also fixed some minor spelling mistakes. 

**Fixed icon missing**
- There was some issues regarding the image-path within some systems where /lenaSYS/ were missing. Fixed this by starting with ../Shared/ instead of /lenaSYS/.

_**One approach to testing this, and how to get to the functionality:**
Enter course of your choice, ex. Demo-Course and investigate view dropdown, marked in blue:_
bild

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -4829,7 +4829,7 @@ function getLocalStorage() {
 function setViewMode(mode){
   const section = document.getElementById("Sectionlisti")
 
-  // Remove previously applie mode
+  // Remove previously applied mode
   section.classList.remove("scroll-mode", "overview-mode");
 
   // Apply selecte view mode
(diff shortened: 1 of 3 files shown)


b23sebth (changes requested):
**Fixes #17048. Minor change needed.**

## What's changed:
- The icon is now fully clickable as expected and should load properly. 
- An alternative text is also added which improves accessibility. 
- A spelling error for a comment inside of the setViewMode function is rectified.

## Change needed:
- The Dropdown component now contains inline-styling which is not allowed per the React Coding Standard. I suggest the styling to be nested inside of the dropButton class to not affect undesired images.

c22isako:
Removed inline-styling, and instead made the styling be nested within the dropButton class.

b23sebth (approved):
**Fixes #17048. Approved!**
## What's changed:
- Styling is removed from Dropdown component and moved into style.css.

PR #17095, opened 2025-04-30
Fix underfined array key warnings for output fields in microserviceUI.php. #17065

### LINKED ISSUE #17065: Fix underfined array key warnings for output fields in microserviceUI.php.
When viewing microservice details, the following warnings appear due to array key mismatches with the database:

Image

To fix the outdated array keys, either the database or microserviceUI.php needs to be changed. Changing the values in the UI would be more appropriate rather than changing the database. When checking installEndpointDb.php, you notice that the affected table "outputs" has been changed (in another issue) so that it is stored in a similar fashion to "parameters". I would therefore recommend displaying outputs in a similar way parameters is displayed, shown in the picture below:

Image

Output is no longer stored just as "output" and "output_description", it is now stored similar to parameters:
Image

a23oskme:
Added a query to get outputs from the outputs table and added a table that looks like the parameter table to present the outputs. Fixes #17065 

image

### DIFF
--- a/DuggaSys/microservices/endpointDirectory/microserviceUI.php
+++ b/DuggaSys/microservices/endpointDirectory/microserviceUI.php
@@ -104,8 +104,6 @@ if (isset($_GET['id'])) {
         <h2><b>ID: </b><?php echo $microservice['id']; ?></h2>
         <p><b>Description:</b> <?php echo $microservice['description']; ?></p>
         <p><b>Calling methods:</b> <?php echo $microservice['calling_methods']; ?>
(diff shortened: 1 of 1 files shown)


a23dilel (approved):
I can see that you have implemented a header table with "ID" and "Description", and also that "Calling methods" and "Microservices used" are show as parameters with "b"-tag styled text which looks very nice:

image

I also see that you have implemented a second table with headers "Name", "Type", and "Description" and values which looks good also:

image

I can see that "output" and "output_description" have been removed, because this issue #17065 mentioned that output is no longer stored in those names, but is now using to parameters.

Overall looks good!

PR #17096, opened 2025-04-30
Added autosave to input for textareas

a23kevre:
This functionality gives all input textareas an autosave feature so that when you change colour of the background for the entity the text is always saved beforehand. At the same time you write in the input textarea the entity recieves and updates the text inside it. 

Although this functionality makes the save button useless, it still solves the problem at hand:)


d22abdha (approved):
Changing the background colour no longer clears any of the other fields, so the original bug is fixed.
I did notice a new “Stereotype” input showing up in the properties panel. If that field isn’t meant to be part of the class, you may want to hide or remove it. Otherwise everything works perfectly

Issue #16979, opened 2025-04-28
Re-engineer microservice retrieveShowDuggaService


a23louma:
Follow the "coding standard for removing includes" for this file in folder /showDuggaServices. The coding standard can be found in a couple of files in /microservices/Microservices Documentation.

PR #17099, opened 2025-04-30
Re-engineer microservice retrieveShowDuggaService #16979

a23danme:
Fixing issue #16979


a22linre (commented):
Looks good! You could also use the curlService.php but this works too

b23johgu:
I was too fast with the merging heh.. I just saw that we should indeed use curlService.php according to issue [17108](https://github.com/HGustavs/LenaSYS/issues/17108). I can try to revert the merge or create a new issue so you can fix it.

PR #17100, opened 2025-04-30
G3-2025-V5-#17017-Replace-JQuery-Within-contribution.js-Line-510-to-1010

c22isako:
Couldn't find use for the files within the current version of LenaSYS. It is also mentioned within documentation:
https://github.com/HGustavs/LenaSYS/wiki/Contribution-Wiki-Page that it's no longer useable.

**For testing, confirm that it is no longer used.** 
This may pose problems since there aren't any apparent uses for the functionality at this time (from what I could find)


a22maxda (approved):
As far as I can tell contribution.js & .php is not currently in use, it's purpose was probably to log activity for groups but I assume that responsibility has been moved to GitHub. There seems to be no way to access it without using the address bar and there is nothing on the page. They can likely be safely deleted.

HGustavs:
It should not be removed.

This file is used through hot-linking in a course item, and if the test data does not have any such hot links, or if we do not find the links that may appear in the data, this item may appear to be unused.

Issue #17044, opened 2025-04-29
Missing tool-panel options


a23emisn:
When logged in as a non-admin user or logged out and trying to use the "Diagram dugga", some of the diagramming tool options are missing in the tool panel. This is may not be in line with the desired behavior of the application.

Logged in as teacher1 (generic user):
![Image](https://github.com/user-attachments/assets/33e97df3-6133-4f4f-96af-d5ad6aa625bb)
The lifeline options are gone, the IE entity is gone in the first submenu, and the other submenus only contain three selectable items instead of four.

However, if you log in with an admin account all options are available:

Logged in as brom (admin): 
![Image](https://github.com/user-attachments/assets/ad4e5da2-8fea-47cb-ba5b-4e3286a2c590)

Issue to solve: Find out whether this is the desired behavior (unlikely) and if not, fix it so that non-admin users can access the complete version of the diagramming tool.

PR #17101, opened 2025-04-30
Fix Missing tool-panel options #17044

a23matsh:
There was a function showDiagramTypes that was hiding certain tools when logged in as a non admin user, which was removed. Now even non admin users can access the complete diagram and tools. It was unclear what that function was necessary for in this diagram and when, also the description of it did not say much.


a22abdjo (commented):
The changes made has solved the issue and does not present any further problems, therefore this solution is approved.

a22abdjo (approved):
The changes made has solved the issue and does not present any further problems, therefore this solution is approved.

PR #17103, opened 2025-05-01
Replaced jQuery with JavaScript in pushnotifications.js

a23yutta:
Replaced jQuery with JavaScript in `/DuggaSys/pushnotifications.js` .

The functionality should remain unchanged. However, I was unable to verify whether it works because the profile page (of brom) shows the message "Notifications subsystems not installed, notifications unavailable.". When attempting to directly access the file `pushnotifications.php` from [https://localhost/LenaSYS](https://localhost/LenaSYS/DuggaSys/courseed.php) you are directed to this page:
Skärmbild 2025-04-29 144328 

Which does not say much. This behaviour is consistent across both the week branch and the issue sub-branch.

### DIFF
--- a/DuggaSys/pushnotifications.js
+++ b/DuggaSys/pushnotifications.js
@@ -1,6 +1,6 @@
 'use strict';
 
-$(function() {
+document.addEventListener('DOMContentLoaded', function () {
 
 	let sendPushRegistrationToServer = function (subscription, deregister) {
 		let action;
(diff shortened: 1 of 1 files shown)


c22isako (approved):
**I can not tests that this solution fully works. This since the error message: "success":false,"error_message":"Error: Notifications subsystems not installed, notifications unavailable" is present, which is also mentioned in the PR.** 

Whiles searching for a way to test this, I stumbled across PR https://github.com/HGustavs/LenaSYS/pull/15546 which mentions the requirement of push notification support installed to properly be able to test the functionality. 

Since I do not have the required "push notification support", and feel like this is a weird way to go about it since I'm unsure if we can fully confirm that it works without the support installed (which could affect the customers experience), the solution can not be confirmed. 

I do believe that the file can be used, and may be used. But that it should be made into a separate issue to fix the feature in itself, and therefore also confirm that the implemented changes within this PR actually works. This would also benefit testing, since it currently is troublesome and requires additional "packages" installed. In itself, the changes look good. 
_This has also been discussed with group leader, and issue coordinator._

Issue #17079, opened 2025-04-30
Fix the documentation of microservices for /sectionedService


a23louma:
The template has been updated(see the v5 branch) and the files documented so far for this folder have to be updated to follow the new template.

PR #17110, opened 2025-05-02
Updated documentation of microservices for /sectionedService

b23johgu:
Updated existing files in sectionedservice according to the new template.

Fixes issue #17079


a23edvto (approved):
Code was following the template and what was added to the files are right

Issue #17092, opened 2025-04-30
Implement hiding the toolbar for mobile version


a23emijo:
Implement the mockup from the issue #16866.

PR #17112, opened 2025-05-02
Fixed: Implement hiding the toolbar for mobile version

a23gabst1:
This hopefully fixes the issue regarding implementing a way to hiding the toolbar so it does not disturb the user experience and limit the space available for modeling. I have eliminiated most of the overflows that were visible on the mobile version before (see video 1) and this switch with hiding the toolbar happens when the width of the device is 414px or less than 414px. 

This fix could still need som further styling and added functionality for opening the sub-menus. 

Fixed the issue: #17092

The toolbar before (video 1):

https://github.com/user-attachments/assets/2a4c380d-716e-4c07-baaf-f0e22c13a739

The toolbar now (video 2):

https://github.com/user-attachments/assets/98a7cf97-0c8a-4704-ba75-8994cdfdd26e

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -301,6 +301,10 @@
 @media (max-width: 414px){
     .icon-wrapper{
         display: block !important;
+        position: fixed;
+        z-index: 10;
+        bottom: 7%;
+        left: calc(50% - 2%);
     }
 
     #diagram-replay-box{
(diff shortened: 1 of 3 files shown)


b23ebbst (approved):
Looks great and works as expected and shown in video 2! 
Good job, I approve! :) 

_**With ruler, toolbar opened**_
image

_**No ruler, toolbar opened**_ 
image

_**No ruler, toolbar closed**_
image

a23miksk (approved):
I also approve this change. Looked at the desktop version mainly and there seemed to be no change from the weekly branch.

PR #17120, opened 2025-05-02
Change createNewCodeExample to use curlService - Issue 17107

a20artje:
The code didn't work before but now it uses the right function, will create another issue to fix the problem.


a20artje:
I misunderstood the issue and will work on it a bit more.

a20artje:
I have fixed the issue for real this time.

a23arvme (changes requested):
Remove comments in githubCodeExample, _**AND**_
Change createNewCodeExample to use Recieve instead of CallPOST

a23arvme (approved):
Looks Good To Me.

PR #17123, opened 2025-05-02
added styling to mobile buttons

a22maxda:
buttons are smaller and equidistant, as well with slightly more spacing to reduce chance of mistapping. They look good on all the aspect ratios I have tried (most popular phone ARs). Also moved course name and code so it's not in the way. Since only styling has been changed I don't see how functionality should differ but I haven't found any issues.

Screenshot 2025-05-02 at 11 38 15

This is the old styling with larger buttons and overlapping text.

Screenshot 2025-05-02 at 11 39 22

This is the new styling with slightly smaller buttons and centered text.

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -7733,6 +7733,10 @@ only screen and (max-device-width: 568px) {
     font-size: small;
   }
 
+  .submit-button {
+    margin: 12px 2px 4px 2px;
+  }
+
   input.submit-button-newitem {
     font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
     font-weight: bold;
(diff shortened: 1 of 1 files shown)


a23osckr (changes requested):
Looks better and is easier to use. Looks fine down to around 335px width, after that the View dropdown collides with Hidden items button. The buttons should not collide until after 320px since this is the smallest commonly found screen size. The View button does not behave well below 500px width. Below this the button becomes unclickable on its right side, and the problem gets worse the smaller the screen size is. I believe the course-label div is covering the View button. Other than these two issues, everything looks good. Changes required.

a22maxda:
The view button doesn't get blocked by the course name now but I have no ideas for how to remedy the issue of buttons overlapping on small enough screens. I am not sure if it is necessary as phone screens generally are not that small but I will look for a solution if necessary.

a22maxda:
I decided to reduce the button size on small enough screens, they may be a little hard to tap and look a bit weird but they retain functionality and do not overlap until 310px in width which in my opinion is fine.

a22maxda:
The conflict has been resolved, should be fine now.

a23erigu (changes requested):
So the issue seems to be fixed with the buttons being able to work at 310px size.

But there seem to be a lot of issues with the course name and code.
The course name appears to be trying to fly away and the course code has basically escaped the screen entirely.
image

and if you start to scroll then the couse name and code seems to follow the screen.
image

I don't know if we want to fix these problems in this issue as it was supposed to be fixed in this issue [#17064](https://github.com/HGustavs/LenaSYS/issues/17064) but it's definitely a problem that need to be fixed in some way.

a22maxda:
The other issue is interfering with the soluion used in this one. If we want to use the other solution that code can be reinstated, however this solution works as far as I can tell.

Issue #16867, opened 2025-04-23
Re-engineer microservice insertIntoSQLite


a23louma:
Follow the "coding standard for removing includes" for this file in folder /gitCommitService. The coding standard can be found in /microservices/Microservices Documentation
/microserviceCodingStandard.md.

PR #17124, opened 2025-05-02
Re-engineer of insertIntoSQLite_ms.php

a23axehe:
Have Reengineered insertIntoSQLite its invert dependent file getCourseID. Post without answer implemented according to standard in file microserviceRefactorPOSTnoAnswer.md. Have tried testing with no result, the tables from the database which both files use, calling either empty rows or a database which does not seem to exist "gitRepos". Have tried adding a github repo to LenaSYS since the file likely does something with github but that function does not seem to be working on my branch nor the groups weekly branch, unsure how to test further. Not sure if files are used in LenaSYS yet. Fixes issue #16867 .

### DIFF
--- a/DuggaSys/microservices/gitCommitService/getCourseID_ms.php
+++ b/DuggaSys/microservices/gitCommitService/getCourseID_ms.php
@@ -2,8 +2,7 @@
 
 include_once "../../../Shared/sessions.php";
 include_once "../../../Shared/basic.php";
-include_once "./insertIntoSQLite_ms.php";
-
+//include_once "./insertIntoSQLite_ms.php";
 
 function getCourseID($githubURL) {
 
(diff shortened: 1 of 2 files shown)


a23axehe:
If someone else has a good idea of how to test this please do and leave a review :)

a22linre (commented):
On row 41 in getCourseID, "$token" is broken, missing name but I'm not sure if its supposed to be like that or not. That's the only thing I found. The code looks good otherwise. How to test it, I'm not sure either tbh.

a23axehe:
insertIntoSQLite($githubURL, $cid, "") this is the original line where token was left as "" aswell so I just tried to resemble that as good as possible with the replacement

a23axehe:
Have now fixed the token and set it to "" as it was before

Issue #16702, opened 2025-04-16
Swap the jquery in contribution_login.js to javascript


a23erigu:
Start swaping the jquery in contribution_login.js to javascript

from what I can see there is a LOT of jquery in this file 

If it seems to much for a singular issue then I can try to divide it upp

EdvinWerner:
I know this has been unasigned from me but i still looked into it and i think this issue is better solved by going through each method or function and changing it from Jquery to javascript

a23simry:
Good job keeping up with PRs but atm its not necessary for this issue! It's on hold until we've figured out and gotten the right contribution files.

a23osckr:
Yes, just wanted to make sure it's not forgotten!

PR #17128, opened 2025-05-02
G3-2025-v5-#16702  JQuery to JS in contribution_loginbox.js

a23osckr:
Closes issue #16702. All JQuery has been converted to JS in contribution_loginbox.js. Also added a function for Enterkey listeners to shorten code slightly. Not sure how to test if all works since browsing to contribution.php gives a blank page if logged in and a loginbox which does nothing if logged out. I believe it should show graphs, but I am unsure how to make the feature work on my local machine. There are however no errors in the console, which is a good sign.

### DIFF
--- a/DuggaSys/contribution_loginbox.js
+++ b/DuggaSys/contribution_loginbox.js
@@ -29,7 +29,7 @@ function contribution_newGitUserCreation()
     document.getElementById('newpassword').style.display = 'none';
     document.getElementById('UserExistslogin').style.display = 'none';
     document.getElementById('newGit-UserCreation').style.display = '';
-    document.getElementById('newGit-UserCreation_password1').focus;
+    document.getElementById('newGit-UserCreation_password1').focus();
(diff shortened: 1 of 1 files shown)


f23linan (changes requested):
Cannot at this time approve this pull request as its not possible to test the changes. Like Oscar put in the pull request the contribution page is weird on the local machines and not able to be reached. From what I can see the acutal code looks good however non-tested code cannot be merged.

EdvinWerner (changes requested):
As mentioned in the other review the code it looks good but since there is no way of actually testing it I csn't in good consience allow this to be merged. One needs to figure out a way to actually test the code.

a23fanro:
I have contacted András and Henrik about aquiring the file that is required for loading this, according to [](https://github.com/HGustavs/LenaSYS/wiki/Contribution-database-installation). Will update when I know more.

PR #17134, opened 2025-05-02
Re-engineered microservice retrieveFileedService #16980

a22linre:
Re-engineered retrieveFileedService and its inverse dependencies. Also formatted the code using the built-in formatter, so some changes might look more than they are beacuse of the formatting. 

This fixes issue #16980 .


a23danme (approved):
Seems to be working fine when testing through the UI. I am not 100% sure about the structure in the code according to the curlService but it looks good in my opinion. It works and it is using curlService and callMicroserviceGets so should be good.

Issue #17113, opened 2025-05-02
Empty strings and comma added in ER-table when there are no normal attributes


a23miksk:
Bug mentioned here: https://github.com/HGustavs/LenaSYS/issues/17036#issuecomment-2846617397 

> _I did notice that some weird things happen when there are no normal attributes. Normal ER entities gets an additional ", " without removing the added ", " for additional attributes, and the weak entities gets that plus a whole copy of their foreign primary key as a normal foreign key._

![Image](https://github.com/user-attachments/assets/237607d6-0575-4a29-b411-4e228639c574)

PR #17135, opened 2025-05-02
Fixed some issues with ER-table generation

a23miksk:
Fixed some issues with the string generation in the ER-table mentioned here https://github.com/HGustavs/LenaSYS/issues/17113
_Before_
image
_After_
image


b23ebbst (approved):
Looks good! No extra commas or other oddities appear when looking at the properties for the ER-diagram. Well done, I approve! 

image

Issue #17014, opened 2025-04-29
There are problems with the lines loaded from the example diagrams


a23gabst1:
There are issues regarding the lines in the example diagrams. The lines are not rendered correctly when zoomed out, and find out the cause for errors generated in the console.log. Some of the lines in the example diagrams are not even rendered on the diagram (e.g. there should be lines from staff waiter and chef because of inheritance). 

![Image](https://github.com/user-attachments/assets/d8213a3d-b90c-4ee7-ad91-0d424335c1c9)
![Image](https://github.com/user-attachments/assets/bc1bcccf-4c8e-4683-9965-456d17190aa0)

b23ebbst:
Zooming out breaks lines in _**all**_ diagrams, both example diagrams (as mentioned in this issue), but also in diagrams that you have created, see images below. 

_**1x zoom**_
![Image](https://github.com/user-attachments/assets/199d922f-e354-430a-a38c-8c2452288102)

_**0.5x zoomed out**_ 
![Image](https://github.com/user-attachments/assets/13910b3e-219c-423a-ba65-bafb5a675a08)

PR #17136, opened 2025-05-02
17014 there are problems with the lines loaded from the example diagrams

b23danhe:
This solves a lot of the problems in issue #17014.

I've added an if block that recalculates the fx, fy, tx, and ty coordinates when the line is connected to or from an IE relation element. I also refactored the logic slightly to improve readability and maintainability.

The final change in this PR is the implementation of zoomfact to the stroke-width attribute in the SVG elements. This ensures that the line thickness scales proportionally with the zoom level.

~~The only remaining issue, or issue caused by this PR, is that in some ER relations, the lines still don't fully connect to the element as expected (watch end of video). Will add a new issue for that.~~

https://github.com/user-attachments/assets/a27c6e82-bb1b-45fa-91ea-601e980d04a2

### DIFF
--- a/DuggaSys/diagram/draw/line.js
+++ b/DuggaSys/diagram/draw/line.js
@@ -4,6 +4,7 @@
  * @param {Object} line The line object that is drawn.
  * @param {boolean} targetGhost Is the targeted line a ghost line
  */
+
 function drawLine(line, targetGhost = false) {
 
     let lineStr = ""; // only the lines, polylines, arrows etc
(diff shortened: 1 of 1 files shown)


a23thefo (approved):
The PR works as intended. The code structure is following the guidelines. The comments are well described. Good job 👍 

### Demo

https://github.com/user-attachments/assets/12977720-48d2-4b52-b4f4-d2e2cbdad1f9

PR #17139, opened 2025-05-02
Documented getAccessedService_ms and retrieveAccessedService_ms

### LINKED ISSUE #17118: document getAccessedService_ms.php and retrieveAccessedService_ms.php in microservices/accessedService
Document these two files in /accessedService. Follow the new updated template(check the branch of this week). When you're done create another sub-issue for the following files of this folder.

microservices/Microservices Documentation/accessedService.md

a23wilgr:
Documented getAccessedService_ms and retrieveAccessedService_ms

fixes #17118

### DIFF
--- a/DuggaSys/microservices/Microservices Documentation/accessedService.md
+++ b/DuggaSys/microservices/Microservices Documentation/accessedService.md
@@ -148,7 +148,7 @@ Checks if a user exists based on the username, creates a new user if it not and 
 
 
 
-# getAccessedService_ms.php
+# name of file/service
 getAccessedService_ms.php
 
 ## Description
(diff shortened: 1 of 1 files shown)


b23johgu (commented):
The names for each file must be like this according to the newest version of the template:

```
# Name of file/service
example_ms (replace this line with the name of the microservice)
```

I'm not sure if opt could be a string (char/varchar) and not an int? As it is pre-filled as an empty string in the parameters that are sent to retrieveAccessedService.

`echo json_encode(retrieveAccessedService($pdo, $debug, $userid, $cid, $log_uuid, $opt="", $newusers=""));`

I also think that log_uuid would be a char or varchar as they are typically stored as that in mysql databases

b23johgu (approved):
Looks good now! Headlines are correct

PR #17145, opened 2025-05-02
Documented updateUser_ms and updateUserCourse_ms

a23wilgr:
Documented updateUser_ms and updateUserCourse_ms

### DIFF
--- a/DuggaSys/microservices/Microservices Documentation/accessedService.md
+++ b/DuggaSys/microservices/Microservices Documentation/accessedService.md
@@ -145,3 +145,111 @@ Checks if a user exists based on the username, creates a new user if it not and 
 ### Microservices Used
 getUid_ms.php
 retrieveAccessedService_ms.php";
+
+
+# Name of file/service
+updateUser_ms.php
+
+## Description
+Handles updating user's properties in the 'user' table. Can be only be done by superusers.
+
+## Input Parameters
+- Parameter: $opt
+   - Type: String
+   - Description: Operation option
+
+- Parameter: $prop
+   - Type: ?
+   - Description: Property to update
+
+- Parameter: $courseid
+   - Type: int
+   - Description: describe parameter
+
+- Parameter: $uid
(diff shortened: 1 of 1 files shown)


a23danme (commented):
On input parameters in documentation I think it's like a "unwritten rule" that the "Type" is what is expected and does not necessary need to be what it is. An example could be:
- Parameter: $group
   - Type: varchar
   - Description: Group
Where the "Type" VARCHAR could be right practically in the database such as "groups VARCHAR(25)" or something. But in documentation I think it should be written as "string" due to the expected value is just letters. I might be wrong but I searched it up aswell and it seems to be right. 

Overall I think the documenation is  good, sure the "Example of use" could have something but these 2 files doesnt really do much outputs and I dont even know what should be there it just says "CODE" in the template so I think it is fine to let them be blank.
 

On updateUser I wrote the changed parameters so you can just copy and paste them in if you would like to. Maybe on updateUser also add getUID due to you added it to the other files microservices. Also the "sessions" should maybe be included in the microservices? 

Here is the parameters for updateUser:

## Input Parameters
- Parameter: $opt
   - Type: string
   - Description: Operation option

- Parameter: $prop
   - Type: string
   - Description: Property to update

- Parameter: $courseid
   - Type: int
   - Description: describe parameter

- Parameter: $uid
   - Type: int
   - Description: User ID

- Parameter: $firstname
   - Type: string
   - Description: Firstname

- Parameter: $lastname
   - Type: string
   - Description: Lastname

- Parameter: $ssn
   - Type: string
   - Description: Social security number

- Parameter: $user_name
   - Type: string
   - Description: Username

- Parameter: $classname
   - Type: string
   - Description: Class (program) name

- Parameter: $log_uuid
   - Type: string
   - Description: Unique identifier for logging purposes

On updateUserCourse all the parameters beside courseid should in my opinion be "string", and where courseid should stay as "int"

a23danme (commented):
Looks good!

Issue #17086, opened 2025-04-30
The event listener for keypresses in the diagram doesn't have any functionality to prevent registering multiple button presses from holding down a button


a23emijo:
The event listener for keypresses in the diagram doesn't have any functionality to prevent registering multiple button presses from holding down a button. This usually isn't noticeable since not many keybinds are bound to anything that could happen multiple times. However with the new quicksave function the save happens instantly, causing it to happen multiple times if the button is held down for too long. This doesn't seem to have any impact on functionality but is jarring to see when trying to save a file.

PR #17147, opened 2025-05-02
[#17086] Fixed issue with quicksave keybind triggering constantly when held down

a22nevan:
**Quicksaving should no longer trigger constantly when holding down the keybind, and will only trigger once upon releasing the keys.**

This was a slight miss on my part when I initially reported it, not realizing that the "functionality" was there, just not very well documented. There's another function that worked in the same way, which was centerCamera(). This action also happens instantly and only happens once. This was due to not calling upon the function in the keydown-eventListener, instead only calling upon it in the keyup-listener, performing the action once when the key was released. The same functionality was applied to the quicksave keybind, only calling upon the function when the key is released. 

This _does_ feel slightly jarring, with nothing happening upon pressing the key, and only getting any feedback once the key has been released, but this is a simpler way of solving the issue without having to tamper with the already in place system which could mess with the entire eventlistener as a whole. 

**I also added relevant comments to both eventlisteners to add some much needed explanations and context. This should hopefully make them easier to understand and grasp.**

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -584,6 +584,8 @@ document.addEventListener('contextmenu', event => {
     event.preventDefault();
 });
 
+
+//Eventhandler that checks if key has been pressed, needs to be reworked to accomodate the new quicksave.
 document.addEventListener('keydown', function (e) {
     if (isKeybindValid(e, keybinds.LEFT_CONTROL) && !ctrlPressed) ctrlPressed = true;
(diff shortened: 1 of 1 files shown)


a23matsh (approved):
Confirmed working solution. Previously the system would continuously save while the quicksave key was held down. With this implementation, holding the key triggers the save only once upon release. The code has also been properly documented, making the functionality easier to understand and follow compared to before.

Issue #17075, opened 2025-04-30
Snap to grid doesn't align elements with the grid's x-axis


a23aleva:
When using keyboard arrow movement on elements in snap to grid-mode there seems to be a slight mistake to the alignment on the grid's x-axis. When moving the element along the y-axis it works as intended, where the element moves along the actual grid lines which makes it easier to align elements. The same should be the case when moving the element sidewise along the grid.

https://github.com/user-attachments/assets/24d8a3a2-d717-4070-a6f2-d986c1e2de1f

PR #17153, opened 2025-05-05
[#17075] snap to grid doesnt align elements with the grids x axis

a23aleva:
When using snapToGrid rectangular elements now snap to grid lines instead of snapping irregularly beside them. Other shapes snap to their center. Overall it gives the user an easier time to line up elements on the grid which seems to be the original purpose of the feature. I also made some changes to note and added elementTypeNames to its kind to match the other elements. For some reason note was previously just a string.

https://github.com/user-attachments/assets/92aeae88-6d4d-45ab-8845-0ed95f0744a3


a23gabst1 (approved):
Works as intended! The elements as of now snap to grid lines instead of irregularly beside them.

a23thefo (changes requested):
Something with the merge conflict solution broke the aligning on the x axis 

image
image

Issue #17116, opened 2025-05-02
Fix the rulers position so it adapts to mobile versions


a23gabst1:
Fix so that when you toggle the ruler when on the mobile version the position of the ruler is correct, see figure 1. The video demonstrates the how it works now and how it should work (used the inspector beside to illustrate how it should toggle on mobile versions). 

How it should look like: 
![Image](https://github.com/user-attachments/assets/1868c358-4a72-4073-8742-dbbc290ae56a)

How it look like when toggling:

https://github.com/user-attachments/assets/a219b5e0-5f02-4c0e-b3e4-69fb5ef67859

PR #17154, opened 2025-05-05
Fixed: Fix the rulers position so it adapts to mobile versions

a23gabst1:
I have now fixed the rulers position so it adapts to both desktop and mobile version. So, when you toggle on both version they are positioned correctly based on their environment (other elements e.g. navbar on desktop). And if you resize the window they are positioned correctly so you dont need to toggle the ruler to make the position correct. see video 1 to see the result. 

PS: The code is now auto formatted/indented

fixed the issue: #17116

video 1

https://github.com/user-attachments/assets/404e4967-546f-44e3-8090-9a3121f7fbac


a23emijo (approved):
This works as intended and is approved! 👍 

**_How it looks when testing_**
https://github.com/user-attachments/assets/ab086049-3f41-4e29-9b0a-5486c4d33f4d

PR #17155, opened 2025-05-05
G3-2025-V5 #17142 Fix course creation error/prevention

### LINKED ISSUE #17142: Fix Course Creation error/prevention
### There is currently no way to create a "Functional" course (unknown error)

Recommended course for trying to diagnose problem is firstly by trying to create a new course and then checking the console for errros as to what is going wrong, 

### Note! creation of the course works, but futher modification is impossible as there are a bunch of errors when clicking into it and trying to continue with adding the information

I also recommend looking for the reason that an error toast appears at all when trying to enter the MOTD (Message Of The Day)

a23lisjo:
Fixes #17142. 
The function that validates the 'Message of the day' text had a missing variable (errorMsg) that is now declared. 
image

For testing I created a new course, filled in the necessary information and saved the information and started adding items to the course. If set to default, the created version will always be loaded, otherwise the same information has to be filled in again. 

The screenshot below shows that the editing for a course now works:
image

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -3816,10 +3816,11 @@ function validateCourseID(courseid, dialogid) {
   return false
 }
 
-function validateMOTD(motd, syntaxdialogid, rangedialogid) {
+function validateMOTD(motd, syntaxdialogid, rangedialogid, dialogid) {
   var emotd = document.getElementById(motd);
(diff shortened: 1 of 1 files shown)


a23erigu (approved):
The chance has made it possible to fully create courses now so this will be approved

I have found that the create and delete in the new course seems holy broken though so that will need to be made into an issue

Issue #16977, opened 2025-04-28
Re-engineer microservice loadDugga


a23louma:
Follow the "coding standard for removing includes" for this file in folder /showDuggaServices. The coding standard can be found in a couple of files in /microservices/Microservices Documentation. Now the functions in DuggaSys/microservices/curlService.php can be used.

PR #17157, opened 2025-05-05
Re-Engineer of loadDugga_ms.php

a23axehe:
Removed includes and replaced with GET. Have tested a bit and cant find any irregularities when accessing a dugga which was a issue I had before the code was not working properly. Solves issue #16977 .


a22wilas (changes requested):
The solution solves the Issue.

Tested to save a Dugga and use the "Hash" and "Password" from the database.
bild
bild

Loaded the Dugga and got the correct result. 
bild

The one change that can me made is remove the code row 6 from "loadDugga_ms.php"
include_once "retrieveShowDuggaService_ms.php";

This row is now redundant and this inverse dependence is solved by the solution!

a23axehe:
@a22wilas I think loadDugga is dependent of retreiveShowDuggaService aswell i.e they are dependent on each other?

a23axehe:
My issue was just to fix one of dependencies

a22wilas:
I tested it with removing the code line mentioned in the review with success but as you mentioned, your issue was to remove the showDugga dependence so will change the review.

a22wilas (approved):

Issue #17141, opened 2025-05-02
Wrong tool gets highlighted in sub-menu


a23emisn:
When selecting the "Sequence lifeline (object)"-tool (second tool from the left) in the sequence sub-menu, another tool: "Sequence activation" (third tool from the left) gets highlighted instead in the sub-menu. See video below for demonstration.

https://github.com/user-attachments/assets/8cc4e608-2e80-40e6-893f-120e6160c7d8

This is incorrect and should be patched so that the correct tool gets highlighted in the sub-menu upon selection.

PR #17158, opened 2025-05-05
Fixed: Wrong tool gets highlighted in the sub-menu

a23gabst1:
The solution was to remove an class from the sequence activation box in the dropdown that was used to mark what element is active, and then add that class to the correct element that was being active. see figure 1 for result. 

The issue that was fixed: #17141


a23eliax (approved):
I confirm that this solution works as expected!

https://github.com/user-attachments/assets/95bb053e-7b72-4605-97ec-11caf9fabea0

PR #17161, opened 2025-05-05
Added general error handling for events. Use Windows with a Docker.

### LINKED ISSUE #16992: Add general error handling for events
Lenasys can produce many different errors, so it's a good idea to implement that if none of these four errors are the cause, then show the actual error message in a modal.

a23dilel:
Fixes #16992. 

There are four errors which are already implemented which are db connection, copy course files, creation of database and sql error.

I have added operation failed modal, because if none of these four errors are the cause, then show the operation error message in a modal.

To test operation error modal, you need to remove else if to show or

### DIFF
--- a/newinstaller/installer_ui.php
+++ b/newinstaller/installer_ui.php
@@ -242,6 +242,7 @@
 							$modal->showFilePermissionErrorModal("permissionError");
 							$modal->showDbCreationErrorModal("dbCreationError");
 							$modal->showSqlExecutionErrorModal("SqlError");
+							$modal->showOperationErrorModal("operationError");
 						?>
 						<div id="pageButtonContainer">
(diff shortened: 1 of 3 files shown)


a23dilel:
Ok, I know how to get a different error. To get this, follow these steps in new Lenasys installer:
Step 1: Select Windows.
Step 2: Select both checkboxes.
Step 3: Database name: lenadb, MySQL user: Daniel, Hostname: db, MySQL user password: 1234. Check "Use Distributed Environment" and uncheck the rest.
Step 4: Root username: root, Root password: password.
Step 5: Select all options.
Step 6: Click the install button.
Complete installation: press the finish button.
Then go back to the Lenasys installer and do the same thing with same values, but now check "overwrite existing database" in the step 3. 

Then you get a this error:
image

a23axehe (approved):
image

Got the correct error message when trying to overwrite the database with an existing user. Works as it should. OK

PR #17164, opened 2025-05-05
Added edit functionality

### LINKED ISSUE #16967: Implement function for editing a microservice for the Microservice directory UI

a23bensh:
Fixes #16967 and adds edit functionality to the microservice UI. Editable attributes include the microservice name, description, calling methods, and microservices used. The parameters and outputs values are intentionally not editable through this form due to complexity and the need to maintain the structure/associations between microservices. IF these attributes should be made editable despite the potential issues, I recommend creating a separate issue for implementing their edit functionality.

### DIFF
--- a/DuggaSys/microservices/endpointDirectory/microserviceUI.php
+++ b/DuggaSys/microservices/endpointDirectory/microserviceUI.php
@@ -1,4 +1,10 @@
 <?php
+session_start();
+
+if (!isset($_SESSION['token'])) {
+    $_SESSION['token'] = bin2hex(random_bytes(32));
+}
+
 try {
 // database
 $db = new PDO('sqlite:endpointDirectory_db.sqlite');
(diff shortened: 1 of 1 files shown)


a23petny (changes requested):
Add a CSRF token check before merge.
Generate a random token once per user session and store it in the session. Include that token in the edit form as a hidden field. When the form is submitted, compare the submitted token to the session token using a constant-time comparison. If they differ, stop the request with HTTP 403.

a23bensh:
I added the token functionality which can be tested through something like inspect element and changing the token value. I am a little unsure as to why this is relevant/needed for what it is we are trying to create as of now, but I understand if it's for preventing problem with future implementations (such as login/user sessions).

a23dilel (approved):
I have tested with the edit button, changed some values, and pressed the save button. It updated and worked perfect.
student

I have also tested the CSRF token by changing token values, which result in the error “Invalid CSRF token.” This works as intended.

csrf2

I considered editing the ID and checking for duplicate ID, but I guess the ID increments when adding a microservice, so implementing this is not necessary. This is just an option for consideration.

Overall looks good!

Issue #17130, opened 2025-05-02
sequence-actors and objects connects to eachother when they are becoming a gruop


a23eliax:
When marking sequence-actors and sequence-objects for replacing them as a group, they connect to eachother. The error-waring comes up correctly when overlapping but they shouldn't connecnt to eachother.

https://github.com/user-attachments/assets/a5a5f2c9-efae-4fb9-b16e-c851925821fd

PR #17168, opened 2025-05-05
G1 2025 v5 #17130 sequence actors and objects connects to eachother when they are becoming a gruop

a23eliax:
After testing many different things, I finally found out that the only thing needed was to add "if (pos.kind !== elementTypesNames.sequenceActivation) return null;". This ensures that the snapping logic is only applied to elements of the type sequenceActivation. If the current element is of any other type, the function will stop execution at this point and return null.


a23vinge (approved):
As described in issue “sequence-actors and objects connect to each other when they become a group #17130”, the problem occurred when using box selection, as shown in the video. This caused the sequence lifelines to compress and appear distorted. The solution proposed in this  issue—adding the following code:

if (pos.kind !== elementTypesNames.sequenceActivation) return null;

successfully resolves the problem. With this change, selecting multiple actors no longer causes layout issues. I have tried moving around sequence actors and objects  with the box selection and I do not find any problem

That said, I did notice a minor usability issue: occasionally, when moving selected elements (even slightly), a warning appears stating:
"You can't place elements too close together."
The behavior seems inconsistent—I tested various speeds and distances, but couldn’t identify a clear pattern. It may be related to how close the moved elements end up relative to one another, or possibly due to a snapping threshold.

While this doesn’t affect the resolution of the original issue, it might be worth tracking separately as a minor bug.

https://github.com/user-attachments/assets/54d80a21-78c8-4bc4-9161-bd2820f34107

Issue #16944, opened 2025-04-25
Re-engineer microservice retrieveProfileService


a23louma:
Follow the "coding standard for removing includes" for this file in folder /profileService. The coding standard can be found in /microservices/Microservices Documentation
/microserviceCodingStandard.md.

PR #17175, opened 2025-05-05
Refactored retrieveProfileService to use POST

a23gabli:
Solves issue #16944


EdvinWerner:
I think I've been incorrectly assigned on this one, I'm not a part of group 2 therefore I'm removing myself.

PR #17187, opened 2025-05-05
Issue 17180 fix button styling

a23jesja:
I implemented a cleaner approach to button styling by creating CSS variables for common properties like background color, text color, font, and size. By replacing duplicate values with these variables in the three main button styles, I've made the code more maintainable while keeping each button's unique properties intact.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -4829,7 +4829,7 @@ function getLocalStorage() {
 function setViewMode(mode){
   const section = document.getElementById("Sectionlisti")
 
-  // Remove previously applie mode
+  // Remove previously applied mode
   section.classList.remove("scroll-mode", "overview-mode");
 
   // Apply selecte view mode
--- a/Shared/components/Dropdown.js
+++ b/Shared/components/Dropdown.js
@@ -9,7 +9,7 @@ const Dropdown = ({ children, dropdownName, onClick, className = "", id="" }) =>
     return (
         <div className={`dropdown ${className}`}>
             <button onClick={onClick} className="dropButton submit-button">
-                <img src="/LenaSYS/Shared/icons/hamburger_menu_google_fonts.svg"/>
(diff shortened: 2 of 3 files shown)


b23sebth:
This seems to mostly be a duplicate of https://github.com/HGustavs/LenaSYS/pull/17093 which is already approved and merged.

a23fanro:
@b23sebth I see exactly what you mean. This should have already been implemented since [PR #17093](https://github.com/HGustavs/LenaSYS/pull/17093), but for some reason the code does not seem to be there. Something have gone sideways with the merge. But the 'original' PR has been merged (but for some reason it's not really merged) and c22isako got activity for their work, so I don't see any harm in this PR. 
And a23jesje seem to have added some improved styling, I will recommend this for review.

b23sebth:
@a23fanro From what I can see, the [PR](https://github.com/HGustavs/LenaSYS/pull/17093) got merged into the master branch instead of the current group branch. So while it's not currently in our weekly branch, I imagine there could be a conflict during the weekly merge. But like you said: there seems to be more changes than the original PR included.

a23erigu (changes requested):
I will start by saying that this code does work but there are a two problems I can find with the solution.

First of, the variables you created are called button-text, button-font etcetera.
From what I can see these variables are only used for submit buttons so calling them just button can makes it a bit confusing for future developers who might think it affect all buttons.

The second is the variables button-bg and button-test which when created are simply set as the colors color-primary and color-text-header respectively.
These two variables are then set as the colors for all the submit buttons.
Why don't you just use color-primary and color-text-header instead of needing these extra variables?
You even swap out color-primary and color-text-header to button-bg and button-text in some places?
This seem like unnecessary bloat as there already are variables that do the exact same thing

a23jesja:
Based on feedback, I simplified my approach by removing unnecessary variables and directly standardizing the redundant button styles.

a23alera (approved):
The changes look correct and address the previous feedback. The code works as expected

Issue #17140, opened 2025-05-02
Tool sub-menu item order gets rearranged


a23emisn:
When using the sub-menu for ER-relation, UML-inheritence symbols etc, selecting any item but the first in the list reorders the positions of the selectable items. See the pictures below:

Initially, the order of the items looks like this when the ER-relation is selected:
![Image](https://github.com/user-attachments/assets/5380571b-8cde-4bdf-816a-f36ffd13c607)

If you go to select any other item, the items rearrange in the sub-menu:
![Image](https://github.com/user-attachments/assets/e65d7989-bc8e-4044-9768-f7ad841798bb)

If you then reselect the ER relation, (first element) the order resets to its correct order.

The problem to solve here is to get rid of this unsuitable behavior by making sure the items won't reorganize upon use.

PR #17191, opened 2025-05-05
Fixed sub-icon order

a23kevre:
I changed so that the sub-icon order for the relations always are the same order. take a look at issue #17140 to see how it looked before the fix.


a23eliax (approved):
I confirm that this is a working solution! The sub-icons no longer changes places.

https://github.com/user-attachments/assets/c0c1a42f-aa48-4acb-8bef-75069885ce1b

PR #17195, opened 2025-05-06
Changed callMicroserviceGET function parameter in curlService.php #17186

### LINKED ISSUE #17186: Change parameter of callMicroserviceGET function in curlService.PHP
The parameter in this function is at the moment "$microservicePath" and should be changed to "string $path".

a23louma:
There was an error in the callMicroserviceGET function in curlService.php. I have changed the parameter "$microservicePath" to "string  $path" since it's used in the URL and now the function should work.

fixes #17186


a23dilel (approved):
The parameters are now correct and looks good!

Issue #17162, opened 2025-05-05
Implement FAB "button"


b23danhe:
The customer has approved the mockup for the FAB, and it is now ready for implementation. This issue focuses solely on adding the button with the correct placement, as shown in the mockup https://github.com/HGustavs/LenaSYS/issues/16813#issuecomment-2847517767

FAB has predefined code in Section editor, try to reuse.

The sub-menu functionality will be addressed in a separate issue.

![Image](https://github.com/user-attachments/assets/e3321fd9-9fb6-4b4a-8139-c8440c97fa44)

Issue #16813, opened 2025-04-22
Mobile FAB in the modeling tool


HGustavs:
Many mobile apps have a FAB for mobile users, do we have anything in the diagram we can put behind a FAB for mobile users?

FAB has predefined code in Section editor, try to reuse.

**The FAB should only have Save, Save As and Load**

HGustavs:
Mockup or code

HGustavs:
i would say that chevron and fab menus do two different things?

a23emijo:
> i would say that chevron and fab menus do two different things?
> 
> Can we do a mock-up that does not use placeholder icons for this?

This mockup is implemented with "real" icons #17112

a23gabst1:
This is a quick mockup demonstration on how the FAB could look like and function on mobile version. The FAB is placed in a position to not disturb the user when modeling and not take space from the diagram especially when there is limited width and height on mobile versions. It is also not placed to low on the screen to disrupt the user experience. Exclude the "Check" and it is a complete mockup. The colors of the different content boxes are different to enhance the user experience, but it is possible to have them all share a color for instance a lighter shade of purple. 

figure 1 shows how the FAB is positioned on the screen
![Image](https://github.com/user-attachments/assets/fd0369b6-c940-41ac-bf30-185bd3496412)

figure 2 focuses more on showing how the content is being displayed when visible
![Image](https://github.com/user-attachments/assets/692eb32c-d515-4777-8e49-bfe32fcfca81)

video 1 is showing how the FAB toggle works 

https://github.com/user-attachments/assets/b2350e06-5f3a-483d-bacb-48bd9e15b129

HGustavs:
This looks great

PR #17199, opened 2025-05-06
Fixed: Implement FAB "button"

a23gabst1:
I have implemented the FAB-btn. The FAB-btn as of now is correctly positioned based on the mockup https://github.com/HGustavs/LenaSYS/issues/16813#issuecomment-2847517767, and the FAB-btn is only being displayed on mobile versions and not desktop or tablet versions. More specifically when the screens width is 414px or less than that, see figure 1.  I have also tried to reuse as much as possible from the other fab-btns available on section editor. 

Fixed the issue: #17162

figure 1 
image

### DIFF
--- a/DuggaSys/diagram.php
+++ b/DuggaSys/diagram.php
@@ -101,6 +101,11 @@
         <i class="material-icons toggle-chevron">keyboard_arrow_up</i>
     </div>
 
+    <!-- The FAB-btn for the diagram.php, STARTS HERE!-->
+    <div class="fixed-action-button diagram-fab">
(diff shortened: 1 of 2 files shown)


a23vinge (approved):
The addition of the FAB button looks good—it’s positioned correctly according to the mockup in [#16813 (comment)]

(https://github.com/HGustavs/LenaSYS/issues/16813#issuecomment-2847517767), and it's appropriately limited to display only on mobile (screens 414px wide or smaller), as shown in the image. 

Also, the added code comment is a great touch—it provides helpful context for both future developers and for me as a reviewer.


image

PR #17200, opened 2025-05-06
Loading file and saving

a23aleva:
When the user loads another file and proceeds to make changes and save that save now targets the correct loaded file instead of the previous one. 

https://github.com/user-attachments/assets/9bcadeb0-db2c-460c-a706-93e3986fc484


a23matsh (approved):
This is a working solution, quicksaving now targets the active file and saves it there instead of the master version where you could have two files "Test1" and "Test2" for example, you went into Test1 and made two circles then you went into "Test2" and made one circle then it would save to the previous one "Test1" so "Test1" would get that save and nothing would happen in the active file.

Issue #17114, opened 2025-05-02
Fix callMicroserviceGET returntransfer options


a20artje:
The line:

    curl_setopt($ch, CURLOPT_RETURNTRANSFER, false); 

sets the return to false but on the next line:

    $response = curl_exec($ch);

A return is expected. CURLOPT_RETURNTRANSFER should be set to true.

PR #17204, opened 2025-05-06
Fix return options on curlService Get function - Issue #17114

a20artje:
Put the return option to true instead of false on the GET function.

Fixes issue #17114


a23louma (approved):
This looks correct, I approve!

Issue #17156, opened 2025-05-05
The rulers position is out of place when reloading the page on mobile version


a23gabst1:
The problem is that when the page is reloaded on the mobile version the ruler is out position, see video 1 to see the problem visually.

video 1 

https://github.com/user-attachments/assets/276e034a-bae0-45b4-97a0-2aecc890707e

PR #17209, opened 2025-05-06
Fixed: G1 2025 v5 #17156 the rulers position is out of place when reloading the page on mobile version

a23eliax:
The issue should be solved.

Had to change a few things:
- Removed the zoom container when media=414px because I thought it just took up space, and I assume the user uses the fingers to zoom in and out on the phone. But if someone else thinks it should be there, it's an easy fix to put it back.
- Since the toolbar was removed from the left side, I had to change the ruler overlay to "left: 0" to remove the gap between the grid and the ruler on the left.

https://github.com/user-attachments/assets/68f8d9c5-e0e1-46d9-be4f-8ada7bc26cd9

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -326,24 +326,28 @@
     }
 
     #rulerOverlay{
+        left: 0;
         position: fixed !important;
     }
 
-    #diagram-toolbar{
-        position: fixed !important;
+    #ruler-y-svg {
+        left: 0;
+    }
+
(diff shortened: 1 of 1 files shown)


a23aleva (approved):
The ruler doesn't cut off half the screen anymore but instead stays in a fixed position to the left side (and on top) which looks nice and fixes the issue. Approved!


https://github.com/user-attachments/assets/64449e53-75fe-4ad5-937b-949540fa1b4d

PR #17215, opened 2025-05-06
updated readCourseVersions to use cURL

a23arvme:
cURL for these files!


a23gabli (approved):
The code looks good and none of the functionality is broken.

PR #17217, opened 2025-05-06
Added a command text "docker compose up --build" for macOS

### LINKED ISSUE #17214: Add a commando text "docker compose up --build" for macOS in our Docker installation documentation.
We can add the text "docker compose up --build" for macOS in our Docker installation documentation. The [README](https://github.com/HGustavs/LenaSYS/blob/master/README.md#run-containers-in-docker) is already fixed  #17137. 

However there are two that we need to change which are [Docker installation instructions](https://github.com/HGustavs/LenaSYS/blob/master/dockerenv/document/docker_install_instructions.md#run-containers-in-docker), and the [Docker installation instructions on the GitHub wiki](https://github.com/HGustavs/LenaSYS/wiki/Docker-installation-instructions#run-containers-in-docker) in step 5.

a23dilel:
Fixes #17214. I have added macos command in Docker installation instructions and [Docker installation instructions on the GitHub wiki](https://github.com/HGustavs/LenaSYS/wiki/Docker-installation-instructions#run-containers-in-docker) in step 5.

### DIFF
--- a/dockerenv/document/docker_install_instructions.md
+++ b/dockerenv/document/docker_install_instructions.md
@@ -39,7 +39,8 @@ When you have installed your Docker desktop, you can run Docker containers, but 
 4. Now you have the LenaSYS repository, so you need to access the new directory in the terminal by typing: `cd LenaSYS/dockerenv/`.
 ![Change directory](terminal3.png "Change directory")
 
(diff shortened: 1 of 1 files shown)


a23bensh (approved):
This change does what it's supposed to and gives clearer instructions when using docker compose, looks good! 
Although it doesn't really concern this PR, I would suggest going over the grammar one more time in "NOTE" sometime maybe in another issue and changing it to something along the lines of "NOTE: You need to keep the Docker Desktop program/app open and not turn it off; otherwise, this command will show an error and fail to work.", but that's just nitpicking.

Issue #17212, opened 2025-05-06
Add foreign key from microservice table to dependencies table in endpoint database


a23oskme:
Add fields to the tables in the database and update dependencies fill script to get id from microservice (based on name) and add it to the insert

PR #17211, opened 2025-05-06
Add database download button to microserviceUI #17188

a23oskme:
Added a button that allows the user to download the database. Fixes #17188 

image

PR #17228, opened 2025-05-06
Add foreign key from microservice table to dependencies table in endpoint database #17212

a23oskme:
Added foreign key fields in the dependencies table. These can however be null and are intended to make queries easier. 

There are some inconsistencies in how the documentation is created and stored in the database. Some documentation has '_ms' in the name, some has '_ms.php' and some doesn't have any of them. In order to solve this, and still match microservice names to get the id and be able to insert the id in the dependencies table, I created a function. This function check the database for all the variants of the name and returns it when it finds it (or null if it didn't, since this usually means that there isn't any documentation yet). 

The microservice documentation isn't complete at the moment, so the test-data is limited. You can see the results below, but when more documentation has been created, more id's will appear. 
image

To test this, first run 'setupEndpointDirectory.php'. You can then see the data in 'dependenciesList.php', and check if there is data under microservice id and depends on id. 
It would be great to also compare these id's to those that are visible in the 'microserviceUI.php' as well, to ensure that the microservices that has documentation gets registered in the dependenciesList.php. (I did this during testing, but it would be good to get a second opinion).

### DIFF
--- a/DuggaSys/microservices/endpointDirectory/dependenciesList.php
+++ b/DuggaSys/microservices/endpointDirectory/dependenciesList.php
@@ -8,13 +8,15 @@ $dependencies = $query->fetchAll(PDO::FETCH_ASSOC);
 
 echo "<h2>Dependencies in database</h2>";
 echo "<table border='1''>";
-echo "<tr><th>ID</th><th>Microservice</th><th>Depends on</th><th>Path</th></tr>";
+echo "<tr><th>ID</th><th>Microservice</th><th>MS ID</th><th>Depends on</th><th>Depends on ID</th><th>Path</th></tr>";
 
 foreach ($dependencies as $dependency) {
     echo "<tr>";
     echo "<td>" . $dependency['dependency_id'] . "</td>";
     echo "<td>" . $dependency['ms_name'] . "</td>";
+    echo "<td>" . $dependency['microservice_id'] . "</td>";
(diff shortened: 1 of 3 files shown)


a23bensh (approved):
The code does what it claims to, and the "Depends on ID" values match the microservices listed in microserviceUI.php. Once more documentation is added, this will definitely help with presenting the associations between microservices, even in cases of inconsistent naming.
I do wonder, though how did the "Download Database" button end up in microserviceUI.php without being shown in a commit in this PR? Maybe I am just missing something, or perhaps it was added in another issue. @a23oskme 

Overall, looks good!

a23oskme:
> The code does what it claims to, and the "Depends on ID" values match the microservices listed in microserviceUI.php. Once more documentation is added, this will definitely help with presenting the associations between microservices, even in cases of inconsistent naming. I do wonder, though how did the "Download Database" button end up in microserviceUI.php without being shown in a commit in this PR? Maybe I am just missing something, or perhaps it was added in another issue. @a23oskme
> 
> Overall, looks good!


Yes, the download button was added in this PR https://github.com/HGustavs/LenaSYS/pull/17211 
I checked if there was any other issues open with microserviceUI before adding it, I didn't see any at that time, but after this merged I saw that there was one opened earlier. So I hope that it doesn't create any conflicts, if it does, feel free to remove the button and open a new issue to add it.

Issue #17213, opened 2025-05-06
Box selection and highlight inconsistency


a23aleva:
When using box selection to select multiple elements they turn purple to show which files are targeted. However, when it is used on a single target that purple indicating highlight doesn't show at all. This behaviour should be consistent regardless of how many elements are targeted.

https://github.com/user-attachments/assets/b90c19f0-17b4-418c-8c75-9bf1b37d43b7

PR #17241, opened 2025-05-06
Single elements now turn purple upon selection

a23emisn:
Single elements did not turn purple upon selection with either the pointer or the box selection. This patch has fixes this issue and all elements turn purple when selected now. The picture below demostrates this: a single ER entity is selected and is clearly purple.

image

Fixes: #17213


a23eliax (approved):
I confirm that this is a working solution and it's not disturbing any other functions!

image

PR #17242, opened 2025-05-06
Fixed desktop toolbar height

a23kevre:
the new media query for min-width was used to alter the height of the toolbar on desktop mode without making the mobile version change.


a22abdjo (approved):
The solution provided to solve the issue has been fixed and therefore approved.

Issue #16774, opened 2025-04-17
Locate and change the JQuery used in codeviewer.js with normal JS - line 4105 to line 4668


a23erigu:

a23elism:
### NOTE: Ignore all AJAX, this will be made its own issue

a23simry:
lines exclusively include the whole of function resizeBoxes

a23simry:
because there is a lot of the same code for each template, I will be trying to create a helper method which resizeBoxes calls upon

a23simry:
The plan is for the updated functions to work with any kind of template, if successful will also need to change the wiki, as there is a section on adding new templates to resizeBoxes.

a23simry:
![Image](https://github.com/user-attachments/assets/4d44cd7b-eebf-45e7-bc7c-520b3b43f27f)

The solution I have works, I think, identical now to the jquery version. Only issue I cannot for the life of me fix is this, where 3 or more boxes are stacked, there is a white space between them I cant get rid of.

a23simry:
well disregard the previous comments, I couldnt solve the white space issue so opted for a solution much closer to the origina jquery code.

PR #17243, opened 2025-05-06
Swapping all the jquery in resizeBoxes function in codeviewer.js #16774

a23simry:
Ended up not really improving on the code structure, but the functionality should hopefully be the same as before. Testing this specific function is simple, try every codeviewer template and try dragging the boxes to resize them. 
image

I don't think anything outside of these pages should be affected by the changes.


b20isato (approved):
This has not degraded the site at all. The site has it's its own problems but no new has been introduced

PR #17263, opened 2025-05-07
Re-engineer of insertIntoSQLite_ms.php

a23axehe:
Have Reengineered insertIntoSQLite its invert dependent file getCourseID. Post without answer implemented according to standard in file microserviceRefactorPOSTnoAnswer.md. Have tried testing with no result, the tables from the database which both files use, calling either empty rows or a database which does not seem to exist "gitRepos". Have tried adding a github repo to LenaSYS since the file likely does something with github but that function does not seem to be working on my branch nor the groups weekly branch, unsure how to test further. Not sure if files are used in LenaSYS yet. Fixes issue https://github.com/HGustavs/LenaSYS/issues/16867 .

See previous PR #17124 .

### DIFF
--- a/DuggaSys/microservices/gitCommitService/getCourseID_ms.php
+++ b/DuggaSys/microservices/gitCommitService/getCourseID_ms.php
@@ -29,7 +29,21 @@ function getCourseID($githubURL) {
 
     // Check if not null, else add it to Sqlite db
     if($cid != null) {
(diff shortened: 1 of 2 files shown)


a20artje (approved):
I have come to the same conclusion that the file is not working from the start. I approve this since the code should work and I have created a new issue to fix the problem.

Issue #17262, opened 2025-05-07
Update table documentation in microserviceDirectoryDocumentation.md


a23oskme:
Update dependencies table documentation to include the foreign keys that was recently added.

PR #17267, opened 2025-05-07
Update table documentation in microserviceDirectoryDocumentation.md #17262

a23oskme:
Added new fields to the table documentation and updated the relationship description, also made the old ones a little bit clearer.


a23gabli (approved):
Documentation looks good and is clear.

PR #17268, opened 2025-05-07
Fixed issue #15647-snap-to-lifeline-or-object-after-placement

a23eliax:
Fixed this issue **again**. Had to put back some disappearing code in some functions in helpers/mouse.js.

https://github.com/user-attachments/assets/a15def50-c895-4d13-9afa-6ec12e593c15


a23aleva (approved):
The snap to lifeline is back and working. This fixes the issue at hand but for some reason some other old issues that were fixed beforehand has returned. These aren't part of this solution though and I will make some new sub-issues addressing them specifically. However this change is approved!

https://github.com/user-attachments/assets/91d998d9-9561-406c-af6c-a31ea342063c

PR #17271, opened 2025-05-07
refactored retrieveprofileservice to use POST

a23gabli:
Solves issue #16944

### DIFF
--- a/DuggaSys/microservices/profileService/retrieveProfileService_ms.php
+++ b/DuggaSys/microservices/profileService/retrieveProfileService_ms.php
@@ -1,10 +1,14 @@
 <?php
+header('Content-Type: application/json');
 
-function retrieveProfileService($debug, $success, $status){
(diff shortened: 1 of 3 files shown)


a23axehe (changes requested):
There is no POST call in the files dependant on this ms, updateSecurityQuestion and updateUserPassword. POST logic needs to be implemented in the files dependant on retrieveProfileService.

a23gabli:
UpdateUserPassword and updateSecurityQuestion have now been refactored to use POST when using retrieveProfileService

a23axehe (changes requested):
The path used as a parameter is used wrong, currently /LenaSYS/DuggaSys/microservices/../profileService/retrieveProfileService_ms.php, remove ../ before path as parameter that should be good and I think header('Content-Type: application/json');
 needs to be added before echoingjson_encode in the update ms files.

PR #17274, opened 2025-05-07
G3 2025 v6 #17025 jquery to js in profile.js

a22maxda:
Moved all changes from last week's PR (#17073) to this one, the changes requested have been made and the code should work.


a23erigu (approved):
This seems to be correct as all the Jquery is gone and the site seems to works like intended.

I had a bit of a problem with trying to find a legal new password but it eventually worked out.
The "need secure HTTPS" thing is a bit weird of a feature but that has nothing to do with this.

PR #17276, opened 2025-05-07
Add dependencies to microservices in microserviceUI.php #17269

### LINKED ISSUE #17269: Add dependencies to microservices in microserviceUI.php
Add dependencies for selected microservices in the UI and present them in the table with the rest of the data

a23oskme:
Added a table that outputs the microservices that the chosen microservice is depending on. I also made the name of the microservice clickable, if clicked the user will be taken to that microservice instead. Fixes #17269 

There was some styling for the a-tags, so there are currently buttons in the table. If that is undesirable we could create a issue that changes the styling (create a class that holds the styling instead of a-tags). Otherwise, leave as is. 

To test this, make sure to have the database installed and fille, then select a microservice that has documented dependencies in microserviceUI.php. 
 
image

If there are no dependencies this will be outputted instead. 
image

### DIFF
--- a/DuggaSys/microservices/endpointDirectory/microserviceUI.php
+++ b/DuggaSys/microservices/endpointDirectory/microserviceUI.php
@@ -187,7 +187,23 @@ if (isset($_GET['id'])) {
         <?php } else { ?>
             <p>No outputs</p>
         <?php } ?>
-
+        <?php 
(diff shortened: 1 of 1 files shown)


a23louma (approved):
The changes look good and now the text is clearer, I approve!

Issue #17184, opened 2025-05-05
ToggleElements button should indicate when in use and when not


a23fanro:
![Image](https://github.com/user-attachments/assets/81c6dd68-4f24-4e04-b16b-511b5c521ce5)
The button for showing hidden items should indicate when it is active and when it is not. 
Either by closing the eye or some other nice way of doing it.

PR #17278, opened 2025-05-07
#17184 ToggleElements button should indicate when in use and when not

f23linan:
Added a new SVG called eye_closed_icon. It shows up after you have opened up the hidden elements, you can close them by pressing the new SVG (which is stillin the same spot). Its a bit more clearer now!

image
Elements closed(by default).

image
Elements open.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -29,7 +29,7 @@ var numberOfItems;
 var backgroundColorTheme;
 var isLoggedIn = false;
 var inputColorTheme;
-let showHidden = true;
+let showHidden = false;
 let count = 0;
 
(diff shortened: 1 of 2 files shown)


b23sebth (approved):
**Fixes #17184. Approved!**

## What's changed:
A crossed-out version of the original icon is added to indicate the change in visibility.

## Recommendation:
The current solution changes the default visibility to hidden, which was necessary for the toggle to function correctly according to the PR author. After merging this, I recommend a new issue to be opened that investigates and fixes the default behavior while toggling the button.

Great work on the icon!

b23sebth:
https://github.com/HGustavs/LenaSYS/issues/17297
I have created an issue for the toggle.

Issue #17247, opened 2025-05-07
Floating course title (mobile-view)


a23fanro:
The course name is floating out from the course-header. Make it stay.

![Image](https://github.com/user-attachments/assets/2be88fc7-0a09-4e51-8388-0c62557d77a8)

![Image](https://github.com/user-attachments/assets/de56e39d-dc22-4bfb-a68a-74559745dcf7)

![Image](https://github.com/user-attachments/assets/4b0464a1-31c0-45b3-ab04-38c4f511ff4f)

PR #17284, opened 2025-05-07
Floating course title (mobile-view) #17247

f23linan:
The fix was to update position, however this does put it under the buttons. The buttons are still too big to fit perfectly however i feel like that is another issue.Its as good as it can get right now :)

image


b20isato (approved):
The course title now stays within the header

Issue #17260, opened 2025-05-07
(BUG) new item button not usable


a23fanro:
When clicking the new item button, it is impossible to select any of the options, they disappear before the user can select them. The options for new item need to be clickable.

https://github.com/user-attachments/assets/2446ab6f-50f1-4acb-99bc-2ea484dbd614

PR #17285, opened 2025-05-07
Moving item creation for new item button #17260

a23erigu:
All I have done is move the item create menu 75px to the right.

It looks like this now
image


EdvinWerner (approved):
This works exaktly as advertised, approved for merge

Issue #17273, opened 2025-05-07
Change the microservice retrieveUsername to use "curlService"


b23johgu:
Continuation of  Change the microservice retrieveUsername to use "curlService" (HARD) #17109 #17163

PR #17286, opened 2025-05-07
Changed microservice retrieveUsername to use curlService #17273

a23danme:
Fixing issue #17273


a22linre (commented):
The files are now using curlService, looks good!

Issue #17181, opened 2025-05-05
Error when trying to delete a group activity


a23erigu:
more info found in #17167

c22isako:
Error stems from the selection-check which is done when the dugga is not an section/header/moment, to find its parent. If no parent exists, then a "uncheck" is run upon a null, which creates the seen error ( originally discovered within https://github.com/HGustavs/LenaSYS/issues/17167 )

![Image](https://github.com/user-attachments/assets/870f7a6d-2ffd-4e51-9830-c757b5575781)

PR #17289, opened 2025-05-07
G3-2025-V6-#17181-error-when-trying-to-delete-a-non-section-header-moment-element

c22isako:
Originally was https://github.com/HGustavs/LenaSYS/pull/17205#event-17537352313 .

### Removed unnecessary searches (and unchecks) of null
- Since the issue stemmed from an uncheck upon an null, made sure that if the unchecked element doesn't have a "parent" (such as a connected section/header/moment), the journey to uncheck the parent shouldn't be conducted.
---
Select course of your choice (which has duggor, ex. Testing Course) 
-> Log in 
-> Via fab button, create a new element, ex. a code dugga. _Note that link elements seem to be broken within the weekly branch, so I would advice to not create one of those._
-> Make sure that the new element doesn't have any section/moment above them. (I believe that the default location is at the top, and if so, nothing has to be moved.) It is also applicable on the type header, but only if the header is configured to a dugga within the dugga-list, not the page-title. 
-> Check & uncheck, test it out! The error message should be avoided with this fix applied. 
Original error can be found within: https://github.com/HGustavs/LenaSYS/issues/17181#issuecomment-2851363141

(image of selection box & placement)
bild


a23yutta:
I am not encountering the same error message described in the related issue [#17181](https://github.com/HGustavs/LenaSYS/issues/17181#issuecomment-2851363141). Instead, I receive a different error message in both the weekbranch and the sub-branch (PR).
Skärmbild 2025-05-08 153630 


It appears that this issue only occurs for me. It's unclear whether the cause is related to my PC or a result of incorrect testing. After discussing with the group leader, another person will be assigned to review this PR on my behalf.

a23alera (approved):
I tested the checkbox functionality and encountered no issues — it works as expected when selecting and deselecting duggas.

PR #17225, opened 2025-05-06
Locate and Change AJAX call to normal Fetch() -= sectioned.js =- Line 3049 & 3086 #17172

a23elism:
### This PR is connected (clearly) to the issue of the same name (#17172) and simply focuses on converting 2 functions (and AJAX calls) to normal JS instead of JQuery and AJAX

Issue #17172, opened 2025-05-05
Locate and Change AJAX call to normal Fetch() -= sectioned.js =- In functions - retrieveCourseProfile() & getStudents() 


a23elism:
### The lines, to be converted in this Issue are the AJAX calls starting on line 3049 and 3086, These are grouped based on them both seemingly being related to UI (a little bit iffy on that still, recommend checking where these are used (if used))

Highly recommend looking at the parent issue for a quick run-down on how to convert AJAX to normal Fetch() (normal JS)

Line 3049 (6th $.ajax instance):

![Image](https://github.com/user-attachments/assets/7053f3b3-9752-480e-8b00-06cafd92487f)

Line 3086 (7th $.ajax instance):

![Image](https://github.com/user-attachments/assets/097dbd3a-e91c-4b3e-b2b5-ce305bd8906d)

These images dont show the entire ajax call but simply a start for help with localizing, note the entire '$.ajax' call is to be converted just as the parent issues PR and not simply switch the '$.ajax(' with 'fetch('

Issue #17259, opened 2025-05-07
Add the "Announcements" button to the navbar on desktop


a22maxda:
The "Announcements" button currently does not appear on the navbar in sectioned.php (course page) on desktop, however it does show up on mobile. The button should appear on both views.

Desktop:
<img width="1016" alt="Image" src="https://github.com/user-attachments/assets/ecd55617-e4c1-4b5e-988d-8be06b3e0923" />

Mobile:
<img width="425" alt="Image" src="https://github.com/user-attachments/assets/cbfdad8a-a0b0-4e66-9d59-c3eee27033fc" />

a23simry:
Not sure exactly what "Announcements" is for, every course I tested only had an empty space for it, so possibly before adding it, try to do some research on if this is actually needed.

a23fanro:
Other issues/PRs that are related to the announcement that may be affected by Henriks input:
PR: https://github.com/HGustavs/LenaSYS/pull/17290
Issue: https://github.com/HGustavs/LenaSYS/issues/17172

Issue: https://github.com/HGustavs/LenaSYS/issues/17223
PR: https://github.com/HGustavs/LenaSYS/pull/17371

HGustavs:
I am unsure if we previously removed the use of announcements or not.

Please refer to previous issues on this topic, but from my foggy recollection, we decided that announcements would be handled by canvas.

a23fanro:
After examining old issues about announcement, issue https://github.com/HGustavs/LenaSYS/issues/17501 was found, declaring announcement are no longer supposed to be implemented with LenaSys, as they are handled  by canvas.

We will begin the process of removing announcement from all views, and this issue will be closed.

PR #17290, opened 2025-05-07
Locate and change ajax call to normal fetch() -= sectioned.js =- line 3049 & 3086 #17172

a23elism:
### This is the updated version of the same PR (org PR #17225)

Recommend reading comments inside original PR for info on testing


c22isako (approved):
From what I can see, the functions most apparent usecase is within announcements (ex. via retrieveAnnouncementAuthor().) 

Since announcements don't work at this moment (see this issue: https://github.com/HGustavs/LenaSYS/issues/17259) I can not confirm that this solution works, nor that it doesn't work. But from the look of it, the AJAX is removed and the replacement code looks both functional and has some comments to boost understandability. (discussions has been made with the group leader, and this is decided to be approved.)

Note that the documentation regarding this within microservice_connections.md will be a bit outdated now, but the group responsible for the document has been notified and everything is in the clear. 

Great work!

Issue #17221, opened 2025-05-06
Edit microservice documentation


a23edvto:
The types of the parameters should be the expected type and not the exact one. So for example string instead of char/varchar. In the parameter description you can write what type it is saved as in the database. Here is a typical example of how I think:

---

Parameter: $param

Type: String
Description: Parameter name. Stored as VARCHAR (30) in database

---

The file is: LenaSYS/DuggaSys/microservices/Microservices Documentation/

PR #17296, opened 2025-05-07
Corrected the documentation for the microservices #17221

a23wilgr:
Corrected the documentation for sectionedService, sharedMicroservices and accessedService files in Microservice Documentation folder. Changed to and added the correct types, and edited sectionedService to follow to template.

### DIFF
--- a/DuggaSys/microservices/Microservices Documentation/accessedService.md
+++ b/DuggaSys/microservices/Microservices Documentation/accessedService.md
@@ -1,3 +1,5 @@
+# accessedService Documentation
+
 # Name of file/service
 addClass_ms.php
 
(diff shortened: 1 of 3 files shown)


a23axehe (changes requested):
Looks good! Have checked some of the services and compared them to the documentation and they all seem correct. The structure of the documentation looks good aswell. While we´re at it though it might be a good idea to just remove line 87 in sectionedService.md to get rid of the thick line seperating the two microservices since thin lines are used everywere else in the three files. Just that easy fix then it should be complete.

b23johgu:
We've decided that the thick line should be after each microservice to separate the microservices in the documentation which makes it more readable! @a23wilgr will add lines to the rest of the documents where it wasn't added before. :)

a23axehe (approved):
Thick lines added, looks good. I have no more comments.

PR #17298, opened 2025-05-07
Added text in the fields of the modal when the user get a dbConnectionError, and also add text in the modal for all errors. 

### LINKED ISSUE #17229: Add text in the fields of the modal when the user get a dbConnectionError, and also add text in the modal for all errors.
I will only create text in the fields in the modal when the user gets a dbCreationError. Add text in the modal when the user gets a database connection error, copy course files error, database creation error, SQL error, or operation error. The text in the modal to show values for the fields entered so that easy to remember what they put before input.

a23dilel:
Fixes #17229. I have implemented that title says and here a screenshots what text and fields looks like:
image

To get this "Database Connection Failed" with fields, need to type wrong hostname, root username and/or root password. The summary (dropdown) and the input fields can be updated. To do this need to enter input fields and press the "change settings" or press "retry" button and enter input fields there. When you did this, then the summary and the input fields should updated.

When installer show a different error, then it should appear summary and updated.

I have tested with xampp and docker.

### DIFF
--- a/newinstaller/style.css
+++ b/newinstaller/style.css
@@ -399,6 +399,10 @@ input[type=checkbox], input[type=radio] {
     padding: 0.75rem;
 }
 
+details {
+    margin-top: 10px;
+}
+
 details,
 summary {
     cursor: pointer;
--- a/newinstaller/style.css
+++ b/newinstaller/style.css
@@ -399,6 +399,24 @@ input[type=checkbox], input[type=radio] {
     padding: 0.75rem;
 }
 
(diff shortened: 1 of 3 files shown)


a23louma (approved):
I approve this PR! Everything works as expected and I'm getting the same error message as the screenshot above. This makes it easier to change the things that were not inserted correctly. Just a thought for improvement(maybe a later issue if agreed), maybe the password shouldn't be displayed in plain text?

PR #17299, opened 2025-05-07
G2 2025 v6 #16827

a23arvme:
Changed all the files listed under inverse dependencies.
- Removed include in _copyCourseVersion_, the function is unused.


a20artje (changes requested):
This ain't working Arv. 

I cannot create a new code example on the website, works on master (although you need to reload the page), also creating a new link seems to break the whole page, don't know if this is related to this PR.

Issue #17234, opened 2025-05-06
Test type element timeline broken


b23sebth:
These problems seem to be mostly related to Testing-course in LenaSYS and is a continuation of [Switching a element to test](https://github.com/HGustavs/LenaSYS/issues/17171). The timeline becomes visible when changing an elements type to test. The element(s) does not appear on the timeline however. The main objective of this issue is to find out why the elements are not showing correctly in the timeline. If possible, also try locating the following issues aswell:

The date seems to be hardcoded to only accept dates between a small intervall in 2020 when choosing an absolute deadline. A relative deadline does not show up on the timeline either.

Lastly, the date does not show up unless you manually refresh.

## Empty timeline:
![Image](https://github.com/user-attachments/assets/6a527e57-fafc-4212-a97e-822adc71fecd)

## Date input:
![Image](https://github.com/user-attachments/assets/d4224a78-70b8-459d-944e-e19d355b589b)

## Before refresh:
![Image](https://github.com/user-attachments/assets/fcd626b2-141e-4353-87ee-c3becf717bb6)

## After refresh:
![Image](https://github.com/user-attachments/assets/95501ed3-1a49-4f67-a850-7927267987a5)

EdvinWerner:
After a more thourough examination of the code I've come to the conclusion that this isn't an issue, since after one does all the required steps the dugga will be shown in the swinlane. The swinlane is alway created even if the dugga isn't associated with a course moment but sectioned.js doesn't render the dugga inside of the swinlane until it's associated with a moment. So in order to get the dugga too show up in the timeline one needs to assign it to a moment and reload the page. If the dugga still doesn't show up in the timeline one simply has to edit the course version by selecting a start and enddate to the course. See attached screenshot for an example of how it should look

So too put it shortly the main issue described in this issue isn't an issue but my PR has along with something included in last weeks merge solved one of the subissues described. 

My recommendation is therefore to put a new reviewer on my PR and after that PR is approved we should merge and close this issue. 

![Image](https://github.com/user-attachments/assets/f0a58f74-8010-41a6-87d5-601439993e3a)

PR #17303, opened 2025-05-07
Updated updateDeadlineFunction

EdvinWerner:
Made it so that the function updates the deadline when kind==2 as well the code worked when switching from code to test in the dropdown, I think this solves the issue #17234

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -1348,6 +1348,9 @@ function updateDeadline() {
   if (kind == 3) {
     AJAXService("UPDATEDEADLINE", prepareItem(), "SECTION");
   }
+  else if(kind==2){
+    AJAXService("UPDATEDEADLINE", prepareItem(), "SECTION");
+  }
 }
 
 function setActiveLid(lid) {


a23osckr (changes requested):
I was not able to reproduce the deadline not updating when switching from code to test on the main branch, so I cannot test for improvement there. But, this issue solves the problem of the deadline not updating when switching an element from test to code. 

However, when reading issue #17234 I understood the main objective of the issue being to fix (or locate) the broken timeline which appears when switching an element to test. As far as I can tell this has not been fixed. I am unsure if it warrants a separate issue. But for now I do not want to approve this pr since it has not fixed the issue. Please correct me if I am wrong.

a23simry:
yes there's been some confusion on this, the main issue was party that the swimlane section shows up and is empty when changing a element in the list to test.
image
image

EdvinWerner:
Yupp I've missunderstod the issue, working on a fix for the actual issue but i think I'm keeping this in the code since it solves an issue as well.

a23erigu (approved):
The change fixes the issue of not auto updating when switching test to code and maybe something with moments not being seen when created (I don't think this has anything to do with this change but it is a thing that happens for me)

There is still an issue of the update happening before the change (so it appears nothing happens) but this probably has to do with the AJAXService being to fast or slow on some computers.

Issue #16936, opened 2025-04-25
Name of duggor can't be seen on mobile


a23erigu:
If you are logged in as a admin then there is to many things being shown on a dugga so the name of it is hidden

pls change so that you can see the name as a admin  

pictures of how the duggor looks on mobile
![Image](https://github.com/user-attachments/assets/c7a7fb07-a8c6-4bec-8cad-29b1bc3c6dfc)

a23erigu:
@EdvinWerner We still see this as a issue as you can't know which dugga is which so we would like this fixed

EdvinWerner:
I think that the PR that I created for this issue (#16990) has been incorrectly reviewed since the fix that i proposed works on both my own computer as well as a friend of mine's, this is what the attached screenshot clearly shows. 

![Image](https://github.com/user-attachments/assets/1476545b-fd4f-4286-a9b7-be1a43ec6e24)

EdvinWerner:
Guess I'll have to create a third identical PR for this issue then...

PR #17305, opened 2025-05-07
Updated dateSize 3.0

EdvinWerner:
Did exactly the same change for a third time, make sure put a reviewer who know how to branch correctly this time and go to the mobile version of sectioned.php for an example of how this solution works check issue #16936

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -936,6 +936,7 @@ p {
   .dateSize {
     width: 25%;
     font-size: 10pt;
+    display: none;
   }
 	.internal-link {
         font-size: 10pt;


b23sebth (approved):
**Fixes https://github.com/HGustavs/LenaSYS/issues/16936. Approved.**

## What's changed:
- The date is no longer visible for mobile devices.

image

## Recommendation:
- Do not use apostrophes in the branch name as Git tries to read it as a string starting, at least when using the command line. May be an issue earlier reviewers may have encountered.

a23fanro:
Soooo... Sure... This fixes the issue with names not being shown. But it also removes the date, which is also important for mobile-view users to see. I will merge this and make another issue about having the dates also be included in mobile-view.

Issue #17237, opened 2025-05-06
Hovering over the save buttons doesn't show a tooltip


a22nevan:
There's an issue with the page where hovering over the two save buttons on the sidebar doesn't show any tooltip information for the buttons. The other buttons show their tooltips when hovered over. The save buttons have tooltips written in the php file, although they don't show up on the page.

![Image](https://github.com/user-attachments/assets/9200efce-c7a6-49f0-8f6b-33fcaf16554f)

PR #17307, opened 2025-05-07
Tooltip fix issue #17237

a23matsh:
There was missing functionality specifically for "disabledicons" when hovering hence why the tooltips was not being shown, so this was implemented and the tooltips started showing for "save" and "save as", as well. After this was solved some slight tweaks was made with the transition delays when hovering so all the buttons behave the same and consistent.

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -550,7 +550,13 @@
     top: -43px;
 }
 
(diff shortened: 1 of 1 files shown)


a22abdjo (approved):
The changes provided to solve the issue works as intendended and therefore approved.

PR #17308, opened 2025-05-07
Updated style.css

d22abdha:
The lock is now visible in dark coloured class boex.
image


a22abdjo (approved):
The solution provided works as intended and is therefore approved.

a22abdjo (approved):
The solution provided works as intended and is therefore approved.

Issue #16750, opened 2025-04-17
Locate and change the JQuery used in sectioned.js with normal JS - Setup and Event listeners - Line 3014 - 3528


a23elism:
Locate and change the JQuery used in sectioned.js under the Setup and Event listeners "chapter" starting form line 3014 til 3528 into normal JavaScript

a23elism:
### NOTE: Ignore all AJAX, this will be made its own issue

a22carsv:
After testing, it seems that the ability to change places of elements (drag-and-drop) was removed in the process of converting to vanilla JS, going to review the code to fix this problem then it is probably ready for PR.

PR #17310, opened 2025-05-07
 Changed jquery between lines 3014-3528 in sectioned.js for issue #16750

a22carsv:
A new PR based on previous week PR review feedback, made to match current weeks group branch. Please refer to following old PR for info beyond what is given in this PR:
https://github.com/HGustavs/LenaSYS/pull/17037
https://github.com/HGustavs/LenaSYS/pull/17076

This solves the original issue #16750 and works to address the given feedback in PR review feedback linked above, now only changes should be within the originally referenced region of code. It solves the feedback of affecting the MOTD given in review, or at least i was not able to recreate the problem.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -29,7 +29,7 @@ var numberOfItems;
 var backgroundColorTheme;
 var isLoggedIn = false;
 var inputColorTheme;
-let showHidden = false;
+let showHidden = true;
 let count = 0;
 
 function initInputColorTheme() {
@@ -883,6 +883,8 @@ function markedItems(item = null, typeInput) {
   // handles selections & deselections
   if (selectedItemList.length != 0) {
     let tempSelectedItemListLength = selectedItemList.length;
+
+    //for every item in selectionList
     for (var i = 0; i < tempSelectedItemListLength; i++) {
       var tempKind = item ? item.closest('tr').getAttribute('value'): null;
 
@@ -890,16 +892,18 @@ function markedItems(item = null, typeInput) {
       if (selectedItemList[i] === active_lid && typeInput != "trash") {        
         document.getElementById(selectedItemList[i] + "-checkbox").checked = false;
         selectedItemList.splice(i, 1);
-        var removed = true;
+        removed = true;
 
-        // deselection of child -> deselect parent
+        // deselection of child -> deselect parent (if parent exists)
         if (tempKind != "section" && tempKind != "moment" && tempKind != "header"){ 
           var parent = recieveCodeParent(active_lid);
 
-          for (var i = 0; i < tempSelectedItemListLength; i++) {
-            if (selectedItemList[i] == parent){
-              document.getElementById(parent + "-checkbox").checked = false;
-              selectedItemList.splice(i, 1);
+          if(parent != null){
+            var indexNr = checkIfInList(selectedItemList, tempSelectedItemListLength, parent);
+          
+            if(indexNr != null){
+              document.getElementById(selectedItemList[indexNr] + "-checkbox").checked = false;
+              selectedItemList.splice(indexNr, 1);
             }
           }
         }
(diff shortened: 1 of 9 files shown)


c22isako (changes requested):
**Issues within the PR:**
From looking at the code, there seems to be JQuery which is introduced between the lines 3067 and 3160, as well as 3216 and 3244. I'm guessing this might be because of the merge or something similar, but needs some fixing.

**Fixes by this PR:**
I do believe that the MOTD works now, as it is displayed on courses which contain test-data for it. (more info below) (issue mentioned within https://github.com/HGustavs/LenaSYS/pull/17076)

---

**Issues which isn't linked to this PR.**

I can not see the MOTD banner for testing course and Demo course, but can see it for other courses such as Webbutveckling - datorgrafik. 
- Thought from further investigation, this seems to be the case within the weekly branch as well, so I don't believe that this fix affected it. It might be linked to the fact that testing course is missing its version name (testdata error!) 

I couldn't find (or create) any applicable announcements to test out that they work, so I can not confirm the solution in that case. see https://github.com/HGustavs/LenaSYS/issues/17259 for further information. (Group leader has been notified)

a22carsv:
Feedback has now been addressed, there was one function "retrieveAnnouncementAuthor" that had not been converted to vanilla JS from JQuery. Although there were some other ones within the region of 3014-3528 that contained JQuery, namely: retrieveCourseProfile, getStudents, retrieveAnnouncementCards, updateReadStatus and toggleFeedbacks. 
Although these contain JQuery, it was specified in the original issue conversation to ignore all parts utilizing AJAX (#16750), hence it has been ignored thus far as to not overreach the scope of this issue. Although, I should have mentioned it in the description of this PR.

b23sebth (changes requested):
**Changes needed - not approved.**

Although you fixed a bug introduced by Javascript in [#16750](https://github.com/HGustavs/LenaSYS/issues/16750). The solution should not have reintroduced jQuery, even though it concerned AJAX handling. jQuery was not ignored, it was added.

## What's changed: 
 - Functions updated to use modern JS conventions.
 - jQuery changed to vanilla JS. 

## Changes needed:
- Revert all jQuery to the prior Javascript. A new issue will be created to fix the drag and drop instead.
- Inconsistency with variables. While use of const is great, let should be used instead of var if the variable does not require global scope.
- Remove inline styling and use style.css instead.

a22carsv:
### What I changed:
- Moved most of the announcement components inline styles into style.css
- Reverted three functions that had inadvertently re‑introduced JQuery  
- Left a handful of inline styles untouched  
Those belong to elements outside the scope of Issue #16750, maybe this can be made into a separate issue, otherwise i can continue doing that in this PR.

I believe every change relevant to the original issue is addressed now, I briefly tested around but couldn't see anything that broke because of changes made.

DuggaSys/sectioned.js
```diff
@@ -3273,172 +3276,186 @@ function updateannouncementForm(updateannouncementid, cid, versid, tempFuction)
...
+      } else {
+        cards.slice(6).forEach(c => (c.style.display = "none"));
+        btn.querySelector(".showmore").textContent = "Show more";
+      }
+    });
+  }
 }
+
 function updateReadStatus(announcementid, cid, versid) {
   var uname = $("#userName").html();
```
a22maxda, line 3460: This line is still JQuery

DuggaSys/sectioned.js
```diff
@@ -3460,52 +3477,54 @@ function updateReadStatus(announcementid, cid, versid) {
...
-  }).mousemove(function (e) { e.preventDefault() });
+    setTimeout(() => (this.scrollTop = scroll), 0);
+    this.focus();
+  });
+  sel.addEventListener("mousemove", e => e.preventDefault());
 }
+
 // Start of recent feedback from the teacher
 function toggleFeedbacks() {
   let uname = $("#userName").html();
```
a22maxda, line 3530: This line is also JQuery

a22maxda (changes requested):
The code looks to have been converted except for 2-3 lines which still utilize JQuery. These are commented in the Files Changed tab and are on lines 3460, 3530, 3621. The last one may be outside the scope of the PR but it is visible in the changes so i'm including it. I also didn't find any issues with the parts that have been changed.

a22carsv:
I have addressed the feedback given and resolved a conflict that appeared just recently.

EdvinWerner (approved):
Seems to me that the code is working as it should now the swimlanes and all other associated stuff seems to be working as they should therefore it's approved for merge

PR #17311, opened 2025-05-07
Create sectionedService.md

a23petny:


a20artje (approved):
Documentation of both files is complete and correct.

b23johgu:
Here is the template we're currently following: DuggaSys/microservices/Microservices Documentation/template.md

The file sectionedService.md already exists in there as well. I saw that the issue was a little unclear on where you were supposed to put the file, sorry for that! Follow the template in the right file and make a new review when that is done! :) 
@a23petny @a20artje

PR #17313, opened 2025-05-07
Burger menu added in dugga for mobile

a23yutta:
Burger menu has been added to dugga.js for mobile. The burger dropdown menu contains the navigation tools: home, toggle darkmode/lightmode and return/back.

It's an ugly solution but it works. I am aware that the burger icon is not consistent with the others, the positioning of the burger menus are different for each burger menu(there are three in total), in fact the burger menu in sectioned and courseed differed from the start.

The navheader.php file is currently messy, unstructured, and not scalable for long-term use. It contains a significant amount of redundant code that PHP may tolerate. All these greatly hinders readability and maintainability.

There are currently three burger menus in total, all using duplicate and inconsistent IDs. This is the reason why the hamburger icon fails to appear in showDugga.php which led to this ugly solution. A more refined solution would be to consistently and uniquely name the element IDs for each burger menu. The current approach is not a viable long-term solution and may lead to further issues in the future.

### DIFF
--- a/DuggaSys/courseed.php
+++ b/DuggaSys/courseed.php
@@ -49,6 +49,7 @@ if(isset($_SESSION['uid'])){
 <body>
 
 	<?php
+	$courseed = true;
 	include '../Shared/navheader.php';
 	?>
 
--- a/DuggaSys/showDugga.php
+++ b/DuggaSys/showDugga.php
@@ -113,6 +113,7 @@
 
 	<?php
 		$noup="SECTION";
+		$showDugga = true;
 		include '../Shared/navheader.php';
 	?>
 
--- a/DuggaSys/showDugga.php
+++ b/DuggaSys/showDugga.php
@@ -20,6 +20,7 @@
 		<!-- To enable dark mode, these 2 files were added. -->
 	<link id="themeBlack" type="text/css" href="../Shared/css/blackTheme.css" rel="stylesheet">
 	<script src="darkmodeToggle.js"></script>
+	<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css">
 
 	<script src="../Shared/js/jquery-1.11.0.min.js"></script>
 	<script src="../Shared/js/jquery-ui-1.10.4.min.js"></script>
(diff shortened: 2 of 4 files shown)


a23lisjo (changes requested):
I have also implemented a hamburger menu and I totally agree that the code in navheader.php is quite a mess. The code works but I have found a way to implement this with less added code.

The main problem is that the link to the hamburger menu icon is missing in showDugga.php, by adding this link :
`<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css">`  
the icon should appear and it should be in the same style as the others as well work with $noup='SECTIONED'.

The functions in dugga.js should still be the way you added them.

### In summary
When testing I removed the added code in 
- showDugga.php line 116.
- navheader.php lines 394-432

and got this result:
image

---

Another thing to fix would be to make the navBurgerBox fit all icons 
image

The last thing I noticed was that when logged in as admin the hamburger menu disapears, that is something to do with superUser, but as logged out or logged in as a student this works!

a23simry:
@a23yutta its up to you if you want to implement the requested changes if they improve on your solution, otherwise if you are satisfied enough with your code and it works we can merge it.

a23yutta:
I have now added the style link for the burger menu that was suggested in the review. However, I decided to keep my code but with a few modifications, these modifications involves  support of the logged in/out version of the page. I have also fixed some display issues of the burger menus in general, before , I have added a default burger menu to be displayed with the contents: home, toggle dark/light mode and return.

a23yutta:
@a23lisjo I have NOW added the suggested style link for the burger menu from the review. However, I chose to keep my original implementation with some modifications. These changes focus on displaying the burger menu content based on whether the user is logged in or out. As part of this update, I introduced a default burger menu that includes the items: Home, Toggle Dark/Light Mode, and Return. This is replacing the previous logged-out version of the default burger menu(`courseed.php` burger menu) but still keeps the `courseed.php` menu intact.

Additionally, I’ve addressed and resolved some general display issues related to the burger menu across the logged out version of different pages(_see attached images_).

### Before change:

Skärmbild 2025-05-09 131424 
### After change:

image

## This PR is ready to be re-reviewed.

a23lisjo (approved):
I like this solution and the code works and looks more structured now! Good job!

Issue #17245, opened 2025-05-07
Name of elements disappears


a23eliax:
When marking, moving or clicking on an element the name of it disappears, and then it comes back when you unclick/unmark the element. Of what I can find, this is the case for sequence object, sequence condition and UML super state. Please note that this was not an issue at the beginning of this project so this has happened under these recently weeks.

Illustration:
https://github.com/user-attachments/assets/624fefa8-35ce-43d9-b34c-55216d3b44ed

PR #17314, opened 2025-05-08
Fixed: Sequence elements turn purple upon selection.

a23emisn:
A previous pull request I made (#17241) fixed an issue where single-selected elements didn't turn purple upon selection. However, it also introduced a new bug that caused the text color to turn white for certain sequence elements and the UML super state element. This became problematic because the background color wasn't updating correctly to purple on selection, making the white text appear to disappear.

I've now patched this issue by extending the earlier fix to include these elements as well, ensuring they also turn purple when selected. Additionally, I’ve verified that everything continues to work correctly with dark mode enabled.

Before: 
image

After: 
image

Fixes: #17245

### DIFF
--- a/DuggaSys/diagram/theme.js
+++ b/DuggaSys/diagram/theme.js
@@ -137,7 +137,18 @@ function updateCSSForAllElements() {
                     }
                     // Update normal elements, and relations i.e default case
                 } else { 
-                    fillColor = elementDiv.children[0].children[0];
+                    if (element.kind == elementTypesNames.sequenceObject) {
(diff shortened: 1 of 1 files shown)


a23aleva (approved):
Looks like a reasonable fix and seems to work as intended. Tested on multiple elements and the box selection now turns them all purple as they are targeted. Also tried dark mode that looks good too. This change is approved!

https://github.com/user-attachments/assets/26a8e2cc-b525-400e-9a19-c27b3f0a989c

Issue #17291, opened 2025-05-07
markedItems() within sectioned.js could benefit from simplification via functioncalls


c22isako:
The function **checkIfSelected()** was created within the fix to issue https://github.com/HGustavs/LenaSYS/issues/17222#issuecomment-2854309696, which could benefit markedItems (found within sectioned.js.)

Check it out, it might need modification. Try to simplify the long if-statement within markedItems().

PR #17318, opened 2025-05-08
G3-2025-v6-#17291-markeditems-simplification

c22isako:
**Unsure if this was an simplification, but I do believe that the code is (if not only) more readable now.** 
Also changed the function to be able to handle not only selectedItems (list), so reuse is better.
There should be no new functionalities, and the changes affect the selections which can be done on "duggor" within courses.

select course of your choice (which has duggor, ex. Testing Course) -> log in -> test selection box 

(Image of selection boxes)
bild
_note: I'm highly unsure if this is better or not. Please give feedback if changes not needed!_

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -896,11 +896,14 @@ function markedItems(item = null, typeInput) {
         if (tempKind != "section" && tempKind != "moment" && tempKind != "header"){ 
           var parent = recieveCodeParent(active_lid);
 
-          for (var i = 0; i < tempSelectedItemListLength; i++) {
-            if (selectedItemList[i] == parent){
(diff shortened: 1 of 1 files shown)


a23simry:
This has conflicts that need to be resolved before I can add a reviewer. @c22isako

c22isako:
"Merged" the bugfix involving error when unchecking element without a header (https://github.com/HGustavs/LenaSYS/pull/17289) with this, so that both functionalities are working together.

f23linan (approved):
Functionality works as expected! The selection of specific duggor works with the delete function and the undo. The simplification looks good in my eyes but im not super familiar with javascript. 

All in all I think its good to be merged, I cannot find any reason why it shouldnt be.

PR #17321, opened 2025-05-08
Undo/Redo now affects functions as well

b23ebbst:
Updated the save()-function to be able to hande undo/redo for functions along with the pre-existing name and attributes. A user can now undo/redo parts of words rather than only single letters.

If the last part of the word is less than 4 symbols (or the word is less than 4 symbols), it will not be redone as it falls below the minimum-limit of what is commited to memory.

https://github.com/user-attachments/assets/4b1cb1e7-6d00-465c-bd5a-225ca40d7f06

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -60,16 +60,48 @@ class StateMachine {
                 });
                 break;
             case StateChange.ChangeTypes.ELEMENT_ATTRIBUTE_CHANGED:
-                for (const element of StateChange.ElementsAreLocked()) {
-                    if (Array.isArray(id)) id = getItemsFromNestedArrays(id)[0];
+                // Normalize ID if it is in an Array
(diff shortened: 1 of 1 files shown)


a23vinge (approved):
As mentioned by the developer, the save() function can now handle undo/redo actions within functions in the UML class diagram. I have reviewed the changes, and it works under the criteria (more than 4 symbols) but as mentioned by the developer if a person would do less than 4, the symbols do not come back after redo.  

I do believe that this is approvable because it builds clear building stones for future issues/sub issues that can improve the code on more stable ground.  The code is functional and is an upgrade, but have small bugs.





https://github.com/user-attachments/assets/e94892ba-4e42-4a91-b366-616a037ef3ea

Here it can be seen that the code does work in some sense and have given great building stones.



https://github.com/user-attachments/assets/25c9416a-1136-4ad3-95a2-ac8421251a76

Here it can be seen that "wife" return as "w" meaning 3 symbols are missing.

PR #17325, opened 2025-05-08
Fixed: G1 2025 v6 17169 the arrows is not visible on straight line between state diagram entities

a23eliax:
Removed duplicated logic for arrow positions and drawing arrow shapes for both the start and end of a line.

Added a new function: calculateArrowPosition
which calculates the arrowhead position at the start or end of a line.

https://github.com/user-attachments/assets/52db799a-99f8-4a4a-b433-a406ddce9c95


a23gabst1 (approved):
It works as intended! The arrows are as of now visible when the line type is straight, see figure 1. 

image

Issue #17179, opened 2025-05-05
Links crash LenaSYS


a23erigu:
When you try to create a link it fails and if you reloade then LenaSYS stops working

The problem seems to be with this line 
if (item['link'].substring(0, 4) === "http") 
at line 1956

I think it has to do with substring

PR #17327, opened 2025-05-08
G3-2025-V6-#17179 Links crash lenasys

a23osckr:
Links can now be created without crashing LenaSYS
image

Added a null and undefined check to stop the if statement executing .substring on a null or undefined link. If one of these evaluates to false, the if statements stops and throws an error instead of crashing. To test, create a new link in sectioned.php.


a22maxda (approved):
The website no longer crashes from creating links. I have not been able to find any issues with the implementation.

Issue #17183, opened 2025-05-05
Locate and Change AJAX call to normal Fetch() -= sectioned.js =-  In Function - fetchFileContent() & fetchFileName()


a23elism:
## The lines, to be converted in this Issue are the AJAX calls starting on line 4611 and 4662, These are grouped based on them both seemingly being related to Github API fetching

Highly recommend looking at the parent issue for a quick run-down on how to convert AJAX to normal Fetch() (normal JS)

### Note Lines are subject to change since we keep adding code into sectioned.js so use the instances instead when trying to find relevant AJAX (Below is how to find through "instances" [ctrl+f or MAC equivalent])

![Image](https://github.com/user-attachments/assets/8cdb55aa-3db1-4d91-986c-9b83fa124e4b)

Line 4611 (15th $.ajax instance):

![Image](https://github.com/user-attachments/assets/b546e204-64e8-4bc8-9464-daa533c9a4c7)

Line 4662 (16th $.ajax instance):

![Image](https://github.com/user-attachments/assets/26498f51-752d-4d21-92ff-3700bf7e72ab)

a23fanro:
I tried out filling in some credentials, used the lenasys repo.
![image](https://github.com/user-attachments/assets/5105e1e3-f38b-4fd0-a756-8549579df271)

Then I got this.
First, when opening the github pointer, the 'childTd is null' shows up. When, after adding the credentials and pressing save: 'Failed to fetch file names: ReferenceError: data is not defined' and 'Uncaught (in promise) TypeError: data.filter is not a function' shows up.
![image](https://github.com/user-attachments/assets/b6f2a50c-b6e3-463b-88ca-1231f7a51c5a)

After further analysis, both of our local lenasys and the [LenaSys](https://dugga.iit.his.se/DuggaSys/sectioned.php?courseid=7&coursevers=82518&embed) that is used in an actual course, I hava a hard time figuring out what the github fetch is supposed to do, and if it is use.

Will await input from the customer.

HGustavs:
I am thinking that the github fetch either 

* checks if the github repo connected has been updated (or if it even exists?)
* checks if specific code example has been updated (or if it needs to be downloaded?)

In this case I assume it is a specific code example on github connected to that specific code viewer entry.

The in-use version does afaik not have github synchronization.

a23fanro:
We ar working on understanding the github fetching: https://github.com/HGustavs/LenaSYS/issues/17496

PR #17334, opened 2025-05-08
Swap AJAX to JS - fetchFileContent() & fetchFileName() #17183

a23erigu:
I believe I have swaped all the ajax to JS and LenaSYS dosen't crash but I can't find how to really test this code.

This is also the first ever ajax I have touched so it's probably wrong in some way.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -4663,18 +4663,17 @@ function fetchGitCodeExamples(courseid){
     //Foreach loop to fetch each file in the filteredFiles array
     filteredFiles.forEach(function(filename){
       var apiGitUrl = 'https://api.github.com/repos/' + owner + '/' + repo + '/contents/' + folderPath + '/' + filename;
-      // Ajax request to fetch filecontent of current file in foreach loop. fetched file is pushed into results array.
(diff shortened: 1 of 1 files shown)


a23simry:
Will add a reviewer so they might figure out how to test this, if not we are also waiting on Henrik's input so no major worries.

a23yutta (commented):
I'm not sure how to go about testing of the converted code from AJAX to JavaScript. The code is being used in the `sectioned.php` file.
Github Icon sectioned

Skärmbild 2025-05-08 185119

**As you can see, valid credentials are required, but I'm not sure where to obtain them.**

a23fanro:
I tried out filling in some credentials, used the lenasys repo.
image

Then I got this.
First, when opening the github pointer, the 'childTd is null' shows up. When, after adding the credentials and pressing save: 'Failed to fetch file names: ReferenceError: data is not defined' and 'Uncaught (in promise) TypeError: data.filter is not a function' shows up.
image

After further analysis, both of our local lenasys and the [LenaSys](https://dugga.iit.his.se/DuggaSys/sectioned.php?courseid=7&coursevers=82518&embed) that is used in an actual course, I hava a hard time figuring out what the github fetch is supposed to do, and if it is use.

Will await input from the customer.

a23erigu:
The data that is referenced in the error seems to be from the change I made

Issue #17244, opened 2025-05-06
MOTD reshow button isn't shown on course pages


a23simry:
When closing message of the day on start page.
![Image](https://github.com/user-attachments/assets/9336f624-0e37-4f90-a1b6-f884051f9a0b)

When closing message of the day on course pages.
![Image](https://github.com/user-attachments/assets/25ad501e-d8e8-4465-bb3b-f50a115a39b6)

The button to reshow message of the day doesn't show up.

This is not a glaring problem but it is a UX/consistency issue.
(The button is there in the mobile burger menu but it gives an error when used - this will be turned into its own issue.)

PR #17335, opened 2025-05-08
Fixed MOTD button #17244

a23alera:
Fixed the motd button on course pages


EdvinWerner (approved):
The button to reshow the MOTD is now visible in both navbars and doesn't produce an error message when used, I therefore think that this PR solves the isssue!

Approved for merge

PR #17336, opened 2025-05-08
Fixed acccess denied user, when go to a course page using new LenaSYS installer with docker

### LINKED ISSUE #16672: Docker doesn't work when go to a course page when installed from a new installer
I have tested old lenays installer and it was successful to install lenasys. When I pressed a green-button which lead to a course page which works fine. I did same thing with the new installer and when I pressed finish-button which also lead to the course page, but it shows an error.

Image

a23dilel:
Fixes #16672. 

In the step 3 there is a checkbox `Use distributed enviroment` which it should updated the `coursesyspw.php` file located at `C:/LenaSYS/dockerenv/coursesyspw.php`, but the file never updated value `define("DB_USING_DOCKER", "")`. The `DB_USING_DOCKER` never add values or never updated, it doesn't matter if you check or uncheck `Use distributed enviroment`. 

In `installer_ui.php` there is a checkbox which says `distEnvironment`, but in an `install_engine.php` look for `distributed_environment` line 127. These names are not the same which `DB_USING_DOCKER` never uppdated which lead to acccess denied user.

I also updated `new_component.php` and `test.html`, but these are never used.

To test this here are all the steps:
Step 1: Select OS
Step 2: Select both checkboxes
Step 3: Database name: lenaDB, MySQL user: Daniel, Hostname: localhost, MySQL user password: 1234 and check Use Distributed Enviroment, and for the rest, uncheck.
Step 4: Root username: root and Root password: password
Step 5: Select all options
Step 6: Click install-button.

### DIFF
--- a/newinstaller/installer_ui.php
+++ b/newinstaller/installer_ui.php
@@ -100,7 +100,7 @@
 								inputFieldWithTip('hostname', 'Hostname:', 'Tip: Usually set to "localhost"');
 								inputField("password", "MySQL user password:", "password");
 
-								checkbox("distEnvironment", "Use Distributed Environment", "https://github.com/HGustavs/LenaSYS/blob/8be873ef4ccb3cdb2fc28e72b2a30a79aa52c2f9/Shared/Documentation/newinstaller/documentation.md#use-distributed-environment");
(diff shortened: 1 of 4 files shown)


a23louma (approved):
I approve! I'm using Docker and Windows and the installation works well, there is no error message showing after clicking the finish button. I followed the steps in the PR (except that it should be 'db' and not 'localhost' as Hostname).

PR #17347, opened 2025-05-08
G3 2025 V6 changed default visibility of objects

a22maxda:
Changed the default visibility new objects to be true, as before it was automatically hiding new objects. Also switched which icon should be used at which point to reflect the actual state of the button.

Before:
Screenshot 2025-05-08 at 14 57 34
Now:
Screenshot 2025-05-08 at 14 58 12


a23lisjo (approved):
Added objects are now shown as default and the changes made for the icon makes the toggle function more clear and understandable of what state it's in.

PR #17350, opened 2025-05-08
Remove header from sectionedService.md #17329

### LINKED ISSUE #17329: Remove header from sectionedService.md
The header on line 1 in sectionedService.md (documentation) creates an unwanted entry in the endpoint directory database. 

Image

a23oskme:
Removed the header in sectionedService.md that created an unwanted entry in the database. Fixes #17329 

If you run 'setupEndpointDirectoryDb.php' the entry in the image below should no longer be in the database (only the middle entry in the image). 

Image

### DIFF
--- a/DuggaSys/microservices/Microservices Documentation/sectionedService.md
+++ b/DuggaSys/microservices/Microservices Documentation/sectionedService.md
@@ -1,5 +1,3 @@
-# sectionedService Documentation
-
 # Name of file/service
 createListEntry_ms
 


a23danme (changes requested):
I am confused, first of all your branch is named G2-2025-v4-etc.. And, the change and fix you did is for me already implemented in this weeks branch, G2-2025-v6. The line "# sectionedService Documentation" is not there for me even when I was going to compare it to the group branch. 

Maybe it is bugged for me with the branches where it opens up ur changes anyway or it is already implemented earlier. Double check this because I tried on both of my computers, both looking at the file in vs code where the line does not appear on G2-2025-v6 either, and it does not appear when loading the UI (the setup and then the microUI).

a23oskme:
> I am confused, first of all your branch is named G2-2025-v4-etc.. And, the change and fix you did is for me already implemented in this weeks branch, G2-2025-v6. The line "# sectionedService Documentation" is not there for me even when I was going to compare it to the group branch.
> 
> Maybe it is bugged for me with the branches where it opens up ur changes anyway or it is already implemented earlier. Double check this because I tried on both of my computers, both looking at the file in vs code where the line does not appear on G2-2025-v6 either, and it does not appear when loading the UI (the setup and then the microUI).

Sorry, the branch name is my mistake. There was already an issue removing this header when I created this issue. Therefore this PR will be removed.

PR #17355, opened 2025-05-08
Change the styling of a-tags in MicroserviceUI #17340

### LINKED ISSUE #17340: Change the styling of a-tags in MicroserviceUI
Change the styling of a-tags in MicroserviceUI. 

"Regarding the buttons in the table. I think that we should create an issue that changes the styling of a-tags, so that we can use links without them turning into buttons. (Create a class and apply it to a-tags that should be buttons)"

_Originally posted by @a23oskme in https://github.com/HGustavs/LenaSYS/issues/17276#issuecomment-2858873655_

a23oskme:
Added a class to a-tags (in microserviceUI.php) that are buttons and added the class to the stylesheet instead of the a-tag.  A-tags that are not supposed to be buttons are now links. Fixes #17340 

image

### DIFF
--- a/DuggaSys/microservices/endpointDirectory/microserviceUI.php
+++ b/DuggaSys/microservices/endpointDirectory/microserviceUI.php
@@ -247,7 +247,7 @@ if (isset($_GET['id'])) {
             echo $service['description'];
         }
         echo '</td>';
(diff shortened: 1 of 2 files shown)


a23bensh (approved):
Changes the hyperlink buttons to just the link, looks good!

Issue #17223, opened 2025-05-06
Locate and Change AJAX call to normal Fetch() -= sectioned.js =- In functions - RetrieveAnnouncementCards(), updateReadStatus() & toggleFeedback()


a23elism:
### The lines, to be converted in this Issue are the AJAX calls starting on too many lines (read the description för locations) These are grouped based on them all seemingly starting with some sort of retriever file (php), this one is smaller than it really seems since they are smaller AJAX calls, but in a larger quantity.

Highly recommend looking at the parent issue for a quick run-down on how to convert AJAX to normal Fetch() (normal JS)

**Images wont be included (like the other issues as there are alot of them)**

### Note Lines are subject to change since we keep adding code into sectioned.js so use the instances instead when trying to find relevant AJAX (Below is how to find through "instances" [ctrl+f or MAC equivalent])

![Image](https://github.com/user-attachments/assets/8cdb55aa-3db1-4d91-986c-9b83fa124e4b)

a23fanro:
This has conflicts that need to be resolved before merge, moving it to 'to do'.

PR #17371, opened 2025-05-08
Locate and Change AJAX call to normal Fetch() -= sectioned.js =- In functions - RetrieveAnnouncementCards(), updateReadStatus() & toggleFeedback() #17223

a23elism:
### Check the issue for info on what is to be changed and where


a23fanro:
The function `RetrieveAnnouncementCards()` is likely going to be deleted since we are working on removing announcement from the project, so there is no need for the reviewer to test this section, just fyi.

Issue #17264, opened 2025-05-07
Re-engineer microservice retrieveShowDuggaService


b23johgu:
Continuation on #17236

a23axehe:
To the next dev working on this, at my commit [d874616](https://github.com/HGustavs/LenaSYS/pull/17628/commits/d874616da07c9eefcbd9c85e933c0cb7de183bf2) The solution was working however that solution is not compatible with the current state of retrieveShowDuggaService and processDuggaFiles. I've run out of ideas on how to solve this.

a23danme:
I tried @a23axehe 's solutions and did them more step by step to while testing to see whats making it not work. Tried alot of different solutions combined with his and could not understand why its not working. For me its always stops working when chaning showDugga, but by trying different methods to fix it I could still not make it work. Just by looking by the pictures below it can be shown that with my and axehe solution it shows "null" (inspection tool on Network) and in the original version it displays as it should. Im afraid I once again could not fix this solution to re-enginer retrieveShowDuggaService.

![Image](https://github.com/user-attachments/assets/6ac4a4f3-4c85-4e87-8322-b2d88e529bac)



![Image](https://github.com/user-attachments/assets/e18e8200-d2b9-45da-b86c-e5262c65fa02)

PR #17372, opened 2025-05-08
Changed back and repaired the 5 files for retrieveShowDuggaService #17264

a23danme:
#17264 
Changed files back to their original state, then added curlService. Works as it did in their orignal state (1 week ago) when not being logged in, aka being "guest" and using dugga. When logged in as "brom" you are "superuser" and it always comes "localhost: undefined" in browser. I have tried fixing it but I can not find the problem, maybe it is caused by something else. You can see all this easily if you compare logged in and not and going to Dugga 1 or 2 and clicking save button etc..

My opinion is to review and test this solution I have done. If it seems to be alright then I suggest to merge it and then make a new issue for this undefined error and then I can try to  fix it from there as a new issue. If still cannot fix it maybe someone else that are better than me or have other perspectives on the problem can try solve it. The link bellow this is the pull request where you can see the original file and compare it to the one I have done now to see my changes back. 

[Re-engineer microservice retrieveShowDuggaService #16979 #17099] (https://github.com/HGustavs/LenaSYS/pull/17099)

### DIFF
--- a/DuggaSys/microservices/showDuggaService/getShowDugga_ms.php
+++ b/DuggaSys/microservices/showDuggaService/getShowDugga_ms.php
@@ -1,12 +1,9 @@
 <?php
 date_default_timezone_set("Europe/Stockholm");
 
-// Include basic application services!
-//---------------------------------------
 include_once "../../../Shared/sessions.php";
 include_once "../../../Shared/basic.php";
-include_once('./retrieveShowDuggaService_ms.php');
-
+include_once "../curlService.php";
 
 pdoConnect(); // Connect to database and start session
 session_start();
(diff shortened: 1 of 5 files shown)


a22wilas (changes requested):
Have tried the different functionalities on the application.

In this PR the functionality to Load a dugga does not work. 
To test this I saved a dugga and then loaded using the hash and password from the table "useranswer".
bild
bild
bild
This is the outcome, no dugga is loaded in.

Now if I change to the weekly branch and refresh the web-application this is the outcome
bild


From going through the files the calling for a dugga comes from the file getShowDugga and is processed in retrieveShowDuggaService.

a23danme:
Saving the dugga and load it only works on the basic files without my edits, when not logged in, otherwise "could not save dugga" comes. I am writing to the issue responsible to see if someone else can take this task due to me being trash on it :D

Issue #17373, opened 2025-05-08
Fix function callMicroservicePOST


a22wilas:
At the moment the function callMicroservicePOST in curlService.php is broken. 
This needs to be fixed to it can be used to POST data between microservices.

HGustavs:
High prio

PR #17376, opened 2025-05-08
Update curlService.php

a22wilas:
Fixes issue https://github.com/HGustavs/LenaSYS/issues/17373

The function callMicroservicePOST should now work as intended.
Problem before was that the function retrieves an array as a parameter but then forces the retrieving microservice to read it as a JSON. 

Also updated the comments for the function to better explain what it does.

### DIFF
--- a/DuggaSys/microservices/curlService.php
+++ b/DuggaSys/microservices/curlService.php
@@ -1,22 +1,34 @@
 <?php
+
+//----------------------------------------------------------------------------------
+// callMicroservicePOST - Sends a POST request to a microservice.
+//----------------------------------------------------------------------------------
+
 function callMicroservicePOST(string $path, array $dataToSend, bool $returnValue = false) {
-    //set url course path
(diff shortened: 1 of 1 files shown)


a20artje (approved):
Services using this now work.

Is it correct to remove https and use http though?

b23johgu:
@a22wilas is there a reason to why you changed https to http? We probably should use https if it's possible.

a22wilas:
If it's on a web-server with a certificate we should use HTTPS.
For local production and testing most people probably does not have a certificate and thus need to use HTTP.

We could still have HTTPS and make in the instruction for the function to change to HTTP when creating the call to the microservice and in the review process. This would make sure that we have the correct setup for a server call while also describing what needs to be done locally to make the function work.

b23johgu:
Sounds good! Will plan for the future to change to HTTPS when we are provided with a webserver with a certificate. Until then, it's enough with HTTP.

Issue #17302, opened 2025-05-07
Weird Lable iteraction with "<"


a23carpa:
Noticed an odd interaction in the label text field. 
When numbers are after a < then some kind if figure is created.
<img src="https://github.com/user-attachments/assets/3bdec8d4-5a1c-441f-b618-6a84bda452e0" alt="Image" width="700" />

PR #17377, opened 2025-05-08
[#17302] Fixed unintended label interactions with "<" and ">"

a22nevan:
There's a converter built into the drawLine function that checks if the user has input the symbols "<" or ">". If they had, the input symbols would be converted to their HTML number values, most likely to avoid conflicting with any html that was to be sent by the rest of the function. There were however semicolons missing at the end of the HTML numbers, which caused further digits to change the HTML number completely. For example, if the user input "<24", the converter **should** have converted the string into "**_&# 60;24_**" which would've displayed as "<24" (_Had to put a space between "&#" and the numbers since it got converted in the text as well_). However since the semicolons were missing, the string was converted into "**_&# 6024_**", which is the HTML number for "ឈ", a Cambodian letter.

Semicolons have been added now so the label should no longer display incorrect symbols if "<" and ">" are included in the input.

image


a23matsh (approved):
This indeed seemed to solve the problem, approved

image

PR #17378, opened 2025-05-08
Documented changeActiveCourseVersion_sectioned_ms and createGithubCodeExample_ms

a23wilgr:
Documented changeActiveCourseVersion_sectioned_ms and createGithubCodeExample_ms. Not 100% sure of the outputs in either of them


a23danme (commented):
For changeActiveCourseVersion_sectioned:
Sorry for pointing out such small flaws :D.
on line 41 and 45 realtivedeadline is misspelled, should be relativedeadline (I think?).
on line 7 in description relvent should be relevant.

Otherwise I think the doc is good, dont find any faults and the different paramters etc.. is enough describing to understand.

createGithupCodeExample looks great couldnt find anything I would really change. 

The file has conflicts? Im not sure why I dont really know how that works but maybe look into it if you know how it works due to groupleader said we should fix it ourselves I think?

b23johgu:
I'm aware that this document will have conflicts and will help out with these! In this file it looks like a false conflict, I can solve that when it's time to merge :)

a23danme (approved):
Looks good!

PR #17387, opened 2025-05-09
G1 2025 v6 17382 delete detection of overlapping

a23thefo:
I have removed overlapping of all the elements. You can resize, drag, create and move with the arrow keys over other elements. No error is shown either.

image

### Known issue
The z-index dose not update when you pick up a new element making it rather jarring to use and should be solved in a different issue. 

image


a23eliax (approved):
I confirm that this is a working solution, and no error messages is shown!

image

PR #17389, opened 2025-05-09
Implement sub menu for the fab a23carpa

a23carpa:
Added four buttons to the Fab as per the mock up.
Buttons should appear when main button is clicked
image


a23emijo (approved):
This works as intended and shown in the video. I approve this! 💯

https://github.com/user-attachments/assets/836c4db7-f184-417a-bf66-34f75461bdee

PR #17397, opened 2025-05-09
Deleted a document and all images

### LINKED ISSUE #17235: Organize installation instructions to one place
The instructions for the installation of LenaSYS are currently all over the place. For easier maintanance and to minimize future confusion we should keep it to one place, which would be **README.md**. 

The instructions that are outdated or duplicates **outside of README.md** should be removed. E.g., the installation instructions for Docker is one of them which exists in both README, Wiki and in dockerenv/document/docker_install_instructions.md.

a23dilel:
Fixes #17235. I deleted wiki and Docker documents.


a23bensh (approved):
This removes all irrelevant and duplicated documentation for setting up Docker. I couldn't find much other documentation for XAMPP for example that could be removed, so this seems to resolve the issue. Looks good!

PR #17398, opened 2025-05-09
G3 2025 V6 #16696 change the jquery in accessed.js to javascript

### LINKED ISSUE #16696: Change the jquery in accessed.js to javascript
Swap the jquery in accessed.js to javascript

I think most of the things in it has to do with the top bar but not fully sure

If this seems to big then tell me and I can try to split this issue

a23lisjo:
Fixes #16696. These are changes from previous weeks that has not been reviewed yet.
There is a problem to create a new user. I'm not sure if changes in other parts of the code are conflicting with this vanilla JS.

### DIFF
--- a/DuggaSys/accessed.js
+++ b/DuggaSys/accessed.js
@@ -66,14 +66,13 @@ function setup() {
 
 //displays dropdown when hovering search bar
 function hoverSearch() {
-	$('#dropdownSearch').css({display:'block'});
-	$('#dropdowns').css('display', 'none');
-	$('#dropdownc').css('display', 'none');
+	document.querySelector('#dropdownSearch').style.display = 'block';
+	document.querySelector('#dropdowns').style.display = 'none';
+	document.querySelector('#dropdownc').style.display = 'none';
 }
 
 //stops displaying the dropdown when removing cursor from search bar
 function leaveSearch() {
-	$('#dropdownSearch').css({display:'none'});
 }
 
 // displays dropdown for the filter-button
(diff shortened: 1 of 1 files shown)


a23osckr (changes requested):
Code appears fine, but some errors were found on the site. As mentioned in the pull request, creating users does not work. When viewing the console the below error can be seen when trying to create a user.
image

The error seems to originate from this code trying to access a null variable.
image

A second issue I found was that the example text dropdown that appears when hovering the search bar does not disappear when not hovering anymore.
image

a23lisjo:
### Hover
The hover does work it is just very slow and not so reactive. Somehow it only disappears when you unhover on the right side. 

https://github.com/user-attachments/assets/be2017be-d728-4761-b482-7982364c0601

---
### Adding a user
The problem with adding a user is that the form has a function "addSingleUser" where it is expected a value like first name, last name, email and term. But it also expects values from elements with id: "addSsn", "addCid", "addPid". After checking in accessed.php there are no inputs for these:
image

When this happens in JQuery, the program can still process the form, only when changing to vanilla JS this is not working and errors occur because these are actually NULL. 

I've tried a few attempts with no real success:
**Attempt 1:**
I have tried to add hidden inputs just like "<input id='uid'>" for these id's. 
image
I do not get an error but the form is not sent away and does not add the new user.

**Attempt 2:** 
Only comment out the .push and for the ones that has no input
image
Error:
image


**Attempt 3:**
 fully comment out the elements not having inputs and changing the array order
image

At first this worked but added the user information in the wrong order (Johan is correct, Lars is not). After trying again it has stopped working. 
image

PR #17406, opened 2025-05-09
Add Microservice directory database installation to New Installer #17386

### LINKED ISSUE #17386: Add Microservice directory database installation to New Installer
Right now you need to go to the folder microservices/endpointDirectory to install the database and fill it with data. Include it in the new installer so that users get it installed directly.

a23oskme:
In 'newinstaller', I created a script (setup_endpoint_directory.php) in the tools folder, that script will be included in the foreach loop in 'install_engine.php'. I also created an operation in 'install_engine.php' that runs the setup script. Fixes #17386 

To test this, run the new installer. After installing you should be able to open [microserviceUI.php](http://localhost/LenaSYS/DuggaSys/microservices/endpointDirectory/microserviceUI.php) and see data. If it's empty, something went wrong. 
The idea is that 'endpointDirectory_db.sqlite' should be created and filled with data after running the new installer. To easily confirm this you can manually remove it in the 'endpointDirectory' folder before running the new installer.

### DIFF
--- a/newinstaller/tools/install_engine.php
+++ b/newinstaller/tools/install_engine.php
@@ -155,6 +155,9 @@ class InstallEngine {
 		$operations["init_db"] = function() use ($installer, $verbose) {
 			$installer->execute_sql_file("SQL/init_db.sql", verbose: $verbose);
 		};
+		$operations["setup_endpoint_directory"] = function() use ($verbose) {
+			setup_endpoint_directory($verbose);
+		};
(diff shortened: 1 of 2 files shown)


a23dilel (approved):
Here is MicroserviceUI without data:
data

When I use the new installer and install LenaSYS, the data is created and works as intended when I return to MicroserviceUI:
data2

PR #17415, opened 2025-05-12
Changed to curlService in retrieveHighscoreService_ms and highscoreservice_ms  #17117

### LINKED ISSUE #17117: Change the microservice retrieveHighscoreService  to use "curlService"
This file in /highscoreService has been re-engineered but has to be changed to use the new "curlService" functions. You will need to include the the "curlService" file. You need to go through both the file you get in the issue and the inverse dependencies files of the file you get.

a23louma:
Fixes #17117 . I've changed the microservice retrieveHighscoreService_ms and its inverse dependency file highscoreservice_ms to use the curlService functions for POST. These files weren't re-engineered correctly before since the cURL options for POST weren't set and the data wasn't received in retrieveHighscoreService_ms. From what I have found it's still the old highscoreservice.php being used and not these microservices so I haven't been able to test it properly. It should be changed in the future to use these files instead. The reviewer can tell me if you find something different.

### DIFF
--- a/DuggaSys/microservices/highscoreService/highscoreservice_ms.php
+++ b/DuggaSys/microservices/highscoreService/highscoreservice_ms.php
@@ -37,23 +37,13 @@ logServiceEvent($log_uuid, EventTypes::ServiceServerStart, "highscoreservice.php
 // Services
 //------------------------------------------------------------------------------------------------
 
-//Re-engineer
-$baseURL = "http://" . $_SERVER['HTTP_HOST'];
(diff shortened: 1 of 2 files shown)


a23dilel (changes requested):
I got this error:
image

This funtion is undefined, because you put wrong directory curlService.php

To fix in highscoreservice_ms you need to change directory from `include_once "../microservices/curlService.php";` to `include_once "../curlService.php";`. Same apply for retrieveHighscoreService_ms.

a23louma:
Changes are done and this PR can be merged now.

Issue #17316, opened 2025-05-08
Snapping to lifeline occurs infinitely along the vertical line


a23aleva:
When snapping a sequenceActivation element onto lifeline outside of the lifeline's own bounds a snap still occurs. This should not be the case and therefore ought to be fixed.

https://github.com/user-attachments/assets/1f214442-2cec-43bf-aac2-6faf6107c21f

PR #17417, opened 2025-05-12
Fixed: G1 2025 v6 #17316 snapping to lifeline occurs infinitely along the vertical line

a23aleva:
Added some check conditions that were missing on the y-axis of the lifeline causing elements to snap infinitely along the vertical line. This should not be possible anymore and snaps only occur to the actual lifeline.

https://github.com/user-attachments/assets/49bde8a2-cec7-4743-8530-47c2d5ffe69b


a23thefo (approved):
The solution solves the issue elegantly with minor changes to the code.
ezgif-45a148b5aa468a

PR #17418, opened 2025-05-12
Null-check for window resize in codeviewer.js

a23yutta:
Added null-checks for window resize in `codeviewer.js` to not throw unexpected error.


a23alera (approved):
The change works as expected, i consider this approved.

PR #17419, opened 2025-05-12
Refactored the diagram toolbar for mobile versions

a23gabst1:
I have refactored and improved the diagram toolbar for the mobile version of the diagram page. video 1 shows how the toolbar works as of now. You can now open sub menus and switch the active drawing element by clicking one of the sub menu elements.
 
This solution works by using data-attributes to store information about the different toolbar boxes and the active toolbar box switches data-attributes if one of the sub menu boxes are clicked. When choosing a element to draw, it shows which sub menu element that is active by checking if the active and sub menu element have the same dataset. 

test this solution on different browsers. 

video 1

https://github.com/user-attachments/assets/42b46832-c3a5-4f8f-8ee0-42bb0437b1ec

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -320,8 +320,9 @@
         display: block !important;
         position: fixed;
         z-index: 10;
-        bottom: 7%;
-        left: calc(50% - 2%);
+        bottom: 5%;
+        left: 50%;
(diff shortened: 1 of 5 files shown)


a22abdjo (approved):
After reviewing and testing the changes made on different browsers, the solution provided has been apporoved.

PR #17423, opened 2025-05-12
Fab button functionality works

a23carpa:
Added event listeners to each of the buttons, to call the same functions as the desktop buttons do.

[Screencast from 2025-05-12 11-04-35.webm](https://github.com/user-attachments/assets/f89a48ac-c050-4f45-8002-21778c5fb710)


a22nevan (approved):
The mobile FAB buttons now have functionality implemented and work just like their toolbar counterparts. Works just like the video above shows. 

Pull request can be merged without problems. 👍

Issue #17343, opened 2025-05-08
Undo button blocks other elements in course view


a22maxda:
The undo button, which appears when something is deleted in a course blocks other elements like the hide/show arrow and part of the view button.

Undo not visible:
<img width="365" alt="Image" src="https://github.com/user-attachments/assets/0301838d-7c6f-477a-9a7e-46440d9baed4" />
Undo visible:
![Image](https://github.com/user-attachments/assets/03e88dd5-e275-4c82-a336-396eaade495f)

The button should be moved so it is not in the way or removed entirely as there is a popup with the same functionality.
Popup:
<img width="458" alt="Image" src="https://github.com/user-attachments/assets/ce094ed2-7d05-441e-a017-2d2a719621d5" />

It can be permanently brought up by finding the #undoButton element using the inspect tool and changing the display value from none to block.

a23fanro:
Since the undoButton does something that is already implemented (in a better way) with the pop-up notification, the undoButton should be removed.

![Image](https://github.com/user-attachments/assets/145a0c55-1ee8-4728-8293-6d0fcfe1704b)

PR #17430, opened 2025-05-12
Deleted the extra undo button #17343

a23erigu:
I have deleted all mentions of the extra undo button in section.php, section.js and style.css

This is the button that has been deleted
image

How the button is found to test can bee seen in the issue #17343


b20isato (approved):
Yepp, button is gone and nothing else broke

Issue #17436, opened 2025-05-12
Adding more alternatives to sequence loop element breaks selection color


a23emisn:
An earlier fix I implemented (#17314) the issue with sequence elements not turning purple on selection. This seemingly fixed the issue for all sequence elements at the time. However it has been discovered that when adding alternatives to this sequence option loop, the color breaks yet again upon selection. See pictures for demonstration of what happens with the selected element.

Working:
![Image](https://github.com/user-attachments/assets/050e4f48-43ae-4caa-9381-15f100db1b1a)

Broken, not purple and text is white and can't be seen:
![Image](https://github.com/user-attachments/assets/42c92fc3-3bac-4439-ad79-af5fa05052dc)

PR #17439, opened 2025-05-12
Fixed multiple alternative sequence-loop highlight color

a23emisn:
Fixes issue: #17436, the issue was that when adding alternatives to sequence loops, it broke the color manipulation of the element (my fault).

The small text area dynamically changes position relative to its parent (in the html-structure) when alternatives are added to the element. But since the color was manipulated with a static child reference, the color manipulation broke when adding alternatives. To fix it I added an ID to the <path> for the small top left text area and changed the color-updating logic filter on that instead of a statically defined child. This solved the issue and is much more reliable.

Before: 
image

After:
image

### DIFF
--- a/DuggaSys/diagram/draw/element.js
+++ b/DuggaSys/diagram/draw/element.js
@@ -873,18 +873,19 @@ function drawElementSequenceLoopOrAlt(element, boxw, boxh, linew, texth) {
     }
     //svg for the small label in top left corner
     content += `<path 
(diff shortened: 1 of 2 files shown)


a23emijo (approved):
This works as intended and shown in the video, I approve! 🥇 

https://github.com/user-attachments/assets/71ef7aaf-f346-4b71-b2c3-f393e2562c9b

PR #17441, opened 2025-05-12
Fixed undo and redo saved names

a22abdjo:
Fixed undo/redo logic for elements names to handle full words. Redo now removes entire words instead of fixed character chunks and accidental deletion of elements with short names. It also ignores whitespace as a change trigger and improves text change tracking by logging per-word edits only.


a22abdjo:
https://github.com/user-attachments/assets/e759acd0-350a-4514-85c8-5a7bd6426d43

a23eliax (approved):
I confirm that this is a working solution and it removes letter by letter and not just a part of the word!

PR #17442, opened 2025-05-12
Name of current file is being displayed to the user

b23ebbst:
Added an element that displays the name of the current file for the user. If no file has been opened, the div is not displayed at all. When a user is changing between files the name updates along with the file being loaded. All of the logic is handled at the same time as the file is loaded (in the same function), so there is no delay in showcasing the name or loading the diagram.

Currently the only way to display the name (so that the only place it shows up on is the diagram pages) is to display it in the drawing-area (or on top of the ruler) which might not be the most convenient place but seems to work well enough for the most part. 

https://github.com/user-attachments/assets/326194ef-2464-4435-be22-895e0d2feb30

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -71,6 +71,17 @@
     overflow: hidden;
 }
 
+#diagramFileName {
+    position: absolute;
+    z-index: 1000;
+    margin-left: 50vw;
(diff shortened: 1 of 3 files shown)


a23kevre (approved):
Great feature, changes approved.

Issue #17402, opened 2025-05-09
FIX error when trying to input a github repo link (Popup window) -= sectioned.php & sectioned.js =- Found at - <div id="githubPopupWindow"....</div> & validateForm()


a23elism:
### A new problem has appeared with the "Refresh Github Repo" button in the nav-bar
Seems someone forgot to update the actual popup window for adding a repo link and didnt include the repo key field causing errors 
and only edited the background logic

### Currently the pop-up when no repo is linked to the course only has 1 field for input (the repo url)
The problem is that the internal code currently asks for the repo url **AND** a repo key without actually letting the user (most likely teacher) enter a "Github Repo Key" in the field, from initial inspection, all that is needed is that a field for entering repo key also needs to be added (possibly also to actually make the addition of a repo key optional, currently seems like its "optional" whilst not being optional)

![Image](https://github.com/user-attachments/assets/c4cd4283-3137-4cda-a903-dc3454734da9)

### Whilst directly editing the course from the courseed.php site has the key as optional:

![Image](https://github.com/user-attachments/assets/995010b6-87e6-46dd-821f-66527c1a2ce9)

### Here is the structure of the pop-up window in sectioned.php
 
![Image](https://github.com/user-attachments/assets/472b454d-7578-4cc5-939a-a0f2f6adde53)

### And here is the part that validates the information in sectioned.js from what you enter in the popup in the function validateForm()

![Image](https://github.com/user-attachments/assets/fcceacee-3747-47c0-aa00-0cbea28cba3c)

a23elism:
### Recommend checking console for additional information

a23fanro:
We are working on figuring out github fetching: https://github.com/HGustavs/LenaSYS/issues/17496
Will close and move this issue to our backlog in the meantime.

Issue #14781, opened 2024-04-22
Update documentation regarding Github integration


a22willi:
After completing issues:
- #14391 
- #14563 
- #14339 
- #14370 
- #14526 
- #14630 

The github integration should be more or less working, update the [wiki page](https://github.com/HGustavs/LenaSYS/wiki/Github-Integration) to reflect these changes.

Issue #13845, opened 2023-05-08
Roadmap - Github integration


d21emijo:
# Roadmap For all Github functions

The Github integration roadmap is for how the fetched data from Github and put into the sqlite db is being used.

![image](https://user-images.githubusercontent.com/102797583/236827312-d29d1423-79be-4102-aa4f-fd2f330df86c.png)

one should be able to use the Github icon and choose a dir in the Github repo to get all the codeexamples... tbc 

##  Order of operation
- [x] Investigate use of subfolders for github examples in file editor 
#13511
- [x] Make refresh button run the download script 
#13943 
- [x] Github Folder Structure for Github Integration 
#12951
- [x] Connect the github button in moment with ability to generate examples 
#13897
- [x] Code examples should automatically be created and named after the files in the folder 
#13549
- [x] The refresh button on moment type needs a URL to function 
#13456
- [x] Code example New Github File 
#13879
- [x] Moments should automatically be created and named after the directories in the github folder 
#13690
- [x] .md or .txt .index file that we can use to control the ordering of code examples 
#13016
- [x] New code example should not affect existing order of code examples 
#13017

##  Handling of code examples from github

Created from: #13875, #13877, #13879, #13880  
###  If a code example comes from the github repo there is a certain locking behavior that is important
Reference case from Henrik: https://github.com/LenaSYS/Webbprogrammering-Examples/tree/master

Issues from Henrik:
- [x] It should be possible to hide the example in the course view
- [ ] It should be possible to reorder examples within a heading but not between headings
#13932
- [x] We should be able to rename examples in the course view at will.
#13933
- [ ] Code example locking for github files 
#13875
- [x] When a new example is added, it should be placed added at the bottom of the list. To then be manually reorder with the normal ordering operation. 
#13935
- [x] When an example is removed from github it should be hidden in the course view and not deleted
#13931
- [x] If a new file is added on github in a folder that is tracked by the synchronization module we want this file to be added automatically to the course view. (Might be broken down)
#13934
- [x] New files should not be hidden by default, but be shown immediately after the first refresh (manual or automatic refresh).
#13936

##  Manual and automatic synchronization option 
Created from: #13880
- [x] Synchronization of github repo should run automatically if the course is "loaded"
#13987 
- [x] Oversaturate github if a user re-clicks the refresh button/buttons
#13999
- [x] Stop the user from sending to many api request under a short amount of time. 
#14012
[...]

PR #17448, opened 2025-05-12
Fixing GitHub repo key #17402

a23erigu:
I have just added a way for you to insert a github repo key.

I **think** this has solved the issue (at least it appears to work with normal repos now)
But I can't check if this works with a repo that actually needs a key as the keys are weird/hard to set up.

Just want to add that the entire linking github repos feature seems unstable at the best of times and I'm not even sure what it even does/achieves.

picture of how it looks now
image

### DIFF
--- a/DuggaSys/sectioned.php
+++ b/DuggaSys/sectioned.php
@@ -774,6 +774,7 @@
 				<input type='hidden' id='cidTrue' value='<?php echo $_GET["courseid"];?>'/>
 				<form action="" method="POST" id="repoLink">
 					<div class= 'inputwrapper'><span>Github repo link:</span><input onchange="checkGithubLinkClue('gitRepoURL')" oninput="checkGithubLink('gitRepoURL')" type="text" id="gitRepoURL" class="textinput" name="reponame" placeholder="https://github.com/username/repository"/></div>
+					<div class= 'inputwrapper'><span>Github repo key (optional):</span><input type="text" id="gitAPIKey" class="textinput" name="repokey" placeholder="12345"/></div>
 				</form>
 			</div>
 			<div class="formFooter">


c22isako (commented):
**It does seem like the linking gives the support of creating code-examples "automatically" from the inserted repo.** 
Additional information about github integration can be found here: https://github.com/HGustavs/LenaSYS/issues/13845

When testing, I created a new repo just to try the features out. To be completely honest, I'm unsure if this is testable at all, since several issues within https://github.com/HGustavs/LenaSYS/issues/13845 mentions going through a web-server, which we do not have access to. Either way, the documentation doesn't seem to be updated as well (see https://github.com/HGustavs/LenaSYS/issues/14781) which makes it even more complicated to test.

**I do believe that the "fixable" issues within this PR is error-handling. Ex. providing an error message for when the wrong length of an key is being typed, or when only a key is inserted.** - more on this within title "Issues which can be seen without having a valid key / webserver:" below.

---
### Fixes done within this PR:
- **Github repo key field is added, providing the useability of repo keys to access certain private repos in a protected manner.** That said, I cannot confirm the functionality without a repo with a ssh key.
- **You can enter a Github repo link without having a key - so the keys are fully optional now.** (From what I can see, unsure how the database treats it)
---
### Issues which can be seen without having a valid key / webserver:
Worth to note, as said in the PR, the entire feature seems unstable. I'm not sure how many errors are created from this issue rather than overall issues within the implementation of "linking github repos"

- **No error displayed when only repo key is inserted, without repo link.** 
The process is not mitigated though, so it's not really a large problem. That said, it could be more user-friendly with feedback regarding the wrong-doings. For reference, this **_might_** be handled for the link via the "onchange="checkGithubLinkClue('gitRepoURL')"  within the input wrapper for the **github repo link**, and oninput="checkGithubLink('gitRepoURL').

- **You can enter a repo link together with an invalid key, which gives the error message seen within _image1_.** 
There is no feedback regarding if you have access to the repo or not. To fix this, the user could go to the "main menu" (../courseed) & edit the course from there, providing it with the right key. That said, it would be hard to know why nothing works without an feedback message. There also seems to be more error-messages within the course-editor found within courseed, seen within _image2_.

image
_(image1)_

image
_(image2)_

- I encountered the errors within image3 when having created a new course with a github repo link, and then pressing "refresh github repo" and entering the repo once again (with an incorrect key as well.) **This might be an entirely different, and probably bigger issue.**

image
_(refresh github repo)_

image
_(image3)_

c22isako:
OBS review not done

a23fanro:
We are working on figuring out github fetching: https://github.com/HGustavs/LenaSYS/issues/17496

Issue #17364, opened 2025-05-08
document microservices in microservices/gitFetchService


b23johgu:
Files to document:

- [ ] bfs_ms
- [ ] downloadToWebServer_ms
- [ ] getGithubURL_ms
- [ ] getIndexFile_ms
- [ ] insertToFileLink_ms
- [ ] insertToMetadata_ms

Use the latest template, this can be found at: /DuggaSys/microservices/Microservices Documentation/template.md

PR #17449, opened 2025-05-12
G2 2025 v6 a20artje 17364

a20artje:
Did the documentation for the files:

bfs_ms
downloadToWebServer_ms
getGithubURL_ms
getIndexFile_ms
insertToFileLink_ms
insertToMetadata_ms

Fixes issue #17364

### DIFF
--- a/DuggaSys/microservices/Microservices Documentation/gitFetchService.md
+++ b/DuggaSys/microservices/Microservices Documentation/gitFetchService.md
@@ -19,7 +19,7 @@ Traverses a GitHub repository using a breadth-first search (BFS) approach to ret
 ## Calling Methods
 - POST
 ## Output Data and Format
-- Output
+- Output: $response, $json['sha']
    - Type: JSON
    - Description: If option is GETCOMMIT the script returns a the commits SHA, otherwise no return unless an error occurs in which case an error message is echoed.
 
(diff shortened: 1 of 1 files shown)


a23louma (changes requested):
In the documentation of bfs_ms.php the output data is missing outputName “$response” (?) to follow the template, in the database UI it shows as “No outputs” for this file. In the template:
“- Output: outputName”.

The outputName is also missing for getGithubUrl_ms.php and getIndexFile_ms.php. 
The same for downloadToWebServer_ms.php, this will show as “No outputs” in the database UI but in the documentation: “No output unless error, in which case error message is logged.” 

The documentation for insertToMetadata_ms.php has two headers for “## Input Parameters”, just remove one of them.

Also in general (which I don’t know if it’s necessary) but according to the latest update of the template, the description of the input parameters should include how it’s stored in the database. I think it’s fine as it is now (also since it’s Tuesday) but another issue can be created to add this later for all the files:  
“- Description: Describe parameter. Stored as *varchar(256)* in the database”.

And if you have time (also according to the new template) add “---" between each microservice for clarity.

The rest looks good!

a20artje:
I fixed the issues brought up in the requested changes (I hope), I will create a new issue to add database storage description of parameters.

a23louma (approved):
Now it looks good, I approve!

Issue #17429, opened 2025-05-12
A4 template options change-up


a23aleva:
The vertical and horizontal option buttons, when the user clicks on A4 template in the options menu, look out of place and are located too close to the darkmode button. To keep the option menu solid and unified in its design please change these buttons to a dropdown that looks like the one for example diagrams.

![Image](https://github.com/user-attachments/assets/ab2e0108-5c22-4af7-95af-aa3136951725)

<img width="258" alt="Image" src="https://github.com/user-attachments/assets/6fb7a618-1864-4de4-aaec-c971654e02bf" />

<img width="256" alt="Image" src="https://github.com/user-attachments/assets/25aa7e4a-b8d8-4df5-826d-73d15a117085" />

PR #17452, opened 2025-05-12
A4 template options change-up (#17429)

b23ebbst:
Fixed the HTML, JS, and CSS to match with the "Example diagram"-button and functionality. The A4 template can now be toggled from a dropdown, and is not too close to the Darkmode-button. 

image

image

image

image


a23matsh (approved):
This looks much better and is indeed fulfilling the request from the issue, approved.

image
image
image

Issue #17384, opened 2025-05-09
Course-version dropdown not fully clickable


c22isako:
The dropdown for course-version isn't activated if you press its icon (see image) This seems to only be a problem on the "desktop" view.

![Image](https://github.com/user-attachments/assets/bc2be6ef-8bbb-47fb-89f7-5b8ad2161374)

PR #17453, opened 2025-05-12
G3-2025-V6-#17384-course-version-dropdown-not-fully-clickable

c22isako:
**Adjustments:** 
- Changed the width styling of _.course-dropdown-div_ to 100% instead of 90%, since it got brought down to the drop-down and removed interactivity in the last 10% (which housed the dropdown-arrow)
- Since this essentially made the width larger, I changed the max width to be a bit smaller - to possibly make up for it.
- I also added padding which prevents the text from covering the dropdown arrow, boosting readability.

---
**How to test:**
-> login
-> Enter course of your choice (ex. Testing-Course)

The affected element is the CourseVersion-dropdown marked in red within the image below:
image
It should now be fully "clickable" - so the "button" is expected to respond with an dropdown.

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -4371,8 +4371,7 @@ span.arrow {
   box-shadow: var(--standard-shadow);
   background-position: right;
   margin-top: 7px;
-  margin-left: 3px;
-  width: 90%;
+  width: 100%;
 }
 
 .course-dropdown-div select {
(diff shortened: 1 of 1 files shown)


a23elism (approved):
The simple changes have fixed the issue of the course version dropdown not being completely clickable.

The original version did not allow for clicking on part of the right side (around or on the dropdown arrow), whilst the new version fixed this and allows for complete interactivity whilst not breaking anything else

Approved for merge 👍

Issue #17080, opened 2025-04-30
Re-engineer microservice setAsActiveCourse using new "curlService.php"


a23louma:
This file has been re-engineered but has to be changed again using the new "curlService.php". OBS! You have to include curlService.php.

PR #17456, opened 2025-05-12
Active version using curlService

a22wilas:
Fixes issue https://github.com/HGustavs/LenaSYS/issues/17080

How to test:
Update the active version for a course and check the table course in the database!


a23bensh (changes requested):
It was a bit difficult to test this due to somewhat unclear instructions. From what I could determine, setting a course version as the default does not update the active version in the database.

I attempted to change the version to default and also looked through the code for any logic or call to setAsActiveCourse_ms.php, but couldn’t find anything that would actually trigger the update. Either additional changes are needed to fix the checkbox functionality, or I may have misunderstood how this feature is expected to work. Some clarification or adjustments would be helpful.

a23bensh (approved):
I checked once more and I seemed to have been in the wrong db, it works as intended, looks good!

PR #17459, opened 2025-05-12
Documented updateCourseVersion_sectioned_ms and updateListEntriesGrad…

a23wilgr:
…esystem_ms

Documented updateCourseVersion_sectioned_ms and updateListEntriesGradesystem_ms in the sectionedService.md. In updateCourseVersion_sectioned_ms the input parameter coursevers is included in the file, but is never used. Should it perhaps be removed from the documentation, file, or both?


a23axehe (approved):
Looks good, follows format and structure set in template

PR #17464, opened 2025-05-12
Deletes one or more list-entries from a sectioned list. Re-indexes th…

a23petny:
…e entries of a sectioned list to a new order.


a23louma (changes requested):
A new folder has been created "sectionedService" with four new files including a new template: template_ms.md, README.md, removeListEntries_ms.md and reorderListEntries_ms.md. This is not correct according to the standard.

Instead you should:
1. Continue the documentation for removeListEntries_ms.php and reorderListEntries_ms.php in the already existing file Microservices Documentation
/sectionedService.md
2. Follow the already existing template.md in the same folder(see the weekly branch for the updated template). This is necessary to show the correct information in the database UI.

A new review can be done when these things have been fixed.

PR #17466, opened 2025-05-13
Admin panel for endpointDirectory_db.sqlite #17414

### LINKED ISSUE #17414: Admin panel for endpointDirectory_db.sqlite
Create a page with buttons to delete, install and refill the database.

a23oskme:
Created a page where you can install, fill and delete the endpoint directory database. Fixes #17414 

On 'admin-endpointDirectory_db.php' you can do things based on the current status of the database. If its installed, you can delete it, if its not installed you can install it, if its installed but not filled you can delete it or fill it etc. 
When testing this you can use 'microserviceUI.php' to see if the database if installed or not, and to check if its filled. 

At the moment there are no login requirements.

### DIFF
--- a/DuggaSys/microservices/endpointDirectory/admin-endpointDirectory_db.php
+++ b/DuggaSys/microservices/endpointDirectory/admin-endpointDirectory_db.php
@@ -12,6 +12,14 @@ try {
     $db->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);
     $services = $db->query("SELECT * FROM microservices")->fetchAll();
 
+    // get microservices
+    $services = $db->query("SELECT * FROM microservices")->fetchAll();
+
+    // check if the table is empty
+    if (empty($services)) {
+        $dbEmpty = "Database is installed but empty.";
+    }
+
 } catch (PDOException $e) {
     $dbError = "Database is not installed.";
 }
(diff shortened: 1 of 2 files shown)


a23dilel (changes requested):
It looks good and works, but I would like to implement a pop up says “begin to fill database” or something, when I pressed the `Install database -> Fill database` and same thing `Setup database`. The `Fill database` and `Setup database` doesn’t have a pop up, otherwise user going to believe this link doesn’t work. 

Also, I don’t like it says only `fill` and `don’t fill`, and I know what these means, but not new user. I want to describe what buttons/links does.

The unlink code delete database, but not filename `(endpointDirectory_db.sqlite)`, so I would also like to remove filename also, so that `endpointDirectory` folder take less files.

a23oskme:
> It looks good and works, but I would like to implement a pop up says “begin to fill database” or something, when I pressed the `Install database -> Fill database` and same thing `Setup database`. The `Fill database` and `Setup database` doesn’t have a pop up, otherwise user going to believe this link doesn’t work.
> 
> Also, I don’t like it says only `fill` and `don’t fill`, and I know what these means, but not new user. I want to describe what buttons/links does.
> 
> The unlink code delete database, but not filename `(endpointDirectory_db.sqlite)`, so I would also like to remove filename also, so that `endpointDirectory` folder take less files.


Unlink is the proper way to delete a file using php <i>(see https://www.php.net/manual/en/function.delete.php)</i>. However, there can be issues with permission when using a local copy with XAMPP, which probably is the cause for it not getting removed (it doesn't get removed for me either). In an properly setup environment it should be removed. 

I added 'with data' to some texts, to make it a little bit clearer. Maybe we can make a separate issue if we still feel like we need to make it more user friendly?

There should be popups when filling the database. This one pops up when you `Install database -> Fill Database`
image
And this one when you use `Setup database`
image
Maybe you can check again? If this continues to be a problem I will have to research why they wont show up.

a23dilel (approved):
What I meant is that when the user presses the `Install database -> Fill database ` or  `Setup database ` buttons, a popup should appear point out that the process of filling data to the database has begun. This will give a information to the user that the database filling process has started and that they need to wait a few seconds, so let the user know, but I can approve this because the merge will start soon. So, please create an issue for this, if you want to, or scrap it.

Issue #17469, opened 2025-05-13
Elements snapping to lifeline


a23aleva:
After the collision detection was removed from the code it seems like all elements can snap to lifeline. Ideally the user should be able to place elements above the lifeline but the snap should not happen if the element in question isn't a sequenceActivation. Look into this.

https://github.com/user-attachments/assets/caa4746e-176e-4672-8d2b-c7aadb59bfba

PR #17470, opened 2025-05-13
Fixed: G1 2025 v6 #17469 elements snapping to lifeline

a23aleva:
Now sequenceActivation should be the only element to actually snap to different lifelines. Just added an extra check before snapping.

https://github.com/user-attachments/assets/444ed525-798b-4021-a38f-d5096ba31e39


a23kevre (approved):
snapping is fixed with lifeline, i approve.

Issue #17272, opened 2025-05-07
Lines only connect to the middle of sequence activation


a23eliax:
The lines between activations seems to only connect to the middle of the activation. This is a problem because you want to be able to have more lines (and straight ones) between the same activation boxes, just like in picture nr.2.

https://github.com/user-attachments/assets/1f8bb64a-7e71-4c18-a020-d25d6dad79cc

<img width="167" alt="Image" src="https://github.com/user-attachments/assets/81f05b1a-9dc9-4f39-8160-55aa3694ff49" />

PR #17484, opened 2025-05-13
Made sequence-diagram lines behave appropriately 

a23emisn:
The issue( #17272) was that the sequence diagram functionality was borderline useless in terms of how sequence diagrams are supposed to look and be used. Lines would automatically connect to the center of the sequence activation elements, just as they do with other elements. This would however defeat the purpose of a UML-diagram type that is meant to have clear, distinguishable and customizable transitions between activations along a timeline. 

The implemented fix has resolved the main parts of the original problem. Lines can now be drawn to user-specified anchoring points on the activations. Lines won't be rearranged by removal of other lines and the arrowheads are correctly placed on the line ends instead of being misaligned etc. However, repetitively using the undo/ctrl+z as well as redo/ctrl+y operations will result in strange behavior. But this issue was present prior to these implemented fixes. Given the major improvement of the sequence diagram behavior I suggest merging this now and opening up a new issue for fixing the potentially faulty undo-stack later. 

Pictures of using the sequence diagram before and after. 

Before: 
image
Note that the arrowheads for the lines are all attached to the center of the middle line and is in the wrong direction. The line anchoring points are centered to the middle of the each activation, regardless of user input. The lines will crease if the elements aren't placed at the exact same Y-coordinates etc.

After:
image
Exact same settings applied here, with the absence of the issues mentioned above.

### DIFF
--- a/DuggaSys/diagram/classes/element.js
+++ b/DuggaSys/diagram/classes/element.js
@@ -31,8 +31,6 @@ class Element {
         this.id = id;
         this.x = x;
         this.y = y;
-        this.anchorX;
-        this.anchorY;
         this.width = width;
         this.height = height;
         this.minWidth = minWidth;
--- a/DuggaSys/diagram/classes/element.js
+++ b/DuggaSys/diagram/classes/element.js
@@ -31,6 +31,8 @@ class Element {
         this.id = id;
         this.x = x;
         this.y = y;
+        this.anchorX;
+        this.anchorY;
         this.width = width;
         this.height = height;
         this.minWidth = minWidth;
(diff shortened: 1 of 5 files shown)


a23matsh (approved):
Indeed a working solution. You can now place lines anywhere on the object rather than them automatically repositioning to the center. Also the arrows appear to align with the lines better and in a more natural way than before.

 I did notice some strange behavior with undo/redo actions as mentioned after placing lines. This occurs when you place lines, undo them, and then redo the action. However, if you undo all the way until no lines remain, then place new lines and perform undo/redo again, the strange behavior doesn't seem to appear. Anyhow this seems to be a pre-existing issue from the master version and should be addressed separately in a new issue.

Approved.

image

Issue #17359, opened 2025-05-08
document microservices in microservices/courseedService


b23johgu:
Files to document:

- [ ] - changeActiveCourseVersion_courseed_ms
- [ ] - copyCourseVersion_ms
- [ ] - createCourseVersion_ms
- [ ] - createMOTD_ms
- [ ] - createNewCourse_ms
- [ ] - deleteCourseMaterial_ms
- [ ] - getCourseed_ms
- [ ] - retrieveAllCourseedServiceData_ms
- [ ] - retrieveCourseedService_ms.php
- [ ] - specialUpdate_ms
- [ ] - updateActiveCourseVersion_courseed_ms
- [ ] - updateCourseVersion_ms
- [ ] - updateCourse_ms

Use the latest template, this can be found at: /DuggaSys/microservices/Microservices Documentation/template.md

PR #17490, opened 2025-05-13
G2 2025 v6 #17359 Document microservices in microservices/courseedService

d22marwi:
Document microservices in microservices/courseedService, fixes issue #17359.

### DIFF
--- a/DuggaSys/microservices/Microservices Documentation/courseedService.md
+++ b/DuggaSys/microservices/Microservices Documentation/courseedService.md
@@ -1,30 +1,28 @@
 # Name of file/service
 changeActiveCourseVersion_courseed_ms.php
 ## Description
-*Description of what the service do and its function in the system.*
+Changes the active course version by updating the activeversion column in the course table.
 
 ## Input Parameters
-*Parameters will be described in lists*
-- Parameter: paramName
+- Parameter: $opt
+   - Type: string
+   - Description:
+
+- Parameter: $cid
    - Type: int
-   - Description: describe parameter
+   - Description: Course ID
 
-- Parameter: paramName
-   - Type: int
-   - Description: describe parameter
-
-
+- Parameter: $vers
+   - Type: string
+   - Description: New version ID
 
 ## Calling Methods
 - GET
-- POST
-- etc.
 
 ## Output Data and Format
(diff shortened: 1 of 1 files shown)


a23louma (changes requested):
The input parameter $versid for changeActiveCourseVersion_courseed_ms.php has type “varchar”, this is how it’s stored in the database which can be added in the description of parameter. For Type it should be for example string/int according to the latest update of the template (only in the weekly branch, I guess you branched out before this was merged).
I saw that it’s the same for all the documented files, so all “varchar” and “datetime” and so on should be changed but added to the description. As the template was changed during this week, I think it’s fine for now and a **new issue** can be created to change this next week instead. I have been checking the GetOP() function in basic.php and I think they all default to string when written like this “$opt = getOP('opt');”, just something to check on and keep in mind.

Also, for all files “## Examples of Use `CODE`” can be added, right know they are all empty. This can be fixed in a **new issue** instead.

In the latest update of template.md, microservices should be separated with “---” for clarity. This can also be fixed in a **new issue** (since it’s Tuesday and maybe not enough time).

**Things to check/change now before merge:**
For createMOTD_ms.php, I couldn’t find the input parameter $makeactive, am I missing it somehow? Otherwise, this should be removed.

The input parameter “$userid” is missing for createNewCourse_ms.php.

Remove “NO INPUT PARAMETERS” from getCourseed_ms.php, it’s better to follow the template, I don’t know if it can give issues in the database UI otherwise. Also remove basic.php and sessions.php from “### Microservices Used” for this file.

The same for retrieveAllCourseedServiceData_ms.php, remove “NO INPUT PARAMETERS”. Also remove basic.php and sessions.php from “### Microservices Used” for this file.

For retrieveCourseedService_ms.php, I think the input parameter “$userid” should be there as well? I’m also wondering about the Output data and format for this file, isn’t it only an array returned? And for microservices used basic.php and sessions.php can be removed. It also says “logServiceEvent()”, is this function call related to a microservice that should be there instead or should it be removed as well?

For specialUpdate_ms.php, updateActiveCourseVersion_courseed_ms.php, updateCourse_ms.php and updateCourseVersion_ms.php, remove basic.php and sessions.php from microservices used.

PR #17491, opened 2025-05-13
Display message positioning fixed on mobile

a23emijo:
Fixed diagram message positioning on mobile because it got cut off by the options-panel. I moved it to the top left corner instead and changed the width on the box. I think the top left corner is the best place because the toolbar and the FAB is at the bottom of the screen. 😄

https://github.com/user-attachments/assets/63f6d423-6ee1-4211-9c85-1ea80ea299fd

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -337,6 +337,13 @@
         transition: var(--normal-transition);
     }
 
+    #diagram-message{
+        position: absolute !important;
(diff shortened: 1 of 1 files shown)


a23gabst1 (approved):
This works as a charm, brilliantly done Emil! The code looks neat and nice, and follows mobile standard. 

ChillGuyMyNewCharacterGIF

PR #17502, opened 2025-05-13
17390 fix z index on latest interacted with element

a23vinge:
!!(I should note that the original purpose of the issue has evolved—from automatically bringing an element to the front when selected, to giving users manual control over which elements should appear in front or behind. This gives users more flexibility when arranging their diagrams.)!!

This pull request, that has been given life by me, adds two buttons in the options panel that let you change the layer order of elements on the canvas. When one element overlaps another, you can now choose to move it forward or backward by adjusting its z-index.

The functions work by finding the current highest or lowest z value and updating the selected element accordingly. Then the entire canvas is redrawn to reflect the change.

I’ve tested this with different kinds of elements and it behaves as expected. This should make arranging elements much easier, without needing to delete and recreate them.

I have added comments that should give context to the code

https://github.com/user-attachments/assets/8266d829-5af5-4282-b1be-2becceb1fbee

The video shows that the objects are being pushed forward and backwards depending on what button is pressed.

### DIFF
--- a/DuggaSys/diagram/draw/element.js
+++ b/DuggaSys/diagram/draw/element.js
@@ -11,7 +11,7 @@ function drawElement(element, ghosted = false) {
     let boxw = Math.round(element.width * zoomfact);
     let boxh = element.height ? 
     Math.round(element.height * zoomfact) : 0; // Only used for extra whitespace from resize
-    let zLevel = 2;
+    let zLevel = element.z ?? 2; /*nsures that elements without a defined z-index get a default value (2)*/
     let mouseEnter = '';
 
     canvas = document.getElementById('canvasOverlay');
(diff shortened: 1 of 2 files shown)


a23aleva (commented):
For some elements this new feature seems to work as intended, at least when moving up or down one layer. Unfortunately there also appears to be some issues with the new feature as some elements get stuck to the grid after moving them back too many steps through many other elements. UML initial and final states don't really switch layer at all as far as I've tried. For some reason elements disappear completely when sending them to back and there is nothing behind them (at least sometimes, though this is somewhat inconsistent). As long as you don't target something else, bringing them to front brings them back again but otherwise they evaporate.

Overall this new feature is a welcome one and the developer clearly has put thought and effort into it. Regrettably it appears quite buggy but perhaps it can be merged as a sub-issue to this parent issue, where another sub-issue is added to mend these problems in a later fix - seeing as all bugs are connected to the feature itself and won't affect things outside of it. Since the merge is closing in it'd be a pity to bereave the developer of activity. I don't know how much time these bugs take to fix but ultimately it'd be best if someone within leadership takes the final decision regarding whether this is mergeable or not - or if there is a possible workaround.

https://github.com/user-attachments/assets/8fb4bb28-d39e-4925-be32-ad3fbdffa4ae

a23miksk:
I will merge this pull request based on Alexanders comment since it doesn't break anything in the current program. The implementation seems to be filled with bugs, but they can be fixed with further specific issues.

PR #17510, opened 2025-05-13
Fixed-sequence-table-recursive-relation-labels

a22abdjo:
Removed the redundant label from options.js. Added a condition statement in line.js to customize label positioning for SE recursive lines, also the label should now appear after saving.


a22abdjo:
https://github.com/user-attachments/assets/05e9edd9-6fe0-41fb-ad15-9296a4b65776

a23thefo:
The solution works just as demonstrated in the video. The code is up to standard! 👍

a23thefo (approved):
The solution works just as demonstrated in the video. The code is up to standard! 👍

Issue #17445, opened 2025-05-12
Documentation/comments in diagram.php


b23danhe:
Review diagram.php and add comments or documentation where necessary.
Many functions currently lack clear explanations, making the code harder to follow. Keep comments concise and aim to describe the purpose of multiple lines of code when possible.

Clean up and remove any comments that are unclear or no longer relevant, if necessary.

PR #17512, opened 2025-05-13
Documentation/comments in diagram.php (issue #17445)

b23ebbst:
Went through the entire _diagram.php_ document to add comments to make it easier to follow and read the code. 
Also updated indentation where I though necessary to up the readability. 

It is worth mentioning that the parts of the code that have been commented-out, has been left as-is. This is because I am unsire of their importance to the current goal of making a mobile version.

### DIFF
--- a/DuggaSys/diagram.php
+++ b/DuggaSys/diagram.php
@@ -1346,9 +1346,10 @@
         </div>
     </div>
     </div>
-    <!-- Replay-mode -->
+    <!-- REPLAY-MODE -->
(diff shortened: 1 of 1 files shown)


a23thefo (approved):
Wonderful comments makes the experience of understanding what is going on that much easier! Every little part has a overarching comment giving an overview on what is happening and that is enough because diagram.php is just an glorified HTML file.

Good job! Gold star ⭐ 👍

Issue #17345, opened 2025-05-08
Unable to log out while in codeviewer.php


a23yutta:
The log out button does not function while in codeviewer.php

PR #17531, opened 2025-05-13
Unable to log out while in codeviewer.php #17345

a23elism:
Added the necessary script src for react into codeviewer to enable the logout feature (since it was react based)

Logging out should work now from the codeviewer


a23elism:
### For testing simply go into any codeviewer.php page and attempt to logout using the "logout button" in the navheader


image

a22carsv (approved):
After testing and comparing before/after, this seems to solve the issue of not being able to log out while in any codeviewer.php (button not responding), does not seem to affect any other component of codeviewer of what i can see. Great job.

Issue #17437, opened 2025-05-12
Fix so you can see all colors in mobile version


a23gabst1:
As of now it is only possible to choose black and almost blue as color when the user is on the mobile version of the diagram page.  Fix so the user can see all possible selection of colors. 
figure 1 

![Image](https://github.com/user-attachments/assets/e7934c4f-1848-4c00-b1b5-d6c7c5e78a56)

PR #17544, opened 2025-05-13
Fixed: G1 2025 v6 #17437 fix so you can see all colors in mobile version

a23eliax:
This change simplifies the positioning logic and reduces unnecessary calculations. The whole color menu is now visible on both desktop version and mobile version. When the user clicks on a color, the color menu closes as it was before.

Desktop version:
image

Mobile version:
image

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -574,9 +574,7 @@
     border-radius: 5px;
     position: absolute;
     left: 0;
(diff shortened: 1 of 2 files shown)


b23ebbst (approved):
Looks good! The panel for all colors look and act the same on mobile and desktop. Well done, I approve this change!

image
image

PR #17549, opened 2025-05-13
Hid the panels on mobile

a23emijo:
Hid the Instructions, General information and the whole header with all its funtionality. To do this, I had to add an ID to the header to be able to use it in a media query. The other panels already had IDs so I just used them in already existing media queries.

_**This is how it looked before**_
image

_**This is how it looks right now**_
image

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -315,7 +315,6 @@
     }
 }
 
-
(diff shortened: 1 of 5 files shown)


a23miksk:
A better solution has been implemented here https://github.com/HGustavs/LenaSYS/pull/17765, so I will close this.

PR #17550, opened 2025-05-13
G1 2025 v6 17428 documentationcomments in optionsjs

a23thefo:
## What I actually solved
It was mostly what i will assume AI typos which made some comments very hard to understand because they were barley English. I have updated them to actually describe what they do.  Thankfully Sol Milleding (My GOAT🙏) last year made some outstanding comments on the majority of the file so it was not that much work!

Like this was one of the comments this inst even English like what!?
image

## Flaws with the code it self that should be checked out
We should consider renaming:
select()
radio()
and option()

They do a way more specific job then their names makes you believe. 

But the opposite is for:
lineLabel()

Which is way more generic then what its name entails. 

All so I do question the necessity for the function setLineLabel() and where its used. Kinda feels like there is a easier way of doing that but I'm still new to the code

### DIFF
--- a/DuggaSys/diagram/draw/options.js
+++ b/DuggaSys/diagram/draw/options.js
@@ -443,7 +443,7 @@ function includeSELabel(line) {
 }
 
 /**
- * @description Creates two editable labels for a lines cardinalities.
+ * @description Creates two editable labels for a line's cardinalities.
  * @param {object} line A line that have an start and end cardinality as properties.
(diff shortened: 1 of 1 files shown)


a23thefo:
After making the PR I realized i missed somethings so dont be alarmed by the additional commits 😅

b23ebbst (approved):
The documentaion-comments are actually readable and understandable now. Really well done, I approve of the changes! 

Good job!

PR #17551, opened 2025-05-13
[#17444] Revised comments and documentation in diagram.js

a22nevan:
Went through the entirety of diagram.js and updated the grammar and punctuation of existing comments and function descriptions. Added new comments and function descriptions where it was deemed necessary. Also updated the region tags in the document to show in the sidebar, providing easier navigation. Added a new region at the bottom for mobile code, where most of it seemed to be located.

Removed some comments and descriptions that didn't seem to have ties to any still existing code. Also added documentation to a function in helper.js that was _only_ used in diagram.js and nowhere else.

This should hopefully improve the clarity and ease of understanding when it comes to the many functions and classes in the file. Most, if not close to all undocumented code should be covered now. There are probably still some minor improvements to be made that I might've missed or not thought of, but this should suffice for a general overhaul/revision of the documentation.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1,4 +1,6 @@
-//#region ================================ CLASSES =============================================
+// #region CLASSES
+// ============================================================================================
+// ========================================= CLASSES ==========================================
 
 /**
  * @description Handles storage and retrieval of usage history allowing undoing and redoing changes. Internal data should ONLY be modified through class methods to prevent weird behaviour.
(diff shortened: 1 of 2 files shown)


a23miksk (approved):
Very good, a lot of work is done here to make sure that the comments follows a high standard both in description of the code and English. 
Descriptions has been added to functions that previously didn't have any, and regions has been implemented to distinguish between blocks of code.
Approved!

PR #17568, opened 2025-05-14
Change 'setupEndpointDirectory.php' to use delete script #17467

### LINKED ISSUE #17467: Change 'setupEndpointDirectory.php' to use delete script
Change so it uses 'deleteEndpointDb.php' instead of manually deleting it before setting the database up.

a23oskme:
Removed the old code that removes the database (if it exists) before reinstalling it. It does the same thing as before, but there was a delete script made for a different issue, so it might as well be used here too. Fixes #17467 

To test this you can run `'setupEndpointDirectory.php'` and check when the `'endpointDirectory_db.sqlite'` was created. If you didn't have the database before running `'setupEndpointDirectory.php'` you need to run it two times.


a23gabli (approved):
Looks good, everything works as intended.

Issue #17396, opened 2025-05-09
 (desktop-view) BUG, Only animated in one direction


b20isato:
![Image](https://github.com/user-attachments/assets/0fede5ad-64f3-4d36-b5ce-099a1bffc562)
The View menu has a nice animation for when opening but nothing for for closing

a23simry:
@a23alera rebase or create new PR for V7 please.

a23fanro:
https://github.com/HGustavs/LenaSYS/pull/17570#issuecomment-2880006899

> Instead of creating a new CSS rule for the the class `.dropdown-content` at line 8696, please apply the necessary changes to the existing rule that begins at line 875. `dropdown-content.dropdown-content-show` should also be moved from line 8704 to somewhere closer to the 'original' `.dropdown-content` rule at line 875, I suggest right under it.

PR #17570, opened 2025-05-14
animated view menu for closing #17396

a23alera:
Replaced max-height with scaleY transform for smoother open/close animations. Added slight delay to opacity during close to enhance fade-out effect.

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -872,7 +872,7 @@ p {
 }
 
 .dropdown-content {
-  display: none;
+  display: flex;
   position: absolute;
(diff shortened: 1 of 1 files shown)


f23linan (approved):
The animation looks great! It makes such a difference. Nothing broken as far as i can tell either! Approved for merge :D

a23alera:
Changed the code like you said Changed the code as requested @a23fanro

a23erigu (approved):
As the first review said it works now with the code in the corect place and the animation look decently fine to.

PR #17572, opened 2025-05-14
Locate and change AJAX call to normal fetch() -= sectioned.js =-  in functions  - retrieveAnnouncementCards(), updateReadStatus() & toggleFeedbacks() #17223

a23elism:
Updated PR from G3-2025-V6 to G3-2025-V7

Original PR #17371 see this for comments on testing or general comments


a23simry:
When reviewing: testing might be difficult, some of the code is probably not used, but make sure site doesn't break.

a23fanro:
This has conflicts that need to be resolved before merge, moving it to 'to do'.

a23elism:
### Conflicts have been resolved recommend re-testing

f23linan (approved):
All the AJAX have been removed and the functionailty has stayed the same. Approved for merge!

Issue #17554, opened 2025-05-14
Align ER attribute name to the left


a23miksk:
In the acceptance test the customer expressed that he would rather that the first part of the name in an ER attribute be shown than the center of it when the name is larger than its container. Currently the name is centered which means that the only part that is shown when the name is large is the middle of it. 

_Video where the name fills out from the center which is unwanted behaviour_
https://github.com/user-attachments/assets/b06840d6-ec2e-476a-aae1-46b7bc4cc2d6

PR #17580, opened 2025-05-14
Fixed: G1-2025-v7-#17554-align-er-attribute-name-to-the-left

a23eliax:
This change makes the name of the attribute centered but if the name is too long, it shows the first part of the attributes name instead of the middle part of it.

### DIFF
--- a/DuggaSys/diagram/draw/element.js
+++ b/DuggaSys/diagram/draw/element.js
@@ -632,11 +632,32 @@ function drawElementERAttr(element, textWidth, boxw, boxh, linew, texth) {
             extra = `class='underline'`;
             break;
     }
+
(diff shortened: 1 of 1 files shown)


a23vinge (changes requested):
Hi, my apologies for the misunderstanding — the code does work as intended!

a23vinge (approved):
The original video highlights the issue where the attribute name fills out from the center, which is the undesired behavior. The new implementation correctly fixes this, as shown in the before-and-after videos below:

Before:

https://github.com/user-attachments/assets/4b6980d4-5e43-4dae-a81c-d0879622eb79

After:

https://github.com/user-attachments/assets/471046ea-21f2-4d1a-9cbb-45c50f597016

PR #17583, opened 2025-05-14
Replace table based navigation with semantic nav and ul elements.

b23sebth:
**This is a recreation of the now closed PR [#17543](https://github.com/HGustavs/LenaSYS/pull/17543).**
## What's changed:
- Navheader is now semantically correct.
- Inline styling removed and moved to style.css.
- Removed excess margin/padding.2

Some elements are not placed inside of the &lt;nav> element because it should only be used for actual site navigation. Functionality like editing/adding courses are instead grouped and placed in a separate unordered list. 

## How to test:
Everything should look mostly the same as before so please make sure to inspect the actual elements as well as checking desktop/mobile view to see that everything works as expected.  

image

## Known issues:
There are some problems in the navheader.php file. These are not introduced by me and also appear on the group branch. The hamburger menu also shows up to the right of the icon instead of directly under it on very small phone screens, but only when logged out.

image

image


a23alera (approved):
The changes work as expected. I tested the navigation and could not find any issues. The refactoring from table-based to semantic nav and ul elements looks clean and correct. I cannot find any problems.

Issue #17513, opened 2025-05-13
convert jquery in \DuggaSys\templates\3d-dugga.js


a23fanro:
Locate jquery in the file and convert it to javascript. 
DO NOT convert jquery that is used with AJAX, this will be done later.
Also remove code that has been commented out if it is jquery.

a23fanro:
For the next developer working on this: a23alera has come a long way translating this from jquery to js, but the functionality of the dugga seem to be trickier to solve. I suggest continuing from their work in PR https://github.com/HGustavs/LenaSYS/pull/17587 or starting over from the beginning, it's up to you. 

The broken functionality is well documented in this [PR:s comments](https://github.com/HGustavs/LenaSYS/pull/17587#pullrequestreview-2850347512).

PR #17587, opened 2025-05-14
Removed all active jQuery from 3D dugga to JavaScript #17513

a23alera:
Removed all active jQuery from 3D dugga to java script.

### DIFF
--- a/DuggaSys/templates/3d-dugga.js
+++ b/DuggaSys/templates/3d-dugga.js
@@ -1,9 +1,9 @@
 //----------------------------------------------------------------------------------
-// Globals
+// Globals (declared elsewhere so must be vars)
 //----------------------------------------------------------------------------------
-let score = -1;
-let dataV;
-let timer=0;
+var score = -1;
+var dataV;
+var timer=0;
 const DEFAULT_CAMERA_POSITION_Z = 1000;
 const	DEFAULT_CAMERA_POSITION_X = 500;
 const	DEFAULT_CAMERA_POSITION_Y = 200;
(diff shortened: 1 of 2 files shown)


b23sebth (changes requested):
**Changes needed - not approved**

## What's changed: 
- jQuery have been converted to regular JS
- Commented out jQuery have been removed

## Changes needed:
- Inconsistency with variables. While use of `const` is great, `let` should be used instead of `var` if the variable does not require global scope. Use `const` for variables that won't be reassigned.
- Remove inline styling and use style.css instead.

a22carsv (changes requested):
### Changes needed, not approved
Existing JQuery seems have been converted in entirety to vanilla JS. However, some issues have been caused due to changes made in this PR. When comparing functionality of 3D dugga in group branch contra the one for this PR, it seems like changes have caused some functionality to break. For example, when pressing the "LÄGG TILL" button on the PR branch, nothing happens:
image
While in the group-branch rendition of the '3d-dugga.js' file the functionality of the button is intact:
image

Furthermore, nothing occurs after pressing the "Triangle" button following the changes of this PR:
image

## Summary
### Changes made:
- Conversion to vanilla JS from JQuery.
- Addressed some of the earlier feedback.
### Changes needed:
- Troubleshoot cause of break in functionality and revise changes to address this.

a22maxda:
In case merging breaks this in the future (like it has before) these are the affected lines in style.css.

Hidden and Visible are from this PR and the rest is from the group branch.

Screenshot 2025-05-20 at 15 21 42

PR #17593, opened 2025-05-14
Fixed and reduced whitespace in UML Class, IE and SD elements

a22abdjo:
Reduced the whitespace in the bottom section under "Attribute" and "Function", to make the UML class element more compact. The same proccess was implemented to the IE ("Primary" and "Attribute") and SD ("do: func") elements. The changes made were done in the element.js file.


a22abdjo:
Before:
image

a22abdjo:
After: 
image

Issue #17562, opened 2025-05-14
Fix the symbols size


a23miksk:
The customer thinks that the symbols are too big. He thinks that they should probably have the same height as the font used in the elements. Is there a standard written down somewhere that explains how big the symbols should be?

_Image where the diamonds probably are too big_
![Image](https://github.com/user-attachments/assets/431de25d-f4c3-4117-a4c9-73ca4dce5c96)

PR #17596, opened 2025-05-14
Fixed: G1-2025-v7-#17562-fix-the-symbols-size

a23eliax:
Fixed so the diamonds has the same height as the triangles, like the costumer wanted. I just hade to change the coordinates a bit. Personally I think it looks a bit small, but if the costumer wants it that way, then that way it is!

image

zoomed in with straight line as comparator:
image


a23kevre (approved):
they are the same height now, i approve.

Issue #17516, opened 2025-05-13
convert jquery in \DuggaSys\templates\boxmodell.js


a23fanro:
Locate jquery in the file and convert it to javascript. 
DO NOT convert jquery that is used with AJAX, this will be done later.
Also remove code that has been commented out if it is jquery.

PR #17605, opened 2025-05-14
Convert JQuery in \DuggaSys\templates\boxmodell.js - #17516

a23elism:
### This PR is focused on replacing JQuery inside boxmodell.js with normal JS for future development into react


f23linan (approved):
I will approve this for merge, however i think the dugga in general is pretty broken. But its not because of Elidons changes. But I think the functionality of this dugga needs to be looked over as its impossible to do the dugga and if youre logged in as administrator all the text disapear 

image
logged in as admin

image
not logged in

But these changes are okay as its not them thats breaking it. I tested it on our group branch and it was the same thing there.

a23simry:
I tested previous branches and as far back as last year and it seems like the dugga was broken even then.

PR #17606, opened 2025-05-14
G3 2025 v7 17515 convert jquery in bit dugga.js

a22maxda:
Converted all JQuery to JS, have tested the functions I could think of and haven't found any issues.
One thing to note is that saving duggas can only be done by students, not teachers or admins. It is also possible to be logged out.

### DIFF
--- a/DuggaSys/templates/bit-dugga.js
+++ b/DuggaSys/templates/bit-dugga.js
@@ -28,12 +28,14 @@ var response;
 
 function setup()
 {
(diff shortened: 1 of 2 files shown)


a23elism (approved):
After trying everything i could (unless im missing some hidden features), everything works as intended
---

The JQuery also looks correctly converted and functional (since the dugga works it is to be expected), nothing major is changed or added (feature wise) from what i could see in the code (since only conversion was the purpose this is a good thing).

---
Approved for merge, well done 👍

Issue #17111, opened 2025-05-02
Change the microservice retrieveCodeviewerService to use "curlService"


a23louma:
This file in /CodeViewerService has been re-engineered but has to be changed to use the new "curlService" functions. You will need to include the the "curlService" file.

PR #17608, opened 2025-05-14
G2 2025 v7 #17111

a22wilas:
Fixes issue https://github.com/HGustavs/LenaSYS/issues/17111

### DIFF
--- a/DuggaSys/microservices/codeviewerService/deleteCodeExample_ms.php
+++ b/DuggaSys/microservices/codeviewerService/deleteCodeExample_ms.php
@@ -3,7 +3,6 @@ date_default_timezone_set("Europe/Stockholm");
 
 // Include basic application services
 include_once "../sharedMicroservices/getUid_ms.php";
-//include_once "./retrieveCodeviewerService_ms.php";
 include_once "../../../Shared/sessions.php";
 include_once "../../../Shared/basic.php";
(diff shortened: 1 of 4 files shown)


a23danme (commented):
I am not 100% sure how to test it fully, but I tested via UI on diffrent codeviewers and examples, and I tested different stuff such as deleting, editing and saving. Also tested on the original branch which gave same results which it should. Also checked the database and the inspection tool on browser and it seems to be working fine. Only reason I would think this is wrong is if I am testing on the wrong places which as said I could be.

a22wilas:
So I am not entirely sure were this code is applicable on the application. The codeviewer does not seem to be affected from these files/microservices. The solution should still be correct but would be good if it could be tested.

https://github.com/HGustavs/LenaSYS/pull/16906
This is the original microservice solution, my suggestion would be that either the one that worked on this or one of the reviewers test this new solution.

a22linre (commented):
The code looks good to me and seems to follow the re-engineering standard. I tried testing it with a few custom test files, but I couldn't get it to work properly probably because of missing database entries or that I didn’t have the right exampleid or boxid set up. But I don’t think it’s an issue with the code itself.

b23johgu:
Unsure if I can merge this until the testing has worked properly.

Issue #17567, opened 2025-05-14
Reseting the view on an example diagram breaks the app


a23miksk:
Diagram breaks completely when the user attempts to reset view on an example diagram. Only seems to affect the state chart diagram and the UML diagram.

Observed events: 
- The y-ruler gets removed and the x-ruler is rotated to the right (is the y-ruler replacing the x-ruler?)
- All elements gets clumped together and set to the top of the window.
- The user cant move up or down.
- The user cant place elements (or they are placed along the top line of the window where they aren't visible).

https://github.com/user-attachments/assets/dd6b7ada-540c-4e17-a535-93c1b1c0eaf5

PR #17612, opened 2025-05-14
[#17567] Fixed issue where app breaks when centering camera on example diagrams

a22nevan:
**TLDR:** Implemented special handling for certain elements when getting largest y-value since some elements don't have default height values.

The problem was caused by the _centerCamera()_ function in camera.js. It turns out that certain elements (_UML Entities and SD Entities_) don't have a default height value assigned to them, which clashed with how the function got the positional values for centering the camera (_This basically just means that the element objects have no height value to retrieve_).

The function works by getting the smallest and largest values of the entire diagram, which it then uses to determine zoom level and where to position the ruler bars. Since UML- and SD-entity elements don't have heights, the largest y-value became **NaN** (_Apparently, y-pos (which had a value) + element height (which was NaN) = NaN_). This is what caused the all elements to clump up at the top of the screen, and why the y-axis ruler disappeared. Unsure what exactly caused the scrolling to lock as well but it had something to do with this too.

The fix to this was to implement special handling for these two elements. They have their own arrays in globals.js called **UMLHeight** and **SDHeight**, which stores the total height of the elements whenever they are drawn. You can retrieve the height of these elements by searching the array using the ID of the element. The heights are retrieved from those arrays and added together with the element's y-position and put into a separate array. The largest value is then taken from that array instead of from an array created with the _.map()_ function like the other values. Also added some comments to explain the code.

The previously affected example diagrams now behave like the others when centering the camera, not clumping up at the top or locking the scroll function.

https://github.com/user-attachments/assets/71bd9489-edc1-4988-a57f-57e6fc93daeb

Note: The UML-diagram zooms so far out due to the way the system decides what zoom level to be at when centering the camera on a diagram. It just barely **doesn't** fit inside the viewport at 0.5x zoom, so it zooms all the way out to 0.25x instead, which makes it very small. Probably able to be changed and improved on, but it isn't a part of this issue, nor is it that big of a problem realistically.

### DIFF
--- a/DuggaSys/diagram/camera.js
+++ b/DuggaSys/diagram/camera.js
@@ -24,11 +24,31 @@ function centerCamera() {
             y: window.innerHeight / 2
         };
 
-        if(emptyDiagram == true){
+        if (emptyDiagram == true){
+            elementHeights = []; // Special array for height of elements since not all elements have a default height assigned to them
+
+            // Get smallest/largest x-value in the diagram (leftmost point and rightmost point), as well as smallest y-value
             const minX = Math.min.apply(null, data.map(i => i.x));
-            const maxX = Math.max.apply(null, data.map(i => i.x + i.width));
+            const maxX = Math.max.apply(null, data.map(i => i.x + i.width)); // Get rightmost point by taking the largest sum of x-position + element width
             const minY = Math.min.apply(null, data.map(i => i.y));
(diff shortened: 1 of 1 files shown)


a23eliax (approved):
This is a very good and smart solution for this issue! It solves the problem and the programmer didn't have to add so much more code to camera.js. I approve!

https://github.com/user-attachments/assets/13915a51-aab1-4ad2-8d83-d4c368008dba

PR #17613, opened 2025-05-14
Fixed unexpected behavior on ER-attribute text when weakKey

a23emijo:
Fixed the unexpected behavior on ER-attribute text when weakKey is selected. This issue happened because the wrong child were selected and the colorchecks were not correct for the weakKey. I fixed this by removing the hardcoded childrens and instead used the "querySelector("line");". I also changed so the blue color has white text and underline because I found it more nice compared to the black text and line on the blue color.

https://github.com/user-attachments/assets/07936c43-4428-489e-89ab-d7afc34898ad

### DIFF
--- a/DuggaSys/diagram/theme.js
+++ b/DuggaSys/diagram/theme.js
@@ -151,11 +151,12 @@ function updateCSSForAllElements() {
                         fillColor = elementDiv.children[0].children[0]; // Accessing the whatever needs to be accessed
                     }
                     fontColor = elementDiv.children[0];
-                    weakKeyUnderline = elementDiv.children[0].children[2];
+                    weakKeyUnderline = elementDiv.querySelector("line");
                     if (contextLengthCheck()) {
                         fillColor.style.fill = color.LIGHT_PURPLE;
                         fontColor.style.fill = color.WHITE;
-                        // If its a weakKey attribute, update underline color
+                        
+                        // If the weakKey attribute is selected, update underline color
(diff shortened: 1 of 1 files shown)


a23aleva (changes requested):
The fix does indeed make the text of the attribute element remain the correct color, instead of turning to white. Unfortunately, something seems to break the diagram if you try to place a UML Inheritance triangle. I checked our group's main branch and the problem wasn't present there, so it has got to do with something inside these code changes. I checked different elements and this bug only happens with the UML Inheritance (if I didn't miss something). Everything gets stuck to the grid and if you try to pick other elements to place you get nothing.

I don't know why this happens, perhaps something about changing: 
**weakKeyUnderline = elementDiv.children[0].children[2];** 
to 
**weakKeyUnderline = elementDiv.querySelector("line");**

But this is just a guess, mainly because the other code changes are just setting colors but it could also be something else completely. See attached video snippets for better understanding of the problem.

https://github.com/user-attachments/assets/c7e9374f-c0c9-48ca-b3e6-f409e72ad4b9

https://github.com/user-attachments/assets/11bbe19c-b94e-49af-9891-8f8ef2c8f1bf

https://github.com/user-attachments/assets/237a6a8d-75b5-4d86-92f8-9f6e608d47e3

a23aleva (approved):
The newest addition to the fix addresses the previous problem and now the bug appears ironed out!
Changes are approved!

Issue #17304, opened 2025-05-07
Updating the microservice UI design/structure


a23bensh:
The UI design for the microservice will need to be updated again after all functions have been implemented. (This should be on hold until all related sub-issues are complete.)

The edit microservice form should preferably appear as an overlay on top of the viewed microservice, with a relevant design and structure.

The add microservice form should follow a similar style to the edit form but appear larger and be displayed on the main page.

Additionally, the filter dropdown should be redesigned to align visually with the new UI.

All changes should follow the provided mockup for consistency.

PR #17615, opened 2025-05-15
Updating the microservice UI design/structure #17304

a23oskme:
- Made edit microservice form to an overlay
- Made add microservice form to an overlay (a little bit bigger than edit)
- Changed styling for filter button to match the other buttons

<b>This wasn't part of this issue</b>
I also changed the button menu a little bit, so the buttons wont rearrange when the reset buttons appear (if this isn't a wanted change we can remove the last commit and create a new issue) 
It now looks like this: 
Skärmavbild 2025-05-15 kl  09 03 24

And when the reset button appears it now looks like this:
Skärmavbild 2025-05-15 kl  09 03 29

Before these changes when the reset button appeared it looked like this: 
Skärmavbild 2025-05-15 kl  08 27 01

### DIFF
--- a/DuggaSys/microservices/endpointDirectory/microserviceUI.php
+++ b/DuggaSys/microservices/endpointDirectory/microserviceUI.php
@@ -119,19 +119,23 @@ if (isset($_GET['id'])) {
 <body>
 
     <?php if (isset($editMicroservice)) { ?>
-        <div class="line">
-            <h1>Edit Microservice</h1>
+        <div class="edit_modal">
+            <div class="edit_modal_content">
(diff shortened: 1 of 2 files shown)


a23bensh (approved):
This fixes the issue, and even though changing the location of the filtering wasn’t part of the issue, it’s a good addition. Maybe removing the dotted border and adjusting the uneven button sizes to better match the mockup would have been a good addition, but that wasn’t part of the issue. Looks good!

b23johgu:
Nice change of the button placements as well, it appears more put together, will merge this!

Issue #17569, opened 2025-05-14
Cardinality text size isn't scaling properly with zoom


b23danhe:
The cardinality text doesn't zoom properly when zooming in and out.
Make changes so that the text follows the same zoom properties as the other elements.

Edit: This appears to be a Firefox-specific issue, as scaling works correctly in both Safari and Chrome.

<img width="292" alt="Image" src="https://github.com/user-attachments/assets/0b1cce68-6eb7-43d8-b050-ce761c19428f" />

https://github.com/user-attachments/assets/f516393b-ba9a-4959-a4f1-60b166210b66

PR #17621, opened 2025-05-15
Fixed: Cardinality labels do not scale correctly with Firefox

a23emisn:
This patches an issue for Firefox users where zooming didn't correctly scale the cardinality text labels. See the attached video in the issue for a problem demonstration (#17569).

The issue was that the cardinality text was styled using a CSS font-size property, and apparently Firefox does not consistently apply this styling to SVG <text> elements, causing the text to remain the same size regardless of zoom level. This was changed to dedicated font-size and fill SVG attributes being used for the cardinality texts. 

**This change should be tested specifically in Firefox to confirm the fix.**

### DIFF
--- a/DuggaSys/diagram/draw/line.js
+++ b/DuggaSys/diagram/draw/line.js
@@ -814,10 +814,11 @@ function drawLineCardinality(line, lineColor, fx, fy, tx, ty, f, t) {
                 height='${(textheight - 4) * zoomfact + zoomfact * 3}'
             /> 
             <text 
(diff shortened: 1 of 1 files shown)


b23ebbst (approved):
**Tested specifically in firefox.**
Works really well and zooming does not weirdly scale the cardinality-labels any more. 

Really well done, I approve this change! 

https://github.com/user-attachments/assets/0ecc5d00-e908-4d07-92b8-d8e5465e3cf8

Issue #17486, opened 2025-05-13
Examine the tooltips for refactoring


a23miksk:
The tooltips can be refactored for the diagram to make it more readable and easy to use. To do this the code needs to be researched so that the rest of the site isnt affected.

PR #17622, opened 2025-05-15
Fixed: Examine the tooltips for refactoring

a23gabst1:
My solution consists of having only one element used to contain information about the elements and act as a tooltip. When the user hovers on an element the tooltip's position is calculated based on where the hovered element are. For example the tooltips associated with the options panel gets a different position in contrast to tooltips on the toolbar. The tooltips information gets filled based on the hovered element. if the pointer element is being hovered then the tooltip gets filled with information about the pointer from the file that contains all information about the elements (tooltip_information.js). 

With this i have removed every old tooltip that existed for every element, this reduces the code especially for the toolbar. 

I think i have found a great refactoring solution to the tooltips #17486

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -1204,6 +1204,9 @@
     background-color: #060606cb;
     z-index: 5000;
     text-align: center;
-    line-height: 1.6;
-    margin-block: .5rem;
+    line-height: 1.5;
+
+    >*{
+        margin: 0;
+    }
 }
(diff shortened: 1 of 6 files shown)


a23eliax (approved):
This is a smart and clean refactoring! Centralizing the tooltip logic into a single element makes the codebase much easier to maintain and reduces redundancy, especially in the toolbar.
I really liked what you did with the new tooltip_information.js file and how your solution reduces some code in diagram.php. Great job, I approve!

PR #17625, opened 2025-05-15
Documented all files in showDuggaService-map

a23wilgr:
Documented all files in showDuggaService-map:
- getShowDugga_ms
- loadDugga_ms
- processDuggaFile_ms
- retrieveShowDuggaService_ms
- saveDugga_ms
- updateActiveUsers_ms

Documented them in a different order than in the actual microservice map, since one file relies on another etc. Not certain of AUtoken parameters, the different variant parameters and the contactable parameter.


a22linre (commented):
This follows the template and looks good to me!

PR #17626, opened 2025-05-15
Replaced jQuery with JS in kryss.js

a23yutta:
Replaced jQuery with JS in `/DuggaSys/templates/kryss.js`. 

I have not been able to verify whether the JavaScript code works, the quiz (pen icon) button for the dugga does not work in both the week branch and the issue sub-branch(that's if you're logged in). Directly accessing the `kryss.html` file located in `/DuggaSys/templates/kryss.html` shows a non-interactable page with quiz instructions. Quiz instructions and the actual quiz displays while logged out but what is the actual purpose of this? It does not do anything when selecting an option. 

Comment on `kryss.js `file:
The `kryss.js` file seems like an experimental file, unfinished and casual.

### DIFF
--- a/DuggaSys/templates/kryss.js
+++ b/DuggaSys/templates/kryss.js
@@ -59,32 +59,27 @@ function quiz(parameters) {
 			}
 		}
 		
-		$("#output").html(app);
-	}
-	else {
+		document.getElementById("output").innerHTML = app;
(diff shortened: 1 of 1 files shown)


b20isato (approved):
Seems good. Does not break anything as far as i can see

Issue #17553, opened 2025-05-14
Examine options.js


a23miksk:
Examine and possible fix the semantic issues described https://github.com/HGustavs/LenaSYS/pull/17550, where a couple functions maybe should be renamed or removed. If the functions really should be renamed and it can be done cleanly, do that. Otherwise, examine the pros and cons of renaming the functions and write a comment about it here.

PR #17627, opened 2025-05-15
Fixed G1-2025-v7-#17553-examine-options.js

a23eliax:
As requested, I changed some functionnames, like "radio()" "select()" and "option()". I did not change the name of the function "lineLabel" becuase it had preferences to other files and my opinion is that it's unnecessary to do all that work for that function.

"setLineLabel()" sets the lines label to "< < include > >" via the include-button on the picture below so I left that function as it is.

image

### DIFF
--- a/DuggaSys/diagram/draw/options.js
+++ b/DuggaSys/diagram/draw/options.js
@@ -282,7 +282,7 @@ function drawElementProperties(element) {
  * @param {Object} object What types of value the element have.
  * @return Returns a dropdown menu.
  */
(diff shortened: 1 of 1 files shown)


a23aleva (approved):
The mentioned function names have been successfully changed for the better and are no longer unintelligible and uninformative. Also checked that they are not being called by their old names in other files and that seems fine as far as I can tell. Furthermore, I stand behind the decision to leave setLineLabel() as it is. These changes are all approved!

PR #17628, opened 2025-05-15
Repaired duggas and reengineered loadDugga and retrieveShowDuggaService

a23axehe:
I reset all the files to their state in week 3 before re-engineering of them started to get some clarity on what was wrong and then i Re engineered loadDugga using my previous changes in PR #17157 I also reengineered retrieveShowDuggaService from scratch and decided to not use curlservice for any re engineering purposes because of the possibility of reusing previous code. I have tested by loading dugga with hash **AcIj-Ld6** and password **OjOEh8R** from the database and then playing around with it trying to figure out how to solve the dugga, tried resetting and saving which both worked. @a23danme contributed to this solution via his  PR #17372 . Something that is bothering me is that in my original PR of #17157 when accessing the 3D dugga in demo course a triangle spinning in a black box appeared next to the ui, which seemed like a dugga function? However im unsure of this and it did not appear when accessing the group branch for week 3. I guess it might have something to do with that dugga not being loaded using hash and password? However the solution fixes issue #17264 .

### DIFF
--- a/DuggaSys/microservices/showDuggaService/getShowDugga_ms.php
+++ b/DuggaSys/microservices/showDuggaService/getShowDugga_ms.php
@@ -99,7 +99,52 @@ if($courseid != "UNK" && $coursevers != "UNK" && $duggaid != "UNK" && $moment !=
 	$debug="Could not find the requested dugga!";
 }
 
-echo json_encode(
+
+header("Content-Type: application/json");
+//set url for setAsActiveCourse.php path
+$baseURL = "https://" . $_SERVER['HTTP_HOST'];
+$url = $baseURL . "/LenaSYS/DuggaSys/microservices/sharedMicroservices/setAsActiveCourse_ms.php";
+$ch = curl_init($url);
+    //options for curl
+curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
+curl_setopt($ch, CURLOPT_POST, true);
+curl_setopt($ch, CURLOPT_POSTFIELDS, http_build_query([
+    'moment' => $moment, 
+    'courseid' => $courseid, 
+    'hash' => $hash, 
+    'hashpwd' => $hashpwd, 
+    'coursevers' => $coursevers,
+    'duggaid' =>  $duggaid,
+    'opt' =>  $opt,
+    'group' =>  $group,
+    'score' => $score,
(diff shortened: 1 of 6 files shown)


a23axehe:
Realised i better remove comments of old code, will do that now quickly

a23danme (commented):
Just testing around for 20 min gave me some questionmarks.
First of all, in the original, not with these changes, the saving, loading and useage of duggor is working but only WHEN NOT LOGGED IN. When you are logged in barely anything works, saving doesnt work, error comes "unable to save dugga" and loading does only reset it and some duggor, an for example dugga 4, does not even show it self when being logged in, but it does when not logged in.

In this 17264, saving and loading does not work for me even when not logged in, it just resets the page. And for example the dugga 4 which shows itself on the original when NOT logged in, does not show either when you are logged in or not now. According to your description it feels like you have tested yourself and it is working? For me it does not, if it does please say so and tell me if im doing something wrong! I tried on my laptop aswell and got the same results. 

In my opinion, as the problem as I had myself sitting with these files, there are something about the "superuser" or "if logged in" or these stuff that is the problem. Because as I said in the original files, the duggas are working when not  logged in but when you get logged in it does not work. And now for this 17264 it feels like it is as "when you are logged in in the original" with other words, it stops working.

This is a very  fast response from checking it out, due to I started to check it out even before I got assigned to review due to me being intrested in the files due to me sitting in them for so long before. So I am going to check more, and look more at the code and details to see if I get a better grip on the situation

a23axehe:
When logged in and saving: 
image

When not logged in and saving:

image

When not logged in and loading dugga 3 with hash AcIj-Ld6 and password OjOEh8R it works

When logged in and loading a dugga from db it works

Tried saving dugga 3 when not logged in, correct hash and pwd appeared in db userAnswer with correct timestamp and everything. Then loaded in to that dugga with hash and pwd and it worked.

When logged in it does not work, I dont know if it should, i refer to line 112 in saveDugga_ms.php where the comment says teachers/superusers cannot submit a dugga and if you read the code you see there is no way for the data to be inserted into db.

a23danme (commented):
I cant find or see what these newest commits changed! The displaying of loaded duggor and for example dugga 4 not showing graphically is still the same

PR #17631, opened 2025-05-15
added media query check for fab button margin

a23kavis:
**Rebased PR**
Previously the fab button had an extensive right margin causing it to almost become centered on smaller width resolutions. This has been adjusted for smaller devices to have a more subtle margin.

Previously
image

Now
image

This solution however may not be the best, may be worth looking into making the fab movable by the user, as it may be in the way depending on what page you are on.

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -8000,6 +8000,7 @@ only screen and (max-device-width: 568px) {
     transform: rotate(90deg);
     transform-origin: top left;
   }
+  
 
   .fab-btn-list2 li {
     transform: rotate(225deg);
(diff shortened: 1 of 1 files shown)


a23kavis:
> This is not a media query. Please make it into a media query. I am pretty sure there already are existing media queries somewhere in the css-file, find them, and implement the fix there. example of how it should look something like:
> 
> ```
> @media (max-width: <what is already set in existing media queries>) {
>   .fabButton {
>     right: 20px;
>   }
> }
> ```

Not sure if i am missing something but I'm quite sure it is in a media query? 
image

a23fanro:
@a23kavis OMG I am so sorry, forgot to check the code over it! 
Very well done, thank you!

Issue #17561, opened 2025-05-14
IE Relation gets cut of at max zoom


a23miksk:
As the title says, the IE Relation element gets cut of at max zoom

_Normal zoom_
![Image](https://github.com/user-attachments/assets/28a3cb7f-73dd-42ac-9a31-16bb710d76d3)

_Max zoom where the element gets corrupted_
![Image](https://github.com/user-attachments/assets/e12347b6-a522-4f9d-bf28-da58a6767d84)

PR #17634, opened 2025-05-15
[#17561] Fixed IE relation getting cut of at max zoom

a23emijo:
Fixed the issue where IE relation got cut of at max zoom. For some reason, only the IE relation had this problem while the other relations didn't.

_**Max zoom**_
bild


a23eliax (approved):
A very smart and simple solution for an important issue! It works sas expected, I approve!

https://github.com/user-attachments/assets/07881965-d0e1-4a7f-a1ad-c1b6e834cc76

PR #17635, opened 2025-05-15
Fixed the too light hover

f23linan:
Added a courseRow hover in blackTheme as there was none and therefor used the lightmode hover. It now looks much better and readable when youre hovering over a listitem in sectioned.php.

image
Hover now

image
Hover before


a23alera (approved):
Fix works as intended, much better hover visibility!

Issue #17501, opened 2025-05-13
Remove announcement


a23fanro:
As of PR https://github.com/HGustavs/LenaSYS/issues/12401 announcement are no longer used in LenaSys, as they are handled by canvas. 
They forgot to remove announcement from all views.

We need someone to remove the button and related code for announcement on mobile view.

PR #17637, opened 2025-05-15
Removed the button for announcements on mobile view for issue #17501

a22carsv:
### Description
For mobile view, this PR removes the button for viewing announcements in the burger menu. Although I am unsure if the page displaying announcements is still accessible, it is probably outside the scope of this issue. This is what it looks like before/after changes:
### Before:
image

## After:
image

### Testing
For testing, simply navigate into any course, CTRL + SHIFT + I and enter mobile view. The previous "Announcements" button should not be visible anymore within the burger menu (three horizontal stripes) and presumably no related code should be remaining within the codebase.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -3051,19 +3051,6 @@ window.addEventListener("DOMContentLoaded", function () {
 
   });
   document.addEventListener("click", function (e) {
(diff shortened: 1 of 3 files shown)


EdvinWerner (approved):
The button for viewing announcments in the mobile view is no longer visible when in mobile view, approved for merge.

PR #17638, opened 2025-05-15
Fixed "Send to Back" elements placed behind the canvas

a22abdjo:
Elements should now not dissappear behind the canvas no matter how many times "Send to Back" button is pressed. This solution has been implemented by setting a cap to the minimum "z" in the "sendtoBack" function. These changes were made in options.js file.

### DIFF
--- a/DuggaSys/diagram/draw/options.js
+++ b/DuggaSys/diagram/draw/options.js
@@ -1949,11 +1949,18 @@ function sendToBack(id) {
     const elem = data.find(e => e.id === id);
     if (!elem) return;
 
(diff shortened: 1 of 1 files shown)


a22abdjo:
https://github.com/user-attachments/assets/3b66b501-8be4-45a6-b3bf-31e7e44dfe41

a23gabst1 (approved):
The solution works as intended! I have tested all of the elements and none of them was placed behind the canvas when pressed "send to back". I approve these changes.

The code looks great but would really love to see more comments that explains how it works.

Issue #17537, opened 2025-05-13
convert jquery in \DuggaSys\templates\feedback_dugga.js


a23fanro:
Locate jquery in the file and convert it to javascript. 
DO NOT convert jquery that is used with AJAX, this will be done later.
Also remove code that has been commented out if it is jquery.

a23fanro:
Please remove the function 'freturneddugga' as mentioned in the [comments](https://github.com/HGustavs/LenaSYS/pull/17640#issuecomment-2890723641), then I will merge!

PR #17640, opened 2025-05-15
Convert jquery #17537

f23linan:
Converted Jquery in in \DuggaSys\templates\feedback_dugga.js to JavaScript.

I had some trouble testing if anything broke during the convertion, but as far as I can tell it should be fine.


a23osckr (approved):
Not sure if I tested this correctly, but I went to the demo-course feedback dugga and tested there. As broken as it already was I could not find anything additional that had broken and no extra errors. Only thing I noticed in the code is an extra funtion "freturnedDugga", seems to be the old "returnedDugga"? Was this just forgotten or done on purpose? However, since everthing seems to "work" I am approving this.

f23linan:
The freturneddugga is the old one i forgot to remove!

f23linan:
I have removed the old function that was left in it now!

PR #17642, opened 2025-05-15
G3 2025 v5#17222 deleting dugga then selecting all shows the deleted dugga for a few seconds

c22isako:
**Please note;**
This PR does not fix the current problem where deleting the dugga, and selecting all, displays the deleted dugga. This is because of an issue within another group which affects reload, and therefore the functionality of created/deleted elements. We are aware that this problem exists, and that it is expected to function when the reload-issue is fixed. The same issue is currently present on the weekly-branch. _- Groupleader has been notified._

**That said, this PR introduces another functionality which lets the user see the deleted items whiles the undo-timer is running. This gives the user the opportunity to spot items which weren't meant to be destroyed, and enhances the feeling of "control".**
- The items are removed after the timer has run, but can be visually "brought back" via select all - caused by reload-issue.
---
**How to get to the functionality:**
select course of your choice (which has duggor, ex. Testing Course) -> log in 
from here, you have access to both delete (trashcans) and select (whereas select-all is marked in red in image below)
image

When pressing delete, and confirm, you should be met with this a greyed-out version of the duggas:
image
When 15 seconds (16 seconds) has passed, the items should disappear from view.


a23simry:
As this can at the moment not be properly tested, if you are reviewing this, please at least make sure the new functionality works and that the code looks okay/doesn't break any other functionality.

Issue #17639, opened 2025-05-15
Diagram-toolbar is not completely visible in showDugga


b23danhe:
The "Load" button is partially hidden in the diagram toolbar when in showDugga mode.

Issue: Adjust the layout so the entire diagram toolbar is fully visible in showDugga.

Also check so that the changes only affects the diagram dugga.

![Image](https://github.com/user-attachments/assets/a1b87bb0-75d0-4c72-b77b-668e4d31ec12)

PR #17646, opened 2025-05-15
Fixed: Diagram-toolbar is not completely visible in showDugga

a23gabst1:
The solution was to replace the containers height from 50.1rem to 100dvh and then have other elements inside it inherit that height. And the ifram inherits from .canvas_right and needed to explicit set the height of #close_open_border to 100dvh.  see figure 1. 

I fixed the issue #17639

figure 1 
image

### DIFF
--- a/DuggaSys/templates/diagram_dugga.html
+++ b/DuggaSys/templates/diagram_dugga.html
@@ -7,7 +7,7 @@
     <style>
         .container {
             display: flex;
-            height: 50.1rem;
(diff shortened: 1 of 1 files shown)


a23emijo (approved):
When I first tested, my resolution was 2560x1440 and the toolbar got to wide. I then tested on my second monitor which has 1920x1080 and it works perfectly. This was tested in Chrome and Edge! 👍 

thumbs-up-approval

Issue #17644, opened 2025-05-15
create SSL-generator script


a23arvme:
A script that generates localhost.key and localhost.cert and places them in a convenient location. It is to be integrated with the installer.

PR #17647, opened 2025-05-15
SSL-generator #17644

a23arvme:
This is a SSL cert generator that generates localhost.key and localhost.crt.
These files will later be used for a tutorial that explains how to enable HTTPS on your local machine.
The .gitignore needs to be updated with _*.crt_ and _*.key_ since these files are supposed to be local, and not be pushed to remote.


a23louma (approved):
I approve! Code looks good and after running the script these files appear as expected:
image

Issue #17579, opened 2025-05-14
The UML super state crashes the canvas and the toolbar


a23emijo:
When an UML super state is placed, it crashes the canvas, the toolbar can't be clicked anymore, only hovered and the "Current file:" text doesn't update, probably more things too.
It is possible to save after the crash and load the save but the savefile is still destroyed (not shown in the video, it was found after).

https://github.com/user-attachments/assets/49d4e1db-568e-484d-b1e7-2775d99587df

PR #17650, opened 2025-05-15
[#17579] Fixed issue with app freezing after placing an UML Super State object

a22nevan:
A recent change seems to have changed the way the _updateCSSForAllElements()_ function looked for elements in an entity that were supposed to be affected when changing the fill color of it. I assume this was related to the change where all selected objects now get a purple fill color. The sequence loop object and UML super state entity had the affected elements retrieved by finding the element with the ID "_loopLabel_". This ID had however not been added to the element in the UML super state object, so when the system tried to draw the super state entity the element to fill returned as null, which froze the app since the function for updating the CSS stopped working. 

The ID has now been added to the correct element in the super state object, in addition to changing the ID name from "_loopLabel_" to "_cornerLabel_" to better reflect both objects that use labels instead of just the loop object. The app should no longer freeze when placing out a super state object.

 
https://github.com/user-attachments/assets/524bf8bd-08f4-41e2-8a2e-f4e5abc467fe


a22abdjo (approved):
The solution provided works as intended and therefore is approved.

PR #17651, opened 2025-05-15
Document retrieveAllCourseVersions_ms.php and retrieveSectionedServic…

a23petny:
…e_ms.php


a22wilas (changes requested):
The documentation for the microservices looks good. I would prefer a minor change when it comes to the output for the service retrieveSectionedService, change the order for in the documentation to follow the same order as the output array. Easier to follow the documentation in that way. 

A much bigger change is that this implementation removes all other documentation in the file sectionedService.md. This needs to be changed before I can approve this PR.

Look at the added and removed code.
bild
****

Issue #17575, opened 2025-05-14
Impossible to change varient after it has been changed once


a23emijo:
The varient can't be changed after it has been changed once. This is the case for every element that can change varient not just the "ER Attribute" that is shown in the video.

EDIT: This also happens when the color is changed before so I think that issue #17511 is the same issue as this one.

https://github.com/user-attachments/assets/ec72eb92-e895-494f-9676-edbee6a83365

a23emijo:
> I'm pretty sure that this issue stems from the same problem as [#17511](https://github.com/HGustavs/LenaSYS/issues/17511) and can be closed with the same pull request.

Both issues should be tested in the PR for one of these issues

Issue #17511, opened 2025-05-13
Problem with variants of elements


a23miksk:
It is now impossible to change the variants of elements in the diagram after changing the name of the element. Changing the variant first, saving, and then changing the name works.
Known elements with the issue: ER entity, ER relation and ER attribute.

_Video where the first element is "locked" and is unable to change variant_
https://github.com/user-attachments/assets/c50eed2e-5954-4a8e-8545-7933fcd58e63

a23emijo:
> It is now impossible to change the variants of elements in the diagram after changing the name of the element. Changing the variant first, saving, and then changing the name works. Known elements with the issue: ER entity, ER relation and ER attribute.
> 
> _Video where the first element is "locked" and is unable to change variant_ https://github.com/user-attachments/assets/c50eed2e-5954-4a8e-8545-7933fcd58e63

Take a look at issue #17575 also

PR #17652, opened 2025-05-15
Problem with variants of elements (#17511 and #17575)

b23ebbst:
The changeState()-function has been fixed so that it can handle a name-change or color-change before changing the "variant" (state) of an element (see issues #17511 and #17575). 
Now it handles any change to the selected elements state, and then redraws the canvas to display the change, making it possible to change name or color and then change the state.

This should affect any elements/entities that utilize the "Variant"-option in the options-panel.

https://github.com/user-attachments/assets/34c2083c-e8a3-4ea4-b7e9-87a39113dc2d


a23emijo (approved):
The solution provided works perfectly. I approve this! 💯

https://github.com/user-attachments/assets/1a19547f-03da-4ce4-8aa9-37a663cdfdd3

PR #17654, opened 2025-05-15
Convert JQuery in \DuggaSys\templates\dugga5.js - #17534

a23elism:
### PR is focused on converting JQuery inside dugga5.js (if existent)

last instance of JQuery is to be left alone, too big for this issue, since it requires creating its own complete function + possibly touching CSS


a23yutta:
@a23elism Line 830 (end of file) has still not been converted to JS yet. I will review the PR when it has been converted.

Issue #17185, opened 2025-05-05
Undo and redo includes individual letters after having renamed element


a22nevan:
Unsure when this was implemented but renaming elements now show the letters being added to the element in 
real-time, which is a very neat quality of life addition. This has however messed with the way renaming is being handled in the state machine. Every single individual letter being added counts as its own separate history state in the state machine, causing the undo and redo functionality to have to cycle though each letter in a name when undoing a name change. This doesn't happen if you paste in an entire word at once, but more than often users are going to write the names by themselves, which'll cause them to encounter this issue if they need to undo/redo anything.

Ideally it should undo/redo the entire name change, instead of each individual letter at a time.

https://github.com/user-attachments/assets/c428cd39-be58-4dd4-a4a6-fbbcc12dc96e

b23danhe:
This issue was reopened since the problem is still present.

PR #17655, opened 2025-05-15
Fix undo/redo name issue #17185

a23matsh:
Changed undo/redo behavior from character-by-character updates which felt unnatural, to using a blur event. Now when the user finishes input (e.g. clicks save or clicks away on the grid), pressing undo will remove the entire input (e.g. full name) instead of one character at a time. Redo behaves the same.

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1216,7 +1216,7 @@ function mmoving(event) {
             }
 
             // Store the changes in the history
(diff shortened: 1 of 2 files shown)


b23ebbst (approved):
Works as expected! 🌟 
If pressing save or enter, the change is saved and can be undone/redone in one go rather than one letter at a time. I approve!

PR #17657, opened 2025-05-16
Changed to curlService in retrieveResultedService_ms #17416

### LINKED ISSUE #17416: Change the microservice retrieveResultedService_ms to use "curlService"
This file in /resultedService has been re-engineered but has to be changed to use the new "curlService" functions. You will need to include the the "curlService" file. You need to go through both the file you get in the issue and the inverse dependency file resultedService\getUserAnswer_ms.php.

a23louma:
Fixes #17416 . I've changed to curlService in retrieveResultedService_ms and its inverse dependency microservice getUserAnswer_ms. At the moment these files are not used and it's the old resultedservice.php being used when you go to "Edit student results". Also, retrieveResultedService_ms just sends the same data back to getUserAnswer_ms as it received. To test that the communication works between these two files I created a testfile to send real values from the database for "cid" and "coursevers" to getUserAnswer_ms. I had to comment out the "isSuperUser().." line of code and change "(kind=3)"  to "(kind=2)" in the query in getUserAnswer_ms, then I was able to get a response with data of the two arrays.

### DIFF
--- a/DuggaSys/microservices/resultedService/getUserAnswer_ms.php
+++ b/DuggaSys/microservices/resultedService/getUserAnswer_ms.php
@@ -72,34 +72,14 @@ if(isSuperUser($userid) || hasAccess($userid, $cid, 'w')){
 
     	array_push($tableInfo, $tableSubmissionInfo);
 	}
-	/*
-    // Set up POST call to retrieveResultedService_ms.php
-    header("Content-Type: application/json");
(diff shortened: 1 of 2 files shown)


a23bensh (approved):
This fixes the issue, curlService is now being used and the inverse file validates input via recieveMicroservicePOST, looks good!

PR #17666, opened 2025-05-16
Replaced jQuery with JS in exempeldugga1-7.js files

a23yutta:
The jQuery has been replaced with JS. I have not been able to verify whether the JS code works as intended. All exampledugga1.js-exampledugga7.js files are the exact copy of each other. I have no idea what purpose the files serves and where they are used(or if they are actually used in the first place).


a23fanro:
Since these duggs do not appear to be used anywhere in the system and there is no clear reason for their existence, I recommend that another developer test the system with their complete removal. 
This should involve thorough testing: specifically, checking each available course to ensure that no duggss disappear or break as a result of the files removal.

If the system functions correctly and no issues are found, I suggest we permanently remove these files.

PR #17668, opened 2025-05-16
Convert jquery in transforms-dugga.js #17520

a23erigu:
I have converted the Jquery that was used in transforms-dugga.js to javascript.

The easiest way to test this is in Transforms Dugga in Demo-Course.


a23osckr (approved):
Previous issues have been fixed and as far as I can tell nothing else has been broken. Approved.

PR #17675, opened 2025-05-16
Implemented sidebar for mobile

a23carpa:
Followed the mobile mock-up and made the the sidebar similar to the bottom menu.
Note: I was not able to implement the logout button, this will need a sub issue in the future.
The other buttons should work as their desktop versions. 
Second note: Save, load and reset dugga will not work in only the diagram.php file.
They need showdugga for specific functions.

[Screencast from 2025-05-16 13-18-10.webm](https://github.com/user-attachments/assets/6bd75813-17b9-4da9-90ab-71b39ed9e45a)

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -338,7 +338,7 @@
         transition: var(--normal-transition);
     }
 
-     .icon-wrapper-2{
+     .icon-wrapper-sidebar{
         display: block !important;
         position: fixed;
         z-index: 10;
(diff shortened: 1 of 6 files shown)


a23matsh (approved):
The sidebar looks fine in regards to the mock up and matches the bottom menu. The buttons work as intended except for the already mentioned save load and reset who needs the specifications for showdugga. Login works but logout needs to be implemented in a future issue since it as of now has no functionality as mentioned.

PR #17685, opened 2025-05-16
Converted jquery in the group-assignment.js

f23linan:
Converted all jquery(except AJAX) to javascript in the group-assignment.js.

Quick note:
The instructions for this dugga in only available in admin view and being able to send in the files is availble in all the other access except for admin, im not sure if this is on purpose or not but its noted that it was like this before the convertion.


c22isako (changes requested):
**Some small changes are needed:**
There seems to be some conversion within ajax, which will need to be converted back to its original state (line 171). This is since there will be an separate issue to change the ajax, and was not to be done within this issue.

**Otherwise, looks good! Can not find any differences within functionality between the group-branch and the PR.** 
That said, I couldn't open it without being admin, since the other account types has no access to the specific course which houses the dugga. _As I couldn't access "load dugga" without being admin, cannot confirm issue with instructions not visible as other users._
- From the admins (and users) point of view, it seems as functional as the group branch - given that I have no applicable dugga to load & test it on.

f23linan:
Fixed the change! good to go now :)

b23sebth (approved):
**Fixes #17539 - Approved!**

This branch and the `G3-2025-V7` branch behaves the same and I was not able to find any issues that are not already present on the group branch.

## What's changed:
- jQuery has been converted to vanilla JS.
- Previous AJAX removal has been reverted.

Issue #17648, opened 2025-05-15
Add a fallback to http in curlService.php


a23arvme:
It is possible to update curlService to check if the server has a SSL-certificate (allowing for https) and fallback to http in the case it does not. I think this would be nice so the page doesn't break down if you do not have SSL, and still allow for it if it does.

PR #17686, opened 2025-05-16
Added fallback to http in curlService.php

a23gabli:
solves issue #17648


a23dilel:
Your code looks correct, but how to test your code is working or not?

PR #17687, opened 2025-05-16
curlService for processDuggaFile

a22wilas:
Fixes issue https://github.com/HGustavs/LenaSYS/issues/16978

To test this try to Load a dugga with and without the curlService include in the retieveShowDuggaService file.
This is the only place I could see that this code effected.

bild

### DIFF
--- a/DuggaSys/microservices/showDuggaService/processDuggaFile_ms.php
+++ b/DuggaSys/microservices/showDuggaService/processDuggaFile_ms.php
@@ -9,7 +9,7 @@ include_once "../../../Shared/basic.php";
 include_once "../../../Shared/sessions.php";
 
 // Connect to database
-pdoConnect(); 
+pdoConnect();
 
 $courseid;
 $coursevers;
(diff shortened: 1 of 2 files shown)


a23dilel (changes requested):
Your code is working and using the process dugga file and I received data which you can see in the Docker terminal, that I’m using:
image

However, in processDuggaFile_ms.php, you are using this code:
```
// Handle incoming POST request
if ($_SERVER['REQUEST_METHOD'] === 'POST') {
	if (isset($_POST['courseid'], $_POST['coursevers'])) {
		$courseid = $_POST['courseid'];
		$coursevers = $_POST['coursevers'];
		$duggaid = $_POST['duggaid'];
		$duggainfo = $_POST['duggainfo'];
		$moment = $_POST['moment'];
	}
}
```
There is already a function that you can use called `recieveMicroservicePOST()` from `curlservice.php`. This function is the same thing your code. To implement this function looks something like this:
```php
include_once "../curlService.php";
...
$data = recieveMicroservicePOST(['courseid', ...]);
$courseid = $data['courseid'];
...
```
If you use this `recieveMicroservicePOST()` function, then I will approve.

Overall, looks good!

a22wilas:
Have fixed the requested changes. Should now use the function recieveMicroservicePOST() to verify the received data.

a23dilel (approved):
Works as intended to receive all data, and the code is working well. Good job!
image

Issue #17508, opened 2025-05-13
Delete-button is too far out when the filename is too long


a23emijo:
The delete-button is too far out to the right when the filename is too long.

![Image](https://github.com/user-attachments/assets/18808e20-0aff-4650-abfa-8ad25727cff1)

PR #17690, opened 2025-05-16
[#17508] Fixed wide load button issue on in mobile view

a22nevan:
The load button didn't have special formatting to ensure that load buttons fit within the load modal window. This caused the button to extend outside of the window, bringing the delete button with it. The button was still accessible if you scrolled horizontally, but ideally it should just fit into the available window. 

A condition has been added so that the text in the buttons receive special formatting if the browser is in mobile view, limiting the displayed text to 12 characters of the file name. This, admittedly, isn't the most elegant solution to the problem, but it does its job.

A global variable called _isMobile_ has also been added for future mobile specific formatting.

**After fix - mobile view:** 
image

**After fix - pc view:**
image

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -2125,7 +2125,17 @@ function showModal() {
         for (let i = 0; i < diagramKeys.length; i++) {
             let wrapper = document.createElement('div');
             const btn = document.createElement('button');
-            const btnText = document.createTextNode(diagramKeys[i]);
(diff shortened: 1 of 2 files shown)


a23eliax (approved):
This is a very good solution on this issue with the ellipsis when the name of the file is too long. The delete buttons are visible and the user doesn't have to scroll to the side to push it. Works as expected!

image

Issue #17577, opened 2025-05-14
"Bring to Front" and "Send to Back" doesn't work on "Sequence lifline (actor)"


a23emijo:
The "Bring to Front" and "Send to Back" feature doesn't work on "Sequence lifline (actor)" as shown in the video.

https://github.com/user-attachments/assets/ef1ac057-2e3e-4b8f-9020-f0704f4e6f55

a23miksk:
A bit unclear job statement IMO. Specifically which elements has this issue? Has the developer fixed the issue when every element in the diagram is fixed?

a23emijo:
> A bit unclear job statement IMO. Specifically which elements has this issue? Has the developer fixed the issue when every element in the diagram is fixed?

I can agree and I'm sorry for that. I didn't try every element, just found out after I published the issue that a lot of elements can't be moved to the front or to the back. I think the developer should focus on element, then try the rest to check that every element works fine 😄.

PR #17691, opened 2025-05-16
All elements now work with bring/send to front/back

a23emisn:
It was discovered that certain elements did not respond to the bring/send to front/back operations via the buttons in the options panel. This was due to static zLevel values having been attributed to these elements, which would override any attempted z-ordering changes. By removing these static zLevel values, the elements now respond as expected.

The specific elements that were affected are: Sequence Actor, Sequence Object, Sequence Activation, Sequence Loop, UML Superstate, UML Initial State, and UML Final State. These have been modified to respond correctly in this patch.

An alternative approach could be to remove the relevant option buttons for these elements **IF** they are intended to remain locked at certain z-levels. One inherent issue with the implemented solution is that Sequence Activations can now get unintentionally buried under Sequence Lifelines, among other flaws. However, this patch resolves the immediate issue and can be merged if deemed necessary.

Check the attached video for issue demo: #17577

### DIFF
--- a/DuggaSys/diagram/draw/element.js
+++ b/DuggaSys/diagram/draw/element.js
@@ -112,7 +112,6 @@ function drawElement(element, ghosted = false) {
         case elementTypesNames.sequenceActor:
             divContent = drawElementSequenceActor(element, textWidth, boxw, boxh, linew, texth);
             mouseEnter = 'mouseEnterSeq(event);';
-            zLevel = 2;
             break;
         case elementTypesNames.sequenceObject:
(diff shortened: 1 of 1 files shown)


a22nevan (approved):
The mentioned elements now are now all able to have their z-index modified with the bring to front/send to back buttons. As mentioned in the PR, sequence activation elements can get buried below lifelines, among some other elements. This is kind of awkward to experience during usage, but is not a major enough issue that it needs be addressed in this solution. 

The changes are good and work as described. They can be merged without any problems. 👍

PR #17692, opened 2025-05-16
17565 remove or extend the stereotype length

d22abdha:
Limit stereotype input to 50 characters and Stereotype text will now automatically wrap to multiple lines and grow the UML header’s heigh

Before:
before
After:
after

### DIFF
--- a/DuggaSys/diagram/draw/element.js
+++ b/DuggaSys/diagram/draw/element.js
@@ -360,32 +360,43 @@ function drawElementUMLEntity(element, boxw, boxh, linew, texth) {
 
     // Header
     const headerLines = splitFull([element.name], maxCharactersPerLine);
-    let height = texth * (headerLines.length + 1) * lineHeight;
+
+if (element.stereotype != "" && element.stereotype != null) {
+    // Split on newlines, wrap each, flatten
+    const raw = String(element.stereotype).split(/\r?\n/);
+    let stLines = [];
+    raw.forEach(l => stLines = stLines.concat(splitFull([l], maxCharactersPerLine)));
+
+    // total lines = 1 padding + stereotype lines + name lines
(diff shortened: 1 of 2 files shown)


d22abdha:
Limited stereotype input to 50 characters to prevent SVG overflow, 99% of real world UML stereotypes fit well under this. This prevents users from entering chunk of text that would break the diagram’s SVG layout.

a23emisn (changes requested):
The implemented solution does not meet the criteria of the issue concerned, and needs to be improved before merging. It does set a higher limit for amount of characters allowed, but it is mentioned in the issue that the stereotype shouldn't cause any line breaks, cut-offs etc. The behavior now is that is looks like multiple stereotypes are created, which isn't the desired behavior: 
image

 Also, adding at least one character to the stereotype makes an empty row appear below it for no reason:
image

Please adjust these flaws so that the stereotype sits fixed at the top row and that the class shape widens dynamically if it needs to.

a23emisn (approved):
The solution has improved with the latest commit and the behavior is more in line with what was requested in the issue. The extra empty row issue is fixed. 50 characters are allowed, the text doesn't overflow/cutoff. The text dynamically follows the resizing of the class it sits in etc. The line will however break with narrow class entities and/or long text inputs, but this is easily adjustable by widening the class. The single set of start/end angle brackets do however make it clear that one stereotype is concerned. 

No line breaks and a dynamic width instead of height would have been ideal, however the solution has improved to a degree that I now consider it reasonable to merge.

I approve!

a23miksk:
The reviewer approved this change but I as responsible for issues will deny this, the solution is not solved. There should only be _one_ line with no row breaks. And there is _a lot_ of new code added in this commit for such a small change that has unclear purpose, which means I can't merge this to the weekly branch in good conscience. Will close this but keep the branch open if it can be used in future development.

PR #17693, opened 2025-05-19
Added inverse dependencies for getUid_ms.php in Microservices_inverse_dependencies.md #17664

### LINKED ISSUE #17664: Add inverse dependencies for getUid_ms.php in Microservices_inverse_dependencies.md
Some files are missing from the documentation and have to be added before I can continue with the re-engineering of getUid_ms.php.

a23louma:
Fixes #17664 . Some inverse dependencies were missing for getUid_ms.php and have been added to Microservices_inverse_dependencies.md. It's important that all of them are correct before the re-engineering can be done for this microservice.

### DIFF
--- a/DuggaSys/microservices/Microservices_inverse_dependencies.md
+++ b/DuggaSys/microservices/Microservices_inverse_dependencies.md
@@ -364,9 +364,10 @@ createListEntry_ms.php
 
 ##### sharedMicroservices
 - sharedMicroservices/createNewCodeExample_ms.php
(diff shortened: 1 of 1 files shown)


a23axehe (changes requested):
Looks good but there is a few things missing, updateSecurityQuestion_ms.php seems to be listed twice. courseedService title has 6 #'s which makes it more grey, the other titles have 5 #'s. The microservice retrieveCodeviewerService_ms.php in codeViewerservice is also dependant on getUid but seems to be missing from the documentation. After these changes its all good.

Issue #17555, opened 2025-05-14
Fix the lines floating on ER attributes


a23miksk:
Currently the lines float a little bit when attached below an ER attribute. 

_Image where the line from Attribute to Appartment1 anchors some pixels below where it should_
![Image](https://github.com/user-attachments/assets/36e69d29-56e2-42ab-acbc-c719b24c1fbd)

PR #17696, opened 2025-05-19
Fixed: G1 2025 v7 #17555 fix the lines floating on er attributes

a23aleva:
Lines will no longer stop short and create a gap when trying to attach them to an ER Attribute element. Instead the line should now run all the way to the ellipse regardless of its position.

https://github.com/user-attachments/assets/538fdd2a-e24b-4a15-bb7c-b141df541e8a


a23kevre (approved):
the line is fixed. approved!

Issue #17431, opened 2025-05-12
Implement mockup for the option panel


b23danhe:
Implementation of the mockup of the option panel when the mockup is approved by the customer.

Check comment in issue #17413 of what to implement.

b23danhe:
The mockup in issue #17413 has been approved by the customer, so this issue can now move forward.

PR #17700, opened 2025-05-19
Fixed: G1 2025 v7 #17431 implement mockup for the option panel

a23eliax:
Made a working fab-button for the options menu that is only visible at mobile version.

Added a close-button to the options
menu. Also only visible when user is on mobile version.

https://github.com/user-attachments/assets/ab271a48-5ab5-4385-a5a6-a532d3868f13

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -423,10 +423,6 @@
     .hide-options-pane {
         right: -500px;
     }
-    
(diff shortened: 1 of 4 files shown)


a23gabst1 (approved):
The implemented mockup for hiding the option panel on mobile version works as a charms! The implementation looks fine and I approve this!

PR #17704, opened 2025-05-19
Fix Lines move weirdly with recursive

a23carpa:
Changed the part where to recursive lines should be sorted out of the offset.
The logic was broken and the recursive line where not excluded from the offset.
With this change the recursive line is now not part of the line offset and the lines behave as before and multiple recursive lines can not be stacked on each other.

[Screencast from 2025-05-19 11-59-21.webm](https://github.com/user-attachments/assets/3db9b40c-5f0a-4d07-9d4c-d21140a27a26)


b23ebbst (approved):
Looks great! The lines behave as they should, and the recursive lines do not stack in a ladder-formation any more. 

Really well done, I approve! ✨✨

PR #17707, opened 2025-05-19
SSL integration into newinstaller

a23arvme:
The SSL script is integrated as a 6th step in newinstaller. To test, just run through the installer, then check if there is a _localhost.key_ and a _localhost.crt_ at _.../newinstaller/tools/ssl/_.

In the future this installation step will refer to tutorials for docker and XAMPP on how use the SSL keys.


a23gabli (approved):
Everything works as intended and the files are present.

Issue #17590, opened 2025-05-14
Sequence activation lines doesn't work properly


a23emijo:
The lines between sequence activations doesn't work properly as shown in the videos and the picture down below. I think that the offset distance has something to do with the amount of zoom the user uses at the current moment.

_**Video to show the issue**_
https://github.com/user-attachments/assets/46e9990b-862c-450e-88b6-110124d1d76d

_**Zoomed in**_
https://github.com/user-attachments/assets/65197061-ad7d-41d5-b4fe-674e6d30ff70

_**Zoomed out**_
https://github.com/user-attachments/assets/abdf7c88-81bb-4056-beba-951384888ec9

_**This abomination I managed to create**_ (the line at the top seems to be connected between the first sequence activation and the Object box but that is not the case, it is actually connected to the second sequence activation, the one at the bottom of the line, on the object line to the right)

![Image](https://github.com/user-attachments/assets/3384ccc0-6806-4344-b0db-a078fb14d579)

PR #17711, opened 2025-05-19
Fixed the sequence diagram lines #17590

a23miksk:
closes https://github.com/HGustavs/LenaSYS/issues/17590.

Added bounds detection so that the lines only can be drawn to a position in an element that is within the first clicked elements y-coordinates.
Implemented a way of visually showing how the lines will look while dragging it around. Also removed the 3 line limit on activation elements.

_Video of the implementations_
https://github.com/user-attachments/assets/95b7ce5c-5b66-41ef-8c78-66a27017b395

### Known issues that still exists are:
Resizing the activation element will move the lines to unwanted positions. Arrows will also stick around weird places when resizing.
Moving an activation element will only carry with it the lines that were drawn _to_ it.
Undo/redo still a bit broken in sequence diagram, the line might not be added to statemachine or something.

### DIFF
--- a/DuggaSys/diagram/draw/line.js
+++ b/DuggaSys/diagram/draw/line.js
@@ -601,6 +601,13 @@ function getLineAttributes(line, f, t, ctype, fromElemMouseY, toElemMouseY) {
     
     // Special case to handle sequence activation lines
     if (f.kind === elementTypesNames.sequenceActivation) {
+        //console.log(f.y, f.height, t.y);
+        
+        if (f.y > t.y || f.y + f.height < t.y){
+            console.log("utanför");
+            
+        }
+
         const fromKey = `from:${line.id}`;
         const toKey = `to:${line.id}`;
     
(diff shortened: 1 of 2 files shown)


a23aleva (approved):
Lines now look cleaner than before and, most importantly, they remain straight instead of tangled, which is especially critical for a sequence diagram. While they appear to only be drawable between activation elements I get some strange automatic regulation when placing them so that they are never really touching the upper edge of the element. Sometimes the line doesn't really land where it is indicated to (sometimes with one end outside the element boundary). The line being placed where the user puts it is something that should be assessed.

While not perfect these changes still look better than before and the known issues mentioned by the developer persist. I'd also like to add that labels get inconsistent placement along the line to this list of known issues. But since all these things can be focused in new issues, to be fixed at a later stage, I still approve these particular changes.

https://github.com/user-attachments/assets/10c763ed-e384-400d-80a9-6c3aaf75294e

PR #17712, opened 2025-05-19
Fixed option menu does not show all options

a23emijo:
The problem were not that the elements didn't show, because they are always "visible". The problem was that depending on the users screen size, they may had to scroll down in the whole browserwindow. This made some options "disappear" when they in reality only were outside of the screen. To fix this, I implemented a scrollbar in the options panel and styled it a bit to make it fit in more (it can still be a lot nicer).

**NOTE:** This scrollbar only works in web browsers that uses chromium. The other ones gets the default ugly looking gray scrollbar.

https://github.com/user-attachments/assets/06e32b75-9005-49f8-b4b8-ac401f651fa0

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -165,6 +165,8 @@
     padding: 10px;
     transition: right .3s ease-in-out;
     user-select: none;
+    overflow-y: auto;
+    overflow-x: hidden;
 }
 
 #options-pane-button {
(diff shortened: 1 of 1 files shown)


b23ebbst (approved):
Works really well, and the scrollbar fits nicely with the rest of the design on the page (not too bad in firefox either, it just looks grey and a little boring). 

Well done!! I approve!

https://github.com/user-attachments/assets/115bbcd1-3d97-47ae-bd10-79fd42b6e9d8

PR #17713, opened 2025-05-19
Add dependencies search UI #17633

### LINKED ISSUE #17633: Add dependencies search UI
Create a page where you can search for a microservice and get a list of the other microservices depends on that microservice. It should look like microserviceUI and be accessible from it. Microservices should be clickable and displayed (on microserviceUI) when clicked.

a23oskme:
Created an UI where users can search for complete dependencies for a microservice. Also created a form that lets the user insert dependencies into the database. Fixes #17633 

The path field is optional in the form because I was unsure wether it is important or not. 
We should create a button on microserviceUI that takes the user to dependenciesUI when it is merged and completed.

### DIFF
--- a/DuggaSys/microservices/endpointDirectory/dependenciesUI.php
+++ b/DuggaSys/microservices/endpointDirectory/dependenciesUI.php
@@ -23,6 +23,19 @@ if (isset($_GET['ms_name'])) {
 
     if ($result) {
         $id = $result['id'];
+
+        $dependingServices = [];
+        $dependentServices = [];
+    
+        // query to get dependencies towards the searched microservice
+        $stmt1 = $db->prepare("SELECT depends_on, path FROM dependencies WHERE microservice_id = ?");
+        $stmt1->execute([$id]);
+        $dependingServices = $stmt1->fetchAll(PDO::FETCH_ASSOC);
+    
+        // query to get what microservices the searched microservice is depending on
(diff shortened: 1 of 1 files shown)


a23louma (commented):
The dependencies UI works as expected. Both the search and insertion works inside the dependencies UI and you can return to the microservice UI by clicking the button. However, after an insertion in the dependencies UI only "Dependencies" are updated in the microservice UI and not "Microservices used". Shouldn't it be updated as well @a23oskme ?

a23oskme:
> The dependencies UI works as expected. Both the search and insertion works inside the dependencies UI and you can return to the microservice UI by clicking the button. However, after an insertion in the dependencies UI only "Dependencies" are updated in the microservice UI and not "Microservices used". Shouldn't it be updated as well @a23oskme ?

The "Microservices used" is based on data from the documentation in the markdown-files. Maybe we can create a new issue to insert new entries (from dependenciesUI) in the microservice entry. 
Also, only inverse dependencies are displayed on microserviceUI at the moment. We could also create an issue so that dependencies are listed as well, and maybe remove the "Microservices used" part?

a23louma (approved):
Yes, I agree with you. I approve this PR and new issues can be created so that the dependencies/inverse dependencies in microserviceUI correspond to the ones in dependenciesUI.

Issue #17495, opened 2025-05-13
(mobile-view) BUG wrong darkmode background color in sectioned.php dropdown


a23osckr:
When in darkmode in a course, the dropdown that appears on the left has the wrong brightness of purple regardless of access level. This image was taken while logged in as admin since this makes the dropdown larger and the issue is seen more clearly.
![Image](https://github.com/user-attachments/assets/71ded051-ee5b-4122-b215-f1a730bbb56f)

PR #17715, opened 2025-05-19
Changed the background color of burger drop down on mobile #17495

a23erigu:
I think this is what the issue wanted.
Only added the hamburgerBox tag to the dark mode file and gave it the correct color

pic of how the dropdown looks in dark mode now.
image

I personally think it looks worse but this is what the issue wanted (unless I have misunderstood something)

### DIFF
--- a/Shared/css/blackTheme.css
+++ b/Shared/css/blackTheme.css
@@ -67,7 +67,6 @@
 	background-color: var(--color-background);
 }
 
-
 #picktemplate{
 	position: absolute;
 	top: 0px;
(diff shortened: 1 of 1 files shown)


a23simry:
image
I would prefer a colour like this but this might be a costumer/group leader question

a23erigu:
I have now made the background black.
@a23simry 

image

EdvinWerner (approved):
Nothing more to say other than it looks good to me, approved for merge!

PR #17716, opened 2025-05-19
Converted jquery in \DuggaSys\templates\dugga2.js #17530 

f23linan:
Converted jquery to JavaScript. 

Removed 2 functions that served no function as well.


a23erigu (changes requested):
Now it seems that all the Jquery (except the Ajax) has been removed.
The code makes me want to cry but that has nothing to do with this so I'm gonna ignore that.
I don't get any errors (from this js).
But I did find this bug that forces me to deny this PR.

When you try and change a number in dugga 2 it normally looks like this.
image

After your changes it looks like this.
image

The table for the choosing what to replace with seems to be trying to hide at the bottom of the screen.
Sorry to say but this has to be fixed to be able to accept this PR
It shouldn't be that hard of a fix from what I can see tho.

Issue #17702, opened 2025-05-19
Fix GitHub functionality


b23johgu:
The GitHub functionality seems to be broken, or not implemented. Check out what needs to be done to fix/implement it and fix it, or if it's too big, write a comment about it and/or create sub-issues for what needs to be done in order to fix it.

> When creating a new course with the webprogramming repo linked, none of the code examples are present. And it seems like no data is pulled from the repo. There is no way to test when and how frequently the code is updated because there is no code showing in the first place. Unless i am missing something, the github functionality needs to be fixed/implemented before we can start working on acceptance testing. 

 Originally posted by @a23gabli in [#16514](https://github.com/HGustavs/LenaSYS/issues/16514#issuecomment-2830404934)

b23johgu:
This was apparently the responsibility of Group 3, so I will close this issue!

PR #17717, opened 2025-05-19
Tried to fix GitHub functionality #17702

a22linre:
The problem I found was that when a new course is created, the files from the GitHub repo are not downloaded into the course folder (courses/"cid"/Github/) like they should be.

I’m committing the changes I’ve made so far related to the GitHub file download when a new course is created.

I don’t even know if this is the actual problem we’re supposed to fix, but I tried to make it so the files from the GitHub repo are downloaded automatically. I got it working through a separate test file, but couldn’t get it to work properly from the normal course creation automatically.

I’ll commit what I have anyway, just in case it helps or someone else knows how to continue from here. I’m a bit lost. 

#17702

### DIFF
--- a/DuggaSys/gitfetchService.php
+++ b/DuggaSys/gitfetchService.php
@@ -289,4 +289,18 @@ function bfs($url, $cid, $opt)
     }
 }
 
+if ($_SERVER['REQUEST_METHOD'] === 'POST') {
+    $cid = $_POST['cid'] ?? null;
+    $url = $_POST['githubURL'] ?? null;
+
+    if ($cid && $url) {
+        // Call bfs to download files from GitHub
+        bfs($url, $cid, "DOWNLOAD");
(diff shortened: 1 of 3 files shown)


a22linre:
When you create  a new course, the course get a cid which can be found if you go in to localhost/phpmyadmin and look for courses. There you can find cid and then you have to replace it in the test file with it.

a23oskme (approved):
Thank you for the clarification! 
I did this and can confirm that the code you provided works. 

image

<i>Screenshot shows the course i created (with 1895 as an ID) and that the github folder got downloaded after running</i> `test_gitfetch.php`.

Although, if it doesn't solve the issue you might want to create a new one and reference this one, I'm sure it would be helpful when working on it.

b23johgu:
I'll merge this so we can continue working on it in next weeks branch if needed. I won't close the issue just yet if you want to create a sub-issue that solves the rest. But either way it seems to work better than before and that's a win!

Issue #17703, opened 2025-05-19
Fix so fab buttons doesn't disappears when loading file


a23eliax:
When the user wants to load a file, and then clicks on "latestChange", it works as expected. But when they presses a saved file they want to load, the diagram fab button AND the option menu fab button disappear. This should not be happening.

https://github.com/user-attachments/assets/a0a9ac52-3b54-4f4c-9a5f-0ca85c74cba8

a23gabst1:
This only occur when the name of the saved file is longer than the width, the solution would be to implement a max-width and a text-overflow ellipsis on the p-tagg on wrapped in the diagramFileName div. So the buttons are still there but at different positions due to the causing overflow.

PR #17720, opened 2025-05-19
Fixed: Fix so fab buttons doesn't disappears when loading file

a23gabst1:
The buttons did not disappear, they were out of the screen because the diagramFileName created overflow. The solution was to set a max-width on it and then place a overflow hidden and text-overflow ellipsis on the p-tag. This creates an ellipsis implying that the file name is longer only when the name is longer than the width of the diagramFileName. see video 1 for illustration. 

Fixed the issue #17703 

video 1

https://github.com/user-attachments/assets/2fe3ca35-899f-4b61-83ed-457eb11e715b

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -74,12 +74,18 @@
 #diagramFileName {
     position: absolute;
     z-index: 1000;
-    margin-left: 50vw;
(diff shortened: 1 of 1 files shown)


a23eliax (approved):
Work as expected! I appove.

PR #17723, opened 2025-05-19
G3 2025 v7 #17532 convert jquery to js in dugga4.js

a22maxda:
Converted JQuery code to JS and fixed a bug where the text on the operation button was sometimes not updated correctly. As far as I can tell nothing is broken or works incorrectly.


a22maxda:
I found a bug in the code so it should not be reviewed yet.

a22maxda:
The bug has been fixed.

a22maxda:
SaveClick has correct indentation now

a23yutta (approved):
The code works as intended and does not seem to break any existing features.

I haven’t been able to verify whether the `showFacit()`  and `closeFacit()` functions works. However, the code appears to be correctly converted, and it seems that the "show/close facit" feature hasn’t been introduced into the page currently. I believe it's safe to merge this PR.

Good job!

[APPROVED]

Issue #17533, opened 2025-05-13
convert jquery in \DuggaSys\templates\dugga3.js


a23fanro:
Locate jquery in the file and convert it to javascript. 
DO NOT convert jquery that is used with AJAX, this will be done later.
Also remove code that has been commented out if it is jquery.

c22isako:
Additional information:
File is used within the course: webbutveckling - datorgrafik, section: Geometri.

PR #17725, opened 2025-05-19
G3-2025-v7-#17533-convert-jquery-in-dugga3.js

c22isako:
### Changes made:
Replaced JQuery within the file dugga3.js, which were represented via the course: _Webbutveckling - datorgrafik_, and its _Geometri_ section. Both Linjedugga 1 & 2 were affected by the changes, and should match the weekly branch in terms of functionality. 
_Note that other duggas might have been affected, though I haven't noticed any problems._

**Ignored functions:**
- Functions which contained AJAX, which will be replaced within another issue.
- Functions connected to feedback, which seems to be removed and will be replaced within another issue.

_That said, some of the functions within the most recent commit might have affected code related to feedback as well - ignore if so._

### DIFF
--- a/DuggaSys/templates/dugga3.js
+++ b/DuggaSys/templates/dugga3.js
@@ -97,7 +97,6 @@ function returnedDugga(data) {
 		alert(data['debug']);
 
 	if(data['opt']=="SAVDU"){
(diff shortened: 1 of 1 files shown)


a23lisjo (approved):
The JQuery is changed to JS and as the description says, the only JQuery that is left is about AJAX and Feedback. The code looks good and I have tested the duggor and it works like it did before. There is one more dugga that uses this template in _Demo-Course_ --> _Dugga 3_, I have not found any problem with this dugga as well.

Issue #17705, opened 2025-05-19
Toolbar cuts off the "Load"-button


b23ebbst:
The toolbar has become "too long" to fully show all tools. The last button, "load", gets cut-off in the middle because of this. 

Note that the button still fully functions, it is just a "superficial" issue where the look of the toolbar is affected. A solution might be to allow scrolling in the toolbar (this also allows for more tools being added in the future without affecting the visuals, or having to adjust for them), or to lengthen the entire drawing-/toolbar-area to fit the full button for the load-function. 

![Image](https://github.com/user-attachments/assets/878d6eb7-1581-4a28-97f4-c09cb0563067)

a23matsh:
There is no issue with the toolbar cutting off any icons, closing.

![Image](https://github.com/user-attachments/assets/89f14327-fb30-4480-8a73-c1bec38d5306)

PR #17728, opened 2025-05-19
Adjusted toolbar issue #17705

a23matsh:
Adjusted the toolbar icons sizing, they all now fit 

image

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -303,13 +303,10 @@
 }
 
 .diagramIcons {
-    position: relative;
-    margin: auto;
-    width: 3.7vh;
-    height: 3.7vh;
-    cursor: pointer;
-    border: solid 1px var(--color-primary);
-    margin-top: 3px;
+    width: 3.8vh;
+    height: 3.8vh;
+    box-sizing: border-box;
+    transition: background-color 0.3s;
 }
 
 @media (max-height: 650px) {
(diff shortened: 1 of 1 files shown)


a23emijo (changes requested):
- To start of, my laptops screen is very small compared to other laptops so my operating system recommends me to use 150% zoom in it at all time. This makes the icons appear smaller in the toolbar, I don't think that this is your problem though, but because you moved the icons to the left of the toolbar, a lot of emptiness appears to the right of the icons. A solution to this (which should be another issue) is to make the icons re-size with the toolbar, because the toolbar already re-sizes when you zoom (it gets thinner or thicker) but the icons stay the same size at all time. I also think that the icons should still be in the center like they were before and not to the left. Maybe make the whole toolbar scrollable, but make the scrollbar invisible, instead of moving or re-sizing the icons?

image

- Why did you remove the white triangle in the bottom right corner of the icons that has a dropdown menu? It was there to indicate that the icon has more icons hidden behind it.

- Why did you remove the "cursor: pointer" in the css? I think it's nicer to have the cursor change to a hand pointing at the object the user hovers. I do like the short transition you added on the icon hover though. 👍

a23matsh:
Adjusted the toolbar for the requested changes

-Toolbar still fits all the icons in it without any cut offs
- The cursor is back
- The icons were previously shifted to the left and not the middle since resizing, this is now fixed they remain in the middle

b23danhe:
The new issue handling scrollbar is #17756

Issue #17538, opened 2025-05-13
convert jquery in \DuggaSys\templates\generic_dugga_file_receive.js


a23fanro:
Locate jquery in the file and convert it to javascript. 
DO NOT convert jquery that is used with AJAX, this will be done later.
Also remove code that has been commented out if it is jquery.

PR #17729, opened 2025-05-19
Convert jquery in \DuggaSys\templates\generic_dugga_file_receive.js #17538

a22carsv:
Converted JQuery for 'generic_dugga_file_recieve.js', there should be none remaining except for AJAX related.


f23linan (changes requested):
Looks good! Everything works as expected. 

Just a small change needed.

"document.getElementById("snus").innerHTML=result;
// Placeholder code
var pl = duggaParams.placeholders;" 

Needs to be reverted back to the original code as this is also part of the Ajax jquery that is not to be changed.

When this is changed it can be approved!

PR #17733, opened 2025-05-20
Fixed: Multiple activation element snap bug

a23aleva:
Now multiple sequenceActivations are allowed to snap, at the same time, to a lifeline when they are box selected. However if there are different elements within the box selection (in addition to sequenceActivations) the erratic behaviour still persists for these. This problem was there before this fix but at least sequenceActivations now behave correctly. The bug is a minor one as the user will seldom attempt to box select-snap a mixture of elements not belonging to a sequence diagram when that diagram type is the focus. This could be targeted in a future issue should it be deemed necessary.

https://github.com/user-attachments/assets/dc4c0083-72e8-4654-a61a-e6c2bc6f4418

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1143,14 +1143,25 @@ function mmoving(event) {
 
                 // Check coordinates of moveable element and if they are within snap threshold
                 const moveableElementPos = screenToDiagramCoordinates(event.clientX, event.clientY);
+                // Lifeline coordinates to visualize snap for selected activation elements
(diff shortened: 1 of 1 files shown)


b23ebbst (approved):
As said in the description, when only selecting Sequence activations with box select, they snap beautifully to the lifeline. If there is another object (not a sequence activation) in the selection the behaviour is a little odd. However, I agree with that the users rarely will try to snap other elements to a sequence lifeline, so that feels like a very small issue. 

Good job! I approve! ⭐⭐⭐

PR #17735, opened 2025-05-20
Add dependencies to microserviceUI #17721

### LINKED ISSUE #17721: Add dependencies to microserviceUI
Only inverse dependencies are displayed for a microservice at the moment (microservices that depends on the chosen microservice). We should create a table that also show dependencies for the chosen microservice (microservices that the chosen microservice is depending on).

a23oskme:
Added a query to get dependencies and added another table to the dependencies. Now both inverse dependencies and dependencies for a microservice are displayed. 
Only inverse dependencies was displayed before. Fixes #17721 

image

### DIFF
--- a/DuggaSys/microservices/endpointDirectory/microserviceUI.php
+++ b/DuggaSys/microservices/endpointDirectory/microserviceUI.php
@@ -103,6 +103,12 @@ if (isset($_GET['id'])) {
         $stmt->execute([$_GET['id']]);
         $dependencies = $stmt->fetchAll();
     }
+
+    if ($microservice) {
+        $stmt = $db->prepare("SELECT * FROM dependencies WHERE dependency_id = ?");
+        $stmt->execute([$_GET['id']]);
+        $depending_on = $stmt->fetchAll();
+    }
 }
 
 } catch (PDOException $e) {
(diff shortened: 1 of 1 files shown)


a23louma (changes requested):
It's good to be able to see both inverse dependencies and dependencies for a microservice! I was just wondering, shouldn't the same microservices appear under "dependencies" as for "Microservices used"? For example, if I go to "updateCourse_ms.php", it has "deleteDugga_ms.php" under dependencies. But shouldn't it be "retrieveCourseedService_ms.php, getUid_ms.php and retrieveUsername_ms.php"?

a23oskme:
> It's good to be able to see both inverse dependencies and dependencies for a microservice! I was just wondering, shouldn't the same microservices appear under "dependencies" as for "Microservices used"? For example, if I go to "updateCourse_ms.php", it has "deleteDugga_ms.php" under dependencies. But shouldn't it be "retrieveCourseedService_ms.php, getUid_ms.php and retrieveUsername_ms.php"?

It should be fixed now. 
image

However, not all microservices listed under 'Microservices used' in your example are outputted due to inconsistencies in the documentation.

a23louma (approved):
I approve, now it looks correct! The inconsistencies in the documentation cannot be fixed now but at least there are correct microservices listed under Dependencies.

PR #17736, opened 2025-05-20
Fixed URL and removed header in sectionedService/createListEntry_ms.php #17708

### LINKED ISSUE #17708: Wrong url in sectionedService/createListEntry_ms.php
I have found a problem with createListEntry, the url is set wrong:
header("Content-Type: application/json");
$baseURL = "https://" . $_SERVER['HTTP_HOST'];
$url = $baseURL . "/LenaSYS/DuggaSys/microservices/sharedMicroservices/createNewCodeExample_ms.php";

Header should not be there and the url should be this:

$url = "sharedMicroservices/createNewCodeExample_ms.php";

a23louma:
Fixes #17708 . I've removed "header("Content-Type: application/json");" and changed the URL so it works for the curlService function callMicroservicePOST. I also removed "global $pdo" as this gave an error.

### DIFF
--- a/DuggaSys/microservices/sectionedService/createListEntry_ms.php
+++ b/DuggaSys/microservices/sectionedService/createListEntry_ms.php
@@ -34,9 +34,6 @@ $templateNumber = getOP('templateNumber');
 $exampleid = getOP('exampleid');
 $debug = "NONE!";
 
-global $pdo;
-
-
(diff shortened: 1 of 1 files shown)


a23axehe (approved):
Fixes the issue by correctly setting the url and removing a faulty header. Approved

Issue #17734, opened 2025-05-20
Mobile view: Infinite number of instances of the page can be loaded inside the digram iframe


b23danhe:
If you press "Home" while in showDugga on the mobile version, the page loads inside the same iframe that contains the diagram.
This results in a new instance of the page being nested inside the existing iframe, and this can be repeated indefinitely, effectively allowing an infinite number of frames to be loaded within each other.

https://github.com/user-attachments/assets/1b8e9a6f-6a1b-4bc8-adbe-2fbc2a5b495c

<img width="414" alt="Image" src="https://github.com/user-attachments/assets/5b0e5573-1622-48ec-8d31-9b08bb5a2be7" />

PR #17743, opened 2025-05-20
Fixed home-screen instances on mobile (#17734)

b23ebbst:
When using the home-button on mobile, it would nestle the home-page inside the diagram-area rather than in the full window, see issue #17734 . Fixed the issue so it affects the full screen/window instead, and it no longer nestles the page within the diagram-area. 

https://github.com/user-attachments/assets/d0ea6726-27fe-46d7-9d28-73ddc817069c


a23aleva (approved):
This effectively fixes the issue of infinite windows opening inside the grid viewport. It is a small and elegant change and it is approved!

https://github.com/user-attachments/assets/02f5f469-fb87-40f2-8464-0ca3ca33fb5d

PR #17752, opened 2025-05-20
G3 2025 v7 #17535 convert jquery to javascript dugga test

### LINKED ISSUE #17535: convert jquery in \DuggaSys\templates\duggaTest.js
Locate jquery in the file and convert it to javascript. 
DO NOT convert jquery that is used with AJAX, this will be done later.
Also remove code that has been commented out if it is jquery.

a23lisjo:
Fixes #17535. Converted all found JQuery to vanilla JS in duggaTest.js. I have not been able to find a page that uses this code so there have not been any testing.


a23elism (approved):
As the original author of the PR brought up, there doesn't seem to be a dugga that uses this file (JS or html version) although the JQuery seems to be properly converted, everything looks to be in order, and thus will be approved for merge. 

Approved for merge 👍
---

### Recommend making a document for easier access on how the duggor look like or where they can be found. It is kind of an easter egg hunt at the moment.

PR #17753, opened 2025-05-20
Fixed elements placed on the canvas is above the FAB

a23emijo:
This solution is not good but does the job. I changed the z-index from 10 to 5000 but the user can still make the object appear above the FAB and the menu-buttons if the "Bring to Front" is pressed enough times. I can try to implement a better solution for this but it has to be made next week (G1-2025-v8).

https://github.com/user-attachments/assets/7e0d2d47-d76d-4ad5-8c17-242c64fbed73

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -331,7 +331,7 @@
     .icon-wrapper{
         display: block !important;
         position: fixed;
-        z-index: 10;
+        z-index: 5000;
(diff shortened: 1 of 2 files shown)


a23thefo (approved):
It dose work! Like Emil is saying its a band aid solution but I spoke with him yesterday about a proper solution that can be implemented next week so I approve

Issue #17731, opened 2025-05-19
Proposal of feature: Automatically toggle options menu when selecting an item


a23emisn:
Proposal of feature.
This might be a bit subjective, but something that bugs me a bit using the diagramming tool is that if you want to edit an attribute for a certain entity, you first need to select the item and then open the options panel manually. 

To clarify, this panel:
![Image](https://github.com/user-attachments/assets/c22eac3d-66e3-47c4-9ab4-b16b1db6209e)

This feels clunky and I believe it would be beneficial for user friendliness if the options panel opened up automatically when selecting an element. However it might not be that straightforward to implement this, as some questions need to be answered first. What should happen if the element is de-selected? Should the options panel close? Remain open? If the user had the options panel opened to begin with, should it close when de-selecting the elements? Should box selection also be included? etc. 

The current behavior is logical and works quite well. But I believe it would be worth investigating the possibility of implementing a more dynamic behavior of this panel, as it would further enhance the usability of the diagram dugga.

HGustavs:
Not a bad idea this should be pretty simple to implement.

PR #17760, opened 2025-05-20
Implemented feature: Automatic options-panel toggle

a23emisn:
This is an implementation of the feature mentioned in the following issue: (#17731). It allows for the options panel to toggle automatically upon selecting elements, which allows for a smoother interface and a faster and easier UX. 

**It works as follows:** 
By default the options panel will open when selecting an element to edit, and it will hide as soon as either the canvas or the toolbar is clicked. 

The user can also lock the toolbar into a desired position i.e. they can open the toolbar before placing any elements and it will remain open until the user closes it. During this state it stays unaffected by any automatic toggle-attempts following selection and de-selection of elements.

Similarly, the user can close the toolbar if they wish to keep it closed while editing elements, and it will remain closed until the user opens it again. Same here, not being affected by any automatic toggle-attempts.

 To reset the toolbar behavior the default behavior, manually opening/closing the toolbar again will reset it, depending on what position it was locked in.

**View this as a partially complete solution, that can be extended upon if deemed suitable for merging.**

Additional notes: 
* This is implemented for single-selection of elements, but can easily be extended to account for box-selection as well if needed to. 
* The locking of whatever toolbar behavior is in use (staying closed or remaining open) isn't inherently clear to the user right now. But a lock icon or a message dialog or similar can be implemented with ease to clearly signal this status to the user. Similarly, the locking can be attributed to a designated button instead of resulting from operating the toolbar manually, which might not be inherently obvious.

### DIFF
--- a/DuggaSys/diagram.php
+++ b/DuggaSys/diagram.php
@@ -810,7 +810,7 @@
     <!-- OPTIONS PANE -->
     <!-- Yellow panel on the right side of the screen -->
     <div id="options-pane" class="hide-options-pane"> 
-        <div id="options-pane-button" class="tooltip-target" onclick="toggleOptionsPane();" data-toolmode="Option_Panel" data-toolid="24">
+        <div id="options-pane-button" class="tooltip-target" onclick="toggleOptionsPane(true);" data-toolmode="Option_Panel" data-toolid="24">
             <span id='optmarker'>&#9660;Options</span>
         </div>
 
--- a/DuggaSys/diagram.php
+++ b/DuggaSys/diagram.php
@@ -810,7 +810,7 @@
     <!-- OPTIONS PANE -->
     <!-- Yellow panel on the right side of the screen -->
     <div id="options-pane" class="hide-options-pane"> 
-        <div id="options-pane-button" class="tooltip-target" onclick="toggleOptionsPane(true);" data-toolmode="Option_Panel" data-toolid="24">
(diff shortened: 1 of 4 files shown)


a23carpa (commented):
My testing shows that the option panel works as stated and it feels good.
Also as mentioned in the pull request, if i did not know i could lock it and unlock it, it might have been more confusing, i dont know.

Side not. Possible issue, i experienced some times that the item drag effect would stop working. But i could not forcefully recreate it for the video and only experienced it once or twice. 

[Screencast from 2025-05-21 08-35-50.webm](https://github.com/user-attachments/assets/be538dbc-ca21-456f-95be-96c7124c5431)

a23emisn:
> My testing shows that the option panel works as stated and it feels good. Also as mentioned in the pull request, if i did not know i could lock it and unlock it, it might have been more confusing, i dont know.
> 
> Side not. Possible issue, i experienced some times that the item drag effect would stop working. But i could not forcefully recreate it for the video and only experienced it once or twice.
> 
>  Screencast.from.2025-05-21.08-35-50.webm

Thanks for the input! The concerns you mentioned definitely hold weight, and I also think the locking/unlocking behavior is a bit unclear. As for the bug with the drag effect—I haven’t experienced it myself, but I agree it’s possible that it could be related to the current solution in some way.

Personally, I think it might be worth spending a bit more time polishing this before merging, as the current version may not be quite ready yet.

Issue #17714, opened 2025-05-19
Re-engineer setAsActiveCourse to use recieveMicroservicePOST


a22wilas:
At the moment the microservice setAsActiveCourse manually verifies the POSTED data instead of uses the function recieveMicroservicePOST from curlService.php. 

Fix the code so the uses curlService.php to receive and verify the data.

PR #17761, opened 2025-05-20
Added receieveMicroservicePOST for setAsActiveCourse

a22wilas:
Fixes issue #17714

To test the solution change the active course version and check the database table "course".
To change active version in the database first change the version on the application, then press the cogwheel and fill in the parameter "change this to the default value. The active version should now be seen in the database.
bild
bild
bild


a23dilel (approved):
I followed these steps and it worked as intended. I created a course 2222 and you can see these screenshots:
v
v2

Also, I received setAsActiveCourse data in my Docker terminal.
v3

Overall, great job!

Issue #17563, opened 2025-05-14
Improve the segmented lines


a23miksk:
The customer thinks that the segmented lines should probably be completely straight lines when the elements "basically" are aligned with each other. This means that the segmented line probably should have some "break point" where it snaps to a straight line that goes up/down or left/right.

_This line should become straight according to the customer_
![Image](https://github.com/user-attachments/assets/0882cc54-7ddc-4c5e-b0c7-bb93072dfb10)

PR #17766, opened 2025-05-20
changes to drawLineSegmented

a23vinge:
The solution to the issue [#17563](https://github.com/HGustavs/LenaSYS/issues/17563) "Improve the segmented lines" did not require redefining how a line’s start and end points are set, adding complex proximity logic, or rebuilding the line rendering system to switch between bent and straight modes.

Instead, the fix was much simpler: by updating the drawLineSegmented function, the line (or arrow) now automatically draws straight when the connected elements are nearly aligned. This approach improves the visual result without altering any core behavior—just a smart enhancement to the drawing logic.

https://github.com/user-attachments/assets/50d32af2-d4df-4a1e-9def-48f4e95c851b

As the video shows, the solution did fix the problem, but there is a small bug that makes the arrow look weird.

### DIFF
--- a/DuggaSys/diagram/draw/line.js
+++ b/DuggaSys/diagram/draw/line.js
@@ -64,7 +64,8 @@ function drawLine(line, targetGhost = false) {
         ty = event.clientY;
     }
 
-    [tx, ty] = isAligned(fx, fy, tx, ty, offset, 10);
+    // Checks if the lines are "basically" aligned, and modifies tx and ty so the line is straight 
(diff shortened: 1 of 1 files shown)


a23miksk:
Had time over so made an attempt to fix the arrows

a23thefo (approved):
Code looks good and solution as well! I have been wrestling with this file for a couple days now so i know how shit it is! Good job!

ezgif-65d77fbd8a0fff

Issue #17718, opened 2025-05-19
Hover effect on options panel


b23danhe:
The hover effect doesn't seem to be implemented for the "Grid" and "Snap to Grid" buttons. Additionally, the hover effect on the "Ruler" button stops working after it has been clicked once.

To fix: Ensure the hover effects are consistent and work correctly within the option panel.

https://github.com/user-attachments/assets/bff24214-2401-4382-9dfe-d13688fb3853

PR #17771, opened 2025-05-20
Fixed hover effect in options panel

a23kevre:
A function in toggle.js made inline styling which overrid diagram.css. Resulting in the hover effect being useless. Added !important to the hover class to make sure it is always active. Make sure to check issue #17718 for a video of before this change as reference.

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -520,12 +520,12 @@
 }
 
 .saveButton:hover {
-    background: #815e9d;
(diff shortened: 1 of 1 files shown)


a23aleva (approved):
The hover effect now works consistently for all buttons in the options menu. The effect also stays for the ruler button even after it has been clicked. I approve these changes!

https://github.com/user-attachments/assets/ce4a8424-9754-49d2-9c0b-bb16165579e7

Issue #17616, opened 2025-05-15
Fix so the lines connect to the arrows in state diagram and lines does not move


a23eliax:
This issue is about the arrows and lines in the state diagram. When the user moves around the states, it can disturb the other lines. It's a little better when the line is segmented but the issue still remains. The video below illustrates the straight lines, which is the biggest issue. When a state comes close to a line, it moves, when it shouldn't move. And the arrows seems to not connect to the line anymore when the user moves it in some directions.

The issue https://github.com/HGustavs/LenaSYS/issues/17559 handles the arrow part.

https://github.com/user-attachments/assets/87b0f89b-627a-420a-a8cf-df49dc24829a

b23danhe:
The issue #17559 handles the arrow part, this issue only needs to adress the part where lines moves when it shouldn't move.

PR #17772, opened 2025-05-20
[#17616] Decreased sensitivity of the way the system determines line connection type

a22nevan:
With the way the entire system for drawing lines and determining the connection type between them, it is essentially impossible to implement logic to "fix" the line moving issue described. To do that a large amount of code would have to be rewritten to accommodate a better way of determining it, as it's currently determined by the middle point between two elements instead of the lines position relative to the elements. This code also affects all lines which could cause a lot of issues if all lines aren't accounted for.

In order to somewhat alleviate the problem presented, the leniency of the determining system has been increased. There is now a dead zone which retains the current connection type until the border of the opposite type is crossed. This is as mentioned not a complete fix to the issue and is admittedly more of a band-aid fix than anything, but it should be less eccentric when moving an element in a diagram. This has however made it so that if you manage to move it to the wrong side, it takes quite a bit more movement to get it back to the original connection type. It's an unfortunate side effect of this but all in all it should be easier to move elements and lines around now without accidentally changing what side the line connects to.

**How it works in the diagram shown in the issue description (still kinda janky, but a bit more flexible):**

https://github.com/user-attachments/assets/19e73787-99dd-432d-9a13-a63807ccfa78

**Demonstration between two SD elements:**

https://github.com/user-attachments/assets/4d1fea37-cbdc-4a99-b02f-7acd02d83c8f

### DIFF
--- a/DuggaSys/diagram/draw/line.js
+++ b/DuggaSys/diagram/draw/line.js
@@ -1244,13 +1244,18 @@ function determineLine(line, targetGhost = false) {
     if (felem.y1 > telem.y2 || felem.y2 < telem.y1) overlapY = false;
     let overlapX = true;
     if (felem.x1 > telem.x2 || felem.x2 < telem.x1) overlapX = false;
-    let majorX = true;
-    if (Math.abs(line.dy) > Math.abs(line.dx)) majorX = false;
+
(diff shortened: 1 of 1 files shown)


a23aleva (approved):
The line connections now feel less janky and the deadzone applied seems to help quite a bit with creating a smoother experience when moving line-attached elements around. The developer mentions that this isn't a complete fix but regardless I can't find any reason not to approve these changes. I endorse this fix!

https://github.com/user-attachments/assets/f7d97d6d-24d1-4646-baac-088d203b4440

PR #17773, opened 2025-05-20
G1 2025 v7 17559 arrows are wrong in state diagram

a23thefo:
You are now able to see the arrows properly and they align with the line

image

### Known issue
This is a known issue that was here before i started and unfortunately has stayed I would love to try and fix it after the weekly merge
image


b23ebbst (approved):
The arrows align with the lines, and look really good!

image

There is an issue, as said in this PR, that still remains (and is known) so I think it is a good idea to continue working on this after the merge! This is the remaining issue: 

image

But, all in all, really well done! I approve!

Issue #17739, opened 2025-05-20
The "navbar" icon for the mobile version appears in the upper-left corner.


b23danhe:
The "navbar" icon for the mobile version appears in the upper-left corner when the browser window is resized.

However, it should only be visible when the window size falls within the defined limits for the mobile view.

https://github.com/user-attachments/assets/b8d02820-123d-4a10-afce-579761641daa

<img width="443" alt="Image" src="https://github.com/user-attachments/assets/0a09e235-4bcb-41bd-b889-b0ae93125edc" />

PR #17791, opened 2025-05-21
Removed visible icon when over the mobile width limit

a23kevre:
Added a media query that had the width values inbetween mobile and desktop and displayed it as none to remove its existance. Make sure to check issue #17739 for reference.


a23emijo (approved):
Works perfectly, I approve! 👍

PR #17792, opened 2025-05-21
Removed the mobile view of the rectangular fab button

a23yutta:
The small (rectangular) FAB button menu in mobile view should now function and appear the same as in the desktop view.

To achieve this, the mobile version of the rectangular FAB menu was removed. Additionally, the round FAB button was also removed from the mobile view, as it caused item text to display incorrectly in the rectangular FAB menu.

The default window size for mobile testing is 320x600.
##
### Info about the round FAB menu
The round FAB button will be completely removed from the mobile version in a separate issue, as discussed with the group leader. Furthermore, not all menu items were visible when using the round FAB button.

### DIFF
--- a/DuggaSys/sectioned.php
+++ b/DuggaSys/sectioned.php
@@ -218,7 +218,7 @@
 			</div>
 		</div>
 
-		<!-- + button --->
+		<!-- + button -->
 
 
 		<div id='Sectionlist'>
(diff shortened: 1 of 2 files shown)


a23lisjo (approved):
**_Copy from last review_** 

(I can not create the problem in Chrome or Edge but in Firefox)
The button is now working in mobile-view. If both the rectangular and the circular FAB is toggled on, their menus are overlapping, but like the description of this pull request, the circular FAB will be removed so this is fine!

PR #17793, opened 2025-05-21
G3-2025-V8-bugfix-(mobile view)-bug-darkmode-text-hard-to-read-when-editing-course-visibility-17576# (V2)

c22isako:
**NOTE: This was originally https://github.com/HGustavs/LenaSYS/pull/17763.**

### Changes made:
- Added styling regarding placeholders - making sure that github URL and Key text doesn't fall back to default (lightmode)
- Added specialized styling for the dropdown menu, since other files (like the contribution files from what I've gathered) might use the general select-styling within blackTheme.css.
  - Made sure that the dropdown menu selected item is the same color as other inserted items 
  _(such as course name and course code within image1 below.)_

Note that these change has affected the desktop dark mode view as well.

image
_(image1, displaying old (left) and updated (right) edit menu)_

---

### Possible errors present within both groupbranch and this PR:
There seems to appear a white background behind the Github URL when inserted (and valid) when testing on chrome _(image2.)_ I'm unsure if this is a me-problem, since I seem to have the same issue on the group-branch. 
- If possible, please confirm if the issue exists.

image
_(image2)_

---
### How to get to the functionality:

enter courseed.php (main menu) -> login as brom -> click the cogwheel _(image3)_ to edit course of your choice (ex. Datorns grunder) and press the visibility dropdown. 
image
_(image3)_

-> This will lead you to one of the edit menus in _image1_


a23yutta (approved):
The text visibility for the visibility dropdown menu, as well as the Github URL and Key, is now improved. It does not break any existing features, but a few notes are to be taken into consideration: 

- I wasn't able to replicate  the issue where  the Github URL and Key text reverted to the default (light mode) before the change. However, I do notice a slight change in the text color, it appears to be a lighter gray now.

- I'm having a hard time figuring out what "contribution files" you're referring to and how they relate to the visibility drop down menu. 

##

### Regarding the White Background Appearing for the Github URL
The issue with the white background isn't only affecting the Github URL but also the Course name and code, and likely the Github Key in both the group branch and this PR. This behaviour only occurs when selecting a previous stored value on the page, not when manually typing or pasting the text. Taking this into consideration, this PR should still be clear to merge. For instance:

Skärmbild 2025-05-21 143618

Skärmbild 2025-05-21 143631
## Example of Testing

### Before Change:
Skärmbild 2025-05-21 141336

### After Change: 
Skärmbild 2025-05-21 141504

## Other Issues to be Addressed not related to this PR
There are several significant bugs when attempting to save a course after making modifications:
- The course doesn't save at all for existing courses(default) but saves for your own created course sometimes
- The page occasionally freezes during the save process, requiring a manual refresh.
- Saving takes an unusually long time.
- Updating the Github URL sometimes results in it being set to "undefined"

Additionally, the page layout is not responsive and designed for window sizes 320x600, manual scrolling or resizing the window is required to view the entire page.

Good job!

[APPROVED]

PR #17800, opened 2025-05-21
Fixed the pop up messages (er toggle, error message and m.m.) are hidden

a23emijo:
Increased the z-index to 5000 to make the pop up message appear above the filename box.

image


a23emisn (approved):
Clean and effective fix! A z-index of 5000 might be a bit excessive, but it solves the issue without any apparent side effects. Easily adjustable if stacking issues arise in the future too.

Approved!

PR #17801, opened 2025-05-21
Converted jquery in \DuggaSys\templates\dugga2.js #17530 Rebased

f23linan:
Converted jquery to JavaScript.

Removed 2 functions that served no function as well.

REBASED 
Already reviewed in #17716

### DIFF
--- a/DuggaSys/templates/dugga2.js
+++ b/DuggaSys/templates/dugga2.js
@@ -269,12 +269,12 @@ function hexClick(divid)
 	const dpos = elem.getBoundingClientRect(); 
 	const dwid = elem.offsetWidth;
 	const dhei = elem.offsetHeight;
-	bw=Math.round(dwid)*1.50;
-	if(bw<180) bw=180;	// Ensure minimum width of the HEX-box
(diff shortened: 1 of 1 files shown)


c22isako (changes requested):
**Everything seems to work as expected, but for one thing.** 
- It seems that the toggle for instructions was used, as it is toggle-able within the group branch.

From current group-branch:

https://github.com/user-attachments/assets/08ad22e2-0889-44ce-b45f-f7d463643f7d

- I know that the feedback functions were discovered to be unused - so I'm unsure if its to be removed within this issue, or further along within an issue specifically targeted for feedback-removal. That said, it's unused, and a small issue. Confirm with issue coordinator just in case.

**Other than that, everything looks good!**
The issue from the last PR is fixed as I cannot replicate the bug from the earlier change request. Good work!

---

Note: It seems that save/reset/load dugga -bar is in the middle within the group branch, but at the left side within this PR. I'm unsure if this is because of new functionality introduced within group branch which isn't present in this PR, or an issue. _Just a headsup!_
image

PR #17802, opened 2025-05-21
Dark mode now works for load dugga ##17630

a23erigu:
Dark mode is now functional on validateHash.php 
You can't press the darkmode button in validateHash.php without closing the load window but it those work now.
It's easiest to test if you change to dark mode before trying to load a dugga.

picture of how it looks now
image


a23elism (approved):
After testing on multiple different duggor, i can confirm that darkmode now does work on the load dugga window (be it validateHash.php or simply pressing the load dugga button as brom)
---
### Well done, Approved for merge!

### Stitch approved 👍
Happy Stitch

PR #17803, opened 2025-05-21
Convert JQuery to vanilla JS in shapes-dugga.js

a23lisjo:
The page is empty or not really functional so this is how the page looked like before I started and after I changed JQuery. 
image

The only change is that something  in the dugga has to be changed for the user to be able to save and reset the dugga. 
image


f23linan (changes requested):
Works as intended! Howeverjust a small line of Jquery that was missed.

On line 75:

"dta=jQuery.parseJSON(data['param']);"

After that quick change it will be clear for merge! :)

f23linan (approved):
All the changes requested have been made and the functionality (which there wasnt a lot to begin with) is still properly working. Good job, approved!

PR #17804, opened 2025-05-21
Replaced jQuery with vanilla JS in html_css_dugga.js #17540

a23alera:
The feedback toggle was difficult to test since I couldn't find where it's triggered in the interface. However, the logic and CSS implementation should be correct based on the code.

### DIFF
--- a/DuggaSys/templates/html_css_dugga.js
+++ b/DuggaSys/templates/html_css_dugga.js
@@ -69,7 +69,7 @@ function setup()
 function returnedDugga(data)
 {
 
-	$("#content").css({"position":"relative"});
+	document.getElementById("content").style.position = "relative";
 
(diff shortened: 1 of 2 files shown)


a23simry:
Feedback functionality should be in the future removed so its not a huge issue to test that, as long as the rest works and nothing breaks everything is fine.

f23linan (approved):
Looks good! The only thing was to maybe move the added CSS to where the other duggors CSS is (If they have any which they probably do), but I dont think its enough to warrant a                                                             denial. Unless the issuemaster think otherwise I think its okay for merge!

a23simry:
Personally I think style.css is already a mess so adding the extra code at the end wont do much, it will be a future issue to remove all uses and mentions of feedback.

PR #17805, opened 2025-05-21
Changed the background color of burger drop down on mobile #17495 V2

a23erigu:
I think this is what the issue wanted.
Only added the hamburgerBox tag to the dark mode file and gave it the correct color

pic of how the dropdown looks in dark mode now.
image


a23alera (approved):
Everything works as expected and matches the issue description.

PR #17812, opened 2025-05-21
Fixed: Implement the mobile tooltip from mockup

a23gabst1:
I have implemented the mobile tooltip and have made so that when the user holds in the toolbar modes or the replay buttons, the tooltips appear on the top right corner of the screen with reasons to cause as low intereference to the modeling. see video 1 for illustration on how it works. If the tooltips should work on the option panel button as well, just add the mb-tooltip class and an if-statement on `tooltipPosition()`. 

video 1

https://github.com/user-attachments/assets/7ba47946-069c-490b-9a20-9ec735d31999

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -2554,6 +2554,7 @@ elementToolbarBoxs.forEach(elementBox=>{
     elementBox.addEventListener("click", handleToolbarClick);
 });
 
+const allToolbarBox = document.querySelectorAll(".mb-tooltip");
 
 // Select all toolbar boxes inside the sub menu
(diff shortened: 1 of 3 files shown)


a23eliax (approved):
This works as expected! It's very nice for the user experience that you just can hold on a button and then the tooltip comes up in the corner, and disappears when you let go of the button. I also noticed that you removed the keybinding-text for the tooltip in the mobile verison, which is good because it had been unnecessay since the user can't press that keys on the phone anyways. Well done!

https://github.com/user-attachments/assets/5ccbbfb6-d08f-4b47-8ffa-4d36337e2121

PR #17818, opened 2025-05-21
Convert jquery in \DuggaSys\templates\generic_dugga_file_receive.js #17538

a22carsv:
Converted JQuery for 'generic_dugga_file_recieve.js', there should be none remaining except for AJAX related. Feedback has been addressed, this is only a new PR for rebase.
PR: #17729 
Issue: #17538

### DIFF
--- a/DuggaSys/templates/generic_dugga_file_receive.js
+++ b/DuggaSys/templates/generic_dugga_file_receive.js
@@ -289,13 +289,18 @@ function closeFacit()
 //                                  Local Functions
 //--------------------================############================--------------------
 
+
+//--------------------================############================--------------------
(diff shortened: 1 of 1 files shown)


EdvinWerner (approved):
Nothing that breaks as far as I can test at least, approved for merge!

a23fanro:
Everything looks good except this part (line 322): 
image
\t is a tab that sometimes is used in tags, but it should not be used when doing css-styling. Worst case scenario it that it can break the code. Please fix this: then I will merge it!

a22carsv:
Okay, I have removed the use of \t for line 322 @a23fanro.

PR #17819, opened 2025-05-21
 Convert jquery in \DuggaSys\templates\default.js #17524

a22carsv:
Converted the JQuery in default.js, I actually dont know where this is used or even if it is used in the current rendition of LenaSYS. So for testing purposes, I assume it might suffice to ensure that nothing breaks as a result of these changes. This PR solves issue https://github.com/HGustavs/LenaSYS/issues/17524. Changes from last weeks branch is that it has been rebased.
Previous PR: #17750 
Issue: #17524

### DIFF
--- a/DuggaSys/templates/default.js
+++ b/DuggaSys/templates/default.js
@@ -20,24 +20,16 @@ function loadQuestion(alternativ, question){
 	app += "<br>";
 	app += "<section class='bottom'><button class='submit' onclick='checkAnswer();'>Submit</button></section>";
 
-	$("#content").append(app);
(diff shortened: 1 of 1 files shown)


b20isato (approved):
I really do not find a way to actually test it since i can' t access it and really do not wanna dig through the possible files that could lead to this being used if it even is used. 

Even making a new dugga doesn't result in the default, there is no way to select default as far that i see. But also the site and other duggas hasn't broken and it's only a JQuery -> Vanilla Javascript so doubt it will become a problem

Issue #17663, opened 2025-05-16
Showdugga has lost functionality, missing title, missing components, pls patch


a23simry:
### Some investigations has been made on the master branch to find where this stems from.

 - The commit where the functionality seemingly broke: https://github.com/HGustavs/LenaSYS/commit/2cf51378a66c113ba298f3800b8e1377a3369a2d

    Also a picture showing how it looks now, see also the error message:
    ![Image](https://github.com/user-attachments/assets/38acfa5d-705c-4725-8c68-b50d6f6fedd8)

- This is the commit right before which still has the functionality: https://github.com/HGustavs/LenaSYS/commit/fb7d8870322f40969d6f1091f0767677b09767a6

    Also a picture showing how it looked then and **should** look now, see also no error message:
    ![Image](https://github.com/user-attachments/assets/e8955200-859f-45b0-9f19-463162eff5ab)

The issue most definitely comes from somewhere between these two commits!

PR #17821, opened 2025-05-21
Showdugga has lost functionality, missing title, missing components #17663

a23elism:
### This is an updated version of #17762, read this PR and the Issue for any information


a23elism:
Everything should be fixed, for testing simply try to use any dugga, save and load it again
---
### This was mainly tested in linjedugga 1 & 2 in webbutveckling - datorgrafik so i recommend starting there

a22maxda (approved):
Having tested most duggas, I haven't found any problems. Some duggas throw errors related to their own code but that is separate from this issue. It seems that the fixes work and haven't created any new issues.
Well done, Approved!

Issue #17695, opened 2025-05-19
Real-time typing in elements


b23ebbst:
Don't know when this disappeared, but now when the user inputs anything in an element's input-fields (name, attribute, function, etc.), the text isn't being shown in the element until the user presses enter or save. 
Previously the element would show, in real-time, the letters being typed out even before the user pressed save/enter.  

https://github.com/user-attachments/assets/80d3b627-3583-4393-83ac-402fca038617

a23miksk:
This got reintroduced https://github.com/HGustavs/LenaSYS/pull/17655 here, a23matsh seems to have implemented some "blur event" that doesn't seem to anything new. But maybe it can be used to make the real-time type work AND have good automatic saves of the text (and not individual letters).

PR #17823, opened 2025-05-21
Fixed G1-2025-v8-#17695-real-time-typing-in-elements

a23eliax:
The real-time typing for the elements is back, AND the undo/redo works. In other words; the undo/redo doesn't remove one letter at the time, it removes the latest save or when the user presses "enter". Also the user can se the typing directly in the element. I had to remove the stateMachine.save in the function saveProperties() and only have it in the blur event.

https://github.com/user-attachments/assets/a6813a5c-93d7-4236-bd62-1a531035c0f6

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1497,7 +1497,6 @@ function saveProperties() {
         addToLine("functions", "+");
     }
     // Saves the changes and updates relevant graphics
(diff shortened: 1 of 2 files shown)


a23kevre (changes requested):
Fixed the issue that was given. The real time text is working wonderfully but i simply can't approve these changes since after a press of the save button all text that was written dissappears and the element/entity resets itself to default through a lot of automated revert changes.

PR #17824, opened 2025-05-21
Added mobile styling to code

### LINKED ISSUE #17764: (mobile) Fix did not fix duggs in all courses
Since PR https://github.com/HGustavs/LenaSYS/pull/17662 for issue https://github.com/HGustavs/LenaSYS/issues/17383 there is a nice implementation of the visibility of the duggs, they now have two lines instead of one.

However this does not seem to be applied to all courses.

WHile testing, demo-course has the new implementation, while testing-course does not.

We want the new implementation to be applied to all courses.

Testing-course (old and what we do **not** want):

Image

Demo-course (new and what we want):

Image

a22maxda:
Added styling for code examples on mobile. Only applies to mobile screens and while logged in. Being on desktop or being logged out shows normal style.

### DIFF
--- a/DuggaSys/sectioned.js
+++ b/DuggaSys/sectioned.js
@@ -1798,6 +1798,22 @@ function returnedSection(data) {
             str += "</td>";
           }
         }
+
+        //Mobile view for code examples
+        if (itemKind === 2 && (data['writeaccess'] || data['studentteacher'])){
+           var param = {
+            'exampleid': item['link'],
+            'courseid': querystring['courseid'],
+            'coursename': querystring['coursename'],
+            'cvers': querystring['coursevers'],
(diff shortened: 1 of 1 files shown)


a22maxda:
Code examples exist in courses like Testing-Course, these are different from duggas which present in Demo-Course among others.

Code examples
Skärmdump 2025-05-21, 15 58 01

Duggas
Skärmdump 2025-05-21, 15 59 26

To verify this works, shrink the screen to 320px width (smallest accepted screen size) and check that full names are visible.

a23erigu (approved):
This seems to be fully functional.
All the button still do what they are supposed to and it's easier to see and interact with duggar.

Nice work

PR #17825, opened 2025-05-21
Fixed stereotype length issue#17565 stereotype

a23matsh:
Since a 10-character limit on stereotypes would feel too restrictive from the customer’s perspective the maximum length has now been extended to 30 characters. While it may be unlikely that such long stereotypes will be used frequently, it is still  beneficial to support them to ensure flexibility and avoid limitations for those cases that it may occur. The goal was to make this increase possible without breaking the layout such like cut offs, overflow or text continuing in a new row.

What has been done:

- Implemented logic to dynamically shrink stereotype text when it exceeds 20 characters

- Used lengthAdjust and textLength to fit long text within the box without unnatural behavior occuring.

- Preserved default size and readability for shorter texts (i.e ≤20 characters)

Down below you see images of different lengths and how they layout.

image
image
image

### DIFF
--- a/DuggaSys/diagram/draw/element.js
+++ b/DuggaSys/diagram/draw/element.js
@@ -363,7 +363,20 @@ function drawElementUMLEntity(element, boxw, boxh, linew, texth) {
 
     // Conditionally display stereotype above name if one is set
     if (element.stereotype != "" && element.stereotype != null) {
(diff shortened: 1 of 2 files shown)


a23gabst1 (approved):
This fix works as intended! Text under 20 characters stays the same and text over shrink making it still somewhat readable for users! The code looks great and has great comments! I approve these changes!

Normal length of an stereotype name
image

Longer than normal
image

PR #17829, opened 2025-05-21
Fixed: Fix the toolbars z-value and position in mobile

a23gabst1:
I have now made so the top menu sidebar has the highest priority when it is open. So, all the other elements are either placed under it or hidden. The fab button was easy to fix because you needed to change its z-index to something lower than the sidebar, however that solution didnt work with the other to buttons (general information and instructions). Because when you tried to lower the z-index on them it disappeared and was behind the iframe, then the solution was to access the elements from the iframe and make buttons disappear when the toggle icon sidebar was clicked, and for every click you check if the sidebar is open or not. If it is open then hide the buttons, and if not display them only after a delay to make it more smoothly and less snappy. 

see video 1 

video 1

https://github.com/user-attachments/assets/91f1a1c7-e3c2-4bf8-a7fd-5302140dbb8a

### DIFF
--- a/DuggaSys/templates/diagram_dugga.html
+++ b/DuggaSys/templates/diagram_dugga.html
@@ -307,6 +307,35 @@
             if(!infoContainer) return; //Null check
             infoContainer.classList.remove("show");
         }
+
+        const iframe = document.getElementById("diagram-iframe");
+        //Waits for the whole iframe to load in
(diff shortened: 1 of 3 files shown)


a23thefo (approved):
Works like just like the video provided in the pr and the code is well documented and following guidelines! 👍

PR #17830, opened 2025-05-21
Created test for updateCodeExampleTemplate_ms.php

### LINKED ISSUE #17807: create a test for codeviewerService/updateCodeExampleTemplate_ms.php

a23bensh:
Fixes #17807 and adds testing for updateCodeExampleTemplate_ms.php following the same standard as the other tests in tests/microservices.

To test it, go into tests/microservices/codeviewerService/updateCodeExampleTemplate_ms_test.php and see if it passes all the tests, and if it does, I assume documentation will need to be added in "codeviewerService_documentation.md" (in another issue?) that the service passed the tests.

### DIFF
--- a/DuggaSys/tests/microservices/codeviewerService/updateCodeExampleTemplate_ms_test.php
+++ b/DuggaSys/tests/microservices/codeviewerService/updateCodeExampleTemplate_ms_test.php
@@ -1,0 +1,37 @@
+<?php
+
+include_once "../../../../Shared/test.php";
+
+$testsData = array(
+    'updateCodeExampleTemplate' => array(
+        'expected-output' => '{"status":"success"}',
+        
+        'query-before-test-1' => "INSERT INTO codeexample (exampleid, examplename, cid, uid, cversion, templateid) VALUES (999995, 'TestCodeExample', 1885, 101, 1337, 5)",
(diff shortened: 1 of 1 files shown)


a23bensh:
I won't change this while someone is still reviewing but I missed putting _ms at the end of the 'updateCodeExampleTemplate' title. I'll add this after it's been reviewed.

d22marwi:
@a23bensh Every test passed except the last: 'Test 3 (assertEqual): failed with error: no valid values to compare.' I’m seeing the same error in my test issue, and I noticed that many other tests have the same error. Is it supposed to be like that? Sorry — this is my first test issue and PR, so I don’t fully understand yet

a23bensh:
@d22marwi it's my first one aswell and I'm working one a second one aswell at the moment. When looking at previous tests and documentation of these tests being approved/passed from other students, they also seem to fail on the 3rd test. That would mean both the old tests would need to be updated, and you would need to add it to the new tests. 

@a20artje you created the issues, do you know if the already created tests are outdated? Or is it supposed to fail on the third test??

a20artje:
The 3rd tests compares the data you give it to the data outputed by the microservice I don't know if this microservice outputs something. 

The test handler test 3 is pretty bad since you have to hardcode expected output.

Try calling the service from a command line using curl and see what the output is. You can usually do something like: 

curl -X POST http://localhost/LenaSYS/DuggaSys/microservices/codeviewerService/updateCodeExampleTemplate_ms.php 

but you would need to give it some data

d22marwi (approved):
I approve this for merge

Issue #17806, opened 2025-05-21
Create a test for codeviewerService/retrieveCodeviewerService_ms.php


a20artje:
Create a test following the same standard as the other tests in tests/microservices

PR #17833, opened 2025-05-21
Createed a test for codeviewerService/retrieveCodeviewerService_ms.php #17806

a22linre:
Made a test file for retrieveCodeviewerService_ms.php 

#17806


a23axehe (changes requested):
image

Test failed on step 3. To test I browsed to http://localhost/lenasys/duggasys/tests/microservices/codeviewerService/retrieveCodeviewerService_ms_test.php and the tests were run.

a22linre:
Maybe the issue isn’t with the test file but with retrieveCodeviewerService_ms.php itself, it seems to be returning an empty response? Do I have to fix the problem here?

a22linre:
It looks like all the tests in codeviewerService, they all fail at test 3

a22linre:
The test file for retrieveCodeviewerService_ms should work now but the other tests still failed at step 3 in codeviewerService

a23axehe (approved):
image

Now with the changes all tests passed. Looks good

PR #17836, opened 2025-05-21
Updated diagram.css

d22abdha:
Increased the margin of the diagram message from 5px to 20px so that the alert now sits above the zoom buttons.

Before:
image

After:
image

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -604,7 +604,7 @@
 
 #diagram-message>div {
     pointer-events: auto;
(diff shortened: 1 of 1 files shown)


a23thefo (changes requested):
You have pushed it up well and that looks good. However moving the error message to the right is unnecessary and looks bad move it back to the left so it aligns with the zoom like it did before but keep the height!

a23thefo (approved):
It has now been fixed and i approve
image

PR #17838, opened 2025-05-21
 COMPLEMENTS PR #17723: G3 2025 v7 #17532 convert jquery to js in dugga4.js

a23yutta:
This PR complements PR [#17723](https://github.com/HGustavs/LenaSYS/pull/17723#event-17737125675) (which I reviewed), as I realized there were some remaining jQuery I initially missed since it was not the usual `$` sign syntax. I noticed this while working on PR #17527, which involves converting jQuery to JS in the `dugga1.js ` file - a file with a similar structure to dugga4.js.


f23linan (approved):
Tiny but necessary change, nothing breaks so good to merge!

Issue #17787, opened 2025-05-21
Remove the nav bar in mobile


a23miksk:
Another group implemented the nav bar for mobile across the site which means that we now have two nav bars in diagram. 
So: hide the other groups nav bar in mobile mode

_Image of the other groups mobile nav bar_
![Image](https://github.com/user-attachments/assets/82470bf0-cd65-48a0-88e5-7abcb9e38787)

PR #17841, opened 2025-05-21
[#17787] Removed showDugga navigation header bar in mobile view

a22nevan:
**For reviewer:** You need to be in showDugga.php for this (_/DuggaSys/courseed.php -> "Demo Course" -> "Diagram Dugga"_).

The navheader now gets hidden if the system detects that the chosen dugga is the Diagram Dugga and the user is in the mobile view. The current way this is implemented is in the same way the mobile view is implemented in diagram.js (checking if the innerWidth is less than 414), if this were to change it should probably be changed here as well.

**Mobile view:**

image

**Regular view:**

image

### DIFF
--- a/DuggaSys/showDugga.php
+++ b/DuggaSys/showDugga.php
@@ -216,6 +216,20 @@ if(!isset($_SESSION["submission-$cid-$vers-$duggaid-$moment"])){
 		
 		?>
 	</div>
+
+	<script>
+		// Hides the navheader if the chosen dugga is the diagram dugga since it has its own "header"
+		const hideNavIfDiagramMobile = () => {
+			const temp = '<?php echo $duggafile; ?>';
+			let nav = document.getElementsByClassName("navheader")[0];
+
+			if (window.innerWidth < 414 && temp === 'diagram_dugga') nav.style.display = "none";
+			else nav.style.display = "flex"; // Mainly if the window is resized after having loaded in to get the bar back
+		}
+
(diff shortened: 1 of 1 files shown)


a23vinge (approved):
As described by the developer, the navheader is now conditionally hidden when the selected dugga is the Diagram Dugga and the window width indicates a mobile view (window.innerWidth < 414). This is a sensible and clean solution, consistent with how mobile mode is handled elsewhere (e.g., diagram.js).

The implementation is well-commented, which provides valuable context for current and future developers.

https://github.com/user-attachments/assets/b4ee12e6-5d2b-4df8-9ba7-e63dab5111bd

As shown in the attached video, the feature works as intended — the header is removed in mobile view, and no visual bugs are introduced.

(This in some sense have no direct connetion to the issue)
That said, I feel it's important to point out a potential UX issue introduced by this change:

By hiding the navheader, the familiar hamburger menu (visible in the original design, see attached image) is no longer available. This menu previously offered a clear and intuitive way for users to return to the course or navigate away from the dugga.

In the updated version, navigation relies on an arrow icon that resembles a “scroll” or “toolbar expansion” button, not a return-to-menu control. For users unfamiliar with the interface, this could cause confusion or frustration, as the exit path is no longer immediately obvious.

I approve the merge request — the implementation is solid and meets the requirements.
However, it might be worth considering a follow-up improvement to clarify the exit/navigation UI for mobile users.

before:

image

after:

https://github.com/user-attachments/assets/b8663ca0-e792-4051-812c-4c2ced89eaea

Issue #17790, opened 2025-05-21
Arrows no longer show up on sequence lines


a23aleva:
Since the latest batch of patches arrows have suddenly stopped to appear when applying them to SE (sequence) lines. The user should be able to use arrows when drawing lines between activations in a sequence diagram so this needs to be fixed.

https://github.com/user-attachments/assets/8bbd68e1-e12b-4af9-ac43-8ede422e5593

PR #17843, opened 2025-05-22
Fixed: #17790 Arrows no longer show up on sequence lines

a23aleva:
Arrows should now be visible again when using SE lines between sequenceActivity elements. Had to update icon draw logic with an extra check condition for straight lines since SE and SD are situated within the same code block. Also updated the iconX/YModifier variable to not start as null which caused errors. Since this whole if-statement wasn't properly indented I also enhanced its readability.

https://github.com/user-attachments/assets/ae2beec8-4c48-4d11-825f-97582ad41a6f

### DIFF
--- a/DuggaSys/diagram/draw/line.js
+++ b/DuggaSys/diagram/draw/line.js
@@ -12,8 +12,9 @@ function drawLine(line, targetGhost = false) {
     let fromElemMouseY;
     let toElemMouseY;
 
-    //For straight SD lines only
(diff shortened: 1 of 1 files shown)


a22nevan (approved):
Solution works just as described. Arrows now show up on lines between sequence activation elements. 

**Before:**
image

**After:**
image

Code is also easier to read after the indentation changes. This PR can be merged without problems. Good job 👍

Issue #17848, opened 2025-05-22
Codeviewer, darkmode buttons not clearly visable


f23linan:
When in darkmode a few of the icons when youre editing the codeblock are not clearly visable. You need to be in admin access to be able to edit the code!

![Image](https://github.com/user-attachments/assets/23fff3d3-a258-4109-874d-20355ea93846)
Lightmode

![Image](https://github.com/user-attachments/assets/8fe71403-dfa9-4a17-bfa9-6f16aa408416)
Darkmode

a23yutta:
During testing, it seems like all buttons in dark mode are not clearly visible to me, as they appear black and blend into the background. 

@f23linan Do you have the same issue of all buttons displaying unclearly? I want to make sure it's not just a ME problem before I actually do any changes.

### Light mode:

![Image](https://github.com/user-attachments/assets/ddec1c80-efe9-4386-91dc-79ab89329593)

### Dark mode:

![Image](https://github.com/user-attachments/assets/69f34f9d-9fcf-40f5-aac3-bbd462a62be7)

f23linan:
Sorry for late reply! But yes it looks like yours does!

Issue #17455, opened 2025-05-12
(desktop-view) when editing a codeblock in codeviewer its not in darkmode


f23linan:
When you go to edit a codeblock in codeviwer one half of the edit page is not in darkmode.

![Image](https://github.com/user-attachments/assets/d052682d-9d44-4811-98c0-11a6dcc58049)

Go to the edit page through this button.

![Image](https://github.com/user-attachments/assets/cc8df6bc-4e81-4b10-9ae0-ba52cb9220f5)

Not in darkmode.

PR #17849, opened 2025-05-22
Editing a codeblock is not in darkmode #17455

f23linan:
Fixed the darkmode of the codeblock when editing in codeviewer.

image
Before

image
After

This did however add a new problem where some of the icons arent clearly visable in darkmode. Ive already made a new issue for that [here](https://github.com/HGustavs/LenaSYS/issues/17848)

### DIFF
--- a/Shared/css/blackTheme.css
+++ b/Shared/css/blackTheme.css
@@ -565,7 +565,9 @@ body {
 
 .loginBox,
 .logoutBox,
-.formBox{
+.formBox,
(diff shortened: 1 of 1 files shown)


b23sebth (approved):
**Fixes [17455](https://github.com/HGustavs/LenaSYS/issues/17455) - approved!**

## What's changed:
- Darkmode is now displayed correctly.

Works and looks good!

PR #17852, opened 2025-05-22
Replaced jQuery with JS in dugga1.js

a23yutta:
Replaced jQuery with JS in `/DuggaSys/templates/dugga1.js`. 

Additionally, the toggle feedback and show/close facit features has not been introduced to the page yet, making it impossible to verify whether the JS conversion works as intended. Related functions: `toggleFeedback()`,  `showFacit()`, `closeFacit()`

The `toggleInstructions()` and `toggleFeedback()`  functions in the end of the file have not been converted as there is no direct JavaScript equivalent for jQuery's `slideToggle()`  function, replacing it would require a custom implementation. This is too big part this issue and should therefore be addressed in a separate issue.

### DIFF
--- a/DuggaSys/templates/dugga1.js
+++ b/DuggaSys/templates/dugga1.js
@@ -265,16 +265,16 @@ function closeFacit(){
 function bitClick(divid) {
 	if (typeof ClickCounter != 'undefined') ClickCounter.onClick();
 
-	const divid = document.getElementById(divid);
+	const el = document.getElementById(divid);
 
-	if (divid.innerHTML === "1") {
-		divid.innerHTML = "0";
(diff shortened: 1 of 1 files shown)


a23lisjo (approved):
The code and functions is working as intended during testing. Like the author describes the only jQuery left is `toggleInstructions()` and `toggleFeedback()` at the bottom of the code file. The author also mentioned,  that the functions `showFacit()` and `closeFacit()` has no implementation and therefore have not been able to be tested. 

Still the code looks correct and should not be a problem to merge!

a23fanro:
All code related to feedback is going to be removed from the system, this is a separate issue. Facit is a discussion that needs to be had, good catch!

Issue #17620, opened 2025-05-15
alert NONE be gone


a23fanro:
there is an alert that pops up whenever we create a new item with the the fabBt, we want the alert removed.

![Image](https://github.com/user-attachments/assets/bf2b49d5-b0da-4adb-aff5-6c5525850b16)

This is the alert:
![Image](https://github.com/user-attachments/assets/460ae04c-116f-4a0f-890a-c24177db02da)

We suspect that it has something to do with one of the various returned*(data) functions in `sectioned.js`, but the issue can be somewhere else. Find out, remove it.
Examples of functions that we suspect might be causing this:
![Image](https://github.com/user-attachments/assets/5ebd4faf-efb5-4961-b7da-0f8a0a2e3de1)

![Image](https://github.com/user-attachments/assets/aa7bd574-7133-4951-8082-b902a2609722)

![Image](https://github.com/user-attachments/assets/351432ca-0553-4d7f-917d-6a631dcc9e1a)

The issue was found it PR: https://github.com/HGustavs/LenaSYS/pull/17299

PR #17859, opened 2025-05-22
got rid of the NONE alert in section.js #17620

a23erigu:
I got rid of the NONE alert that appears when trying to create a element


a23osckr (approved):
The NONE alert is gone and nothing is broken. Approved.

Issue #17832, opened 2025-05-21
mobile buttons overlap


a23fanro:
Since commit [dcd0ac93a39a5ac62cf61025dc079932bdab8755](https://github.com/HGustavs/LenaSYS/commit/e9229217ceba68cf2b872bf53b249e9e22e258d7) the buttons in courseHeader (logged in as user brom) overlap each other, specificly the view- and toggleElements-buttons. There might have been some styling deleted in the [commit](https://github.com/HGustavs/LenaSYS/commit/e9229217ceba68cf2b872bf53b249e9e22e258d7) that needs to be implemented again.

Make the buttons stop touching each other. It should work with a 320x600 mobile view.

Buttons touching:

![Image](https://github.com/user-attachments/assets/71908323-d7d5-4e80-8b1b-65996d39c38d)

Buttons not touching (what we want):

![Image](https://github.com/user-attachments/assets/6feeadbd-3ad9-4b64-b328-bd642f6358dc)

PR #17861, opened 2025-05-22
mobile buttons overlap #17832

a23elism:
This PR is connected to the relevant Issue #17832 
--
This PR also simply removed a "}" that broke the styling of the buttons relevant to the Issue


a23elism:
Testing is simply entering the site, and going to any course as brom and looking if the buttons are correct and not overlapping
---
image

b23sebth (changes requested):
There appears to still be some overlap on certain scren sizes.

### 352px width:
image

### 320px width:
image

## Change needed:
- Please fix the overlap on screen sizes >= 320px

Issue #17842, opened 2025-05-22
The tooltip overlaps the submenu


b23danhe:
I find it distracting when the tooltip overlaps the submenu *(figure 1)* while hovering over the main menu button. It would be better if the tooltip always appeared in the same position as shown in *figure 2* within each respective menu.

![Image](https://github.com/user-attachments/assets/49fc5ec0-6a28-4d0c-bd9c-3c8a9f3af82b)
*figure 1, hovering main menu: tooltip overlapping submenu*

![Image](https://github.com/user-attachments/assets/efebcb75-b87b-4dfd-b87d-85930c71e1ac)
*figure 2, hovering submenu: tooltip not overlapping submenu*

PR #17863, opened 2025-05-22
Fixed: Tooltip overlaps submenu

a23emisn:
This PR fixes issue: #17842

Previously, hovering over a toolbar button with a submenu caused the tooltip to overlap the submenu, demonstrated below:

image
_Tool tip overlaps the sub menu._ 

An if-statement has been added that repositions the tooltip-box position if a submenu is present, eliminating the overlap. Additionally, the margin differences were also adjusted and the tool tip box now sits just as flush to the submenu as to the tool bar. This is what the behavior looks like now:

image
_Overlap eliminated by the repositioned tool tip. Margin between tool tip and sub menu reduced (compare this with weekly branch)._

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -2698,7 +2698,7 @@ function tooltipPosition(element){
 
         return{
             top: `${dropdownPosition.top - dropdownPosition.height/2}px`,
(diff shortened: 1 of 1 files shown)


a23emijo (approved):
The implementation works like a charm. The hover-boxes is in a much nicer spot now compared to before. I approve! 😊👍

Issue #17629, opened 2025-05-15
(mobile-view) BUG darkmode codeviewer hard to read


a23osckr:
The darkmode codeviewer is hard to read with non contrasting text colors against the background. This was taken from a javascript example, but others are also not great. Perhaps trying to copy the colors from an IDE such as VS would be a good idea since their darkmode works well.

![Image](https://github.com/user-attachments/assets/f24d8017-7c52-4be1-b3d1-cd2d79ee3c3e)

PR #17866, opened 2025-05-22
G3-2025-V8-#17629 darkmode codeviewer hard to read

a23osckr:
I have changed some of the colors for the darkmode codeviewer since it was hard to read before. Lightmode has been slightly affected as well since I could not find a darkmode option for HTML, so I changed the colors to work better in darkmode while also being readable in lightmode. The lightmode comment color has been slightly darkened for better readability. The colors used for darkmode are colors usually used in IDEs when darkmode theme is used.

JS Before:
image

JS After:
image

HTML Before:
image

HTML After:
image

### DIFF
--- a/DuggaSys/codeviewer.js
+++ b/DuggaSys/codeviewer.js
@@ -2062,14 +2062,14 @@ function rendercode(codestring, boxid, wordlistid, boxfilename) {
 				switch(coloringcode) {
 					case "<html>":
 					case "</html":
-						fontcolor = "red";
+						fontcolor = "#fc2d2d";
 						break;
 					case "<link ":
 						fontcolor = "blue";
 						break;
 					case "<h1>":
 					case "</h1":
-						fontcolor = "darkorchid";
+						fontcolor = "#4f5aff";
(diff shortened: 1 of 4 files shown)


a23elism (approved):
Addition/changes to the colors inside the codeviewer have made the code alot more readable, whilst not being overwhelming to the eyes or simply "weird"  (it feels more like looking at an actual code editor now)
---
### Approved for merge 👍, Well Done! 

### Stitch Approved
Happy Stitch

a23fanro:
Resolve conflicts please

a23fanro:
Have you thought about creating a css-rule for the code that is only supposed to be affected when in dark mode? instead of changing it for both regular mode and dark mode.

a23osckr:
Yes I though about doing this, but figured I would just solve the immediate problem and leave it for a future issue. Only HTML and PHP had the issue of a shared light and dark theme, so I made the colors more appropriate for both themes on those two. JS and SQL have separate theme colorings, so this was not an issue with them.

PR #17867, opened 2025-05-22
Added: overflow-y: auto;

a23vinge:
The #diagram-toolbar element already contained enough content to require scrolling on smaller screens.
However, without overflow-y: auto, the browser didn’t allow scrolling, which caused the layout to break or clip icons.

With overflow-y: auto and height: 100vh, the browser now detects when the content overflows and automatically shows a scrollbar as needed.

https://github.com/user-attachments/assets/71869b0c-4d0a-43e1-95a1-698a4ffd4627

As demonstrated in the video, when the screen height becomes too small to fit all toolbar icons, vertical scrolling is enabled

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -26,6 +26,9 @@
 }
 
 @media screen and (min-width: 768px) {
+    /*
+    Make toolbar scrollable when content exceeds screen height.
+    */
     #diagram-toolbar {
         position: fixed;
         top: 0;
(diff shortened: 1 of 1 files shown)


a23eliax (changes requested):
This almost works as expected. The issue now is that the scrollbar takes up to much place and the whole toolbar isn't visible. It takes up more place when your screen is zoomed out a bit. I think the scrollbar has to be thinner, then I can approve! 

https://github.com/user-attachments/assets/3ad07a7d-d922-4aeb-860c-cf3fa067e18c

a23eliax (approved):
Now the issue is totally fixed! And very nice with the yellow color on the scrollbar, I approve!
image

PR #17873, opened 2025-05-22
Fixed: Make the showDugga buttons be behind ui elements in mobile

a23gabst1:
I added a function so it can be reused in the click listeners. The function checks if either the sidebar or the option panel is open, if it true then it hides the buttons, if not then it displays the buttons after a delay. I needed to add this function to both the option fab and the close btn because otherwise the buttons would stay hidden because i could not press the option fab since it is under the option panel when it is active. see video 1 on how the solution works. 

video 1

https://github.com/user-attachments/assets/d2e519ef-2a29-4e33-8fbc-f6a6b5c64037


a23eliax (approved):
This works very well and fixes the issue in a very smooth way, I approve!

Issue #17844, opened 2025-05-22
Refine the appearance of the "Save" and "Save As" buttons


b23danhe:
As shown in *figure 1*, a visible border appears, which should not be present. Additionally, should disabled icons have a hover effect? 
They could also appear greyed out to clearly indicate that they are inactive, similar to the appearance in *figure 2*.

![Image](https://github.com/user-attachments/assets/c2b59190-3450-4a4e-b9df-2977a172590e)
*figure 1, how the icons appear now*

![Image](https://github.com/user-attachments/assets/f5d9b063-41b7-4935-b044-419d0a963d09)
*figure 2, example of how a disabled icon can appear*

PR #17874, opened 2025-05-22
Fixed: G1-2025-v8-#17844 refine the appearance of the save and save as buttons

a23eliax:
Previously, users could hover over a disabled icon. This could give a misleading impression that the icon was clickable or active. With these changes; opacity and no hover, it is now clear that the icon is inactive and cannot be clicked. The message "You don't have anything to save!" still appears if the user presses Ctrl + S. Additionally, the border around the Save and Save As icons has been removed, making them more consistent with the rest of the icons.

image

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -658,11 +658,6 @@
     transition-delay: 700ms;
 }
 
-.disabledIcon:hover .toolTipText {
-    visibility: visible;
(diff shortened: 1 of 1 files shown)


a23gabst1 (approved):
This solution works smoothly! I really like how the save and save as buttons changes from an disable style to an enable style as soon as the user starts modeling! The styling looks fine and I approve these changes!

Issue #17498, opened 2025-05-13
(desktop-view) BUG darkmode splits instructions


a23erigu:
The instruction text gets split into two section when switching to dark mode.
This seems to be a problem for all duggor.

pic of light mode version
![Image](https://github.com/user-attachments/assets/a6844cc3-72b4-435c-bf42-068f06d3a042)

pic of dark mode version
![Image](https://github.com/user-attachments/assets/d964764f-e2cf-4c26-9ced-30a886993e96)

PR #17875, opened 2025-05-22
Fixed the bug in dark mode for instructions #17498

a23alera:
Instructions-content text now stays in one column


a23erigu (approved):
From what I can tell this those solve the issue.
Seems like a weirdly easy fix to the problem but it those work.

Nice work

PR #17876, opened 2025-05-22
Add button from microserviceUI to dependencyUI #17854

### LINKED ISSUE #17854: Add button from microserviceUI to dependencyUI
The user should be able to go from microserviceUI to dependencyUI

a23oskme:
Simply added a button on microserviceUI that takes the user to dependenciesUI.php. Fixes #17854

### DIFF
--- a/DuggaSys/microservices/endpointDirectory/microserviceUI.php
+++ b/DuggaSys/microservices/endpointDirectory/microserviceUI.php
@@ -185,6 +185,7 @@ if (isset($_GET['id'])) {
         </div>
         <div class="search-container">
             <div class="button-container">
+                <button onclick="openFilterModal()">Filter</button>
                 <form method="get">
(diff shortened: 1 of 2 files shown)


a23axehe (commented):
Button works perfectly,

image

I however think we should resize the button and the text to match the buttons to the left and right of it. See attatched image

a23oskme:
> Button works perfectly,
> 
> image
> 
> I however think we should resize the button and the text to match the buttons to the left and right of it. See attatched image

I agree! I can fix the menu in general, because it does not look so good.

a23oskme:
I reworked the menu a bit by making the filter options an overlay (the filter button will bring it up) and unified the styling of the buttons

image

a23axehe (approved):
Looks good now. 

image

PR #17879, opened 2025-05-22
Correct the command: permission error in new installer #17659

### LINKED ISSUE #17659: Correct the command: permission error in new installer
When installing the system with the new installer you can get a error message that you need to add permissions using a command. This command is re-used from the old installer and now looks like `chmod -R 755 /LenaSys/install/courses`. This won't work as you're in the new installer and should look like this  `chmod -R 755 /LenaSYS/newinstaller/courses`  (LenaSYS should also be changed to use capital letters in "SYS" as that's what the repo is named when it's cloned from git).

It could also be a good thing to add something to ensure that the person is in the correct folder in the terminal when trying to run the command as it won't work otherwise. For example: I had to use this to get it to work as intended; `chmod -R 755 /Applications/XAMPP/xamppfiles/htdocs/LenaSYS/newinstaller/courses` but this depends entirely on where the person has LenaSYS located.

a23oskme:
I updated the text that as instructed in the issue. Fixes #17659  
However, I did not manage to get the error myself. The error that i receive when removing permissions by running `sudo chmod 000 /Applications/XAMPP/xamppfiles/htdocs/LenaSYS/courses` is the one in the image below. 

Skärmavbild 2025-05-22 kl  15 33 39

I would appreciate if someone could test this. And if they are unable to get the error mentioned in the issue (to see the updated text), I could investigate further. 

The error I got is fixed by running the same command that I added to the feedback text though, so maybe we shall add that text to this error as well?

### DIFF
--- a/newinstaller/tools/modal.js
+++ b/newinstaller/tools/modal.js
@@ -1,4 +1,4 @@
-document.addEventListener('DOMContentLoaded', function () {
+document.addEventListener('DOMContentLoaded', function() {
     let modal;
     let closeModalBtn = document.getElementById("closeModal");
     let openModalBtn = document.getElementById("openModal");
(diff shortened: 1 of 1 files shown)


a23axehe (approved):
image

Changes are correctly made and appear as they should, however box is to narrow and it does not look good. I will create a issue for this which the frontend group can pick up.

Issue #17808, opened 2025-05-21
create a test for duggaedService/retrieveDuggaedService


a20artje:
Create a test following the same standard as the other tests in tests/microservices

PR #17880, opened 2025-05-22
test file for retrieveDuggaedService, 

a23danme:
for issue #17808
Created the testfile, tried alot of different things for the file to get more results. I would appreciate other peoples input on this one due to the file retrieveDuggaedService_ms.php (as i understand it) only defines a function and does NOT call it or echo any output(??). It doess not use json and echo as most others.
Due to this, the result of the test gives that the service returns null and this test cannot fully pass.
I tested and tried to do a temporary "wrapper" or "nonused" file of retrieveDuggaedService using json giving some kind of output and use the tester on that one instead but I did not get it to work. 

As said, would be great to get another input on this file and test and maybe I am just bad or me being home with fever is making me thinking slow.

### DIFF
--- a/DuggaSys/tests/microservices/duggaedService/retrieveDuggaedService_ms_test.php
+++ b/DuggaSys/tests/microservices/duggaedService/retrieveDuggaedService_ms_test.php
@@ -1,0 +1,30 @@
+<?php
+
+include_once "../../../../Shared/test.php";
+
+$testsData = array(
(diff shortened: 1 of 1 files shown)


d22marwi (approved):
The third test case is currently failing, which is consistent with other existing tests since the database schema update last year. While it's possible to make the test pass by hardcoding values that match the current database, this approach is not sustainable. I have discussed this with our group leader and the other tests passed, so this is correct.

PR #17881, opened 2025-05-22
Updated diagram.css

d22abdha:
fixed the issue where the bottom toolbar icons would disappear when resizing.
https://github.com/user-attachments/assets/53d5e6c6-c30a-4166-bf95-3ec88c0dbeaf


a23emisn (approved):
I can confirm that the bottom of the toolbar doesn't magically disappear when the window is resized with this patch. A simple yet effective fix, that is unlikely to cause any unwanted side effects elsewhere. 

I approve for merge!

PR #17882, opened 2025-05-22
Now in right file and has convertad all jquery minus Feedback

b20isato:
This is for https://github.com/HGustavs/LenaSYS/issues/17541


a23lisjo (changes requested):
There are some missed jQuery in these lines:

Line 83
image

Line 194 and 195
image

Line 222
image

It is okay to keep the jQuery for the function `toggleFeedBack()` as there is no direct translation to vanilla JavaScript.

The page does not work fully and gets error messages when trying to look at the preview. I do not think the problem lies in this issue since the problem comes up even in the group branch.

PR #17883, opened 2025-05-22
G3 2025 v8 convert jquery in dugga sys templates daily minutes.js #17522

EdvinWerner:


a22maxda (approved):
Conversion looks correct and nothing appears to have changed. (Dugga was broken before but changes have not changed how it works)

Approved, well done.

PR #17884, opened 2025-05-22
Created test for createGithubCodeExample_ms_test.php

### LINKED ISSUE #17813: create a test for sectionedService/createGithubCodeExample

a23bensh:
Fixes #17813 and adds a test file for createGithubCodeExample_ms. You'll find it in tests/microservices/sectionedService/createGithubCodeExample_ms_test.php. The third test case currently fails, consistent with other existing tests due to the new database schema since last year. The third test could be passed by hardcoding values that match the current database, but this would not be sustainable (as we notice happen with the old tests not passing the third test). This was mentioned in PR #17830.

### DIFF
--- a/DuggaSys/tests/microservices/sectionedService/createGithubCodeExample_ms_test.php
+++ b/DuggaSys/tests/microservices/sectionedService/createGithubCodeExample_ms_test.php
@@ -1,0 +1,39 @@
+<?php
+include_once "../../../../Shared/test.php";
+
+$testsData = array(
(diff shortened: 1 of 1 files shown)


a23axehe (approved):
Test 1 and 2 passed no problem. Test 3 failed which is consistent with your statement that there is a db error and not a error regarding this test. This test is OK

image

b23johgu:
Great choice to not hardcode values to make test 3 work! I agree that it's not sustainable and will merge this, the group next year can continue working on the tests.

PR #17886, opened 2025-05-22
Update sectionedService.md

a23petny:


a23louma (changes requested):
First of all, which issue does this fix? If the issue was to document retrieveAllCourseVersions_ms.php and retrieveSectionedService_ms.php the documentation looks good in general but some things have to be fixed to not give problems with the microservice database UI. Check in the template how it should be:
-  "Name of file/service" is missing for both files
- The output data and format has to be changed to follow the template
- If possible, for the input parameters add how they are stored in the database (see template)
- Under "Microservices Used" for retrieveSectionedService_ms.php, I think "(via callMicroserviceGET)" can be removed as it might give issues in the microserrvice UI.

Also, there are conflicts which would be good if you could try to resolve :)

b23johgu:
These microservices are already documented in sectionedService.md, will close this PR.

PR #17887, opened 2025-05-23
Re-engineered retrieveSectionedService_ms.php #17598

### LINKED ISSUE #17598: Re-engineer microservice retrieveSectionedService
Follow the "coding standard for removing includes" for this file in folder /sectionedService. The coding standard can be found in a couple of files in /microservices/Microservices Documentation. Now the functions in DuggaSys/microservices/curlService.php can be used.

a23louma:
Fixes #17598 . retrieveSectionedService_ms.php has been re-engineered and 25 files have been changed to use curlService, according do documentation these ones are used at the moment:

- \sectionedService\deleteListEntries_ms.php
- \sectionedService\readCourseGroupsAndMembers_ms.php
- \sectionedService\readUserDuggaFeedback_ms.php
- \sectionedService\removeListEntries_ms.php
- \sectionedService\updateActiveCourseVersion_sectioned_ms.php
- \sectionedService\updateListEntries_ms.php
- \sectionedService\updateListEntriesTabs_ms.php
- \sectionedService\updateListEntryOrder_ms.php
- \sectionedService\updateVisibleListEntries_ms.php
- \sectionedService\createGithubCodeExample_ms.php
- \sectionedService\createListEntry_ms.php

Not used according to documentation deprecated_microservices.md:

- \sectionedService\updateQuizDeadline_ms.php 
- \sectionedService\getListEntries_ms.php
- \sectionedService\getCourseGroupsAndMembers_ms.php
- \sectionedService\changeActiveCourseVersion_sectioned_ms.php
- \sectionedService\getDeletedListEntries_ms.php
- \sectionedService\readGroupValues_ms.php
- \sectionedService\getGroupValues_ms.php
- \sectionedService\getUserDuggaFeedback_ms.php
- \sectionedService\reorderListEntries_ms.php
- \sectionedService\retrieveAllCourseVersions_ms.php
- \sectionedService\setVisibleListentries_ms.php
- \sectionedService\updateCourseVersion_sectioned_ms.php
- \sectionedService\updateListEntriesGradesystem_ms.php


a23dilel (changes requested):
createListEntry_ms = working
image

updateListEntriesTabs_ms = working
image

deleteListEntries_ms = working
image

updateListEntries_ms = working
image

createGithubCodeExample_ms = not working, because createGithubCodeExample_ms has a few errors and in master branch also error. However using curlService funtions looks good.

updateListEntryOrder_ms = working
image

readCourseGroupsAndMembers_ms = working
image

readUserDuggaFeedback_ms = working
image

I could't find and test these microservices: removeListEntries_ms, retrieveSectionedService_ms, updateActiveCourseVersion_sectioned_ms and updateVisibleListEntries_ms. I would like to show a guide, about how to get these microservices in LenaSYS instead of trying to find these microservices.

I feel like removeListEntries_ms and deleteListEntries_ms are same thing and only different is removeListEntries_ms doesn't include `lid`.

In updateCourseVersion_sectioned_ms you have a few `header("Content-Type: application/json")` and it's enough to have one header. You can add header under include_once, but you don't need to change this.

a23louma:
image

updateVisibleListEntries_ms should be called when you change visibility but right now I think it’s only updateListEntries_ms being used, when changed to hidden, the value of visible is updated to 0.

image

But when you do this, you can see an example of when retrieveSectionedService_ms is being called in Docker logs:

image

retrieveSectionedService_ms.php is used by all the inverse dependencies in this PR and not called anywhere else except for in these microservices. According to the documentation it is responsible for retrieving all updated data from the database in the format of an array. It´s basically gathering and sharing information about courses with users, making sure it fits what they are allowed to see and do. For this I tried adding this at the end of retrieveSectionedService_ms before ”header('Content-Type: application/json');”
`error_log(
  'retrieveSectionedService_ms.php response array: '
  . json_encode($array)
);`
And then in Docker logs I could see the output including the updated value after changing the visibility to public or hidden:

image

For removeListEntries_ms, it’s supposed to be that you can remove a listentry, but it doesn’t get deleted from the database, you should be able to restore it. I haven’t been able to find this either in LENASYS so I don’t think it’s used at the moment. It should update the table listentries and set visible = '3'. But I have only been able to change it to 0, 1 or 2 when you change to hidden/public/login.

deleteListEntries_ms deletes a listEntry completly from the database.

updateActiveCourseVersion_sectioned_ms is supposed to be called when you change a course version to be the default version, but at the moment it doesn’t seem to be used and it’s setAsActiveCourse_ms instead which does the same thing: "UPDATE course SET activeversion=:vers WHERE cid=:cid". It can be seen in the database that activeversion is updated in the table course after doing this:

image

In Docker logs after changing a version to default version:

image

You are right about that there shouldn’t be several ”header("Content-Type: application/json")” in updateCourseVersion_sectioned_ms, I’ve removed them and now I only have it in the beginning after the includes.

a23louma:
Regarding createGithubCodeExample_ms, I will create an issue for fixing the errors in this file.

a23louma:
I will create an issue for removeListEntries_ms.php, the other ones I'm unsure if they are necessary.

PR #17889, opened 2025-05-23
Fixed failed to connect to the database. Test on Linux with docker and maybe macOS.

### LINKED ISSUE #17870: Failed to connect to the database. Problem on Linux with docker and maybe macOS.
I'm using Linux with Docker and after finish the installation using the new LenaSYS installer, then the course page shows a database connection error, which you can see the error in the screenshot below.

Image

a23dilel:
Fixes #17870.

I removed the command from `init.sh` because Docker can’t change the owner and permissions on whole LenaSYS folder in an apache server. Instead, these need to change on a local computer. That’s why you see the error `chown: changing ownership of '/var/www/html/coursesyspw.php': Operation not permitted.` However, on Windows, this doesn’t matter, because Windows doesn’t have permissions and ownership, so it still works, but not on Linux. I have moved some commands from `init.sh` to `setup.sh` and refined some commands.

To test this, you to have a Linux with Docker, and run the file `./setup.sh` or `bash ./setup.sh` (if your first command doesn’t work). Also, you need to have Sudo command to run this file. This file is located at `LenaSYS/dockerenv/setup.sh`. 

When you run this file, you will see some outputs in the terminal showing about files and folders being created, ownership and permissions beginning to change all folders in the LenaSYS folder. Which looks like this: -rwxrwxrwx _username_:www-data in local computer and apache server.

Then, go to the [new installer](http://localhost/LenaSYS/newinstaller/installer.php) and follow all these steps and do not forget to check “Distributed environment”. When you have finished the setup with the new LenaSYS installer, then press the finish-button which lead to the course page and the error should not appear. 

Also check owner and permissions in dockerenv folder and type `ls -l` and there you can your username, group and permission has changed. This also change all folders in the LenaSYS folder.

### DIFF
--- a/dockerenv/Dockerfile
+++ b/dockerenv/Dockerfile
@@ -43,8 +43,6 @@ RUN echo '<Directory /var/www/html>' > /etc/apache2/conf-available/custom-direct
     && echo '</Directory>' >> /etc/apache2/conf-available/custom-directory-listing.conf \
     && a2enconf custom-directory-listing
 
-RUN chown -R www-data:www-data /var/www/html/
-
 # Copy the initialization script
 COPY ./init.sh /usr/local/bin/init.sh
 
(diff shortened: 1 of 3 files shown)


a23dilel:
Ok, now I have added a dseditgroup for macOS. Test again if this works!

PR #17893, opened 2025-05-23
Replaced jQuery with JS in curve-dugga.js 

a23yutta:
Replaced jQuery with JS in `/DuggaSys/templates/curve-dugga.js`. 

Additionally, the toggle feedback and show facit features has not been introduced to the page yet, making it impossible to verify whether the JS conversion works as intended. Related functions: `toggleFeedback()`,  `showFacit()`

The `toggleFeedback()`  function in the end of the file have not been converted as there is no direct JavaScript equivalent for jQuery's `slideToggle()`  function, replacing it would require a custom implementation. This is too big part this issue and should therefore be addressed in a separate issue.

### DIFF
--- a/DuggaSys/templates/curve-dugga.js
+++ b/DuggaSys/templates/curve-dugga.js
@@ -516,10 +516,10 @@ function handler_mousedown(ev) {
 	clickstate = 1;
 	// Figure out if we clicked in an object
 	document.querySelectorAll("#operations > option").forEach(function (option) {
-		var opArr = this.value.split(" ");
+		var opArr = option.value.split(" ");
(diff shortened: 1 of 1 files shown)


b23sebth (approved):
**Fixes [#17521](https://github.com/HGustavs/LenaSYS/issues/17521) - approved!**

The original issue states that commented out jQuery code should be removed. The function `fitToContainer()` is commented out and contained jQuery. I've spoken about this with the PR author and that particular function appears to be slated for future implementation.

I was not able to test `showFacit()` or `closeFacit()` but have reviewed the code changes.

Everything works the same on this branch as on the group branch, great job!

## What's changed:
- jQuery has been changed where possible

PR #17897, opened 2025-05-23
Added check to course visiblity

a22maxda:
Added a check to exclude Demo-Course and Testing-Course from access filters and make them accessible to everyone, regardless of whether they are signed up or not.

(Alvik has no active courses)

Group:
Screenshot 2025-05-23 at 11 14 32

PR:
Screenshot 2025-05-23 at 11 15 16

### DIFF
--- a/DuggaSys/courseed.js
+++ b/DuggaSys/courseed.js
@@ -699,9 +699,11 @@ function returnedCourse(data) {
 	if (data['entries'].length > 0) {
 		for (i = 0; i < data['entries'].length; i++) {
 			var item = data['entries'][i];
-			// Do not show courses the user does not have access to.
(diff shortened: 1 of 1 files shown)


a23erigu (changes requested):
Now this has made the courses appear for every user so that is a success.
But i'm still going to rejet this as its hard coded that these courses always appear.
The problem of public visibility not showing still exist.
This solution just hides the problem.

a22maxda:
For context, the root of the issue is that no users are given access to the courses and therefore they are hidden as they are not registered and not guests. This change to course access gives all users access to the courses.

a23alera (approved):
Everything seems to work as expected now. The courses are visible for all relevant user types.

Issue #17850, opened 2025-05-22
Limited icon options on sequence lines


a23aleva:
As of now the only arrow icon available on Sequence Diagram lines is ARROW:

![Image](https://github.com/user-attachments/assets/0f4af240-f475-4b5a-8fd3-5cdbaa39a573)

But the standard arrow icon to use when modelling reply messages is this one:

![Image](https://github.com/user-attachments/assets/bd73af9e-e8cb-4f43-b2c0-e8ec299059d4)

Both these arrows are accessible on other types of lines so it should be possible to implement them for Sequence lines as well. Please see to this and make both arrow icons usable through the line dropdown.

PR #17899, opened 2025-05-23
Fixed: G1 2025 v8 #17850 limited icon options on sequence lines

a23aleva:
Line icon dropdown for Sequence Diagrams now has two alternatives: Arrow and Small_Triangle. I changed name from ARROW to Small_Triangle since ARROW and Arrow looked ugly and were confusing as dropdown options.

https://github.com/user-attachments/assets/6c53ff11-a590-479d-8eae-5aa708bff8ac


a23carpa (approved):
A good addition that works exactly as shown in the video.

PR #17904, opened 2025-05-23
create a test for profileService/retrieveProfileService_ms.php #17810

d22marwi:
The test is located at: tests/microservices/profileService/retrieveProfileService_ms_test.php. 

The third test case is currently failing, which is consistent with other existing tests since the database schema update last year. While it's possible to make the test pass by hardcoding values that match the current database, this approach is not sustainable.

### DIFF
--- a/DuggaSys/microservices/profileService/retrieveProfileService_ms.php
+++ b/DuggaSys/microservices/profileService/retrieveProfileService_ms.php
@@ -8,3 +8,6 @@ function retrieveProfileService($debug, $success, $status){
 		);
     return $array;
 }
+
+header("Content-Type: application/json");
+echo json_encode($array);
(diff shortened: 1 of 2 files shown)


a23bensh (approved):
Test file looks mostly good. It matches the function structure and correctly checks the expected output. However, the test sends success, status, and debug as part of the POST data, which doesn’t really pass the tests for what this service would be used for. This test basically just echoes back what you inserted.

BUT even though all of this may be true, when looking at 'retrieveProfileService_ms.php' there isn't really much to go after and there is pretty much nothing to even test with this service as of now. I would say that a test for this service is pretty unnecessary and is not needed at the moment. It could be kept as a placeholder for now and changed later when more is added to the service!

Only change needed is adding what issue this fixes, I assumed it's #17810.

PR #17907, opened 2025-05-23
Changed css for fab buttons

a23eliax:
The issue was to fix the spacing between the option-fab button and the dugga-button so they don't overlap. I added some more margin between them so that shouldn't be a problem. Although (at least for me) it wasn't the browser that was the problem. it was the size of the screen on the phone you used. When you are on a bigger phone, like Iphone XR (figure 2 below), the spacing is a little bit bigger than if you use an Iphone SE (figure 1). But I see this as a positive thing because the two dugga-buttons is a group and that should be visible, while the fab-option button is for it self.

Iphone SE:
image
_figure 1_

Iphone XR:
image
_figure 2_

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -173,7 +173,6 @@
     line-height: 80px;
     text-align: center;
     border-radius: 40px;
-    background-color: #eb4;
     user-select: none;
     cursor: pointer;
     z-index: 5000;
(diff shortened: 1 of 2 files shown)


a23emijo (approved):
The buttons are more grouped now compared to before and doesn't look like the Firefox picture in the issue did. I tested this is Chrome and Edge and both look nice. I approve! 👍

a23emijo (changes requested):
I was a bit too quick and saw after I approved that the code is using "px" and not "vw, vh, em or similar" like the customer wanted us to do. Please change this 😊

a23emijo (approved):
After the change to 1.5vh, everything is okay and approved!

Issue #17885, opened 2025-05-22
A petty lil issue, change word to use capital letter


b23johgu:
"generate local SSL certificate" should start with a capital letter (Generate). So sorry for being petty haha.

<img width="647" alt="Image" src="https://github.com/user-attachments/assets/237da847-2200-4efa-8ba0-6ab30af16daf" />

PR #17908, opened 2025-05-23
change word to use capital letter

d22marwi:
Fixes issue #17885. 
In the new installer, step 6, "generate local SSL certificate" now starts with a capital letter.


a20artje (approved):
Hmmm, after careful consideration and testing this is approved.

PR #17909, opened 2025-05-23
G3 2025 v8 #17599 fix course dropdown length

### LINKED ISSUE #17599: course dropdown fix (desktop)
The course dropdown is to small in sectioned, it needs to be made to take up more space in the navbar.

This is how it looks right now:
Image

This is how it should look like, does not need to be this big though, but the text in dropdown should be visible:
Image

a23lisjo:
Fixes #17599. 

### What has been changed
- A `max-width` for the dropdown has been removed making it possible to see the entire text inside the dropdown window.
- Extra: An extra closing curly-bracket was spotted in the file and was removed. 

Before:
image

After:
image

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -7878,7 +7878,7 @@ only screen and (max-device-width: 568px) {
     bottom: 100px; /* Nudges the FAB 10px up from the bottom */
   }
 }
-}
+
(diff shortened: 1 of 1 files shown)


a22carsv (approved):
Works as intended and doesn't break anything, approved.

PR #17910, opened 2025-05-23
Fixed snapping of activation elements

a22abdjo:
The element should now continue snapping when moving from the breaking point, thus being placed beneath the cursor. Instead of calculating with deltas ( startx, targetPos) etc, the element is placed directly under the mouse using its center. These changes were implemented and made in the diagram.js file.


a23emijo (approved):
The snapping now works great and the box is following the cursor everytime. Approved! 💯

PR #17912, opened 2025-05-23
Geting rid of x and stäng button in dugga5

a23erigu:
I have deleted the x and stäng buttons in dugga5.html
I couldn't find any reference to the buttons in the js file.
Both buttons only hade generic css for buttons so no change there.

pic of how it looked
image

pic of how it is now
image


EdvinWerner (approved):
This PR has gotten rid of the stängbutton just as advertised but if one was to be really "picky" it might have been a good idea to go through the css file and delete any style associated with said button so we don't leave anything unused inside of codebase but since the PR solves the original issue it's approved for merge!

Issue #17754, opened 2025-05-20
Search dropdown not disappearing on edit course access page


b23sebth:
This problem appears to have been introduced by https://github.com/HGustavs/LenaSYS/pull/17398, but fixing the issue with creating and deleting users take precedent so this gets made into it's own separate issue. The example dropdown doesn't disappear correctly when the user stops hovering, unless leaving the dropdown from the right side. Please make the hover dissappear correctly, regardless where the mouse exits the search box.

## This problem is located in the edit course access and can be reached by:
Selecting a course -> Clicking the lock icon in the navheader while logged in
![Image](https://github.com/user-attachments/assets/53f32810-5908-42d3-bdfe-642d7435bb6b)

## Example from https://github.com/HGustavs/LenaSYS/pull/17398: 
[...]

PR #17913, opened 2025-05-23
Search dropdown not disappearing #17754

f23linan:
So this was a silly issue. The problem was that that function that removes the dropdown was empty. It was an easy fix and should now work as intended.

### DIFF
--- a/DuggaSys/accessed.js
+++ b/DuggaSys/accessed.js
@@ -73,7 +73,8 @@ function hoverSearch() {
 
 //stops displaying the dropdown when removing cursor from search bar
 function leaveSearch() {
-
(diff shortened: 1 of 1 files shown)


a23fanro (approved):
The issue seems to have 'evolved', stated in https://github.com/HGustavs/LenaSYS/issues/17754#issuecomment-2909939752

This PR now makes it so the dropdown suggestions almost never appear, only way I can make it appear is by hovering over AND inputting text in the search field at the same time. But it does fix the (new) issue that you can not input any text in the field. And since inputing text is more important that a dropdown with suggestions is, I will approve this for merge. But the issue is still in need of fixing, and a new PR is needed to fix it.

PR #17919, opened 2025-05-23
Fixed clearGitFiles and refreshGithubRepo

a23danme:
Fixed clearGitFiles and refreshGithubRepo for issue #16861

I tried testing it or looking up how to test it, maybe im dumb and its easy to test but as I understood it there iss this file "metadata2.db" that has tables on its own. Tried to download extensions as sqlite to add a row so I could delete it but I could not add one. The extension sqlite viewer worked good to view the tables but they are empty. With this said I am not 100% certain howw to test this properly.

### DIFF
--- a/DuggaSys/microservices/gitCommitService/clearGitFiles_ms.php
+++ b/DuggaSys/microservices/gitCommitService/clearGitFiles_ms.php
@@ -6,19 +6,25 @@ include_once "../../../Shared/sessions.php";
 
 global $pdo;
 
+header("Content-Type: application/json");
+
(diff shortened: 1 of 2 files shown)


d22marwi (commented):
The files are currently not in use. The old file, GitCommitService, is the one in use. This should probably be changed later, but it makes testing this solution difficult at the moment.

ClearGitFiles seems to be working, Following this path: http://localhost/lenaSYS/DuggaSys/microservices/gitCommitService/clearGitFiles_ms.php
the output is this:
success

Issue #17921, opened 2025-05-23
Sidebar should cover FAB-button on mobile


a23emijo:
The FAB-button should not be visible or clickable when the sidebar is opened in the mobile version of the diagram
![Image](https://github.com/user-attachments/assets/48bb64e9-531b-4473-b9b9-4e423be1326b)

PR #17922, opened 2025-05-23
Fixed mobile view: Make menu auto-close

a23emijo:
The menus now closes when a new menu is opened.

_**See the video down below.**_
https://github.com/user-attachments/assets/8be5f97d-8167-4b0b-a310-52353eaa8899

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -522,9 +522,8 @@ function getData() {
 
     //Main mobile FAB-button
     document.getElementById("diagram-fab").addEventListener("click", () =>{
-        document.querySelectorAll('.fab-inner').forEach(button => {
-            button.style.display = button.style.display === 'flex' ? 'none' : 'flex';
-          });
+        let fabList = document.querySelector(".fab-btn-list");
+        fabList.style.display = fabList.style.display === 'none' ? 'block' : 'none';
     });
 
     //Side navbar buttons. (save, load and reset are inside diagram_dugga)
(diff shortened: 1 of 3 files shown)


a23vinge (changes requested):
The code looks great and functions as intended — I don’t see any major issues. However, I did notice that when you press the button, the overlay menu from the right doesn’t disappear. According to the pull request, it’s mentioned that “the menus now close when a new menu is opened,” so I believe this behavior needs to be addressed, if it is not now, that the button is not defined as a menu. (this can then be ignored)


https://github.com/user-attachments/assets/2473a3ab-dc1e-4c2f-b652-acbb58b23a43

As can be seen, everything seems to be working but the button when pressed deos not get prioritesed.

a23emijo:
> The code looks great and functions as intended — I don’t see any major issues. However, I did notice that when you press the button, the overlay menu from the right doesn’t disappear. According to the pull request, it’s mentioned that “the menus now close when a new menu is opened,” so I believe this behavior needs to be addressed, if it is not now, that the button is not defined as a menu. (this can then be ignored)
> 
>  Inspelning.2025-05-26.112930.mp4 
> As can be seen, everything seems to be working but the button when pressed deos not get prioritesed.

I have made an issue about this here https://github.com/HGustavs/LenaSYS/issues/17921. My isse was not to fix this 😄

a23vinge (approved):
I APPROVVEE NOW!

PR #17934, opened 2025-05-26
Codeviewer single box view fixed #17746

### LINKED ISSUE #17746: Viewing specific boxes in codeviewer looks bad
In codeviewer if you press the hamburger menu in the top bar you can select to only see specific boxes.

If you do then you get a view of only that box but it looks really bad with it just stopping where the code ends.
Pls at least extend the background a bit down to make it look better (maybe even to the end of the screen depending on what looks the best).

pic of how it looks when viewing all the boxes.
Image

pic of how it looks when viewing one box.
Image

a23erigu:
I have fixed the issue of the singular view in codeviewer looking horrible.
it now covers the entire screen.
I also fix the extra vertical scroll that always existed in code viewer.

How it looks now
image

how it uset to look
image

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -283,7 +283,6 @@ p {
 
 #content {
   background: var(--color-background);
-  margin: 0px 0px 15px 0px;
 }
 
 #accessedcontent {


a23simry:
change from master to group branch pls!!

a23jesja:
approved

PR #17936, opened 2025-05-26
Redo recursive relations into its own element

a23carpa:
A recursive line (self call) is now no longer a function of each element but its own separate element to be placed.
Now the user gets a error when trying to draw a line to the same element and is advised to use Self call instead.
Its named Self Call and can be selected from the menu. 
At the moment its a cube shape, but this is very easy to change in the future.

* The main issue wanted the element to only be visible when selected or having no lines connected, i where unable to do this and a issue will have to be made for that functions specifically.

* Arrows(icons), cardinality and label are not optimally placed and might need to be looked into to in future issues.

* I removed much of the old recursive functions, but not all. This requires a deeper dive and i feel a separate issue would suit it better.

I feel that this works and solves the basic problems of the recursive, its not perfect or fully complete, but also makes it easier for new implementation and fixes as it is its own element.  Hopefully the next year they can continue evolving it.
I feel i could make it better and fix the smaller issues, but i have spent to much time already on this issue and feel it would be better to make problems there own separate issues instead. 

[Screencast from 2025-05-26 11-46-40.webm](https://github.com/user-attachments/assets/531a5b52-268e-45c8-bcf8-5c94ac529fbe)

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -88,6 +88,17 @@
     }
 }
 
+
+.selfcallBox {
+  pointer-events: all;
+  cursor: pointer; /* if you want the hover cursor */
+  opacity: 0;
+}
+
+.selfcallBox:hover {
+  opacity: 1;
+}
+
 #svgbacklayer {
     position: absolute;
     left: 0;
--- a/DuggaSys/diagram.php
+++ b/DuggaSys/diagram.php
@@ -624,6 +624,14 @@
         </fieldset>
         <!-- MODES FIELD ENDS HERE!! -->
 
+        <!-- Self Call -->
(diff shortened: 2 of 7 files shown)


a23kevre (approved):
I agree that the issue is solved but in another way. The arrows and labels need some polishing just like you said: example : 
image

me personally I like the menu placement, definitly agree that the issue need to be split apart. The warning or message comes up when you double click an entity with the line which was stated in the issue and is solved. There are a few bugs that will need to be figured out in the future just like you said, here is one for reference:
image

All in all even though this change needs a lot of fixing and a few bugs here and there I still approve these changes since the main issue has been solved for the most part, and it can be worked on in the future because of easy access.

Issue #17545, opened 2025-05-13
Can't move elements with pointer


a23eliax:
On mobile version, you can't drag around an element with the pointer. You can only move the entity when you are using the box selection.

https://github.com/user-attachments/assets/53fb1b9e-cbf2-48ce-b84e-8d1ebe8c6376

a23miksk:
To further specify this issue: Make elements selectable with touch in mobile mode.

PR #17937, opened 2025-05-26
Fixed pointer-tool on mobile (#17545)

b23ebbst:
Fixed the issue with being unable to select and move an element using the pointer-tool on the mobile version. The user is now able to select an element, and then move it around the grid-area. 

Is this the perfect solution? Honestly no. 
However, the main problem is solved - a user can select an element, drag it around, and can open options if they press the options-button while the element is selected. Sure, the element is not visible while dragging it around (like it is when using box-select), but at this point that feels like a minor problem that can be fixed later on (and should be it's own issue as I have not managed to fix that in the four days I've been trying to solve this). 

https://github.com/user-attachments/assets/a3bd75c1-705c-4ae4-892d-25d40bc8886e

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -940,7 +940,22 @@ function setupTouchAsMouseSupport() {
         if (event.touches.length === 1) {
             const touch = event.touches[0];
 
+            // If the Pointer-tool is used, a selection can be done on one element, and movement is possible
             if (mouseMode === mouseModes.POINTER) {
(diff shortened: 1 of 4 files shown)


a23emijo (approved):
The implemented solution kinda works, it's not complete but because the user can move the elements and work with it I'm going to approve this. For some reason, when using the Pointer-tool, the element gets an offset when trying to move it. This doesn't happen when moving with Box Selection though. This problem can be solve in another issue though! 👍 

https://github.com/user-attachments/assets/b60993f1-4cff-4da3-9819-b346c393f5c6

Issue #17901, opened 2025-05-23
Add a limit to characters in savefile


a23kevre:
at the moment you can write endless amounts of characters when saving a file. This affects the notification popup:

![Image](https://github.com/user-attachments/assets/2e8fe2c3-75a7-4928-b61c-e60148ce05e2)

and it also affects the load file area badly: 

![Image](https://github.com/user-attachments/assets/9ffb2fdb-c7c2-4785-b082-2037b0aff376)

![Image](https://github.com/user-attachments/assets/593248ed-e61a-43a1-b200-a99d7dffdb3c)

not only does it affect the forced width by adding the delete button to the right of it. It also counts towards the scroll bar but with invicible text:

![Image](https://github.com/user-attachments/assets/7cb63477-6073-4c96-b823-98fe07738937)

a23emijo:
Didn't this https://github.com/HGustavs/LenaSYS/pull/17690 PR fix this?

a23emijo:
This PR https://github.com/HGustavs/LenaSYS/pull/17690 should have fixed this

a23miksk:
> This PR [#17690](https://github.com/HGustavs/LenaSYS/pull/17690) should have fixed this

Why?

PR #17940, opened 2025-05-26
Fixed G1-2025-v8-#17901 add a limit to characters in savefile

a23eliax:
Added an ellipsis on the popup so the text overflow doesn't get outside the box.

image

Added an ellipsis to the filenames too so they fit in the load container.

desktop:
image

mobile:
(the arrow-chevron disturbs me because it's over the load container. It's not a part of this issue so I'm making a new issue about this)
image

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -1271,7 +1271,10 @@ button.saveButton {
   font-size: 16px;
   margin-top:10px;
   padding-left: 5px;
-  width: 70px;
+}
+
+.deleteLocalDiagram {
+    width: 70px;
 }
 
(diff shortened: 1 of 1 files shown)


a23gabst1 (approved):
The solution works as intended and adds a ellipsis (three dots) if the number of characthers exceeds the width of the element.  The code looks fine but would love to see less code repeated by chaining css style with commas e.g., but I still approve these changes either way! See figure 1 and 2. 

`#diagram-message  .error,
#diagram-message .succes,
... {...}
`

figure 1 

image

figure 2

Skärmbild 2025-05-26 195552

Issue #17927, opened 2025-05-26
Tool popouts no longer showing up when hovering over toolbar


a23aleva:
There seems to be a new bug where tools can't be picked from an appearing sidebar when hovering over their representative in the toolbar. The problem is that the sidebar doesn't even appear. Fix this since the only way to change tool right now is via keybinding.

https://github.com/user-attachments/assets/eada67b4-b59a-4217-a1a3-8ff05f8d9031

Issue #17756, opened 2025-05-20
Add a scrollbar to the toolbar when it doesn't fit within the window


b23danhe:
The issue is that the toolbar layout breaks and becomes cluttered when the browser window is small or when using a device with a smaller screen.

The customer has approved a solution where a scrollbar appears in the toolbar if all elements cannot fit within the available window size.

The task is to implement a scrollbar in the toolbar and also ensure that the size of the toolbar icons adjusts dynamically based on the toolbar's dimensions and with the same proportions, whether due to window resizing or browser zooming.

https://github.com/user-attachments/assets/30be8551-775e-497f-b39e-6802a1ce2f31

<img width="440" alt="Image" src="https://github.com/user-attachments/assets/73cc64b1-6432-447f-85c2-92396b5e0e3d" />
<img width="895" alt="Image" src="https://github.com/user-attachments/assets/1d6cd937-21ee-4cfd-8954-53b93d466286" />

b23danhe:
Since the solution to this issue caused the popouts in the menu to malfunction, the issue needs to be reopened in order to explore an alternative fix that works correctly with the popouts. The scrollbar was removed in PR #17952

a23gabst1:
I believe that the positioning of some elements in the elements in the toolbar hinders the scrollbar. Absolute and fixed both takes the element out of the document flow however absolute still depends on the closest relative parent whereas fixed does not. The solution would be to tweak a little bit with the positions on the diagramPopOut or its first child.

PR #17952, opened 2025-05-26
Fixed: Tool sidebar no longer showing up when hovering over toolbar #17927

a23aleva:
Unfortunately, the scrollbar and popouts are both situated in the same targeted div, which made the css overflow code - that creates the scrollbar - override the popout's overflow on the x-axis. The only way I found to solve this would have been to rewrite a large chunk of the toolbar html-code inside the diagram.php-file to incorporate a specific div for the scrollbar. I had a talk to the group leader and he said that it was better to prioritize the popout menus rather than having a scrollbar for the toolbar. So while this solves the problem of the popouts not showing, it isn't really a complete solution. Hence the Issue #17756 for the scrollbar will probably get reopened for someone to try to implement in a better way that doesn't remove the popouts. 

https://github.com/user-attachments/assets/de22df4f-3298-4aae-b075-5f1c1422e94d

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -34,35 +34,13 @@
         top: 0;
         left: 0;
         bottom: 0;
-        width: 60px; 
+        width: 50px; 
         height: 100vh; 
         z-index: 1000;
-        overflow-y: auto; /* Scrollbar appears only when needed */
-         overflow-x: hidden;
+        display: flex;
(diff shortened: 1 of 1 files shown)


a23gabst1 (approved):
This solution solves the issue and works fine! The code looks fine to me and I approve these changes. See figure 1. 

figure 1
bild

PR #17956, opened 2025-05-26
Fixed selection conflicts when using UML Superstate and Sequence Loop templates

a23emijo:
Created a for loop that iterates through the elements placed on the canvas and if the bring to front button is pressed on a Loop, Alt or Super State, nothing happens and the code return nothing.

Also removed some unnecessary console logs in the code.


a22abdjo (approved):
After reviewing the changes made and the solution provided, the selection conflicts when using UML Superstate and Sequence Loop templates is solved.

PR #17961, opened 2025-05-26
making font size and zoom work

a23erigu:
I think I have solved the issue.
Now the zoom reset button goes to the level set as font size.

I hope that this is what the issue wanted me to fix


b20isato (approved):
changing font size and zooming functions well togeter

PR #17962, opened 2025-05-26
G3 2025 v8 #17860 fix h3 styling fileed

### LINKED ISSUE #17860: Header 3 styling in fileed.php gets overwritten by other CSS
When editing the code in fileed.php the header 3 is not in the correct styling.

Image

How to access the code editor (the pen):
Image

### What has to be fixed

- The H3 has an underline that should not be there
- The H3 font-size is smaller than H4 

Image

### What causes this

The CSS file that should be used is the one in style.css
Image

But instead the styling in markdown.css is used which gives the H3 an underline and a bigger font size.

Image

There might be another problem that I have not found.

a23lisjo:
Fixes #17860 

### What has been fixed
- Multiple CSS on the same element has been removed so only styling from `style.css` is now active 
- All styling about the headers in `markdown.css` has been removed
- Added styling for headers with no styling or different styling from the rest

Before:
image

After:
image

### DIFF
--- a/Shared/css/markdown.css
+++ b/Shared/css/markdown.css
@@ -176,16 +176,6 @@
 	border: 1px solid #ddd;
 }
 
-#markdown h4, #markdown h5, #markdown h6 {
-	margin: 0;
-}
-
-#markdown h3{
-    font-size:16px;
-    margin: 2px 0px 2px 0px;
-    border-bottom: 2px solid #aaa;
-}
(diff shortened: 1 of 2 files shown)


a23yutta (approved):
The code works as intended fixing the issue stated in issue #17860 as well as in this PR #17962. 

[APPROVED]

PR #17963, opened 2025-05-26
Added fucntionality in mouseMode_onMouseUP()

a23thefo:
## What I did
mouseMode_onMouseUp() now closes the open mobile submenu (if there is one) after it finishes placing the element.

## Demo
Screencast from 2025-05-26 16-31-21


b23ebbst (approved):
Works just as intended with the sub-menu closing after placing an element. Well done, I approve!! ⭐⭐

https://github.com/user-attachments/assets/c1564731-1368-45cb-a224-1aabfc150c43

Issue #17888, opened 2025-05-23
Make class height resizable


a23miksk:
When an user tries to resize the class elements height the height stays the same no matter what. If the user drags on one of the top corners, the elements position is moved with they height staying the same. Make the class elements height resizable when the user attempts to drag one of its corners.

_Video of the element moving weird when dragging_
https://github.com/user-attachments/assets/ca908640-ccd6-4221-bd0d-91c3f8311ab1

PR #17967, opened 2025-05-26
[#17888] Fixed issue with certain elements not resizing properly

a22nevan:
Certain elements, the ones that consisted of multiple "layers", i.e. UML Class, SD State, and IE Entity elements, were behaving weird when trying to resize them.

The cause of the bug lied in the way the elements were drawn. The _boxh_ value had been removed from the function that drew each box "layer" in the element. Without this, the elements wouldn't resize properly since boxh is crucial for all elements to resize properly. Boxh has been readded to the affected elements, while trying to keep the box-layer height ratios as similar as possible to what they were during the bug.

**The affected elements now resize properly again:**
https://github.com/user-attachments/assets/aab2c62e-0ce0-42d3-b0e5-ae3df0d816d6

### DIFF
--- a/DuggaSys/diagram.js
+++ b/DuggaSys/diagram.js
@@ -1169,6 +1169,7 @@ function mmoving(event) {
             }
             break;
         case pointerStates.CLICKED_NODE:
+            console.log("hello");
             const index = findIndex(data, context[0].id);
             const elementData = data[index];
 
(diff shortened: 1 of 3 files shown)


a23thefo (approved):
This fixes the problem! The code is edited in a clean way and is following guidelines!

Screencast from 2025-05-27 09-34-58

Issue #17814, opened 2025-05-21
create a test for sectionedService/removeListEntries


a20artje:
Create a test following the same standard as the other tests in tests/microservices

PR #17968, opened 2025-05-26
G2 2025 v8 #17814

a22wilas:
Fixes issue https://github.com/HGustavs/LenaSYS/issues/17814

To test comment out the delete part:
`'query-after-test-1' => "
      DELETE FROM listentries WHERE lid = 9999;
    ",`
    
Check the database table "listentries".
A new row should appear and the value visible should change to "3".

Had to add "include_once "../curlService.php";" in updateCourseVersion_sectioned_ms.php because it uses the function callMicroservicePOST.
The file removeListEntries_ms.php has an include to updateCourseVersion so would get error without the new include.


a23bensh (approved):
All tests pass and work as intended. Sometimes if a previous test run failes before reaching the delete statement (DELETE FROM listentries WHERE lid = 9999;) it causes an error. This could be fixed by adding the statement before the tests run aswell? It doesn't happen constistently enough though so I wouldn't really say this change is that necessary. Looks good overall!

PR #17969, opened 2025-05-26
Updated diagram.css

d22abdha:
Changed width and height from vh to px and the icons keeps the same size and position regardless of window height

https://github.com/user-attachments/assets/6e56308e-66ce-4d29-baf1-1645784aba2c

### DIFF
--- a/DuggaSys/diagram.css
+++ b/DuggaSys/diagram.css
@@ -385,8 +385,8 @@
     position: relative;
     transition: background-color 0.3s;
     margin: auto;
-    width: 3.2vh;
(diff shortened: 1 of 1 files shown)


a23aleva (approved):
The change from viewport height to pixels seems to have done the trick. There is some small artifact close to the bottom of the toolbar when pulling the window downwards but it quickly resets itself so it's really negligible. Good work and approved!

https://github.com/user-attachments/assets/8f3510ce-35f1-4050-9b2b-e1443c678f0c

Issue #17178, opened 2025-05-05
Locate and Change AJAX call to normal Fetch() -= sectioned.js =- Line 1198, 3615 & 4711


a23elism:
### The lines, to be converted in this Issue are the AJAX calls starting on line 1198, 3615 and 4711, These are grouped based on them all using opt and action with internal POSTs
Highly recommend looking at the parent issue for a quick run-down on how to convert AJAX to normal Fetch() (normal JS)

Line 1198:

![Image](https://github.com/user-attachments/assets/568fb3ea-2094-4a36-9ca7-41bebddfc1d5)

Line 3615:

![Image](https://github.com/user-attachments/assets/5ddb269d-fa14-4761-bb1c-1755381a8c7a)

Line 4711:

![Image](https://github.com/user-attachments/assets/28921dc6-69be-41a9-bd22-54d5c6e34b06)

PR #17970, opened 2025-05-26
G3 2025 v8 #17178 locate and change ajax a23kavis

a23kavis:
Entry 4, 14 and 17 have been changed to use the modern javascript fetch method, while maintaining headers and toasts as best as possible.
 
Entry 4
image
image

Entry 14
image
image

Entry 17
image
image


a23erigu (approved):
Its a bit hard to check these as it hard to find where they are used but from what I can tell it seems to work just fine.

Good work

Issue #17831, opened 2025-05-21
Bug: saving modifications of the Github URL does not work for courses


a23yutta:
Saving modifications of the Github URL does not work for courses and results in it being set to "undefined".  Investigate this issue and try to find a possible solution to fix this. If the investigation suggests that the page may not have introduced this feature yet - report to the group leader about this issue.

Where to find edit course dialog:

Login -> courseed.php -> find edit course: 

![Image](https://github.com/user-attachments/assets/0939fd4c-b544-4553-8352-53ce90de6f5d)

a23yutta:
Documentation of this issue can be found at PR [#17976](https://github.com/HGustavs/LenaSYS/pull/17976).

PR #17976, opened 2025-05-27
Investigation of saving modifications of the Github URL of a course

a23yutta:
### This PR does not close or resolve the issue #17831 BUT it introduces safer handling of the Github URL input and adds console log to aid future developers with debugging. 

**This PR is a documentation of issue** [#17831](https://github.com/HGustavs/LenaSYS/issues/17831).

After some digging, I'm still not sure what's causing the GitHub URL to be undefined or empty nor have I found a solution for this issue.

When testing each function related to the GitHub URL individually, they all appear to work as expected. However, the whole process of updating the GitHub URL doesn't seem to work for some reason. 

##

### Sample testing of the issue after applying debugging changes
**Note for when testing:** I filtered the GitHub URL to safely handle double quotes and avoid injecting undefined values into HTML attributes which is why the value is empty instead of "undefined".

**Disclaimer:** Initially intended to attach a recording of the test, but the file exceeded the 10 MB limit.

**Sample test:**
When selecting a course to edit with a valid GitHub URL, then saving it after updating the URL, and finally reselecting the same course, the GitHub URL field appears empty. However, if you reload the page after these steps, the updated GitHub URL displays correctly - WHICH does not make much sense. Could it be an issue with the server, the database, or the backend? Maybe it's something I'm overlooking - honestly, I’m not sure.

**Console log of the sample test:**
Skärmbild 2025-05-27 092156

### This concludes my documentation of the issue. Anyone is welcome to continue the investigation and, hopefully, identify the root cause.

Github URL related files: 
`courseed.js`
`courseed.php`

Github URL related functions in `courseed.js`:
`updateGithubRepo()`
`selectCourse()`
`fetchGitHubRepo()`
`updateCourse()`
`quickValidateForm()`
`validateForm()`
`returnedCourse()`

### DIFF
--- a/DuggaSys/courseed.js
+++ b/DuggaSys/courseed.js
@@ -64,6 +64,9 @@ function updateCourse() {
 	const courseid = "C" + cid;
 	const token = document.getElementById("githubToken").value;
 
+	console.log("updateCourse() => Input element:", courseGitURL);
+	console.log("updateCourse() => courseGitURL value before fetch:", courseGitURL || "NO INPUT");
+
 	const url = "../DuggaSys/gitcommitService.php";
 	const params = {
 		githubURL: courseGitURL,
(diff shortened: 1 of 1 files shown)


EdvinWerner (approved):
Looks good to me, not sure if it's required inside of the main but since it doesn't break anything as far as i can tell, it's approved for merge!

PR #17983, opened 2025-05-27
DragDrop made rougt with reload

b20isato:
This sure is a rough solution but it does atleast update, really hope someone will make a better solution. This just does a reload of page when moved


a23erigu (approved):
As stated by the PR owner it's a very quick fix and not good at all but it works if we need a quick fix.

Issue #17957, opened 2025-05-26
Make the "Bring to Front" not clickable and look disabled


a23emijo:
Style the button to make it understandable that the button is disabled and not clickable on Loop, Alt and Super State.

![Image](https://github.com/user-attachments/assets/c24eab42-dec2-415e-86d7-d633b2407466)

PR #17990, opened 2025-05-27
[#17957] Changed visibility of buttons in options pane for UML SuperState and Sequence Condition elements

a22nevan:
The **Bring to Front** button is no longer displayed in the options pane for the UML SuperState and Sequence Condition elements. 

The description of the issue was to make it greyed out and unclickable, but that seemed a bit uneccessary, since it would've just made it so that the options pane included an "empty" button with no functionality. This felt like it would've been cluttered and a bit misleading so I chose to just remove the button displaying for those elements altogether. Readding the button with the greyed out style and unclickability isn't a difficult thing to do, _just add the button to the else statement with the proper modifications_.

**How the options pane looks now (The Sequence Condition element is identical from the save button down):**

image

There is an issue with the bring to front/send to back buttons in general where it sometimes stops working after sending to back too many times, especially when sending behind a superstate/condition element. The affected element just stops being selectable and trying to send other elements behind it also prevents them from being selectable. 

Sometimes the buttons just don't work in general.

These bugs existed before the changes made in the pull request and weren't part of this issue, I just thought I'd mention them.

### DIFF
--- a/DuggaSys/diagram/draw/options.js
+++ b/DuggaSys/diagram/draw/options.js
@@ -40,7 +40,7 @@ function generateContextProperties() {
         showProperties(true, propSet, menuSet);
         str += drawElementProperties(element);
 
-        // Bring to front / send to back buttons, not shown for UML SuperState or Sequence Loop elements
+        // Bring to front / send to back buttons. Bring to front isn't displayed for UML SuperState or Sequence Loop elements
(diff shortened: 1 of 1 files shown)


a23emisn (approved):
I can verify that the buttons are no longer displaying for the mentioned elements, as described in the PR. I agree on that hiding the buttons instead of leaving them paralyzed is a more rational solution. 

Sequence loop options, before and after (same for UML superstate).



<br><br>

Nicely done! 

Approved for merge!

PR #17994, opened 2025-05-27
G3 Added dots and commas as legal names

f23linan:
Added dots and commas as legal names when editing a section name.

image
Name can now contain . or , and still be valid.


c22isako (approved):
### **Works as expected!** 
Not much more to be said, there doesn't seem to be any problems with the sections having dots and commas. **Good work!** 

---
**Additional issue creation:** (PR owner can ignore)
- There might be some benefits from changing the error-message to include the newly added characters, but at the same time, the "-" wasn't included before. This will be made into its separate issue, so that other error messages might be confirmed as well.
- I found some issues regarding illegal characters within feedback whiles testing, but that's not related to this PR and so will be made its own issue.

Issue #17993, opened 2025-05-27
Remove unused files in endpointDirectory


a23oskme:
These files can be removed: 

- descriptionSearch.php
- nameQuery.php
- parameterQuery.php

These were intended to be used to query the endpoint directory database, but are no longer necessary since we created an UI which has database searching built in instead. 
They are located in the endpointDirectory folder.

PR #17995, opened 2025-05-27
Remove unused files in endpointDirectory #17993

a23oskme:
Removed these files and added them to gitignore. 

- descriptionSearch.php
- nameQuery.php
- parameterQuery.php

They can be removed from gitignore after the weekly merge.


a22wilas (approved):
Looks good, fixes the issue.
Good job!

PR #18012, opened 2025-05-27
Removed "README.md" from the links

### LINKED ISSUE #17998: Remove README.md from links inside the README file
In step 12 under `Run containers in Docker` there are two links `README.md#new-installer-steps` and `README.md#old-installer-steps`. When I clicking these links on GitHub, then GitHub will open a page and reload instead of jumping directly to section in the same file. I don't know if there are more these links.

To fix this simply update both links by remove the README.md, like so: `#new-installer-steps` and `#old-installer-steps`, and test if these links works.

Note: This change may conflict with PR #17996, maybe wait until the PR is merged before to fix this.

a23bensh:
Fixes #17998 and stops sending the user to a new "reloaded" page of README.


a23axehe (approved):
Tested in github, works as it should and solves the issue.

PR #18026, opened 2025-05-27
Fixed chevron overlapping load save

a23thefo:
## What I did
Made the chevron have display none when showModal is opened and be set back to display block when its closed

## Demo
2025-05-27_15-59-06-ezgif com-video-to-gif-converter


a23kevre (approved):
Neat fix, I approve!

Issue #17926, opened 2025-05-26
Fix errors in microservice /sectionedService/createGithubCodeExample_ms.php


a23louma:
This file has 19 errors that have to be fixed, most of them related to $pdo, "Expected type 'object'. Found 'null' ". There is also an error related to $exampleid on line 139.

PR #18047, opened 2025-05-28
Fixed errors in microservice /sectionedService/createGithubCodeExample_ms.php

a23axehe:
Solved exampleid error by defining it in function.

Solved pdo error by silencing intelephense.

Fixes issue #17926 .


a22linre (commented):
image
image

I also tried to debug this before and the same error still comes up here. I made a test file to send data to createGithubCodeExample_ms.php, and it still returns Missing required POST parameter: cid (see screenshots).

The code looks fine , so I’m not sure if the issue is in the code itself or somewhere else (like how the data is passed or something with sessions?). Maybe I did the test wrong or it’s supposed to work some other way. I'm not sure if I should put it as "request change" or not, so I will just leave it as a comment.

PR #18058, opened 2025-05-28
Create a test for setAsActiveCourse_ms.php

### LINKED ISSUE #18032: Create a test for sharedMicroservices/setAsActiveCourse_ms.php

a23bensh:
Fixes #18032. Added a folder for sharedMicroservices and then a file for setAsActiveCourse_ms_test.php. To test this go into tests/microservices/sharedMicroservices/setAsActiveCourse_ms_test.php, and verify that all tests pass.


a22wilas (approved):
The solution works as intended.

Removed the delete query and checked the database.
The result:
bild

The activeversion was updated with the right value from the test service.

Good job!

PR #18059, opened 2025-05-28
Fix initial page on newinstaller/installer.php  #18052

### LINKED ISSUE #18052: Fix initial page on newinstaller/installer.php
For some reason both page 1 and page 7 are visible when first arriving on 'newinstaller.php'. However, it doesn't seem to break the installation itselft and will disappear when going to page 2 (it doesn't reappear when going back to page 1 from page 2 either).

Image

a23oskme:
The page class had `display: flex` when the page loaded which caused page 1 and 7 to be visible and didn't get display: none until the user changes page. Fixes #18052 
I replaced flex with none on the page class and gave page1 `display: flex`. 

Only page 1 should be visible now and the page navigation will still work.

### DIFF
--- a/newinstaller/installer_ui.php
+++ b/newinstaller/installer_ui.php
@@ -20,7 +20,7 @@
 <body>
 
 	<form id="installer_form">
-		<div id="page1" class="page">
+		<div id="page1" class="page" style="display:flex;">
 			<div class="banner">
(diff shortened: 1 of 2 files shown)


a22linre (commented):
I can now only see step 1 page when enter the page. Looks good!

PR #18065, opened 2025-05-28
G3 Fixed the horizontal scroll in sectioned.

f23linan:
Fixed the horizontol scroll that appeared in sectioned while in mobile view.


a23alera (approved):
Everything looks good and works as intended. The horizontal scroll issue is resolved in mobile view.

PR #18072, opened 2025-05-28
Fixed styling and logic of burger menu in mobile

a23yutta:
Moved burger menu dropdown and the icon to the left and fixed some burger menu logic in mobile.


a23fanro:
I found out that this PR made it so the hamburgermeny and coruse-selection 'overlaps'
image

a23fanro:
This PR will be reverted.

Issue #18064, opened 2025-05-28
Remove/change unnecessary code from microservice createNewCodeExample_ms.php


a23louma:
In createNewCodeExample_ms.php there is first a curlService call with callMicroserviceGET to get the username. Then there is another call with callMicroservicePOST, but the $url is never set and I think this this line should just be removed and the following line `$data = json_decode($response, true);` as well.

Also there is a logUserEvent call on line 21 which I think should be removed since there is another call at the end, and on line 21 $sectname isn't set yet.

PR #18098, opened 2025-05-30
Removed unnecessary lines of code in createNewCodeExample_ms.php 

a23axehe:
Removed unnecessary lines of code which were either duplicates or already implemented. Fixes issue #18064


a23bensh (approved):
Cleans up code and still works, looks good!

PR #18099, opened 2025-05-30
XAMPP ssl tutorial

### LINKED ISSUE #17643: Create SSL tutorial for XAMPP
A tutorial on how to integrate the generated .key and .cert file so you can develop with https enabled. 
This is needed for both the microservices and for push-notifications.

a23arvme:
This is a tutorial for integrating the keys into XAMPP.
To test, run newinstaller, then follow the link in the installer to the tutorial. I tested with the latest XAMPP version for Windows, 8.2.12. I noticed that the latest version already has SSL keys enabled. Some versions don't.

NOTE: the link will not work until it is merged into master

fixes #17643

### DIFF
--- a/Shared/Documentation/newinstaller/documentation.md
+++ b/Shared/Documentation/newinstaller/documentation.md
@@ -167,8 +167,16 @@ This page also conatins the options:
 # Options
 This section covers the various options available in the installer.
 
+## Step 6 - SSL
+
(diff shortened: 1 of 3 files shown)


a23gabli (approved):
Looks good. The link doesnt work yet, but as you said it will probably work when merged.

PR #18109, opened 2025-05-30
Removal of whitespace under MOTD

a23kavis:
Added a minimal javascript function that changes margin to adjust for the removal of the motd. 

A secondary approach to this would have been to rebuild the layout of the header and motd, to make them fall after each other in order. That way when motd closes the gap would close by itself.
Before
image

After
SbQyx8TJVg

### DIFF
--- a/DuggaSys/courseed.js
+++ b/DuggaSys/courseed.js
@@ -39,6 +39,10 @@ function closeDeleteForm() {
 	document.getElementById("myForm").style.display = "none";
 }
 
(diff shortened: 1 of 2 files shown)


a23fanro (approved):

a23fanro:
Now when MOTD is reopend after being closed, it hides some of the courseed, but this is acceptablt.

https://github.com/user-attachments/assets/aa006c6e-caf8-4254-8707-e34585d64164

Issue #18089, opened 2025-05-28
Make instruction for function recieveMicroservicePOST


a22wilas:
Needs a clear instruction on how the function recieveMicroservicePOST in curlService.php should be called.
What needs to be added, both to access the file and make a call to the function.

PR #18128, opened 2025-05-30
Update curlService.php

a22wilas:
Fixes issue https://github.com/HGustavs/LenaSYS/issues/18089

Added description for how to use the function recieveMicroservicePOST in curlService.php


a22linre (commented):
On the first row "recieveMicroservicePOST", a small misspelling (recieve -> receive). Idk if it matters that much, if you have time and want to fix, you can but I don't like its important because its just a header. 

Otherwise, looks good and easy to follow!

a22wilas:
Fixed the spelling mistake!

a22linre (commented):
Can't find anything else that needs to be fixed. Looks good!

PR #18139, opened 2025-06-17
Body lock on popup so background cannot be scrolled 

### LINKED ISSUE #18083: (mobile-view) BUG scrollbar visible ontop of popups
### The scrollbar is visible on top of "new course" popup (seen within image1, red marking) 
When having an popup open, such as edit course / new course and scrolling, the scrollbar is visible over the popup. This is mostly apparent whiles in smaller resolutions, such as mobile view, as the popups don't cover the whole sites otherwise.
- It should be kept in the background as it affects the background page, not the popup. 

Image
(image1)

---
### **How to get to functionality:**

Enter courseed.php (main menu) via mobileview -> login as BROM -> press cogwheel (shown in _image2_)
From here, you will be able to see the scrollbar whiles scrolling via mobile view. The issue is mainly encountered at the minimal resolution 320x600.

Image
(_image2_)

a23simry:
I added a function into dugga.js that locks and unlocks the main body. 
Function is called with lockScroll(true) to lock and lockScroll(false) to unlock.
I have added these calls to a few different onclick functions for opening and closing popups.

### DIFF
--- a/DuggaSys/courseed.js
+++ b/DuggaSys/courseed.js
@@ -379,6 +379,7 @@ function createVersion() {
 
 function selectCourse(cid, coursename, coursecode, visi, vers, edvers, gitHubUrl) {
 	console.log("selectCourse() => Selecting course to edit - gitHubUrl:", gitHubUrl || "NO INPUT");
+	lockScroll(true);
 	let items = document.querySelectorAll(".item");
 	items.forEach(item => {
 		item.style.border = "none";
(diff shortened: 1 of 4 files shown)


a23simry:
Most popups use their own onclick function for opening, but a shared one for closing, though not all of them do?? why

a23simry:
There might be more popups to find and lock, would be A LOT nicer if they all used the same open and close functions

a23simry:
I am supposed to review this but can't assign myself as reviewer. 
Code works, it does not lock itself, and does not mess with any other functionality.

PR #18140, opened 2025-07-02
added spacing between burger buttons

### LINKED ISSUE #18079: Unnecessary spacing below dropdown elements via courseed.php
### The dropdown menu found within courseed.php has some unnecessary spacing below its dropdown elements, as shown via _image1_.
- The spacing should be removed, so that unnecessary space isn't taken up whiles in mobileview (and to generally make it look better)

Image
(_image1_)

---
### How to get to functionality:

**The functionality can be viewed via courseed.php, within small resolutions (or mobile view.)**
It is recommended to test via **mobile view,** since the issue is more apparent at the minimal resolution 320x600 
_(which is when the dropdown is visible)_

a23simry:
Decided to add spacing between the buttons instead of removing the extra space, feels like better UX.
image

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -6520,7 +6520,7 @@ only screen and (max-device-width: 568px) {
   }
   
   .navBurgerButt{
-    margin: 0 0 0.4rem 0; 
+    margin: 0 0 0.3rem 0; 
   }
   
   #motdNav {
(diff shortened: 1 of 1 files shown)


a23simry:
reviewing my own code. its a small style.css addition, cant find it changes or messes with anything else

Issue #18141, opened 2025-07-04
Selected items bunch up when dragged with box selection tool


a23gabli:
When multiple items are selected with the box selection tool and then dragged, they all bunch up under the cursor and lose their relative position to each other.

PR #18142, opened 2025-07-11
Relative positions of multiple elements are now maintained when dragged.

a23gabli:
You can test this by simply creating multiple elements in diagram.php, selecting them all with the box selection tool, and dragging them to a different location.
Solves issue #18141


a23gabli (commented):
The solution works as described, the relative positions of elements is kept when selected using the box selection tool and dragged.

Issue #18143, opened 2025-07-27
Self Call button retains selection indicator when switching tools


a23gabli:
When selecting the "Self Call" button, the visual selection indicator does not get removed when switching to another tool unlike the other buttons.

<img width="354" height="382" alt="Image" src="https://github.com/user-attachments/assets/91101834-9175-4cd3-9cc6-779e7c0e06de" />

PR #18144, opened 2025-07-27
Self Call button now properly clears selection indicator on tool switch

a23gabli:
Solves #18143


a23gabli (commented):
Works as intended, the self call button no longer remains (visually) selected when switching to other placement types.

PR #18149, opened 2025-08-20
18085 Codeviwer heading cut off on mobile

a23kavis:
moved mobile heading to reposition it outside of navbar, since it doesnt fit in the navbar on mobile width. This makes use of pervious empty white space and ensures long filenames will be displayed properly. See attached images for before/after

Before
image

After
image


a23kavis:
Upon testing normal behaviour the title seems to remain in the navbar as previously implemented
image


Mobile header seems to work as intended 
image

PR #18150, opened 2025-08-20
18088 Make edges of burgerbutton clickable

### LINKED ISSUE #18088: (mobile-view) BUG codeviewer.php edges of burger button can be pressed without dropdown
#### Issue Description
This issue was found in JS-Test template 1 for Testing-Course (codeviewer.php) while not logged in (login-state doesn't seem to affect issue). Essentially, when pressing the edges of the burger button: 
Image
it can be pressed (and shows it being pressed) but not actually do anything as seen in the following clip:

https://github.com/user-attachments/assets/068ddb44-8d8a-4de3-918b-de7b10ca3fdb

a23kavis:
As stated in issue https://github.com/HGustavs/LenaSYS/issues/18088 the burgerbutton is clickable but does not open the menu. The reason for this being the logic of a table-cell display format. Adding some padding makes sure the whole area is clickable, and triggers the onclick also.

chrome_p1ELUD86O6

### DIFF
--- a/Shared/css/style.css
+++ b/Shared/css/style.css
@@ -1835,7 +1835,15 @@ div .navButt:hover {
 
 #codeBurger {
   display: table-cell;
+  padding: 8px; /* Required for full click coverage, pls dont remove */
+  cursor: pointer;
+  text-align: center;
+  vertical-align: middle;
 }
+#codeBurger img {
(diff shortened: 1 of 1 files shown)


a23kavis:
Upon testing this is a simple solution with maybe a few pixels too much padding, but whole area is clickable where user would expect it to.

PR #18151, opened 2025-08-20
18075 Cog wheels different colours

a23kavis:
After figuring out the logic for the svg filters on the cogwheels, i have applied it to the cog wheel that manages motd. It is now matching the rest.

image


a23kavis:
Solution is simple css and works in dark mode too. Ready for merge
image

PR #18153, opened 2025-08-20
18093 Diagram SelfCall CSS Fix

### LINKED ISSUE #18093: Diagram SelfCall CSS fix
The SelfCall button in the desktop menu does not lose it's selected color when its no longer selected.
Make it so it works like the others.

https://github.com/user-attachments/assets/038b1bb2-c2bb-4fea-b387-da1e1b56ba5a

a23kavis:
The reason for the selfcall not losing its active class is due to how active class is being removed in mouseMode.js

image

The only purpose of that class is to check and uncheck active mode.
So adding that class to the SelfCall element solves this issue
chrome_G4vKDfTE3T

### DIFF
--- a/DuggaSys/diagram.php
+++ b/DuggaSys/diagram.php
@@ -629,7 +629,7 @@
         <!-- Self Call -->
         <fieldset id="localLoadField">
             <legend aria-hidden="true">Self Call</legend>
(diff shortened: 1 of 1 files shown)


a23kavis:
I have tested the diagram and what this solution possibly may have affected but can not find any issues.

PR #18154, opened 2025-08-22
17950 Fix darkmode in SVG Dugga

### LINKED ISSUE #17950: Fix darkmode in duggor: SVG Dugga
Image

Darkmode needs some fixing. Most likely needs some changes in the darktheme stylesheet.

a23kavis:
Currently the example SVG Dugga in testing course is brigh white due to the css being written straight into the stylesheet and no variant in the blacktheme.css file. 
 I have added styling to make sure the svg editor also toggles theme on theme swap, and tried to keep the colour scheme as best as possible.
 
chrome_DppOQE2fgM

### DIFF
--- a/Shared/css/blackTheme.css
+++ b/Shared/css/blackTheme.css
@@ -1466,4 +1466,34 @@ border-left: 14px solid var(--color-sectioned-table-hi);
 .confirm { 
     background-color: #e9e9e9;
 }
+
+
+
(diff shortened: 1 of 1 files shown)


a23kavis:
The affected classes dont appear anywhere else in the project so I believe nothing else is possibly broken. Only suggestion I have would be to add syntax highlighting, but that is a new issue in itself.

PR #18155, opened 2025-08-22
#18077 Buttons in Dugga 5 are not darkmode 

### LINKED ISSUE #18077: (mobile-view) BUG Buttons in Dugga 5 are not darkmode
The buttons used to add vertices and triangles are not being displayed in darkmode.
Image
Image

a23kavis:
Added styling to make the darkmode version easier on the eyes and match the rest of the site. A lot of the html in this page is missing classes and proper styling, so had to use some alternative selectors. 

chrome_y3KcLLYGkv


a23kavis:
Works good, only potential thing i could think to fix is the small reposition arrows to fully match darkmode.

PR #18156, opened 2025-08-22
18094 Selfcall description implementation

### LINKED ISSUE #18094: SelfCall description implementation
The SelfCall button in the desktop version do not have a description when hovered over like the others.
Add it using the same way and classes used in the other buttons. 

https://github.com/user-attachments/assets/c99d1384-46b6-453f-878a-cb652f4dd0ca

a23kavis:
Selfcall was missing a description on hover, and was missing attributes for it as well. I suspect it was added after tooltips were set up and was overlooked. I have now added it however it is missing a keybind as i was unaware of what keys were available still, may be looked into for another issue.

image

### DIFF
--- a/DuggaSys/diagram.php
+++ b/DuggaSys/diagram.php
@@ -629,7 +629,7 @@
         <!-- Self Call -->
         <fieldset id="localLoadField">
             <legend aria-hidden="true">Self Call</legend>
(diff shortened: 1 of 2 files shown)


a23kavis:
Simple solution, works great but is missing a keybind.

